PR #7902, opened 2019-04-29
Hide getButton method for TextArea and Lookup

alexander-kotov-dx:


RomanTsukanov (dismissed):

mrjohn0011 (approved):

IlyaKhD:
LGTM
Note that the method won't be hidden in TS declarations

IlyaKhD (approved):

PR #7911, opened 2019-04-30
Update API definitions

Romzay:


Romzay:
It's impossible to name an object with the "Options" suffix ("Options" is reserved for Widget options objects). The declaration tool generates an error in this case.

maxerokh (approved):

PR #7934, opened 2019-05-06
Fix cellData calculation in Scheduler when groupByDate = true

Yarovaya:


testing/tests/DevExpress.ui.widgets.scheduler/workSpace.tests.js
```diff
@@ -530,6 +530,35 @@ QUnit.testStart(function() {
         });
     });
 
+    QUnit.test("Date table cells shoud have right cellData, groupByDate = true without groups", function(assert) {
```
Olexi: Date table cells **should**

Olexi (changes requested):
One of tests has misspelling in naming

testing/tests/DevExpress.ui.widgets.scheduler/workSpace.tests.js
Yarovaya: fixed

Olexi (approved):

jsdmitry (approved):

PR #7944, opened 2019-05-06
Fix toolbox dragging in scroll box + refactor left panel options

Romzay:

### DIFF
--- a/js/ui/diagram/ui.diagram.js
+++ b/js/ui/diagram/ui.diagram.js
@@ -12,7 +12,6 @@ import DiagramContextMenu from "./ui.diagram.contextmenu";
 import NodesOption from "./ui.diagram.nodes";
 import EdgesOptions from "./ui.diagram.edges";
 import { getDiagram } from "./diagram_importer";
-import { getWindow } from "../../core/utils/window";
 import { hasWindow } from "../../core/utils/window";
 
 const DIAGRAM_CLASS = "dx-diagram";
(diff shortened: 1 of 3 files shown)


js/ui/diagram/ui.diagram.js
```diff
@@ -672,6 +675,14 @@ class Diagram extends Widget {
             this._nodesOption.remove(key, callback);
         }
     }
+    _raiseToolboxDragStart() {
+        const body = getWindow().document.body;
```
romanresh: I'd check the server mode via the `hasWindow()` method because a diagram's instance is initialized in server mode as well.

romanresh: Do we really need document.body? Can we add the class on the toolbox container (this._leftPanel.$element())?

js/ui/diagram/ui.diagram.leftpanel.js
```diff
@@ -27,15 +29,9 @@ class DiagramLeftPanel extends Widget {
...
-    _getDataSources() {
-        return this.option("dataSources") || {};
-    }
-    _getCustomShapes() {
-        return this.option("customShapes") || [];
-    }
     _getAccordionDataSource() {
         var result = [];
-        var categories = ShapeCategories.load(this._getCustomShapes().length > 0);
+        var categories = ShapeCategories.load(this._customShapes.length > 0);
```
romanresh, line 34: Can we use the options("shapes") itself instead of caching its value in the instance._shapes? Its value is already cached in the options.

js/ui/diagram/ui.diagram.js
```diff
@@ -672,6 +675,14 @@ class Diagram extends Widget {
...
+        body.className += " dx-skip-gesture-event";
+    }
+    _raiseToolboxDragEnd() {
+        const body = getWindow().document.body;
+        body.className = body.className.replace(" dx-skip-gesture-event", "");
```
romanresh: Try to replace it with the toggleClass() method. Its available in dxt/jquery

romanresh (approved):

PR #7972, opened 2019-05-07
FileManager - Fix dialogs sizes on different screen sizes

vladkovl:


testing/tests/DevExpress.ui.widgets/fileManagerParts/adaptivity.tests.js
```diff
@@ -80,4 +88,24 @@ QUnit.module("Adaptivity", moduleConfig, () => {
...
     });
 
+    test("dialog size corrent on different window size", function(assert) {
+        this.wrapper.getToolbarButton("Copy").trigger("dxclick");
+        this.clock.tick(400);
+
+        let $dialog = $(".dx-filemanager-dialog-folder-chooser:visible");
+        assert.equal($dialog.length, 1, "dialog is shown");
+
+        const dialogWidth = $dialog.get(0).offsetWidth;
```
maxerokh, line 98: Better way to use the $dialog.offset().width construction

maxerokh (changes requested):

testing/tests/DevExpress.ui.widgets/fileManagerParts/adaptivity.tests.js
vladkovl: This approach will not work, because we redefine `jQuery.width()` function (`renderer.fn.width`). Therefore, we can use only the native functions of the DOM element.

maxerokh (approved):

PR #7994, opened 2019-05-08
List: Fix keyboard navigation using down key if select all checkbox is focused in header filter of DataGrid (T739247)

vconst:


DokaRus (approved):
LGTM

AntonSermyazhko (approved):

PR #8030, opened 2019-05-14
Fix position of input in tagbox (T739332)

babich-a:

### DIFF
--- a/styles/widgets/material/tagBox.material.less
+++ b/styles/widgets/material/tagBox.material.less
@@ -8,6 +8,7 @@
     @MATERIAL_TAGBOX_SELECT_ALL_MARGIN_BOTTOM: 3px;
     @MATERIAL_TAGBOX_TAG_MARGIN: 8px;
     @MATERIAL_TAGBOX_TAG_CONTAINER_PADDING: 4px 12px 12px 12px;
+    @MATERIAL_TAGBOX_TAG_HEIGHT: 32px;
 }
 .dx-size-compact() {
     @MATERIAL_TAGBOX_TAG_CONTENT_PADDING_LEFT: 8px;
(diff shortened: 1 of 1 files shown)


alexander-kotov-dx (approved):

hakimio:
@babich-a 
Just some thoughts. Current styling logic seems a little over-complicated and negative margin-top for the tag-container is causing the field to shift up when tags are added or user starts typing. 
Here is a bit simpler fix I would suggest:

- Remove negative margin-top from tag-container.
- Change tag margins to the following values (equal top and bottom margins):
```css
.dx-tag {
    margin: 4px 8px 4px 0;
}
```
- Tag content padding to the following (equal top and bottom padding):
```css
.dx-tag-content {
    padding: 8px 32px 8px 12px;
}
```
- Tagbox input to this (set height to auto and switch top and bottom paddings):
```css
.dx-texteditor-input {
    height: auto;
    padding: 6px 0 7px;
}
```
Basically the important part is "height: auto" for the input and removing negative margins for the tag-container.
This seems to work quite nicely in Chrome. Are there any issues in IE or FF? Why was it necessary to add the negative margin to tag-container.

EDIT: I can see that the negative margin was part of fix for T736668, but with my solution the clear button still stays aligned correctly even without the negative margin hack.

babich-a:
Hi, @hakimio 
Feel free to make a PR.
I have tested your solution and found some issues.
1. The text when `searchEnabled:true` is not aligned with the tags.
2. The height of the TagBox, when it is empty and the height when at least one of the items is selected will be different (the fix for [T736668](https://www.devexpress.com/Support/Center/Question/Details/T736668/tagbox-the-clear-icon-is-misaligned-if-the-material-blue-light-theme-is-used) will be broken).
3. A validation icon is misaligned.
4. TagBox with `stylingMode: "filled"` or `stylingMode: "outlined"` will be broken.

hakimio:
@babich-a Here is an example without those issues:
https://codepen.io/hakimio/pen/mYBNQa
Anyway, it's just a simple prototype with only one goal: to show that the styling could be simplified.

Btw, doesn't your solution cause the tagbox to go up when user enters text or selects some tags?

babich-a:
@hakimio now your solution contains more code and has negative margins too. It doesn't look like simpler solution anymore. Is there any issue related to negative margins you faced? If so, please provide us an example here on in Support Center.

PR #8039, opened 2019-05-14
DataGrid - filter panel icon element outline should have correct width

AntonSermyazhko:


vconst (changes requested):
Several commits are mixed in this PR

vconst (approved):

PR #8062, opened 2019-05-16
Use 'generic' theme for 'android' and 'win' devices

AlekseyMartynov:
Reason: [BC4928](https://www.devexpress.com/Support/Center/VersionHistory?TechnologyName=DevExtreme&StartBuildName=19.1.3#BC4928), 'Mobile device-specific themes' paragraph

### DIFF
--- a/js/ui/themes.js
+++ b/js/ui/themes.js
@@ -241,9 +241,12 @@ function current(options) {
 function themeNameFromDevice(device) {
     var themeName = device.platform;
 
-
-    if(themeName === "ios") {
-        themeName += "7";
+    switch(themeName) {
+        case "ios":
+            return "ios7";
+        case "android":
+        case "win":
+            return "generic";
     }
 
     return themeName;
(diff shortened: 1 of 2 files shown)


San4es (approved):
LGTM

AlekseyMartynov:
@BingoRUS highlighed another broken case:

```html
<link rel="stylesheet" type="text/css" href="css/dx.common.css" />
<link rel="dx-theme" data-theme="material.blue.light" data-active="true" href="css/dx.material.blue.light.css" />
<link rel="dx-theme" data-theme="material.blue.dark" data-active="false" href="css/dx.material.blue.dark.css" />
```

I'm interested in your opinions. Do you think this should be fixed? If yes -- in this PR or separately?

BingoRUS:
I would say it should be fixed but in separate PR. When we try to find an appropriate theme for a device and fail, we need to use an active theme (data-active attribute is true) or the first theme from the list.

PR #8112, opened 2019-05-20
Fix datebox test for IE11 and mobile devices

alexander-kotov-dx:


DokaRus (approved):

mrjohn0011 (changes requested):
Convert all file to the ES6

mrjohn0011 (approved):

PR #8114, opened 2019-05-21
Add api for waiting web fonts (T736793)

babich-a:


js/ui/themes.js
```diff
@@ -328,6 +329,60 @@ function checkThemeDeprecation() {
...
+    const testElement = document.createElement("span");
+
+    testElement.style.position = "absolute";
+    testElement.style.top = "-9999px";
+    testElement.style.left = "-9999px";
+    testElement.style.visibility = "hidden";
+    testElement.style.fontFamily = etalonFont;
+    testElement.style.fontSize = "250px";
+    testElement.style.fontWeight = fontWeight;
+    testElement.innerHTML = text;
```
BingoRUS, line 349: Why can’t we always use the same text?

babich-a: Google font is divided by unicode-range. 'm' can be loaded while 'м' is not

BingoRUS: That's awful 😑

babich-a: They do it now with its own guidelines :) https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization?hl=en#optimization_checklist

jsdmitry (approved):

groshenkovamarina (approved):

PR #8128, opened 2019-05-22
Don't enlarge dxDropDownButton padding to 10px when splitButton is true

mrjohn0011:
image

**GENERIC**

image

**MATERIAL**

image


DokaRus (changes requested):
It looks weird with the RTL mode

```
DevExpress.config({ rtlEnabled: true });
```
image

jsdmitry (approved):

EugeniyKiyashko (approved):

DokaRus (approved):

PR #8149, opened 2019-05-24
Don't hide custom editor buttons in readonly state (T743776)

mrjohn0011:
This PR adds following behavior:

1. If disabled state of the button is not specified by a user, the CUSTOM button will be disabled when the `readOnly` option is `true` and enabled otherwise.

2. If disabled state is specified by the user, the CUSTOM button ignores the `readOnly` option and gets the disabled state from the user defined one.

The user can change disabled state of the button in run time with the `optionChanged` event and the `getButton` method.

3. Predefined buttons are always invisible in readOnly state

### DIFF
--- a/js/ui/button.js
+++ b/js/ui/button.js
@@ -347,7 +347,7 @@ var Button = Widget.inherit({
     _renderClick: function() {
         var that = this,
             eventName = eventUtils.addNamespace(clickEvent.name, this.NAME),
-            actionConfig = { };
+            actionConfig = { excludeValidators: ["readOnly"] };
 
         if(this.option("useSubmitBehavior")) {
             actionConfig.afterExecute = function(e) {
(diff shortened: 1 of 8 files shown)


DokaRus:
This PR will allow focusing or hovering editor buttons but will phohibit clicks. I think we should solve a whole issue at once.
image

alexander-kotov-dx:
I think custom buttons should be visible but have a readOnly/disabled state by default. And the user needs to have the possibility to change this behavior (by the onOptionChanged handler or by direct button instance access)

DokaRus (dismissed):

alexander-kotov-dx (dismissed):

mrjohn0011:
Waiting for Designer's review

RomanTsukanov (commented):
The logic is ok

DokaRus (approved):

PR #8172, opened 2019-05-27
Hide vertical scroll bar in auto resize mode (T740645)

gvrnmntdept:

### DIFF
--- a/js/ui/text_area.js
+++ b/js/ui/text_area.js
@@ -8,7 +8,8 @@ var $ = require("../core/renderer"),
     TextBox = require("./text_box");
 
 var TEXTAREA_CLASS = "dx-textarea",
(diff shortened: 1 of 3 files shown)


js/ui/text_area.js
```diff
@@ -123,6 +124,8 @@ var TextArea = TextBox.inherit({
     _createInput: function() {
         var $input = $("<textarea>");
```
mrjohn0011, line 125: This value can be stored in `this._$input` in the editor.js
We can remove find operator in this._input function and use the cached value

mrjohn0011 (approved):
The code may be refactored, performance may be improved (Non blocking changes)

DokaRus (approved):

PR #8223, opened 2019-05-29
Update SelectBox behavior after DataSource change (T728469)

alexander-kotov-dx:

### DIFF
--- a/js/ui/select_box.js
+++ b/js/ui/select_box.js
@@ -451,10 +451,24 @@ var SelectBox = DropDownList.inherit({
         return result;
     },
 
+    _isOldDataSourceValue: function() {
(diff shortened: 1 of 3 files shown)


js/ui/select_box.js
```diff
@@ -451,12 +451,26 @@ var SelectBox = DropDownList.inherit({
         return result;
     },
 
+    _isOldDataSourceValue: function() {
+        return this.option("text") && this.option("text") === this.option("value");
```
DokaRus: Is it works correctly with the DS contains object items? 
I can define `displayExpr` that represents `text`, and `valueExpr` that represents `value`

alexander-kotov-dx: Thanks. I have check this function and _isOldDataSourceValue is not required

mrjohn0011 (approved):

DokaRus (approved):

PR #8245, opened 2019-05-30
TreeView ignores updating a visible option for items (T735450)

EugeniyKiyashko:


js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -1307,6 +1307,24 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
+        const $node = this._getNodeElement(node);
+        const parentNode = this._dataAdapter.getNodeByKey(node.internalFields.parentKey);
+
+        $node.toggleClass(INVISIBLE_STATE_CLASS, state !== undefined && !state);
+
+        if(parentNode) {
+            const $parentNode = this._getNodeElement(parentNode);
+            const $icon = $parentNode.children(`.${TOGGLE_ITEM_VISIBILITY_CLASS}`);
+
+            if(this._dataAdapter._isAllChildInvisible(parentNode)) {
```
mrjohn0011, line 1320: Use icon.toggle with a parameter instead of show/hide

PR #8261, opened 2019-05-31
dxDiagram - replace selectboxes in the toolbar with selectbuttons (T739191)

romanresh:
The diagram's toolbar contains unnamed selectboxes (without any labels). 
This fix replaces them with selectbuttons. These are the selectboxes where textbox and each item in dropdown is replaced with clear icons.


images/widgets/common/diagram/connector-end-arrow.svg
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 21.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
```
babich-a: Can we remove some garbage from svg, like this comment?

images/widgets/common/diagram/connector-end-arrow.svg
```diff
@@ -0,0 +1,6 @@
...
+<svg version="1.1" id="_x31_" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
```
babich-a: Do we need 'id' here and in the 'path'?

styles/widgets/base/diagram.less
```diff
@@ -0,0 +1,32 @@
...
+    
+    .dx-diagram-i-connector-straight {
+        .diagram-icon-colored2(data-uri('image/svg+xml;charset=UTF-8', 'images/widgets/common/diagram/connector-straight.svg'), @new-color, 32px, 22px);
+    }
+}
+
+.diagram-icon-colored2(@data-uri, @current-color, @width, @height) {
+    @uri: replace("@{data-uri}", "currentColor", escape(@current-color));
+    background-image: e(@uri);
+    width: @width;
```
babich-a, line 30: It seems we can use 32px for width and 22px for height and decrease the number of mixin parameters.

babich-a (dismissed):
Check that all files have a newline at the end (connector-straight.svg, styles/widgets/base/diagram.less, styles/widgets/common/diagram.less, styles/widgets/generic/diagram.generic.less, styles/widgets/material/diagram.material.less).

images/widgets/common/diagram/connector-end-arrow.svg
romanresh: done

images/widgets/common/diagram/connector-end-arrow.svg
romanresh: removed. it was auto-generated by illustrator

styles/widgets/base/diagram.less
romanresh: This parameter will be needed for the next feature (replace the colorboxes with colorbuttons) because image size of color buttons is different.

Romzay (approved):

PR #8300, opened 2019-06-05
DataGrid - Should fire onKeyDown event for each keyboard supported element

igormalzew:


AntonSermyazhko (dismissed):

js/ui/grid_core/ui.grid_core.pager.js
```diff
@@ -43,6 +43,7 @@ var PagerView = modules.View.inherit({
             pagerOptions = that.option("pager") || {},
             dataController = that.getController("data"),
             options = {
+                parent: that,
```
AntonSermyazhko: This is an extra code

AntonSermyazhko (approved):

PR #8318, opened 2019-06-06
Save SelectBox input text after clear of the dataSource (T728469)

alexander-kotov-dx:

### DIFF
--- a/js/ui/drop_down_editor/ui.drop_down_list.js
+++ b/js/ui/drop_down_editor/ui.drop_down_list.js
@@ -449,22 +449,18 @@ var DropDownList = DropDownEditor.inherit({
         this.option("selectedItem", null);
     },
 
-    _processNonexistentItem: function() {
-        this._clearSelectedItem();
-    },
-
     _processDataSourceChanging: function() {
         this._setListDataSource();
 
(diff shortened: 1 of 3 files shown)


js/ui/drop_down_editor/ui.drop_down_list.js
```diff
@@ -449,18 +449,18 @@ var DropDownList = DropDownEditor.inherit({
...
 
-    _processNonexistentItem: function() {
-        this._clearSelectedItem();
-    },
-
     _processDataSourceChanging: function() {
         this._setListDataSource();
 
+        if(this.option("value") && this.option("selectedItem")) {
+            this.oldValue = this.option("selectedItem");
```
DokaRus: Why it without underlining, like a public prop?

js/ui/select_box.js
```diff
@@ -451,14 +451,19 @@ var SelectBox = DropDownList.inherit({
         return result;
     },
 
+    _isOldDataSourceValue: function() {
+        return (this.oldValue && this.option("value") === this._valueGetter(this.oldValue));
```
DokaRus: When `oldValue` resets?

PR #8347, opened 2019-06-10
DataGrid - incorrectly exports group summary values if the column.allowExporting is false (T746913)

EugeniyKiyashko:


IgnatovDan (approved):

js/ui/data_grid/ui.data_grid.export.js
```diff
@@ -374,6 +374,10 @@ exports.ExportController = dataGridCore.ViewController.inherit({}).include(expor
         return result;
     },
 
+    _needColumnExporting: function(column) {
+        return !column.command && (column.allowExporting || column.allowExporting === undefined);
```
jsdmitry, line 378: What if the allowExporting equal the null value? Maybe to use the `typeUtilis.isDefined()`?

jsdmitry (approved):

PR #8350, opened 2019-06-11
DataGrid - cell should preserve focus after Enter key pressed if enterKeyDirection is 'none' and enterKeyAction is 'moveFocus'

AntonSermyazhko:


js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -511,6 +511,10 @@ var KeyboardNavigationController = core.ViewController.inherit({
             rowIndex = this.getVisibleRowIndex(),
             $row = this._focusedView && this._focusedView.getRow(rowIndex);
 
+        if(!this._allowEditingOnEnterKey() && !this._getEnterKeyDirection(eventArgs)) {
```
vconst: Why is expand functionality is skipped in this case?

vconst (approved):

PR #8351, opened 2019-06-11
Diagram - update list editors(like PageSize, GridSize, etc) items via…

Romzay:
… diagram core

### DIFF
--- a/js/ui/diagram/ui.diagram.commands.js
+++ b/js/ui/diagram/ui.diagram.commands.js
@@ -190,14 +190,8 @@ const DiagramCommands = {
                 command: DiagramCommand.PageSize,
                 text: "Page Size",
                 widget: "dxSelectBox",
-                getValue: (v) => {
-                    const vParts = v.replace("[", "").replace("]", "").split(",");
-                    return {
-                        width: parseInt(vParts[0]),
-                        height: parseInt(vParts[1])
-                    };
-                },
-                setValue: (v) => v.toString()
+                getValue: (v) => JSON.parse(v),
+                setValue: (v) => JSON.stringify(v)
             },
             {
                 command: DiagramCommand.PageLandscape,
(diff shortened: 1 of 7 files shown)


js/ui/diagram/ui.diagram.commands.js
```diff
@@ -181,128 +181,56 @@ const DiagramCommands = {
...
+                text: "Units",
+                widget: "dxSelectBox"
+            },
             {
                 command: DiagramCommand.PageSize,
                 text: "Page Size",
                 widget: "dxSelectBox",
                 getValue: (v) => {
-                    const vParts = v.split("|");
+                    const vParts = v.replace("[", "").replace("]", "").split(",");
```
romanresh: I'd suggest to use JSON.parse instead of manually parsing JSON.

styles/widgets/material/diagram.material.less
```diff
@@ -12,6 +12,9 @@
             .dx-accordion-item {
                 border-left: none;
                 border-right: none;
+                .dx-diagram-right-panel-begin-group {
+                    padding-top: 24px;
```
romanresh, line 16: Should this padding be same for both compact and regular material themes? If it's not, you can use some existing constant.

js/ui/diagram/ui.diagram.commands.js
Romzay: dxSelectBox doesn't work with objects (in core we use Size), so for items we use
var value = typeof item.value === "object" ? item.value.toString() : item.value;
And when a values is changed we return an object similar to Size 

                    const vParts = v.replace("[", "").replace("]", "").split(",");
                    return {
                        width: parseInt(vParts[0]),
                        height: parseInt(vParts[1])
                    };

styles/widgets/material/diagram.material.less
Romzay: I tried to find something similar but failed. Could you please give me an example?

romanresh: You don't need a constant if the padding should be same for both compact and regular material themes. Should it be same? 
If it should, it's perfect, keep it as is. If it should not, you can add a new variable in the "size-schemes" folder.

romanresh (approved):

PR #8352, opened 2019-06-11
Remove doT dependency

San4es:
Motivation:
https://github.com/DevExpress/DevExtreme/network/alerts
https://hackerone.com/reports/390929

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "cldrjs": "^0.5.0",
     "del": "^2.2.2",
     "devextreme-internal-tools": "^1.0.3",
-    "dot": "^2.0.0-beta",
     "eslint": "^5.9.0",
     "eslint-plugin-spellcheck": "0.0.11",
     "exceljs": "^1.7.0",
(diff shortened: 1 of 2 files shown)


nightskylark:
@San4es  This is a `devDependency` and it's used only for testing. I can't see how this vulnerability could affect in this case. And I don't think we should test unstable version because of it.

San4es:
We have a private discussion with @nightskylark. His suggestion is to remove the "doT" dependency and tests for it.

vconst (dismissed):

San4es:
@nightskylark @vconst thank you for your feedback!
I am going to merge this PR. We can separate third-party template engines integrations and move them to demos level in future major releases.

nightskylark:
I still think that using beta-version of dependency is a hack solution that is not needed.

nightskylark:
Also, as I can see, the `dot` v 2.0 uses `new Function` too.
https://github.com/olado/doT/blob/v2.0.0/doT.js#L69
https://github.com/olado/doT/blob/v2.0.0/doT.js#L83
https://github.com/olado/doT/blob/v2.0.0/doT.js#L133

nightskylark (approved):

San4es:
@nightskylark Oh, thank you!

>His suggestion is to remove the "doT" dependency and tests for it.

you win! :wink:

PR #8358, opened 2019-06-13
Ignore change of timezone when winter|summer time change (T739951)

Olexi:


Olexi:
Previous behavior was correct, rework is not needed

PR #8365, opened 2019-06-14
File Manager - Validate Move and Copy operations in JSON provider

maxerokh:

### DIFF
--- a/js/ui/file_manager/ui.file_manager.common.js
+++ b/js/ui/file_manager/ui.file_manager.common.js
@@ -1,5 +1,6 @@
 import { when } from "../../core/utils/deferred";
 import { noop } from "../../core/utils/common";
+import typeUtils from "../../core/utils/type";
 
 const ErrorCode = {
     NoAccess: 0,
@@ -13,22 +14,6 @@ const whenSome = function(arg, onSuccess, onError) {
     onSuccess = onSuccess || noop;
     onError = onError || noop;
 
-    const createResult = function(result, success, canceled, error) {
-        return {
-            result: result,
-            success: success,
-            canceled: canceled || false,
-            error: error || null
-        };
-    };
-
-    const createErrorInfo = function(index, result) {
-        return {
-            index: index,
-            errorId: result.errorId,
-        };
-    };
-
     if(!Array.isArray(arg)) {
         arg = [ arg ];
     }
@@ -36,20 +21,15 @@ const whenSome = function(arg, onSuccess, onError) {
     const deferreds = arg.map((item, index) => {
         return when(item)
             .then(
-                result => {
-                    const resultObj = createResult(result, true);
-                    if(resultObj.error && onError) {
-                        onError(createErrorInfo(index, resultObj.error));
-                    }
-
-                    if(resultObj.success && onSuccess) {
-                        onSuccess();
-                    }
+                () => {
+                    typeUtils.isFunction(onSuccess) && onSuccess();
                 },
                 error => {
-                    if(error) {
-                        onError(createErrorInfo(index, error));
+                    if(!error) {
+                        error = { };
                     }
+                    error.index = index;
+                    typeUtils.isFunction(onError) && onError(error);
                 });
     });
 
--- a/js/ui/file_manager/ui.file_manager.common.js
+++ b/js/ui/file_manager/ui.file_manager.common.js
@@ -22,34 +22,30 @@ const whenSome = function(arg, onSuccess, onError) {
     };
 
     if(!Array.isArray(arg)) {
-        return when(arg)
-            .then(onSuccess,
-                error => {
-                    if(error) {
-                        onError(createErrorInfo(0, error));
-                    }
-                });
+        arg = [ arg ];
     }
 
     const deferreds = arg.map((item, index) => {
         return when(item)
-            .then(result => createResult(result, true),
+            .then(
+                result => {
+                    const resultObj = createResult(result, true);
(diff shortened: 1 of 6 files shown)


js/ui/file_manager/ui.file_manager.common.js
```diff
@@ -22,34 +30,31 @@ const whenSome = function(arg, onSuccess, onError) {
...
+        arg = [ arg ];
     }
 
     const deferreds = arg.map((item, index) => {
         return when(item)
-            .then(result => createResult(result, true),
+            .then(
+                result => {
+                    const resultObj = createResult(result, true);
+                    if(resultObj.error && onError) {
```
vladkovl: `resultObj.error` always  is null so the condition also always is null. Can we remove it?

js/ui/file_manager/ui.file_manager.common.js
```diff
@@ -22,34 +30,31 @@ const whenSome = function(arg, onSuccess, onError) {
...
+                        onError(createErrorInfo(index, resultObj.error));
+                    }
+
+                    if(resultObj.success && onSuccess) {
```
vladkovl: `resultObj.success` always is true, so the condition can be changed to `if(onSuccsess)`.
At summary, `resultObj` is not need at all, neither the `createResult()` function. Can we remove any of it?

js/ui/file_manager/ui.file_manager.editing.js
```diff
@@ -179,13 +174,18 @@ class FileManagerEditingControl extends Widget {
...
                     () => this._raiseOnSuccess(action.getSuccessMessage(items), onlyFiles),
-                    info => {
-                        const fileItem = items[info.index];
-                        this._raiseOnError(info.errorId, fileItem);
-                    });
-            });
+                    info => this._onFileProviderError(info, items)
+                );
+            })
+            .fail(info => this._onFileProviderError(info, items));
```
vladkovl: Can we use reject callback function instead of the separate `.fail()`? E.g.:
```
then(result => {
    whenSome(
        result,
        () => this._raiseOnSuccess(action.getSuccessMessage(items), onlyFiles),
        info => this._onFileProviderError(info, items)
	);
},
    info => this._onFileProviderError(info, items)
);
```

js/ui/file_manager/ui.file_manager.files_tree_view.js
```diff
@@ -325,7 +324,7 @@ class FilesTreeViewModel {
             return this._expandAndGetChildrenForGeneralItem(item);
         });
 
-        return whenSome(deferreds)
+        return when.apply(null, deferreds)
```
vladkovl: We need to wait all errors (rejects) before selected item loading, because the errors can correct selected item. E.g. if selected folder is not found, then we need select the parent folder.

vladkovl (changes requested):

js/ui/file_manager/ui.file_manager.common.js
```diff
@@ -22,34 +30,31 @@ const whenSome = function(arg, onSuccess, onError) {
...
+                        onSuccess();
+                    }
+                },
                 error => {
                     if(error) {
                         onError(createErrorInfo(index, error));
                     }
-                    return createResult(null, false, !error, error);
                 });
     });
```
vladkovl, line 34: I found that the new implementation does not return `Deffered ` as a result. Without this, it is impossible to achieve the desired in [this comment](https://github.com/DevExpress/DevExtreme/pull/8365#discussion_r293697464).

maxerokh: Please, review new version of code.

js/ui/file_manager/ui.file_manager.files_tree_view.js
maxerokh: The whenSome method internally call the when.apply(null, deferreds) method. What is the differences?

js/ui/file_manager/ui.file_manager.common.js
maxerokh: Fixed

js/ui/file_manager/ui.file_manager.editing.js
maxerokh: Fixed

js/ui/file_manager/ui.file_manager.files_tree_view.js
vladkovl: The difference is as follows: `when()` has a fail-fast behavior, while` whenSome()` has a fail-last behavior.
Actions that are performed after the function `whenSome()` in `_expandAndGetChildrenForRootItem()` method should occur after all the preceding actions are completed, regardless of the status of their completion (success or error). This is necessary in order to keep the selected folder up to date.

js/ui/file_manager/ui.file_manager.common.js
vladkovl: It's good.

js/ui/file_manager/ui.file_manager.files_tree_view.js
maxerokh: Ok, I have reverted the code.

vladkovl: A more detailed description of the fail-last approach is in[ this article](https://nmaggioni.xyz/2016/10/13/Avoiding-Promise-all-fail-fast-behavior/).

vladkovl (approved):

PR #8375, opened 2019-06-14
DataGrid - should focus on editing cell with enabled focused row (T747484)

igormalzew:


styles/widgets/generic/gridBase.generic.less
```diff
@@ -618,12 +618,12 @@
             border-top: @GENERIC_GRID_BASE_ROW_BORDER;
             border-bottom: @GENERIC_GRID_BASE_ROW_BORDER;
 
-            &.dx-edit-row:first-child > td {
+            &.dx-edit-row:first-child > td:not(.dx-command-edit) {
```
vconst: Making selectors stronger is not good to fix bug

styles/widgets/generic/gridBase.generic.less
```diff
@@ -737,6 +737,16 @@
...
 
+            & .dx-command-edit {
+                background-color: ~"@{@{widgetName}-row-focused-bg}";
+                border-top: 1px solid ~"@{@{widgetName}-row-focused-bg}";
+                border-bottom: 1px solid ~"@{@{widgetName}-row-focused-bg}";
+
+                & .dx-link {
+                    color: ~"@{@{widgetName}-row-focused-color}";
+                }
+            }
```
vconst, line 740: This styles is already defined in another place. So, this new styles are duplicated.

vconst (changes requested):

PR #8387, opened 2019-06-17
Fix pivotGrid rendering to container with css transformation (T749971)

roman-simionov:


vconst (approved):

fiantsev:
@roman-simionov @vconst guys, does this solution work for scaling > 1.0 e.g.: transform: scale(1.8) ?

roman-simionov:
Hi @fiantsev,
 
Yes, this solution will work for any scale value.

PR #8404, opened 2019-06-18
Extender recursive stack of doom fix.

branc116:


San4es:
Please clarify what scenario you want to fix.

branc116:
The object named "source" had property called "default" and property "default" was a reference to a property "source", and, this would keep extending itself forever, or until the stack overflows.
This is the error:
image

San4es:
Can you use non-deep extending in your case?

branc116:
What is non-deep extending?

branc116:
To clarify, this happens when I include anything from devextreme library. eg. 

import * as D from "devextreme-react/pie-chart";


And it wasn't just for react, it happens whenever I try importing anything from your libraries. 

The quick fix I provided, made everything behave properly.

San4es:
@branc116

Unfortunately, we cannot merge this PR because it breaks a lot of tests:
https://devextreme-ci.devexpress.com/DevExpress/DevExtreme/17023

However, we want to help you solve this issue. Would you please submit a ticket in [our Support Center](https://www.devexpress.com/Support/Center/) and share a simple working example that reproduces the issue?

nightskylark:
I've reproduced the `Maximum call stack size exceeded` exception by extending an object by the object with a circular link.
```
QUnit.test("Extend with circular objects", (assert) => {
    const source = {};
    source.default = source;
    extend(true, { }, source);
    assert.ok(true, "Does not throw an exception");
});
```
But this behavior is normal for the `extend` utility. The _jQuery extend_ behaves the same way.
Can you please provide an example to demonstrate how it affects in your scenario?

PR #8474, opened 2019-06-24
DataSource: Fix map callback was not called after push to store (T751537)

jtoming830:


vconst (approved):

Nisden:
Would you not want to run mapFunc on "update" as well?

jtoming830:
> Would you not want to run mapFunc on "update" as well?

Hello,
In this case, the change is applied to the already mapped item. 
DataSource creates map projections only for new items.

PR #8477, opened 2019-06-24
DataGrid - should focus on editing cell with enabled focused row (T747484)

igormalzew:


styles/widgets/generic/gridBase.generic.less
```diff
@@ -720,16 +720,24 @@
...
+            &.dx-data-row {
+                & > td,
+                & > tr > td,
+                & .dx-command-edit {
                     background-color: ~"@{@{widgetName}-row-focused-bg}";
                     color: ~"@{@{widgetName}-row-focused-color}";
 
-                    & .dx-@{widgetName}-group-opened,
-                    & .dx-@{widgetName}-group-closed {
+                    input {
```
vconst: Using our class dx-texteditor-input is better because our styles for input can broke customer's template

styles/widgets/generic/gridBase.generic.less
```diff
@@ -720,16 +720,24 @@
...
+                        border-radius: 0;
+                    }
+
+                    .dx-link {
                         color: ~"@{@{widgetName}-row-focused-color}";
+
+                        & .dx-@{widgetName}-group-opened,
+                        & .dx-@{widgetName}-group-closed {
+                            color: ~"@{@{widgetName}-row-focused-color}";
```
vconst: Is this block of styles correct? The .dx-@{widgetName}-group-opened and .the dx-@{widgetName}-group-closed classes can not be inside .dx-link

vconst (changes requested):

vconst (approved):

PR #8480, opened 2019-06-24
Stylelint: Ignore all files except *.css and *.less

San4es:
To avoid linting unnecessary files in the stylelint plugin for VS Code:
image

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,6 +1,6 @@
 {
     "extends": "stylelint-config-standard",
-    "ignoreFiles": "**/*.json",
+    "ignoreFiles": "!**/*.{css,less}",
     "rules": {
         "indentation": [4, { ignore: ["inside-parens"] }],
         "max-empty-lines": 2,


nightskylark (approved):

nightskylark:
Does it ignore `node_modules` folder?

San4es:
`node_modules` is ignored in our workspace for VS Code:
https://github.com/DevExpress/DevExtreme/blob/19_1/devextreme.code-workspace#L20

nightskylark:
Does `.stylelintrc` use only in VS Code?

nightskylark:
Now I'm afraid I should not approve so fast :)

San4es:
> Does .stylelintrc use only in VS Code?

`.stylelintrc` also is used in the `lint-css` npm script, but we specify the `styles` folder there:
https://github.com/DevExpress/DevExtreme/blob/19_1/package.json#L118

>Now I'm afraid I should not approve so fast :)

No worries! We can always make a separate PR ;)

PR #8494, opened 2019-06-25
Diagram - focus document after each click on widget (T739199)

pgrivachev:

### DIFF
--- a/js/ui/diagram/ui.diagram.js
+++ b/js/ui/diagram/ui.diagram.js
@@ -16,6 +16,7 @@ import { getDiagram } from "./diagram_importer";
 import { hasWindow, getWindow } from "../../core/utils/window";
 import eventsEngine from "../../events/core/events_engine";
 import eventUtils from "../../events/utils";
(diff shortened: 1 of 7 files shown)


romanresh (commented):
Does it still allow to focus toolbar inputs via mouse? It looks like diagram.core will catch the "mouseup" event and focus a document instead.

js/ui/diagram/ui.diagram.js
```diff
@@ -152,6 +154,13 @@ class Diagram extends Widget {
         });
     }
 
+    _onPanelPointerUp({ event }) {
+        const preventRefocusSelector = ".dx-textbox";
+        if(!$(event.target).closest(preventRefocusSelector).length) {
```
romanresh: I'd move such condition inside panel's code because diagram widget should not know about implementation of its inner  parts. Also, I suggest to extract the selector const out of method to the beginning of a file (near other constants)

romanresh (approved):

PR #8510, opened 2019-06-25
DataGrid - A page scrolls after scrolling in the header filter on MAC (T733974)

EugeniyKiyashko:


js/ui/overlay/ui.overlay.js
```diff
@@ -1103,6 +1106,49 @@ var Overlay = Widget.inherit({
...
+        return (isReachedMax || isReachedMin);
+    },
+
+    _tryGetScrollableContainer(element) {
+        const scrollableClass = `.dx-scrollable-container`;
+
+        if($(element).is(scrollableClass)) {
+            return element;
+        } else {
+            return $(element).parents(scrollableClass).get(0);
```
DokaRus: `parents` always get all ancestors of the element. If you'll need a first matched element, better to use `closest`

js/ui/overlay/ui.overlay.js
```diff
@@ -1103,6 +1106,49 @@ var Overlay = Widget.inherit({
...
+            }
+        }
+
+        return (isReachedMax || isReachedMin);
+    },
+
+    _tryGetScrollableContainer(element) {
+        const scrollableClass = `.dx-scrollable-container`;
+
+        if($(element).is(scrollableClass)) {
```
DokaRus: I think that event-based interaction between widgets will help to implement a more elegant solution.

DokaRus (dismissed):

jsdmitry (dismissed):

IgnatovDan (approved):

PR #8549, opened 2019-06-27
Add dxGantt widget

kruglikovstepan:


maxerokh (dismissed):

build/gulp/vendor.js
```diff
@@ -41,12 +41,15 @@ const JS_VENDORS = [
...
+    },
+    {
+        path: '/devexpress-gantt/dist/dx-@(gantt|gantt.min).js'
     }
 ];
 
 const CSS_VENDORS = [
     {
-        path: '/devexpress-diagram/dist/dx-@(diagram|diagram.min).css'
+        path: '/devexpress-gantt/dist/dx-@(gantt|gantt.min).css'
```
vconst, line 55: Why did diagram replace to gantt?

js/ui/gantt/ui.gantt.js
```diff
@@ -0,0 +1,152 @@
...
+import { extend } from '../../core/utils/extend';
+
+const GANTT_CLASS = 'dx-gantt';
+const GANTT_SPLITTER_CLASS = 'dx-gantt-splitter';
+const GANTT_VIEW_CLASS = 'dx-gantt-view';
+
+class Gantt extends Widget {
+    _init() {
+        super._init();
+    }
```
vconst: There is no logic in this _init overload. Remote it.

js/ui/gantt/ui.gantt.view.js
```diff
@@ -0,0 +1,102 @@
...
+    getGanttTickSize() {
+        return { width: 100, height: 34 };
+    }
+    getGanttContainer() {
+        return this.$element().get(0);
+    }
+    getGanttViewStartDate() {
+        let min = new Date();
+        const tasks = this.getGanttTasksData();
+        tasks.forEach(t => {
```
vconst: Using Array,reduce is better to find min value

ts/dx.all.d.ts
```diff
@@ -2673,6 +2680,22 @@ declare module DevExpress.ui {
...
+    /** @name dxGantt.Options */
+    export interface dxGanttOptions extends WidgetOptions<dxGantt> {
+        /** @name dxGantt.Options.dependenciesDataSource */
+        dependenciesDataSource?: any;
+        /** @name dxGantt.Options.resourceAssignmentsDataSource */
+        resourceAssignmentsDataSource?: any;
+        /** @name dxGantt.Options.resourcesDataSource */
+        resourcesDataSource?: any;
+        /** @name dxGantt.Options.taskDataSource */
+        taskDataSource?: any;
```
vconst: Why is task prefix singular but dependencies, resourceAssignments and resources are plural

vconst: Also, "any" type is not good for data sources.

js/ui/gantt/ui.gantt.view.js
```diff
@@ -0,0 +1,102 @@
...
+            if(t.start < min) {
+                min = t.start;
+            }
+        });
+        return min;
+    }
+    getGanttViewEndDate() {
+        let max = new Date();
+        const tasks = this.getGanttTasksData();
+        tasks.forEach(t => {
```
vconst: Using Array,reduce is better to find max value

js/ui/gantt/ui.gantt.js
```diff
@@ -0,0 +1,152 @@
...
+        });
+    }
+
+    _optionChanged(args) {
+        switch(args.name) {
+            case 'taskDataSource':
+            case 'dependenciesDataSource':
+            case 'resourcesDataSource':
+            case 'resourceAssignmentsDataSource':
+                break;
```
vconst, line 142: Why is no logic here?

js/ui/gantt/ui.gantt.js
```diff
@@ -0,0 +1,152 @@
+import $ from '../../core/renderer';
```
vconst: Use everywhere double quotes because devextreme widgets use them.

js/ui/gantt/ui.gantt.js
```diff
@@ -0,0 +1,152 @@
...
+            scrollable.off('scroll', (e) => { this._onTreeListScroll(e); });
+            scrollable.on('scroll', (e) => { this._onTreeListScroll(e); });
+        }
+    }
+
+    _getDefaultOptions() {
+        return extend(super._getDefaultOptions(), {
+            /**
+            * @name dxGanttOptions.taskDataSource
+            * @type object
```
vconst: The object type is bad for datasource. Mostly our widgets have the following type for datasource:
* @type Array<Object>|DataSource|DataSourceOptions

vconst (changes requested):

testing/tests/DevExpress.ui.widgets/gantt.tests.js
```diff
@@ -0,0 +1,34 @@
...
+        this.instance = this.$element.dxGantt("instance");
+    }
+};
+
+QUnit.module("Gantt", moduleConfig, () => {
+    test("should render treeList", (assert) => {
+        const treeListElement = this.$element.find(TREELIST_SELECTOR);
+        assert.ok(treeListElement.length === 1);
+    });
+});
```
vconst, line 101: Add tests for each Gantt option initialization and for each option changing

vconst (changes requested):

js/ui/gantt/ui.gantt.js
```diff
@@ -0,0 +1,152 @@
...
+            */
+            taskDataSource: null,
+            /**
+            * @name dxGanttOptions.dependenciesDataSource
+            * @type object
+            * @default null
+            */
+            dependenciesDataSource: null,
+            /**
+            * @name dxGanttOptions.resourcesDataSource
```
RomanTsukanov: Do we need that "dataSource" postfix in every option name? IMHO, simple names like `tasks`, `dependencies`, `resources`, and `resourceAssignments` (or `resourceAssignees`, see my comment below) are much better

js/ui/gantt/ui.gantt.js
```diff
@@ -0,0 +1,152 @@
...
+            * @type object
+            * @default null
+            */
+            resourcesDataSource: null,
+            /**
+            * @name dxGanttOptions.resourceAssignmentsDataSource
```
RomanTsukanov: Does this data source contain assignees, like employees?

RomanTsukanov (dismissed):
Doccomments LGTM

js/ui/gantt/ui.gantt.view.js
```diff
@@ -44,36 +44,24 @@ export class GanttView extends Widget {
...
     getGanttViewStartDate() {
-        let min = new Date();
         const tasks = this.getGanttTasksData();
-        tasks.forEach(t => {
-            if(t.start < min) {
-                min = t.start;
-            }
-        });
-        return min;
+        return tasks.reduce((min, t) => t.start < min ? t.start : min, tasks[0].start);
```
vconst: Null ref if tasks.length is 0. May be, use new Date() instead of tasks[0].start

vconst (changes requested):

testing/tests/DevExpress.ui.widgets/gantt.tests.js
kruglikovstepan: Tests will be added after the optionsChanged method implementation

vconst: Are you not using TDD? Tests about options initialization can be added now.

vconst (changes requested):

vconst (approved):

PR #8565, opened 2019-06-28
Toolbar: fix header width recalculation (T752661, T752306, T752400)

viterobk:


testing/tests/DevExpress.ui/themes.tests.js
```diff
@@ -568,17 +568,6 @@ QUnit.module("web font checker", () => {
         assert.equal(diff, 0, "Element's count are the same after method call");
     });
 
-    test("waitWebFont: function reject by timeout if the font is not loaded", (assert) => {
```
babich-a, line 571: I suggest rewriting this test to make sure that promise will be resolved if the font is not loaded.

js/ui/toolbar/ui.toolbar.base.js
```diff
@@ -149,12 +149,38 @@ var ToolbarBase = AsyncCollectionWidget.inherit({
...
+    },
+
     _render: function() {
+        var that = this;
         this.callBase();
         this._renderItemsAsync();
 
         if(themes.isMaterial()) {
-            this._checkWebFontForLabelsLoaded().then(this._dimensionChanged.bind(this));
+            this._waitParentAnimationFinished().then(function() {
```
babich-a: We can use an arrow function. We do not need 'that' in this case. My suggestion is `Promise.all([this._waitParentAnimationFinished(),this._checkWebFontForLabelsLoaded()]).then(this._dimensionChanged.bind(this))`

js/ui/toolbar/ui.toolbar.base.js
```diff
@@ -149,12 +149,38 @@ var ToolbarBase = AsyncCollectionWidget.inherit({
...
     },
 
+    _waitParentAnimationFinished: function() {
+        var $element = this.$element(),
+            timeout = 15;
+        return new Promise(function(resolve) {
+            const check = function() {
+                var animated = true;
+                $element.parents().each(function(_, parent) {
+                    if($(parent).data("dxAnimData")) {
```
babich-a: If somebody rename 'dxAnimData' constant the trouble will happen. I suggest using https://js.devexpress.com/Documentation/ApiReference/Common/utils/fx/Methods/#isAnimatingelement method from 'fx' that do completely the same thing.

js/ui/toolbar/ui.toolbar.base.js
```diff
@@ -149,12 +149,38 @@ var ToolbarBase = AsyncCollectionWidget.inherit({
         this.setAria("role", "toolbar");
     },
 
+    _waitParentAnimationFinished: function() {
+        var $element = this.$element(),
+            timeout = 15;
+        return new Promise(function(resolve) {
```
babich-a: Will modern syntax be better `... Promise(resolve => { ...` ?

js/ui/toolbar/ui.toolbar.base.js
```diff
@@ -149,12 +149,38 @@ var ToolbarBase = AsyncCollectionWidget.inherit({
         this.setAria("role", "toolbar");
     },
 
+    _waitParentAnimationFinished: function() {
+        var $element = this.$element(),
```
babich-a: const/let instead of var?

babich-a (changes requested):

babich-a (approved):

BingoRUS (approved):

hakimio:
@viterobk 
Using the latest build (19.1.5-build-19191-0321) the issue doesn't seem to be fixed. 
Demo -> https://codesandbox.io/s/popup-editing-devextreme-data-grid-rzpy4
Click the edit button to open edit dialog.

viterobk:
Hello @hakimio 
Thank you for the feedback. We reproduced this issue and [fixed](https://github.com/DevExpress/DevExtreme/pull/8903) it.

PR #8568, opened 2019-06-28
Scheduler: Changing drop target calculation when dragging appointment

vsimonaitis:
The issue derived from support ticket https://www.devexpress.com/Support/Center/Question/Details/T728235/when-dragging-events-in-timelinemonth-boxes-shifts

Problem
In current implementation, when dragging events in scheduler, dropping point is always under mouse cursor. With small draggable items it is sufficient to drop under cursor, but when event box is wide, drop area under mouse cursor becomes unintuitive.

To reproduce
See demo https://codepen.io/anon/pen/RdXqJv
When grabbing a multi-day event (3 days or more) in scheduler month view grab the bottom-right corner. 
Slightly move the mouse (mouse may stay in the same cell area) and release it. Notice that events start day is shifted by events initial duration. What user would expect is that accidental minor drag would not change the event and drop it to the same date cell. 

Fix
The fix changes calculation of dropping point for scheduler event boxes. Instead of dropping box under the mouse cursor, the item is dropped under its top-left edge position. This is way more intuitive for the users.

### DIFF
--- a/js/ui/scheduler/ui.scheduler.appointments.js
+++ b/js/ui/scheduler/ui.scheduler.appointments.js
@@ -790,6 +790,11 @@ var SchedulerAppointments = CollectionWidget.inherit({
                 });
 
                 translator.move($appointment, coordinates);
+            },
+            adjustDropPosition = function(event, $element) {
+                var offset = $element.offset();
+                event.pageX = offset.left;
+                event.pageY = offset.top;
             };
         this.notifyObserver("getDraggableAppointmentArea", {
             callback: function(result) {
@@ -818,6 +823,7 @@ var SchedulerAppointments = CollectionWidget.inherit({
             },
             onDrag: function(args) {
                 correctCoordinates(args.element);
+                adjustDropPosition(args.event, args.element);
             },
             onDragEnd: function(args) {
                 correctCoordinates(args.element, true);
--- a/js/ui/scheduler/ui.scheduler.appointments.js
+++ b/js/ui/scheduler/ui.scheduler.appointments.js
@@ -791,8 +791,8 @@ var SchedulerAppointments = CollectionWidget.inherit({
 
                 translator.move($appointment, coordinates);
             },
-            adjustDropPosition = function(event, $element) {
-                var offset = $element.offset();
+            adjustDropPosition = function(event, element) {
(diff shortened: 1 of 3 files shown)


andrewmakarov (commented):
Your request breaks existing tests and doesn't contain any new tests. So it's not clear how it works in other existing scenarios. We cannot accept it in this format.

testing/tests/DevExpress.ui.widgets.scheduler/appointments.tests.js
```diff
@@ -543,7 +543,7 @@ QUnit.test("Allday appointment should stay in allDayContainer after small draggi
     var $appointment = $("#allDayContainer .dx-scheduler-appointment"),
         pointer = pointerMock($appointment).start();
 
-    pointer.dragStart().drag(0, -30);
+    pointer.dragStart().drag(0, 30);
```
vsimonaitis, line 546: Drop target is anchored under top-left corner. Dragging the appointment up would move the appointment to another row, so I had to change the dragging direction.

js/ui/scheduler/ui.scheduler.appointments.js
```diff
@@ -818,6 +823,7 @@ var SchedulerAppointments = CollectionWidget.inherit({
             },
             onDrag: function(args) {
                 correctCoordinates(args.element);
+                adjustDropPosition(args.event, args.element);
```
vsimonaitis, line 826: Adjusting drop target anchor by the position of top-left corner of appointment being dragged.

vsimonaitis:
I have changed the code and fixed the UI tests. The drag and drop behavior change is now localized to Scheduler only and is not affecting other generic drag and drop functionality.

vsimonaitis:
@andrewmakarov Please review my last changes

andrewmakarov:
> @andrewmakarov Please review my last changes

Hi,

I reviewed your code and found that the suggested fix won't work in certain scenarios. We're going to address this issue in  a new pull request that should cover all problems. I'll make an update here once it's ready.

andrewmakarov:
> @andrewmakarov Please review my last changes

We're going to address this issue #9617 :
I'm closing this request

PR #8591, opened 2019-06-28
Resolve https://github.com/DevExpress/DevExtreme/issues/8552

### LINKED ISSUE #8552: compileGetter create incorrect getter for array of nested properties
### Steps to Reproduce:

1. Go to https://codepen.io/anon/pen/QXOgQL 
Example code:
```
var getter = DevExpress.data.utils.compileGetter(["B.Id",  "B.Key"]);
var obj =  { B: { Id:"B", Key: "key"}};
var idObj = getter(obj);
var idStr = JSON.stringify(idObj);
if(idObj.B.Id !== "B"){
  console.log("idObj.B.Id is incorrect. idObj.B.Id has to be 'B', but was", idObj.B.Id)
}
console.log("idObj -", idStr);
```
2. Check that function DevExpress.data.utils.compileGetter create getter and when getter executed part of nested properties in result has been lost.
3. I assume that error occurs because of line
`current = current[path[i]] = {}`
in function combineGetters, that line would overwrite part of result object
i think solution would be:
`current = current[path[i]] || (current[path[i]] = {});`

AlekseyMartynov:
Resolves https://github.com/DevExpress/DevExtreme/issues/8552

TODO: pick down to `19_1`, `18_2`

### DIFF
--- a/js/core/utils/data.js
+++ b/js/core/utils/data.js
@@ -127,17 +127,15 @@ var combineGetters = function(getters) {
             path = name.split(".");
             last = path.length - 1;
 
-            for(i = 0; i <= last; i++) {
+            for(i = 0; i < last; i++) {
                 pathItem = path[i];
-                if(i === last) {
-                    current[pathItem] = value;
-                } else {
-                    if(!(pathItem in current)) {
-                        current[pathItem] = { };
-                    }
-                    current = current[pathItem];
+                if(!(pathItem in current)) {
+                    current[pathItem] = { };
                 }
+                current = current[pathItem];
             }
+
+            current[path[last]] = value;
         });
         return result;
     };
--- a/js/core/utils/data.js
+++ b/js/core/utils/data.js
@@ -115,6 +115,7 @@ var combineGetters = function(getters) {
             var value = this(obj, options),
                 current,
                 path,
+                pathItem,
                 last,
                 i;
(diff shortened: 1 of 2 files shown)


nightskylark (dismissed):

js/core/utils/data.js
```diff
@@ -126,11 +127,17 @@ var combineGetters = function(getters) {
             path = name.split(".");
             last = path.length - 1;
 
-            for(i = 0; i < last; i++) {
-                current = current[path[i]] = {};
+            for(i = 0; i <= last; i++) {
```
IlyaKhD: What about moving the assignment outside the loop?
```
path = name.split(".");
last = path.length - 1;
if(last < 0) {
    return;
}

for(i = 0; i < last; i++) {
    pathItem = path[i];
    current = pathItem in current
        ? current[pathItem]
        : current[pathItem] = {};
}

current[path[last]] = value;
```

AlekseyMartynov: Moved.

IlyaKhD (approved):

js/core/utils/data.js
IlyaKhD: :ok_hand:

dxvladislavvolkov (approved):

PR #8607, opened 2019-07-01
DataGrid - The pager's info text displays incorrectly in RTL (T753000)

igormalzew:


js/ui/pager.js
```diff
@@ -520,6 +520,7 @@ var Pager = Widget.inherit({
         if(this.option("showInfo") && typeUtils.isDefined(infoText)) {
             this._$info = $("<div>")
                 .css("display", this._isInfoHide ? "none" : "")
+                .css("direction", this.option('rtlEnabled') ? 'rtl' : 'ltr')
```
vconst: It is better to move this logic to css. Mostly, we use css to customize styles for rtl. If the rtlEnabled option is true, the specific class is added to widget,

vconst (changes requested):

vconst (approved):

PR #8622, opened 2019-07-02
Load next page when downArrow pressed in the dxSelectBox (T750253)

mrjohn0011:


KsyushaM (dismissed):

DokaRus (commented):
I think we can find the best way to fix this. If you apply these changes, then when you open SelectBox, only the items of the current data source page will be rendered and you will not be able to load the items from the previous pages.

KsyushaM (dismissed):

DokaRus (approved):

PR #8642, opened 2019-07-03
Publish valueFormat option in the dxDropDownBox

mrjohn0011:


DokaRus (dismissed):

js/ui/drop_down_box.js
```diff
@@ -137,6 +137,12 @@ var DropDownBox = DropDownEditor.inherit({
              * @default "change"
              */
 
+            /**
+             * @name dxDropDownBoxOptions.valueFormat
+             * @type function(value)
+             * @type_function_param1 value:string|Array<any>
+             * @type_function_return string
+             */
             valueFormat: function(value) {
```
DokaRus: @AlexJameson 
Should we rename this option before publish?
I think the current name can confuse other people. They may decide that this option accept value with the [Format](https://js.devexpress.com/Documentation/ApiReference/Common/Object_Structures/format/) type.
Another point - description will conflict with other [published](https://js.devexpress.com/Documentation/ApiReference/UI_Widgets/dxDataGrid/Configuration/summary/totalItems/#valueFormat) `valueFormat` option which accept  a value with the `Format`type

AlexJameson: I suggest changing the option name to valueFormatter, since we already have formatter function and this name describes the actions quite precisely.

DokaRus (approved):

AlexJameson (approved):

PR #8686, opened 2019-07-05
Fix DateBox resizing after changing the "readOnly" option (T751811) (#8682)

DokaRus:


DokaRus:
Isn't reproduced in 18.2

PR #8743, opened 2019-07-09
File Manager - Detecting uniqueness of file items by key

maxerokh:


js/ui/file_manager/file_items_controller.js
```diff
@@ -0,0 +1,265 @@
...
+
+        const providerDirKey = parentDirectoryInfo.fileItem.isRoot ? "" : parentDirectoryInfo.fileItem.key;
+        loadItemsDeferred = when(this._fileProvider.getItems(providerDirKey))
+            .then(fileItems => {
+                parentDirectoryInfo.items = [ ];
+                for(let i = 0; i < fileItems.length; i++) {
+                    const fileItem = fileItems[i];
+                    const info = fileItem.isDirectory && this._createDirectoryInfo(fileItem, parentDirectoryInfo) || this._createFileInfo(fileItem, parentDirectoryInfo);
+                    parentDirectoryInfo.items.push(info);
+                }
```
vladkovl: You can use `.map()` instead of `for`:
`parentDirectoryInfo.items = fileItems.map(fileItem => fileItem.isDirectory && this._createDirectoryInfo(fileItem, parentDirectoryInfo) || this._createFileInfo(fileItem, parentDirectoryInfo));`

js/ui/file_manager/file_items_controller.js
```diff
@@ -0,0 +1,265 @@
...
+                parentDirectoryInfo.itemsLoaded = true;
+                return parentDirectoryInfo.items;
+            });
+
+        const that = this;
+        this._loadedItems[dirKey] = loadItemsDeferred;
```
vladkovl, line 125: Why are you using `that `instead of `this`? Arrow functions don't change the current context.

js/ui/file_manager/file_items_controller.js
```diff
@@ -0,0 +1,265 @@
...
+        loadItemsDeferred.then(() => {
+            delete that._loadedItems[dirKey];
+        });
+
+        return loadItemsDeferred;
+    }
+
+    createDirectory(parentDirectoryInfo, name) {
+        const that = this;
+        return when(this._fileProvider.createFolder(parentDirectoryInfo.fileItem, name))
```
vladkovl, line 134: You can use `this` in all arrow functions.

js/ui/file_manager/file_provider/file_provider.js
```diff
@@ -45,15 +52,7 @@ class FileProvider {
...
-    getFolders(path) {
-        return this.getItems(path, "folder");
-    }
-
-    getFiles(path) {
-        return this.getItems(path, "file");
-    }
-
-    getItems(path, itemType) {
+    getItems(path) {
```
vladkovl, line 53: I think we should not remove `itemType` parameter, because almost in all providers we can get files separately of directories.

js/ui/file_manager/file_provider/array.js
```diff
@@ -36,40 +37,60 @@ class ArrayFileProvider extends FileProvider {
...
-        item.dataItem.name = name;
+        if(!item) {
+            return;
+        }
+        this._nameSetter(item.dataItem, name);
+        item.name = name;
+
+        let key = this._keyGetter(item.dataItem);
+        if(!key) {
+            key = String(new Guid());
```
vladkovl, line 67: Why not use `_getKeyByDataObject` instead of `Guid()` ?

js/ui/file_manager/file_provider/file_provider.js
```diff
@@ -1,11 +1,13 @@
 import { compileGetter } from "../../../core/utils/data";
-import { pathCombine, getFileExtension, getParentPath, getName } from "../ui.file_manager.utils";
+import { pathCombine, getFileExtension } from "../ui.file_manager.utils";
 import { ensureDefined } from "../../../core/utils/common";
 import { deserializeDate } from "../../../core/utils/date_serialization";
 import { each } from "../../../core/utils/iterator";
 
 const DEFAULT_FILE_UPLOAD_CHUNK_SIZE = 200000;
 
+const FILE_MANAGER_ROOT_DIR_KEY = "__DX-FM-ROOT__";
```
vladkovl: Not reliable way. This value can intersect with custom id value from a custom file provider.

js/ui/file_manager/file_provider/file_provider.js
```diff
@@ -185,5 +188,14 @@ class FileManagerItem {
     }
 }
 
+class FileManagerRootItem extends FileManagerItem {
+    constructor() {
+        super("", "Files", true);
+        this.key = FILE_MANAGER_ROOT_DIR_KEY;
+        this.isRoot = true;
+    }
+}
```
vladkovl, line 194: This class doesn't contain any new logic. I think we can replace it with the factory method.

js/ui/file_manager/ui.file_manager.files_tree_view.js
```diff
@@ -58,26 +49,40 @@ class FileManagerFilesTreeView extends Widget {
 
         this._filesTreeView = this._createComponent($treeView, TreeViewSearch, treeViewOptions);
 
-        eventsEngine.on($treeView, "click", this._raiseClick.bind(this));
+        eventsEngine.on($treeView, "click", treeViewOptions.onItemClick.bind(this));
```
vladkovl: Is it correct? `treeViewOptions.onItemClick` refers to the "onDirectoryClick" action, and I'm not sure we can change the context for this call.

js/ui/file_manager/ui.file_manager.files_tree_view.js
```diff
@@ -122,39 +126,19 @@ class FileManagerFilesTreeView extends Widget {
...
-
-    _updateFocusedElement($element) {
+    _updateFocusedElement() {
+        const directoryInfo = this._getCurrentDirectory();
+        const $element = this._getItemElementByKey(directoryInfo.fileItem.key);
         if(this._$focusedElement) {
             this._$focusedElement.toggleClass(FILE_MANAGER_DIRS_TREE_FOCUSED_ITEM_CLASS, false);
         }
         this._$focusedElement = $element || $();
+        this._$focusedElement.dirInfo = directoryInfo;
```
vladkovl: Can we use .data() function here? For example:
`this._$focusedElement.data("dirInfo", directoryInfo);`

js/ui/file_manager/ui.file_manager.js
```diff
@@ -43,9 +40,11 @@ class FileManager extends Widget {
...
         this._onSelectedFileOpenedAction = this._createActionByOption("onSelectedFileOpened");
 
-        this._provider = this._getFileProvider();
-        this._currentFolder = null;
-
+        this._controller = new FileItemsController({
+            "rootText": "",
+            "fileProvider": this.option("fileProvider"),
+            "onSelectedDirectoryChanged": this._onSelectedDirectoryChanged.bind(this)
+        });
```
vladkovl, line 47: Quotes for options names are redundant.

vladkovl:
I see these major problems:
- there is no support for the command to update the entire tree data with maintaining the state of the expanded folders
- there is no resolution of possible conflicts in the folder structure associated with the lack of folders after loading data
- there is no support for loading the tree data according to the specified initial path
- there is no support for a complete update of the tree data after the editing operation

All this features were in previous version of the code. 
Each of these problems requires solving the following tasks:
- adaptation of the previous version of the code to the new approach implemented in this PR
- refinement of the algorithm for using `key` instead of `relativeName`

maxerokh:
You main remarks notify about new implementation doesn't store expanded state of tree view and file manager doesn't have ability to set custom initial directory.

I think we should create new commit with PR to resolve it with reach set of unit tests. So, previous version didn't have public API with this functionality. We should create the functionality with tech writers cooperation. If we try to solve the problems in the PR, we will solve it partially.

What do you say?

vladkovl:
I think we can address this issues in the future PRs.

vladkovl (approved):

PR #8756, opened 2019-07-10
Toolbar - ComboBox in dropdown menu has decreased dropdown buttons (T739065)

EugeniyKiyashko:


DokaRus (approved):
It would be nice to attach small screenshots "before"/"after" for such PR

jsdmitry (approved):

PR #8806, opened 2019-07-15
 Remove data sources + fix shape categories after refactoring

Romzay:


romanresh (approved):
Less code, better code

PR #8821, opened 2019-07-16
Wrap item text in the dxDropDownButton

mrjohn0011:


RomanTsukanov (dismissed):

styles/widgets/common/list.less
```diff
@@ -9,6 +9,10 @@
     }
 }
 
+.dx-wrap-item-text .dx-item-content {
```
DokaRus: How about item template with a nested collection widget which has its own items?

js/ui/list/ui.list.base.js
```diff
@@ -1105,6 +1112,9 @@ var ListBase = CollectionWidget.inherit({
             case "groupTemplate":
                 this._invalidate();
                 break;
+            case "wrapItemText":
+                this._itemElements().toggleClass(WRAP_ITEM_TEXT_CLASS, args.value);
```
DokaRus: Maybe we should add this class to the root element?

js/ui/collection/ui.collection_widget.base.js
```diff
@@ -898,6 +899,10 @@ var CollectionWidget = Widget.inherit({
         $container = $container || this._itemContainer();
         var $itemFrame = this._renderItemFrame(index, itemData, $container, $itemToReplace);
         this._setElementData($itemFrame, itemData, index);
+        if(this.option("showItemDataTitle")) {
+            var displayValue = this._displayGetter ? this._displayGetter(itemData) : itemData;
+            $itemFrame.attr("title", displayValue);
```
DokaRus: Perhaps we should move L902-L906 to the `_renderItemFrame`?

DokaRus (dismissed):

DokaRus (approved):

PR #8851, opened 2019-07-18
Add treeListWidth option

kruglikovstepan:


js/ui/gantt/ui.gantt.js
```diff
@@ -92,12 +103,26 @@ class Gantt extends Widget {
         return this._treeList.getVisibleRows().map(r => r.data);
     }
     _initScrollSync(treeList) {
-        const scrollable = treeList.getScrollable();
-        if(scrollable) {
-            scrollable.off("scroll", (e) => { this._onTreeListScroll(e); });
-            scrollable.on("scroll", (e) => { this._onTreeListScroll(e); });
+        const treeListScrollable = treeList.getScrollable();
+        if(treeListScrollable) {
+            treeListScrollable.off("scroll", (e) => { this._onTreeListScroll(e); });
```
vladkovl: Is it correct? AFAIK `(e) => { this._onTreeListScroll(e); }` is a new anonymous function, so it can't be found inside the event handlers list. Maybe this code will be enough:
`treeListScrollable.off("scroll");`

kruglikovstepan: You're right. I fixed it

vladkovl (approved):

PR #8878, opened 2019-07-20
Add tests for the default button/menu template styles (T747772, T738205, T742980, T636219)

IgnatovDan:

### DIFF
--- a/testing/helpers/checkStyleHelper.js
+++ b/testing/helpers/checkStyleHelper.js
@@ -7,8 +7,6 @@ function testInChromeOnDesktop(name, testCallback) {
     } else {
         QUnit.testInActiveWindow.call(null, name, testCallback);
     }
-
-    return true;
 }
 
 function checkTextOverflow(assert, element, expectedValue) {
--- a/testing/helpers/checkStyleHelper.js
+++ b/testing/helpers/checkStyleHelper.js
@@ -2,7 +2,7 @@ import devices from "core/devices";
 import browser from "core/utils/browser";
 import typeUtils from "core/utils/type";
 
-function testInChromeOnDesktop(name, testCallback) {
+export function testInChromeOnDesktop(name, testCallback) {
     if(devices.real().deviceType === "desktop" && browser.webkit) {
         QUnit.testInActiveWindow.call(null, name, testCallback);
     } else {
@@ -10,19 +10,19 @@ function testInChromeOnDesktop(name, testCallback) {
     }
 }
 
-function getTextOverflow(element) {
+export function getTextOverflow(element) {
     return window.getComputedStyle(element).textOverflow;
 }
 
-function getWhiteSpace(element) {
+export function getWhiteSpace(element) {
     return window.getComputedStyle(element).whiteSpace;
 }
 
-function getOverflowX(element) {
+export function getOverflowX(element) {
     return window.getComputedStyle(element).overflowX;
 }
 
-function getBackgroundColor(element) {
+export function getBackgroundColor(element) {
     let elementBackgroundColor = window.getComputedStyle(element).backgroundColor;
     let currentElement = element.parentNode;
     while(currentElement.parentNode !== document.documentElement && typeUtils.isDefined(currentElement.parentNode)) {
@@ -36,15 +36,6 @@ function getBackgroundColor(element) {
     return elementBackgroundColor;
 }
 
-function getColor(element) {
+export function getColor(element) {
     return window.getComputedStyle(element).color;
 }
-
-export default {
-    testInChromeOnDesktop: testInChromeOnDesktop,
-    getTextOverflow: getTextOverflow,
-    getWhiteSpace: getWhiteSpace,
-    getOverflowX: getOverflowX,
-    getBackgroundColor: getBackgroundColor,
-    getColor: getColor,
-};
--- a/testing/helpers/checkStyleHelper.js
+++ b/testing/helpers/checkStyleHelper.js
@@ -1,0 +1,53 @@
+import devices from "core/devices";
+import browser from "core/utils/browser";
+
+function isChromeOnDesktop(assert) {
+    if(devices.real().deviceType !== "desktop" || !browser.webkit) {
+        assert.ok(true, "test is designed to run in chrome on desktop only");
+        return false;
+    }
+    return true;
+}
+
+function checkTextOverflow(assert, element, expectedValue) {
+    let style = window.getComputedStyle(element.parentNode);
+    assert.equal(style.textOverflow, expectedValue, "textOverflow");
+}
+
+function checkWhiteSpace(assert, element, expectedValue) {
+    let style = window.getComputedStyle(element);
+    assert.equal(style.whiteSpace, expectedValue, "whiteSpace");
+}
+
+function checkOverflowX(assert, element, expectedValue) {
+    let style = window.getComputedStyle(element.parentNode);
+    assert.equal(style.overflowX, expectedValue, "overflowX");
+}
+
+function checkBackgroundColor(assert, element, expectedValue) {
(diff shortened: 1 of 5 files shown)


EugeniyKiyashko (dismissed):

testing/helpers/checkStyleHelper.js
```diff
@@ -0,0 +1,53 @@
...
+    assert.equal(style.textOverflow, expectedValue, "textOverflow");
+}
+
+function checkWhiteSpace(assert, element, expectedValue) {
+    let style = window.getComputedStyle(element);
+    assert.equal(style.whiteSpace, expectedValue, "whiteSpace");
+}
+
+function checkOverflowX(assert, element, expectedValue) {
+    let style = window.getComputedStyle(element.parentNode);
```
churkin: You can extract `checkStyle` method with `checkParents` param for the background color style

testing/helpers/checkStyleHelper.js
```diff
@@ -0,0 +1,53 @@
...
+    assert.equal(style.overflowX, expectedValue, "overflowX");
```
churkin: I believe we should use `strictEqual` anywhere, even if it's not critical here

testing/helpers/checkStyleHelper.js
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function checkBackgroundColor(assert, element, expectedValue) {
+    let elementBackgroundColor = window.getComputedStyle(element).backgroundColor;
+    let currentElement = element.parentNode;
+    for(let i = 0; i < 5; i++) {
```
churkin: What does `5` mean?

testing/helpers/checkStyleHelper.js
```diff
@@ -0,0 +1,53 @@
+import devices from "core/devices";
+import browser from "core/utils/browser";
```
churkin, line 2: import { webkit } from "core/utils/browser";

testing/helpers/checkStyleHelper.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+    assert.equal(elementBackgroundColor, expectedValue, "backgroundColor");
+}
+
+function checkColor(assert, element, expectedValue) {
+    let style = window.getComputedStyle(element);
+    assert.equal(style.color, expectedValue, "color");
+}
+
+export default {
```
churkin: Can we export each function separately?

testing/helpers/checkStyleHelper.js
```diff
@@ -0,0 +1,53 @@
...
+    if(devices.real().deviceType === "desktop" && browser.webkit) {
+        QUnit.testInActiveWindow.call(null, name, testCallback);
+    } else {
+        QUnit.skip.call(null, name, testCallback);
+    }
+}
+
+function checkTextOverflow(assert, element, expectedValue) {
+    let style = window.getComputedStyle(element.parentNode);
+    assert.equal(style.textOverflow, expectedValue, "textOverflow");
```
churkin: we can return boolean value instead of using assert in the helpers

churkin (changes requested):

testing/helpers/checkStyleHelper.js
IgnatovDan: The 'browser.' prefix makes it easier to understand what does the 'webkit' variable value mean:
  'if(browser.webkit) {'  vs  'if(webkit) {'
The 'import { webkit }' and 'if(webkit) {...' approaches are not used in our code so I don't want to introduce them for the first time if I am not sure that they are better.

churkin: I think everyone is very clear that `webkit` means in the test`s context :). Our code is not ideal, if we can do it better - we should.

testing/helpers/checkStyleHelper.js
IgnatovDan: Boolean result from this function means this code in test:

assert.ok(checkTextOverflow(style.textOverflow, expectedValue), "textOverflow");

With this code I will not see the 'expected' and 'actual' values in the test failure details.
This is why I want to use assert.equal with both these values.

However, I can change my functions to the 'getTextOverflow' approach and write my tests in this way:

assert.equal(CheckStyleHelper.getTextOverflow($template[0]), "clip", "textOverflow");

It keeps both values and allows to check styles of several elements in one test and provide different message for each assert . This approach looks better and I will change my code to implement it.

testing/helpers/checkStyleHelper.js
IgnatovDan: Yes, I used equal instead of strictEqual by a mistake. I will change it.

testing/helpers/checkStyleHelper.js
churkin: I like it!

testing/helpers/checkStyleHelper.js
IgnatovDan: The 'for(let i = 0; i < 5; i++) {' seems to be an excessive optimization: I used '5' to break enumeration through the entire hierarchy in the case when the background color is lost.
The 'while(currentElement.parentNode != document) {' cycle requires nearly the same time and I will remove '5'.

churkin: `while(currentElement.parentNode != document.documentElement)` I think

testing/helpers/checkStyleHelper.js
IgnatovDan: I don't see any advantage in using the 'if(webkit)' template instead of the widely used 'if(browser.webkit)' template. For a single place among a lot of others I see disadvantages: it can make it harder to review all the 'browser.webkit' occurrences and can add some time to make sure that the 'webkit' value means the usual 'browser.webkit' value.
We should use the same condition syntax in this file as in other files.

churkin: It doesn't matter to me, let it be `browser.webkit`

testing/helpers/checkStyleHelper.js
IgnatovDan: I have changed code and now the only 'element' argument is required:
function getOverflowX(element) {
    return window.getComputedStyle(element).overflowX;
}

In tests, it used in this way:
    assert.strictEqual(CheckStyleHelper.getOverflowX($template[0].parentNode), "hidden", "overflowX");

This means: Currently our code declares the 'hidden' value for the 'overflowX' CSS property so content of my template is hidden and anyone can expect that this behavior will not be changed in future (w/o bc).

Whether or not this code splitting is better?

testing/helpers/checkStyleHelper.js
IgnatovDan: Done.

testing/helpers/checkStyleHelper.js
IgnatovDan: Yes, it was a typo. Done.

testing/helpers/checkStyleHelper.js
IgnatovDan: Yes, we can export each function separately, without 'export default {'
It will be used in this way:
- import * as CheckStyleHelper from "../../helpers/checkStyleHelper.js";
- CheckStyleHelper = require("../../helpers/checkStyleHelper.js");

But currently all the functions are used in each test.
Are you sure it is worth it at this stage with the only 46 lines of code?

testing/helpers/checkStyleHelper.js
IgnatovDan: Done

testing/helpers/checkStyleHelper.js
IgnatovDan: Done

churkin (dismissed):

EugeniyKiyashko (dismissed):

EugeniyKiyashko (approved):

andrewmakarov (approved):

PR #8888, opened 2019-07-22
Make overlay wrapper full screen and transparent when shading is disabled (T758340)

mrjohn0011:


js/ui/overlay/ui.overlay.js
```diff
@@ -1286,7 +1281,12 @@ var Overlay = Widget.inherit({
     _getContainer: function() {
         var position = this._position,
             container = this.option("container"),
-            positionOf = position ? position.of || window : null;
+            positionOf = null;
+
+        if(!container && position) {
+            var isEvent = !!(position.of && position.of.type === "positionEvent");
```
DokaRus: > position.of.type === "positionEvent"

There's no such thing. You can adopt logic from [animation/position](https://github.com/DevExpress/DevExtreme/blob/19_2/js/animation/position.js#L328)

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -808,6 +808,19 @@ testModule("position", moduleConfig, () => {
         assert.strictEqual($overlayWrapper.css("position"), devices.real().ios ? "absolute" : "fixed");
     });
 
+    test("wrapper should have 100% width and height when shading is disabled", (assert) => {
```
DokaRus, line 811: Add the `shading` option change test at runtime. And check the `pointer-events` value

DokaRus (changes requested):
Should we update or remove `_renderShadingDimensions` from [Popover](https://github.com/mrjohn0011/DevExtreme/blob/overlay_shading/js/ui/popover.js#L539)?

styles/widgets/common/list.less
```diff
@@ -51,7 +51,6 @@
     cursor: pointer;
     display: table;
     width: 100%;
-    table-layout: fixed;
```
DokaRus: it's gonna cause a problem with the ListItem badge

DokaRus (approved):

jsdmitry (approved):

PR #8895, opened 2019-07-23
DataGrid - added row should be scrolled to the top of the grid (T748046)

igormalzew:


AntonSermyazhko (dismissed):

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -3948,14 +3948,14 @@ QUnit.testInActiveWindow('Insert Row after scrolling', function(assert) {
         that.click(headerPanelElement, '.dx-datagrid-addrow-button');
 
         // assert
-        assert.strictEqual(rowsView.getTopVisibleItemIndex(), 1);
-        assert.equal(getInputElements(testElement.find('tbody > tr').eq(1)).length, 4);
+        assert.strictEqual(rowsView.getTopVisibleItemIndex(true), 0);
```
vconst: This test check that row inserted in not zero position after scrolling. So, it will be better to change scrollTo position in this test instead of changing asserts

vconst (changes requested):

vconst (approved):

PR #8919, opened 2019-07-24
dxSpeedDialAction - add 'visible' option

groshenkovamarina:

### DIFF
--- a/js/ui/speed_dial_action/speed_dial_action.js
+++ b/js/ui/speed_dial_action/speed_dial_action.js
@@ -32,9 +32,8 @@ const SpeedDialAction = Widget.inherit({
 
             /**
             * @name dxSpeedDialActionOptions.visible
-            * @hidden
             */
-            visible: false,
+            visible: true,
 
             /**
             * @name dxSpeedDialActionOptions.width
@@ -91,6 +90,11 @@ const SpeedDialAction = Widget.inherit({
             case "icon":
                 initAction(this);
                 break;
+            case "visible":
+                args.value ?
+                    initAction(this) :
+                    disposeAction(this);
+                break;
             case "animation":
             case "id":
                 break;
@@ -100,6 +104,10 @@ const SpeedDialAction = Widget.inherit({
     },
 
     _render() {
+        this._toggleVisibility(false);
+
+        if(!this.option("visible")) return;
+
         if(!getSwatchContainer(this.$element())) {
             ready(() => initAction(this));
         } else {
--- a/js/ui/speed_dial_action/speed_dial_action.js
+++ b/js/ui/speed_dial_action/speed_dial_action.js
@@ -93,7 +93,7 @@ const SpeedDialAction = Widget.inherit({
             case "visible":
                 args.value ?
                     initAction(this) :
-                    disposeAction(this);
+                    disposeAction(this._options.id);
                 break;
             case "animation":
             case "id":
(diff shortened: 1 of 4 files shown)


testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
```diff
@@ -339,4 +338,25 @@ QUnit.module("check action buttons click args", (hooks) => {
     });
 });
 
+QUnit.module("add visible option", (hooks) => {
```
babich-a, line 341: I want you to make test with one SDA.
Simple scenario:
1) Create, 2) set `visible` to` false`, 3) set `visible` to `true`
It is interesting because in this case we completely dispose and remove the SpeedDialMainItem

js/ui/speed_dial_action/speed_dial_action.js
```diff
@@ -91,6 +90,11 @@ const SpeedDialAction = Widget.inherit({
             case "icon":
                 initAction(this);
                 break;
+            case "visible":
```
babich-a, line 93: What about this scenario:
1. Create SDA with `visible:false`
2. Change `icon` option
It seems that this https://github.com/DevExpress/DevExtreme/pull/8919/files#diff-645315b92c27d92aebe52d8df754b0e0R91 initAction will fire. Or not?

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
groshenkovamarina: Add test and fix

js/ui/speed_dial_action/speed_dial_action.js
groshenkovamarina: Add test and fix

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
```diff
@@ -339,4 +338,35 @@ QUnit.module("check action buttons click args", (hooks) => {
...
+        $("#fab-one").dxSpeedDialAction("instance").dispose();
+        $("#fab-two").dxSpeedDialAction("instance").dispose();
+    }),
+
+    test("check rendering", (assert) => {
+        const firstSDA = $("#fab-one").dxSpeedDialAction().dxSpeedDialAction("instance");
+
+        firstSDA.option("visible", false);
+
+        assert.equal($("." + FAB_CLASS).length, 0, "invisible if change visible option to false");
```
babich-a: I suggest to make some variable from "." + FAB_CLASS

babich-a (dismissed):

BingoRUS:
When we have two buttons and change their visibility, the main FAB has a wrong position after re-render. I'd suggest to fix it before merging this PR.

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
groshenkovamarina: Variable will store the first value

groshenkovamarina:
@BingoRUS added fix

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
babich-a: Not for length for ('.'+FAB_CLASS). This is const.

babich-a (approved):

BingoRUS (approved):
👍

PR #8930, opened 2019-07-24
DataGrid - onFocusedCellChanged event arguments should contains valid row object is scrolling mode is virtual (T801041)

AntonSermyazhko:


js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -1603,7 +1604,8 @@ var KeyboardNavigationController = core.ViewController.inherit({
             columnIndex = that.getView("rowsView").getCellIndex($cellElement),
             rowIndex = this._getRowIndex($cellElement && $cellElement.parent()),
             isEditingCell = that.getController("editing").isEditCell(rowIndex, columnIndex),
```
vconst: isEditiCell method should use localRowIndex

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -1603,7 +1604,8 @@ var KeyboardNavigationController = core.ViewController.inherit({
...
-            row = dataController.items()[rowIndex];
+            localRowIndex = Math.min(rowIndex - dataController.getRowIndexOffset(), dataController.items().length - 1),
```
vconst, line 1606: Expression

rowIndex - dataController.getRowIndexOffset() +  dataController.getRowIndexDelta()

should be used to convert global row index to local. The getRowIndexDelta method call needs for rowRenderingMode virtual

vconst (changes requested):

PR #9004, opened 2019-07-29
Call change event handler on focusout in the dxTextEditor with mask (T759547)

mrjohn0011:
TextEditor with mask prevents default behavior on keydown. 
It blocks change event after the focusout.
We are going to use our valueChange handler both on 'change' and 'focusout' events


DokaRus (approved):

San4es (approved):
LGTM

PR #9018, opened 2019-07-30
DataGrid: Editing should be prevented for Boolean column in cell mode if cancel is set in onRowUpdating event (T755201)

jtoming830:


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1703,11 +1703,22 @@ var EditingController = modules.ViewController.inherit((function() {
                     type: DATA_EDIT_DATA_UPDATE_TYPE
                 };
 
+                var isShowEditorAlwaysWithoutForceUpdate = options.column.showEditorAlways && !forceUpdateRow;
+                var isUpdateInCellMode = editMode === EDIT_MODE_CELL && options.row && !options.row.inserted;
```
AntonSermyazhko: Use one "var" for variable declarations here

AntonSermyazhko: updateFieldValue is too big. Maybe split it into cell and row update types?

vconst (approved):

PR #9020, opened 2019-07-30
Set ellipsis for dxDropDownEditor's placeholder (T802325)

mrjohn0011:
image

### DIFF
--- a/styles/widgets/common/dropDownEditor.less
+++ b/styles/widgets/common/dropDownEditor.less
@@ -56,6 +56,12 @@
     }
 }
 
+.dx-placeholder:before {
+    max-width: 100%;
+    white-space: nowrap;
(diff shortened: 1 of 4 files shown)


styles/widgets/common/dropDownEditor.less
```diff
@@ -56,6 +56,12 @@
     }
 }
 
+.dx-placeholder:before {
```
DokaRus: I think we should move it to the [textEditor](https://github.com/mrjohn0011/DevExtreme/blob/dde_placeholder/styles/widgets/common/textEditor.less#L28) common styles

DokaRus (dismissed):

alexander-kotov-dx (dismissed):

DokaRus (approved):

PR #9029, opened 2019-07-30
Add showResources and taskTitlePosition options to Gantt

kruglikovstepan:


js/ui/gantt/ui.gantt.js
```diff
@@ -58,6 +59,8 @@ class Gantt extends Widget {
 
         this._treeList = this._createComponent(this._$treeList, dxTreeList, {
             dataSource: this.option("tasks.dataSource"),
+            keyExpr: this.option("tasks.keyExpr"),
```
DokaRus:
```
const { keyExpr, dataSource, parentIdExpr } = this.option("tasks");
<...>
{
  keyExpr,
  dataSource,
  parentIdExpr
}

File: js/ui/gantt/ui.gantt.js [hunk 2]
```diff
@@ -304,6 +328,12 @@ class Gantt extends Widget {
             case "treeListWidth":
                 this._updateWidth(args.value);
                 break;
+            case "showResources":
+                this._ganttView && this._ganttView.option("showResources", args.value);
```
DokaRus: Perhaps we can define a separate method that will change the GanttView options if view exists (to avoid duplication of `this._ganttView && this._ganttView.option(<...>)`).

DokaRus (approved):

PR #9064, opened 2019-08-01
DataGrid - Should hide filter row menu after losing it's focus

igormalzew:


js/ui/grid_core/ui.grid_core.filter_row.js
```diff
@@ -538,11 +543,15 @@ var ColumnHeadersViewFilterRowExtender = (function() {
...
-
                     eventsEngine.trigger($menu, "blur");
-                    menu && menu.option("focusedElement", null);
-                    isCellWasFocused && that._focusEditor($editorContainer);
+                    restoreFocus();
+                },
+                onContentReady: function(e) {
+                    eventsEngine.on($menu, "blur", () => {
+                        var menu = e.component;
+                        menu._hideSubmenu(menu._visibleSubmenu);
```
vconst, line 552: Is it posiible not use private API of Menu widget?

igormalzew: There are no public methods to close the menu. I also tried to trigger various events on the menu element, but this did not produce results.

testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -178,6 +178,33 @@ QUnit.test("Accessibility columns id should not set for columns editors (T710132
...
+    $menu.focus();
+
+    menuInstance = $menu.dxMenu("instance");
+    $root = $(menuInstance.itemElements().get(0));
+    menuInstance._showSubmenu($root);
+
+    $menu.trigger("blur");
+
+    // assert
+    assert.notOk(!!menuInstance._visibleSubmenu, "submenu is hidden");
```
vconst: It is bad assert. If _visibleSubmenu is renamed test will not fail. It is better to check menu overlay element visiibility

vconst (changes requested):

testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
igormalzew: I added double check for the submenu visibility which covers the "private field is renamed" case.

vconst (approved):

AntonSermyazhko (approved):

PR #9079, opened 2019-08-02
File Manager - Ability to set current directory by path

maxerokh:

### DIFF
--- a/js/ui/file_manager/ui.file_manager.js
+++ b/js/ui/file_manager/ui.file_manager.js
@@ -38,7 +38,7 @@ class FileManager extends Widget {
     _initMarkup() {
         super._initMarkup();
 
-        this._onPathChangedAction = this._createActionByOption("onPathChanged");
+        this._onCurrentDirectoryChangedAction = this._createActionByOption("onCurrentDirectoryChanged");
         this._onSelectedFileOpenedAction = this._createActionByOption("onSelectedFileOpened");
 
         this._controller = new FileItemsController({
@@ -373,14 +373,14 @@ class FileManager extends Widget {
             customizeDetailColumns: null,
 
             /**
-            * @name dxFileManagerOptions.onPathChanged
+            * @name dxFileManagerOptions.onCurrentDirectoryChanged
             * @extends Action
             * @type function(e)
             * @type_function_param1 e:object
             * @default null
             * @action
             */
-            onPathChanged: null,
+            onCurrentDirectoryChanged: null,
 
             /**
             * @name dxFileManagerOptions.onSelectedFileOpened
@@ -453,8 +453,8 @@ class FileManager extends Widget {
             case "permissions":
                 this.repaint();
                 break;
-            case "onPathChanged":
-                this._onPathChangedAction = this._createActionByOption("onPathChanged");
+            case "onCurrentDirectoryChanged":
+                this._onCurrentDirectoryChangedAction = this._createActionByOption("onCurrentDirectoryChanged");
                 break;
             case "onSelectedFileOpened":
                 this._onSelectedFileOpenedAction = this._createActionByOption("onSelectedFileOpened");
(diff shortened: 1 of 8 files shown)


js/ui/file_manager/ui.file_manager.js
```diff
@@ -419,6 +419,24 @@ class FileManager extends Widget {
...
+
+    /**
+     * @name dxFileManagetMethods.getPath
+     * @publicName getPath()
+     * @return string
+     */
+    getPath() {
+        return this._controller.getCurrentPath();
+    }
+
```
vladkovl: How about `"initialPath"` option? This option will open the specified folder immediately without loading root folder content into the item list view.

I think it's better to use `"path"` option instead of the `getPath()/setPath()` methods. So we will be able to replace `"initialPath"` with this `"path"` option.

maxerokh: I think about it, but the methods approach coverages more scenarios. For instance,

- How to complete something, after change folder? Event?
- How to get current path? In your case, we'll update options object

vladkovl: > How to complete something, after change folder? Event?

We need `currentFolderChanged` and `errorOccured` events regardless of the current PR. So, this events may be a solution for this case too.

> How to get current path? In your case, we'll update options object

`fileManager.option("path")`
AFAIK we shouldn't have problems with this aproach.

maxerokh: So, I have implemented your suggestions. What do you think about this version?

vladkovl: Not really. In my previous comment I mentioned the `currentFolderChanged` event, not the `pathChanged`, because:
- `currentFolderChanged` event gives full information about the current folder - it's id, path, lastWriteTime, in future - metadata and access rights
- removes the problem when the folder changed from one to another with the same name (relative path), so the path will not change, but the folder will change.

By the way, we can address `currentFolderChanged` event implementation in the next PR.

maxerokh: onPathChanged -> onCurrentDirectoryChanged

vladkovl: OK

vladkovl (dismissed):

js/ui/file_manager/ui.file_manager.js
```diff
@@ -314,6 +316,13 @@ class FileManager extends Widget {
             */
             fileProvider: null,
 
+            /**
+            * @name dxFileManagerOptions.path
```
RomanTsukanov: Path to what? Could you give an example?

js/ui/file_manager/ui.file_manager.js
```diff
@@ -363,6 +372,16 @@ class FileManager extends Widget {
             */
             customizeDetailColumns: null,
 
+            /**
+            * @name dxFileManagerOptions.onCurrentDirectoryChanged
+            * @extends Action
+            * @type function(e)
+            * @type_function_param1 e:object
```
RomanTsukanov, line 379: Any object? Can we be more specific?

js/ui/file_manager/ui.file_manager.js
maxerokh: This is path to current directory that file manager displays.

js/ui/file_manager/ui.file_manager.js
maxerokh: The event needs to update something on a web page when current directory was changed. All required info, you can get from file manager's instance, such as current directory path.

js/ui/file_manager/ui.file_manager.js
RomanTsukanov: Shouldn't it be `currentPath`?

maxerokh: Renamed.

RomanTsukanov (approved):

PR #9093, opened 2019-08-05
dxSpeedDialAction - add "label" option

groshenkovamarina:


build/style-compiler/LessRegistry.cs
```diff
@@ -181,7 +181,7 @@ static string[] GenerateWidgetsLessFileList(string themeName)
                 "drawer",
                 "card",
                 "htmlEditor",
-                "floatingActionButton",
+                "speedDialAction",
```
babich-a: The corresponding changes should be done in theme.less. This file used only for themebuilder assets generation.

js/ui/speed_dial_action/speed_dial_item.js
```diff
@@ -38,10 +39,29 @@ const SpeedDialItem = Overlay.inherit({
...
         this._renderClick();
     },
 
+    _renderLabel() {
+        !!this._$label && this._$label.remove();
+        this.$content().css("maxWidth", "");
+
+        const labelText = this.option("label");
+
+        if(labelText === "") {
```
babich-a: What about null and undefined?

js/ui/speed_dial_action/speed_dial_main_item.js
```diff
@@ -67,6 +69,16 @@ const SpeedDialMainItem = SpeedDialItem.inherit({
         this._renderClick();
     },
 
+    _renderLabel() {
+        this.callBase();
+        this.$element().toggleClass(FAB_MAIN_CLASS_HAS_TEXT, !!this._$label);
+        if(this._$label) {
+            const contentMaxWidth = Math.ceil(this._$icon.outerWidth() + this._$label.outerWidth());
```
babich-a: Was this tested for react, vue (cases when widget can initialized before dom is ready)?

js/ui/speed_dial_action/speed_dial_action.js
```diff
@@ -30,6 +30,13 @@ const SpeedDialAction = Widget.inherit({
 
             onClick: null,
 
+            /**
+            * @name dxSpeedDialActionOptions.label
+            * @type string
+            * @default ""
+            */
+            label: "",
```
babich-a, line 38: We can set only text? What about the position?

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
```diff
@@ -388,4 +389,59 @@ QUnit.module("add visible option", (hooks) => {
...
+        firstSDA = $("#fab-one").dxSpeedDialAction({
+            label: "first action"
+        }).dxSpeedDialAction("instance");
+
+        assert.equal($(FAB_MAIN_SELECTOR).find(FAB_LABEL_SELECTOR).text(), "first action", "FAB has label");
+        assert.ok($(FAB_MAIN_SELECTOR).hasClass("dx-fa-button-has-text"), "FAB has class");
+
+        let $fabMainContent = $(FAB_MAIN_SELECTOR).find(".dx-overlay-content");
+
+        assert.equal($fabMainContent.offset().top, $(window).height() - ($fabMainContent.outerHeight() + 16), "default position top");
```
babich-a: I think we need a named variable for $(window).height() - ($fabMainContent.outerHeight() + 16) and better comment for this assert to understand what this assert is doing.

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
```diff
@@ -388,4 +389,59 @@ QUnit.module("add visible option", (hooks) => {
...
+        assert.equal($fabMainContent.offset().left, $(window).width() - ($fabMainContent.outerWidth() + 16), "default position left");
+
+        secondSDA = $("#fab-two").dxSpeedDialAction({
+            label: "second action"
+        }).dxSpeedDialAction("instance");
+
+        $fabMainContent = $(FAB_MAIN_SELECTOR).find(".dx-overlay-content");
+
+        assert.equal($(FAB_MAIN_SELECTOR).find(FAB_LABEL_SELECTOR).length, 0, "FAB has no label if create second SDA");
```
babich-a: Look! This comment (FAB has no label if create second SDA) is much better than previous (default position left).

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
```diff
@@ -388,4 +389,59 @@ QUnit.module("add visible option", (hooks) => {
...
+        assert.ok(!$(FAB_MAIN_SELECTOR).hasClass("dx-fa-button-has-text"), "FAB has no class if create second SDA");
+        assert.equal($(FAB_SELECTOR).find(FAB_LABEL_SELECTOR).eq(0).text(), "first action", "first SDA has label");
+        assert.equal($(FAB_SELECTOR).find(FAB_LABEL_SELECTOR).eq(1).text(), "second action", "second SDA has label");
+        assert.equal($fabMainContent.offset().top, $(window).height() - ($fabMainContent.outerHeight() + 16), "default position top");
+        assert.equal($fabMainContent.offset().left, $(window).width() - ($fabMainContent.outerWidth() + 16), "default position left");
+
+        config({
+            floatingActionButtonConfig: {
+                label: "fab"
```
babich-a: Hm. We have an opportunity to set label via floatingActionButtonConfig. Where is the doc comment to document it? I see only https://github.com/DevExpress/DevExtreme/pull/9093/files#diff-645315b92c27d92aebe52d8df754b0e0R33 for SDA option.

babich-a (changes requested):

build/style-compiler/LessRegistry.cs
groshenkovamarina: Fixed theme.less

js/ui/speed_dial_action/speed_dial_action.js
groshenkovamarina: Fixed. Position changes according to FAB position if the position set is left label shows on the right.

js/ui/speed_dial_action/speed_dial_main_item.js
groshenkovamarina: For react and vue approaches there is this code https://github.com/DevExpress/DevExtreme/blob/e70199df4b9cb6272457e107be60afe6284a16f3/js/ui/speed_dial_action/speed_dial_action.js#L117

js/ui/speed_dial_action/speed_dial_action.js
babich-a: We have added some magic instead of adding a simple option. 'left' in position.at does not mean that I want to render the labels to the right. Anyway, the 'at' option can contain 'x' and 'y' (https://js.devexpress.com/Documentation/ApiReference/Common/Object_Structures/positionConfig/at/) and you code [here](https://github.com/DevExpress/DevExtreme/pull/9093/files#diff-a57151fa32bbc9ff906be3840d2c859bR70) can cause the error when you try to call `startsWith` on the object.

js/ui/speed_dial_action/speed_dial_item.js
```diff
@@ -38,10 +40,38 @@ const SpeedDialItem = Overlay.inherit({
...
         this._renderClick();
     },
 
+    _renderLabel() {
+        !!this._$label && this._$label.remove();
+        this.$content().css("maxWidth", "");
+
+        const labelText = this.option("label");
+
+        if(!labelText || labelText === "") {
```
babich-a: I think `!labelText` will be enough.

babich-a (changes requested):

babich-a (dismissed):

babich-a (dismissed):

babich-a (dismissed):

babich-a (dismissed):

js/ui/speed_dial_action/speed_dial_item.js
```diff
@@ -38,10 +41,40 @@ const SpeedDialItem = Overlay.inherit({
...
+
+        if(!labelText) {
+            this._$label = null;
+            return;
+        }
+
+        const $element = $("<div>").addClass(FAB_LABEL_CLASS);
+
+        this._$label = $element
+            .html(labelText)
```
BingoRUS: Does it mean `label` works just like `template`? How about script tags injected in this string?

js/ui/speed_dial_action/speed_dial_main_item.js
```diff
@@ -8,6 +8,7 @@ import SpeedDialItem from "./speed_dial_item";
 import themes from "../themes";
 
 const FAB_MAIN_CLASS = "dx-fa-button-main";
+const FAB_MAIN_CLASS_HAS_TEXT = "dx-fa-button-has-label";
```
BingoRUS: I don't remember if we use verbs in class names.
```suggestion
const FAB_MAIN_CLASS_HAS_TEXT = "dx-fa-button-with-label";
```

styles/widgets/base/floatingActionButton.less
```diff
@@ -1,57 +0,0 @@
-.dx-fa-button-mixin(@main-action-button-size, @action-button-size, @shadow, @icon-size) {
```
BingoRUS, line 1: Use [git-mv](https://git-scm.com/docs/git-mv) next time please.

js/ui/speed_dial_action/speed_dial_main_item.js
babich-a: https://github.com/DevExpress/DevExtreme/blob/3f6e5b98c8c9e6192581a2ed56b3e081adb79bc0/styles/widgets/material/toolbar.material.less#L180 dx-button-has-text for example

styles/widgets/material/speedDialAction.material.less
```diff
@@ -0,0 +1,40 @@
...
+                font-size: @MATERIAL_BASE_FONT_SIZE;
+                text-transform: uppercase;
+                letter-spacing: 1.25px;
+                font-weight: 500;
+            }
+        }
+    }
+
+    .dx-overlay-content {
+        .dx-fa-button-label {
```
BingoRUS: These class rules look the same in generic and material. Isn't it better to move them in the mixin?

js/ui/speed_dial_action/speed_dial_main_item.js
BingoRUS: 💩

BingoRUS (changes requested):
Let's remove `html` method usage in speed_dial_item.js.

js/ui/speed_dial_action/speed_dial_main_item.js
```diff
@@ -130,15 +140,25 @@ const SpeedDialMainItem = SpeedDialItem.inherit({
                 this._clickHandler();
             });
 
+
```
babich-a, line 143: I suggest moving this arithmetical code to separate method like _calculatePosition

babich-a (dismissed):

babich-a (approved):

BingoRUS (approved):

PR #9178, opened 2019-08-08
Fix selection behavior if null assigned to selectedRowKeys option

KiriosK:


testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -6134,6 +6134,26 @@ QUnit.test("empty selection should be restored from state storing if selectedRow
     assert.deepEqual(dataGrid.getSelectedRowsData(), [], "getSelectedRowsData result");
 });
 
+QUnit.test("assign null to selectedRowKeys option unselect selected items", function(assert) {
```
AntonSermyazhko, line 6137: Is it possible to write this test in "datagrid/selection.tests.js"?
In this case, the test should take less time and will be placed in the right place.

mrjohn0011 (approved):

vconst (approved):

vconst (approved):

testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
KiriosK: It's not possible to place this test in "selection.tests.js", I tried already.

AntonSermyazhko (approved):

PR #9187, opened 2019-08-09
Detach focus events before fieldTemplate rendering in SelectBox (T751314)

KsyushaM:


mrjohn0011 (approved):

mrjohn0011 (changes requested):
Use ES5 or translate all tests to ES6

mrjohn0011 (approved):

PR #9245, opened 2019-08-13
FileManager - Edit operation progress and notification about completion

vladkovl:
Added new UI elements for the operation progress tracking:
- notification popup;
- extended toolbar item based on "refresh" command toolbar item;
- progress panel.


js/ui/file_manager/file_items_controller.js
```diff
@@ -281,14 +312,14 @@ export default class FileItemsController {
...
-        return extend(this._createFileInfo(fileItem, parentDirectoryInfo), {
+        return extend(this.createFileInfo(fileItem, parentDirectoryInfo), {
             icon: "folder",
             expanded: fileItem.isRoot,
             items: [ ]
         });
     }
 
-    _createFileInfo(fileItem, parentDirectoryInfo) {
+    createFileInfo(fileItem, parentDirectoryInfo) {
```
maxerokh: The `createFileInfo` method should be private. The main idea, that only controller object should get info about file items. Also, the uploading mechanism goes throw file provider. Therefore, I couldn't see any reasons to make the method as public.

js/ui/file_manager/ui.file_manager.editing.js
```diff
@@ -160,43 +213,188 @@ class FileManagerEditingControl extends Widget {
...
+        };
     }
 
     _tryUpload(destinationFolder) {
-        this._uploadFolder = destinationFolder && destinationFolder[0] || this._getCurrentDirectory();
+        this._uploadDirectoryInfo = destinationFolder && destinationFolder[0] || this._getCurrentDirectory();
         this._fileUploader.tryUpload();
     }
 
+    _getItemInfosForUploaderFiles(files, parentDirectoryInfo) {
```
maxerokh: I think, that best home for the method is controller class. Main concept of controller class was described above.

js/ui/file_manager/file_items_controller.js
```diff
@@ -147,50 +147,81 @@ export default class FileItemsController {
     }
 
     createDirectory(parentDirectoryInfo, name) {
-        return when(this._fileProvider.createFolder(parentDirectoryInfo.fileItem, name))
-            .done(() => this._resetDirectoryState(parentDirectoryInfo));
+        return this._fileProvider.createFolder(parentDirectoryInfo.fileItem, name);
+    }
+
+    completeCreateDirectory(parentDirectoryInfo) {
```
maxerokh, line 153: It is a bad approach when you call action, that should always call from another entity. It is potentional place for a big bugs in future. I think, the best solution is revert the complete... changes, and solve the problem in future PR.

In object that call complete... methods you can define empty promise.

maxerokh (changes requested):

js/ui/file_manager/file_items_controller.js
vladkovl: We can't revert the complete... changes without big refactoring. The previous version of the code already has many bugs that are fixed in the current version of the code. So, I think we should do this kind of refactoring in the future PR.

js/ui/file_manager/file_items_controller.js
vladkovl: Fixed

js/ui/file_manager/ui.file_manager.editing.js
vladkovl: Agree. Fixed.

maxerokh (dismissed):

js/ui/file_manager/ui.file_manager.editing.js
```diff
@@ -160,40 +212,174 @@ class FileManagerEditingControl extends Widget {
             return;
         }
 
-        if(isFunction(action)) {
-            action(arg);
+        if(!action.dialog) {
+            return action.action(arg);
         } else {
-            this._tryEditAction(action, arg);
+            return this._tryEditAction(action, arg);
```
jsdmitry: This need will make simpler: `!action.dialog ? action.action(arg) : this._tryEditAction(action, arg)`

vladkovl: Changed to this version.

jsdmitry (approved):

PR #9360, opened 2019-08-20
Blur body when bluring of an element is impossible (for IE)

mrjohn0011:
IE browser throws an exception after click on the "test" word:

```
<body>
    <div id="test">
        <div id="ftest" tabindex="0">
            test
        </div>
    </div>
</body>

<script type="text/javascript">
    var el = document.getElementById("test");
    var el2 = document.getElementById("ftest");
    el.onclick = function () {
        el.msRequestFullscreen();
        setTimeout(function () {
            el2.focus();
            setTimeout(function () {
                // try {
                    el2.blur();
                 // } catch (e) {
                 //    document.body.blur();
                 // }
            }, 500)
        }, 100)
    }
</script>
```

DevExtreme widgets call blur manually in some cases, for example, on overlay's hidding. This pull request is a temporary solution before microsoft fixed this problem in IE or MS IE 11 support stopped

### DIFF
--- a/js/core/utils/dom.js
+++ b/js/core/utils/dom.js
@@ -12,9 +12,16 @@ var $ = require("../../core/renderer"),
     elementStrategy;
 
 var resetActiveElement = function() {
-    var activeElement = domAdapter.getActiveElement();
-    if(activeElement && activeElement !== domAdapter.getBody() && activeElement.blur) {
-        activeElement.blur();
+    var activeElement = domAdapter.getActiveElement(),
+        body = domAdapter.getBody();
+
+    // todo: remove this hack after msie 11 support stopped
+    if(activeElement && activeElement !== body && activeElement.blur) {
+        try {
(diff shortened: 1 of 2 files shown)


js/core/utils/dom.js
```diff
@@ -12,9 +12,16 @@ var $ = require("../../core/renderer"),
     elementStrategy;
 
 var resetActiveElement = function() {
-    var activeElement = domAdapter.getActiveElement();
-    if(activeElement && activeElement !== domAdapter.getBody() && activeElement.blur) {
-        activeElement.blur();
+    var activeElement = domAdapter.getActiveElement(),
+        body = domAdapter.getBody();
```
ovchinnikov: This variable can be moved into catch block

ovchinnikov (approved):

San4es (approved):

PR #9361, opened 2019-08-20
Accordion is not working when it is not a first item of TabPanel (T804231)

EugeniyKiyashko:


jsdmitry (dismissed):

js/ui/collection/ui.collection_widget.edit.js
```diff
@@ -306,8 +306,10 @@ var CollectionWidget = BaseCollectionWidget.inherit({
 
         this.callBase();
 
-        var selectedItemIndices = this._getSelectedItemIndices();
-        this._renderSelection(selectedItemIndices, []);
+        deferRender(() => {
```
nightskylark: This change should be covered in `collectionWidget.tests`

nightskylark (dismissed):

nightskylark:
Why do we use different solutions for the same issues? We have used `collection_widget.async` in toolbar and `deferRender` in many other places.
Why don't we get rid of getting elements from DOM here? We could add the selected classes on the item rendering stage. It'll make widget's lifecycle more healthy.

jsdmitry (approved):

PR #9411, opened 2019-08-23
List: Use default color for icons nested into buttons (T808708)

San4es:
Fixes the [T808708](https://www.devexpress.com/Support/Center/Question/Details/T808708/) regression after #7577.

### DIFF
--- a/styles/widgets/generic/list.generic.less
+++ b/styles/widgets/generic/list.generic.less
@@ -99,10 +99,6 @@
             color: @list-item-focus-color;
         }
 
-        .dx-button .dx-icon {
-            color: @button-icon-color;
-        }
-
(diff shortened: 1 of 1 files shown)


styles/widgets/generic/list.generic.less
```diff
@@ -117,6 +121,10 @@
             color: @list-item-active-color;
         }
 
+        .dx-button .dx-icon {
+            color: @button-icon-color;
```
alexander-kotov-dx:
```
&.dx-state-focused, 
&.dx-state-active {
  .dx-button .dx-icon {
    color: @button-icon-color;
  }
}
```
?

alexander-kotov-dx (approved):

PR #9413, opened 2019-08-23
DataGrid - Focusing into rowsView should focus element with tabIndex (T805341)

AntonSermyazhko:


js/ui/grid_core/ui.grid_core.adaptivity.js
```diff
@@ -1130,15 +1130,16 @@ module.exports = {
...
-                            $dataCells = getDataCellElements($row),
-                            $targetCell = direction === "next" ? $dataCells.last() : $dataCells.first(),
-                            rowIndex = $row.get(0).rowIndex,
-                            adaptiveController = this._adaptiveController,
-                            key = this._dataController.getKeyByRowIndex(direction === "next" ? rowIndex : rowIndex - 1),
-                            isCellElementsEquals = $currentCell && $targetCell && $currentCell.get(0) === $targetCell.get(0);
-
-                        return adaptiveController.isAdaptiveDetailRowExpanded(key) && isCellElementsEquals;
+                        var $row = $currentCell.parent();
+                        if($row.length) {
```
jtoming830: To decrease 'if' count you can use variable assignment in condition like this.
 
`if(!result && $currentCell && ($row = $currentCell.parent()) && $row.length)`

IMHO, 2 'if' blocks only because of one variable assignment are overkill. 
However, this is just a small suggestion)

AntonSermyazhko: Corrected conditions

jtoming830 (approved):

vconst (approved):

vconst (approved):

PR #9416, opened 2019-08-23
Fix unexpected scrollbar appearance when zooming  (T691054)

ksercs:
* Problem: unnecessary scrollbar appearance as result of browser zooming
* Reason: size recalculation's uncertainty because of work with float numbers
* Solution: rounding the resulting sizes to hundredths and scrollbar showing/hiding depending of the ratio of the container and content when zoom is 100%

### DIFF
--- a/js/ui/scroll_view/ui.scrollbar.js
+++ b/js/ui/scroll_view/ui.scrollbar.js
@@ -175,7 +175,7 @@ var Scrollbar = Widget.inherit({
             baseContainerSize = Math.round(this.option("baseContainerSize")),
             baseContentSize = Math.round(this.option("baseContentSize"));
 
-        // NOTE: if current scrollbar's using outside of scrollbalbe
+        // NOTE: if current scrollbar's using outside of scrollable
         if(isNaN(baseContainerSize)) {
             baseContainerSize = containerSize;
             baseContentSize = contentSize;
--- a/js/ui/scroll_view/ui.scrollbar.js
+++ b/js/ui/scroll_view/ui.scrollbar.js
@@ -35,6 +35,7 @@ var Scrollbar = Widget.inherit({
             visibilityMode: SCROLLBAR_VISIBLE.onScroll,
             containerSize: 0,
             contentSize: 0,
+            isScrollUsedInsideOfScrollable: false,
             expandable: true,
             scaleRatio: 1
         });
@@ -173,9 +174,10 @@ var Scrollbar = Widget.inherit({
         var containerSize = Math.round(this.option("containerSize")),
             contentSize = Math.round(this.option("contentSize")),
             baseContainerSize = Math.round(this.option("baseContainerSize")),
-            baseContentSize = Math.round(this.option("baseContentSize"));
+            baseContentSize = Math.round(this.option("baseContentSize")),
+            isScrollUsedInsideOfScrollable = Math.round(this.option("isScrollUsedInsideOfScrollable"));
 
-        if(isNaN(baseContentSize)) {
+        if(!isScrollUsedInsideOfScrollable) {
             baseContainerSize = containerSize;
             baseContentSize = contentSize;
         }
@@ -235,6 +237,8 @@ var Scrollbar = Widget.inherit({
                 break;
             case "scaleRatio":
                 this._update();
+                break;
+            case "isScrollUsedInsideOfScrollable":
                 break;
             default:
                 this.callBase.apply(this, arguments);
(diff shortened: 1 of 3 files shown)


js/ui/scroll_view/ui.scrollbar.js
```diff
@@ -35,6 +35,7 @@ var Scrollbar = Widget.inherit({
             visibilityMode: SCROLLBAR_VISIBLE.onScroll,
             containerSize: 0,
             contentSize: 0,
+            isScrollUsedInsideOfScrollable: false,
```
mrjohn0011: Is it possible to detect if scroll is used inside of scrollable without new option?

js/ui/scroll_view/ui.scrollbar.js
```diff
@@ -216,13 +227,19 @@ var Scrollbar = Widget.inherit({
                 this.option()[args.name] = this._normalizeSize(args.value);
                 this._update();
                 break;
+            case "baseContentSize":
```
mrjohn0011, line 229: This options have no default value. If it is required to add them, they should be defined in _defaultOptions method and tested in defaultOptions.tests.js

It is strongly recommended don't add any new options or flags to the code for bug fixing. The new option can be added when there is no way to fix a bug without it.

mrjohn0011 (changes requested):
1. It is strongly recommended to fix a bug without adding new options/flags or using private functions/variables of DevExtreme components

2. Changes in JS files without unit or/and functional tests can not be merged

js/ui/scroll_view/ui.scrollbar.js
ksercs: It's possible to set default baseContainerSize and baseContentSize to impossible values e.g. -1, and detect if scroll is used inside of scrollable with these values.

ksercs: > Is it possible to detect if scroll is used inside of scrollable without new option?

It's possible to set default baseContainerSize and baseContentSize to impossible values e.g. -1 and detect if scroll is used inside of scrollable based on these values.

js/ui/scroll_view/ui.scrollbar.js
ksercs: > This options have no default value. If it is required to add them, they should be defined in _defaultOptions method and tested in defaultOptions.tests.js
> 
> It is strongly recommended don't add any new options or flags to the code for bug fixing. The new option can be added when there is no way to fix a bug without it.

It's possible to solve the task without adding new options, but _getBaseDimension function will be dupclicated in two classes. Is it better than adding a new option?

mrjohn0011: No, duplication is bad idea too. Please try to find better solution

jsdmitry: I agree with @mrjohn0011. Why these options are used in the processing of the options changed if none updates via the option method outside? And I don't see **tests** on these changes!

ksercs:
> 2. Changes in JS files without unit or/and functional tests can not be merged

The problem is strongly connected with a browser behavior features, so it's impossible to reproduce the bug programmatically, to my mind.

js/ui/scroll_view/ui.scrollbar.js
ksercs: isScrollUsedInsideOfScrollable option is removed now. Is the current version ok?

ksercs: > Is it possible to detect if scroll is used inside of scrollable without new option?

isScrollUsedInsideOfScrollable option is removed now.

hakimio:
> The problem is strongly connected with a browser behavior features, so it's impossible to reproduce the bug programmatically, to my mind.

It is: `document.body.style.zoom = 1.5` ([source](https://stackoverflow.com/a/10464073/3529279))

testing/tests/DevExpress.ui.widgets/scrollableParts/scrollable.scrollbar.tests.js
```diff
@@ -600,3 +601,19 @@ QUnit.test("content size should be rounded to prevent unexpected scrollbar appea
 
     assert.ok(scrollbar.$element().is(":hidden"), "scrollbar is not visible");
 });
+
+QUnit.test("scrollbar should be hidden when container size is almost similar to content size when zooming", function(assert) {
+    let scrollable = new Scrollable($('#scrollable'), {
+        'useNative': false
+    });
+
+    let verticalScroller = scrollable._strategy._scrollers['vertical'];
```
mrjohn0011: Please use less private variables (._variable). Try to mock getClientRect function

jsdmitry (dismissed):

testing/tests/DevExpress.ui.widgets/scrollableParts/scrollable.scrollbar.tests.js
ksercs: I minimized use of private variables. But I cannot use it at all cause I don’t see another way to specify the necessary methods in sinon.stubs.

mrjohn0011 (approved):

jsdmitry (approved):

PR #9480, opened 2019-08-28
DataGrid.ExcelJS - Set excel column width in accordance with DataGrid column width

IgnatovDan:

### DIFF
--- a/testing/tests/DevExpress.exporter/exceljsParts/exceljs.tests.js
+++ b/testing/tests/DevExpress.exporter/exceljsParts/exceljs.tests.js
@@ -1,4 +1,5 @@
 import $ from "jquery";
+import browser from "core/utils/browser";
 import ExcelJS from "exceljs";
 import ExcelJSTestHelper from "./ExcelJSTestHelper.js";
 import { exportDataGrid } from "exporter/exceljs/excelExporter";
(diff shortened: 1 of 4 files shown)


testing/tests/DevExpress.exporter/exceljsParts/exceljs.tests.js
```diff
@@ -36,25 +43,40 @@ const moduleConfig = {
...
+                    worksheet: this.worksheet,
+                    topLeftCell: topLeftCell,
+                    customizeCell: (eventArgs) => {
+                        if(typeUtils.isDefined(expectedCustomizeCellArgs)) {
+                            helper.checkCustomizeCell(eventArgs, expectedCustomizeCellArgs, this.customizeCellCallCount++);
+                        }
+                    },
+                    excelFilterEnabled: excelFilterEnabled,
+                };
+                if(exportColumnWidth !== undefined) {
```
EugeniyKiyashko:
```suggestion
                if(typeUtils.isDefined(exportColumnWidth)) {
```

IgnatovDan: In this case it's better to change arguments to "const getDataGridConfig = (options)" and use the passed values.
I have tried this approach but it requires a lot of changes and makes this commit too large.
Currently, there are explicit arguments instead of such object and this is why I used 'undefined': it is assigned to an argument if the calling code doesn't specify its value.
'null' value means a set of another use cases that should not be covered by this code.

jsdmitry (dismissed):

EugeniyKiyashko (dismissed):

jsdmitry (approved):

EugeniyKiyashko (approved):

PR #9358, opened 2019-08-20
File Manager and Gantt - Add splitter and extract common parts of its functionality

Staziz:

PR #9481, opened 2019-08-28
Add sankey to WidgetsCreation.tests

alexanderPolosatov:


roman-simionov (approved):

roman-simionov:
@romanresh @kruglikovstepan @groshenkovamarina  @nightskylark 
What about Diagram, File manager, SpeedDialAction?
I have noticed that they are missed in the widgets list.

nightskylark:
I was talking with @kruglikovstepan, and he was going to add the Gantt Diagram here. As for the rest of these widgets, ofcourse it should be added to before publising as a stable API. I suppose we should automate it somehow in future.

kruglikovstepan:
In 19.2 the gantt has already been added
https://github.com/alexanderPolosatov/DevExtreme/blob/19_2/testing/tests/DevExpress.serverSide/widgetsCreation.tests.js

What about File manager - cc to: @maxerokh @vladkovl

vladkovl:
FileManager should be added in 19_2 version in the context of [this PR](https://github.com/DevExpress/DevExtreme/pull/9358/files).
Do we need to add this widget to the widgets list in 19_1 version too?

roman-simionov:
> Do we need to add this widget to the widgets list in 19_1 version too?

yes, since it is added in v19.1
https://js.devexpress.com/Documentation/ApiReference/UI_Widgets/dxFileManager/

PR #9529, opened 2019-08-30
Move imports from bundle to widgets

babich-a:


styles/widgets/generic/validation.generic.less
```diff
@@ -1,3 +1,4 @@
+@import (once) "./overlay.generic.less";
 @import (once) "../base/validation.less";
```
BingoRUS, line 1: Maybe it's better to strengthen CSS selectors in base/validation.less and import theme-specific less after the base one?

styles/widgets/ios7/popup.ios7.less
```diff
@@ -1,3 +1,7 @@
+@import (once) "./button.ios7.less";
+@import (once) "./overlay.ios7.less";
+// @import (once) "./toolbar.ios7.less";
```
BingoRUS: Why commented?

styles/widgets/generic/validation.generic.less
babich-a: I'll change the order

styles/widgets/ios7/popup.ios7.less
babich-a: Yes. My fail. I'll fix

BingoRUS (approved):

PR #8948, opened 2019-07-25
Fix 'renderedCallback' behavior when creating nested templates during the execution of the callback (T758209)

DokaRus:

PR #9540, opened 2019-08-30
Fix lost aspnet.createComponent calls (T810336)

AlekseyMartynov:
Issue: https://devexpress.com/issue=T810336

After the PR https://github.com/DevExpress/DevExtreme/pull/8948, the order of `createComponent` routines has changed to depth-first.

The test case reproduces the issue. The callback registered to initialize the second button (`#b2`) is fired during the render of the default template of the first button (`#b1`). The scoped jQuery wrapper `$("#" + id, container)` prevents `#b2` from being found within the contents of `#b1`. 

I considered 3 options to fix the issue:
- Remove the context argument: https://github.com/DevExpress/DevExtreme/commit/338e0d4ac3d5ac05b117d1e0b2d8a3e7892a6fac (it has been there since May 2016).
- Keep the handler in queue until the element is found: https://github.com/DevExpress/DevExtreme/commit/6169fbe36b3b543cb155f723fdcb62ee31e857ff
- Use a single `templateRendered` callback and collect handlers in an internal array (`pendingCreateComponentRoutines`). I guess this is the best option.

I'm interested in your opinions.

### DIFF
--- a/js/aspnet.js
+++ b/js/aspnet.js
@@ -27,6 +27,7 @@
     }
 })(function($, setTemplateEngine, templateRendered, Guid, validationEngine, iteratorUtils, extractTemplateMarkup, encodeHtml) {
     var templateCompiler = createTemplateCompiler();
+    var pendingCreateComponentRoutines = [ ];
 
     function createTemplateCompiler() {
         var OPEN_TAG = "<%",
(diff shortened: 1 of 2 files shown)


nightskylark (approved):
LGTM

PR #9543, opened 2019-09-02
Speed Dial Action - Fix visible option

groshenkovamarina:


js/ui/speed_dial_action/speed_dial_main_item.js
```diff
@@ -87,30 +87,37 @@ const SpeedDialMainItem = SpeedDialItem.inherit({
...
+        const currentActions = actions || this.option("actions");
+
+        return currentActions.filter((action) => action.option("visible"));
+    },
+
     _clickHandler() {
-        const actions = this._actionItems;
-        actions.forEach(action => {
-            action.toggle();
+        const actions = this._actionItems.filter((action) => action.option("actionVisible"));
```
babich-a, line 118: Why we can't use 'visible' option and should save and then use additional 'actionVisible' option?

groshenkovamarina: 'visible' option use to show or hide actions by FAB click, so I add 'actionVisible' option to keep actual information about 'visible' option

js/ui/speed_dial_action/speed_dial_main_item.js
```diff
@@ -33,10 +33,9 @@ const SpeedDialMainItem = SpeedDialItem.inherit({
             hint: "",
             label: "",
             actions: [],
-            visible: true,
             activeStateEnabled: true,
             hoverStateEnabled: true,
-            indent: 56,
+            indent: 55,
```
BingoRUS, line 38: Why?

js/ui/speed_dial_action/speed_dial_main_item.js
BingoRUS: Aren't they equal? You call `action._options.actionVisible = action._options.visible;` on render...

js/ui/speed_dial_action/speed_dial_main_item.js
groshenkovamarina: By generic design, indent between the main button and action button should be 16px, but we have 17px now

js/ui/speed_dial_action/speed_dial_main_item.js
```diff
@@ -195,7 +214,7 @@ const SpeedDialMainItem = SpeedDialItem.inherit({
     _optionChanged(args) {
         switch(args.name) {
             case "actions":
-                this._renderIcon();
+                if(this._isVisible()) this._renderIcon();
```
babich-a, line 217: It looks strange that we do not render an icon for an invisible button, but render close icon, set position, etc.

babich-a (approved):

js/ui/speed_dial_action/speed_dial_main_item.js
groshenkovamarina: It is needed for icon and label because they rerender early then FAB hide. There is this condition in set position too because of the show method.

PR #9563, opened 2019-09-03
PivotGrid: Fix virtual scrolling did not work if there is no data (T810822)

jtoming830:


testing/tests/DevExpress.ui.widgets.pivotGrid/pivotGrid.tests.js
```diff
@@ -2955,6 +2955,90 @@ QUnit.test("Scrolling when virtual scrolling is enabled", function(assert) {
     scrollable.scrollTo({ left: 10, top: 1 });
 });
 
+// T810822
+QUnit.test("Virtual scrolling should work on columns area if data is empty", function(assert) {
+    var done = assert.async();
```
vconst: Async test is not good test. So, it will be better to check hasScroll and check data area scrollable options.

vconst (changes requested):

vconst (dismissed):

vconst (approved):

roman-simionov (approved):

PR #9596, opened 2019-09-05
dxViz: API for zooming in dxPolarChart

dxBeardedBear:


js/viz/chart.js
```diff
@@ -7,7 +7,8 @@ import {
     map as _map, getLog, getCategoriesInfo,
     updatePanesCanvases, convertVisualRangeObject, PANE_PADDING,
     normalizePanesHeight,
-    checkElementHasPropertyFromStyleSheet
+    checkElementHasPropertyFromStyleSheet,
+    rangesIsEqual
```
AlexJameson: Is it ok that plural 'ranges' goes with singular 'is', should it be rangesAreEqual? 

Or is it by design?

AlexJameson (approved):

alexanderPolosatov (approved):

PR #9611, opened 2019-09-09
Add 'widgets' option for devextreme-themebuilder

babich-a:


build/gulp/style-compiler.js
```diff
@@ -20,9 +20,13 @@ const commentsRegex = /\s*\/\*[\S\s]*?\*\//g;
 
 const compileBundles = (bundles) => {
     const paths = path.join(process.cwd(), 'styles');
+
     return gulp
         .src(bundles)
-        .pipe(plumber())
+        .pipe(plumber(e => {
+            console.log(e);
```
BingoRUS, line 27: Do we need logging?

styles/theme.less
```diff
@@ -9,8 +9,11 @@
 @import (once) "./widgets/@{base-theme}/common.@{base-theme}.less";
 @import (once) "./widgets/@{base-theme}/icons.@{base-theme}.less";
 @import (once) "./widgets/@{base-theme}/widget.@{base-theme}.less";
+@import (once) "./widgets/@{base-theme}/card.@{base-theme}.less";
+@import (once) "./widgets/@{base-theme}/fieldset.@{base-theme}.less";
 
-// Widgets
+// Widgets: this list (to the end of file) used by themebuilder, do not remove comment below
```
BingoRUS:
```suggestion
// Widgets: this list (to the end of file) used by DevExtreme ThemeBuilder, do not remove a comment below
```

themebuilder/modules/modules-handler.js
```diff
@@ -0,0 +1,96 @@
...
+        const fileName = importString.substr(lastSlashIndex + 1);
+        const dotIndex = fileName.indexOf(".");
+        return fileName.substr(0, dotIndex);
+    }
+
+    availableWidgets(less) {
+        const widgets = [];
+        const widgetsListIndex = less.indexOf(this.SPECIAL_COMMENT);
+
+        if(widgetsListIndex < 0) {
```
BingoRUS: Let's invert: if widgetsListIndex >= 0 then process `less`. There will be only one return statement.

themebuilder/modules/modules-handler.js
```diff
@@ -0,0 +1,96 @@
...
+            return widgets;
+        }
+
+        less
+            .substr(widgetsListIndex + this.SPECIAL_COMMENT.length)
+            .split("\n")
+            .filter(item => !!item)
+            .forEach((importString) => {
+                const lastSlashIndex = importString.lastIndexOf("/");
```
BingoRUS: Do you suppose to use `getWidgetFromImport` method here?

themebuilder/modules/modules-handler.js
```diff
@@ -0,0 +1,96 @@
...
+                    name: fileName.substr(0, dotIndex).toLowerCase(),
+                    import: importString
+                });
+            });
+
+        return widgets;
+    }
+
+    lessPlugin() {
+        const getFileNameFromContext = (context) => {
```
BingoRUS, line 39: Why keep `getFileNameFromContext` and `removeUnnecessaryWidgets` methods incide `lessPlugin`?

build/gulp/style-compiler.js
babich-a: Yes. It is useful to see the error :)

themebuilder/modules/modules-handler.js
babich-a: great!

themebuilder/modules/modules-handler.js
babich-a: I do not want to make them the ModulesHandler class methods. But I can't define variables inside class ModulesHandler {}. The solution - to define this functions outside the class, but I do not like it.

themebuilder/modules/modules-handler.js
babich-a: I'll change. But I do not like this because this condition makes the nesting bigger.

styles/widgets/material/button.material.less
```diff
@@ -15,6 +15,7 @@
     @MATERIAL_BUTTON_ACTIVE_BOX_SHADOW_SIZE: 0 4px 6px;
 }
 
+@MATERIAL_EDITOR_CUSTOM_BUTTON_MARGIN: 5px;
```
BingoRUS: Move to styles/widgets/material/textEditor.material.less?

themebuilder/modules/modules-handler.js
BingoRUS: So bad we don't have `private` in ES6...

build/gulp/style-compiler.js
BingoRUS: Okay...

BingoRUS (approved):

PR #9637, opened 2019-09-10
DataGrid - focused row style is incorrect on hover in multiple selection mode

igormalzew:


AntonSermyazhko (dismissed):

vconst (changes requested):
Material theme has the same problem

AntonSermyazhko (approved):

vconst (approved):

PR #9684, opened 2019-09-11
FileManager - Fix metadata name error and remove OneDriveProvider from namespace

vladkovl:


js/bundles/modules/file_providers.js
```diff
@@ -1,8 +1,6 @@
 import DevExpress from "./core";
 import WebApiFileProvider from "../../ui/file_manager/file_provider/webapi";
-import OneDriveFileProvider from "../../ui/file_manager/file_provider/onedrive";
```
maxerokh, line 3: Why didn't you remove the OneDriveFileProvider file?

maxerokh (changes requested):

js/bundles/modules/file_providers.js
vladkovl: Removed

maxerokh (approved):

PR #9701, opened 2019-09-13
Add treeListColumns option

kruglikovstepan:

### DIFF
--- a/js/ui/gantt/ui.gantt.js
+++ b/js/ui/gantt/ui.gantt.js
@@ -563,7 +563,7 @@ class Gantt extends Widget {
                 this._refreshDataSource("resourceAssignments");
                 break;
             case "treeListColumns":
-                this._setTreeListOption("columns", args.value);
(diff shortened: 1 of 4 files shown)


js/ui/gantt/ui.gantt.js
```diff
@@ -551,6 +562,9 @@ class Gantt extends Widget {
             case "resourceAssignments":
                 this._refreshDataSource("resourceAssignments");
                 break;
+            case "treeListColumns":
+                this._setTreeListOption("columns", args.value);
```
vconst: Do not use args.value here because user can change inner property. For example after calling gantt.option("treeListColumns[0].visible", false) args,name will be "treeListColumns, args,fullName will be "treeListColumns[0].visible", args.value will be false and this handler will be wrong.

vconst (changes requested):

vconst (approved):

IlyaKhD:
LGTM

PR #9725, opened 2019-09-16
Popup FIX Widget opens without animation for the first time (T814523)

ksercs:

### DIFF
--- a/js/ui/popup.js
+++ b/js/ui/popup.js
@@ -808,11 +808,12 @@ var Popup = Overlay.inherit({
             case "useDefaultToolbarButtons":
             case "useFlatToolbarButtons":
                 var isPartialUpdate = args.fullName.search(".options") !== -1;
-                var isInitialize = args.fullName === "toolbarItems";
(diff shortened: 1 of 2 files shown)


js/ui/popup.js
```diff
@@ -808,10 +808,11 @@ var Popup = Overlay.inherit({
             case "useDefaultToolbarButtons":
             case "useFlatToolbarButtons":
                 var isPartialUpdate = args.fullName.search(".options") !== -1;
+                var isInitialize = args.fullName === "toolbarItems";
```
alexander-kotov-dx: it does not work for this case:
```
    $("#button").dxButton({
        text: "Open",
        onClick: () => {
          popup.option("visible", true);
          popup.option("toolbarItems[0]", {
            widget: "dxButton",
            options: {text: 'Supprimer', type:'danger'}
          });
        }
    });
```

alexander-kotov-dx (approved):

San4es (approved):

PR #9784, opened 2019-09-19
Add localization constants for Diagram

LanaDX:
Add dxDiagram localization strings


DokaRus (commented):
Add localization constants to other localization files

Constants that require translation can be set as follows:
"dxDiagram-newMessage": "TODO"

romanresh (dismissed):

DokaRus (approved):
LGTM

PR #9795, opened 2019-09-20
Floating Action Button - Add shading option

groshenkovamarina:


js/ui/speed_dial_action/speed_dial_main_item.js
```diff
@@ -235,6 +240,10 @@ const SpeedDialMainItem = SpeedDialItem.inherit({
             case "icon":
                 if(this._isVisible()) this._renderIcon();
                 break;
+            case "shading":
```
babich-a: I suggest not to handle 'shading' option for the main action. But read shading from global config and use it only on click to set and reset the shading.

babich-a (changes requested):

babich-a (approved):

PR #9801, opened 2019-09-20
Don't select disabled items when select all is checked

mrjohn0011:


js/ui/selection/selection.strategy.js
```diff
@@ -55,6 +56,12 @@ module.exports = Class.inherit({
         return commonUtils.equalByValue(key1, key2);
     },
 
+    nonDisabled: function(items) {
```
DokaRus:
```suggestion
    getSelectableItems: function(items) {
```

DokaRus (approved):

alexander-kotov-dx (approved):

PR #9929, opened 2019-09-27
DataGrid - should focus cell after changing value of previous cell with showEditorAlways (T816039)

igormalzew:


AntonSermyazhko (approved):

JingGeGe77:
when dx-editing mode is cell，after changing value of previous cell with showEditorAlways，click the next cell to edit, can not focus current cell  （devextreme and devextreme-vue version: 19.1.7）

igormalzew:
@JingGeGe77 

To solve this problem, you need to set the repaintChangesOnly setting to true.

PR #9937, opened 2019-09-27
File Manager - Fix itemView reseting state after renaming some item

Staziz:

### DIFF
--- a/js/ui/file_manager/ui.file_manager.js
+++ b/js/ui/file_manager/ui.file_manager.js
@@ -189,8 +189,8 @@ class FileManager extends Widget {
     _initCommandManager() {
         const actions = extend(this._editing.getCommandActions(), {
             refresh: () => this._refreshAndShowProgress(),
-            thumbnails: () => this._switchView("thumbnails"),
-            details: () => this._switchView("details"),
+            thumbnails: () => this.option("itemView.mode", "thumbnails"),
+            details: () => this.option("itemView.mode", "details"),
             clear: () => this._clearSelection(),
             showNavPane: () => this._adaptivityControl.toggleDrawer()
         });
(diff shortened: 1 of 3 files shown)


js/ui/file_manager/ui.file_manager.js
```diff
@@ -258,6 +258,7 @@ class FileManager extends Widget {
         this._disposeWidget(this._itemView);
 
         this._createItemView(this._$itemsPanel, viewMode);
+        this.option("itemView.mode", viewMode);
```
vladkovl: Without additional code modifications this approach leads to one additional excessive` repaint()` method call. We should remove this excessive call in that case.

testing/helpers/fileManagerHelpers.js
```diff
@@ -129,6 +130,10 @@ export class FileManagerWrapper {
         return $(`.${Consts.DROPDOWN_MENU_LIST_CLASS} .${Consts.DROPDOWN_MENU_CONTENT_CLASS} .${Consts.DROPDOWN_MENU_LIST_ITEM_CLASS}`)[childIndex];
     }
 
+    getItemViewPopupListItem(childIndex) {
```
vladkovl: We haven't any popups in `itemView `widget. This method applies to the toolbar view switcher, so it should be reflect it in it's name.

testing/tests/DevExpress.ui.widgets/fileManagerParts/toolbar.tests.js
```diff
@@ -393,4 +393,53 @@ QUnit.module("Toolbar", moduleConfig, () => {
...
+
+        $selectBox.trigger("dxclick");
+        this.clock.tick(400);
+        detailsViewSelector = this.wrapper.getItemViewPopupListItem(0);
+        $(detailsViewSelector).trigger("dxclick");
+        this.clock.tick(400);
+
+        $selectBox = this.wrapper.getGeneralToolbarElements().last();
+        assert.equal($selectBox.val(), "Thumbnails View", "Thumbnails View");
+    });
```
vladkovl, line 439: In the entire code of this test, you can use wrapper methods instead of explicitly defining selectors.

vladkovl (changes requested):

vladkovl (approved):

PR #9978, opened 2019-10-01
Validate - small refactoring

NickMitrokhin:
dxValidator.validate and dxValidationGroup.validate methods returns copies of their internal validation result object. This is done for preventing users from mutating results on their side.


San4es (commented):
LGTM

San4es (approved):

PR #10003, opened 2019-10-03
Select preloaded items in dxTagBox with pre-grouped data (part of T756894)

mrjohn0011:
This is a fix for this problem: T756894.

**The problem**: When a user gives pregrouped data in dxTagBox's dataSource and sets a grouped option to true, tags don't render.

**The solution** Convert all loaded items to a plain structure and search items by id.

This problem is partially fixed because it is impossible to filter items in the remote dataSource when the data is pre-grouped. This PR fixes the problem only when the data is already loaded.

We cannot add pre-grouped data support to the DevExtreme DataSource now because it leads to breaking changes in DataGrid.

### DIFF
--- a/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
@@ -5421,6 +5421,43 @@ QUnit.module("regression", {
         assert.equal($.trim($tagContainer.text()), "Item1Item3", "selected values are rendered");
     });
 
+    QUnit.test("selection should work with pregrouped data without paging and with preloaded datasource", (assert) => {
+        const ds = new DataSource({
+            store: [
+                { key: "Category 1", items: [{ id: 11, name: "Item 11" }, { id: 12, name: "Item 12" }] },
+                { key: "Category 2", items: [{ id: 21, name: "Item 21" }, { id: 22, name: "Item 22" }] }
+            ]
+        });
+        ds.load();
+
+        const $tagBox = $("#tagBox").dxTagBox({
+            dataSource: ds,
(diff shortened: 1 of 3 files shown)


testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -5421,6 +5421,26 @@ QUnit.module("regression", {
...
     });
 
+    QUnit.test("selection should work with pregrouped data without paging and with preloaded datasource", (assert) => {
+        const ds = new DataSource({
+            store: [
+                { key: "Category 1", items: [{ id: 11, name: "Item 11" }, { id: 12, name: "Item 12" }] },
+                { key: "Category 2", items: [{ id: 21, name: "Item 21" }, { id: 22, name: "Item 22" }] }
+            ]
+        });
+        ds.load();
```
DokaRus, line 5431: It's very unobvious behavior. Can we avoid it?
Perhaps, we can turn off the `deferRendering` option for TagBox's Popup if editor has the initial value, so the List will immediately load the data from the DataSource

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -5421,6 +5421,47 @@ QUnit.module("regression", {
...
```
DokaRus, line 5431: So we can remove it?
```suggestion

```

mrjohn0011: No, i have created a new test without load. The TagBox should work both with preloading and without it

DokaRus (approved):

PR #10007, opened 2019-10-03
FileUploader - Add doc comments for custom upload operations

vladkovl:


js/ui/file_uploader.js
```diff
@@ -350,8 +351,41 @@ class FileUploader extends Editor {
...
+            * @type_function_return Promise<any>|any
+            */
             uploadFile: null,
+
+            /**
+            * @name dxFileUploaderOptions.uploadChunk
+            * @type function
+            * @type_function_param1 file:File
+            * @type_function_param2 chunksInfo:object
+            * @type_function_param2_field1 bytesLoaded:Number
```
RomanTsukanov: Maybe "bytes**Up**loaded"?

js/ui/file_uploader.js
```diff
@@ -350,8 +351,41 @@ class FileUploader extends Editor {
...
+            * @type_function_param2 progressCallback(loadedBytes):any
+            * @type_function_return Promise<any>|any
+            */
             uploadFile: null,
+
+            /**
+            * @name dxFileUploaderOptions.uploadChunk
+            * @type function
+            * @type_function_param1 file:File
+            * @type_function_param2 chunksInfo:object
```
RomanTsukanov: Looks like an object with information on a single chuck, not on several. Shouldn't it be "chunkInfo" without "s"?

js/ui/file_uploader.js
vladkovl: "`bytesUploaded`" is better. But "`bytesLoaded`" has already been used in `onProgress` event.

js/ui/file_uploader.js
vladkovl: This members applies to all chunks: `bytesLoaded`, `chunkCount`, `customData`. But I think it should not be a problem if we change the parameter name to `chunkInfo` and at the same time keep all this members inside this object.

RomanTsukanov (approved):

PR #10013, opened 2019-10-03
Fix HtmlEditor scrolling when it is part of Overlay or ScrollView content (T802879)

DokaRus:
We cannot develop a common solution for contenteditable elements that does not result in loss of performance, as it is necessary to validate mouse wheel events for correct operation of widgets that are part of Popup or ScrollView content.
Since the initial appeal revealed a problem with HtmlEditor, it was decided to solve the problem privately only for this widget.


San4es (approved):
LGTM

alexander-kotov-dx (approved):

PR #10025, opened 2019-10-04
Get rid of global Node

dxvladislavvolkov:


nightskylark (dismissed):

ovchinnikov (changes requested):
Don't use ES6 and commonJS module systax in same file. So, replace module.exports with export default

DokaRus (dismissed):
LGTM

PR #10061, opened 2019-10-07
Sortable: Add content template to fix transcluded content in angularjs

vconst:


js/ui/draggable.js
```diff
@@ -858,9 +871,29 @@ var Draggable = DOMComponentWithTemplate.inherit({
         return this["_" + name + "Action"] || this._createActionByOption(name);
     },
 
+    _getAnonymousTemplateName: function() {
+        return "content";
```
jtoming830, line 875: why not global variable instead of func?

jtoming830 (approved):

AntonSermyazhko (approved):

js/ui/draggable.js
vconst: This method is virtual and used in base class DOMComponentWithTemplate

PR #10064, opened 2019-10-07
Fix appearance of the DropDownButton's nested List (T818183)

DokaRus:
The DropDownButton dropdown element should have a consistent appearance with other dropdown-based widgets

generic
gen_compact
material
material_compact


alexander-kotov-dx (approved):

San4es (approved):
LGTM

PR #10078, opened 2019-10-08
Provide a way to resolve template name conflicts between inner and outer widget (T817778)

mrjohn0011:


nightskylark:
1. The name of the PR is incorrect. As we already discussed, the `anonymous` template is not a reason of a problem. We could make a named template and got the same issue.
2. I gave you the `textEditorButtonContent` name as an example. I think it should be more human-oriented name to use it for naming Button templates in a DropDownBox.
3. I think a separate test for the `DOMComponentWithTemplate` should also be written, The changes in base component shold not be testsd only in a separate widget tests.

js/ui/text_box/texteditor_button_collection/custom.js
```diff
@@ -30,7 +30,8 @@ export default class CustomButton extends TextEditorButton {
         this._addToContainer($element);
 
         const instance = editor._createComponent($element, Button, extend({}, this.options, {
-            disabled: this._isDisabled()
+            disabled: this._isDisabled(),
+            defaultTemplatesMap: { content: "customButtonContent" }
```
nightskylark: It's better to approve this name with a techwriter if we are going to publish it. Or we could use an array of template names which should not be looked up in the `integrationOptions.temapltes` instead.

testing/tests/DevExpress.core/domComponent.tests.js
```diff
@@ -1,931 +1,957 @@
...
+                        }
+                    },
+                    name2: {
+                        render() {
+                            return "Markup for name 2";
+                        }
+                    }
+                }
+            },
+            defaultTemplatesMap: { "name1": "name2" }
```
nightskylark: Now looking at this test I can see that the `defaultTemplatesMap` is not suitable name, because it maps not only default templates. Also I think we need to test the case with a default template too.

js/ui/text_box/texteditor_button_collection/custom.js
mrjohn0011: @AlexJameson agree with me

nightskylark: I am doubted about the `custom` word. Any template which customer sets is custom. But we do not use this prefix for other template names. Why we use this word here?

DokaRus (approved):

nightskylark (approved):

PR #10100, opened 2019-10-09
Overlay FIX Overlay-based widget's has an incorrect shader/overlay dimensions when target is container(T821559)

ksercs:


mrjohn0011 (dismissed):

js/ui/overlay/ui.overlay.js
```diff
@@ -1272,21 +1272,44 @@ var Overlay = Widget.inherit({
 
     _renderShading: function() {
         this._fixWrapperPosition();
+        if(this.option("shading")) {
+            this._renderShadingDimensions();
```
DokaRus: If the size of the popup is set by relative values, it will be displayed correctly only when shading is enabled:
image
image

DokaRus (dismissed):

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -2182,6 +2185,24 @@ testModule("container", moduleConfig, () => {
...
+                of: $container
+            }
+        });
+
+        $overlay.dxOverlay("show");
+
+        const $shader = $container.find(toSelector(OVERLAY_SHADER_CLASS));
+        assert.strictEqual($shader.height(), $container.height(), "shader height is correct");
+        assert.strictEqual($shader.width(), $container.width(), "shader width is correct");
+    });
```
DokaRus:
```suggestion
    [false, true].forEach((shading) => {
        test(`wrapper should stretch across container when target is container and shading is ${shading}`, (assert) => {
            const $container = $("#containerT821559");

            const $overlay = $("#overlay").dxOverlay({
                container: $container,
                shading,
                position: {
                    of: $container
                }
            });

            $overlay.dxOverlay("show");
            
            const $shader = $container.find(toSelector(OVERLAY_SHADER_CLASS));
            assert.strictEqual($shader.height(), $container.height(), "shader height is correct");
            assert.strictEqual($shader.width(), $container.width(), "shader width is correct");
        });
    });
```

js/ui/overlay/ui.overlay.js
```diff
@@ -100,6 +100,10 @@ var getElement = function(value) {
     return value && $(value.target || value);
 };
 
+function isPercent(size) {
+    return !!size && (size.toString().slice(-1) === '%');
```
EugeniyKiyashko: Try to use "../core/utils/size"

ksercs: There is no enough functionality, so I supplemented it

styles/widgets/common/drawer.less
```diff
@@ -73,6 +73,14 @@
     }
 }
 
+.dx-rtl .dx-drawer-slide,
```
DokaRus: This styles are specific for the Drawer strategy that uses Overlay. So let's specify it:

```css
.dx-rtl .dx-drawer-overlap.dx-drawer-slide,
.dx-rtl.dx-drawer-overlap.dx-drawer-slide
```

ksercs: Got it, thanks

San4es (approved):
LGTM

mrjohn0011 (approved):

PR #10127, opened 2019-10-10
Validation - Adds extra parameters to validationCallback

NickMitrokhin:
validationCallback accepts the following extra params:

1. **column** - the current column to which the validator belongs. Exists only when you validate a DataGrid or TreeList cell's value.

2. **formItem** - the current Form item to which the validator belongs. Exists only when you validate a Form editor.


DokaRus (approved):
lgtm

San4es (approved):
🙏

PR #10161, opened 2019-10-11
List: Fix item selection after selected item removing if repaintChanges is true (T821093)

San4es:
Fixes T821093.


DokaRus (approved):
LGTM

mrjohn0011 (approved):

PR #10181, opened 2019-10-14
Add dependencies to themebuilder metadata

babich-a:

### DIFF
--- a/themebuilder/modules/metadata-generator.js
+++ b/themebuilder/modules/metadata-generator.js
@@ -111,7 +111,7 @@ const dependenciesPlugin = (dependencies) => {
                             dependencies[widget] = [];
                         }
 
-                        dependencies[widget] = dependencies[widget].concat(parseImports(less));
+                        dependencies[widget].push(...parseImports(less));
                     }
                     return less;
                 }
--- a/themebuilder/modules/metadata-generator.js
+++ b/themebuilder/modules/metadata-generator.js
@@ -4,66 +4,158 @@ const fs = require("fs");
 
 const themes = require("./themes");
 const getBundleName = require("./bundle-resolver");
+const ModulesHandler = require("../modules/modules-handler");
+
+const stylesDirectory = path.join(__dirname, "..", "..", "styles");
+const themesFileContent = fs.readFileSync(path.join(stylesDirectory, "theme.less"), "utf8");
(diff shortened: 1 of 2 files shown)


themebuilder/modules/metadata-generator.js
```diff
@@ -4,66 +4,158 @@ const fs = require("fs");
...
+                process: (less, context) => {
+                    const fullPath = context && context.fileInfo && context.fileInfo.filename;
+
+                    if(typeof fullPath === "string") {
+                        const widget = getWidgetFromFileName(path.basename(fullPath));
+                        if(!dependencies[widget]) {
+                            dependencies[widget] = [];
+                        }
+
+                        dependencies[widget] = dependencies[widget].concat(parseImports(less));
```
BingoRUS:
```suggestion
                        dependencies[widget].push(...parseImports(less));
```

themebuilder/modules/metadata-generator.js
```diff
@@ -4,66 +4,158 @@ const fs = require("fs");
...
+};
+
+const resolveDependencies = (dependencies) => {
+    for(var widget in dependencies) {
+        if(Object.prototype.hasOwnProperty.call(dependencies, widget)) {
+            const widgetDependencies = dependencies[widget];
+
+            for(let i = 0; i < widgetDependencies.length; i++) {
+                const dependencyWidget = widgetDependencies[i];
+                const dependencyList = dependencies[dependencyWidget];
```
BingoRUS, line 67: Shouldn't it be recursive?

themebuilder/modules/metadata-generator.js
```diff
@@ -4,66 +4,158 @@ const fs = require("fs");
...
+            if(selfIndex >= 0) {
+                widgetDependencies.splice(selfIndex, 1);
+            }
+        }
+    }
+
+    return dependencies;
+};
+
+const removeDependenciesInternals = (dependencies) => {
```
BingoRUS: I think `removeInternalDependencies` is a better method name.

themebuilder/modules/metadata-generator.js
babich-a: No. We add the dependency to the end of array and then in any case process it. Theoretically, the algorithm should recursive (it is a tree), but our tree has circle dependencies and recursion can cause the trouble :)

BingoRUS (approved):

PR #10266, opened 2019-10-20
Fix icons alignment in List

BingoRUS:


BingoRUS:
Vertical align is different because icons are smaller than text in Generic but bigger than text in Material.

San4es (approved):
LGTM

PR #10309, opened 2019-10-23
TextEditor FIX: The cursor jumps when typing if the mask is set and maskChar is empty (T825309)

ksercs:


San4es (changes requested):
Equating maskСhar = "" to maskСhar = " " is not a good idea - these are 2 different cases.

PR #10313, opened 2019-10-23
Call onValueChanged callback only once when value option is changed in the dxRangeSlider (T823877)

mrjohn0011:


alexander-kotov-dx (approved):

San4es (approved):
LGTM

PR #10326, opened 2019-10-24
DateBox: Fix native pickers showing if the drop-down button is clicked on a mobile device (T824701)

San4es:
**In 18.2:**

**In 19.1:**

**After this fix in 19.1:**
fix

### DIFF
--- a/styles/widgets/common/dateBox.less
+++ b/styles/widgets/common/dateBox.less
@@ -9,6 +9,15 @@
 .dx-datebox-native {
     .dx-texteditor-buttons-container {
         pointer-events: none;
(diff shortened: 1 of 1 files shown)


alexander-kotov-dx:
It breaks styles for the cases with custom buttons

alexander-kotov-dx (changes requested):
It breaks styles for the cases with custom buttons

San4es:
@alexander-kotov-dx Thank you!

San4es:
@alexander-kotov-dx I have prepared an alternative fix (#10353). Please review it.

PR #10403, opened 2019-10-31
Renovation: dxButton es6 refactoring

churkin:


pavelgruba (dismissed):

js/ui/button.js
```diff
@@ -249,243 +316,126 @@ var Button = Widget.inherit({
...
     }
+}
 
-}).include(ValidationMixin);
+Button.include(ValidationMixin);
 
-registerComponent("dxButton", Button);
+registerComponent('dxButton', Button);
 
 module.exports = Button;
```
MaximKudriavtsev, line 441: Can we use `export class Button ...` instead of `module.exports = Button`?

MaximKudriavtsev (approved):

PR #10452, opened 2019-11-06
DataGrid: Fix row dragging when there is master-detail (T828366)

vconst:


js/ui/grid_core/ui.grid_core.row_dragging.js
```diff
@@ -48,7 +48,7 @@ var RowDraggingExtender = {
             that._sortable = that._createComponent($content, Sortable, extend({
                 component: that.component,
                 contentTemplate: null,
-                filter: "> table > tbody > .dx-data-row",
+                filter: "> table > tbody > .dx-row:not(.dx-freespace-row):not(.dx-virtual-row)",
```
Alyar666, line 51: dx-error-row?

vconst: The error row does not have dx-row class, it has only dx-error-row class.

Alyar666 (approved):

PR #10453, opened 2019-11-06
Default item template of the CollectionWidget should correctly handle zero or empty string values (T827713)

DokaRus:


alexander-kotov-dx (approved):

ksercs (approved):

San4es (commented):
👍

PR #10465, opened 2019-11-07
Move init validation to constructor of editor

dxvladislavvolkov:


DokaRus (dismissed):
How about a small refactoring - replace `var` with `let` or `const`?

dxvladislavvolkov:
@DokaRus fixed

DokaRus (approved):

AlexJameson (approved):

PR #10466, opened 2019-11-07
DataGrid - the form editor should not lose focus when changing its value in setCellValue (T822877)

igormalzew:


js/ui/grid_core/ui.grid_core.column_fixing.js
```diff
@@ -762,7 +762,7 @@ var RowsViewFixedColumnsExtender = extend({}, baseFixedColumns, {
             columns,
             cellIndex = 0;
 
-        if($fixedTable && $fixedTable.find($cell).length) {
+        if($fixedTable && $fixedTable.find($cell).length && $cell.is("td")) {
```
AntonSermyazhko: Maybe change the order?
$cell.is("td") && $fixedTable && $fixedTable.find($cell).length

AntonSermyazhko (approved):

vconst (approved):

PR #10497, opened 2019-11-11
Remove unnecessary doc comments for the reset() method (T823490)

San4es:


AlexJameson (dismissed):

DokaRus (approved):

ts/dx.all.d.ts
```diff
@@ -3378,10 +3378,6 @@ declare module DevExpress.ui {
         field(): DevExpress.core.dxElement;
         /** @name dxDropDownEditor.open() */
         open(): void;
-        /** @name dxDropDownEditor.reset() */
-        reset(): void;
-        /** @name Component.resetOption(optionName) */
-        resetOption(optionName: string): void;
```
San4es, line 3384: @IlyaKhD  have you any ideas why the resetOption method is removed from TS definitions as well?

IlyaKhD (approved):

ts/dx.all.d.ts
San4es: It was a bug in devextreme-internal-tools.

PR #10514, opened 2019-11-12
Renovation: Options manager refactoring

churkin:
Extract most of the option logic from the `component` to the `option manager`.  It will be helpful for MVVM model separation.

### DIFF
--- a/js/core/component.js
+++ b/js/core/component.js
@@ -1,15 +1,20 @@
-import Action from './action';
-import Callbacks from './utils/callbacks';
-import Config from './config';
-import Class from './class';
-import errors from './errors';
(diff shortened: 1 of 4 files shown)


MaximKudriavtsev (approved):

nightskylark:
@churkin Here are a lot of changes. Maybe you could describe them and main goals of this refactoring in a description?

churkin:
@nightskylark 
Extract most of the option logic from the `component` to the `option manager`. It will be helpful for MVVM model separation.

nightskylark:
Discussed personally with @churkin and decided it's better to avoid a cohession with the `device` module.

LazyLahtak (approved):

PR #10524, opened 2019-11-12
Renovation: EventsMixin and EventsStrategy refactor

LazyLahtak:
Remove `Events Mixin`, refactor `Events Strategy` and all components which use `Events Mixin`.


js/core/events_strategy.js
```diff
@@ -1,30 +1,40 @@
...
         if(callbacks) {
             return callbacks.has();
         }
         return false;
-    },
+    }
 
-    fireEvent: function(eventName, eventArgs) {
+    fireEvent(eventName, eventArgs) {
         var callbacks = this._events[eventName];
```
MaximKudriavtsev: Can we use es6 variable types?

`var` -> `const`

js/core/events_strategy.js
```diff
@@ -1,55 +1,65 @@
...
         }
-    },
+    }
 
-    dispose: function() {
+    dispose() {
         each(this._events, function() {
             this.empty();
         });
     }
```
MaximKudriavtsev:
```js
each(this._events, () => {
  this.empty();
});
```

js/data/abstract_store.js
```diff
@@ -309,21 +310,61 @@ var Store = Class.inherit({
...
+    /**
+     * @name StoreMethods.off
+     * @publicName off(eventName, eventHandler)
+     * @param1 eventName:string
+     * @param2 eventHandler:function
+     * @return this
+     */
+    off: function(eventName, eventHandler) {
+        this._eventsStrategy.off(eventName, eventHandler);
+        return this;
```
MaximKudriavtsev: Seems like we should move these methods into EventsStrategy.

And use here something like that

```js
off: function(eventName, eventHandler) {
  this._eventsStrategy.offMethod2(eventName, eventHandler);
  return this;
}
```

testing/tests/DevExpress.core/component.tests.js
```diff
@@ -1442,12 +1442,12 @@ QUnit.module("event API", {
         component.on("event", h1);
         component.on("event", h2);
 
-        component.fireEvent("event");
+        component._eventsStrategy.fireEvent("event");
```
MaximKudriavtsev, line 1445: Should we change the previous classes' API? I mean, we don't change tests exclude events files.

LazyLahtak: The `fireEvent` was a private method, but in some tests it is checked.

LazyLahtak: We decided to remove it from classes and call inside. Should we back it?

churkin: I think it is ok to get rid of `fireEvent` method. But events mixin is a public class. Is it braking change? @NickMitrokhin

js/ui/validation_engine.js
```diff
@@ -751,8 +752,22 @@ const GroupConfig = Class.inherit({
         });
         this._pendingValidators = [];
         this._resetValidationInfo();
-    }
-}).include(EventsMixin);
+    },
+
+    on: function(eventName, eventHandler) {
```
churkin: `on(eventName, eventHandler) {`

js/ui/validation_engine.js
```diff
@@ -751,8 +752,22 @@ const GroupConfig = Class.inherit({
...
+        if(isPlainObject(eventName)) {
+            each(eventName, (function(e, h) { this.on(e, h); }).bind(this));
```
churkin: use arrow functions

js/ui/pivot_grid/data_source.js
```diff
@@ -1723,6 +1723,47 @@ module.exports = Class.inherit((function() {
...
+         * @return this
+         */
+        /**
+         * @name PivotGridDataSourceMethods.on
+         * @publicName on(events)
+         * @param1 events:object
+         * @return this
+         */
+        on: function(eventName, eventHandler) {
+            if(isPlainObject(eventName)) {
```
churkin: Should we move this logic to the event strategy?

js/core/events_strategy.js
churkin:
```javascript
each(this._events, event => event.empty());
``` 
:)

js/core/events_strategy.js
```diff
@@ -1,55 +1,65 @@
...
-    ctor: function(owner) {
+export class EventsStrategy {
+    constructor(owner) {
         this._events = {};
         this._owner = owner;
-    },
+    }
+    static setEventsStrategy(owner, strategy) {
+        if(strategy) {
+            if(typeof strategy === "function") {
```
churkin:
```javascript
return isFunction(strategy) ? strategy(owner) : strategy;
```

js/core/events_strategy.js
```diff
@@ -1,55 +1,65 @@
...
+        } else {
+            return new EventsStrategy(owner);
+        }
+    }
 
-    hasEvent: function(eventName) {
-        var callbacks = this._events[eventName];
+    hasEvent(eventName) {
+        const callbacks = this._events[eventName];
         if(callbacks) {
```
churkin:
```javascript
return callbacks ? callbacks.has() : false;
```

churkin (changes requested):

testing/tests/DevExpress.core/component.tests.js
vconst: The fireEvent method is not public, but some users use it. For example, https://supportcenter.devexpress.com/ticket/details/T656964/customstore-how-to-trigger-the-modified-event

churkin: It is not legal

MaximKudriavtsev (approved):

testing/tests/DevExpress.core/component.tests.js
NickMitrokhin: This is an undocumented method. So, we can change it according to our needs without any notifications.

churkin (approved):

churkin:
@nightskylark ping

nightskylark (commented):
Maybe it's better to rename `EventsStrategy` to avoid confusion with the [DOM events](https://github.com/DevExpress/DevExtreme/blob/20_1/js/events.js). As I can see in fact it's a callbacks object not events.

LazyLahtak:
> Maybe it's better to rename `EventsStrategy` to avoid confusion with the [DOM events](https://github.com/DevExpress/DevExtreme/blob/20_1/js/events.js). As I can see in fact it's a callbacks object not events.

`CallbacksStrategy` or `CallbacksManager`?
Should we also rename methods? (`fireEvent` -> `fireCallback` or `fire`) (same for `hasEvent`)

nightskylark:
> > Maybe it's better to rename `EventsStrategy` to avoid confusion with the [DOM events](https://github.com/DevExpress/DevExtreme/blob/20_1/js/events.js). As I can see in fact it's a callbacks object not events.
> 
> `CallbacksStrategy` or `CallbacksManager`?
> Should we also rename methods? (`fireEvent` -> `fireCallback` or `fire`) (same for `hasEvent`)

It's just thoughts not a change request. I'm not crearly understand what is this module for so I'm not sure. As I can see these events are linked with owner instance. Maybe it could be something like `InstanceEvents` or `InstanceCallbacks`. Ofcourse in case we deside to get rid of `event` word we should do it everywhere.

nightskylark:
Or maybe `ComponentXxx`.

LazyLahtak:
> > > Maybe it's better to rename `EventsStrategy` to avoid confusion with the [DOM events](https://github.com/DevExpress/DevExtreme/blob/20_1/js/events.js). As I can see in fact it's a callbacks object not events.
> > 
> > 
> > `CallbacksStrategy` or `CallbacksManager`?
> > Should we also rename methods? (`fireEvent` -> `fireCallback` or `fire`) (same for `hasEvent`)
> 
> It's just thoughts not a change request. I'm not crearly understand what is this module for so I'm not sure. As I can see these events are linked with owner instance. Maybe it could be something like `InstanceEvents` or `InstanceCallbacks`. Ofcourse in case we deside to get rid of `event` word we should do it everywhere.

This is old `EventStrategy`. I just got rid of `EventMixin` and replace functional into components and `EventStrategy`. `EventStrategy` contains callbacks (events) and fire it (for example `disposing` or `optionChanged`)

LazyLahtak:
@nightskylark,

As I see, with rename of the`eventsStrategy` we have to change (rename) some elements. I think we should do it in other PR.

DokaRus (approved):

PR #10581, opened 2019-11-14
Renovation: fix OptionManager

LazyLahtak:
Replace `for ... of` by `for` cycle

### DIFF
--- a/js/core/option_manager.js
+++ b/js/core/option_manager.js
@@ -104,8 +104,8 @@ export class OptionManager {
 
     _normalizeName(name) {
         if(name) {
-            for(let deprecateName of this._deprecateNames) {
-                if(deprecateName === name) {
+            for(let i = 0; i < this._deprecateNames.length; i++) {
+                if(this._deprecateNames[i] === name) {
                     const deprecate = this._deprecated[name];
 
                     if(deprecate) {


js/core/option_manager.js
```diff
@@ -104,8 +104,8 @@ export class OptionManager {
 
     _normalizeName(name) {
         if(name) {
-            for(let deprecateName of this._deprecateNames) {
-                if(deprecateName === name) {
+            for(let i = 0; i < this._deprecateNames.length; i++) {
```
MaximKudriavtsev, line 107: Maybe change this logic by using something like `reduce()`?

LazyLahtak: We break cycle using `return`. Does `reduce()` fit to us in this case?

churkin (approved):

js/core/option_manager.js
MaximKudriavtsev: Maybe `find` can help us

LazyLahtak: I did't find any use of `Array.find` in code. It seems that it may not work without polyfills (like `for...of` cycle)

LazyLahtak: https://forum.vuejs.org/t/using-polyfill-array-find-for-ie11/40493

MaximKudriavtsev (approved):

PR #10593, opened 2019-11-14
DataGrid: Fix load panel visibility after fixing and unfixing column when the infinite scrolling is enabled (T829901)

Alyar666:
See https://devexpress.com/issue=T829901

### DIFF
--- a/js/ui/grid_core/ui.grid_core.column_fixing.js
+++ b/js/ui/grid_core/ui.grid_core.column_fixing.js
@@ -117,7 +117,7 @@ var baseFixedColumns = {
 
             that._isFixedTableRendering = false;
         } else {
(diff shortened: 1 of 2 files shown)


js/ui/grid_core/ui.grid_core.column_fixing.js
```diff
@@ -89,6 +89,13 @@ var baseFixedColumns = {
         }
     },
 
+    _removeFixedContent: function() {
+        let $element = $(this.element()),
+            $fixedContent = $element && $element.find(`.${this.addWidgetPrefix(CONTENT_FIXED_CLASS)}`);
```
AntonSermyazhko: I guess $element check is unnecessary

vconst (approved):

igormalzew (approved):

PR #10609, opened 2019-11-14
fix 'subclassOf' method for the es6 inheritors

churkin:


AlekseyMartynov (dismissed):
LGTM

testing/tests/DevExpress.core/class.tests.js
```diff
@@ -1,4 +1,4 @@
-var Class = require("core/class");
+import Class from 'core/class';
```
ovchinnikov: I suggest to use double quotes instead of single because all quotes in this file are double

ovchinnikov (changes requested):

testing/tests/DevExpress.core/class.tests.js
ovchinnikov: Same issue with quotes in other changed files

churkin:
FPR

ovchinnikov (approved):

nightskylark (approved):

AlekseyMartynov (approved):

PR #10638, opened 2019-11-15
Renovation: extract TemplateManager module

MaximKudriavtsev:


js/core/template_manager.js
```diff
@@ -0,0 +1,276 @@
...
+            const deviceTemplate = TemplateManager._findTemplateByDevice(templatesMap[templateName]);
+            if(deviceTemplate) {
+                this.saveTemplate(templateName, deviceTemplate);
+            }
+        }
+    }
+
+    saveTemplate(name, template) {
+        const templates = this.option('integrationOptions.templates');
+        templates[name] = this.createTemplate(template);
```
LazyLahtak: I tried to use
```this.option(`integrationOptions.templates.${name}`, this.createTemplate(template));```
instead and seems that it work

File: js/core/dom_component_with_template.js [hunk 2]
```diff
@@ -1,290 +1,75 @@
...
+    _getAnonymousTemplateName: function() {
+        return undefined;
     },
 
     _init: function() {
         this.callBase();
 
-        this._tempTemplates = [];
-        this._defaultTemplates = {};
+        const integrationOptions = this.option('integrationOptions');
```
churkin: const { integrationOptions } = this.option();

js/core/dom_component_with_template.js
```diff
@@ -1,290 +1,75 @@
...
-                    }
-                    return noop;
-                },
-                templates: { "dx-polymorph-widget": DX_POLYMORPH_WIDGET_TEMPLATE },
-                createTemplate: function(element) {
-                    return new Template(element);
-                }
-            }
-        });
+        return extend(this.callBase(), TemplateManager.getDefaultOptions());
```
churkin: create static getter

js/ui/defer_rendering.js
```diff
@@ -83,6 +83,10 @@ var DeferRendering = Widget.inherit({
         return ANONYMOUS_TEMPLATE_NAME;
     },
 
+    _initTemplates: function() {
```
ovchinnikov: It seems that this function is unnecessary

ovchinnikov (commented):
All templates are defined via square bracket and double quotes like this { ["group"]: value }. Why don't use just simple { group: value }?

testing/tests/DevExpress.ui.widgets/responsiveBox.markup.tests.js
```diff
@@ -480,18 +480,15 @@ QUnit.module("templates", () => {
...
+        const $responsiveBox = $("#responsiveBox").dxResponsiveBox({
+            rows: [{}],
+            cols: [{}],
+            itemTemplate: function() {
+                return $("<div>test</div>");
+            },
+            items: [{ location: { row: 0, col: 0 } }]
+        });
 
-        assert.equal(templateContext, "dxResponsiveBox", "Correct context");
```
ovchinnikov: I think here is Breaking Change

MaximKudriavtsev: We discussed this point a week ago and decided that it is not a BC.
- we don't have mentions of this point in docs;
- we don't have tests except for current;
- this conception is not correct by ideology

testing/tests/DevExpress.ui.widgets.editors/dropDownEditor.tests.js
```diff
@@ -1049,7 +1049,7 @@ QUnit.test("should have no errors after value change if text editor buttons were
         value: 1,
         fieldTemplate(value) {
             const $textBox = $("<div>").dxTextBox();
-            return $("<div>").text(value + this.option("value")).append($textBox);
+            return $("<div>").text(value + value).append($textBox);
```
ovchinnikov: This change completely break idea of this test. What do you check after this change??

ovchinnikov (dismissed):

testing/tests/DevExpress.ui.widgets.editors/dropDownEditor.tests.js
ovchinnikov: Ok, It seems I got it. There is no differece for this test what text will be in the template. So I suggest remove value + value and just use simple value

MaximKudriavtsev: I have reverted all similar tests. Without any BC

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+
+    _createTemplate(templateSource) {
+        return this._createElement(TemplateManager.validateTemplateSource(templateSource));
+    }
+
+    getTemplate(templateSource, templates, { isAsyncTemplate, skipTemplates }, context) {
+        if(isFunction(templateSource)) {
+            return new FunctionTemplate((options) => {
+                const templateSourceResult = templateSource.apply(context, TemplateManager._getNormalizedTemplateArgs(options));
+                // const templateSourceResult = templateSource(TemplateManager._getNormalizedTemplateArgs(options)); // accordion tests
```
vconst: Is this comment necessary?

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+        parent._createComponent(widgetElement, widgetName, widgetOptions);
+    } else {
+        widgetElement[widgetName](widgetOptions);
+    }
+
+    return widgetElement;
+});
+const defaultCreateElement = element => new Template(element);
+
+export default class TemplateManager {
```
IlyaKhD, line 43: I suggest move all `static`'s outside the class and use named exports where necessary.

A class is a factory for creating instances, each instance is a set of data fields (state) and methods, performing read\write operation on these fields. If a method doesn't depend on class data - it seems to be better extracted as an external function.

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+import { getElementOptions, normalizeTemplateElement } from './utils/dom';
+import devices from './devices';
+import { Template } from './templates/template';
+import { TemplateBase } from './templates/template_base';
+import { FunctionTemplate } from './templates/function_template';
+import { EmptyTemplate } from './templates/empty_template';
+import { ChildDefaultTemplate } from './templates/child_default_template';
+import { camelize } from './utils/inflector';
+
+const TEXT_NODE = 3;
```
IlyaKhD, line 16: Suggestion: move all constants to the end of the file.

I think the main part of this module is the `TemplateManager` class and it should go first

js/ui/filter_builder/filter_builder.js
```diff
@@ -587,7 +587,7 @@ var FilterBuilder = Widget.inherit({
     },
 
     _initCustomOperations: function() {
-        this._customOperations = utils.getMergedOperations(this.option("customOperations"), this.option("filterOperationDescriptions.between"));
+        this._customOperations = utils.getMergedOperations.bind(this)(this.option("customOperations"), this.option("filterOperationDescriptions.between"));
```
vconst: Why? The getMergedOperations method is static and it does not use this keyword. So it is very strange to bind it to this.

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+const TEMPLATE_WRAPPER_CLASS = 'dx-template-wrapper';
+const DEPRECATED_WIDGET_NAMES = ['button', 'tabs', 'dropDownMenu'];
+const DX_POLYMORPH_WIDGET_TEMPLATE = new FunctionTemplate(({ model, parent }) => {
+    let widgetName = model.widget;
+    if(!widgetName) return $();
+
+    const widgetElement = $('<div>');
+    const widgetOptions = model.options || {};
+
+    if(DEPRECATED_WIDGET_NAMES.some(deprecatedWidgetName => deprecatedWidgetName === widgetName)) {
```
IlyaKhD:
```suggestion
    if(DEPRECATED_WIDGET_NAMES.has(widgetName)) {
```
`Set` better fits here

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+import { FunctionTemplate } from './templates/function_template';
+import { EmptyTemplate } from './templates/empty_template';
+import { ChildDefaultTemplate } from './templates/child_default_template';
+import { camelize } from './utils/inflector';
+
+const TEXT_NODE = 3;
+const ANONYMOUS_TEMPLATE_NAME = 'template';
+const TEMPLATE_SELECTOR = '[data-options*="dxTemplate"]';
+const TEMPLATE_WRAPPER_CLASS = 'dx-template-wrapper';
+const DEPRECATED_WIDGET_NAMES = ['button', 'tabs', 'dropDownMenu'];
```
IlyaKhD: As for me I used to using CAPS_NAMES only for strings and numbers, not complex entities like objects\functions\arrays erc

js/ui/filter_builder/filter_builder.js
vconst: Looks like hack to bind editorTemplate to this. It will be batter pass this as parameter to getMergedOperations method.

testing/tests/DevExpress.ui.widgets/box.markup.tests.js
```diff
@@ -94,17 +94,12 @@ QUnit.module("layouting", () => {
...
-                if(this.option("_layoutStrategy") === "flex") {
-                    assert.equal(this._$element[0].style.flexDirection, "column");
-                } else {
-                    assert.ok(true);
-                }
-            }
-        });
+        }).instance();
+
+        assert.equal($(boxInstance._$element)[0].style.flexDirection, "column");
```
vconst: The _layoutStrategy is fallback for IE. Is this test checked in IE?

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+
+    static templateKey(templateSource) {
+        return (isRenderer(templateSource) && templateSource[0]) || templateSource;
+    }
+
+    get anonymousTemplateName() {
+        return this._anonymousTemplateName;
+    }
+
+    addDefaultTemplate(template) {
```
vconst: Why is name not addDefaultTemplates? The method can get many templates throw template parameter. The template parameter is object map here with templates. So it will be better rename addDefaultTemplate to addDefaultTemplates and template to templates.

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+        });
+        const onlyJunkTemplateContent = $notJunkTemplateContent.length < 1;
+
+        return !onlyJunkTemplateContent
+            ? { template: this._createTemplate($anonymousTemplate), name: this._anonymousTemplateName }
+            : {};
+    }
+
+    _createTemplateIfNeeded(templateSource) {
+        const cachedTemplate = this._tempTemplates.filter(tempTemplate =>
```
IlyaKhD, line 137:
```suggestion
        const cachedTemplate = this._tempTemplates.find(tempTemplate =>
```

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+        this._tempTemplates.push({ template, source: TemplateManager.templateKey(templateSource) });
+        return template;
+    }
+
+    _createTemplate(templateSource) {
+        return this._createElement(TemplateManager.validateTemplateSource(templateSource));
+    }
+
+    getTemplate(templateSource, templates, { isAsyncTemplate, skipTemplates }, context) {
+        if(isFunction(templateSource)) {
```
IlyaKhD: Invert condition to decrease main block indetation
```suggestion
        if(!isFunction(templateSource)) {
          return this._acquireTemplate(
            templateSource,
            this._createTemplateIfNeeded.bind(this),
            templates,
            isAsyncTemplate,
            skipTemplates
            );
          }
          
          return new FunctionTemplate((options) => {
          
          ...
```

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+                    return this._createTemplate(templateSource);
+                }, templates, isAsyncTemplate, skipTemplates);
+
+                const result = template.render(options);
+                dispose && template.dispose && template.dispose();
+                return result;
+            });
+        }
+
+        return this._acquireTemplate(templateSource, this._createTemplateIfNeeded.bind(this), templates, isAsyncTemplate, skipTemplates);
```
IlyaKhD: Move binding to `ctor` (no need to create function object each time)

IlyaKhD: As I got it, the last 3 args are used only by the `_acquireStringTemplate`. If so - better modify the calls to `_acquireTemplate` and `_acquireStringTemplate` in this way (or smth like):

```
return this._acquireTemplate(templateSource, this._createTemplateIfNeeded.bind(this)) ||
       this._acquireStringTemplate(templateSource, templates, isAsyncTemplate, skipTemplates) ||
       this._defaultTemplates[templateSource] ||
       createTemplate(templateSource);
```

May be some methods renaming will be required (`_acquireStringTemplate`)

IlyaKhD (changes requested):

js/core/template_manager.js
MaximKudriavtsev: In my opinion.
A Class is a set of methods related to some functionality. If a static method use only with current functionality it should be placed inside the method.

js/core/template_manager.js
```diff
@@ -0,0 +1,277 @@
...
+        }
+
+        if(templateSource.nodeType || isRenderer(templateSource)) {
+            return createTemplate($(templateSource));
+        }
+
+        return this._acquireStringTemplate(templateSource, createTemplate, templates, isAsyncTemplate, skipTemplates);
+    }
+
+    _renderIntegrationTemplate(templateSource, templates, isAsyncTemplate) {
```
IlyaKhD:
```suggestion
    _renderIntegrationTemplate(integrationTemplate, isAsyncTemplate) {
```
as `templateSource` and `templates` are used only for `const integrationTemplate = templates[templateSource];`

IlyaKhD (dismissed):

js/core/template_manager.js
MaximKudriavtsev: I know that the `eslint` has a special rule - [no-use-before-define](no-use-before-define) and many eslint plugins have this rule. I agree with this.

But it should depend on the team's code guidelines.

js/core/template_manager.js
MaximKudriavtsev: Find is not working on IE

testing/tests/DevExpress.ui.widgets/box.markup.tests.js
MaximKudriavtsev: I have reverted this test back

js/core/template_manager.js
MaximKudriavtsev: It uses as a predefined constant too as your `string` and `number` constants.

js/core/template_manager.js
```diff
@@ -0,0 +1,270 @@
...
+        return typeof templateSource === 'string'
+            ? normalizeTemplateElement(templateSource)
+            : templateSource;
+    }
+
+    static templateKey(templateSource) {
+        return (isRenderer(templateSource) && templateSource[0]) || templateSource;
+    }
+
+    get anonymousTemplateName() {
```
vconst, line 67: It will be better to have method getAnonymousTemplateName because:
1) It will be say by name that it is read only.
2) It will be impossible to make typo error.
3) The getter is magic, method is better to understand.

js/ui/drawer/ui.drawer.js
```diff
@@ -250,7 +251,7 @@ const Drawer = Widget.inherit({
 
         const contentTemplateOption = this.option("contentTemplate"),
             contentTemplate = this._getTemplate(contentTemplateOption),
-            transclude = this._getAnonymousTemplateName() === contentTemplateOption;
+            transclude = this._templateManager.anonymousTemplateName === contentTemplateOption;
```
vconst, line 254: Can we add the isTranscludedTemplate(templateOptionValue) method to _templateManager?
Similar expression:
transclude = this._templateManager.anonymousTemplateName === contentTemplateOption
exists in 5 places.

js/core/template_manager.js
MaximKudriavtsev: When we use any `method` we think that it calculates something and has some logic inside. When we use any `getter` we know that it don't calculate anything and only returns value of a static field.

js/ui/drawer/ui.drawer.js
MaximKudriavtsev: I think no because this method will show only that template is an anonymous template. It is not a greate logic that should be written in the TemplateManager API and this method will be created on each instance creation, this is bad. Current API is little and flexible.

js/core/template_manager.js
IlyaKhD: I guess the thing you described is a module. That exports functions 😉

js/core/template_manager.js
IlyaKhD: What about adding a polyfill?

js/core/template_manager.js
```diff
@@ -0,0 +1,221 @@
...
+    constructor(createElement, anonymousTemplateName) {
+        this._tempTemplates = [];
+        this._defaultTemplates = {};
+        this._anonymousTemplateName = anonymousTemplateName || ANONYMOUS_TEMPLATE_NAME;
+
+        this._createElement = createElement || defaultCreateElement;
+        this._createTemplateIfNeeded = this._createTemplateIfNeeded.bind(this);
+    }
+
+    static get defaultOptions() {
```
IlyaKhD: This seems to be a method, as it returns a new object on each call. `buildDefaultOptions` or smth like.
@vconst, @MaximKudriavtsev I see your discussion below on the topic "getter vs method" and this one seems to me to be a method

js/core/dom_component_with_template.js
```diff
@@ -1,290 +1,76 @@
...
+        const skipTemplates = this.option('integrationOptions.skipTemplates');
+
+        return this._templateManager.getTemplate(
+            templateSource,
+            templates,
+            {
+                isAsyncTemplate,
+                skipTemplates
+            },
+            that
```
IlyaKhD: I guess that can be replaced with `this`, as there is no closure and the argument is passed directly

js/core/template_manager.js
```diff
@@ -0,0 +1,221 @@
...
+
+        const $notJunkTemplateContent = $anonymousTemplate.filter((_, element) => {
+            const isTextNode = element.nodeType === TEXT_NODE;
+            const isEmptyText = $(element).text().trim().length < 1;
+
+            return !(isTextNode && isEmptyText);
+        });
+        const onlyJunkTemplateContent = $notJunkTemplateContent.length < 1;
+
+        return !onlyJunkTemplateContent
```
IlyaKhD:
```suggestion
        return $notJunkTemplateContent.length > 0
```
Seems to be clear enough

js/core/template_manager.js
```diff
@@ -0,0 +1,221 @@
...
+        // TODO: templateSource.render is needed for angular2 integration. Try to remove it after supporting TypeScript modules.
+        if(isFunction(templateSource.render) && !isRenderer(templateSource)) {
+            return addOneRenderedCall(templateSource);
+        }
+
+        if(templateSource.nodeType || isRenderer(templateSource)) {
+            return createTemplate($(templateSource));
+        }
+
+        return this._acquireStringTemplate(templateSource, createTemplate, templates, isAsyncTemplate, skipTemplates);
```
IlyaKhD: What about this?
```suggestion
        return this._acquireIntegrationTemplate(templateSource, templates, isAsyncTemplate, skipTemplates)
            || this._defaultTemplates[templateSource]
            || createTemplate(templateSource);
```
One arg less

js/core/template_manager.js
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    _acquireStringTemplate(templateSource, createTemplate, templates, isAsyncTemplate, skipTemplates) {
+        const nonIntegrationTemplates = skipTemplates || [];
+        let integrationTemplate = null;
+
+        if(nonIntegrationTemplates.indexOf(templateSource) === -1) {
```
IlyaKhD: No need for creating an array each time
```suggestion
        if(!skipTemplates || skipTemplates.indexOf(templateSource) === -1) {
```

IlyaKhD (approved):

churkin (approved):

vconst (approved):

LazyLahtak (approved):

PR #10654, opened 2019-11-18
Renovation: refactor OptionManager and singled out Options

LazyLahtak:
Separate PostponedOperations and Component.
Singled out Options into separated class (now use it unstead OptionManager).

Note: `component`'s (and inheritors') fields `_optionManager` and `_option` will be refactored in other PR.


js/core/options/index.js
```diff
@@ -0,0 +1,223 @@
...
+    //         const match = isFunction(deviceFilter) ?
+    //             deviceFilter(currentDevice) :
+    //             deviceMatch(currentDevice, deviceFilter);
+
+    //         if(match) {
+    //             extend(options, ruleOptions);
+    //         }
+    //         return options;
+    //     }, {});
+    // }
```
MaximKudriavtsev: Should we remove this comment?

js/core/component.js
```diff
@@ -160,7 +122,7 @@ var Component = Class.inherit({
             this._setOptionsByReference();
             this._setDeprecatedOptions();
             this._options = this._getDefaultOptions();
```
MaximKudriavtsev, line 124: Why we keep `this._options` property here?

js/core/component.js
```diff
@@ -160,7 +122,7 @@ var Component = Class.inherit({
...
-            this._optionManager = new OptionManager(
+            this._optionManager = new Options(
```
MaximKudriavtsev, line 125: Why we use `this._optionManager` here for `Options` and the same name `this._optionManager` for `OptionManager` instance inside the Options class?

js/core/options/index.js
```diff
@@ -0,0 +1,223 @@
...
+        );
+
+        return isObject(value) ? clone(value) : value;
+    }
+
+    option(options, value) {
+        const isGetter = arguments.length < 2 && type(options) !== 'object';
+
+        if(isGetter) {
+            return this._optionManager.get(this._options, this._normalizeName(options));
```
MaximKudriavtsev: `this._options` is not defined in this context

js/core/options/index.js
```diff
@@ -0,0 +1,223 @@
...
+        } else {
+            this._optionManager.set(options, value);
+        }
+    }
+
+    silent(options, value) {
+        const isGetter = arguments.length < 2 && type(options) !== 'object';
+
+        if(isGetter) {
+            return this._optionManager.get(this._options, options, undefined, true);
```
MaximKudriavtsev: Same with `this._options`

js/core/options/index.js
```diff
@@ -0,0 +1,223 @@
...
+            options,
+            optionsByReference
+        );
+        this._optionManager.onRelevantNamesPrepared((options, name, value) => this._setRelevantNames(options, name, value));
+
+        this._rules = [];
+    }
+
+    get _options() {
+        return this._optionManager._options;
```
MaximKudriavtsev: Should we use an API to get it? I think yes because `_options` is a private property of the `this._optionManager` and we should use only public API to operate.

js/core/options/index.js
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    get _deprecateNames() {
+        if(!this._cachedDeprecateNames.length) {
+            for(const optionName in this._deprecated) {
+                this._cachedDeprecateNames.push(optionName);
+            }
+        }
+
+        return this._cachedDeprecateNames;
```
MaximKudriavtsev: Seems like `this._cachedDeprecateNames.push(optionName);` we should do in the constructor

js/core/options/utils.js
```diff
@@ -0,0 +1,29 @@
+import devices from '../devices';
+import { isEmptyObject, isFunction } from "../utils/type";
+import { findBestMatches } from '../utils/common';
+import { extend } from "../utils/extend";
+
+export const convertRulesToOptions = (rules) => {
```
MaximKudriavtsev, line 6: @churkin should we create utils functions or static methods would be preferred?

js/core/component.js
LazyLahtak: We cannot just get rid of this field. I'll do it in next PR

js/core/component.js
LazyLahtak: The same (see above):
In next PRs we remove all optionManager from components

js/core/options/index.js
```diff
@@ -0,0 +1,201 @@
...
+        let fieldObject = null;
+
+        do {
+            fieldName = fieldName ? `.${fieldName}` : '';
+            fieldName = getFieldName(fullName) + fieldName;
+            fullName = getParentName(fullName);
+            fieldObject = fullName ? this._optionManager.get(options, fullName, false) : options;
+        } while(!fieldObject);
+
+        fieldObject[fieldName] = value;
```
MaximKudriavtsev, line 84: Should we use `silent` method for these operations?

js/core/options/option_manager.js
```diff
@@ -0,0 +1,100 @@
...
+        this.cachedGetters = {};
+        this.cachedSetters = {};
+    }
+
+    _setByReference(options, rulesOptions) {
+        extend(true, options, rulesOptions);
+
+        for(const fieldName in this._optionsByReference) {
+            if(Object.prototype.hasOwnProperty.call(rulesOptions, fieldName)) {
+                options[fieldName] = rulesOptions[fieldName];
```
MaximKudriavtsev, line 26: Should we use `silent` method for these operations?

LazyLahtak: `_setByReference` is called by `silent`. It will be infinite loop.

nightskylark (approved):

js/core/options/index.js
LazyLahtak: We can't.
`silent` doesn't trigger 'onOptionChanged' in component (but assigning do that). We caught incorrect behaviour.

churkin (approved):

MaximKudriavtsev (approved):

PR #10658, opened 2019-11-18
DataGrid - Updating the element focus on refresh should be correct for focused row if cell or batch editing mode (T830334)

AntonSermyazhko:


js/ui/grid_core/ui.grid_core.focus.js
```diff
@@ -774,13 +772,21 @@ module.exports = {
...
                             if(operationTypes && !operationTypes.paging) {
                                 that.resizeCompleted.remove(that._scrollToFocusOnResize);
                                 that.resizeCompleted.add(that._scrollToFocusOnResize);
                             }
                         }
                     }
                 },
+                _getFocusedRowElement: function() {
+                    const focusedRowKey = this.option("focusedRowKey");
+                    const rowIndex = this._dataController.getRowIndexByKey(focusedRowKey);
```
vconst, line 789: This is wrong. The getRowIndexByKey return visible row index. But focusedRowKey for virtual scrolling return global row index

vconst (changes requested):

js/ui/grid_core/ui.grid_core.focus.js
AntonSermyazhko: Here I need visible index to find focusedRow element in the visible part

vconst (approved):

PR #10687, opened 2019-11-19
Fix closing of DateBox when pressing "esc" key while editing time (T831957)

DokaRus:


js/ui/date_box/ui.time_view.js
```diff
@@ -170,130 +176,149 @@ var TimeView = Editor.inherit({
...
+            editor.option("value", value);
+            editor._resumeValueChangeAction();
+        }
     },
 
     _getNumberBoxConfig: function() {
-        return extend({
+        return {
             showSpinButtons: true,
-            disabled: this.option("disabled"),
```
San4es, line 266: Why do we need this?

js/ui/date_box/ui.time_view.js
```diff
@@ -303,11 +328,6 @@ var TimeView = Editor.inherit({
         }
     },
 
-    _toggleDisabledState: function(value) {
```
San4es, line 306: Why do we need to remove this method?

San4es (dismissed):
LGTM

ksercs (dismissed):

js/ui/date_box/ui.time_view.js
DokaRus: When we create a nested widget via `createComponent` method, we say that root component should sync state of the `rtlEnabled` and `disabled` option with the nested one. So there is no need to set the initial value for this option

js/ui/date_box/ui.time_view.js
DokaRus: This method is not necessary. To synchronize the `disabled` option, it is sufficient to create a nested component using the `createComponent` method

js/ui/date_box/ui.time_view.js
San4es: Ok. Thank you!

js/ui/date_box/ui.time_view.js
San4es: Ok.

alexander-kotov-dx (dismissed):

San4es (approved):

alexander-kotov-dx (approved):

PR #10704, opened 2019-11-19
DataGrid: Fix placeholder was wider than grid if horizontal scroll exists (T830034)

jtoming830:


vconst (changes requested):
Fix the problem in Sortable widget because user can drag custom items in scrollable with horizontal scroll.
Also check dragging when horizontal scroll position is not zero.

js/ui/draggable.js
```diff
@@ -791,29 +797,31 @@ var Draggable = DOMComponentWithTemplate.inherit({
...
-            ownerDocument = $dragElement.get(0).ownerDocument,
             $window = $(window),
             mousePosition = {
                 x: e.pageX - $window.scrollLeft(),
                 y: e.pageY - $window.scrollTop()
             },
-            allObjects;
+            allObjects = that.getElementsFromPoint(mousePosition);
+
+        that.verticalScrollHelper && that.verticalScrollHelper.updateScrollable(allObjects, mousePosition);
```
vconst: Add "_" prefix to verticalScrollHelper  and horizontalScrollHelper because they are not public

vconst (changes requested):

vconst (approved):

Alyar666 (approved):

PR #10707, opened 2019-11-19
Support change of fullScreen popup option while the popup is showing (T825004)

alexander-kotov-dx:

### DIFF
--- a/js/ui/popup.js
+++ b/js/ui/popup.js
@@ -754,6 +754,15 @@ var Popup = Overlay.inherit({
         return this.callBase() || this.option("fullScreen");
     },
 
+    _toggleSafariFullScreen: function(value) {
+        var toggleFullScreenBeforeShown = this._useFixedPosition() && value && !this.isReady();
+        if(toggleFullScreenBeforeShown) {
+            this._bodyScrollTop = value ? window.pageYOffset : undefined;
+        } else {
+            this._toggleSafariScrolling(!value);
+        }
+    },
+
     _renderDimensions: function() {
         if(this.option("fullScreen")) {
             this._$content.css({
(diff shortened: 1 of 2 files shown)


js/ui/popup.js
```diff
@@ -827,8 +827,16 @@ var Popup = Overlay.inherit({
                 break;
             case "fullScreen":
                 this._toggleFullScreenClass(args.value);
-                this._toggleSafariScrolling(!args.value);
+
+                var toggleFullScreenBeforeShown = this._useFixedPosition() && args.value && !this.isReady();
```
DokaRus: Perhaps it would be better to extract this to the separate method

js/ui/popup.js
```diff
@@ -827,8 +827,16 @@ var Popup = Overlay.inherit({
...
+                if(toggleFullScreenBeforeShown) {
+                    this._bodyScrollTop = args.value ? window.pageYOffset : undefined;
+                } else {
+                    this._toggleSafariScrolling(!args.value);
+                }
```
DokaRus: Perhaps it would be better to extract this to the separate method

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -838,6 +838,43 @@ QUnit.test("start scroll position is saved after full screen popup hiding", func
     }
 });
 
+QUnit.test("works correctly with PREVENT_SAFARI_SCROLLING_CLASS class if fullScreen option is changed on showing event in safari (T825004)", function(assert) {
+    if(!IS_SAFARI) {
+        assert.expect(0);
+        return;
+    }
+    let $additionalElement;
```
San4es, line 846: Why do we need this element?

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -838,6 +838,43 @@ QUnit.test("start scroll position is saved after full screen popup hiding", func
...
+
+    try {
+        $additionalElement = $("<div>").height(2000).appendTo("body");
```
San4es: Use `$body` here?

testing/tests/DevExpress.ui.widgets/popup.tests.js
alexander-kotov-dx: to use the body element scrolling

DokaRus (approved):

San4es (approved):

PR #10732, opened 2019-11-20
Renovation: Button - move click subscription to the EventEngine

churkin:


js/ui/button.js
```diff
@@ -208,7 +206,7 @@ class Button extends Widget {
         let needValidate = true;
         let validationStatus = 'valid';
 
-        return this._createAction(({ event: e }) => {
+        return this._createAction(({ event }) => {
             if(needValidate) {
                 const validationGroup = this._validationGroupConfig;
```
MaximKudriavtsev, line 211: Can we don't create this variable, it can reduce the `if` count

MaximKudriavtsev (approved):

LazyLahtak (approved):

PR #10747, opened 2019-11-21
Renovation: move dxClick, resize, visibility event subscriptions to the eventEngine

churkin:


js/events/core/events_engine.js
```diff
@@ -614,6 +614,18 @@ eventsEngine.hover = {
     }
 };
 
+eventsEngine.visibility = {
+    on: ($el, shown, hiding, { namespace }) => {
```
DokaRus, line 618: Doesn't that seem a little confusing to you?

We have to remember the order of the event handlers. Also, don't forget that some parts of EventsEngine have only one handler, while others have several handlers.

Perhaps we can unify the structure of the parameters, for example: `.on(#target, handler, options)` where `handler: function|Object` 
So we can group handlers this way:
```suggestion
    on: ($el, { shown, hiding }, { namespace }) => {
```
And use it like follows:
```js
visibilityEvents.on($el,
  {
    shown: shownHandler,
    hiding: hidingHandler
  },
  { namespace }
);
```

js/ui/widget/ui.widget.js
```diff
@@ -491,7 +489,7 @@ var Widget = DOMComponentWithTemplate.inherit({
     _attachHoverEvents() {
         const { hoverStateEnabled } = this.option();
         const selector = this._activeStateUnit;
-        const namespace = UI_FEEDBACK;
+        const namespace = 'UIFeedback';
```
DokaRus, line 492:
```suggestion
```

I think it's better to use one constant
```js
hover.off($el, { selector, namespace: UI_FEEDBACK });
<...>
active.off($el, { selector, namespace: UI_FEEDBACK });
```

churkin: Why? :) Can I use different namespaces for the different events?

DokaRus: You use the same namespace twice in this file. Using a constant prevents the possibility of being sealed when using this namespace again

churkin: But it is not a constant :) Why these different namespaces should have general value?

js/events/core/events_engine.js
```diff
@@ -614,6 +614,18 @@ eventsEngine.hover = {
...
+        eventsEngine.on($el, addNamespace('dxhiding', namespace), hiding);
```
DokaRus, line 619: Most of the quotes in this file are double. For consistency, it is better to use either double or single quotes in all files.

js/ui/widget/ui.widget.js
DokaRus: > Why these different namespaces should have general value

Why are they different? They have the same value
[L492](https://github.com/DevExpress/DevExtreme/pull/10747/files#diff-de09fda2ed9e5fc71c891890679f5cc1R492): `const namespace = 'UIFeedback';`
[L513](https://github.com/DevExpress/DevExtreme/pull/10747/files#diff-de09fda2ed9e5fc71c891890679f5cc1R513): `const namespace = 'UIFeedback';`

DokaRus (approved):

LazyLahtak (approved):

js/core/dom_component.js
```diff
@@ -162,179 +160,185 @@ var DOMComponent = Component.inherit({
...
-    _createComponent: function(element, component, config) {
-        var that = this;
+    _createComponent(element, component, config = {}) {
+        const synchronizableOptions = grep(
+            this._getSynchronizableOptionsForCreateComponent(),
+            value => !(value in config)
+        );
 
-        config = config || {};
+        let { nestedComponentOptions, integrationOptions } = this.option();
```
MaximKudriavtsev, line 286: Should we use an option API to modify options?

MaximKudriavtsev (approved):

PR #10749, opened 2019-11-21
Use right isShown condition to detect popup showing stage

alexander-kotov-dx:

### DIFF
--- a/js/ui/popup.js
+++ b/js/ui/popup.js
@@ -755,7 +755,7 @@ var Popup = Overlay.inherit({
     },
 
     _toggleSafariFullScreen: function(value) {
-        var toggleFullScreenBeforeShown = this._useFixedPosition() && value && !this.isReady();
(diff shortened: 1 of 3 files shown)


js/ui/popup.js
```diff
@@ -755,7 +755,7 @@ var Popup = Overlay.inherit({
     },
 
     _toggleSafariFullScreen: function(value) {
-        var toggleFullScreenBeforeShown = this._useFixedPosition() && value && !this.isReady();
+        var toggleFullScreenBeforeShown = this._useFixedPosition() && value && !this._isShown;
```
San4es, line 758: this._currentVisible?

DokaRus: The value of this variable is set to true before the animation starts. In this case, it does not seem to be suitable

DokaRus (approved):

San4es (approved):

PR #10770, opened 2019-11-22
TreeView and TreeList - fix for rtl mode with bootstrap css (T822293)

novsstation:
T822293
Bootstrap breaks rtl mode. 
1) Fix for treeView and TreeList
2) Tests


DokaRus (changes requested):
I don't think that's a good idea. Bootstrap does not support [RTL](https://github.com/twbs/bootstrap/issues/28797)

PR #10828, opened 2019-11-25
Renovation: avoid directly using _options field

LazyLahtak:
Using the `_options` field directly is deprecated.
Instead, we have methods:
* `option(option, value)` - Using to set/get option
* `_setOptionSilent(option, value)` - Using to avoid call `_optionChanged` method
* `_getOptionByStealth(name)` - Using to get option directly, without check if the option is deprecated. Also, that should be used if you redefined the `option` method (to avoid wrong result or an infinite loop).
* `_setOptionByStealth(options, value)` - Using to set option directly, without check if the option is deprecated and call any events or methods.
* `_getOptionValue(name, context)` - Using to get the option and bind to the context.

Continuation of #10654


js/ui/form/ui.form.js
```diff
@@ -1181,7 +1181,7 @@ const Form = Widget.inherit({
 
             if(optionFullName === "formData") {
                 if(!isDefined(args.value)) {
-                    that._options.formData = args.value = {};
+                    that.option("formData", args.value = {});
```
DokaRus: Should it be `that._setOptionByStealth`?
@jsdmitry  @dxvladislavvolkov 

@LazyLahtak Re-check similar code fragments, please

testing/tests/DevExpress.ui.widgets.form/form.API.option_formData.tests.js
```diff
@@ -59,8 +59,9 @@ QUnit.test("Set { formData: {}, items: [dataField1] }, call option(formData, {})
 
 QUnit.test("Set { formData: {dataField1: a}, items: [dataField1] }, call option(formData, null)", function(assert) {
     var form = $("#form").dxForm({ formData: { dataField1: "a" }, items: ["dataField1"] }).dxForm("instance");
+    var formData = {};
     form.option("formData", null);
-    assert.propEqual(form.option("formData"), { dataField1: "" });
+    assert.propEqual(form.option("formData"), formData);
```
DokaRus: @jsdmitry  Is it expected?

DokaRus (changes requested):

js/ui/form/ui.form.js
LazyLahtak: You're right, I'll correct it.

testing/tests/DevExpress.ui.widgets.form/form.API.option_formData.tests.js
LazyLahtak: This changes caused by `that.option` instead `that._setOptionByStealth`
I'll revert this.

js/ui/form/ui.form.js
jsdmitry: This code doesn't should lead to the optionChanged event

DokaRus: > I'll correct it.

Do not forget to check the other code fragments (Switch, asyncTemplateMixin, SpeedDialButton etc)

dxvladislavvolkov: And we should use _getOptionByStealth instead of _option("optionName")

LazyLahtak: We also have `_setOptionsSilent` (it works this). When we should use `_setOptionsSilent` and when `_setOptionByStealth`?

DokaRus: I think we should replace `_setOptionsSilent` with `_setOptionByStealth`. It looks like API duplication.

dxvladislavvolkov: _setOptionsSilent doesn't call only  _optionChanged method https://github.com/DevExpress/DevExtreme/blob/20_1/js/core/component.js#L255

LazyLahtak: Do we need `_setOptionsSilent` and `_setOptionByStealth` both? Can we keep one of scenario and get rid of other?

LazyLahtak: @dxvladislavvolkov 
>And we should use _getOptionByStealth instead of _option("optionName")

I tried to use it (for example in `js/ui/speed_dial_action/*`) and it doesn't work (tests crashed)

jsdmitry (dismissed):

DokaRus (dismissed):

dxvladislavvolkov (dismissed):

groshenkovamarina (approved):

PR #10833, opened 2019-11-26
Renovation: keyboardProcessor refactoring

churkin:
- Get rid of keyboard processors transfering from widget to widget
- Use simple api for keyboard events: 
```javascript
const id = keyboard.on(element, () => ...);
...
keyboard.off(id);
```
- Use keyboard processor as an internal tool
- Add private `onKeyboardHandled` event to catch child widget keyboard events

### DIFF
--- a/js/core/dom_component.js
+++ b/js/core/dom_component.js
@@ -12,7 +12,7 @@ import { grep, noop } from './utils/common';
 import { inArray } from './utils/array';
 import { isString, isDefined } from './utils/type';
 import { hasWindow } from '../core/utils/window';
-import { resize as resizeEvent, visibility as visibilityEvents } from '../events/core/events_engine';
+import { resize as resizeEvent, visibility as visibilityEvents } from '../events/index';
 
 const { abstract } = Component;
 
--- a/js/core/dom_component.js
+++ b/js/core/dom_component.js
@@ -12,7 +12,7 @@ import { grep, noop } from './utils/common';
 import { inArray } from './utils/array';
 import { isString, isDefined } from './utils/type';
 import { hasWindow } from '../core/utils/window';
-import { resize as resizeEvent, visibility as visibilityEvents } from '../events/index';
+import { resize as resizeEvent, visibility as visibilityEvents } from '../events/';
 
 const { abstract } = Component;
 
--- a/js/events/index.js
+++ b/js/events/index.js
@@ -33,7 +33,6 @@ export const resize = {
     }
 };
 
-
 export const hover = {
     on: ($el, start, end, { selector, namespace }) => {
         eventsEngine.on($el, addNamespace('dxhoverend', namespace), selector, event => end(event));
@@ -100,18 +99,20 @@ export const click = {
     }
 };
 
+let index = 0;
 const keyboardProcessors = {};
+const getKeyboardProcessorId = () => `keyboardProcessorId${index++}`;
 
 export const keyboard = {
     on: (element, focusTarget, handler) => {
-        const listenerId = Symbol('listenerId');
+        const listenerId = getKeyboardProcessorId();
 
         keyboardProcessors[listenerId] = new KeyboardProcessor({ element, focusTarget, handler });
 
         return listenerId;
     },
 
-    off: (listenerId) => {
+    off: listenerId => {
         if(listenerId && keyboardProcessors[listenerId]) {
             keyboardProcessors[listenerId].dispose();
             delete keyboardProcessors[listenerId];
@@ -119,5 +120,5 @@ export const keyboard = {
     },
 
     // NOTE: For tests
-    _getProcessor: (listenerId) => keyboardProcessors[listenerId]
+    _getProcessor: listenerId => keyboardProcessors[listenerId]
 };
--- a/js/events/index.js
+++ b/js/events/index.js
@@ -1,0 +1,123 @@
+import domAdapter from '../core/dom_adapter';
+import eventsEngine from './core/events_engine';
+import KeyboardProcessor from './core/keyboard_processor';
+
+function addNamespace(event, namespace) {
+    return namespace ? eventsEngine.addNamespace(event, namespace) : event;
+}
+
+export const active = {
+    on: ($el, active, inactive, opts) => {
+        const { selector, showTimeout, hideTimeout, namespace } = opts;
+
+        eventsEngine.on($el, addNamespace('dxactive', namespace), selector, { timeout: showTimeout },
+            event => active.execute({ event, element: event.currentTarget })
+        );
+        eventsEngine.on($el, addNamespace('dxinactive', namespace), selector, { timeout: hideTimeout },
+            event => inactive.execute({ event, element: event.currentTarget })
+        );
+    },
+
+    off: ($el, { namespace, selector }) => {
+        eventsEngine.off($el, addNamespace('dxactive', namespace), selector);
+        eventsEngine.off($el, addNamespace('dxinactive', namespace), selector);
+    }
+};
+
+export const resize = {
+    on: ($el, resize, { namespace } = {}) => {
+        eventsEngine.on($el, addNamespace('dxresize', namespace), resize);
+    },
+    off: ($el, { namespace } = {}) => {
+        eventsEngine.off($el, addNamespace('dxresize', namespace));
+    }
+};
+
+
+export const hover = {
+    on: ($el, start, end, { selector, namespace }) => {
+        eventsEngine.on($el, addNamespace('dxhoverend', namespace), selector, event => end(event));
+        eventsEngine.on($el, addNamespace('dxhoverstart', namespace), selector, event => {
+            start.execute({ element: event.target, event });
+        });
+    },
+
+    off: ($el, { selector, namespace }) => {
+        eventsEngine.off($el, addNamespace('dxhoverstart', namespace), selector);
+        eventsEngine.off($el, addNamespace('dxhoverend', namespace), selector);
+    }
+};
+
+export const visibility = {
+    on: ($el, shown, hiding, { namespace }) => {
+        eventsEngine.on($el, addNamespace('dxhiding', namespace), hiding);
+        eventsEngine.on($el, addNamespace('dxshown', namespace), shown);
+    },
+
+    off: ($el, { namespace }) => {
+        eventsEngine.off($el, addNamespace('dxhiding', namespace));
+        eventsEngine.off($el, addNamespace('dxshown', namespace));
+    }
+};
+
+export const focus = {
+    on: ($el, focusIn, focusOut, { namespace, isFocusable }) => {
+        eventsEngine.on($el, addNamespace('focusin', namespace), focusIn);
+        eventsEngine.on($el, addNamespace('focusout', namespace), focusOut);
+
+        if(domAdapter.hasDocumentProperty('onbeforeactivate')) {
+            eventsEngine.on($el, addNamespace('beforeactivate', namespace),
+                e => isFocusable(e.target) || e.preventDefault()
+            );
+        }
+    },
+
+    off: ($el, { namespace }) => {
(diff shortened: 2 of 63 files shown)


churkin:
Tests are ok, it is problem in the schedular, pls review.

js/events/index.js
```diff
@@ -0,0 +1,123 @@
+import domAdapter from '../core/dom_adapter';
+import eventsEngine from './core/events_engine';
+import KeyboardProcessor from './core/keyboard_processor';
+
+function addNamespace(event, namespace) {
+    return namespace ? eventsEngine.addNamespace(event, namespace) : event;
```
nightskylark: Why `addNamespace` method has been left in the `eventsEngine` module? It looks like the namespace adding is the main feature of the current one.

js/core/dom_component.js
```diff
@@ -12,7 +12,7 @@ import { grep, noop } from './utils/common';
 import { inArray } from './utils/array';
 import { isString, isDefined } from './utils/type';
 import { hasWindow } from '../core/utils/window';
-import { resize as resizeEvent, visibility as visibilityEvents } from '../events/core/events_engine';
+import { resize as resizeEvent, visibility as visibilityEvents } from '../events/index';
```
nightskylark: I did not get why this new module named as `index`. We have a lot of modules which are semanticly connected with the `event` word so it becomes a mess. I'd like to know more about this module is needed for by looking at it's name.

MaximKudriavtsev: Should we remove `/index` from this line because of an `index` file will be exported by default from folder?

testing/functional/tests/editors/list.ts
```diff
@@ -5,7 +5,7 @@ import List from '../../model/list';
 fixture `List`
     .page(url(__dirname, './pages/t718398.html'));
 
-test("Should focus first item after changing selection mode (T811770)", async t => {
+test('Should focus first item after changing selection mode (T811770)', async t => {
```
nightskylark, line 8: I like single quotes too, but I think it's not a good idea to change them in separate files. It should be a common decision with full replacing and lint rules updating. Different quotes in different files causes more problems than profits.

js/core/dom_component.js
churkin: We already have `events.js` file

js/ui/drop_down_box.js
```diff
@@ -249,11 +249,7 @@ var DropDownBox = DropDownEditor.inherit({
         this.callBase();
 
         if(this.option("focusStateEnabled")) {
-            this._popup._keyboardProcessor.push(new KeyboardProcessor({
-                element: this.content(),
-                handler: this._popupElementTabHandler,
-                context: this
-            }));
+            keyboard.on(this.content(), null, e => this._popupElementTabHandler(e));
```
nightskylark, line 252: I found that any other keyboard subscriptions stores an IDs to detach them later. Why it does not happened here? Maybe it discovers a leak? If so, maybe it's possible to store these IDs on each subscription automatically? It could allow to detach them all on disposing.

js/events/index.js
```diff
@@ -0,0 +1,123 @@
...
+    off: ($el, { namespace } = {}) => {
+        eventsEngine.off($el, addNamespace('click', namespace));
+    }
+};
+
+const keyboardProcessors = {};
+
+export const keyboard = {
+    on: (element, focusTarget, handler) => {
+        const listenerId = Symbol('listenerId');
```
MaximKudriavtsev: Is it a problem that `Symbol` doesn't work in IE?

js/ui/lookup.js
```diff
@@ -977,12 +968,13 @@ var Lookup = DropDownList.inherit({
...
+
             this._searchBox = this._createComponent($searchBox, TextBox, {
-                onDisposing: function() {
-                    this._detachSearchChildProcessor();
-                }.bind(this),
-                onFocusIn: this._searchFocusHandler.bind(this),
-                onFocusOut: this._searchBlurHandler.bind(this),
+                onDisposing: () => isKeyboardListeningEnabled = false,
+                onFocusIn: () => isKeyboardListeningEnabled = true,
+                onFocusOut: () => isKeyboardListeningEnabled = false,
```
MaximKudriavtsev, line 976: Should we create a class property and change it by methods instead of variables?

churkin: Why?

js/events/index.js
```diff
@@ -0,0 +1,125 @@
...
+        eventsEngine.on($el, addNamespace('click', namespace), click);
+    },
+    off: ($el, { namespace } = {}) => {
+        eventsEngine.off($el, addNamespace('click', namespace));
+    }
+};
+
+let index = 0;
+const keyboardProcessors = {};
+const getKeyboardProcessorId = () => `keyboardProcessorId${index++}`;
```
IlyaKhD:
```suggestion
function getNextProcessorId() { return index++; }
```

js/events/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+export const keyboard = {
+    on: (element, focusTarget, handler) => {
+        const listenerId = getKeyboardProcessorId();
```
IlyaKhD:
```suggestion
        const listenerId = `keyboardProcessorId${getNextProcessorId()}`;
```

js/events/utils/add_namespace.js
```diff
@@ -0,0 +1,19 @@
+import errors from '../../core/errors';
+
+const addNamespace = (eventNames, namespace) => {
+    if(!namespace) {
+        throw errors.Error('E0017');
+    }
+
+    if(typeof eventNames === 'string') {
```
IlyaKhD: What about rearranging `if`'s?

```
  if(typeof eventNames !== 'string') { // or better use Array.isArray
    return eventNames
        .map(eventName => `${eventName}.${namespace}`)
        .join(' ');
  }

  if(eventNames.indexOf(' ') > -1) {
    return addNamespace(eventNames.split(/\s+/g), namespace);
  }

  return  `${eventNames}.${namespace}`;        
```

js/events/utils/add_namespace.js
```diff
@@ -0,0 +1,19 @@
+import errors from '../../core/errors';
+
+const addNamespace = (eventNames, namespace) => {
```
IlyaKhD, line 3:
```suggestion
function addNamespace(eventNames, namespace) {
```

js/events/utils/add_namespace.js
IlyaKhD: @nightskylark is that case possible: `eventNames` is an array of strings and contains items with spaces (that should be split)?

IlyaKhD (dismissed):

js/events/utils/add_namespace.js
nightskylark: I don't think so.

js/events/index.js
```diff
@@ -0,0 +1,125 @@
+import domAdapter from '../core/dom_adapter';
+import eventsEngine from './core/events_engine';
+import KeyboardProcessor from './core/keyboard_processor';
+import { addNamespace as pureAddNamespace } from './utils';
+
+function addNamespace(event, namespace) {
+    return namespace ? pureAddNamespace(event, namespace) : event;
+}
```
MaximKudriavtsev, line 8: Why do we use `function` here instead of `() => {  }`

js/events/utils/add_namespace.js
```diff
@@ -0,0 +1,19 @@
...
+            `${eventNames}.${namespace}` :
+            addNamespace(eventNames.split(/\s+/g), namespace);
+    }
+
+    return eventNames
+        .map(eventName => `${eventName}.${namespace}`)
+        .join(' ');
+};
+
+export default addNamespace;
```
MaximKudriavtsev, line 21: We can write the next code
```
export default (eventNames, namespace) => { ... }
```

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -140,7 +140,7 @@ var KeyboardNavigationController = core.ViewController.inherit({
             eventsEngine.off($element, eventUtils.addNamespace(pointerEvents.up, "dxDataGridKeyboardNavigation"), clickAction);
             eventsEngine.on($element, eventUtils.addNamespace(pointerEvents.up, "dxDataGridKeyboardNavigation"), clickSelector, clickAction);
 
-            that._initKeyDownProcessor(that, $element, that._keyDownHandler);
+            that._initKeyDownHandler($element, e => that._keyDownHandler(e));
```
MaximKudriavtsev, line 143: We can not create a new function.

`that._initKeyDownHandler($element, e => that._keyDownHandler(e));`
->
`that._initKeyDownHandler($element, that._keyDownHandler);`

js/ui/overlay/ui.overlay.js
```diff
@@ -1406,11 +1407,11 @@ var Overlay = Widget.inherit({
...
     _attachKeyboardEvents: function() {
-        this._keyboardProcessor = new KeyboardProcessor({
-            element: this._$content,
-            handler: this._keyboardHandler,
-            context: this
-        });
+        this._keyboardListenerId = keyboard.on(
+            this._$content,
+            null,
+            opts => this._keyboardHandler(opts)
```
MaximKudriavtsev, line 1413: Also with creating a new function

js/events/utils/add_namespace.js
churkin: No

MaximKudriavtsev (dismissed):

js/ui/widget/ui.widget.js
```diff
@@ -183,6 +182,11 @@ var Widget = DOMComponentWithTemplate.inherit({
     _init: function() {
         this.callBase();
         this._initContentReadyAction();
+        this._initKeyboardHandledAction();
+    },
+
+    _initKeyboardHandledAction() {
+        this._keyboardHandledAction = this._createActionByOption('onKeyboardHandled');
```
vconst: Is it necessary to create action here? It is necessary for public event, but it is private.
Action is not light thing. Every new action in base widget decrease performance.
Also by default action has "disabled" and "readOnly" validators. So in readOnly widget callback onKeyboardHandled will not be executed. Is it expected behavior?

Maybe it will be better avoid creating action here. And instead of 
```
this._keyboardHandledAction(options);
```
call callback directly
```
const onKeyboardHandled = this.option("onKeyboardHandled");
onKeyboardHandled && onKeyboardHandled(options);
```

vconst (dismissed):

MaximKudriavtsev (dismissed):

js/events/utils/add_namespace.js
```diff
@@ -0,0 +1,19 @@
...
+const addNamespace = (eventNames, namespace) => {
+    if(!namespace) {
+        throw errors.Error('E0017');
+    }
+
+    if(Array.isArray(eventNames)) {
+        return eventNames
+            .map(eventName => `${eventName}.${namespace}`)
+            .join(' ');
+    } else if(eventNames.indexOf(' ') !== -1) {
```
IlyaKhD:
```suggestion
    }
    
    if(eventNames.indexOf(' ') !== -1) {
```
No need for `else` since previois `if` block contains `return` statement

MaximKudriavtsev (approved):

IlyaKhD (approved):

js/ui/menu/ui.menu.js
```diff
@@ -541,14 +541,16 @@ class Menu extends MenuBase {
         return submenu;
     }
 
+    _getKeyboardListeners() {
```
IgnatovDan, line 544: FYI
This change makes this regression, it worked correctly before:
[T920355 - Menu - Item in onItemClick handler is incorrect when using keyboard navigation to "click" items](https://supportcenter.devexpress.com/internal/ticket/details/T920355)

PR #10875, opened 2019-11-28
Renovation: EventsStrategy refactor

LazyLahtak:
Rename `setEventsStrategy` method.

Note: duplication of doc comments will be fixed in other PR

Refactor of #10524

### DIFF
--- a/js/bundles/modules/core.js
+++ b/js/bundles/modules/core.js
@@ -27,8 +27,6 @@ DevExpress.Color = require("../../color");
 
 var animationFrame = require("../../animation/frame");
 
-DevExpress.EventsStrategy = require("../../core/events_strategy").EventsStrategy;
-
 /**
  * @name utils
  * @namespace DevExpress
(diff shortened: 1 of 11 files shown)


IlyaKhD (changes requested):
What is the aim of declaring inheritance?
Declarations (TS especially) should fit JS part, and there is no inheritance in the JS part, right?

I mean these:

```
@name DevicesObject 
@inherits EventsStrategy
```
```
@name Store
@inherits EventsStrategy
```
etc.

LazyLahtak:
> What is the aim of declaring inheritance?

It is a try to avoid multiple [doc comments](https://github.com/DevExpress/DevExtreme/pull/10875/files#diff-8ae7d11679885a2279672087e53958b3R38-R81). But if we declare it in the EventsStrategy, we should bound it with components. We decided to declare inheritance.

IlyaKhD:
LGTM

churkin (approved):

IlyaKhD (approved):

PR #10903, opened 2019-11-29
DataGrid -  Add confirmDelete option to allow a user to delete a row without displaying a confirmation message

igormalzew:


RomanTsukanov (approved):

igormalzew:
Decided to add only in 20.1

PR #10921, opened 2019-12-02
File Manager - Checkbox and isDirectory columns width adjusting

Staziz:


js/ui/file_manager/ui.file_manager.item_list.details.js
```diff
@@ -13,6 +13,7 @@ const FILE_MANAGER_DETAILS_ITEM_LIST_CLASS = "dx-filemanager-details";
 const FILE_MANAGER_DETAILS_ITEM_THUMBNAIL_CLASS = "dx-filemanager-details-item-thumbnail";
 const FILE_MANAGER_DETAILS_ITEM_NAME_CLASS = "dx-filemanager-details-item-name";
 const FILE_MANAGER_DETAILS_ITEM_NAME_WRAPPER_CLASS = "dx-filemanager-details-item-name-wrapper";
+const FILE_MANAGER_DETAILS_ITEM_IS_DIRECTORY_CLASS = "dx-filemanager-details-item_is_directory";
```
vladkovl: We should use "-" character instead of the "_". And "`dx-filemanager-details-item-type`" is better then the "`dx-filemanager-details-item-is-directory`" as for me.

maxerokh (approved):

PR #10943, opened 2019-12-03
PivotGrid: Fix error occurs when filtering values by using Field Chooser (T836053)

jtoming830:


testing/tests/DevExpress.ui.widgets.pivotGrid/store.remote.tests.js
```diff
@@ -1386,6 +1385,40 @@ QUnit.test("Filter by number with groupInterval", function(assert) {
     });
 });
 
+// T836053
+QUnit.test("Error should not be thrown during filtering if there are more filter values ​​than fields", function(assert) {
+    var store = new RemoteStore({
+            load: function(loadOptions) {
+                assert.deepEqual(loadOptions.filter,
```
vconst: It will be better assign loadOptions.filter to variable and check it after store.load call because it is no check that this callback is called

vconst (changes requested):

vconst (approved):

AntonSermyazhko (approved):

PR #10945, opened 2019-12-03
Clear last show handler timeout when a new show event is fired (T829575)

alexander-kotov-dx:

### DIFF
--- a/js/ui/popover.js
+++ b/js/ui/popover.js
@@ -595,13 +595,13 @@ var Popover = Popup.inherit({
         return side === "left" || side === "right";
     },
 
-    _clearEventTimeout: function(visibility) {
-        clearTimeout(this._timeouts[visibility ? "show" : "hide"]);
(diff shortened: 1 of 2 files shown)


js/ui/popover.js
```diff
@@ -598,6 +599,11 @@ var Popover = Popup.inherit({
         clearTimeout(this._timeouts[visibility ? "show" : "hide"]);
     },
 
+    _clearEventsTimeouts: function() {
+        this._clearEventTimeout(true);
```
DokaRus: Let's redefine this method to make it clear:

```
_clearEventTimeout: function(name){
<...>
}
```
So we can use it as follow:
`this._clearEventTimeout("hide");`
or
`this._clearEventTimeout("show");`
or
`this._clearEventTimeout(visibilityState ? "show" : "hide");`

DokaRus (approved):

San4es (approved):

PR #10925, opened 2019-12-02
Overlay wrapper should be repositioned regardless of shading (T835358)

ksercs:

PR #10955, opened 2019-12-03
Popover shading container should be closest parent element with defined position (T836650) + fix regression from #10925

ksercs:


testing/tests/DevExpress.ui.widgets/popover.tests.js
```diff
@@ -956,6 +956,33 @@ QUnit.test("popover rendering with position.of and target options", function(ass
     }
 });
 
+QUnit.test("popover shading should cover the parent element with absolute position, not target element", function(assert) {
+    fixtures.simple.create();
+
+    try {
+        var $popover = $("#what"),
```
DokaRus: Use `const` or `let`

DokaRus (approved):

San4es (approved):

PR #10961, opened 2019-12-04
DataGrid: Fix wrong cell height when showCheckBoxesMode is onClick and hoverStateEnabled is true (T837103)

jtoming830:


igormalzew (dismissed):

styles/widgets/generic/gridBase.generic.less
```diff
@@ -434,6 +434,10 @@
                     &.dx-@{widgetName}-group-space {
                         border-right-color: ~"@{@{widgetName}-hover-bg}";
                     }
+
+                    &:first-child {
```
vconst: Selection column can have non-first position. Also, the first cell is not always selection column. So it will be better to use .dx-command-select selector

vconst (dismissed):

Alyar666 (approved):

igormalzew (approved):

PR #10973, opened 2019-12-05
DataGrid / TreeList -  Add appropriate type to options with default value 'auto'

igormalzew:


IlyaKhD:
I've fixed the tool
@igormalzew, make sure you use `devextreme-internal-tools@1.2.16` locally and run `update-ts` again. This should fix the issue

vconst (approved):

IlyaKhD (approved):

PR #10984, opened 2019-12-05
Lookup should display the LoadPanel when the data is being loaded for a long time while searching (T837176)

DokaRus:


ksercs (dismissed):

alexander-kotov-dx (dismissed):

testing/tests/DevExpress.ui.widgets.editors/lookup.tests.js
```diff
@@ -2930,13 +2936,49 @@ QUnit.test("search should be execute after paste", function(assert) {
...
-    } finally {
-        clock.restore();
-        fx.off = originalFX;
-    }
+    [
+        { loadDelay: 200, indicateLoading: false },
+        { loadDelay: 1000, indicateLoading: true }
+    ].forEach(({ loadDelay, indicateLoading }) => {
+        QUnit.test(`search with loading delay = ${loadDelay} should ${indicateLoading ? "" : "not"} lead to the load panel being displayed`, function(assert) {
+            const clock = sinon.useFakeTimers();
```
San4es: Looks like we already do this here:
https://github.com/DevExpress/DevExtreme/pull/10984/files#diff-922b6fcabe4f5ae828645b18174c6277R2919

Can we use `this.clock` instead?

js/ui/autocomplete.js
```diff
@@ -161,8 +161,7 @@ var Autocomplete = DropDownList.inherit({
 
     _listConfig: function() {
         return extend(this.callBase(), {
-            pageLoadMode: "none",
-            indicateLoading: false
+            pageLoadMode: "none"
```
San4es, line 164: Why are we going to remove this?

San4es (changes requested):

testing/tests/DevExpress.ui.widgets.editors/lookup.tests.js
DokaRus: Thank you, I'll fix it

js/ui/autocomplete.js
DokaRus: The value of this option is already set to "false" (in the parent component)

San4es (approved):

alexander-kotov-dx (approved):

PR #10987, opened 2019-12-06
Renovation: Widget es6 refactoring

churkin:


js/ui/widget/ui.widget.js
```diff
@@ -1,30 +1,25 @@
...
-const { hover, focus, active, dxClick, keyboard } = events;
-
-var WIDGET_CLASS = "dx-widget",
-    ACTIVE_STATE_CLASS = "dx-state-active",
-    DISABLED_STATE_CLASS = "dx-state-disabled",
-    INVISIBLE_STATE_CLASS = "dx-state-invisible",
-    HOVER_STATE_CLASS = "dx-state-hover",
-    FOCUSED_STATE_CLASS = "dx-state-focused",
-    FEEDBACK_SHOW_TIMEOUT = 30,
-    FEEDBACK_HIDE_TIMEOUT = 400;
```
LazyLahtak, line 27: Should we save these types, but define them as constants? Maybe in separate file.

churkin: Most of these use once ) I think it is not necessary.

LazyLahtak: As I see, we have similar types in other components (for example, `ACTIVE_STATE_CLASS`, `DISABLED_STATE_CLASS`, `INVISIBLE_STATE_CLASS`).  Maybe we should extract them in `constants.js` or something like that.

LazyLahtak: In other PR of course.

churkin: I don't see the real benefits of this

js/ui/widget/ui.widget.js
```diff
@@ -1,30 +1,25 @@
...
+import { focusable as focusableSelector } from './selectors';
+import { inArray } from '../../core/utils/array';
+import { isFakeClickEvent } from '../../events/utils';
+import { isPlainObject, isDefined } from '../../core/utils/type';
+
+import '../../events/click';
+import '../../events/core/emitter.feedback';
+import '../../events/hover';
+
+function setAttribute(name, value, target) {
```
LazyLahtak, line 17: Should we use an arrow function?
`const setAttribute = (name, value, target) => {...}`

churkin:
@IlyaKhD ping

js/ui/widget/ui.widget.js
```diff
@@ -745,16 +687,25 @@ var Widget = DOMComponentWithTemplate.inherit({
...
     }
 });
 
+Widget.getOptionsFromContainer = ({ name, fullName, value }) => {
+    let options = {};
+
+    if(name === fullName) {
+        options = value;
+    } else {
+        const option = fullName.split('.').pop();
```
IlyaKhD, line 703:
```suggestion
        const option = fullName.substr(fullName.lastIndexOf('.') + 1);
```

IlyaKhD: may be extract a helper function for this

js/ui/widget/ui.widget.js
IlyaKhD: Nope 😉

js/ui/widget/ui.widget.js
```diff
@@ -262,64 +245,52 @@ var Widget = DOMComponentWithTemplate.inherit({
...
-            }
-            return this._renderContentImpl();
-        }).done(() => {
-            if(this._disposed) {
-                return;
-            }
-            this._fireContentReadyAction();
-        });
+    _renderContent() {
+        deferRender(() => !this._disposed ? this._renderContentImpl() : void 0)
```
IlyaKhD, line 255:
```suggestion
        deferRender(() => !this._disposed ? this._renderContentImpl() : undefined)
```
As I got it, this should preserve the global `undefined` variable from been overwritten in some ancient browser. If we think this can a problem we should better create a dedicated PR for updating all such places in devextreme.

Actually it seems to be non-relevant for our days)

js/ui/widget/ui.widget.js
```diff
@@ -555,188 +516,169 @@ var Widget = DOMComponentWithTemplate.inherit({
...
-            option.target.attr(attrName, attrValue);
-        };
-
-        if(!typeUtils.isPlainObject(arguments[0])) {
-            setAttribute({
-                name: arguments[0],
-                value: arguments[1],
-                target: arguments[2] || this._getAriaTarget()
-            });
+    setAria(...args) {
```
IlyaKhD, line 654:
```suggestion
    setAria(arg0, arg1, arg2) {
```
No need for spread

js/ui/widget/ui.widget.js
churkin: Is it more clear you think? It is very simple for the util function.

js/ui/widget/ui.widget.js
churkin: I like to use `void 0`. Not so important but looks better for me. I can rollback it if you want.

js/ui/widget/ui.widget.js
churkin: Why? It looks shortly and hints that the number of parameters is not constant

churkin:
Ready to review

js/ui/drop_down_editor/ui.drop_down_editor.js
```diff
@@ -785,7 +786,7 @@ var DropDownEditor = TextBox.inherit({
     _updatePopupWidth: commonUtils.noop,
 
     _popupOptionChanged: function(args) {
-        var options = this._getOptionsFromContainer(args);
+        var options = Widget.getOptionsFromContainer(args);
```
babich-a, line 790: Should we use const/let instead of var?

churkin: We should, but this file not refactored yet.

js/ui/widget/ui.widget.js
IlyaKhD: The idea is to make different codebase parts connected via such simple helpers. E.g. if I want to check all the places where name building logic is used I can find usages of such a helper.

But I'm not this is such a case, actually

js/ui/widget/ui.widget.js
IlyaKhD: As I get from the code the number of args is constant, as they are accessed only by index, and not processed as collection.

js/ui/widget/ui.widget.js
IlyaKhD: I think we could add this to the next guide-lines-meeting agenda

js/ui/widget/ui.widget.js
churkin: `setArea({}, target)` or `setArea(name, value, target)`

churkin: I'll create new PR for this dispute

LazyLahtak (dismissed):

MaximKudriavtsev (approved):

LazyLahtak (approved):

PR #11035, opened 2019-12-10
DataGrid - Horizontal scrolling should not scroll back if focusedRowEnabled and columnRenderingMode is 'virtual' (T834918)

AntonSermyazhko:


js/ui/grid_core/ui.grid_core.focus.js
```diff
@@ -790,9 +790,12 @@ module.exports = {
                     return $(this.getRowElement(rowIndex >= 0 ? rowIndex : 0));
                 },
 
-                _scrollToElement: function(element) {
+                scrollElementTop: function($element) {
```
vconst: Bad name. May be, scrollToRowElement?

vconst (changes requested):

jsdmitry (dismissed):

js/ui/grid_core/ui.grid_core.focus.js
AntonSermyazhko: Have agreed rename to scrollToElementVertically

vconst (approved):

jsdmitry (approved):

PR #11036, opened 2019-12-10
Renovation: refactor options cache in Widget

LazyLahtak:
Move logic of initializing, setting and getting options cache from `Widget` into `Options`. 
`Options` now have `cache(name, option)` method (works like methods `option(options, value)` and `silent(options, value)`)


LazyLahtak:
Ready to review

churkin (dismissed):

js/core/options/index.js
```diff
@@ -197,4 +199,14 @@ export class Options {
     isDeprecated(name) {
         return Object.prototype.hasOwnProperty.call(this._deprecated, name);
     }
+
+    cache(name, options) {
```
IlyaKhD, line 203: Why not split into two simple methods? E.g. `addToCache` \ `readFromCache` or simpler `cahe` \ `fromCache`

IlyaKhD: No need for this `if(isGetter) {` logic, simpler to read - good points only 😉

LazyLahtak: We discussed this with @churkin and desided to make this method like `option` and `silent` methods.
Should we change them too?

IlyaKhD: No you shouldn't
I don't see how these methods are related to each other actually. `cache` is very simple and should not be over-weighted. `addToCache` and `getFromCache` is very clear, no one is going to be mixed up, I think.

LazyLahtak: Methods `option`, `silent` and `cache` work like getter and setter at same time. They also look very similar. I think we should leave them as they are now or refactor them all.

js/ui/widget/ui.widget.js
```diff
@@ -179,30 +179,23 @@ const Widget = DOMComponentWithTemplate.inherit({
         this._initContentReadyAction();
     },
 
-    _clearInnerOptionCache(optionContainer) {
-        this[`${optionContainer}Cache`] = {};
-    },
+    _getOptionsFromContainer: function({ name, fullName, value }) {
+        var options = {};
```
DokaRus: Use `const` or `let` instead of `var`

LazyLahtak: Done in #10987

DokaRus (approved):

churkin (approved):

js/core/options/index.js
churkin: You should not merge PR if it has not resolved comments.

LazyLahtak: Decided to leave them as they are now

PR #11037, opened 2019-12-10
dxList: nextButton should be removed after search when result items count is smaller then pageSize and repaintChangesOnly is true(T838645)

ksercs:


testing/tests/DevExpress.ui.widgets/listParts/commonTests.js
```diff
@@ -794,6 +795,22 @@ QUnit.module("next button", moduleSetup, () => {
         list.repaint();
         assert.ok($(".dx-list-next-button", this.element).text());
     });
+
+    QUnit.test("nextButton should be removed after search if result items count is smaller than page size, repaintChangesOnly=true (T838645)", assert => {
```
San4es:
```suggestion
    QUnit.test("nextButton should be removed after search if result items count is smaller than page size, repaintChangesOnly=true (T838645)", function(assert) {
```

San4es (approved):

PR #11057, opened 2019-12-11
DataGrid - The widget incorrectly renders borders of the last column's header cell if native scrolling is enabled (T838716)

igormalzew:


js/ui/grid_core/ui.grid_core.grid_view.js
```diff
@@ -568,6 +568,7 @@ var ResizingController = modules.ViewController.inherit({
 
         commonUtils.deferRender(function() {
             that._columnHeadersView && that._columnHeadersView.setScrollerSpacing(vScrollbarWidth);
+            that._columnHeadersView && that._columnHeadersView.toggleScrollerBorder(vScrollbarWidth);
```
vconst: Make toggleScrollerBorder private by adding prefix "_" and call it inside setScrollerSpacing. Small public API is better.

vconst (changes requested):

js/ui/grid_core/ui.grid_core.grid_view.js
igormalzew: Done

vconst (approved):

jtoming830 (approved):

PR #11058, opened 2019-12-11
DataGrid: Fix SelectAll incorrectly select rows if filtering by search and grid has many columns (T843852)

jtoming830:


testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -17445,6 +17445,68 @@ QUnit.test("Click near selectAll doesn't generate infinite loop", function(asser
     assert.ok($($selectAllElement).find(".dx-select-checkbox").hasClass("dx-checkbox-checked"));
 });
 
+// T843852
+QUnit.test("SelectAll should not select all rows if filter by search is applied and grid has many columns", function(assert) {
```
vconst: Move this test to selection.test.js because integration test with Grid's UI is not necessary for this test.

vconst (changes requested):

vconst (approved):

AntonSermyazhko (approved):

PR #11081, opened 2019-12-12
AsyncRule - Fixes reject

NickMitrokhin:
Promise.reject can accept only a message for displaying validation message.


San4es (approved):
LGTM

PR #11111, opened 2019-12-13
DataGrid: Fix validation error disappears when a new portion of data is loaded during scrolling (T838674)

jtoming830:


vconst (changes requested):
Add test for similar case with adding

vconst (approved):

PR #11199, opened 2019-12-18
Fix virtual scrolling if rowTemplate is defined and jquery is not used (T821418)

vconst:


nightskylark:
Why don't use [repeatMethod](https://github.com/DevExpress/DevExtreme/blob/22920a4b3b27417d92f925b3ef432538a8b9db61/js/core/renderer_base.js#L51)?

vconst:
@nightskylark 
Because inverted iteration is used here. I can add boolean flag "inverted" to the repeatMethod but it will become more complex.

nightskylark (approved):

jtoming830 (approved):

PR #11200, opened 2019-12-18
dxDropDownButton: toggleButton should render inside of DDB even if content width is bigger than DDb width (T847072)

ksercs:


alexander-kotov-dx (changes requested):
const toggleButtonWidth = 26;
Is it actual for all themes and all cases?

PR #11239, opened 2019-12-20
DropDownButton: fix css issues (T847072)

ksercs:
Fixed problems:
1) [1 test, generic theme] toggleButton is shown outside of dropDownButton in generic themes
before: image after: image
2) [1 test, all themes] toggleButton width changes dynamically
before: 
image 
after:    
image
3) [no tests, material theme] Button inside of DDB has min-width, bun popup changes its width if width option is defined
before: image after: image
4) [no tests, material theme] dropDownButton has incorrect height
before: image after: image
5) [1 test, all themes] dropDownButton content is not centered vertically when content width is bigger then dropDownButton width
before:
image
after:
image

PR #11262, opened 2019-12-23
DropDownButton: fix css issues (T847072) (#11239)

ksercs:


alexander-kotov-dx (approved):

San4es (changes requested):
We need to fix tests in IE11 first.

PR #11291, opened 2019-12-24
Renovation: add ES2020 features using babel's plugins

LazyLahtak:
Add Nullish Coalescing Operator:
```js
const result = value ?? defaultValue
```
equal
```js
let result;
if (value !== null && value !== undefined) {
  result = value;
} else {
  result = defaultValue;
}
```
_________________________
Add Optional Chaining:
```js
const obj = {
  field: {
    value: 5,
  },
}

obj?.field?.value     // 5
obj?.subfield?.value  //undefined
```

```js
const obj = {
  func: () => { return 'someText'; },
}

obj?.func?.()         // 'someText'
obj?.anotherFunc?.()  // undefined
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "@types/jquery": "^2.0.34",
     "angular": "^1.6.10",
     "babel-core": "^7.0.0-bridge.0",
+    "babel-eslint": "^10.0.3",
     "babel-loader": "^7.1.2",
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "@babel/preset-env": "^7.7.7",
     "@types/jquery": "^2.0.34",
     "angular": "^1.6.10",
-    "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.3",
     "babel-loader": "^7.1.2",
     "babel-plugin-add-module-exports": "^0.2.1",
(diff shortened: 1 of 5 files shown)


LazyLahtak:
Ready to review

vconst (dismissed):

package.json
```diff
@@ -36,9 +36,14 @@
...
   },
   "devDependencies": {
+    "@babel/core": "^7.7.7",
+    "@babel/plugin-proposal-nullish-coalescing-operator": "^7.7.4",
+    "@babel/plugin-proposal-optional-chaining": "^7.7.5",
+    "@babel/preset-env": "^7.7.7",
     "@types/jquery": "^2.0.34",
     "angular": "^1.6.10",
-    "babel-core": "^6.26.0",
+    "babel-core": "^7.0.0-bridge.0",
```
San4es, line 45: Do we need both `@babel/core` and `babel-core` packages?

LazyLahtak: Thank you, fixed!

San4es (dismissed):

IlyaKhD:
LGTM
Btw, does `babel-eslint` take into account which babel plugins are installed?

IlyaKhD (dismissed):

package.json
LazyLahtak: UPD: we still need `babel-core` for `gulp`

LazyLahtak:
> LGTM
> Btw, does `babel-eslint` take into account which babel plugins are installed?

Yes, it does.
We need to add all used plugins in the `babelrc`

churkin (approved):

San4es (approved):
Happy New "ES2020" Year!

PR #11298, opened 2019-12-25
Fix allDay appointments rendering when groupByDate = true (T845632)

Yarovaya:


Krijovnick (dismissed):

js/core/utils/date.js
```diff
@@ -609,8 +609,12 @@ var makeDate = function(date) {
 };
 
 var getDatesOfInterval = function(startDate, endDate, step) {
-    var currentDate = new Date(startDate.getTime()),
-        result = [];
+    let currentDate = new Date(startDate.getTime());
+    let result = [];
```
Krijovnick: `result` variable can be used as const, 
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Statements/const

Olexi: You're right that is can be `const`, but as for me, `result` should be set as `let` because of it's changing value throughout method execution (like in C-like languages with strict type management).

testing/functional/model/scheduler.ts
```diff
@@ -233,6 +239,10 @@ export default class Scheduler extends Widget {
         return this.dateTableRows.nth(rowIndex).find(`.${CLASS.dateTableCell}`).nth(cellIndex);
     }
 
+    getallDayTableCell(cellIndex: number = 0): Selector {
```
Olexi: Missspelling, `getAllDayTableCell`

Krijovnick (dismissed):

andrewmakarov (approved):

Krijovnick (approved):

PR #11310, opened 2019-12-25
DataGrid: Fix cell border if a cell template is used, showRowLines is false and rowAlternationEnabled is true (T838734)

vconst:


styles/widgets/generic/gridBase.generic.less
```diff
@@ -155,6 +155,9 @@
             & > td,
             & > tr > td {
                 background-color: ~"@{@{widgetName}-row-alternation-bg}";
+                // T838734
```
AntonSermyazhko, line 158: Do we realy need this comment? I suppose a change history should be enough.

jtoming830 (approved):

styles/widgets/generic/gridBase.generic.less
vconst: I think yes, because styles are needed for very specific case.

AntonSermyazhko (approved):

PR #11335, opened 2019-12-27
DataGrid: Fix cell was not highlighted after editing another cell and click (T836391)

jtoming830:


js/ui/grid_core/ui.grid_core.editor_factory.js
```diff
@@ -113,13 +113,34 @@ var EditorFactory = modules.ViewController.inherit({
             }
             that._$focusedElement = $element;
 
+            if($element.is('td')) {
```
AntonSermyazhko: It is better to use the "_getFocusCellSelector()" method

PR #11344, opened 2019-12-27
DataGrid - Refactoring - Extract method in the keyboard navigation (T837977)

AntonSermyazhko:


js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -155,6 +151,14 @@ var KeyboardNavigationController = core.ViewController.inherit({
         });
     },
 
+    _setRowsViewAttributes: function($rowsView) {
+        let isGridEmpty = !this._dataController.getVisibleRows().length;
```
igormalzew: Maybe we should use 'const' for all immutable variables for consistently style

AntonSermyazhko: Sure, I'll correct it

igormalzew (approved):

PR #11347, opened 2019-12-27
TreeList: Fix problem with displaying a changed data when repaintChangesOnly is true and scrolling.rowRenderingMode is virtual (T846709)

Alyar666:
See https://devexpress.com/issue=T846709

### DIFF
--- a/js/ui/tree_list/ui.tree_list.rows.js
+++ b/js/ui/tree_list/ui.tree_list.rows.js
@@ -35,6 +35,8 @@ exports.RowsView = rowsViewModule.views.rowsView.inherit((function() {
         for(let i = 0; i <= level; i++) {
             $container.append(createIcon(i === level && row.node.hasChildren, row.isExpanded));
         }
+
+        return $container;
     };
 
     return {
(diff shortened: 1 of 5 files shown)


js/ui/tree_list/ui.tree_list.rows.js
```diff
@@ -29,36 +29,37 @@ exports.RowsView = rowsViewModule.views.rowsView.inherit((function() {
...
     return {
-        _renderExpandIcon: function($container, options) {
+        _renderIconContainer: function($container, options) {
             const $iconContainer = $('<div>')
                 .addClass(TREELIST_EXPAND_ICON_CONTAINER_CLASS)
                 .appendTo($container);
 
-            renderIcons($iconContainer, options.row);
-
             options.watch && options.watch(function() {
```
AntonSermyazhko, line 38: If we don't have 'options.watch' do we need to render icons by 'this._renderIcons($iconContainer, options)' method?
As it was here: https://github.com/DevExpress/DevExtreme/pull/11347/files#diff-810751d755fc6bec8c41133709b7e5acL46

vconst (approved):

AntonSermyazhko (approved):

PR #11356, opened 2019-12-29
dxScheduler - fix server filtering after switching views(T838165)

andrewmakarov:


testing/tests/DevExpress.ui.widgets.scheduler/integration.appointments.tests.js
```diff
@@ -3910,5 +3910,80 @@ QUnit.module('Appointments', () => {
             assert.ok(eventCallCount === 5, 'appointmentTooltipTemplate should be raised');
         });
     });
+
+    QUnit.test('Remote filter should apply after change view type', function(assert) {
```
Yarovaya, line 3914: Is it possible to write such test in a test cafe?

PR #11387, opened 2019-12-31
Renovation: odata/context es6 refactoring 

LazyLahtak:
Part of refactoring odata/* components

(In plans - get rid of odata/mixins)

### DIFF
--- a/js/data/odata/context.js
+++ b/js/data/odata/context.js
@@ -1,15 +1,13 @@
-const Class = require('../../core/class');
-const extend = require('../../core/utils/extend').extend;
-const typeUtils = require('../../core/utils/type');
-const each = require('../../core/utils/iterator').each;
-const errorsModule = require('../errors');
-const ODataStore = require('./store');
-const mixins = require('./mixins');
-const deferredUtils = require('../../core/utils/deferred');
-const when = deferredUtils.when;
-const Deferred = deferredUtils.Deferred;
+import Class from '../../core/class';
+import { extend } from '../../core/utils/extend';
+import { isDefined, isPlainObject } from '../../core/utils/type';
+import { each } from '../../core/utils/iterator';
+import errorsModule from '../errors';
+import ODataStore from './store';
+import { SharedMethods, formatFunctionInvocationUrl, escapeServiceOperationParams } from './mixins';
+import { when, Deferred } from '../../core/utils/deferred';
 
-require('./query_adapter');
+import './query_adapter';
 
 /**
 * @name ODataContext
@@ -19,7 +17,7 @@ require('./query_adapter');
 */
 const ODataContext = Class.inherit({
 
-    ctor: function(options) {
+    ctor(options) {
         const that = this;
 
         /**
@@ -78,7 +76,7 @@ const ODataContext = Class.inherit({
          * @name ODataContextOptions.entities
          * @type object
          */
-        each(options.entities || [], function(entityAlias, entityOptions) {
+        each(options.entities || [], (entityAlias, entityOptions) => {
             that[entityAlias] = new ODataStore(extend(
                 {},
                 options,
@@ -97,7 +95,7 @@ const ODataContext = Class.inherit({
      * @param2 params:object
      * @return Promise<any>
      */
-    get: function(operationName, params) {
+    get(operationName, params) {
         return this.invoke(operationName, params, 'GET');
     },
 
@@ -109,7 +107,7 @@ const ODataContext = Class.inherit({
      * @param3 httpMethod:object
      * @return Promise<void>
      */
-    invoke: function(operationName, params, httpMethod) {
+    invoke(operationName, params, httpMethod) {
         params = params || {};
         httpMethod = (httpMethod || 'POST').toLowerCase();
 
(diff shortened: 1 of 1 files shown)


js/data/odata/context.js
```diff
@@ -19,7 +17,7 @@ require('./query_adapter');
 */
 const ODataContext = Class.inherit({
 
-    ctor: function(options) {
+    ctor(options) {
         const that = this;
```
nightskylark:
```
const that = this;
```
I suppose we do no need this anymore.

js/data/odata/context.js
```diff
@@ -109,7 +107,7 @@ const ODataContext = Class.inherit({
      * @param3 httpMethod:object
      * @return Promise<void>
      */
-    invoke: function(operationName, params, httpMethod) {
+    invoke(operationName, params, httpMethod) {
         params = params || {};
```
nightskylark: Can be replaced with default value in arguments:
```
invoke(operationName, params = {}, httpMethod);
```

js/data/odata/context.js
```diff
@@ -109,7 +107,7 @@ const ODataContext = Class.inherit({
...
         httpMethod = (httpMethod || 'POST').toLowerCase();
```
nightskylark: Same for the `POST` as a default value

js/data/odata/context.js
```diff
@@ -1,15 +1,13 @@
...
-require('./query_adapter');
+import Class from '../../core/class';
+import { extend } from '../../core/utils/extend';
+import { isDefined, isPlainObject } from '../../core/utils/type';
+import { each } from '../../core/utils/iterator';
+import errorsModule from '../errors';
+import ODataStore from './store';
+import { SharedMethods, formatFunctionInvocationUrl, escapeServiceOperationParams } from './mixins';
+import { when, Deferred } from '../../core/utils/deferred';
+
```
nightskylark: Why do we need this empty line?

nightskylark (changes requested):

js/data/odata/context.js
LazyLahtak: As explained to me, we use `that` for babel.
Babel translating `that` -> `t` to make bundle smaller.
I decided to not change `const that = this;` for now.

LazyLahtak: I think we will discuss this on meeting.

nightskylark: I'm not sure it's a good decision. Nobody will return to this variable later. Also I don't think all DevExtreme developers knows and abides by this arrangement. We should make a finally decision while we are making manual rework of a codebase.

js/data/odata/context.js
```diff
@@ -109,27 +106,26 @@ const ODataContext = Class.inherit({
...
      * @return Promise<void>
      */
-    invoke: function(operationName, params, httpMethod) {
-        params = params || {};
-        httpMethod = (httpMethod || 'POST').toLowerCase();
+    invoke(operationName, params = {}, httpMethod = 'POST') {
+        httpMethod = httpMethod.toLowerCase();
 
         const d = new Deferred();
         let url = this._url + '/' + encodeURIComponent(operationName);
```
churkin:
```javascript
`${this._url}/${encodeURIComponent(operationName)}`
```

js/data/odata/context.js
```diff
@@ -109,27 +106,26 @@ const ODataContext = Class.inherit({
      * @param3 httpMethod:object
      * @return Promise<void>
      */
-    invoke: function(operationName, params, httpMethod) {
-        params = params || {};
-        httpMethod = (httpMethod || 'POST').toLowerCase();
+    invoke(operationName, params = {}, httpMethod = 'POST') {
```
churkin, line 107: Can I call `invoke('name', null, null)` or  `invoke('name', void 0, void 0)`?

js/data/odata/context.js
```diff
@@ -19,7 +16,7 @@ require('./query_adapter');
 */
 const ODataContext = Class.inherit({
```
churkin, line 17: `export default const ODataContext = Class.inherit({`

churkin (changes requested):

js/data/odata/context.js
LazyLahtak: You can call `invoke('name', void 0, void 0)`, `invoke('name', undefined, undefined)` or `invoke('name')` and it will be work correcrtly.
But it isn't work with `invoke('name', null, null)`. Do we have cases, when we call it with `null`?

js/data/odata/context.js
LazyLahtak: We cannot use `export default const` (this construction does not applicable)
We should use `export default Class.inherit({...})` or define `const ODataContext = ...` and make `export default ODataContext` (as it is now)

js/data/odata/context.js
LazyLahtak: After research, we found that we can get rid of `const that = this;` construction

LazyLahtak:
Ready to review

nightskylark (approved):

churkin (approved):

PR #11419, opened 2020-01-09
Renovation: odata/utils es6 refactoring

LazyLahtak:
Part of refactoring odata/* components

(In plans - get rid of odata/mixins)

### DIFF
--- a/js/data/odata/utils.js
+++ b/js/data/odata/utils.js
@@ -1,17 +1,13 @@
-const Class = require('../../core/class');
-const extend = require('../../core/utils/extend').extend;
-const typeUtils = require('../../core/utils/type');
-const iteratorUtils = require('../../core/utils/iterator');
-const each = require('../../core/utils/iterator').each;
-const ajax = require('../../core/utils/ajax');
-const Guid = require('../../core/guid');
-const isDefined = typeUtils.isDefined;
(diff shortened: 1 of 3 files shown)


MaximKudriavtsev (approved):

LazyLahtak:
Ready to review

js/data/odata/utils.js
```diff
@@ -446,45 +403,36 @@ function transformTypes(obj, options) {
...
 
-const serializePropName = function(propName) {
-    if(propName instanceof EdmLiteral) {
-        return propName.valueOf();
-    }
+const serializeString = value => `'${value.replace(/'/g, '\'\'')}'`;
 
-    return propName.replace(/\./g, '/');
-};
+export const serializePropName = propName =>
```
nightskylark, line 412: Just a note. I think it's better to use brackets in multiline arrow functions. Here is an argumentation: https://github.com/airbnb/javascript#arrows--paren-wrap
Maybe we'll add this rule in future.

nightskylark (approved):

js/data/odata/utils.js
LazyLahtak: I think we should discuss this rule on meeting.

PR #11447, opened 2020-01-10
Renovation: add BaseWidget component

MaximKudriavtsev:
This component would be a general component and should consider logic from DomComponent and Widget classes.


js/ui/test-widget.tsx
```diff
@@ -0,0 +1,168 @@
...
+    onPointerOut() {
+        this._hovered = false;
+    }
+
+    @Listen("pointerdown")
+    onPointerDown() {
+        this._active = true;
+    }
+
+    @Listen('pointerup', { target: document })
```
roman-simionov, line 159: use document from core/dom_adapter

PR #11463, opened 2020-01-13
Fix recurrence editor with custom start Day Of Week (T847050)

Olexi:

### DIFF
--- a/js/ui/scheduler/ui.scheduler.recurrence_editor.js
+++ b/js/ui/scheduler/ui.scheduler.recurrence_editor.js
@@ -375,7 +375,7 @@ const RecurrenceEditor = Editor.inherit({
                 onValueChanged: this._repeatByDayValueChangeHandler.bind(this)
             });
 
-            this._daysOfWeek[dayIndex] = day;
+            this._daysOfWeek[i] = day;
             this._$repeatOnWeek.append($day);
         }
     },
@@ -392,10 +392,11 @@ const RecurrenceEditor = Editor.inherit({
 
     _repeatByDayValueChangeHandler() {
         let byDayRule = '';
+        const daysOffset = this._getFirstDayOfWeek();
 
         each(this._daysOfWeek, (index, day) => {
             if(day.option('value')) {
-                const dayName = days[index];
+                const dayName = days[(index + daysOffset) % 7];
 
                 if(!byDayRule) {
                     byDayRule = dayName;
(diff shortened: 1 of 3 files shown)


js/ui/scheduler/ui.scheduler.recurrence_editor.js
```diff
@@ -392,10 +392,11 @@ const RecurrenceEditor = Editor.inherit({
 
     _repeatByDayValueChangeHandler() {
         let byDayRule = '';
+        const daysOffset = this._getFirstDayOfWeek();
 
         each(this._daysOfWeek, (index, day) => {
             if(day.option('value')) {
-                const dayName = days[index];
+                const dayName = days[(index + daysOffset) % 7];
```
Yarovaya: Can you extract method for this calculation?
It is better to use a constant instead of a number. For example, const DAYS_IN_WEEK = 7

Yarovaya (dismissed):

js/ui/scheduler/ui.scheduler.recurrence_editor.js
Olexi: Done with const. Method extracting is unnecessary, because of less-productive code results.

andrewmakarov (dismissed):

js/ui/scheduler/ui.scheduler.recurrence_editor.js
```diff
@@ -362,9 +364,9 @@ const RecurrenceEditor = Editor.inherit({
 
         this._daysOfWeek = [];
 
-        for(let i = 0; i < 7; i++) {
+        for(let i = 0; i < DAYS_IN_WEEK; i++) {
             const daysOffset = this._getFirstDayOfWeek() + i;
-            const dayIndex = daysOffset % 7;
+            const dayIndex = daysOffset % DAYS_IN_WEEK;
```
Krijovnick: There is 3 copy-paste in the file. Lines: 369, 401, 876. It can be in the extra function. Something like this:
```
function getDayIndex() {
  const firstDay = this._getFirstDayOfWeek();
  return function(index) {
      return (index + firstDay ) % DAYS_IN_WEEK;
   }
}
```

Krijovnick (dismissed):

Krijovnick (approved):

andrewmakarov (approved):

Yarovaya (approved):

PR #11479, opened 2020-01-14
Renovation: Data Source refactoring

LazyLahtak:
Separate `Operation Manager` and `Data Source` into different files.
Move functions into separate file.


js/data/data_source/utils.js
```diff
@@ -0,0 +1,133 @@
...
+    switch(originalArguments.length) {
+        case 0:
+            return undefined;
+        case 1:
+            return originalArguments[0];
+    }
+    return [].slice.call(originalArguments);
+};
+
+export const generateStoreLoadOptionAccessor = (optionName) => function() {
```
MaximKudriavtsev: `function()` -> `() => ` and other same places

LazyLahtak: We should return `function` to change context (we use `this`).
Should we pass `this` as `context` to a function?

https://github.com/DevExpress/DevExtreme/pull/11479/files#diff-6b83de546e57031c44bc875e92bbfc24L141-L150

https://github.com/DevExpress/DevExtreme/pull/11479/files#diff-6b83de546e57031c44bc875e92bbfc24L579

LazyLahtak: `export const generateStoreLoadOptionAccessor = (optionName, context) => () => ...`

MaximKudriavtsev: It's bad practice using context inside simple functions.
The best solution is to get rid of using context, but I don't know how to achieve it now.

LazyLahtak: I can try define this function in Data Source class (then we will same context for `this`). Or we can get rid of `generateStoreLoadOptionAccessor` (generator) and use result function in class.

nightskylark (dismissed):

MaximKudriavtsev (approved):

PR #11480, opened 2020-01-14
Fix incorrect RTL appearance for editors with Validator (T851279)

DokaRus:
In case the 'rtlEnabled' option is not set through the `DevExtreme.config()`

### DIFF
--- a/js/ui/validator.js
+++ b/js/ui/validator.js
@@ -178,6 +178,8 @@ const Validator = DOMComponent.inherit({
         }
     },
 
+    _toggleRTLDirection() {},
+
     _initMarkup() {
         this.$element().addClass(VALIDATOR_CLASS);
         this.callBase();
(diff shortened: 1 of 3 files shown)


js/ui/validator.js
```diff
@@ -178,6 +178,12 @@ const Validator = DOMComponent.inherit({
         }
     },
 
+    _toggleRTLDirection(isRtl) {
+        const { rtlEnabled } = this.option('adapter')?.editor?.option?.() || { rtlEnabled: isRtl };
```
San4es:
```suggestion
        const rtlEnabled = this.option('adapter')?.editor?.option('rtlEnabled') ?? isRtl;
```
?

San4es (approved):

alexander-kotov-dx (approved):

PR #11522, opened 2020-01-16
Prevent DDBox popup closing after window scrolling for iPad in desktop mode (T845484)

alexander-kotov-dx:


js/ui/drop_down_box.js
```diff
@@ -265,6 +268,16 @@ const DropDownBox = DropDownEditor.inherit({
         return this.callBase();
     },
 
+    _preventCloseHandler: function() {
+        const isInputFocused = !!this._popup.$content().find('input:focus').length;
```
DokaRus:
```js
import { getActiveElement } from '../core/dom_adapter';
<...>
_isNestedElementActive: function() {
  const activeElement = getActiveElement();
  return activeElement && this._popup.$content().get(0).contains(activeElement);
}

_shouldCloseOnTargetScroll: function() {
  return realDevice.deviceType === 'desktop' && this._canShowVirtualKeyboard() && this._isNestedElementActive();
}

```

DokaRus (dismissed):

js/ui/drop_down_box.js
```diff
@@ -213,6 +217,25 @@ const DropDownBox = DropDownEditor.inherit({
         return this.callBase();
     },
 
+    ///#DEBUG
```
San4es: My suggestion is to export `realDevice` object from the current module:

```
///#DEBUG
module.exports.realDevice = realDevice;
```
and use it directly in the test.

So, we can avoid adding a new method to the component.

testing/tests/DevExpress.ui.widgets.editors/dropDownBox.tests.js
```diff
@@ -502,6 +506,39 @@ QUnit.module('popup options', moduleConfig, () => {
...
+                assert.expect(0);
+                return;
+            }
+
+            const $content = $('<input type="text" />');
+            const instance = new DropDownBox($('#dropDownBox'), {
+                contentTemplate: () => $content
+            });
+
+            instance.setRealDevice({ mac: isMac });
```
San4es: We should restore an original real device object at the test end.

San4es (approved):
LGTM

DokaRus (approved):

PR #11541, opened 2020-01-16
 dxLookup - Fix height when a search bar and a cancel button are displayed in the Material theme (T849526)

groshenkovamarina:


babich-a:
We should fix tests at first

babich-a (approved):

hakimio:
@groshenkovamarina can you take a look at T853111?

PR #11542, opened 2020-01-16
DataGrid - DateBox does not update a cell via a keyboard (T848039 / Refix T819067)

igormalzew:


testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -12713,7 +12713,8 @@ QUnit.testInActiveWindow('Datebox editor\'s enter key handler should be replaced
 
     // assert
     const dateBox = rowsViewWrapper.getEditor(0, 0).dxDateBox('instance');
-    assert.equal(dateBox._enterHandler, commonUtils.noop, 'dateBox enter key handler for mask is replaced');
+    const enterKeyHandler = dateBox._supportedKeys().enter;
+    assert.equal(JSON.stringify(enterKeyHandler), JSON.stringify(() => true), 'dateBox enter key handler is replaced');
```
vconst: Does this check work in IE? I think it will be better to check function result:
assert.strictEqual(enterKeyHandler(), true, 'dateBox enter key handler is replaced');

igormalzew: Replaced check condition in test

vconst (approved):

AntonSermyazhko (approved):

PR #11548, opened 2020-01-17
Renovation: devices es6 refactoring

LazyLahtak:
Refactor devices into es6 class.


MaximKudriavtsev (dismissed):

js/core/devices.js
```diff
@@ -156,7 +156,7 @@ const uaParsers = {
  * @module core/devices
  * @export default
  */
-const Devices = Class.inherit({
```
MaximKudriavtsev, line 112: export class Devices... ?

nightskylark (dismissed):

js/core/devices.js
LazyLahtak: We don't use Devices as class in other components. We create `const devices = new Devices();` and export it.
There is no need to `export class Devices`, but we can do it.

MaximKudriavtsev (approved):

PR #11556, opened 2020-01-17
Renovation: window refactor

LazyLahtak:
Get rid of using `this` in `window` (allow named export for all functions)

### DIFF
--- a/js/core/utils/window.js
+++ b/js/core/utils/window.js
@@ -2,49 +2,40 @@
 
 const domAdapter = require('../dom_adapter');
 
-const hasWindow = typeof window !== 'undefined';
-let windowObject = hasWindow && window;
+const windowObject = window ?? { window: windowObject };
 
-if(!windowObject) {
-    windowObject = {};
-    windowObject.window = windowObject;
-}
+const hasWindow = () => typeof window !== 'undefined';
+
+const getWindow = () => windowObject;
+
+const hasProperty = (prop) => hasWindow() && prop in windowObject;
+
+const defaultScreenFactorFunc = (width) => {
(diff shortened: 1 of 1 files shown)


js/core/utils/window.js
```diff
@@ -2,49 +2,40 @@
...
+    const windowWidth = domAdapter.getDocumentElement()['clientWidth'];
+
+    return screenFactorFunc(windowWidth);
+};
+
+const getNavigator = () => hasWindow() ? windowObject.navigator : { userAgent: '' };
+
+module.exports = {
+    hasWindow,
+    getWindow,
```
MaximKudriavtsev, line 36: Why aren't we using `export const ...` ?

LazyLahtak: In some places we use
```js
const windowUtils = require('../core/utils/window');
```
This is incompatible with
```js
export const hasWindow = ...;
export const getWindow = ...;
```
or with
```js
export {
    hasWindow,
    getWindow,
    ...
}
```

If we want to use it, we should get rid of all `windowUtils = require`. Should we do that in this PR?

MaximKudriavtsev: I think we want to do it.

> Should we do that in this PR?

Yes, we can.

churkin (approved):

js/core/utils/window.js
LazyLahtak: We have too many places, where we use `require`. It's quite difficult to change them all in same time (many dependencies between components).

MaximKudriavtsev (approved):

nightskylark (approved):

PR #11572, opened 2020-01-17
Fix compound keys comparison (T852818)

dxvladislavvolkov:

### DIFF
--- a/js/core/utils/common.js
+++ b/js/core/utils/common.js
@@ -296,10 +296,7 @@ const objectsEqualByValue = function(object1, object2, depth, strict) {
 
 const maxEqualityDepth = 3;
 
-const equalByValue = function(object1, object2, depth, strict) {
-    depth = depth || 0;
-    strict = isDefined(strict) ? strict : true;
-
+const equalByValue = function(object1, object2, depth = 0, strict = true) {
     object1 = toComparable(object1, true);
     object2 = toComparable(object2, true);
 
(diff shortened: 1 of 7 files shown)


js/data/utils.js
```diff
@@ -153,14 +153,22 @@ function isConjunctiveOperator(condition) {
...
+function getKeyValue(keyName, key) {
+    let value;
+    keyName.split('.').forEach((name) => {
+        value = value ? value[name] : key[name];
+    });
+
+    return value;
+}
+
 const keysEqual = function(keyExpr, key1, key2) {
```
nightskylark: The `keyExpr` variable was not used in a comparison since the method was written (4c89bcaff145 in hg). @AlekseyMartynov , can you explain, why?

js/core/utils/common.js
```diff
@@ -296,18 +296,22 @@ const objectsEqualByValue = function(object1, object2, depth) {
 
 const maxEqualityDepth = 3;
 
-const equalByValue = function(object1, object2, depth) {
+const equalByValue = function(object1, object2, depth, strict) {
```
nightskylark:
```suggestion
const equalByValue = function(object1, object2, depth = 0, strict = true) {
```

nightskylark (changes requested):

nightskylark (dismissed):

nightskylark (dismissed):

AlekseyMartynov (approved):

PR #11585, opened 2020-01-20
Support debugging in the ES6 code 

jsdmitry:
This PR resolve problems with debugging our code after used the Babel library in the tests and the playground. 

What was done:
1. The devtool option of the WebPack is changed for using the source map
2. Use the gulp-sourcemaps package for debugging tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,7 @@
     "build-themes": "gulp style-compiler-themes",
     "build-themebuilder-assets": "gulp style-compiler-tb-assets",
     "dev": "gulp dev",
+    "dev-map": "gulp dev --sourceMap",
     "test-env": "node testing/launch",
     "update-ts": "dx-tools update-ts-bundle",
     "internal-tool": "dx-tools",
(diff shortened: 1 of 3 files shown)


DokaRus (dismissed):

andrewmakarov (dismissed):

package.json
```diff
@@ -136,6 +136,7 @@
     "build-themes": "gulp style-compiler-themes",
     "build-themebuilder-assets": "gulp style-compiler-tb-assets",
     "dev": "gulp dev",
+    "dev-map": "gulp dev --sourceMap",
```
San4es: I think we need one "dev" script.

nightskylark: I've discussed it with Dima and proposed to rename tasks in such way
dev-map ->dev
dev -> dev-ie (Because it shuold be left only for the IE debugging)

AlekseyMartynov:
@jsdmitry 
Please add a description. What is the connection between source map type and ES6?

I suggest the following argument:
`gulp --webpack-debug-devtool=...`
This will be less implicit.

San4es (approved):

NickMitrokhin (approved):

nightskylark (approved):

andrewmakarov (approved):

PR #11594, opened 2020-01-20
File Manager - 'Refresh' button relocation

Staziz:

### DIFF
--- a/js/ui/file_manager/ui.file_manager.toolbar.js
+++ b/js/ui/file_manager/ui.file_manager.toolbar.js
@@ -145,24 +145,45 @@ class FileManagerToolbar extends Widget {
     }
 
     _getPreparedItems(items) {
-        let groupHasItems = false;
-
-        return items.map(item => {
+        items = items.map(item => {
             const commandName = isString(item) ? item : item.name;
             const preparedItem = this._configureItemByCommandName(commandName, item);
 
-            if(commandName === 'separator') {
-                preparedItem.visible = groupHasItems;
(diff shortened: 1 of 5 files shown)


js/ui/file_manager/ui.file_manager.toolbar.js
```diff
@@ -145,24 +145,45 @@ class FileManagerToolbar extends Widget {
...
+        }
+        return items;
+    }
+
+    _groupHasItems(items, separatorIndex, isBefore) {
+        const delta = isBefore ? -1 : 1;
+        let i = separatorIndex + delta;
+        let result = false;
+        while(items[i] && items[i].name !== 'separator') {
+            result = result || items[i].visible;
```
vladkovl: We can exit from function after first visible item has been found.

js/ui/file_manager/ui.file_manager.toolbar.js
```diff
@@ -330,20 +351,15 @@ class FileManagerToolbar extends Widget {
...
 
         items.forEach((item, index) => {
             const itemVisible = item.available;
 
             let showItem = false;
-            if(item.name === 'separator') {
-                showItem = groupHasItems;
-                groupHasItems = false;
-            } else {
+            if(item.name !== 'separator') {
```
vladkovl, line 358: All separators will have `showItem = false`, so they all will be hidden after this cycle? Maybe we should refactor this code.

vladkovl (approved):

PR #11617, opened 2020-01-21
QUnit: Check for uncleared timers with no jQuery

San4es:


nightskylark (approved):

nightskylark:
How about to get rid of jquery in the entire file?

San4es:
@nightskylark 

>How about to get rid of jquery in the entire file?

Good idea! I think it can be done in a separate PR.

DokaRus (approved):
LGTM

PR #11618, opened 2020-01-21
SelectBox and TagBox should have the correct "event" field of the "onValueChanged" event after adding an custom item (T853677)

DokaRus:


San4es (approved):
LGTM

alexander-kotov-dx (approved):

PR #11629, opened 2020-01-22
Less refactoring for scss generation (icons, typography, button, buttonGroup, scrollView)

babich-a:
This PR is created to simplify less->scss migration.
The following changes were made:

**Move some validation constants to the base section**, **Move base radiuses to the base section** - some constants are used in many places, that's why they were moved to the base section.

**Move @badge-invalid-bg to the place where it used** - @badge-invalid-bg was removed from the theme files to the common.(theme).less (where it used once).

**Change the order of variables** - scss can't use undefined variables.

**Extract S font mixin**, **typography mixin braces** - scss can't use selector as a mixin.

**dx-icon-sizing mixin refactoring** - scss can't use mixin overloading.

**Change the order for parent selectors** - cause of https://github.com/sass/sass/issues/1425 we need to use a workaround in scss generator. But it has some restrictions:

works:

```
@at-root .dx-rtl#{&},
.dx-rtl & {
    margin-left: $size;
}
```

does not work:
```
.dx-rtl &,
@at-root .dx-rtl#{&}, {
    margin-left: $size;
}
```

**Add comments for button group less constants** - helps generator to move the constants to the groupButton, not to the button


package-lock.json
```diff
@@ -0,0 +1,18150 @@
+{
```
San4es: Do we add this file intentionally?

styles/widgets/generic/color-schemes/contrast/generic.contrast.less
```diff
@@ -48,14 +48,20 @@
 @typography-link-color: @base-link-color;
 @base-disabled-opacity: 0.5;
 
-// Validation
```
San4es, line 51: This change is not clear to me

package-lock.json
babich-a: I'll remove it

styles/widgets/generic/color-schemes/contrast/generic.contrast.less
babich-a: @base-invalid-faded-border-color was moved to the base section (above Validation). Git diff here is not clear.

San4es (approved):
LGTM

PR #11649, opened 2020-01-22
DataGrid - A blue border is shown when a cell is clicked several times / Fix row editing mode (T850219)

igormalzew:


js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -748,9 +748,11 @@ const KeyboardNavigationController = core.ViewController.inherit({
 
     _pointerDownHandler: function(e) {
         const $target = $(e.event.target);
+        const isEditedRow = $target.closest('tr').hasClass('dx-edit-row');
```
AntonSermyazhko: I suppose here you can use _isEditorCell_ method instead

AntonSermyazhko (approved):

jtoming830 (approved):

PR #11662, opened 2020-01-23
Temporary workaround for Widget JSX Constructor

roman-simionov:


js/ui/test-button.tsx
```diff
@@ -2,6 +2,9 @@ import { Component, Prop, React } from "../component_declaration/common";
 import { getImageSourceType } from '../core/utils/icon';
 
 import Widget from './test-widget';
+import JSXConstructor from "../component_declaration/jsx";
```
MaximKudriavtsev: We are using `'` quotes instead of `"`

js/component_declaration/jsx.d.ts
```diff
@@ -0,0 +1,6 @@
+import "react";
```
MaximKudriavtsev: We are using `'` quotes instead of `"`

js/component_declaration/jsx-g.ts
```diff
@@ -0,0 +1,4 @@
+
```
MaximKudriavtsev: Empty string

MaximKudriavtsev (approved):

PR #11687, opened 2020-01-24
Fix DropDownEditor Popup position depending the "rtlEnabled" option value (T856114)

DokaRus:

### DIFF
--- a/testing/tests/DevExpress.core/utils.position.tests.js
+++ b/testing/tests/DevExpress.core/utils.position.tests.js
@@ -1,0 +1,31 @@
+import config from 'core/config';
+import { getDefaultAlignment } from 'core/utils/position.js';
+
+const { module: testModule, test } = QUnit;
+
+testModule('getDefaultAlignment', function() {
+    test('getDefaultAlignment should return an alignment depending on the global "rtlEnabled" config or passed value', function(assert) {
+        const originalConfig = config();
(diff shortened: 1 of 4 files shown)


testing/tests/DevExpress.core/utils.position.tests.js
```diff
@@ -0,0 +1,31 @@
+import config from 'core/config';
```
San4es, line 1: Let's merge this file with this one:
https://github.com/DevExpress/DevExtreme/blob/fefd39b7ef1acf829100378d8aec47ea384f1bce/testing/tests/DevExpress.animation/position.tests.js

DokaRus: I don't think we should do that. The positioning utilities located in `animation/utils` are used to support the animation module.
This one (`core/position`) is private utility, used to get start direction or [base alignment](https://github.com/DevExpress/DevExtreme/blob/20_1/js/ui/data_grid/ui.data_grid.export.js#L113) based on passed argument or global `rtlEnabled` config.

San4es: Thanks! My mistake...

San4es (approved):
LGTM

alexander-kotov-dx (approved):

PR #11735, opened 2020-01-28
Renovation: implement Button.stylingMode

churkin:


js/renovation/button.tsx
```diff
@@ -1,56 +1,35 @@
...
-
-    if (model.stylingMode === 'outlined') {
-        classNames.push('dx-button-mode-outlined');
-    } else if (model.stylingMode === 'text') {
-        classNames.push('dx-button-mode-text');
-    } else {
-        classNames.push('dx-button-mode-contained');
-    }
+    const classNames = ['dx-button'].concat(model.classNames);
+    const isValidStylingMode = ['outlined', 'text', 'contained'].indexOf(model.stylingMode) !== -1;
```
MaximKudriavtsev: We can extract these string arrays instead of creating them each time

roman-simionov (dismissed):

LazyLahtak (approved):

js/renovation/button.tsx
```diff
@@ -16,16 +16,18 @@ const getImageContainerJSX = (source: string) => {
 };
 
 const stylingModes = ['outlined', 'text', 'contained'];
+const defaultClassNames = ['dx-button'];
 
 const getCssClasses = (model: any) => {
-    const classNames = ['dx-button'].concat(model.classNames);
-    const isValidStylingMode = stylingModes.indexOf(model.stylingMode) !== -1;
+    const { text, icon, stylingMode, type } = model;
```
MaximKudriavtsev, line 22: We can use argument destruction here

```js
const getCssClasses = ({ text, icon, stylingMode, type })
```

churkin: We will miss 'classNames'

MaximKudriavtsev: We can use alias name
```js
const getCssClasses = ({ text, icon, stylingMode, type, classNames: classes })
```

MaximKudriavtsev (approved):

PR #11711, opened 2020-01-27
Renovation: implement base widget focus and add icon test

LazyLahtak:
1) Implement focus for widget component.
2) Add test for button's icon

PR #11743, opened 2020-01-28
Renovation: Implement default button behavior

LazyLahtak:
Implemen base behavior of button:
1) hoverStateEnabled = true
2) activeStateEnabled = true
3) possible to click on button without `accessKey` property

Will be compatible with focus (#11711) after generator fix.

### DIFF
--- a/js/renovation/widget.tsx
+++ b/js/renovation/widget.tsx
@@ -205,7 +205,7 @@ export default class Widget {
     }
 
     @Effect()
-    clickEffect() {
+    accessKeyEffect() {
         const namespace = 'UIFeedback';
         const isFocusable = this.focusStateEnabled && !this.disabled;
         const canBeFocusedByKey = isFocusable && this.accessKey;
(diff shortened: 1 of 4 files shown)


MaximKudriavtsev:
Should we add some tests?

js/renovation/widget.tsx
```diff
@@ -269,6 +268,18 @@ export default class Widget {
...
+    clickEffect() {
+        const namespace = 'UIFeedback';
+
+        dxClick.on(this.widgetRef,
+            () => this.onClick!(this.clickArgs),
+            { namespace }
+        );
+
+        return () => dxClick.off(this.widgetRef, { namespace });
+    }
```
MaximKudriavtsev: - should we make the subscription without onClick property?
- should we combine `namespace` with `thus.name`?

LazyLahtak: 1) Widget (and Button) still work even if we won't pass `onClick`. Do you mean that?
2) You mean that we should make  'namespace = `UIFeedback${this.name}`'?
Also, the 'this.name' at now is always `Widget` (even if it is `Button`). I think we should fix it.

MaximKudriavtsev: About 2
image
Original button code. namespace = this.name

MaximKudriavtsev (approved):

churkin (approved):

PR #11763, opened 2020-01-29
Less: change icon mixin from 'selector' to 'parametrized'

babich-a:
It is not possible to use a selector as mixin in sass. That's why we create .dx-icon(@icon) mixin.

### DIFF
--- a/styles/widgets/base/icons.less
+++ b/styles/widgets/base/icons.less
@@ -67,271 +67,209 @@
     }
 }
 
-.dx-font-icons-list() {
-    .dx-icon-add { .dx-font-icon("\f00b"); }
-    .dx-icon-airplane { .dx-font-icon("\f000"); }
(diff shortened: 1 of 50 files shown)


styles/widgets/base/icons.less
```diff
@@ -67,271 +67,209 @@
...
+    showpanel: "\f11d";
+}
+// stylelint-enable
 
-    .dx-icon-showpanel {
-        .dx-font-icon("\f11d");
-        .dx-rtl &:before { content: "\f11c"; } // hidepanel icon
-    }
+.dx-icon(@name) {
+    .dx-font-icon(@icons[$@name]);
```
San4es, line 263: fy: https://stackoverflow.com/questions/57823508/how-can-i-get-less-map-value-by-variable#comment102081559_57823508

San4es (approved):
LGTM

PR #11765, opened 2020-01-29
QUnit: Rework API for ignoring uncleared timers

San4es:
See [commits](https://github.com/DevExpress/DevExtreme/pull/11765/commits) for more details.

### DIFF
--- a/testing/helpers/qunitExtensions.js
+++ b/testing/helpers/qunitExtensions.js
@@ -164,17 +164,6 @@
         tooltip: 'Enabling this will test if any test introduces timers (setTimeout, setInterval, ....) and does not cleared them on test finalization. Stored as query-strings.'
     });
 
-    QUnit.timerIgnoringCheckers = (function() {
-        const noop = function() { };
-
-        return {
-            register: noop,
-            unregister: noop,
-            clear: noop,
-            needSkip: noop
-        };
-    })();
-
     const createMethodWrapper = function(method, callbacks) {
         const originalMethod = method;
         const beforeCall = callbacks['beforeCall'];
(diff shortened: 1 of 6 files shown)


DokaRus (approved):

testing/helpers/qunitExtensions.js
```diff
@@ -137,13 +137,30 @@
         });
     };
 
+    const beforeTestDoneCallbacks = [];
+
+    QUnit.beforeTestDone = function(callback) {
```
nightskylark, line 142: Why you are not using arrow functions here?

testing/helpers/qunitExtensions.js
```diff
@@ -363,176 +366,121 @@
...
-                    skip = true;
-                    return false;
-                }
-            });
+    const logTestFailure = function(timerInfo) {
+        const timeoutString = timerInfo.timeout ? '\nTimeout: ' + timerInfo.timeout : '';
 
-            return skip;
-        };
+        const message = [
```
nightskylark, line 435: Why don't use a multiline string?
```
        const message = `Not cleared timer detected.
Timer type: ${timerInfo.timerType}
Id: ${timerInfo.timerId}
Callback:
${timerInfo.callback}
${timeoutString}`;
```

San4es: Interpolated strings are not supported in IE:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

testing/helpers/qunitExtensions.js
San4es: Arrow functions are not supported in IE:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

PR #11800, opened 2020-01-31
Fix passing of the event that caused the value to change when clicking the List item of the DateBox time picker (T858107)

DokaRus:


js/ui/date_box/ui.date_box.strategy.js
```diff
@@ -90,12 +90,6 @@ const DateBoxStrategy = Class.inherit({
         this._widget && this._widget.option('value', this.dateBoxValue());
     },
 
-    _valueChangedHandler: function(args) {
```
DokaRus, line 93: Unused code. This handler is used only in a calendar strategy, but in a calendar strategy this handler is overridden.

alexander-kotov-dx (approved):

San4es (approved):

PR #11847, opened 2020-02-03
DropDownButton: add some SSR tests

ksercs:


testing/tests/DevExpress.serverSide/dropDownButton.tests.js
```diff
@@ -0,0 +1 @@
+require('../DevExpress.ui.widgets/dropDownButton.base.tests.js');
```
DokaRus:
```suggestion
import '../DevExpress.ui.widgets/dropDownButton.base.tests.js';
```

DokaRus (commented):
I think dropDownButton.tests should be extend by the "optionChange" tests module. Some of tests can have a different count of assertions in SSR and client mode (you can check the `core/utils/window => hasWindow` state)

testing/tests/DevExpress.serverSide/dropDownButton.tests.js
```diff
@@ -0,0 +1 @@
+import '../DevExpress.ui.widgets/dropDownButton.base.tests.js';
```
nightskylark: We named same tests for other widgets as `*.markup.tests.js`. Why did you use another name?

ksercs: Because there are not only tests for markup in this file, but also option change tests, etc. But I agree that I should still name it in accordance with tests for other widgets

San4es (commented):
Some tests are failed with:
"Error: Window fields using is prevented"

js/ui/drop_down_button.js
```diff
@@ -331,7 +331,7 @@ const DropDownButton = Widget.inherit({
                     y: -1
                 }
             }
-        }, this._options.cache('dropDownOptions'));
+        }, this._options.cache('dropDownOptions'), { visible: this.option('opened') });
```
San4es, line 337: Is this change necessary? What test covers this?

js/ui/drop_down_button.js
```diff
@@ -519,14 +523,18 @@ const DropDownButton = Widget.inherit({
...
                 break;
             case 'keyExpr':
                 this._compileKeyGetter();
+                this._setListOption(name, value);
                 break;
             case 'buttonGroupOptions':
                 this._innerWidgetOptionChanged(this._buttonGroup, args);
                 break;
             case 'dropDownOptions':
+                args.value.visible && delete args.value.visible;
```
San4es: Is this change necessary? What test covers this?

ksercs: https://github.com/DevExpress/DevExtreme/pull/11847/files#diff-d9fa2b0f966e8a8b29e5b1d170e9ea05R599

js/ui/drop_down_button.js
ksercs: https://github.com/DevExpress/DevExtreme/pull/11847/files#diff-0360e44628ce8bae07db8d6360fc7ba4R103

js/ui/drop_down_button.js
```diff
@@ -424,6 +424,10 @@ const DropDownButton = Widget.inherit({
     },
 
     toggle(visible) {
+        if(!windowUtils.hasWindow()) {
```
San4es: Let's get rid of this and use hasWidow() check here:
https://github.com/DevExpress/DevExtreme/pull/11847/files#diff-cc46f04faa146e60076fc47eff5a47c1R309

js/ui/drop_down_button.js
```diff
@@ -588,7 +598,7 @@ const DropDownButton = Widget.inherit({
                 this._createSelectionChangedAction();
                 break;
             case 'deferRendering':
-                if(!value && !this._popup) {
+                if(windowUtils.hasWindow() && !value && !this._popup) {
```
San4es: See this comment:
https://github.com/DevExpress/DevExtreme/pull/11847/files#diff-cc46f04faa146e60076fc47eff5a47c1R427

js/ui/drop_down_button.js
San4es: Do we need to do this for `deferRendering` as well?

js/ui/drop_down_button.js
```diff
@@ -588,10 +597,8 @@ const DropDownButton = Widget.inherit({
                 this._createSelectionChangedAction();
                 break;
             case 'deferRendering':
-                if(!value && !this._popup) {
-                    this._renderPopup();
-                    this._renderContent();
-                }
+                this._renderPopup();
```
San4es: Do we need to render popup if it is already rendered? To answer this question, let's cover this case by a separate test.

js/ui/drop_down_button.js
ksercs: No, `deferRendering` works correct, because `dropDownOptions.deferRendering` doesn't affect on popup:
- `deferRendering` is true: because the popup widget has not been created yet. Option will affect only on popup init.
- `deferRendering` is false: because the popup widget has already been created

js/ui/drop_down_button.js
ksercs: fixed

js/ui/drop_down_button.js
ksercs: fixed

js/ui/drop_down_button.js
ksercs: fixed it

testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
```diff
@@ -633,6 +633,16 @@ QUnit.module('common use cases', {
         assert.strictEqual(getActionButton(this.dropDownButton).text(), 'Trial for Visual Studio', 'action button has been changed');
     });
 
+    QUnit.test('deferRendering should not do anything if popup has already been rendered', function(assert) {
```
San4es, line 636: Wouldn't it be better to check the popup element strict equality between deferRendering changes to avoid private method usage?

San4es (dismissed):
LGTM

testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
ksercs: It's good idea, thanks

js/ui/drop_down_button.js
```diff
@@ -519,14 +522,20 @@ const DropDownButton = Widget.inherit({
...
             case 'keyExpr':
                 this._compileKeyGetter();
+                this._setListOption(name, value);
                 break;
             case 'buttonGroupOptions':
                 this._innerWidgetOptionChanged(this._buttonGroup, args);
                 break;
             case 'dropDownOptions':
+                if(args.value.visible !== undefined) {
+                    delete args.value.visible;
```
DokaRus, line 540: `instance.option('dropDownOptions.visible', true)` raises `valueChanged` event with args: 
```
{
  name: "dropDownOptions",
  fullName: "dropDownOptions.visible",
  value: true
}

Comment (ksercs) [reply js/ui/drop_down_button.js:540]: fixed it, thanks

File: testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js (hunk 8 above)
Comment (San4es) [reply testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js:636]: I meant popup elements :)

Comment (ksercs) [reply testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js:636]: I understood, and even tested it in the playground, but it didn't stop me from screwing up :-D

File: testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js [hunk 10]
```diff
@@ -794,6 +633,16 @@ QUnit.module('common use cases', {
...
+        const $dropDownButton = this.dropDownButton._popup.$element();
```
San4es: What about using `getPopup`:
https://github.com/DevExpress/DevExtreme/blob/ebc39a2da672b7e3dfd16eabe72012f1548896b1/testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js#L28-L30

San4es (approved):

DokaRus (approved):

PR #11879, opened 2020-02-04
Update ja.json

AlisherAmonulloev:
Added localizations for Max and Min

You can verify translations by using the https://www.microsoft.com/en-us/language/Search service.


AlekseyMartynov:
Rooks good to me 🌏

BingoRUS (changes requested):
We have to localize `dxDataGrid-summaryAvg` too.

BingoRUS (approved):

PR #11855, opened 2020-02-04
DOTNET_USE_POLLING_FILE_WATCHER=true (docker-ci)

AlekseyMartynov:
Got this error:
```
Application startup exception: System.IO.IOException: The configured user limit (128) on the number of inotify instances has been reached.
   at System.IO.FileSystemWatcher.StartRaisingEvents()
   at System.IO.FileSystemWatcher.StartRaisingEventsIfNotDisposed()
   at System.IO.FileSystemWatcher.set_EnableRaisingEvents(Boolean value)
   at Microsoft.Extensions.FileProviders.Physical.PhysicalFilesWatcher.TryEnableFileSystemWatcher()
   at Microsoft.Extensions.FileProviders.Physical.PhysicalFilesWatcher.CreateFileChangeToken(String filter)
   at Microsoft.AspNetCore.Mvc.RazorPages.Internal.PageActionDescriptorChangeProvider.GetChangeToken()
   at Microsoft.Extensions.Primitives.ChangeToken.OnChange(Func`1 changeTokenProducer, Action changeTokenConsumer)
   at Microsoft.AspNetCore.Mvc.Internal.ActionDescriptorCollectionProvider..ctor(IEnumerable`1 actionDescriptorProviders, IEnumerable`1 actionDescriptorChangeProviders)
--- End of stack trace from previous location where exception was thrown ---
   . . .
   at Microsoft.AspNetCore.Builder.MvcApplicationBuilderExtensions.UseMvc(IApplicationBuilder app, Action`1 configureRoutes)
   at Runner.Program.<>c__DisplayClass0_1.<Main>b__1(IApplicationBuilder app) in /drone/src/github.com/AlekseyMartynov/DevExtreme/testing/runner/Program.cs:line 49
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.BuildApplication()
The configured user limit (128) on the number of inotify instances has been reached.
```

Relevant issues:
- https://github.com/dotnet/aspnetcore/issues/7531
- https://github.com/dotnet/aspnetcore/issues/8449
- https://github.com/dotnet/runtime/issues/27272

From https://github.com/aspnet/Announcements/issues/186:
> The polling file watchers are mostly used in Docker when running code from a shared volume.

This is our case (`testing/runner` runs from a volume).

Official SDK images have this flag enabled: https://github.com/dotnet/dotnet-docker/pull/428

PR #11897, opened 2020-02-05
Don't hang if QUnit runner is broken (docker-ci)

AlekseyMartynov:
Lesson learned from #11855

When runner fails to start:
i1

When runner doesn't respond:
i2

### DIFF
--- a/docker-ci.sh
+++ b/docker-ci.sh
@@ -62,8 +62,6 @@ function run_test {
 
     dotnet ./testing/runner/bin/runner.dll --single-run & runner_pid=$!
 
-    echo "URL: $url"
-
     for i in {15..0}; do
         if [ -n "$runner_pid" ] && [ ! -e "/proc/$runner_pid" ]; then
             echo "Runner exited unexpectedly"
@@ -80,6 +78,8 @@ function run_test {
         sleep 1
         echo "Waiting for runner..."
     done
+
+    echo "URL: $url"
 
     case "$BROWSER" in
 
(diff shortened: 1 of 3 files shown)


docker-ci.sh
```diff
@@ -62,8 +62,21 @@ function run_test {
 
     dotnet ./testing/runner/bin/runner.dll --single-run & runner_pid=$!
 
-    while ! httping -qc1 $url; do
+    for i in {15..0}; do
```
AntonSermyazhko: How about extracting this code into a separate function?

docker-ci.sh
```diff
@@ -62,8 +62,21 @@ function run_test {
...
+        httping -qsc1 $url && break
+
+        if [ $i -eq 0 ]; then
+            echo "Runner not reached"
+            exit 1
+        fi
+
         sleep 1
+        echo "Waiting for runner..."
     done
```
AntonSermyazhko, line 80: How about extracting this code into a separate function?

AlekseyMartynov: I had this idea but rejected it because of trade-offs:
- https://stackoverflow.com/q/6212219
- https://stackoverflow.com/q/9893667

San4es (approved):

docker-ci.sh
AntonSermyazhko: Ok

AntonSermyazhko (approved):

PR #11899, opened 2020-02-05
Renovation: implement Button.template

LazyLahtak:
Default value of the `template` property changed from `'content'` to `''` (possibly BC)


testing/jest/button.tests.tsx
```diff
@@ -210,15 +210,46 @@ describe('Button', () => {
...
+                expect(buttonContent.props().text).toBe('My button');
+                expect(buttonContent.render().text()).toBe('My button123');
+
+                button.setProps({ text: 'New value' });
+                buttonContent = button.find(contentRender);
 
-                expect(buttonContentChildren.props().text).toBe('My button');
-                expect(buttonContentChildren.render().text()).toBe('My button123');
+                expect(buttonContent.props().text).toBe('New value');
+                expect(buttonContent.render().text()).toBe('New value123');
```
MaximKudriavtsev: Why we should call `render` method manually?

LazyLahtak: Remained from previous tests.
Refactored.

js/renovation/dist/button.j.jsx
```diff
@@ -10,22 +14,30 @@ class Button extends Widget {
...
-                        template.render({ model: data, container: parent });
-                        parent.appendChild(element);
-                    }
-                }}/>);
+
+        if(props.template) {
+            props.contentRender = ({ text, icon, contentRef, children }) => {
+                const data = { text, icon };
+                let $content = $('<div>');
+                $content.addClass('dx-button-content');
```
pavelgruba: We expect this file to be generated, so it can not contain any control specific content like class names (dx-button-content) etc.
Is there any way to move it somewhere or get it from declarative component?

LazyLahtak: We can move them into properties:
```jsx
props.contentRender = ({ text, icon, className, wrapperClassName, contentRef, children }) => {
    // ...
    $content.addClass(className);
    // ...
    if ($template.hasClass(wrapperClassName)) { }
    // ...
}
```

pavelgruba (dismissed):

js/renovation/button.tsx
```diff
@@ -16,6 +16,16 @@ const getImageContainerJSX = (source: string, position: string) => {
     }
 };
 
+const CustomRender = ({ text, icon, renderFn, contentRef, children }) => {
+    const result = renderFn({
+        model: { text, icon },
+    });
```
MaximKudriavtsev: Why have we created a  `model` object?
```js
const result = renderFn({ text, icon });
```

testing/jest/button.tests.tsx
```diff
@@ -210,15 +210,46 @@ describe('Button', () => {
...
+            it('should rerender contentRender in runtime', () => {
+                const button = mount(<Button text='My button' />);
+                
+                expect(button.exists('CustomRender')).toBe(false);
+
+                button.setProps({ contentRender });
+                expect(button.exists('CustomRender')).toBe(true);
+                
+                button.setProps({ contentRender: undefined });
+                expect(button.exists('CustomRender')).toBe(false);
```
MaximKudriavtsev: It is not clear what does 'CustomRender' mean

js/renovation/button.tsx
MaximKudriavtsev: I think it is a bad name - CustomRender

testing/jest/button.tests.tsx
```diff
@@ -210,15 +210,46 @@ describe('Button', () => {
...
+            });
+
+            it('should change properties in runtime', () => {
+                const button = mount(<Button text='My button' contentRender={contentRender} />);
+                let buttonContent = button.find('CustomRender');
```
MaximKudriavtsev: Similar with `CustomRender`

js/renovation/button.tsx
LazyLahtak: > I think it is a bad name - CustomRender

I think a better name will be the `Template`. But the generator has a bug with translation all `template` words into `render` in the declaration.

LazyLahtak: > Why have we created a model object?

It can be passed from 0 to 4 properties into the template.
We pass them using spread:
```jsx
// params = { model, index, onRendered, parent }
const data = { container: getPublicElement($container), ...params };
```

LazyLahtak:
Ready to review

pavelgruba (approved):

churkin (approved):

MaximKudriavtsev (approved):

PR #11906, opened 2020-02-06
Drone CI - Fix setting timezone

AntonSermyazhko:


AntonSermyazhko:
Setting the _TZ_ environment variable to the desired time zone doesn't work altough the _date_ command shows the correct date.
Some applications reads this info from the _/etc/localtime_ (https://unix.stackexchange.com/a/452566).
This file is a symbol link to the required timezone file. Expected [solution](https://github.com/DevExpress/DevExtreme/pull/11906/files#diff-4851192b271a019436fa8db6807c9daeR51-R52).

AlekseyMartynov (approved):

docker-ci.sh
```diff
@@ -48,8 +48,8 @@ function run_test {
     [ -n "$PERF" ] && url="$url&include=DevExpress.performance&workerInWindow=true"
 
     if [ -n "$TZ" ]; then
-        echo "Time-Zone: $TZ"
-        date
+        ln -sf "/usr/share/zoneinfo/$TZ" /etc/localtime
```
San4es, line 51: Can we use `timedatectl set-timezone [TIMEZONE]` instead?

docker-ci.sh
```diff
@@ -48,8 +48,8 @@ function run_test {
     [ -n "$PERF" ] && url="$url&include=DevExpress.performance&workerInWindow=true"
 
     if [ -n "$TZ" ]; then
-        echo "Time-Zone: $TZ"
```
San4es, line 51: How to make sure that that the desired timezone is applied correctly now?

docker-ci.sh
AntonSermyazhko: Our ubuntu image has no `timedatectl`

docker-ci.sh
AntonSermyazhko: [`dpkg-reconfigure --frontend noninteractive tzdata`](https://github.com/DevExpress/DevExtreme/pull/11906/files/7d384da68582c91d06596c8fdb2d5d1aa71df472#diff-4851192b271a019436fa8db6807c9daeR52) shows UTC and applied date

docker-ci.sh
AlekseyMartynov: `timedatectl` is not part of our build image
1. https://askubuntu.com/a/1081158
2. https://github.com/DevExpress/DevExtreme/blob/e3c0b5bd4e7a0c9379ba98469ac8d5f9eac5f487/build/docker-image/Dockerfile#L7-L8

AlekseyMartynov: Я первый!

andrewmakarov (approved):

docker-ci.sh
AntonSermyazhko: Too big solution for a little TZ :)

San4es: >timedatectl is not part of our build image

ok

I have met many occurrences of this command in this form:
`ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone`

Do we need:
- `-n` for `ln`?
- to create `/etc/timezone`?

docker-ci.sh
San4es: Got it, tnx!
image

docker-ci.sh
AntonSermyazhko: > -n for ln?

No, `/etc/timezone` will be created without this flag.
I tested this command and looks like `dpkg-reconfigure` is not needed in this case.

PR #11927, opened 2020-02-07
DropDownEditor: Move 'contentTemplate' option to DropDownBox (T860163)

San4es:
https://trello.com/c/SHRBaaGo/

### DIFF
--- a/js/ui/drop_down_editor/ui.drop_down_editor.js
+++ b/js/ui/drop_down_editor/ui.drop_down_editor.js
@@ -183,7 +183,6 @@ const DropDownEditor = TextBox.inherit({
             dropDownButtonTemplate: 'dropDownButton',
 
             fieldTemplate: null,
-            contentTemplate: null,
 
             /**
              * @name dxDropDownEditorOptions.openOnFieldClick
@@ -556,6 +555,8 @@ const DropDownEditor = TextBox.inherit({
         this._renderPopupContent();
     },
 
+    _renderPopupContent: noop,
+
     _renderPopup: function() {
         this._popup = this._createComponent(this._$popup, Popup, extend(this._popupConfig(), this._getInnerOptionsCache('dropDownOptions')));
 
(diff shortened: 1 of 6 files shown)


js/ui/drop_down_editor/ui.drop_down_editor.js
```diff
@@ -603,7 +604,8 @@ const DropDownEditor = TextBox.inherit({
             showCloseButton: false,
             toolbarItems: this._getPopupToolbarItems(),
             onPositioned: this._popupPositionedHandler.bind(this),
-            fullScreen: false
+            fullScreen: false,
+            contentTemplate: null
```
DokaRus, line 608: Was it made to avoid problems with the named template (`content`)? Should we add a test for ddEditor with nested dxTemplate (e.g. for KO)?

NickMitrokhin: @DokaRus, what should we check using this test? That named templates are not working in ddEditors except ddBox? Since this option is not documented for ddEditors, maybe the test will be redundant.

DokaRus: >what should we check using this test?

We should check that named template (`content` in this case) does not replace `contentTemplate` of the nested Popup widget.
@San4es has resolved this confusing behavior at [R608](https://github.com/DevExpress/DevExtreme/pull/11927/files#diff-daa74ef81b21e0ce80dd1b08baaeeb9cR608) but without test this line can be reverted after some kind of refactoring or renovating.

San4es: >Should we add a test for ddEditor with nested dxTemplate

Done

DokaRus (approved):
LGTM

alexander-kotov-dx (approved):

NickMitrokhin:
LGTM

NickMitrokhin (approved):

PR #11928, opened 2020-02-07
Draggable: fix clone had incorrect direction if rtlEnabled (T859557 refix)

jtoming830:


Alyar666 (approved):

js/ui/draggable.js
```diff
@@ -475,7 +475,9 @@ const Draggable = DOMComponent.inherit({
             }).appendTo(result);
         }
 
-        return result.toggleClass(this._addWidgetPrefix(CLONE_CLASS), result.get(0) !== $element.get(0));
+        return result
+            .toggleClass(this._addWidgetPrefix(CLONE_CLASS), result.get(0) !== $element.get(0))
+            .toggleClass('dx-rtl', this.option('rtlEnabled'));
```
vconst, line 480: This class should not be added if clone is false.

vconst (changes requested):

vconst (approved):

PR #11983, opened 2020-02-12
DataGrid - Wrong freeSpace row height calculation in FF 70.0+

igormalzew:


js/ui/grid_core/ui.grid_core.rows.js
```diff
@@ -830,8 +830,7 @@ module.exports = {
                                     const rowsHeight = that._getRowsHeight(contentElement.children().first());
                                     const $tableElement = $table || that.getTableElements();
                                     const borderTopWidth = Math.ceil(parseFloat($tableElement.css('borderTopWidth')));
-                                    const heightCorrection = browser.webkit && that._getDevicePixelRatio() >= 2 ? 1 : 0; // T606935
-                                    const resultHeight = elementHeightWithoutScrollbar - rowsHeight - borderTopWidth - heightCorrection;
+                                    const resultHeight = elementHeightWithoutScrollbar - rowsHeight - borderTopWidth - that._getHeightCorrection.call(that);
```
vconst: It is not necessary to use call method here. Change `that._getHeightCorrection.call(that)` to `that._getHeightCorrection()`.
Also, assign call result to heightCorrection constant as previously.

vconst (changes requested):

js/ui/grid_core/ui.grid_core.rows.js
igormalzew: Done

vconst (dismissed):

AntonSermyazhko (approved):

PR #11986, opened 2020-02-12
TS: Skip default library compilation checks

San4es:
We get the following error:

```
TypeScript error: 
.../node_modules/@types/node/globals.d.ts(318,65): error TS2339: Property 'toPrimitive' does not exist on type 'SymbolConstructor'.
```

after this change in @types/node@10.17.15:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42111/files#diff-ad3552bc3f967859ad249ffc9b9cf5c1R358

This PR enables `skipLibCheck` option for TS compiler to avoid default library compilation in our Gulp tasks (see https://www.typescriptlang.org/docs/handbook/compiler-options.html).


nightskylark (approved):
👍

AlekseyMartynov (approved):

PR #12053, opened 2020-02-17
DataGrid - Async validation (row/cell/batch)

NickMitrokhin:


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -2636,8 +2666,11 @@ module.exports = {
                     }
 
                     const modifiedValues = parameters.row && (parameters.row.isNewRow ? parameters.row.values : parameters.row.modifiedValues);
-
-                    if(modifiedValues && modifiedValues[columnIndex] !== undefined && parameters.column && !isCommandCell && parameters.column.setCellValue) {
+                    const isCellInvalidInNewRow = parameters.row && parameters.row.isNewRow && editingController.isInvalidCell({
+                        rowKey: parameters.key,
+                        columnIndex: parameters.column.index
+                    });
+                    if(((modifiedValues && modifiedValues[columnIndex] !== undefined) || isCellInvalidInNewRow) && parameters.column && !isCommandCell && parameters.column.setCellValue) {
```
vconst: Create virtual method _isCellModified instead of isValidCell and move logic with calling isValidCell to it. Editing should never know about validation.

js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -106,51 +117,41 @@ const ValidatingController = modules.Controller.inherit((function() {
...
 
             if(this._isValidationInProgress) {
                 return deferred.resolve(false).promise();
             }
 
             this._isValidationInProgress = true;
+            let fullValidating;
             if(isFull) {
+                fullValidating = new Deferred();
+                editingController.addDeferred(fullValidating);
```
vconst: Why us the fullValidating variable needed? May be call

`editingController.addDeferred(deferred);`

The deferred variable is resolved after full validation.

js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -177,12 +179,16 @@ const ValidatingController = modules.Controller.inherit((function() {
             const editMode = this._editingController.getEditMode();
 
             if(FORM_BASED_MODES.indexOf(editMode) === -1) {
+                if(editData.type === EDIT_DATA_INSERT_TYPE && (!isDefined(editData.data) || isEmptyObject(editData.data))) {
```
vconst: `(!isDefined(editData.data) || isEmptyObject(editData.data))`
Why is such strange condition needed?

js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -203,44 +209,112 @@ const ValidatingController = modules.Controller.inherit((function() {
...
             let editIndex;
-            const defaultValidationResult = function(options) {
-                if(options.brokenRule) {
-                    options.brokenRule.columnIndex = column.index;
-                    options.brokenRule.column = column;
+            let visibleColumns;
+            let columnsController;
+            const editingController = this._editingController;
+            const column = parameters.column;
+            const renderCellPendingIndicator = () => {
```
vconst: The createValidator method became big. Move inner functions outside.

js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -592,8 +752,23 @@ module.exports = {
                         callBase.call(this, $cell);
                     }
                 },
+
                 getEditDataByKey: function(key) {
                     return this._editData[getIndexByKey(key, this._editData)];
+                },
+
+                getEditData: function() {
```
vconst: I can't find any usage of this method.

js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -718,7 +895,7 @@ module.exports = {
                             .addClass(INVALID_MESSAGE_CLASS)
                             .addClass(INVALID_MESSAGE_ALWAYS_CLASS)
                             .addClass(this.addWidgetPrefix(WIDGET_INVALID_MESSAGE_CLASS))
-                            .text(message)
+                            .html(message)
```
vconst: Why? BC? Security?

js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -819,46 +996,63 @@ module.exports = {
                     },
 
                     focus: function($element, hideBorder) {
```
vconst, line 1049: Method became big. Separate it.

js/ui/widget/ui.errors.js
```diff
@@ -233,11 +233,6 @@ module.exports = errorUtils(errors.ERROR_MESSAGES, {
     */
     E1056: 'The {0} widget does not support buttons of the "{1}" type',
 
-    /**
-    * @name ErrorsUIWidgets.E1057
-    */
-    E1057: 'The {0} widget does not support asynchronous validation rules in the "{1}" edit mode',
```
vconst, line 239: After removing the new added error will have the same error code and the customer with 19.2 version will see the wrong error.

styles/widgets/generic/gridBase.generic.less
```diff
@@ -865,6 +865,30 @@
...
+        td {
+            &.dx-validation-pending {
+                position: relative;
+                padding: 0;
+
+                .dx-highlight-outline {
+                    padding-right: @GENERIC_TEXTEDITOR_INVALID_BADGE_SIZE;
+
+                    .dx-rtl&,
+                    .dx-rtl & {
```
vconst, line 880: Ask Babich about this style, because he is removing such constructions to support SSAS

vconst (changes requested):

styles/widgets/generic/gridBase.generic.less
NickMitrokhin: These constructions will be converted automatically. There should not be problems.

js/ui/grid_core/ui.grid_core.validating.js
NickMitrokhin: It' is not used anymore. I forgot to delete it.

js/ui/grid_core/ui.grid_core.validating.js
NickMitrokhin: It makes sense, deferred should be sufficient.

js/ui/grid_core/ui.grid_core.validating.js
NickMitrokhin: isDefined is redundant here. So, the idea is to prevent the hidden validation from executing for a new unmodified row. I'll move the following condition
```js
isEmptyObject(editData.data)
```
to a separate method and call it isRowDataModified. So, it'll be:
```js
isRowDataModified(editData) {
    return !isEmptyObject(editData.data);
},
...
if(editData.type === EDIT_DATA_INSERT_TYPE && !this.isRowDataModified(editData)) {
    editData.isValid = true;
    return;
}
```

js/ui/grid_core/ui.grid_core.editing.js
NickMitrokhin: OK

js/ui/grid_core/ui.grid_core.validating.js
NickMitrokhin: OK

js/ui/grid_core/ui.grid_core.validating.js
NickMitrokhin: OK

js/ui/widget/ui.errors.js
NickMitrokhin: I'll comment it out. So, if we generate a new ID in the future, we will skip this particular index and will delete the comment.

js/ui/grid_core/ui.grid_core.validating.js
NickMitrokhin: It's is required for displaying several error messages in the tooltip. There should not be problems with security, because I'll encode messages.

vconst (approved):

Alyar666 (approved):

PR #12060, opened 2020-02-18
Sass generator: pivotGrid, contextMenu, multiView, tabPanel, fileUploader

babich-a:


styles/widgets/material/pivotGrid.material.less
```diff
@@ -131,6 +131,12 @@
             min-width: @PIVOTGRID_EMPTY_COLUMN_WIDTH;
         }
     }
+
+    .dx-ie {
```
San4es, line 135: It looks like code duplication :)

babich-a: Unfortunately no. The mixin that we use [in the base file](https://github.com/DevExpress/DevExtreme/blob/20_1/styles/widgets/base/pivotGrid.less#L125) is defined in the [material module](https://github.com/DevExpress/DevExtreme/blob/20_1/styles/widgets/material/common.material.less#L59) and [in the generic module](https://github.com/DevExpress/DevExtreme/blob/20_1/styles/widgets/generic/common.generic.less#L59). We can not import them in the base file, but in sass we also can not use previously undefined mixin. The other question - why this mixins are identical (maybe not for IOS). I add the refactoring checklist to the working card  https://trello.com/c/QlqOcx7F/1192-less-to-sass-converter-draft

San4es (approved):

PR #12067, opened 2020-02-18
DataGrid: Fix cell was editing after adding row and page change if mode is cell (T861092)

jtoming830:


js/ui/grid_core/ui.grid_core.data_controller.js
```diff
@@ -91,6 +91,13 @@ module.exports = {
                             that.option('paging.' + optionName, value);
                             that._skipProcessingPagingChange = false;
 
+                            const editingController = that.getController('editing');
```
vconst: 1) Move this code to editing module
2) Check virtual scrolling because virtual scrolling uses paging

vconst (changes requested):

vconst (dismissed):

Alyar666 (dismissed):

Alyar666 (approved):

PR #12069, opened 2020-02-19
DataGrid - Row should not focus on scrolling by pointer (T861577)

AntonSermyazhko:


js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -125,9 +126,11 @@ const KeyboardNavigationController = core.ViewController.inherit({
             const clickSelector = `.${ROW_CLASS} > td, .${ROW_CLASS}`;
             const $focusedElement = $(':focus');
             const isFocusedElementCorrect = !$focusedElement.length || $focusedElement.closest($rowsView).length || (browser.msie && $focusedElement.is('body'));
+            const isMobile = devices.current().deviceType !== 'desktop';
+            const pointerEventName = !isMobile ? pointerEvents.down : 'dxclick';
```
vconst: import click event from '../events/click' module

AntonSermyazhko: Added. Also added unit test.

vladkovl (dismissed):

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -721,7 +725,7 @@ const KeyboardNavigationController = core.ViewController.inherit({
     // #endregion Key_Handlers
 
     // #region Click_Handler
-    _pointerDownHandler: function(e) {
+    _clickHandler: function(e) {
```
vconst: It is not real click. So, I think that will be better to rename it to _focusClickHandler or something like that

vconst (changes requested):

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
AntonSermyazhko: This handler is not only for focus, if editing enabled e.g.
Suggest rename it to pointerEventHandler

AntonSermyazhko: Renamed: _clickHandler -> _pointerEventHandler

jtoming830 (approved):

Alyar666 (approved):

vconst (approved):

PR #12087, opened 2020-02-20
Turn rebase off for clean-css (T862834)

babich-a:


themebuilder/modules/clean-css-options.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
+    // eslint-disable-next-line
```
tar-g: I suggest not to disable eslint for the next line, but add the word `rebase` to ignore list in _eslint-plugin-spellcheck_ config

tar-g (approved):

groshenkovamarina (approved):

PR #12152, opened 2020-02-27
Localization: Add unlimitedIntegerDigits parameter for LDML number formats (for dashboards)

vconst:


nightskylark (commented):
Don't you think the LDML standard should be supported by default?

vconst:
@nightskylark 
I think that we should support LDML, but now it will be BC for NumberBox without workaround.
It is necessary more time to fix it out of the box.

nightskylark (approved):

PR #12156, opened 2020-02-27
Add single entry point for requesting external libraries 

DokaRus:
https://trello.com/c/kY4qQ043

Issue
---
We have a lot of "code smells" related to requesting external libraries (e.g. [HtmlEditor](), [Gantt](https://github.com/DevExpress/DevExtreme/blob/20_1/js/ui/gantt/gantt_importer.js), [Diagram](https://github.com/DevExpress/DevExtreme/blob/20_1/js/ui/diagram/diagram.importer.js))

Possible solution
---
We can add a local registry which should provide a way to add and request external libraries. Components should access external components by requesting them through a registry.

Proposed API
---
Bundle approach:
```js
DevExpress.libraryRegistry.add('awesomeLib', window.AwesomeLib);
<...>
const awesomeLib = DevExpress.libraryRegistry.get('awesomeLib');
awesomeLib.doSomething();
```
Modular approach:
```js
import { addLibrary } from 'devextreme/core/library_registry';
import AwesomeLib from 'awesome';

addLibrary('awesomeLib', AwesomeLib);

<...>
import { getLibrary } from 'devextreme/core/library_registry';

const awesomeLib = getLibrary('awesomeLib');
awesomeLib.doSomething();
```

What will change for those who add the DevExtreme scripts via `script` tag
---
Nothing will change. 

What will change for those who use DevExtreme in a modular environment
---
There we have a small BC.
Before they should just import external library:
```js
import 'awesome';
```
For now they should import external library and register it in DevExtreme:
```js
import { addLibrary } from 'devextreme/core/library_registry';
import AwesomeLib from 'awesome';

addLibrary('awesomeLib', AwesomeLib);
```


ts/dx.all.d.ts
```diff
@@ -711,6 +711,8 @@ declare module DevExpress {
         /** @name TransitionExecutor.stop() */
         stop(): void;
     }
+    /** @name addLibrary(libraryName, library) */
+    export function addLibrary(libraryName: string, library: any): void;
```
San4es, line 715: Should we extract the `libraryRegistry` interface?

playground/jquery.html
```diff
@@ -22,18 +22,1375 @@
...
     <script src="https://cdnjs.cloudflare.com/ajax/libs/jszip/3.1.2/jszip.min.js"></script> -->
-
+    <!-- <script type="text/javascript" src="../artifacts/js/dx-gantt.js" charset="utf-8"></script> -->
     <script type="text/javascript" src="../artifacts/js/dx.all.debug.js" charset="utf-8"></script>
+    <script type="text/javascript" src="../artifacts/js/dx-gantt.js" charset="utf-8"></script>
+
 </head>
 <body>
-    <div id="button"></div>
+    <div id="gantt"></div>
```
San4es: Should we revert changes in this file?

vladkovl (approved):

kruglikovstepan (approved):

DokaRus:
We'll solve the problem in another way to avoid breaking changes

PR #12158, opened 2020-02-27
Default option rules

roman-simionov:


pavelgruba (dismissed):

js/core/options/utils.js
```diff
@@ -27,3 +27,7 @@ export const deviceMatch = (device, filter) => isEmptyObject(filter) || findBest
 export const getFieldName = fullName => fullName.substr(fullName.lastIndexOf('.') + 1);
 
 export const getParentName = fullName => fullName.substr(0, fullName.lastIndexOf('.'));
+
+export default function createDefaultRules(options = []) {
+    return options;
+}
```
MaximKudriavtsev: I think we should not combine `old` and `new` styles. Use arrow functions

testing/jest/button.tests.tsx
```diff
@@ -414,4 +440,43 @@ describe('Button', () => {
 
         expect(tree.is('.dx-button')).toBe(true);
     });
+
+    describe('DefaultOptionRules', () => {
+        function getDefaultProps() {
```
MaximKudriavtsev: arrow function

testing/jest/button.tests.tsx
```diff
@@ -414,4 +440,43 @@ describe('Button', () => {
...
+            defaultOptions({
+                device: () => false,
+                options: {},
+            });
+            return Button.defaultProps;
```
MaximKudriavtsev, line 450: It's very bed using global functions and global variables. Can we rewrite it to pure function?

roman-simionov: No, We can't

pavelgruba (approved):

MaximKudriavtsev (approved):

PR #12173, opened 2020-03-02
Fix sass generator

babich-a:
1. The number of small fixes (fade, parent selector, calc, some expressions)
2. Base filterBuilder
3. recurrenceEditor for scheduler


San4es (approved):
LGTM

PR #12174, opened 2020-03-02
Renovation: add onContentReady property to button and widget

LazyLahtak:


js/renovation/widget.tsx
```diff
@@ -126,6 +129,10 @@ export const viewModelFunction = ({
 };
 
 export const viewFunction = (viewModel: any) => {
+    useLayoutEffect(() => {
+        !viewModel.disabled && viewModel.onContentReady?.();
+    }, [viewModel.children]);
```
MaximKudriavtsev: - why do we use `children` as a dependency?
- why do we use `!viewModel.disabled` here?

js/renovation/widget.tsx
```diff
@@ -16,6 +16,7 @@ import { each } from '../core/utils/iterator';
 import { extend } from '../core/utils/extend';
 import { focusable } from '../ui/widget/selectors';
 import { isFakeClickEvent } from '../events/utils';
+import { useLayoutEffect } from 'preact/hooks';
```
pavelgruba: We shouldn't use any react/preact things. Why Effect decorator is not used here?

js/renovation/widget.tsx
```diff
@@ -126,6 +129,10 @@ export const viewModelFunction = ({
 };
 
 export const viewFunction = (viewModel: any) => {
+    useLayoutEffect(() => {
```
pavelgruba: View function should only return JSX markup, any side effects are prohibited

js/renovation/widget.tsx
MaximKudriavtsev: Should we use something like `@layoutEfffect`?

js/renovation/dist/button.j.jsx
```diff
@@ -53,6 +53,10 @@ class Button extends Widget {
             }
         });
 
+        props.onContentReady = this._createActionByOption('onContentReady', {
+            excludeValidators: ['disabled', 'readOnly']
+        });
+
```
MaximKudriavtsev, line 59: Why we are using this code twice?

js/renovation/widget.tsx
pavelgruba: I don't think so, there should be strong reason to use LayoutEffect instead of just Effect

MaximKudriavtsev: Okay

js/renovation/widget.tsx
LazyLahtak: * `children` is content. I've made it to avoid call `onContentReady` on state changes
* `!viewModel.disabled` - you're right, I'll remove this

js/renovation/dist/button.j.jsx
LazyLahtak: The same was implemented for `onClick` property
In fact, I think we can get rid of these properties for widget

MaximKudriavtsev: Keeping this in the `Widget` maybe more convenient for us instead of in the `Button`

LazyLahtak: If I remove this from Button - it doesn't work correctly. `onContentReady` from Widget rewrite `onContentReady` from Button's props.
If I remove this from Widget - it still work.

js/renovation/widget.tsx
```diff
@@ -254,6 +257,13 @@ export default class Widget extends JSXComponent<WidgetInput> {
         return () => dxClick.off(this.widgetRef, { namespace });
     }
 
+    @Effect()
+    contentReadyEffect() {
+        const { onContentReady } = this.props;
+
+        onContentReady?.({});
```
MaximKudriavtsev, line 264: Should we pass here the following [object](https://js.devexpress.com/Documentation/20_1/ApiReference/UI_Widgets/dxButton/Configuration/#onContentReady)?

`component, element, model`?

roman-simionov: No, ContentReady is an action that passes these parameters

roman-simionov: https://github.com/DevExpress/DevExtreme/pull/12174/files#diff-03a401c56907a4c058f12b32b61523fdR56

MaximKudriavtsev (approved):

pavelgruba (approved):

roman-simionov (approved):

PR #12183, opened 2020-03-03
Use preventDefault for Home and End keydowns for the DateBox widget only (T865192)

alexander-kotov-dx:

### DIFF
--- a/js/ui/date_box/ui.date_box.base.js
+++ b/js/ui/date_box/ui.date_box.base.js
@@ -54,18 +54,7 @@ const STRATEGY_CLASSES = {
 const DateBox = DropDownEditor.inherit({
 
     _supportedKeys: function() {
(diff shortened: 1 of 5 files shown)


js/ui/date_box/ui.date_box.base.js
```diff
@@ -54,7 +54,18 @@ const STRATEGY_CLASSES = {
...
+                e.preventDefault();
+                return true;
+            }
+            return false;
+        };
+
+        return extend(this.callBase(), {
+            home: homeEndHandler,
+            end: homeEndHandler
+        }, this._strategy.supportedKeys());
```
DokaRus: Should we prevent `home`/`end` behavior for all strategies by default? Or for Calendar-based only?

DokaRus (approved):
LGTM

San4es (approved):

PR #12254, opened 2020-03-06
DataGrid - keyboard navigation should work correctly in virtual rendering modes

AntonSermyazhko:


igormalzew (dismissed):

NickMitrokhin:
Fixed in #14396

PR #12332, opened 2020-03-17
Update exceljs to the latest version

San4es:


.github/workflows/CI.yml
```diff
@@ -1,6 +1,6 @@
 name: QUnit IE
 
-on: [push]
+on: [push, pull_request]
```
IgnatovDan, line 3: Change doesn't look to be related to exceljs version change.
Do we need it?

San4es: No, but we saw problems with the new ExcelJS in IE. So I closed this PR.

PR #12344, opened 2020-03-18
Diagram - hide unsupported api members + fix the focus() method

Romzay:


San4es:
FYI: https://github.com/DevExpress/DevExtreme/pull/12342

LanaDX (approved):

PR #12346, opened 2020-03-18
Renovation: jest tests refactoring

churkin:


MaximKudriavtsev (approved):

js/renovation/widget.tsx
```diff
@@ -213,7 +217,7 @@ export default class Widget extends JSXComponent<WidgetInput> {
                 e => !e.isDefaultPrevented() && (this._focused = true),
                 e => !e.isDefaultPrevented() && (this._focused = false),
                 {
-                    isFocusable: el => focusable(null, el),
+                    isFocusable: focusable,
```
LazyLahtak, line 220: `focusable` method pass element as second parameter. Did this checked in IE?
In old component we used
```jsx
isFocusable: el => $(el).is(focusableSelector)
```
(that pass into function `index` and `element`, but we use only `element)
```js
focusable: function(index, element) {
  return focusable(element, $(element).attr('tabIndex'));
}
```

js/ui/widget/ui.widget.js
```diff
@@ -419,7 +419,7 @@ const Widget = DOMComponent.inherit({
             e => this._focusInHandler(e),
             e => this._focusOutHandler(e), {
                 namespace: `${this.NAME}Focus`,
-                isFocusable: el => $(el).is(focusableSelector)
+                isFocusable: (index, el) => $(el).is(focusableSelector)
```
LazyLahtak, line 422: This might break in IE too.

LazyLahtak:
Also, we may skip async QUnit tests for new button (template tests).
They are qute unstable.

LazyLahtak (approved):

PR #12352, opened 2020-03-18
Add Angular and React generators 

roman-simionov:
- Add gulp tasks that generate React and Angular components
- Add Angular and React applications in the playground


build/gulp/generator/gulpfile.js
```diff
@@ -0,0 +1,80 @@
...
+                quiet: true,
+                fix: true,
+                useEslintrc: true
+            })
+        ))
+        .pipe(lint.format())
+        .pipe(gulp.dest(DEST));
+});
+
+function addGenerationTask(approach, knownErrors = []) {
```
MaximKudriavtsev: approach -> frameworkName?

build/gulp/generator/ts-configs/angular.tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+    "compilerOptions": {
+        "target": "ES2016",
+        "module": "ESNext",
```
MaximKudriavtsev: We can extract the base part as it did [here](https://github.com/DevExpress/devextreme-reactive/blob/master/packages/dx-react-scheduler/tsconfig.json#L2)

playground/react/App.js
```diff
@@ -0,0 +1,13 @@
...
+import Button from 'devextreme/renovation/button';
+
+const App = () => {
+    return <Button
+        text={'Click Me!'}
+        onClick={() => alert('Clicked')}
+        icon={'download'}
+        iconPosition={'right'}
+    ></Button>;
+};
```
MaximKudriavtsev: It would be more 'reactive' and better for our programmers. Or what about adding the `tslint` file here?

```jsx
const App = () => {
    return (
      <Button
        text="Click Me!"
        onClick={() => alert('Clicked')}
        icon="download"
        iconPosition="right"
        enabled // boolean
      />
  )
};
```

LazyLahtak (approved):

MaximKudriavtsev (approved):

PR #12355, opened 2020-03-18
DateBox: Use 'beforeInput' event for mask on Android devices (T838638)

San4es:

### DIFF
--- a/js/ui/date_box/ui.date_box.mask.js
+++ b/js/ui/date_box/ui.date_box.mask.js
@@ -150,7 +150,28 @@ const DateBoxMask = DateBoxBase.inherit({
         return result;
     },
 
-    _maskInputHandler(e) {
+    _maskBeforeInputHandler(e) {
+        this._maskInputHandler = null;
+
+        const { inputType } = e.originalEvent;
+
+        if(inputType === 'insertCompositionText') {
+            this._maskInputHandler = () => {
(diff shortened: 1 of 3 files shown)


DokaRus (approved):

alexander-kotov-dx (approved):

js/ui/date_box/ui.date_box.mask.js
```diff
@@ -124,30 +125,82 @@ const DateBoxMask = DateBoxBase.inherit({
...
+            this._maskInputHandler = null;
+        }
+    },
+
+    _processInputKey(key) {
+        if(this._isAllSelected()) {
+            this._activePartIndex = 0;
+        }
+        this._setNewDateIfEmpty();
+        if(isNaN(parseInt(key))) {
```
ksercs, line 199: It will be false on "123someLetters". Is it ok? Should we use `isNaN(key)` here?

San4es: The `key` argument is always one char - we have `!this._isSingleCharKey(e)` conditions before the `_processInputKey` call:
https://github.com/DevExpress/DevExtreme/pull/12355/files#diff-e23fb8153c04867cc50fe7f201618494R142
https://github.com/DevExpress/DevExtreme/pull/12355/files#diff-e23fb8153c04867cc50fe7f201618494R175

ksercs (approved):

PR #12422, opened 2020-03-24
Show deprecated warnings for SlideOut, SlideOutView and NavBar widgets to console when they are initialized

jsdmitry:
Deprecate the `dxSlideOut`, `dxSlideOutView `and `dxNavBar`
Warnings:
1. W0000 - 'dxSlideOut' is deprecated in 20.1. Use the 'dxDrawer' widget instead. See:
http://js.devexpress.com/error/20_1/W0000
2. W0000 - 'dxSlideOutView' is deprecated in 20.1. Use the 'dxDrawer' widget instead. See:
http://js.devexpress.com/error/20_1/W0000
3. W0000 - 'dxNavBar' is deprecated in 20.1. Use the 'dxTabs' widget instead. See:
http://js.devexpress.com/error/20_1/W0000

**NOTE:** The `dxSlideOut` thrown two warnings to the console: own and from the `dxSlideOutView`

### DIFF
--- a/js/core/component.js
+++ b/js/core/component.js
@@ -95,7 +95,7 @@ const Component = Class.inherit({
             this._options.onChanging(
                 (name, previousValue, value) => this._initialized && this._optionChanging(name, previousValue, value));
             this._options.onDeprecated(
-                (option, info) => this._logDeprecatedWarning(option, info));
+                (optionName, info) => this._logDeprecatedWarning('W0001', { ...info, optionName }));
             this._options.onChanged(
                 (name, value, previousValue) => this._notifyOptionChanged(name, value, previousValue));
             this._options.onStartChange(() => this.beginUpdate());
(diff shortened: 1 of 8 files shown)


js/core/component.js
```diff
@@ -125,9 +125,15 @@ const Component = Class.inherit({
         });
     },
 
-    _logDeprecatedWarning(option, info) {
-        const message = info.message || (`Use the '${info.alias}' option instead`);
-        errors.log('W0001', this.NAME, option, info.since, message);
+    _logDeprecatedWarning(error, options) {
```
nightskylark: Why we need common method to warn about deprecated widgets and deprecated options? As I can see this method does not contain any common logic. And it uses string comparison to find out what to do. Not cool.

js/ui/nav_bar.js
```diff
@@ -9,6 +9,11 @@ const NAVBAR_ITEM_CLASS = 'dx-nav-item';
 const NAVBAR_ITEM_CONTENT_CLASS = 'dx-nav-item-content';
 
 const NavBar = Tabs.inherit({
+    _init: function() {
+        this.callBase();
+        this._logDeprecatedWarning('W0000', { since: '20.1', alias: 'dxTabs' });
```
nightskylark: Maybe it's better to move warning in a constructor? Just in case if the code will fall before widgets initialization.

EugeniyKiyashko (dismissed):

js/core/component.js
jsdmitry: 1. The `_logDeprecatedWarning` doesn't contain the option word so it has a common name for warnings.
I can agree to create a new method for a deprecated component and rename the `_logDeprecatedWarning` to `_logDeprecatedOptionWarning` as an example.
2. Comparison of a string?  The error argument is an id of warning

js/ui/nav_bar.js
jsdmitry: I will move this code to the constructors of widgets.

js/core/component.js
nightskylark: 1. Let's do it
2. I know. But we do not need redundant string comparison in case we have two methods.

nightskylark (approved):

EugeniyKiyashko (approved):

PR #12432, opened 2020-03-24
DataGrid: Fix highlight outline after editing cell when cellRender is used (T858174)

Alyar666:


vconst (changes requested):
In the adaptivity module call _cellPrepared in the done of renderTemplate when cellTemplate is defined.

PR #12469, opened 2020-03-26
Lookup: Remove dxPopoverOptions type form dropDownOptions option

San4es:
To avoid the issue in Vue components:
image


San4es:
Fixed in metadata tools.

PR #12487, opened 2020-03-27
Force TagBox tags rerender on repaint after dataSource change (T873372)

alexander-kotov-dx:

### DIFF
--- a/js/ui/tag_box.js
+++ b/js/ui/tag_box.js
@@ -736,7 +736,7 @@ const TagBox = SelectBox.inherit({
         const selectedItemsAlreadyLoaded = filteredItems.length === values.length;
         const d = new Deferred();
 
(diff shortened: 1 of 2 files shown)


DokaRus (dismissed):

js/ui/tag_box.js
```diff
@@ -736,7 +736,7 @@ const TagBox = SelectBox.inherit({
         const selectedItemsAlreadyLoaded = filteredItems.length === values.length;
         const d = new Deferred();
 
-        if(selectedItemsAlreadyLoaded) {
+        if(!this._isCacheDisabled && selectedItemsAlreadyLoaded) {
```
ksercs: Why `_isCacheDisabled`? To my mind, it's better to name it `_isDataSourceChanged` or smth like this

ksercs (approved):

PR #12495, opened 2020-03-27
DateBox should not validate the intermediate value when the "dateSerializationFormat" option is defined (T874472)

DokaRus:


San4es (approved):
LGTM

alexander-kotov-dx (approved):

PR #12543, opened 2020-04-02
[WIP] Renovation: allow render button in different containers

LazyLahtak:


js/renovation/widget.tsx
```diff
@@ -119,6 +119,7 @@ export class WidgetInput {
     @Event() onKeyboardHandled?: (args: any) => any | undefined;
     @Event() onKeyPress?: (e: any, options: any) => any;
     @Event() onVisibilityChange?: (args: boolean) => undefined;
+    @OneWay() rootNode?: any = 'div';
```
roman-simionov, line 122: type should be a string

LazyLahtak: If I set type `string`, we get an error:
`JSX element type 'viewModel.props.rootNode' does not have any construct or call signatures`

LazyLahtak: If I pass render function instead string:
```jsx
rootNode === (props) => Preact.h(nodeName, props)
```
then, we can't use JSX (generator doesn't recognise correctrly)

roman-simionov: ok

PR #12627, opened 2020-04-08
DataGrid - Fixes SelectAll state when all items in the header filter are deselected (T875471)

NickMitrokhin:
1. Added the required condition to reset the **filterValues** array to make the **updateSelectAllState** method not to set the SelectAll checkbox value to undefined. 
2. Added a test.

### DIFF
--- a/js/ui/grid_core/ui.grid_core.header_filter_core.js
+++ b/js/ui/grid_core/ui.grid_core.header_filter_core.js
@@ -287,14 +287,21 @@ exports.HeaderFilterView = modules.View.inherit({
                         const items = e.component.option('items');
                         const selectedItems = e.component.option('selectedItems');
 
+                        const filterValuesOnAnotherPage = !!items.length && !!options.filterValues && options.filterValues.some(filterValue => {
+                            return gridCoreUtils.getIndexByKey(filterValue, items, 'value') < 0;
+                        });
+
                         if(!e.component._selectedItemsUpdating && !e.component.option('searchValue') && !options.isFilterBuilder) {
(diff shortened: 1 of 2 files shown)


js/ui/grid_core/ui.grid_core.header_filter_core.js
```diff
@@ -288,7 +288,9 @@ exports.HeaderFilterView = modules.View.inherit({
                         const selectedItems = e.component.option('selectedItems');
 
                         if(!e.component._selectedItemsUpdating && !e.component.option('searchValue') && !options.isFilterBuilder) {
-                            if(selectedItems.length === 0 && items.length && (!options.filterValues || options.filterValues.length <= 1)) {
+                            const filterValues = options.filterValues || [];
+                            const isExclude = options.filterType === 'exclude';
+                            if(selectedItems.length === 0 && items.length && (filterValues.length <= 1 || isExclude && filterValues.length === items.length - 1)) {
```
jtoming830, line 293: Maybe move this huge condition to the const with charactonym? 
Or at least this part
`(filterValues.length <= 1 || isExclude && filterValues.length === items.length - 1)`

vconst (approved):

js/ui/grid_core/ui.grid_core.header_filter_core.js
NickMitrokhin: But how to name this constant? needChangeFilterType? I think that the name will not be so obvious.

jtoming830 (approved):

PR #12676, opened 2020-04-13
Renovation: add checking of parent node on render

LazyLahtak:


pavelgruba (dismissed):

roman-simionov:
Why is there no test on this case?

pavelgruba (approved):

PR #12714, opened 2020-04-15
dxToolbar - Fix internal menu styles

groshenkovamarina:


styles/widgets/generic/toolbar.generic.less
```diff
@@ -180,3 +189,21 @@
...
 }
+
+.dx-toolbar-button {
+    .dx-menu {
+        .dx-menu-item {
+            border: 1px solid transparent;
+            border-radius: 4px;
+
+            .dx-menu-item-content {
+                padding: @GENERIC_TOOLBAR_MENU_TOP_PADDING @GENERIC_MENU_ITEM_PADDING_BASE @GENERIC_TOOLBAR_MENU_BOTTOM_PADDING;
```
babich-a: if horizontal menu padding is not changed anywhere before I suggest using` padding-top` and `padding-bottom`. the using of @GENERIC_MENU_ITEM_PADDING_BASE constant here is not good for future module system.

babich-a (dismissed):

styles/widgets/generic/toolbar.generic.less
```diff
@@ -180,3 +189,22 @@
 .dx-toolbar .dx-tab {
     padding: 4px;
 }
+
+.dx-toolbar-button {
+    .dx-menu {
+        .dx-menu-item {
+            border: 1px solid transparent;
+            border-radius: 4px;
```
BingoRUS: Should we use @base-border-radius here?

groshenkovamarina: A menu item in a toolbar should look like a button so I use button's parameters here

BingoRUS (approved):

styles/widgets/generic/toolbar.generic.less
BingoRUS: @button-border-radius is fine too.

PR #12776, opened 2020-04-20
dxLookup - Popup position is not correctly defined in the Material theme

groshenkovamarina:


js/ui/lookup.js
```diff
@@ -489,36 +489,65 @@ const Lookup = DropDownList.inherit({
...
+        const listItemsCount = this._list.option('items').length;
+        const lookupOffset = $(this._list.element()).offset().top;
 
-        const selectedIndex = this._list.option('selectedIndex');
-        const flipped = this._popup._$wrapper.hasClass(LOOKUP_POPOVER_FLIP_VERTICAL_CLASS);
-        if(selectedIndex === -1 || flipped) return;
+        let offset = 0;
 
-        const selectedListItem = $(this._list.element()).find('.' + LIST_ITEM_SELECTED_CLASS);
+        if(1 < selectedIndex && selectedIndex < (listItemsCount - 2)) {
```
babich-a: I think we need to add some info about what this condition means. It is unclear now. Is it two 'border' items at the start and the end of the list?

groshenkovamarina: Yes, it is two items at the start and the end of the list, which cannot be centered. What do you suggest? Add comment?

babich-a: I prefer the self-documenting code. We can extract this to the function for example

groshenkovamarina: We can extract internal code from 'If' condition, but it will not more clear

babich-a: it depends on function name :)

groshenkovamarina: What do you think about 'getCenteringOffset' name?

babich-a: needCentering(){ return selectedIndex > 1 && selectedIndex < (listItemsCount - 2);}
smth like this?

groshenkovamarina: Updated

js/ui/lookup.js
```diff
@@ -489,47 +489,99 @@ const Lookup = DropDownList.inherit({
...
+
+        if(selectedIndex !== -1) {
+            if(this._isCenteringEnabled(selectedIndex, listItemsCount)) {
+                this._scrollToSelectedItem();
+
+                const scrollOffsetTop = (popupHeight - selectedListItem.height()) / 2 - this._getDifferenceOffsets(selectedListItem);
+
+                this._list.scrollTo(this._list.scrollTop() + MATERIAL_LOOKUP_LIST_PADDING / 2 - scrollOffsetTop);
+
+                offsetTop = differenceOfHeights + this._getDifferenceOffsets(selectedListItem);
```
babich-a, line 525: It seems that this line is need in all condition branches (522, 527, 532, 536)

groshenkovamarina: Yes, but  this._getDifferenceOffsets(selectedListItem) need calculate after scroll each time

babich-a: And? Can we calculate it once after all scrollings? Duplicate code 4 times is not good solution

groshenkovamarina: As you can see in the first case this._getDifferenceOffsets(selectedListItem) calls twice. The second time is based on the previous value of this._getDifferenceOffsets(selectedListItem)

groshenkovamarina: I can extract this line from the other two cases

babich-a (dismissed):

babich-a (dismissed):

PR #12846, opened 2020-04-23
Fix the min-width CSS style for the ResponsiveBox item in the row direction (T878630)

jsdmitry:

### DIFF
--- a/testing/helpers/wrappers/tabsWrappers.js
+++ b/testing/helpers/wrappers/tabsWrappers.js
@@ -1,0 +1,78 @@
+import $ from 'jquery';
+import { extend } from 'core/utils/extend';
+
+export class TestTabsWrapper {
+    constructor($element, options) {
+        const { itemsCount } = options;
+        const tabsOptions = extend({
+            items: this._generateItems(itemsCount),
+        }, options);
+
+        this._$tabs = $($element).dxTabs(tabsOptions);
+        this._tabs = this._$tabs.dxTabs('instance');
+    }
+
+    _generateItems(count = 5) {
+        return [...new Array(count)].map((value, index) => ({ text: `item ${index}` }));
+    }
+
+    _checkTabsContent($tabs) {
+        const items = this._tabs.option('items');
+        $tabs.toArray().forEach((tabElement, index) =>
+            QUnit.assert.equal($(tabElement).text(), items[index].text, `text of the ${index} tab`)
+        );
+    }
+
+    set width(value) {
+        this._tabs.option('width', value);
+    }
+
+    setItemsByCount(count) {
+        this._tabs.option('items', this._generateItems(count));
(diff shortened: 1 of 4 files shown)


novsstation (dismissed):

novsstation (dismissed):

testing/helpers/wrappers/tabsWrappers.js
```diff
@@ -0,0 +1,78 @@
...
+    set width(value) {
+        this._tabs.option('width', value);
+    }
+
+    setItemsByCount(count) {
+        this._tabs.option('items', this._generateItems(count));
+    }
+
+    checkTabsWithoutScrollable() {
+        const $tabs = this._$tabs.find('> .dx-tabs-wrapper > .dx-tab');
```
IgnatovDan, line 35: I would prefer this._$tabs.find('.dx-tab') instead of '> .dx-tabs-wrapper > .dx-tab' (full path) if it is possible.
And, the same in the next function: this._$tabs.find('> .dx-tabs-scrollable .dx-scrollable-content > .dx-tabs-wrapper > .dx-tab')

testing/tests/DevExpress.ui.widgets/responsiveBox.tests.js
```diff
@@ -301,6 +301,48 @@ QUnit.module('layouting', moduleConfig, () => {
...
+                { location: { row: 0, col: 1 }, html: '<div class=\'second\'></div>' }
+            ],
+            _layoutStrategy: 'flex'
+        });
+
+        const $first = $responsiveBox.find('.first').closest('.dx-box-item');
+        const $second = $responsiveBox.find('.second').closest('.dx-box-item');
+        const getStyleValue = ($element, styleName) => $element.get(0).style[styleName];
+
+        assert.equal(getStyleValue($first, 'minWidth'), '0px', 'minWidth of the first item');
```
IgnatovDan: I would prefer to check here the resulting coordinates using the getBoundingClientRect with a fixed size of the parent element.
The 'minSize: auto/miSize: 0' are source settings for 'some layouting code', they can lead to different results in IE/Chrome/Edge/ios or we can misunderstand these values as 'incorrect' later.
Imho for T878630 scenario getBoundingClientRect allows to easily create a stable and fast 'screenshot' test (usually 'width/height < 100px' requires '< 1px' rounding)
Something like [FormLayoutTestWrapper.checkElementPosition](https://github.com/DevExpress/DevExtreme/blob/21c5577ddc46de78f287cd6abfa8467b6a28a61e/testing/helpers/FormLayoutTestWrapper.js#L29)

novsstation (approved):

IgnatovDan (approved):

EugeniyKiyashko (approved):

PR #12871, opened 2020-04-27
TextEditor - Prevent custom buttons overlapping in ie11 (T879885)

alexander-kotov-dx:

### DIFF
--- a/js/ui/text_box/ui.text_editor.base.js
+++ b/js/ui/text_box/ui.text_editor.base.js
@@ -211,7 +211,6 @@ const TextEditorBase = Editor.inherit({
         this._renderStylingMode();
         this._renderInputType();
         this._renderPlaceholder();
-        this._compressInputContainer();
 
         this._renderProps();
 
@@ -228,6 +227,7 @@ const TextEditorBase = Editor.inherit({
         this._renderEnterKeyAction();
         this._renderEmptinessEvent();
         this.callBase();
+        this._compressInputContainer();
     },
 
     _renderInput: function() {
--- a/js/ui/text_box/ui.text_editor.base.js
+++ b/js/ui/text_box/ui.text_editor.base.js
@@ -33,7 +33,7 @@ const ALLOWED_STYLE_CLASSES = [
     TEXTEDITOR_STYLING_MODE_PREFIX + 'filled',
     TEXTEDITOR_STYLING_MODE_PREFIX + 'underlined'
 ];
-const TEXTEDITOR_COMPRESSED_CLASS = 'dx-texteditor-compressed';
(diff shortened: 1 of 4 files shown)


js/ui/text_box/ui.text_editor.base.js
```diff
@@ -331,6 +334,14 @@ const TextEditorBase = Editor.inherit({
         });
     },
 
+    _compressInputContainer: function() {
+        const $element = this.$element();
+        const isIE11 = browser.msie && browser.version <= 11;
+        if(isIE11 && $element.css('display') === 'block') {
+            $element.addClass(TEXTEDITOR_COMPRESSED_CLASS);
```
San4es: $element.css('flexBasis', 0)?

DokaRus: It's better to avoid the use of inline styles - they are more difficult to override if necessary. The use of classes is also more convenient in terms of testing

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -331,6 +334,14 @@ const TextEditorBase = Editor.inherit({
         });
     },
 
+    _compressInputContainer: function() {
+        const $element = this.$element();
+        const isIE11 = browser.msie && browser.version <= 11;
+        if(isIE11 && $element.css('display') === 'block') {
```
DokaRus, line 340: Is it works fine with the SSR?
The method is used in the `initMarkup` function, so the test, which checks this behavior, should be in "_markup" file and pass in SSR environment.
Or it should be called after `_initMarkup`.

js/ui/text_box/ui.text_editor.base.js
San4es: Make sense. Maybe we can come up with a better name for the CSS class.

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -32,6 +33,7 @@ const ALLOWED_STYLE_CLASSES = [
     TEXTEDITOR_STYLING_MODE_PREFIX + 'filled',
     TEXTEDITOR_STYLING_MODE_PREFIX + 'underlined'
 ];
+const TEXTEDITOR_COMPRESSED_CLASS = 'dx-texteditor-compressed';
```
San4es: What does the name of this class mean?

alexander-kotov-dx: It means we should use flex-basis: 0; style for texteditor-input-container to minimize the input container width

San4es: compress -> collapse?

DokaRus (dismissed):

San4es (approved):
LGTM

PR #12922, opened 2020-05-06
DataGrid - Fixes focusing command cells on Tab (T884646)

NickMitrokhin:
1. Fixed the issue by adding the tabindex attribute to command cells.
2. Added tests.

### DIFF
--- a/testing/functional/model/dataGrid.ts
+++ b/testing/functional/model/dataGrid.ts
@@ -179,7 +179,7 @@ class CommandCell extends DxElement {
 
     constructor(dataRow: Selector, index: number) {
         const childrenSelector = `td:nth-child(${++index})`;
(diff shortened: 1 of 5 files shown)


testing/functional/model/dataGrid.ts
```diff
@@ -11,6 +11,7 @@ const CLASS = {
     groupRow: 'dx-group-row',
     commandEdit: 'dx-command-edit',
     commandExpand: 'dx-command-expand',
+    selectCommand: 'dx-command-select',
```
vconst: Rename selectCommand to commandSelect like commandEdit/commandExpand

vconst (changes requested):

testing/functional/model/dataGrid.ts
NickMitrokhin: Done.

vconst (approved):

jtoming830 (approved):

PR #12924, opened 2020-05-06
Change export of ODataContext (T886254)

LazyLahtak:
In future we should replace all `require` by `import` to use `export default` for the ODataStore and the ODataContext

### DIFF
--- a/js/data/odata/store.js
+++ b/js/data/odata/store.js
@@ -203,5 +203,6 @@ const ODataStore = Store.inherit({
 
 }, 'odata');
 
-// TODO: replace by "export default" after "odata/context" refactor
+// TODO: replace by "export default" after "js/bundles/modules/data.odata" refactor
+//       (should replace all "require" by "import")
 module.exports = ODataStore;
--- a/js/data/odata/store.js
+++ b/js/data/odata/store.js
@@ -203,6 +203,4 @@ const ODataStore = Store.inherit({
 
 }, 'odata');
 
-// TODO: replace by "export default" after "js/bundles/modules/data.odata" refactor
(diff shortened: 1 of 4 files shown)


MaximKudriavtsev (dismissed):

js/data/odata/store.js
```diff
@@ -203,5 +203,6 @@ const ODataStore = Store.inherit({
 
 }, 'odata');
 
-// TODO: replace by "export default" after "odata/context" refactor
+// TODO: replace by "export default" after "js/bundles/modules/data.odata" refactor
```
nightskylark: Do we need these TODOs? Are these places the only that we'll need to  convert? I'm pretty sure it will be done only if we have a planned task about it. But we already have a lot of senseless TODOs in project that nobody is gonna to do. Let's not add more.

LazyLahtak: I agree.
But I heard that we shouldn't mix `import` and `require` in one file (it may cause problems).
We can

1) Use `export default` and in `js/bundles/modules/data.odata` replace
`DevExpress.data.ODataStore = require('../../data/odata/store');`
by
`DevExpress.data.ODataStore = require('../../data/odata/store').default`;

2) We can replace all `require` in `js/bundles/modules/data.odata` by `import`, but it may cause problems (for example, with integrations).

What way should we choose?

nightskylark: I'm not telling you should fix it now. I just suggest to plan a task instead leaving a TODO.

nightskylark (dismissed):

MaximKudriavtsev (approved):

PR #12934, opened 2020-05-06
Button - Fixes submitting a form when AsyncRule is passed (T887207)

NickMitrokhin:
1. Fixed by resetting the **needValidate** flag after triggering a repeat click.
2. Added a test.

### DIFF
--- a/testing/tests/DevExpress.ui.widgets/button.tests.js
+++ b/testing/tests/DevExpress.ui.widgets/button.tests.js
@@ -542,6 +542,56 @@ QUnit.module('submit behavior', {
         ValidationEngine.registerValidatorInGroup('testGroup', validator);
         this.$element.trigger('dxclick');
     });
+
+    QUnit.test('Form should be submitted only when an async validation rule is passed positively (T887207)', function(assert) {
+        this.clock.restore();
+        let value = 'a';
+        const validValue = 'b';
+        const validator = new Validator($('<div>').appendTo(this.$form), {
+            adapter: sinon.createStubInstance(DefaultAdapter),
+            validationRules: [{
+                type: 'async',
+                validationCallback: function() {
+                    const d = new Deferred();
+                    setTimeout(() => {
+                        d.resolve({
+                            isValid: value === validValue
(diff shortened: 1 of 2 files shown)


DokaRus (dismissed):

DokaRus (approved):

testing/tests/DevExpress.ui.widgets/button.tests.js
```diff
@@ -542,6 +542,57 @@ QUnit.module('submit behavior', {
...
+                            triggerButtonClick();
+                        });
+                    }
+                }
+            }
+        });
+
+        ValidationEngine.registerValidatorInGroup('testGroup', validator);
+        triggerButtonClick();
+    });
```
San4es, line 595: Let's simplify this test.

```suggestion
    QUnit.test('Form should be submitted when an async validation rule is passed positively (T887207)', function(assert) {
        this.clock.restore();

        const done = assert.async();

        const validator = new Validator($('<div>').appendTo(this.$form), {
            adapter: sinon.createStubInstance(DefaultAdapter),
            validationRules: [{
                type: 'async',
                validationCallback: function() {
                    const d = new $.Deferred();
                    setTimeout(() => {
                        d.resolve({ isValid: true });
                    }, 10);
                    return d.promise();
                }
            }]
        });

        ValidationEngine.registerValidatorInGroup('testGroup', validator);

        this.$element.dxButton({ validationGroup: 'testGroup' });

        this.$form.on('submit', () => {
            assert.strictEqual(this.$element.dxButton('option', 'disabled'), false);
            done();
        });

        this.$element.trigger('dxclick');
    });
```

NickMitrokhin: The main idea is to check that the form is not submitted when a value is invalid and submitted only when a value is valid. I would not simplify this test to prevent regressions during renovations and so forth.

San4es: Ok.

San4es (approved):

PR #12951, opened 2020-05-07
Fix lint warnings in the jest tests

churkin:


testing/jest/widget.tests.tsx
```diff
@@ -1,35 +1,39 @@
...
 import {
   clear as clearEventHandlers, defaultEvent, emit, getEventHandlers, fakeClickEvent, EVENT,
 } from './utils/events-mock';
 import Widget from '../../js/renovation/widget.p';
 import type { WidgetRef } from '../../js/renovation/widget.p';
+import type { WidgetProps } from '../../js/renovation/widget';
 
 describe('Widget', () => {
-  const render = (props = {}) => shallow(<Widget {...props} />);
+  const render = (props = {}): ShallowWrapper => shallow(<Widget {...props} />);
```
MaximKudriavtsev, line 12: Can we define types here, like it did in this [example](https://stackoverflow.com/questions/44625581/testing-a-react-component-with-enzyme-typescript-cant-find-instance-methods). Instead of using `as WidgetProps` in multiple places?

MaximKudriavtsev (approved):

PR #12979, opened 2020-05-08
Fix lost aspnet.createComponent calls, part II (fixes T886572)

AlekseyMartynov:
I considered this back in September. Check 'Keep the handler in queue until the element is found' in #9540.

Now, thanks to ticket https://devexpress.com/issue=T886572, we have a test case that confirms the actual need.

### DIFF
--- a/js/aspnet.js
+++ b/js/aspnet.js
@@ -143,17 +143,29 @@
     function createComponent(name, options, id, validatorOptions) {
         var selector = '#' + String(id).replace(/[^\w-]/g, '\\$&');
         pendingCreateComponentRoutines.push(function() {
-            var $component = $(selector)[name](options);
-            if($.isPlainObject(validatorOptions)) {
-                $component.dxValidator(validatorOptions);
+            var $element = $(selector);
+            if($element.length) {
+                var $component = $(selector)[name](options);
+                if($.isPlainObject(validatorOptions)) {
+                    $component.dxValidator(validatorOptions);
+                }
+                return true;
             }
+            return false;
         });
     }
 
     templateRendered.add(function() {
         var snapshot = pendingCreateComponentRoutines.slice();
+        var leftover = [ ];
+
         pendingCreateComponentRoutines = [ ];
-        snapshot.forEach(function(func) { func(); });
+        snapshot.forEach(function(func) {
+            if(!func()) {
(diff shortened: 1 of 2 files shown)


nightskylark (commented):
LGTM

DokaRus (approved):

js/aspnet.js
```diff
@@ -143,17 +143,29 @@
...
+        var leftover = [ ];
+
         pendingCreateComponentRoutines = [ ];
-        snapshot.forEach(function(func) { func(); });
+        snapshot.forEach(function(func) {
+            if(!func()) {
+                leftover.push(func);
+            }
+        });
+        pendingCreateComponentRoutines = pendingCreateComponentRoutines.concat(leftover);
```
San4es, line 168: It looks like `pendingCreateComponentRoutines` is an empty array here. Can we just set `leftover` to it?
```suggestion
        pendingCreateComponentRoutines = leftover;
```

js/aspnet.js
```diff
@@ -143,17 +143,29 @@
...
 
     templateRendered.add(function() {
         var snapshot = pendingCreateComponentRoutines.slice();
+        var leftover = [ ];
+
         pendingCreateComponentRoutines = [ ];
-        snapshot.forEach(function(func) { func(); });
+        snapshot.forEach(function(func) {
+            if(!func()) {
+                leftover.push(func);
```
San4es, line 165: Can we push items to `pendingCreateComponentRoutines` directly?
```suggestion
                pendingCreateComponentRoutines.push(func);
```

js/aspnet.js
San4es: We found that `pendingCreateComponentRoutines` can be a not empty array when it is called recursively (for instance, during nested components rendering).

AlekseyMartynov:
@San4es 

- `snapshot` and `leftover` are function-local variables
- `pendingCreateComponentRoutines` is module-local
- `pendingCreateComponentRoutines` *might be* modified in a recursive `templateRendered` call. 
- `snapshot` isolates the inbound state for correct iteration
- `leftover` isolates outbound changes

You suggestions may be valid, however I didn't succeed in authoring a relevant test case.

San4es (approved):
LGTM

PR #13011, opened 2020-05-12
DataGrid - Prevents focusing hidden cells (T887014)

NickMitrokhin:
1. Added logic for removing the **tabindex** attribute when a cell is hiding and set **tabindex** to 0 when a cell is showing.
2. Added a test.

### DIFF
--- a/js/ui/grid_core/ui.grid_core.keyboard_navigation.js
+++ b/js/ui/grid_core/ui.grid_core.keyboard_navigation.js
@@ -87,7 +87,7 @@ function isMobile() {
 }
 
 function isCellInHeaderRow($cell) {
-    return !!$cell.closest(`.${HEADER_ROW_CLASS}`).length;
+    return !!$cell.parent(`.${HEADER_ROW_CLASS}`).length;
 }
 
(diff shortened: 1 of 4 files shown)


js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -82,6 +86,10 @@ function isMobile() {
     return devices.current().deviceType !== 'desktop';
 }
 
+function isCellInHeaderRow($cell) {
+    return !!$cell.closest(`.${HEADER_ROW_CLASS}`).length;
```
vconst: Use parent(`.${HEADER_ROW_CLASS}`) instead of closest. It is much faster.

vconst (changes requested):

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
NickMitrokhin: ok

vconst (approved):

Alyar666 (approved):

PR #13021, opened 2020-05-13
Refactor DOM utils 

nightskylark:
I faced with circular dependencies trying use an utility from this module. So I find out that here is a mess of methods and most of them should not be here. Here is a list of such methods.

- `setPublicElementWrapper`, `getPublicElement`
Extracted to the `core/element` module. These methods does not work with DOM. But it perfectly fits to the `core/element.d.ts` file. It describes a `dxElement` which can be an `Element` or a `jQuery` instance.

- `triggerShownEvent`, `triggerHidingEvent`, `triggerResizeEvent`
Extracted to the `events/visibility_change` module. These methods are alien here much more than other ones. And they are the main reason of this refactoring. They depend on `evens_engine` module where I was going to use another DOM utility.
 
- `createComponents`, `uniqueId`
Removed. These methods are atavisms that left after SPA.

- `createMarkupFromString`
Removed. This method was made to incapsulate a WinJS hacks. After getting rid of them this method turns into one-line `htmlParser.parseHTML` call, so method becomes a redundant layer. The rest of WinJS-linked code will be removed in a separate PR.

- `getElementOptions`, `extractTemplateMarkup`, `normalizeTemplateElement`
Skipped for now. I think they should be extracted to somewhere too. But this PR becomes too large so I decided to split it.

- `createTextElementHiddenCopy`
I discussed this method with Editors Squad but still not sure where to place it. Maybe it just should gone away after the greate beautiful world of the future will come.


DokaRus (dismissed):
LGTM

js/events/visibility_change.js
```diff
@@ -0,0 +1,22 @@
...
+            add($element.find(VISIBILITY_CHANGE_SELECTOR));
+
+        for(let i = 0; i < changeHandlers.length; i++) {
+            triggerHandler(changeHandlers[i], eventName);
+        }
+    };
+};
+
+
+exports.triggerShownEvent = triggerVisibilityChangeEvent('dxshown'); // TODO: extract to events
```
ovchinnikov: I'm not sure that increasing of TODOs is good idea. We already have [170 TODOs](https://github.com/DevExpress/DevExtreme/search?q=TODO) When it will be fixed? I suggest to use internal planning tool to list improvements

nightskylark: I totally agree about that. But I was going to get rid of them in next PR. I just decided that this one is too large to do it.

ovchinnikov (dismissed):

js/events/visibility_change.js
nightskylark: I was talking about [these TODOs](https://github.com/DevExpress/DevExtreme/pull/13021/files#diff-1ff023ae34a7f57e07f146626ad4e5d0R142). But I've just noticed that this comment is about another ones, which I forgot to remove. [Fixed](https://github.com/DevExpress/DevExtreme/pull/13021/commits/d2dbf2d53f76bb0ef3d0f1d75d0dabe43ca2bc7f).

ovchinnikov (approved):

IlyaKhD (approved):

PR #13060, opened 2020-05-15
DataGrid: Fix column resizing when columnResizingMode is widget and rtlEnabled is true (T837344)

Alyar666:
See https://devexpress.com/issue=T837344

### DIFF
--- a/js/ui/grid_core/ui.grid_core.columns_resizing_reordering.js
+++ b/js/ui/grid_core/ui.grid_core.columns_resizing_reordering.js
@@ -575,17 +575,17 @@ const ColumnsResizerViewController = modules.ViewController.inherit({
         return false;
     },
 
-    _isRtlMode: function() {
-        return this._$parentContainer && this._$parentContainer.parent().css('direction') === 'rtl';
+    _isRtlParentStyle: function() {
+        return this._$parentContainer && this.option('rtlEnabled') && this._$parentContainer.parent().css('direction') === 'rtl';
     },
 
     _pointCreated: function(point, cellsLength, columns) {
         const isNextColumnMode = isNextColumnResizingMode(this);
         const rtlEnabled = this.option('rtlEnabled');
-        const isRtlMode = this._isRtlMode();
-        const firstPointColumnIndex = !isNextColumnMode && rtlEnabled && !isRtlMode ? 0 : 1;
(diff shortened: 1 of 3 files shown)


js/ui/grid_core/ui.grid_core.columns_resizing_reordering.js
```diff
@@ -575,12 +575,17 @@ const ColumnsResizerViewController = modules.ViewController.inherit({
         return false;
     },
 
+    _isRtlMode: function() {
```
vconst: _isRtlMode -> _isRtlParentStyle

vconst: Also, I think that we should check this style only if rtlEnabled is true to improve performance without rtl

vconst (changes requested):

js/ui/grid_core/ui.grid_core.columns_resizing_reordering.js
```diff
@@ -733,7 +733,7 @@ const ColumnsResizerViewController = modules.ViewController.inherit({
             that._trackerView.show();
             const scrollable = that.component.getScrollable();
 
-            if(scrollable && that.option('rtlEnabled') && that._isRtlMode()) {
+            if(scrollable && that.option('rtlEnabled') && that._isRtlParentStyle()) {
```
jtoming830: isRtlParentStyle will check for this.option('rtlEnabled') so check rtlEnabled here is unnecessary

js/ui/grid_core/ui.grid_core.columns_resizing_reordering.js
```diff
@@ -575,17 +575,17 @@ const ColumnsResizerViewController = modules.ViewController.inherit({
         return false;
     },
 
-    _isRtlMode: function() {
-        return this._$parentContainer && this._$parentContainer.parent().css('direction') === 'rtl';
+    _isRtlParentStyle: function() {
+        return this._$parentContainer && this.option('rtlEnabled') && this._$parentContainer.parent().css('direction') === 'rtl';
```
jtoming830: You can minify this condition using optional chaining. 
`this.option('rtlEnabled') && this._$parentContainer?.parent.().css.('direction') === 'rtl';`

vconst (approved):

PR #13173, opened 2020-05-22
Animation position should be calculated correctly if target is svg element (T891214)

ksercs:


js/core/utils/position.js
```diff
@@ -1,9 +1,36 @@
...
 import config from '../config';
+import typeUtils from '../utils/type';
 
 const getDefaultAlignment = function(isRtlEnabled) {
     const rtlEnabled = isRtlEnabled ?? config().rtlEnabled;
 
     return rtlEnabled ? 'right' : 'left';
 };
 
+const getBoundingRect = (instance) => {
```
San4es: - Do we need yet another [size utils](https://github.com/DevExpress/DevExtreme/blob/4a46c3a58e8275d4c3933506efb07c5fb4795334/js/core/utils/size.js#L37)? 
- If yes, do we need to use it [everywhere](https://github.com/DevExpress/DevExtreme/search?q=getBoundingClientRect&unscoped_q=getBoundingClientRect)?
- Can we use our [`outerWidth`](https://github.com/DevExpress/DevExtreme/blob/4a46c3a58e8275d4c3933506efb07c5fb4795334/js/core/renderer_base.js#L167) implementation?

testing/tests/DevExpress.ui.widgets/popover.tests.js
```diff
@@ -636,6 +636,74 @@ QUnit.module('content positioning', () => {
         }
     });
 
+    QUnit.test('Popover should be positioned correctly on the target right side when target is inside of svg (T891214)', function(assert) {
```
San4es, line 639: What about adding tests for the `position` utils (see [position.tests.js](https://github.com/DevExpress/DevExtreme/blob/20_2/testing/tests/DevExpress.animation/position.tests.js))?

js/core/utils/position.js
ksercs: 1) No, this size util is too specific
2) Yeah, I gonna do it
3) No, it's not fully clear how to do it from a technical point of view

testing/tests/DevExpress.ui.widgets/popover.tests.js
ksercs: This module tests common position utils, etc calculatePosition - it's to common for adding a test for my new code. Besides, my new util is used in the main calculatePosition function, so it affects all of these tests

DokaRus (dismissed):

testing/tests/DevExpress.ui.widgets/popover.tests.js
ksercs: I've add test to another test module, see the changes

DokaRus (dismissed):
LGTM

js/core/utils/position.js
```diff
@@ -1,9 +1,41 @@
...
+    const element = $(instance).get(0);
```
nightskylark: I don't think it's a good idea to use renderer in the utility. It enlarges coupling. I'm sure that utility that works with DOM nodes do not need to know about the `renderer` entity. Maybe only DOM nodes should be allowed as an entry argument for this method?

ksercs: I thought about it. I agree that coupling  enlarges, but also util external api becomes more flexible. I decided that the second aspect is more important. But ok, i'll fix it

js/core/utils/position.js
```diff
@@ -1,9 +1,40 @@
...
+import typeUtils from '../utils/type';
 
 const getDefaultAlignment = function(isRtlEnabled) {
     const rtlEnabled = isRtlEnabled ?? config().rtlEnabled;
 
     return rtlEnabled ? 'right' : 'left';
 };
 
+const getBoundingRect = (instance) => {
+    const element = instance.get && instance.get(0) || instance;
```
San4es:
```suggestion
const getBoundingRect = (element) => {
    element = element.get && element.get(0) || element;
```

js/core/utils/position.js
```diff
@@ -1,9 +1,40 @@
...
+            width: element.outerWidth,
+            height: element.outerHeight
+        };
+    }
+
+    let rect;
+    try {
+        rect = element.getBoundingClientRect();
+    } catch(e) {
+        // TODO: add warning "IE throws Unspecified error if there is no such element on the page DOM"
```
San4es: When we are going to add this warning?

nightskylark (dismissed):

js/core/utils/position.js
ksercs: Probably, we will stop supporting the IE, and then the need for warning will disappear. I don’t think it’s worth doing it now - it's, at least, additional work with the documentation. And the scenario is rare. (The mere fact that the IE does not describe this error makes you think)

DokaRus (dismissed):

js/core/utils/position.js
San4es:
```suggestion
        // NOTE: IE throws 'Unspecified error' if there is no such element on the page DOM
```

San4es (approved):

DokaRus (approved):

PR #13219, opened 2020-05-27
NumberBox: removing last digit should change value to 0 even if it's negative when mask is enabled (T892552)

ksercs:


DokaRus (approved):

alexander-kotov-dx (approved):

San4es (commented):
LGTM

PR #13251, opened 2020-05-29
Remove own themebuilder linter

babich-a:


themebuilder-scss/src/modules/compile-manager.ts
```diff
@@ -4,6 +4,7 @@ import PreCompiler from './pre-compiler';
 import resolveBundle from './bundle-resolver';
 import PostCompiler from './post-compiler';
 import BootstrapExtractor from './bootstrap-extractor';
+// eslint-disable-next-line
```
San4es: What rule do we want to disable here?

themebuilder-scss/src/modules/compiler.ts
```diff
@@ -1,4 +1,5 @@
 import * as sass from 'sass';
+// eslint-disable-next-line
```
San4es: What rule do we want to disable here?

themebuilder-scss/src/modules/post-compiler.ts
```diff
@@ -2,6 +2,7 @@ import CleanCSS, { Options } from 'clean-css';
 import AutoPrefix from 'autoprefixer';
 import PostCss from 'postcss';
 import commonOptions from './clean-css-options';
+// eslint-disable-next-line
```
San4es: What rule do we want to disable here?

themebuilder-scss/src/modules/compile-manager.ts
babich-a: The module imported in the next line does not exist. It is generated while themebuilder is built.

churkin (dismissed):

themebuilder-scss/src/modules/compile-manager.ts
San4es: I suggest disabling a specific rule for this line. For instance:
https://github.com/DevExpress/DevExtreme/blob/5c77f06225bd4a73cd071eb1196d2868a69d7672/build/npm-bin/bundler.js#L14

San4es (approved):

churkin (approved):

PR #13302, opened 2020-06-03
DateBox: value should not contain time after value change when type is date and useMaskBehavior is true (T895591)

ksercs:


DokaRus (approved):

alexander-kotov-dx (approved):

San4es (commented):
LGTM

PR #13308, opened 2020-06-03
Stop trim the button aria-label value (T893304)

alexander-kotov-dx:


DokaRus (dismissed):

ksercs (dismissed):

testing/tests/DevExpress.ui.widgets/button.tests.js
```diff
@@ -152,6 +152,16 @@ QUnit.module('options changed callbacks', {
         this.instance.option('text', 'new test button text');
         assert.ok($testElement.parent().hasClass('dx-button'), 'test element is still in button');
     });
+
+    QUnit.test('button with non-string text option value should not rise an error (T893304)', function(assert) {
```
San4es:
```suggestion
    QUnit.test('button with non-string text option value should not raise an error (T893304)', function(assert) {
```

San4es (approved):

PR #13321, opened 2020-06-04
dxScheduler - Renovate Tooltip Item Content

AryamnovEugeniy:


js/renovation/scheduler/appointment-tooltip/tooltip-item-content.tsx
```diff
@@ -0,0 +1,52 @@
...
+  defaultOptionRules: null,
+  view: viewFunction,
+})
+export default class TooltipItemContent extends JSXComponent(TooltipItemContentProps) {
+  get formattedData(): FormattedContent {
+    const {
+      getTextAndFormatDate, appointmentData, currentAppointmentData,
+    } = this.props;
+
+    return getTextAndFormatDate!(appointmentData, currentAppointmentData);
```
pavelgruba: Props `appointmentData` and `currentAppointmentData` are here only to pass to another `getTextAndFormatDate` prop. Why not just pass formatted value as prop?

js/renovation/scheduler/appointment-tooltip/tooltip-item-layout.tsx
```diff
@@ -18,6 +18,9 @@ import TooltipItemContent from './tooltip-item-content';
 
 export const viewFunction = (viewModel: TooltipItemLayout) => {
   const useTemplate = !!viewModel.props.itemContentTemplate;
+  const formattedContent = viewModel.props.getTextAndFormatDate!(
```
pavelgruba: It's better to avoid any calculations in view function, component's getter is better here

pavelgruba (changes requested):

js/renovation/scheduler/appointment-tooltip/marker.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export const viewFunction = (viewModel: Marker) => (
+  <div
+    className={`${TOOLTIP_APPOINTMENT_ITEM_MARKER} ${viewModel.props.className}`}
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    {...viewModel.restAttributes}
+  >
+    <div
+      className={TOOLTIP_APPOINTMENT_ITEM_MARKER_BODY}
+      style={{ background: viewModel.appointmentColor! }}
```
MaximKudriavtsev: I think it would be better to use getter here

```js
getStyles() {
  return ({ background: viewModel.appointmentColor! });
}
```

js/renovation/scheduler/appointment-tooltip/tooltip-item-layout.tsx
AryamnovEugeniy: This calculation depends on value we get from getter (currentData getter). Would it be correct to move this to getters? Because if I call `currentData` getter inside another getter, it will be calculated twice: when it is called inside another getter and before it is passed to the viewFunction

pavelgruba (dismissed):

testing/jest/scheduler/appointment-tooltip/tooltip-item-layout.tests.tsx
```diff
@@ -0,0 +1,326 @@
...
+
+    it('should spread restAttributes', () => {
+      const tree = shallow(<TooltipItemLayoutView
+        restAttributes={{ customAttribute: 'customAttribute' }}
+        props={defaultProps}
+      />);
+
+      expect(tree.childAt(0).prop('customAttribute'))
+        .toBe('customAttribute');
+    });
```
MaximKudriavtsev: All test with `restAttributes` It can be like that
```js
const tree = shallow(<ErrorMessage />);
expect(tree.prop('restAttributes')).toBe('restAttributes');
```

testing/jest/scheduler/appointment-tooltip/tooltip-item-layout.tests.tsx
```diff
@@ -0,0 +1,326 @@
...
+
+    describe('Template', () => {
+      const currentData: dxSchedulerAppointment = { text: 'currentData' };
+      const template = () => null;
+      let container;
+      const render = () => shallow(<TooltipItemLayoutView
+        props={{ ...defaultProps, itemContentTemplate: template, container }}
+        currentData={currentData}
+      />);
```
MaximKudriavtsev: BTW, it looks better

```jsx
const render = () => shallow(
  <TooltipItemLayoutView
      props={{ ...defaultProps, itemContentTemplate: template, container }}
      currentData={currentData}
  />
);
```

churkin (dismissed):

andrewmakarov (dismissed):

pavelgruba (dismissed):

MaximKudriavtsev (dismissed):

Yarovaya (dismissed):

pavelgruba (approved):

AntonSermyazhko (approved):

andrewmakarov (approved):

PR #13326, opened 2020-06-04
DataGrid - Fixes showing error when reject is called in stateStoring.customLoad (T894590)

NickMitrokhin:
1. Implemented showing the error in the error row.
2. Added unit tests

### DIFF
--- a/js/ui/grid_core/ui.grid_core.state_storing.js
+++ b/js/ui/grid_core/ui.grid_core.state_storing.js
@@ -248,10 +248,9 @@ module.exports = {
                 _refreshDataSource: function() {
                     const callBase = this.callBase;
                     const stateStoringController = this.getController('stateStoring');
-                    const showErrorRow = (error) => {
-                        const errorHandling = this.getController('errorHandling');
+                    const showError = (error) => {
                         const errorMessage = error || 'Unknown error';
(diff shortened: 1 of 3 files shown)


jtoming830 (dismissed):

js/ui/grid_core/ui.grid_core.state_storing.js
```diff
@@ -246,25 +246,34 @@ module.exports = {
...
                 _refreshDataSource: function() {
-                    const that = this;
-                    const callBase = that.callBase;
-                    const stateStoringController = that.getController('stateStoring');
+                    const callBase = this.callBase;
+                    const stateStoringController = this.getController('stateStoring');
+                    const showErrorRow = (error) => {
+                        const errorHandling = this.getController('errorHandling');
+                        const errorMessage = error || 'Unknown error';
+                        errorHandling && errorHandling.renderErrorRow(errorMessage);
```
vconst: It will be better to call this._handleLoadError(errorMessage) because the errorRowEnabled option and onDataErorOccured should works in this case..

vconst (changes requested):

jtoming830 (dismissed):

vconst (approved):

PR #13357, opened 2020-06-08
Renovation: Provide only event into onKeyDown handler

LazyLahtak:
Refactor `onKeyDown` event for correct work with Native Angular.
Angular can provide only one argument into event.


js/renovation/button.tsx
```diff
@@ -208,18 +208,18 @@ export default class Button extends JSXComponent<ButtonProps> {
     useSubmitBehavior && this.submitInputRef.click();
   }
 
-  onWidgetKeyDown(event: Event, options) {
+  onWidgetKeyDown(options) {
     const { onKeyDown } = this.props;
-    const { keyName, which } = options;
+    const { originalEvent, keyName, which } = options;
```
MaximKudriavtsev, line 213: Why we use the name `originalEvent`? Can we call it just `event`?

LazyLahtak: The `originalEvent` came from old component. But we can rename this.

LazyLahtak: In fact, `options` - is `event` argument in `onKeyDown(event)`. I think for this reason we have `originalEvent` field.

MaximKudriavtsev: I see, but in the current context, we don't have any `event` variables and can make our code more readable.

MaximKudriavtsev (approved):

pavelgruba (approved):

churkin (approved):

PR #13377, opened 2020-06-10
Renovation: Rename `style` property

LazyLahtak:
The `style` property cannot be used in Vue (reserved name).
The `style` property have replaced with `styles`.

### DIFF
--- a/testing/jest/widget.tests.tsx
+++ b/testing/jest/widget.tests.tsx
@@ -86,28 +86,28 @@ describe('Widget', () => {
     describe('width/height', () => {
       it('should have the ability to be a function', () => {
         const widget = render({ width: () => 50, height: () => 'auto' });
-        const props = widget.props() as WidgetProps;
+        const props = widget.props() as any;
 
         expect(props.style).toEqual({ width: 50, height: 'auto' });
       });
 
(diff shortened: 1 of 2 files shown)


testing/jest/widget.tests.tsx
```diff
@@ -86,28 +86,28 @@ describe('Widget', () => {
     describe('width/height', () => {
       it('should have the ability to be a function', () => {
         const widget = render({ width: () => 50, height: () => 'auto' });
-        const props = widget.props() as WidgetProps;
+        const props = widget.props() as any;
```
MaximKudriavtsev, line 89: It seems like we should not use ts inside tests 😃

MaximKudriavtsev (approved):

pavelgruba (approved):

testing/jest/widget.tests.tsx
LazyLahtak: `widget.props()` doesn't return only `WidgetProps` type. For example, we have no `style` property in `Widget` component, but `widget.props()` has this property.

We can write next:
```tsx
const props = widget.props() as WidgetProps && { style: Object };
```

PR #13398, opened 2020-06-11
DataGrid: Fix validation message showing for a cell when Switch is used as an editor (T897363)

vconst:


js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -764,7 +764,11 @@ module.exports = {
                             rowKey: e.key,
                             columnIndex: e.column.index
                         });
-                        when(currentValidator && validatingController.validateCell(currentValidator)).done(deferred.resolve);
+                        when(currentValidator && validatingController.validateCell(currentValidator))
+                            .done(() => {
+                                this.getController('editorFactory').refocus();
+                            })
+                            .done(deferred.resolve);
```
```suggestion
                        when(currentValidator && validatingController.validateCell(currentValidator))
                           .done((validationResult) => {
                                this.getController('editorFactory').refocus();
                                deferred.resolve(validationResult);
                            });
```
It seems that the second **done** callback is redundant. I would keep only the first one.

NickMitrokhin (changes requested):

NickMitrokhin (approved):

PR #13404, opened 2020-06-11
Renovation: move qunit tests to the 'qunit' folder

churkin:


San4es:
@churkin see also https://github.com/DevExpress/DevExtreme/blob/20_2/build/build-dotnet.sln

PR #13410, opened 2020-06-11
ExcelJS export -  error is thrown when file opens with using the predefined 'billions' or 'millions' format (T899905)

novsstation:
Bugfix t899905


js/exporter/exceljs/export_data_grid.js
```diff
@@ -149,7 +149,7 @@ function _exportRow(rowIndex, cellCount, row, startColumnIndex, dataProvider, cu
 
             let numberFormat = _tryConvertToExcelNumberFormat(format, dataType);
             if(isDefined(numberFormat)) {
-                numberFormat = numberFormat.replace(/&quot;/g, '');
+                numberFormat = numberFormat.replace(/&quot;/g, '"');
```
IgnatovDan, line 152: I think we need an internal ticket in SC for this scenario: why the old code should be changed?
What 'error is thrown'? Where is the XSLX file? Where is the demo to reproduce this error with sources 'before change' and 'after change''? Why 'billion format' only? What chapter in XLSX specification describes the formatting? How 'billion' formatting is described in ExcelJS docs?

IgnatovDan (changes requested):

js/exporter/exceljs/export_data_grid.js
EugeniyKiyashko: I created a separate thread - [DataGrid - the exported file format corrupted when using 'billion' predefined grid column format(T899905)](https://supportcenter.devexpress.com/internal/ticket/details/T899905) . Into this ticket you can find the additional information about our choose.

IgnatovDan (approved):

jsdmitry (approved):

PR #13424, opened 2020-06-15
Lookup: toolbarItems should be passed to the popup even if usePopover is true (T896951)

ksercs:


js/ui/lookup.js
```diff
@@ -585,7 +585,8 @@ const Lookup = DropDownList.inherit({
             closeOnTargetScroll: true,
             width: this._isInitialOptionValue('dropDownOptions.width')
                 ? (function() { return this.$element().outerWidth(); }).bind(this)
-                : this._popupConfig().width
+                : this._popupConfig().width,
+            toolbarItems: this.option('dropDownOptions.toolbarItems')
```
DokaRus: How about other `dropDownOptions`?
Notice: https://github.com/DevExpress/DevExtreme/blob/20_2/js/ui/drop_down_editor/ui.drop_down_editor.js#L477

ksercs: fixed

DokaRus (approved):

San4es (approved):
LGTM

alexander-kotov-dx (approved):

PR #13449, opened 2020-06-16
Scheduler - Virtual Scrolling

AntonSermyazhko:


js/renovation/scheduler/workspaces/base/table.tsx
```diff
@@ -0,0 +1,29 @@
+import {
+  Component, ComponentBindings, JSXComponent, OneWay, Slot,
+} from 'devextreme-generator/component_declaration/common';
+
+export const viewFunction = (viewModel: Table) => (
+  <table
+    // eslint-disable-next-line react/jsx-props-no-spreading
+    {...viewModel.restAttributes}
+    className={`${viewModel.props.className}`}
```
pavelgruba: Why not just `className={viewModel.props.className}`?

js/renovation/scheduler/workspaces/day/date_table/layout.tsx
```diff
@@ -5,47 +5,58 @@ import { DateTableRow } from '../../base/date_table/row';
...
 }
 
 @Component({
   defaultOptionRules: null,
   view: viewFunction,
 })
 export class DayDateTableLayout extends JSXComponent(DayDateTableLayoutProps) {
+  get DateTable(): typeof VirtualTable | typeof Table {
+    return this.props.isVirtual ? VirtualTable : Table;
+  }
```
AryamnovEugeniy: We should either use it or delete it. if we use it, this getter should be tested

AntonSermyazhko: I've already tested this logic tests of the render:
https://github.com/DevExpress/DevExtreme/pull/13449/files/59efef5eafeb3428902e6c7ed2166e28bbddb9be#diff-1b1e8943bf4b91173fa173151f84a343R40-R52
Because checking of this logic needs rendering the components.

js/renovation/scheduler/workspaces/base/table.tsx
AntonSermyazhko: Sure, I've corrected this.

js/renovation/scheduler/workspaces/day/date_table/layout.tsx
AryamnovEugeniy: Getters should not be tested by rendering a component. And in your case it is not tested in render method. View function contains an exactly the same [piece of code](https://github.com/DevExpress/DevExtreme/pull/13449/files#diff-a4cd4307ff5fb4a0544f0a3bfd223077R12). So, it should be removed either from viewFunction or from here. If it is removed from here, there is no need for a new test. If it is removed from viewFunction though, this getter should be tested (without rendering a component).

AntonSermyazhko: Ok than. If it is so principle...
I will replace this getter with a piece of logic in the view function.

AntonSermyazhko: This getter is already redundant. I'll just remove it.

js/renovation/scheduler/workspaces/base/__tests__/table.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+      expect(table.exists())
+        .toBe(true);
+      expect(table.hasClass('some-class'))
+        .toBe(true);
+
+      const tbody = layout.find('tbody');
+      expect(tbody.exists())
+        .toBe(true);
+      expect(tbody.hasClass(''))
+        .toBe(true);
```
AryamnovEugeniy: It seems there is no need for this

AryamnovEugeniy (dismissed):

AryamnovEugeniy (approved):

pavelgruba (approved):

andrewmakarov (approved):

PR #13481, opened 2020-06-18
Make jQuery wrapper backward compatible with old one

pavelgruba:

### DIFF
--- a/js/renovation/button.tsx
+++ b/js/renovation/button.tsx
@@ -23,17 +23,6 @@ import BaseComponent from './preact-wrapper/button';
 
 const stylingModes = ['outlined', 'text', 'contained'];
 
-const getInkRippleConfig = ({ text, icon, type }: ButtonProps) => {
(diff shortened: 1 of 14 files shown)


roman-simionov (approved):

js/renovation/button.tsx
```diff
@@ -242,24 +219,36 @@ export default class Button extends JSXComponent(ButtonProps) {
   }
 
   get aria() {
-    return getAriaLabel(this.props.text, this.props.icon);
+    const { text, icon } = this.props;
```
LazyLahtak, line 222: Is there any reason why we get rid of `getAriaLabel` function?

pavelgruba: This function is small and used only in aria getter, so why do we need it at all? All that stuff is just getter

LazyLahtak (approved):

MikeVitik (approved):

PR #13529, opened 2020-06-22
DataGrid: Fix display  of the "Cancel Changes" button after changing a column which has the setCellValue callback (T897592) (refix T872751)

Alyar666:
See https://devexpress.com/issue=T897592

### DIFF
--- a/js/ui/grid_core/ui.grid_core.editing.js
+++ b/js/ui/grid_core/ui.grid_core.editing.js
@@ -1515,8 +1515,8 @@ const EditingController = modules.ViewController.inherit((function() {
         },
         saveEditData: function() {
             const deferred = new Deferred();
-            const afterSaveEditData = () => {
-                when(this._afterSaveEditData()).done(function() {
+            const afterSaveEditData = (cancel) => {
+                when(this._afterSaveEditData(cancel)).done(function() {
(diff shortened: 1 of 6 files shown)


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1527,7 +1527,7 @@ const EditingController = modules.ViewController.inherit((function() {
                 }
                 when(this._beforeSaveEditData()).done(cancel => {
                     if(cancel) {
-                        afterSaveEditData();
+                        afterSaveEditData(true);
```
NickMitrokhin:
```suggestion
                        afterSaveEditData(cancel);
```
Since **afterSaveEditData** accepts the **cancel** parameter and you call it with **true** when saving is canceled, pass **cancel** to the function instead of **true**.

NickMitrokhin (changes requested):

NickMitrokhin:
In addition, I think it is better to add tests for the same scenario for Batch to prevent regressions in future.

NickMitrokhin (dismissed):

vconst (dismissed):

vconst (dismissed):

vconst (approved):

PR #13535, opened 2020-06-23
DxElement type should be an HTMLElement

ovchinnikov:
Fixes T896721


js/core/element.d.ts
```diff
@@ -4,4 +4,4 @@
  * @hidden
  * @prevFileNamespace DevExpress.core
  */
-export type dxElement = Element & JQuery;
+export type dxElement = HTMLElement & SVGElement & JQuery;
```
IlyaKhD: Doc-tag have to updated as well: `* @type Node|SVGElement|jQuery`

IlyaKhD (dismissed):

IlyaKhD (approved):

PR #13553, opened 2020-06-24
Check rrule license header

Yarovaya:


AntonSermyazhko (dismissed):

.drone.yml
```diff
@@ -27,6 +27,7 @@ matrix:
   include:
     - { TARGET: lint }
     - { TARGET: ts }
+    - { TARGET: test_license }
```
San4es: We have similar configs for Shippable and Travis.

docker-ci.sh
```diff
@@ -195,6 +195,13 @@ function run_test_scss {
     npm i && npm run build && npm run test
 }
 
+function run_test_license {
+    npm i
+    npm run build
+
+    npx gulp check-license-notices
```
San4es: This task is called twice. See:
https://devextreme-ci.devexpress.com/DevExpress/DevExtreme/34035/51

San4es, line 202: This task is called twice: https://devextreme-ci.devexpress.com/DevExpress/DevExtreme/34035/51

docker-ci.sh
```diff
@@ -195,6 +195,13 @@ function run_test_scss {
     npm i && npm run build && npm run test
 }
 
+function run_test_license {
```
San4es: I have doubts about creating a separate task in the CI matrix:
- it builds everything but we check only one file (dx.all.js).
- `npm run build` takes about 8 minutes: https://devextreme-ci.devexpress.com/DevExpress/DevExtreme/34035/51
- the `check-license-notices` task takes only 2 sec.

Looks like we can run the `check-license-notices` task in Gulp tasks where we build any bundles (dx.all, dx.all.debug) with and without uglify.

AntonSermyazhko: If we check the license in the debug version, there is no guaranty that the license header will be presented in the product one. The product bundle is not build by any task of the tests matrix, so we did a separate task to check the license. And in this case there is no excessive consumption of process time.

docker-ci.sh
Yarovaya: I've updated PR and then we call 'check-license-notices' task only in 'test_license' task

docker-ci.sh
Yarovaya: Important notice - the 'check-license-notices' task takes 30 minutes for dx.all.debug.js.

AntonSermyazhko (approved):

San4es (approved):

PR #13554, opened 2020-06-24
Fix weekly recurrence changing in Scheduler, when byDay rule isn't specified (T900434)

Yarovaya:


js/ui/scheduler/ui.scheduler.recurrence_editor.js
```diff
@@ -644,14 +643,18 @@ const RecurrenceEditor = Editor.inherit({
     },
 
     _valueChangedHandler(args) {
-        const value = args.component.option('value');
+        const value = args.value;
+        const previousValue = args.previousValue;
```
AryamnovEugeniy: Let's use destructuring:
`const { value, previousValue } = args;`

js/ui/scheduler/recurrence.js
```diff
@@ -368,9 +368,13 @@ class RecurrenceValidator {
     }
 
     _wrongDayOfWeek(rule) {
+        const byDay = rule['byday'];
         const daysByRule = getRecurrenceProcessor().daysFromByDayRule(rule);
         let brokenDaysExist = false;
 
+        if(byDay === '') {
+            brokenDaysExist = true;
+        }
```
AryamnovEugeniy, line 377: I think, it is better to use ternary operator here

AryamnovEugeniy (dismissed):

andrewmakarov (approved):

AntonSermyazhko (approved):

PR #13604, opened 2020-06-26
Implement pager visible property

MikeVitik:
reanimate QUnit: Pager is not displayed when visible is false
resizable-contaner.test refactoring

### DIFF
--- a/js/renovation/pager/pager-props.tsx
+++ b/js/renovation/pager/pager-props.tsx
@@ -15,7 +15,8 @@ export default class PagerProps {
 
   @OneWay() pagesCountText?: string = messageLocalization.getFormatter('dxPager-pagesCountText')();
 
-  // visible: true,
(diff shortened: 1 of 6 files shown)


js/renovation/pager/pager-props.tsx
```diff
@@ -15,7 +15,8 @@ export default class PagerProps {
 
   @OneWay() pagesCountText?: string = messageLocalization.getFormatter('dxPager-pagesCountText')();
 
-  // visible: true,
```
vconst, line 18: Why is comment not removed?

MikeVitik: Without comment, it's not clear how visible related to test code

MikeVitik: What do you mean? Maybe you mistake?

vconst (approved):

PR #13672, opened 2020-07-02
dxScheduler - move mobile tooltip to window(T900555)

andrewmakarov:


scss/widgets/base/scheduler/_tooltip.scss
```diff
@@ -0,0 +1,14 @@
+$SCHEDULER_APPOINTMENT_BASE_COLOR: null !default;
```
babich-a, line 1: It is not a good idea to have both scheduler/ (folder) and _scheduler.scss file in the `base` folder. It can cause a collision while import `base/scheduler` module. (The same in `common` folder.)

AntonSermyazhko (approved):

AryamnovEugeniy (approved):

Yarovaya (approved):

PR #13692, opened 2020-07-03
RadioGroup - support value selection after changing a data source if the value wasn't in the previous data source (T903212)

alexander-kotov-dx:


ksercs (approved):

DokaRus (approved):

San4es (approved):
LGTM

PR #13725, opened 2020-07-07
Sortable Drag Performance: Fix drag item hang on dragLeave and dragEnter (Fixes T905423)

softworkz:
Happens in DropFeedbackMode Push. Code was wrong because animated row positions must not be reset on dragLeave because:

1. Consider typical drag-drop implementations on Windows:
  - Drag-Drop files in Windows Explorer
  - Drag-Drop text in a text editor
  - Drag-Drop diagrams in word
  - etc.
  You will never see that the source item that is being dragged is getting hidden once you drag the item out of the source window. In the same way, this should not be done here
2. The worst part is this:
  - Currently, immediately after collapsing all rows, they are expanded once again - inside the DragLeave event
  - First, all row positions are reset (collapsing the free-space row) and immediately afterwards, they are expanded again..

Collapsing the free space should only happen when

1. a row is dropped elsewhere
   AND
2. the item is meant to be removed from the source list (drag-move vs. drag-copy)

=> means neither in DragEnter nor in DragLeave

And this is what this commit is changing.

(Fixes T905423)

### DIFF
--- a/js/ui/sortable.js
+++ b/js/ui/sortable.js
@@ -147,15 +147,11 @@ const Sortable = Draggable.inherit({
     dragEnter: function() {
         if(this === this._getTargetDraggable()) {
             this.option('toIndex', this.option('fromIndex'));
-        } else {
-            this.option('toIndex', -1);
         }
     },
 
     dragLeave: function() {
-        if(this === this._getTargetDraggable()) {
-            this.option('toIndex', -1);
-        } else {
+        if(this !== this._getTargetDraggable()) {
             this.option('toIndex', this.option('fromIndex'));
         }
     },


jtoming830:
Hello,

I have noticed that this code disables animation for all rows in certain cases. Currently we don't support virtual scrolling with `dropFeedbackMode: 'push'`. We plan to fix row dragging performance in future releases. I am closing this pull request.

softworkz:
> Hello,
> 
> I have noticed that this code disables animation for all rows in certain cases. Currently we don't support virtual scrolling with `dropFeedbackMode: 'push'`. We plan to fix row dragging performance in future releases. I am closing this pull request.

This is not about virtual scrolling.

This is about your code resetting all rows position on dragLeave and immediately afterwards, you are starting animation on all rows to get right back to the exact situation where it was before. 
With the tiny difference being that this is wasting CPU cycles and causes the UI to hang for several seconds.

This is a bug in your code and IMO it is not acceptable to simply close and tell something about virtual scrolling (which is not related to this at all).

softworkz:
> I have noticed that this code disables animation for all rows in certain cases.

Well - how about getting it right, then? 
I already spent two workdays for what is actually supposed to be your job.

And I can only repeat: what your code does awfully bad. It can be fixed and that does not depend in any way on any architectural changes that you might want to do in the future.

softworkz:
Has been applied here: https://github.com/DevExpress/DevExtreme/commit/3dc5efd0393a9ac3d102a35173dfb75c4b7e1578
(with a slight modification)

PR #13726, opened 2020-07-07
Sortable: Only animate items around the current drop position (Fixes T905623)

softworkz:
Animating hundreds or thousands of rows doesn't really make sense.

What this PR does:

- Only animate few items around the current drop position
- Instantly move all other items instead

Fixes T905623 (needs fix for t905423 as well to be fully resolved)

### DIFF
--- a/js/ui/sortable.js
+++ b/js/ui/sortable.js
@@ -711,6 +711,9 @@ const Sortable = Draggable.inherit({
         const animationConfig = this.option('animation');
         const rtlEnabled = this.option('rtlEnabled');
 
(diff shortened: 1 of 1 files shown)


softworkz:
One more note: Try to reduce the animation duration from 300ms to 200 or 150ms. This will provide visually better results and make things run more fluently.

jtoming830:
Hello,

I am closing this pull request because `animationLowerLimit` and `animationUpperLimit` calculation is incorrect for some configurations. Last visible rows may not be animated. We plan to improve row dragging performance in future releases.

softworkz:
So you think it is a good idea, not to change anything and continue to allow the code to start up to 1000 simultaneous animations when the grid has 1000 rows?

PR #13788, opened 2020-07-13
dxScheduler - add renovated Vertical Group Panel

AryamnovEugeniy:


js/renovation/scheduler/workspaces/utils.tsx
```diff
@@ -0,0 +1,20 @@
+export const getKeyByDateAndGroup = (date: Date, group?: object): string => {
+  const keyFromDate = date.toString();
+  if (!group) return keyFromDate;
```
AntonSermyazhko: In this case, I suggest using a simple `if / else`, since the logic with` return` next to the declarations degrades the readability of the code.

pavelgruba (approved):

AntonSermyazhko (approved):

MaximKudriavtsev (approved):

andrewmakarov (approved):

PR #13808, opened 2020-07-14
dxScheduler - select cells correctly when appointments are grouped vertically by more than one resource

AryamnovEugeniy:


js/ui/scheduler/workspaces/ui.scheduler.work_space.js
```diff
@@ -929,10 +929,9 @@ class SchedulerWorkSpace extends WidgetObserver {
     _getRowCount() { return noop(); }
 
     _getRowCountWithAllDayRows() {
-        const allDayRowsCount = this.option('showAllDayPanel')
-            ? this._getGroupCount() : 0;
+        const allDayRowCounted = this.option('showAllDayPanel') ? 1 : 0;
```
churkin: `const allDayRowCounted = +this.option('showAllDayPanel');` ? :)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Unary_plus

churkin (dismissed):

andrewmakarov (dismissed):

js/ui/scheduler/workspaces/ui.scheduler.work_space.js
AryamnovEugeniy: To me this seems less clear. But if you insist, I will do it)

AntonSermyazhko: `+this.option('showAllDayPanel')` is not clear to me too.
I think we should rename: `allDayRowCounted` -> `allDayRowCount`

AntonSermyazhko (approved):

PR #13867, opened 2020-07-17
Sortable: Fix performance during cross component d&d with large item count (T913673)

vconst:


NickMitrokhin (approved):

Alyar666 (approved):

js/ui/sortable.js
```diff
@@ -17,6 +17,23 @@ const CLONE_CLASS = 'clone';
 
 const isElementVisible = itemElement => $(itemElement).is(':visible');
 
+const animate = (element, config) => {
+    if(!element) return;
+
+    const left = config.to?.left || 0;
+    const top = config.to?.top || 0;
+
+    element.style.transform = `translate(${left}px,${top}px)`;
```
softworkz, line 26: ### This is not a good way

It is much slower than using predefined CSS rules.

Of course you can't know the actual offset values in advance.
That's no problem:

- either pre-create like 50 CSS rules for each direction
  (for the rare cases of higher values, you can still use the style like above)
- or create the required CSS rules dynamically at runtime

Example:

```css
.translateTop_0 {
    transform: translate( 0px, 0px);
}
.translateTop_1 {
    transform: translate( 0px, 1px);
}
.translateTop_2 {
    transform: translate( 0px, 2px);
}
.translateTop_3 {
    transform: translate( 0px, 3px);
}
etc...
```

#### => Then assign the matching css class in the `animate` method

Only that way, you'll achieve the best possible performance.

js/ui/sortable.js
```diff
@@ -17,6 +17,23 @@ const CLONE_CLASS = 'clone';
...
+    element.style.transition = fx.off ? '' : `all ${config.duration}ms ${config.easing}`;
```
softworkz, line 27: ### This is wrong

(or rather: it gets wrong in line 34 where the transition is cleared)

In order to animate elements, the transition setting needs to remain constant.

You must only change the CSS properties that should be transitioned from value X to value Y. 

### Problems
#### 1. Incorrect Animation Appearance
When value Y is the base state (e.g. no translation), you must not clear the transition property (like in line 34), because that avoids the "reverting" animation from happening.
This can be seen with your code applied, where the rows in a 2nd grid immediately jump back instead of closing the gap in an animated way.

#### 2. Performance
Another reason not to change the transition via style and for each element individually is browser performance. (You should not only test this with 100 rows in Chrome)

Best performance will be achieved by assigning this via a CSS rule to all rows and never add or remove that CSS class from items.

#### 3. Limit Animation Properties

Specifying `all` for the transition is not a good way, because it will animate all style changes which is surely undesired. 

Use 

`transform ${config.duration}ms ${config.easing}`;

instead.

softworkz (commented):
Please also see my comments and CodePen demonstrating the flaws of this fix in **T913673**.

Kind regards,
softworkz

PR #13872, opened 2020-07-17
Support active state for non-readonly custom buttons inside readonly editor (T905970)

alexander-kotov-dx:


alexander-kotov-dx:
reopen as https://github.com/DevExpress/DevExtreme/pull/13886

PR #13873, opened 2020-07-18
Renovation: get rid of ts warnings

churkin:


js/renovation/ui/button.tsx
```diff
@@ -40,19 +41,18 @@ const getCssClasses = (model: ButtonProps) => {
...
 };
-
-export const viewFunction = (viewModel: Button) => {
+export const viewFunction = (viewModel: Button): JSX.Element => {
   const {
-    children, icon, iconPosition, template, text,
+    children, icon, iconPosition, template: ButtonTemplate, text,
   } = viewModel.props;
-  const renderText = !template && !children && text;
+  const renderText = !ButtonTemplate && !children && text;
```
vconst, line 48: WTF? Why are tests passed?

js/renovation/ui/button.tsx
```diff
@@ -40,19 +41,18 @@ const getCssClasses = (model: ButtonProps) => {
...
   const isIconLeft = iconPosition === 'left';
-  const iconComponent = !template && !children && viewModel.iconSource
+  const iconComponent = !ButtonTemplate && !children && viewModel.iconSource
```
vconst, line 50: WTF?

js/renovation/ui/button.tsx
```diff
@@ -74,20 +74,20 @@ export const viewFunction = (viewModel: Button) => {
...
     >
       <div className="dx-button-content" ref={viewModel.contentRef as any}>
-        {template
-                && (
-                <viewModel.props.template
-                  data={{ icon, text }}
-                />
-                )}
-        {!template && children}
+        {ButtonTemplate && (<ButtonTemplate data={{ icon, text }} />)}
```
vconst, line 77: Why is template replaced to ButtonTemplate?

js/renovation/ui/button.tsx
```diff
@@ -74,20 +74,20 @@ export const viewFunction = (viewModel: Button) => {
...
+        {!ButtonTemplate && children}
```
vconst, line 78: Why is template replaced to ButtonTemplate?

js/core/devices.d.ts
```diff
@@ -149,6 +149,14 @@ declare class DevicesObject {
...
      */
     real(): Device;
+    /**
+     * @docid DevicesObjectMethods.isSimulator
+     * @publicName isSimulator()
+     * @return Boolean
+     * @prevFileNamespace DevExpress.core
+     * @hidden
+     */
+    isSimulator(): boolean;
```
vconst, line 152: Why is doc comment added here?

js/renovation/ui/button.tsx
roman-simionov: it is just a variable name

https://github.com/DevExpress/DevExtreme/pull/13873/files#diff-6bde74af2450bbb641a700bbb770adf2R46

roman-simionov (dismissed):

js/renovation/ui/button.tsx
churkin: What is wrong?

js/renovation/ui/button.tsx
churkin: WTF?

js/renovation/ui/button.tsx
churkin: JSX requires upper case name for the component

js/core/devices.d.ts
churkin: Where should I add they?

js/renovation/ui/button.tsx
vconst: It is not a type. I think that ButtonTemplate is type of template.

vconst (dismissed):

js/core/devices.d.ts
vconst: Previously, the isSimulator method was internal. We want to publish it? Usually, we do not write doc comments for internal methods.

churkin: It's marked as a "hidden" (https://github.com/DevExpress/DevExtreme/pull/13873/files#diff-5f66e6a1045e4cd65e0e6007dd944fe2R43) @IlyaKhD is everything good here?

IlyaKhD: If the intention is to hide a member from wrappers and other generators, then doc-tags shouldn't be added at all.

roman-simionov (approved):

PR #13894, opened 2020-07-21
DataGrid: Fix revert button (T905677)

Alyar666:

### DIFF
--- a/testing/testcafe/tests/dataGrid/editing.ts
+++ b/testing/testcafe/tests/dataGrid/editing.ts
@@ -1401,22 +1401,25 @@ test('Rollback changes on a click on a revert button  when startEditAction is db
     .doubleClick(cell0.element)
     .expect(cell0.isEditCell).ok()
     .click(cell0.element.find('.dx-checkbox'))
-    .expect($revertButton.exists).ok()
+    .expect($revertButton.exists)
+    .ok()
     .click($revertButton)
-    .expect($revertButton.exists).notOk()
-    .expect(cell0.isEditCell).notOk();
+    .expect($revertButton.exists)
+    .notOk()
+    .expect(cell0.isEditCell)
+    .notOk();
 }).before(() => createWidget('dxDataGrid', {
   dataSource: [{ name: 'test', test: undefined }],
   editing: {
     mode: 'cell',
     allowUpdating: true,
-    startEditAction: 'dblClick'
+    startEditAction: 'dblClick',
   },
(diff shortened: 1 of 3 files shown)


testing/testcafe/tests/dataGrid/editing.ts
```diff
@@ -1389,3 +1389,37 @@ test('Async Validation(Batch) - Validation frame should be rendered when a neigh
...
   }],
 })));
+
+// T905677
+test('Rollback changes on a click on a revert button  when startEditAction is dblclick', async (t) => {
+  const dataGrid = new DataGrid('#container');
+  const dataRow = dataGrid.getDataRow(0);
+  const cell0 = dataRow.getDataCell(1);
+  const $revertButton = Selector('.dx-revert-button');
+
```
NickMitrokhin: It is better to add the **getRevertButton** method to the **DataCell** class that is placed in the "testing\testcafe\model\dataGrid\data\cell.ts" file. For instance:

```ts
const CLASS = {
  ...
  revertButton: 'dx-revert-button',
};

export default class DataCell extends FocusableElement {
  ...
  getRevertButton(): Selector {
    return this.element.find(`.${CLASS.revertButton}`);
  }
}
```

After that, you will be able to get the button and use it in the test in the following manner:
```ts
const revertButton = cell0.getRevertButton();

await t
    ...
    .click(revertButton)
    .expect(revertButton.exists)
```

testing/testcafe/tests/dataGrid/editing.ts
```diff
@@ -1389,3 +1389,37 @@ test('Async Validation(Batch) - Validation frame should be rendered when a neigh
...
+    .expect(cell0.isEditCell).ok()
+    .click(cell0.element.find('.dx-checkbox'))
+    .expect($revertButton.exists)
+    .ok()
+    .click($revertButton)
+    .expect($revertButton.exists)
+    .notOk()
+    .expect(cell0.isEditCell)
+    .notOk();
+}).before(() => createWidget('dxDataGrid', {
```
NickMitrokhin, line 1414: I would add checking the current cell value using the following code:
```suggestion
    .notOk()
    .expect(dataGrid.apiGetCellValue(0, 1))
    .notOk(); //means that the cell value is undefined after clicking the revert button
}).before(() => createWidget('dxDataGrid', {
```

NickMitrokhin (approved):

vconst (approved):

PR #13910, opened 2020-07-22
Renovation: remove testing/jest folder, test utils refactoring

churkin:


jest.config.js
```diff
@@ -25,19 +25,15 @@ module.exports = {
             branches: 0
```
MaximKudriavtsev, line 25: When we update these values?

MaximKudriavtsev (dismissed):

jest.config.js
churkin: Do all tests already 100% cover renovation code?

MaximKudriavtsev: I think no. But we can increase these properties to some values to get rid of adding widgets without testing. And we can ask some guys to increase code coverage rating.

churkin: @ksercs will do it

roman-simionov (dismissed):

MaximKudriavtsev (dismissed):

roman-simionov (approved):

PR #13921, opened 2020-07-22
ExcelJS.PivotGrid - add doс comments for arguments of the 'exportPivotGrid' function

EugeniyKiyashko:


novsstation (approved):

IlyaKhD (approved):

js/excel_exporter.d.ts
```diff
@@ -45,6 +46,33 @@ export interface ExcelDataGridCell {
     value?: any;
 }
 
+/**
+* @docid ExcelPivotGridCell
+* @namespace DevExpress.excelExporter
+* @type object
+* @inherits dxPivotGridPivotGridCell
+*/
+export interface ExcelPivotGridCell extends dxPivotGridPivotGridCell {
```
IgnatovDan, line 55: Do we really need dxPivotGridPivotGridCell object in a handler? Will we change its api?
For DataGrid we did not inherited a similar interface from dxDataGridDataGridCell

IgnatovDan (changes requested):

js/excel_exporter.d.ts
IgnatovDan: It's ok

IgnatovDan (approved):

js/excel_exporter.d.ts
EugeniyKiyashko: Yes. For the PivotGrid we already have a ready-made description for the cell properties, that need for us. In the case of DataGrid widget such description was skipped and we couldn't use the similar interface inheritance.

PR #13943, opened 2020-07-23
Form - Fix changing item properties via itemOption(id, options) (T903774)

novsstation:
Allows the user to disable the partial update of the form (T903774)

### DIFF
--- a/testing/tests/DevExpress.ui.widgets.form/form.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.form/form.tests.js
@@ -455,7 +455,7 @@ QUnit.test('From renders the right types of editors according to stylingMode opt
                 form.repaint();
             }
             if(repaintChangesOnly === false) {
-                assert.equal(form.itemOption('group1.item1'), undefined, 'item1');
+                assert.deepEqual(form.itemOption('group1.item1'), undefined, 'item1');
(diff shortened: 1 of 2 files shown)


testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -423,6 +424,76 @@ QUnit.test('From renders the right types of editors according to stylingMode opt
...
+                items: [ clone(item1) ]
+            }).dxForm('instance');
+
+            form.itemOption('item1', clone(newItem1));
+            if(useRepaint) {
+                form.repaint();
+            }
+
+            assert.deepEqual(form.itemOption('item1'), undefined, 'item1');
+            assert.deepEqual(form.itemOption('newItem1'), newItem1, 'newItem1');
```
IgnatovDan: inho it should not depend on the item' level and there should be the same "if(repaintChangesOnly === false) {"

IgnatovDan (changes requested):

testing/tests/DevExpress.ui.widgets.form/form.tests.js
IgnatovDan: skip for new tickets

IgnatovDan (dismissed):

PR #14059, opened 2020-07-30
Tagbox: selected items and text should not be cleared when searchValue length becomes smaller then minSearchLength and showSelectionControls is true (T898390)

ksercs:


DokaRus (approved):

San4es (approved):
LGTM

PR #14069, opened 2020-07-31
Fix: DataGrid - Resize icon showing then resize isn't possible in Cell and Batch edit modes (T916159)

MikeVitik:
Extract canResize method from _startResizing and call it during icon updating to resize.

### DIFF
--- a/testing/tests/DevExpress.ui.widgets.dataGrid/columnsResizingReorderingModule.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.dataGrid/columnsResizingReorderingModule.tests.js
@@ -1735,6 +1735,38 @@ QUnit.module('Columns resizing', {
         assert.ok(resizeController._isResizing, 'columnsResizer is resizing');
     });
 
+    ['cell', 'batch'].forEach((mode) => {
+        QUnit.test(`Dont show resize cursor during editing in ${mode} mode. T915568`, function(assert) {
+            // arrange
+            const resizeController = this.createColumnsResizerViewController();
+            this.options['editing.mode'] = mode;
+            this.component._controllers.editing._isEditing = true;
+
+            this.renderViews($('#container'));
+            resizeController._pointsByColumns = [
+                { x: -9875, columnIndex: 0, index: 1, y: -9995 },
+                { x: -9750, columnIndex: 1, index: 2, y: -9995 },
+            ];
+
+            // assert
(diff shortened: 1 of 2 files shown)


testing/tests/DevExpress.ui.widgets.dataGrid/columnsResizingReorderingModule.tests.js
```diff
@@ -1735,6 +1735,38 @@ QUnit.module('Columns resizing', {
...
+
+            this.renderViews($('#container'));
+            resizeController._pointsByColumns = [
+                { x: -9875, columnIndex: 0, index: 1, y: -9995 },
+                { x: -9750, columnIndex: 1, index: 2, y: -9995 },
+            ];
+
+            // assert
+            resizeController._columnsSeparatorView.changeCursor = function(cursorName) {
+                assert.notEqual(cursorName, 'col-resize', 'we cannot resize column while editing is active');
```
jtoming830, line 1753: Is it possible to check css attribute instead of mocking 'changeCursor' function?

MikeVitik: I copy it from the other test and I suppose it's normal in the current case.

MikeVitik: Additional info. We have the test that check the css attribute after changeCursor called. So another word checking css attribute in my test is duplication and not necessary.
```
    QUnit.test('changeCursor', function(assert) {
        // arrange
        const columnsSeparator = createColumnsSeparator();

        // act
        columnsSeparator.changeCursor('col-resize');

        // assert
        assert.equal(columnsSeparator.element().css('cursor'), 'col-resize', 'cursor');
    });
```

jtoming830 (approved):

NickMitrokhin (approved):

softworkz:
From my point of view, this goes into an awfully wrong direction.

Especially in cases where you are using cell editing: When all cells are editable, you are practically permanently in a situation where editing is active.

And that means that you'll never be able to resize columns.

Please re-consider that restriction, it doesn't make sense.

Also, you are still allowing to re-order columns via drag-drop. So why allow re-ordering columns but not re-sizing?

That doesn't make sense at all!

Kind regards,
softworkz

softworkz:
BTW: Issue reference is incorrect. Should be T915568!

PR #14132, opened 2020-08-06
Generate cldr-data with ES6 export (hotfix for gulp tasks)

dxvladislavvolkov:


build/gulp/localization.js
```diff
@@ -17,7 +17,7 @@ const weekData = require('cldr-core/supplemental/weekData.json');
 const likelySubtags = require('cldr-core/supplemental/likelySubtags.json');
 const parentLocales = require('../../node_modules/cldr-core/supplemental/parentLocales.json').supplemental.parentLocales.parentLocale;
 
-const getParentLocale = require('../../js/localization/parentLocale.js');
+const getParentLocale = require('./parentLocale.js');
```
nightskylark: Just inline this method here. We don't need such additional file.

nightskylark (changes requested):

nightskylark (approved):

PR #14135, opened 2020-08-06
DataGrid/TreeList: Publish editColumnName option

jtoming830:


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1167,19 +1249,24 @@ const EditingController = modules.ViewController.inherit((function() {
         },
 
         editCell: function(rowIndex, columnIndex) {
+            return this._editCell(rowIndex, columnIndex);
+        },
+
+        _editCell: function(rowIndex, columnIndex, oldState) {
```
NickMitrokhin: The **oldState** argument is confusing a little bit. Isn't better to wrap parameters into objects? For instance: 

```js
_editCell: function(newParams, oldParams)
```
or

```js
_editCell: function(newState, oldState)
```

usage:
```js
const newParams = {
    rowIndex, 
    columnIndex
};
const oldParams = {
    rowIndex,
    columnIndex
};
this._editCell(newParams, oldParams);
```
**_editCellCore** can be modified in the same manner.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -586,6 +578,47 @@ const EditingController = modules.ViewController.inherit((function() {
...
+                if(isDefined(args.value)) {
+                    if(args.value !== args.previousValue) {
+                        this._editRow(rowIndex, oldRowIndex);
+                    }
+                } else {
+                    this.cancelEditData();
+                }
+            } else {
+                if(isDefined(args.value)) {
+                    if(args.value !== args.previousValue) {
```
vconst: Why is not one "if" with "&&"?

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -586,6 +578,47 @@ const EditingController = modules.ViewController.inherit((function() {
...
+                }
+            }
+        },
+
+        _handleEditColumnNameChange: function(args) {
+            const rowIndex = this._getVisibleEditRowIndex();
+
+            if(this.isCellOrBatchEditMode() && rowIndex !== -1) {
+                if(isDefined(args.value)) {
+                    if(args.value !== args.previousValue) {
```
vconst: Why is not one "if"?

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -595,20 +628,26 @@ const EditingController = modules.ViewController.inherit((function() {
...
                 this._resetEditRowKey();
             }
         },
 
         isEditing: function() {
-            return isDefined(this.option('editing.editRowKey'));
+            const isEditRowKeyDefined = isDefined(this.option('editing.editRowKey'));
+
+            if(this.isCellOrBatchEditMode()) {
+                return isEditRowKeyDefined && isDefined(this.option('editing.editColumnName'));
```
vconst: Extract isEditColumnNameDefined constant. It will be consistent with isEditRowKeyDefined and more understandable.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -959,6 +998,48 @@ const EditingController = modules.ViewController.inherit((function() {
...
+        _resetEditColumnName: function() {
+            this._setEditColumnName(null, true);
+        },
+
+        _getEditColumn: function() {
+            const visibleColumns = this._columnsController.getVisibleColumns();
+            const editColumnName = this.option('editing.editColumnName');
+            let editColumn;
+
+            isDefined(editColumnName) && visibleColumns.some(column => {
```
vconst, line 1024: May be filter will be better?
`isDefined(editColumnName) && visibleColumns.filter(column => column.name == editColumnName)[0];`

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1206,7 +1304,7 @@ const EditingController = modules.ViewController.inherit((function() {
             const column = params.column = visibleColumns[columnIndex];
 
             if(column && item && (item.rowType === 'data' || item.rowType === 'detailAdaptive') && !item.removed && !isRowEditMode(that)) {
-                if(that.isEditCell(rowIndex, columnIndex)) {
+                if(!oldState && that.isEditCell(rowIndex, columnIndex)) {
```
vconst: Why?

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -5304,7 +5304,7 @@ QUnit.module('Editing with real dataController', {
             isCloseEditCell = true;
         };
 
-        this.editRow(0);
+        this.editCell(0, 0);
```
vconst, line 5307: Why?

vconst (changes requested):

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
jtoming830: Because `editRow` does not work for batch and cell edit modes. [This is mentioned in our docs](https://js.devexpress.com/Documentation/ApiReference/UI_Widgets/dxDataGrid/Methods/#editRowrowIndex).

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: `some` will be faster and will not create a new array for each `_getEditColumn` method call.
`_getEditColumn` is called every time when `_getVisibleEditRowIndex` is called and `_getVisibleEditRowIndex` has 19 different call sources in the editing.js file. It's called many times so I think this optimization will be nice for cases with a lot of columns.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -959,6 +989,48 @@ const EditingController = modules.ViewController.inherit((function() {
...
+            if(!isDefined(editColumnName)) {
+                return -1;
+            }
+
+            return this._columnsController.getVisibleColumnIndex(editColumnName);
+        },
+
+        _setEditColumnNameByIndex: function(index) {
+            const visibleColumns = this._columnsController.getVisibleColumns();
+            this._setEditColumnName(visibleColumns[index].name, true);
```
vconst: The null ref exception may be thrown if user pass wrong columnIndex in editCell method

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1167,19 +1240,24 @@ const EditingController = modules.ViewController.inherit((function() {
...
                 when(coreResult)
                     .done(d.resolve)
                     .fail(d.reject);
             });
 
             return coreResult !== undefined ? coreResult : d.promise();
         },
-        _editCellCore: function(rowIndex, columnIndex) {
+
+        _editCellCore: function({ rowIndex, columnIndex, oldRowIndex, oldColumnIndex } = {}) {
```
vconst: I think that `= {}`  is not necessary here. The  _editCellCore method call without parameters wrong and impossible.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1189,9 +1267,19 @@ const EditingController = modules.ViewController.inherit((function() {
...
+            let oldColumn;
+            if(isDefined(oldColumnIndex)) {
+                oldColumn = visibleColumns[oldColumnIndex];
+            } else {
+                oldColumn = this._getEditColumn();
+            }
+
+            if(!isDefined(oldRowIndex)) {
+                oldRowIndex = that._getVisibleEditRowIndex();
+            }
```
vconst: The _editCellCore method became big. Extract options processing to separate method like _normalizeEditCellOptions.

vconst (changes requested):

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: Ok

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: Ok

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1167,62 +1240,83 @@ const EditingController = modules.ViewController.inherit((function() {
...
-            const dataController = that._dataController;
-            const items = dataController.items();
-            const item = items[rowIndex];
-            const params = {
-                data: item && item.data,
-                cancel: false
-            };
-            const oldEditRowIndex = that._getVisibleEditRowIndex();
+
+        _normalizeEditCellOptions: function({ oldColumnIndex, oldRowIndex, columnIndex, rowIndex }) {
```
NickMitrokhin: Since we expect that the method returns something, I think it is better to name it **_getNormalizedEditCellOptions**.
```suggestion

        _getNormalizedEditCellOptions: function({ oldColumnIndex, oldRowIndex, columnIndex, rowIndex }) {
```

MikeVitik (dismissed):

MikeVitik (approved):

NickMitrokhin (approved):

vconst (approved):

PR #14149, opened 2020-08-07
TreeView - Fix expandig nodes after repaint() if virtual mode is used (T920415)

novsstation:
Bugfix T920415

### DIFF
--- a/js/ui/hierarchical_collection/ui.data_adapter.js
+++ b/js/ui/hierarchical_collection/ui.data_adapter.js
@@ -323,7 +323,7 @@ const DataAdapter = Class.inherit({
 
         const itemKey = this.options.dataAccessors.getters.key(item);
         each(this._dataStructure, function(_, node) {
-            const isNodeFound = typeUtils.isDefined(itemKey)
+            const isNodeFound = itemKey !== undefined
                 ? node.internalFields.key === itemKey
                 : node.internalFields.item === item;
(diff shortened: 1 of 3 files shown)


js/ui/hierarchical_collection/ui.data_adapter.js
```diff
@@ -321,8 +321,12 @@ const DataAdapter = Class.inherit({
     getNodeByItem: function(item) {
         let result = null;
 
+        const itemKey = this.options.dataAccessors.getters.key(item);
         each(this._dataStructure, function(_, node) {
-            if(node.internalFields.item === item) {
+            const isNodeFound = itemKey
+                ? node.internalFields.key === itemKey
```
IgnatovDan: "itemKey ? node.internalFields.key === itemKey : node.internalFields.item === item" 

[null, undefined, false, 0] key values means that there is no key and "node.internalFields.item === item" should be used
Are you sure?

IgnatovDan (changes requested):

js/ui/hierarchical_collection/ui.data_adapter.js
novsstation: Yep, fixed.
Now, I check the key only if it's value doesn't equals the undefined value

IgnatovDan (approved):

PR #14155, opened 2020-08-07
Fix incorrect imports

egaluza:


js/core/renderer.d.ts
```diff
@@ -0,0 +1,256 @@
+interface JQueryCoordinates {
```
IlyaKhD: I suggest renaming all those `jQueryXXX` types to `jQueryLikeXXX` e.g.:
```suggestion
interface JQueryLikeCoordinates {
```

Maybe we should move all of them to a separate folder.

js/core/renderer.d.ts
```diff
@@ -0,0 +1,256 @@
...
+    
+    attr(attributeName: string, value: string|number|null): JQuery;
+    
+    attr(attributeName: string, func: (index: number, attr: string) => string|number): JQuery;
+    
+    hasClass(className: string): boolean;
+    
+    html(htmlString: string): JQuery;
+    
+    html(func: (index: number, oldhtml: string) => string): JQuery;
```
IlyaKhD: It seems that this overload is not implemented:

```js
initRender.prototype.html = function(value) {
    if(!arguments.length) {
        return this[0].innerHTML;
    }

    this.empty();

    if(typeof value === 'string' && !isTablePart(value) || typeof value === 'number') {
        this[0].innerHTML = value;

        return this;
    }

    return this.append(parseHTML(value));
};
```

IlyaKhD: I think we should check for unsupported declarations and remove them.
And see if it breaks compilation 🍿

js/core/renderer.d.ts
nightskylark: I'd prefer to avoid `jquery` word usage everywhere except the `integrations/jquery` folder.

egaluza: Leave part of the name XXX

js/core/renderer.d.ts
egaluza: Done

dxvladislavvolkov (dismissed):

js/core/renderer.d.ts
```diff
@@ -0,0 +1,133 @@
...
+    
+    wrapInner(wrappingElement: JQuery|Element|string): JQuery;
+    
+    each(func: (index: number, elem: Element) => boolean | void): JQuery;
+    
+    get(index: number): Element;
+
+    index(selector: string|JQuery|Element): number;
+    
+    add(...elements: Element[]): JQuery;
```
IlyaKhD: It seems that our implementation differs from jQuery:
```js
initRender.prototype.add = function(selector) {
    const targets = renderer(selector);
    const result = this.toArray();

    for(let i = 0; i < targets.length; i++) {
        const target = targets[i];
        if(result.indexOf(target) === -1) {
            result.push(target);
        }
    }

    return renderer(result);
};
```

nightskylark (approved):

PR #14219, opened 2020-08-13
Sortable: Fix cross-component dragging between Sortable with different nesting level (T918389)

Alyar666:
See https://devexpress.com/issue=T918389

### DIFF
--- a/js/events/drag.js
+++ b/js/events/drag.js
@@ -286,7 +286,7 @@ const DragEmitter = GestureEmitter.inherit({
             return false;
         }
 
-        if($prevTarget.length && $target.find($prevTarget).length) {
+        if($prevTarget.length && $prevTarget.closest($target).length) {
             return false;
         }
 
--- a/js/events/drag.js
+++ b/js/events/drag.js
@@ -244,7 +244,7 @@ const DragEmitter = GestureEmitter.inherit({
             const $target = $(target);
             iteratorUtils.each(getItemDelegatedTargets($target), function(_, delegatedTarget) {
                 const $delegatedTarget = $(delegatedTarget);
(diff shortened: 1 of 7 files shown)


js/events/drag.js
```diff
@@ -286,6 +286,10 @@ const DragEmitter = GestureEmitter.inherit({
             return false;
         }
 
+        if($prevTarget && $target.find($prevTarget).length) {
```
vconst: Use closest instead of find because it works faster.
`$prevTarget.closest($target).length`
Also $prevTarget always exists because it is passed as $(result). So null ref check can be removed.

testing/tests/DevExpress.ui.widgets.scheduler/integration.allDayAppointments.tests.js
```diff
@@ -405,7 +433,11 @@ QUnit.test('allDayExpanded option of workspace should be updated after dragged o
     const $appointment = $(this.instance.$element()).find('.dx-scheduler-appointment').eq(0);
 
     const pointer = pointerMock($appointment).start().down().move(10, 10);
-    $(this.instance.$element()).find('.dx-scheduler-date-table-cell').eq(0).trigger(dragEvents.enter);
+    const appointmentOffset = $appointment.offset();
+    $(this.instance.$element()).find('.dx-scheduler-date-table-cell').eq(0).trigger($.Event(dragEvents.enter, {
```
vconst: Write helper function for enter event triggering to avoid code duplication.

vconst (changes requested):

vconst (dismissed):

jtoming830 (dismissed):

vconst (approved):

PR #14235, opened 2020-08-14
Late binding DOM improves performance up to 88% for makeTable

jacobboland:
Profiling for my company's usage of the scheduler, I found that several seconds can be spent within the makeTable call in ui.scheduler.table_creator.js. The issues occur with the schedules of 10-12 people for Week view especially. Using console.time measurements within this method, I found that a customer with 10 users' schedules, including hundreds of appointments over Week View could spend 3.5 seconds in this method. After these changes, that measurement dropped to around 400ms. 

The premise is that the later an element is bound to the DOM, the better. Making updates to an element that is within the DOM, like appending cells to a row and rows to a table will recalculate multiple elements each time it is done. The more elements in the DOM, the worse this becomes. The change is to simply move appendChild calls to as late as possible without changing much code.

We have a fork and are starting to use this for 19.2. We look forward to upgrading to 20.2, soon, hence the attempt to get this incorporated into your next release if possible.

### DIFF
--- a/js/ui/calendar/ui.calendar.base_view.js
+++ b/js/ui/calendar/ui.calendar.base_view.js
@@ -97,7 +97,6 @@ const BaseView = Widget.inherit({
             for(let colIndex = 0, colCount = this.option('colCount'); colIndex < colCount; colIndex++) {
                 this._renderCell(rowData, colIndex);
             }
-            this.$body.get(0).appendChild(rowData.row);
         }
     },
 
@@ -105,6 +104,7 @@ const BaseView = Widget.inherit({
         const row = domAdapter.createElement('tr');
 
         this.setAria('role', 'row', $(row));
+        this.$body.get(0).appendChild(row);
 
         return row;
     },
--- a/js/ui/calendar/ui.calendar.base_view.js
+++ b/js/ui/calendar/ui.calendar.base_view.js
@@ -97,6 +97,7 @@ const BaseView = Widget.inherit({
             for(let colIndex = 0, colCount = this.option('colCount'); colIndex < colCount; colIndex++) {
                 this._renderCell(rowData, colIndex);
             }
+            this.$body.get(0).appendChild(rowData.row);
         }
     },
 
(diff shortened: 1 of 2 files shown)


andrewmakarov:
Hi @jacobboland ,
Thank you for the detailed explanation. The solution you suggested looks valid. However, the overall issue may be more complex depending on the configuration of the component.

For instance, we won't use this code (and this fix) in our upcoming [`Virtual Scrolling`](https://github.com/DevExpress/DevExtreme/issues/14285) mode. However, it is possible that you may decide to use this mode in v20.2, if you need to render a lot of appointments/resources.

Therefore, we'd like to test the case where you see the performance degradation to decide if we can merge of these changes. For this, please create a stand-alone example that demonstrates your configuration and the performance issue.

You can use one of our [demos](https://js.devexpress.com/Demos/WidgetsGallery/Demo/Scheduler/SimpleArray/jQuery/Light/) for this purpose.

 

* Press "Copy To …" to extract a demo into a stand-alone example.
* Modify the example and use the "Save" button to apply changes and generate a new link.
* Post the updated link here for further examination.

 


Thanks,
Andrew

jacobboland:
Hello @andrewmakarov, I'm sorry for the delay in getting back to you. [Here is a scenario](https://codesandbox.io/s/peaceful-noyce-gfjzp) that illustrates some of the issues we are experiencing with large customers of ours. It can take over 10 seconds in this scenario and that time is blocking. This is a bit simpler than our scenario, so we are working to make things a bit simpler on our side. We are looking forward to 20.2. Do you have an idea for when to expect that to be released?

AntonSermyazhko:
Hello,

To enhance the performance in v20.2, we will implement a [Virtual Scrolling feature](https://github.com/DevExpress/DevExtreme/issues/14285). However, it will only support vertical grouping. Is it possible for you to use vertical grouping? If no, I suggest you track the following ticket in our support center to automatically get a notification when we support horizontal grouping in Virtual Scrolling: [Scheduler - Virtual Scrolling does not work when horizontal grouping is enabled](https://supportcenter.devexpress.com/internal/ticket/details/t936668).

jacobboland:
We are going to test out the latest Beta to see how that will work for us. I will check with our product team about vertical grouping.

PR #14264, opened 2020-08-17
DataGrid/TreeList: Implement new editing events

jtoming830:


testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -9012,6 +9012,333 @@ QUnit.module('Editing with real dataController', {
...
+            rowsView.render($testElement);
+
+            // act
+            this.editRow(0);
+            this.cellValue(0, 0, 'new value');
+            this.saveEditData();
+
+            // assert
+            assert.ok($(this.getRowElement(0)).hasClass('dx-edit-row'), 'row is edited');
+            assert.equal(onSaving.callCount, 1, 'onSaving');
```
MikeVitik: Can you add some additional comment because if something goes wrong it's too difficult to understand which assert is fail

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -205,6 +205,7 @@ const EditingController = modules.ViewController.inherit((function() {
             const that = this;
 
             that._editData = [];
+            that._isEditingEndInProgress = {};
```
NickMitrokhin: I didn't find a place where this option is used.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1515,51 +1525,70 @@ const EditingController = modules.ViewController.inherit((function() {
...
-            const that = this;
-            const store = that._dataController.store();
-            let isDataSaved = true;
+            const onSavingParams = {
+                cancel: false,
+                promise: null
+            };
+            this.executeAction('onSaving', onSavingParams);
+            const d = new Deferred();
+            when(fromPromise(onSavingParams.promise))
```
NickMitrokhin, line 1533: **fromPromise** is not necessary. You can use the following code:
```js
when(onSavingParams.promise).done(...)
```
I'm not sure I understand why we need to use the **promise** parameter. Is it possible to keep only the **cancel** parameter that might be boolean or a promise that we can check directly after calling **onSaving**?

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1655,130 +1685,155 @@ const EditingController = modules.ViewController.inherit((function() {
...
-                        const rowIndex = this._dataController.getRowIndexByKey(key);
-                        if(rowIndex !== -1) {
-                            for(let columnIndex = 0; columnIndex < columnsCount; columnIndex++) {
-                                this._rowsView._getCellElement(rowIndex, columnIndex).removeClass(CELL_MODIFIED);
-                            }
+
+                if(deferreds.length) {
+                    dataSource?.beginLoading();
+
+                    when.apply($, deferreds).done(() => {
```
NickMitrokhin:
```suggestion
                    when(...deferreds).done(() => {
```

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1848,16 +1903,28 @@ const EditingController = modules.ViewController.inherit((function() {
...
         cancelEditData: function() {
-            const that = this;
-            const editMode = getEditMode(that);
+            const params = {
+                promise: null,
+                cancel: false
+            };
+
+            this.executeAction('onEditCanceling', params);
+            if(!params.cancel) {
```
NickMitrokhin, line 1909: It seems that checking **params.promise** is missing after calling **onEditCanceling**.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1515,51 +1525,70 @@ const EditingController = modules.ViewController.inherit((function() {
...
+                cancel: false,
+                promise: null
+            };
+            this.executeAction('onSaving', onSavingParams);
+            const d = new Deferred();
+            when(fromPromise(onSavingParams.promise))
+                .done(() => {
+                    if(!onSavingParams.cancel) {
+                        const isDataSaved = this._processEditData(deferreds, results, changes);
+                        d.resolve(isDataSaved);
```
NickMitrokhin: **_processEditData** has the following code where **isDataSaved** is calculated in the **deferred.always** callback:

```js
                    deferred
                        .always(function(data) {
                            isDataSaved = data !== 'cancel';
                            results.push({ key: editData.key, result: data });
                        })
                        .always(doneDeferred.resolve);
```
is it possible that **deferred.always** will be called after **_processEditData** return the result?

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: We discussed with @vconst and decided not to add `promise` to onEditCanceling. The only one scenario where `promise` in onEditCanceling is needed can be implemented via `cancel` and flag.

NickMitrokhin: OK, in this case the _promise_ option should be removed from the **params** object that is passed to **onEditingCancelling**.

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: Yes, it is possible, but 2 separate parameters are better because:
1. 1 parameter - 1 type, it is easier for TS and looks like _true_ static type system. 
2. Code like `e.cancel = new Promise((resolve) => { resolve(true) })` is not obvious. 
3. These parameters accomplish different tasks

We don't have many parameters in this callback so I think it's OK.

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: **deferred.always** is called after return in all cases. This is because of **setTimeout** call in the **_executeEditingAction** method.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1655,130 +1685,155 @@ const EditingController = modules.ViewController.inherit((function() {
...
+                        this._saving = false;
+                        deferred.resolve(result);
+                    }).fail(deferred.reject);
                 }).fail(deferred.reject);
             }).fail(deferred.reject);
             return deferred.promise();
         },
+
+        _resolveAfterSaveEditDataComplete: function(deferred, { cancel, error } = {}) {
+            when(this._afterSaveEditData(cancel)).done(function() {
```
vconst, line 1713: Add fail handler

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -9012,6 +9012,333 @@ QUnit.module('Editing with real dataController', {
...
+            assert.equal(onSaved.callCount, 0, 'onSaved');
+
+            // act
+            this.clock.tick(1000);
+
+            // assert
+            assert.ok($(this.getRowElement(0)).hasClass('dx-edit-row'), 'row is edited');
```
vconst, line 9402: Asserts are not changed after tick(1000). Why? I think we should check that load panel is hidden after this tick.

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: I have noticed that isDataSaved is useless, seems like it was not removed in one of old refactorings. 
I have removed it and condition for cell focus which was never executed.

MikeVitik (dismissed):

NickMitrokhin (dismissed):

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1655,130 +1680,151 @@ const EditingController = modules.ViewController.inherit((function() {
...
-                    afterSaveEditData();
+                if(this.isSaving()) {
+                    this._resolveAfterSaveEditDataComplete(deferred);
                     return;
                 }
                 when(this._beforeSaveEditData()).done(cancel => {
                     if(cancel) {
-                        afterSaveEditData(cancel);
+                        this._resolveAfterSaveEditDataComplete(deferred, { cancel });
+                        this._saving = false;
```
vconst: Why?

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1655,130 +1680,151 @@ const EditingController = modules.ViewController.inherit((function() {
...
+                }
+
                 dataController.updateItems();
             } else {
                 this._focusEditingCell();
             }
 
-            afterSaveEditData();
-            return result.promise();
+            this.executeAction('onSaved');
```
vconst, line 1810: Is onSaved event should be fired if saving is canceled?

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1655,130 +1680,151 @@ const EditingController = modules.ViewController.inherit((function() {
...
-            const afterSaveEditData = (error) => {
-                when(this._afterSaveEditData()).done(function() {
-                    result.resolve(error);
-                });
-            };
 
-            if(!this._saveEditDataCore(deferreds, results, changes) && editMode === EDIT_MODE_CELL) {
-                this._focusEditingCell();
-            }
+                    this.executeAction('onSaved');
```
vconst: Why is onSaved fired before all deferreds are completed?

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1515,51 +1524,69 @@ const EditingController = modules.ViewController.inherit((function() {
...
+            const d = new Deferred();
+            when(fromPromise(onSavingParams.promise))
+                .done(() => {
+                    if(!onSavingParams.cancel) {
+                        this._processEditData(deferreds, results, changes);
+                        d.resolve();
+                    } else {
+                        d.reject();
+                    }
+                });
```
vconst, line 1543: The fail is not processed. d should be rejected and error row should be shown.

vconst: .fail(createFailureHandler(deferred))

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1655,130 +1680,151 @@ const EditingController = modules.ViewController.inherit((function() {
...
-            const resetModifiedClassCells = () => {
-                if(editMode === EDIT_MODE_BATCH) {
-                    const columnsCount = this._columnsController.getVisibleColumns().length;
-                    editData.forEach(({ key }) => {
-                        const rowIndex = this._dataController.getRowIndexByKey(key);
-                        if(rowIndex !== -1) {
-                            for(let columnIndex = 0; columnIndex < columnsCount; columnIndex++) {
-                                this._rowsView._getCellElement(rowIndex, columnIndex).removeClass(CELL_MODIFIED);
-                            }
+            when(this._saveEditDataCore(deferreds, results, changes)).done(() => {
```
vconst: fail is not processed

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1655,130 +1680,151 @@ const EditingController = modules.ViewController.inherit((function() {
...
+                    return result.always(() => {
+                        this._focusEditingCell();
+                        this._saving = false;
+                    }).promise();
+                }
 
-            if(deferreds.length) {
-                this._saving = true;
+                this._cancelSaving();
+                this._resolveAfterSaveEditDataComplete(result);
```
vconst: I think that _resolveAfterSaveEditDataComplete should be called inside _cancelSaving similar to _endSaving.

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -9012,6 +9012,336 @@ QUnit.module('Editing with real dataController', {
...
+            this.options.onSaving = onSaving;
+            this.options.onSaved = onSaved;
+            this.editingController.optionChanged({ name: 'onSaving' });
+            this.editingController.optionChanged({ name: 'onSaved' });
+            rowsView.render($testElement);
+
+            // act
+            this.editRow(0);
+            this.cellValue(0, 0, 'new value');
+            this.saveEditData();
```
vconst, line 9131: saveEditData result should be tested

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -9012,6 +9012,336 @@ QUnit.module('Editing with real dataController', {
...
+        });
+
+        QUnit.test('Promise in onSaving', function(assert) {
+            // arrange
+            const rowsView = this.rowsView;
+            const $testElement = $('#container');
+            const onSaving = sinon.spy(e => {
+                e.promise = new Deferred();
+                setTimeout(() => {
+                    e.promise.resolve();
```
vconst, line 9176: There is no test with e.promise.reject("Blablabla")

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -9012,6 +9012,336 @@ QUnit.module('Editing with real dataController', {
...
+            this.editingController.optionChanged({ name: 'onSaving' });
+            this.editingController.optionChanged({ name: 'onSaved' });
+            this.editingController.optionChanged({ name: 'onEditCanceling' });
+            this.editingController.optionChanged({ name: 'onEditCanceled' });
+            rowsView.render($testElement);
+
+            // act
+            this.editRow(0);
+            this.cellValue(0, 0, 'new value');
+            this.saveEditData();
```
vconst: Result should be tested

js/ui/grid_core/ui.grid_core.editing.js
vconst: I think that _saving = true and _saving = false should be in one place. But now it is changed in different places.

vconst (changes requested):

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: No.

vconst (approved):

NickMitrokhin (approved):

MikeVitik (approved):

PR #14304, opened 2020-08-19
TagBox: space entering should increase input element size (T923429)

ksercs:


js/ui/tag_box.js
```diff
@@ -680,6 +680,7 @@ const TagBox = SelectBox.inherit({
         if(isEmptyInput && canTypeText) {
             const $calculationElement = createTextElementHiddenCopy($input, value, { includePaddings: true });
 
+            $calculationElement.css('whiteSpace', 'pre');
```
DokaRus: Are there any limitations to doing this by default for the `createTextElementHiddenCopy` utility?

ksercs: I fixed, but it's strange to me that `whiteSpace` was defined in element styles as `nowrap`. What is the likelihood that this definition just appeared there without any reason?

DokaRus (approved):

alexander-kotov-dx (approved):

PR #14350, opened 2020-08-24
Toast: widget should have correct position on mobile devices with pitch zoom (T924939)

ksercs:


testing/tests/DevExpress.animation/position.tests.js
```diff
@@ -940,20 +941,25 @@ const testCollision = (name, fixtureName, params, expectedHorzDist, expectedVert
         try {
             window.innerHeight = 500;
             window.outerHeight = 200;
+            window.visualViewport = { height: 1000 };
```
DokaRus: Should we set an initial object at the `finally` block?

testing/tests/DevExpress.animation/position.tests.js
```diff
@@ -967,20 +973,25 @@ const testCollision = (name, fixtureName, params, expectedHorzDist, expectedVert
         try {
             window.innerWidth = 500;
             window.outerWidth = 200;
+            window.visualViewport = { width: 1000 };
```
DokaRus: And there

js/ui/toast.js
```diff
@@ -146,13 +146,10 @@ const Toast = Overlay.inherit({
             },
             {
                 device: function(device) {
-                    const isPhone = device.deviceType === 'phone';
-                    const isAndroid = device.platform === 'android';
```
DokaRus: Is there a reason to remove it? Please note - the new position config will be applied to all phones.

ksercs: I tested it in iphone and android. Should we specify only these platforms?

testing/tests/DevExpress.animation/position.tests.js
ksercs: yeah, thx

alexander-kotov-dx (dismissed):

js/ui/toast.js
ksercs: fixed

DokaRus (approved):

alexander-kotov-dx (approved):

PR #14392, opened 2020-08-27
List: it should be possible to subscribe on `itemSwipe` event using `on` method (T926077)

ksercs:


DokaRus (dismissed):

DokaRus (dismissed):

js/ui/list/ui.list.base.js
```diff
@@ -892,6 +901,9 @@ export const ListBase = CollectionWidget.inherit({
             case 'badgeExpr':
                 this._invalidate();
                 break;
+            case '_swipeEnabled':
+                this._toggleSwipeEvent(args.value);
```
DokaRus: Let's test it?

ksercs: fixed

DokaRus (approved):

alexander-kotov-dx (approved):

PR #14438, opened 2020-09-03
DataGrid PDF Export - Grouping

KuznetsovVN:


EugeniyKiyashko (dismissed):

EugeniyKiyashko (dismissed):

js/exporter/jspdf/export.js
```diff
@@ -108,37 +114,47 @@ export const Export = {
...
+        const setStyle = (styleName, value) => {
+            if(isDefined(value)) {
+                styles[styleName] = value;
+            }
+        };
+        if(rowType === 'header') {
+            setStyle('halign', columnAlignment);
+        } else {
+            setStyle('halign', horizontalAlignment);
+            setStyle('fontStyle', bold ? 'bold' : undefined);
```
jsdmitry: Why do you use a condition? If there is a check on the undefined value inside the setStyle method.

KuznetsovVN: this block of code is rewritten

EugeniyKiyashko (approved):

jsdmitry (approved):

PR #14519, opened 2020-09-08
Scheduler - Add method for getting timezone list

Yarovaya:


AntonSermyazhko (dismissed):

AryamnovEugeniy (dismissed):

js/ui/scheduler.d.ts
```diff
@@ -804,6 +804,21 @@ interface JQuery {
     dxScheduler(options: dxSchedulerOptions): JQuery;
 }
 }
+
+/**
+ * @docid ui.dxScheduler.getTimeZones
+ * @publicName getTimeZones(date)
+ * @param1 date:Date|undefined
+ * @return Array<Object>
```
RomanTsukanov, line 812: This object has a defined structure. Maybe we should create a type for it?

pavel-2020 (approved):

andrewmakarov (approved):

js/ui/scheduler.d.ts
RomanTsukanov: Turns our we can't do this if the method is static.

RomanTsukanov (approved):

PR #14547, opened 2020-09-10
Add templates to axes

alexanderPolosatov:
cc @AlexJameson


js/viz/axes/base_axis.js
```diff
@@ -1959,28 +1972,66 @@ Axis.prototype = {
...
+        offset = that._adjustDateMarkers(offset);
+        that._adjustTitle(offset);
+        that.__rendered = true;
+    },
+
+    getTemplatesDef() {
+        return this._templatesRendered;
+    },
+
+    drawn(state) {
```
dxBeardedBear: You must explicitly return the value

js/viz/axes/base_axis.js
```diff
@@ -149,7 +150,8 @@ function updateGridsPosition(ticks, animate) {
 }
 export const measureLabels = function(items) {
     items.forEach(function(item) {
-        item.labelBBox = item.label ? item.label.getBBox() : { x: 0, y: 0, width: 0, height: 0 };
+        const label = item.templateContainer || item.label;
```
dxBeardedBear: Better to make a method from "tick.templateContainer || tick.label"

dxBeardedBear (approved):

PR #14555, opened 2020-09-10
Adds applyChanges method

NickMitrokhin:
Added the **applyChanges** method to the **DevExtreme.data** namespace.


js/data/array_utils.js
```diff
@@ -78,21 +79,36 @@ function setDataByKeyMapValue(array, key, data) {
     }
 }
 
-function applyBatch(keyInfo, array, batchData, groupCount, useInsertIndex) {
+function applyBatch(keyInfo, array, batchData, groupCount, useInsertIndex, immutable) {
+    const isDataImmutable = isDefined(immutable) && immutable === true;
```
vconst: Why is not `const isDataImmutable = immutable === true;`?

vconst: Also, we can pass immutable into update without normalization.

js/data/array_utils.js
```diff
@@ -108,7 +124,13 @@ function update(keyInfo, array, key, data, isBatch) {
             if(index < 0) {
                 return !isBatch && rejectedPromise(errors.Error('E4009'));
             }
+
             target = array[index];
+
+            if(immutable === true && isDefined(target) && isPlainObject(target)) {
```
vconst: Why is isPlainObject? What is about class objects? Sometimes users use classes instead of plaint objects in DataGrid. 
I think we should use the following utils method: https://github.com/DevExpress/DevExtreme/blob/42ce857c16b4c64ca19cd7042b132ce8fb247eca/js/ui/grid_core/ui.grid_core.utils.js#L412.

vconst (changes requested):

js/data/array_utils.js
```diff
@@ -78,21 +79,36 @@ function setDataByKeyMapValue(array, key, data) {
...
+            case 'update': update(keyInfo, items, item.key, item.data, true, isDataImmutable); break;
             case 'insert': insert(keyInfo, items, item.data, useInsertIndex && isDefined(item.index) ? item.index : -1, true); break;
             case 'remove': remove(keyInfo, items, item.key, true); break;
         }
     });
+    return resultItems;
+}
+
+function applyChanges(data, changes, options) {
+    const { keyExpr = 'id', immutable = true } = isDefined(options) && isPlainObject(options) ? options : {};
```
jtoming830: You can define default value for the options and remove all these checks. 
```
function applyChanges(data, changes, options = {}) {
    const { keyExpr = 'id', immutable = true } = options;
```

js/data/array_utils.js
```diff
@@ -78,21 +79,36 @@ function setDataByKeyMapValue(array, key, data) {
...
+function applyBatch(keyInfo, array, batchData, groupCount, useInsertIndex, immutable) {
+    const isDataImmutable = isDefined(immutable) && immutable === true;
+    const resultItems = isDataImmutable ? [...array] : array;
+
     batchData.forEach(item => {
-        const items = item.type === 'insert' ? array : getItems(keyInfo, array, item.key, groupCount);
+        const items = item.type === 'insert' ? resultItems : getItems(keyInfo, resultItems, item.key, groupCount);
 
-        generateDataByKeyMap(keyInfo, items);
+        !isDefined(immutable) && generateDataByKeyMap(keyInfo, items);
```
jtoming830: Why !isDefined here?

testing/tests/DevExpress.data/applyChanges.tests.js
```diff
@@ -0,0 +1,110 @@
+import { extend } from 'core/utils/extend';
+import applyChanges from 'data/apply_changes';
+
+QUnit.module('Apply Changes', {
```
jtoming830, line 4: Why no test for the case when keyExpr is not defined and no 'id' field in objects?

jtoming830 (changes requested):

js/data/apply_changes.js
```diff
@@ -0,0 +1,3 @@
+import { applyChanges } from './array_utils';
```
jtoming830, line 1: Why don't you isolate the applyChanges and applyBatch functions in this file instead of placing them in array_utils?

jtoming830: With current exports we have two internal sources of applyChanges.

NickMitrokhin: > Why don't you isolate the applyChanges and applyBatch functions in this file instead of placing them in array_utils?


I do not think that the applyBatch method should be moved to apply_changes. The method is intended to arrays and should be placed in array_utils as it is. 

> With current exports we have two internal sources of applyChanges.

We have many places not only in grid where we re-export API and there is no problem with it.

js/data/array_utils.js
NickMitrokhin: applyBatch is called not only by applyChanges. However, only applyChanges can pass the immutable parameter. So, when this parameter is defined this means that it is passed by applyChanges, we should not add any extra info to items.

js/data/array_utils.js
NickMitrokhin: makes sense:)

testing/tests/DevExpress.data/applyChanges.tests.js
NickMitrokhin: I'll implement it as a separate task. We need to discuss how we will notify a user. Moreover, applyBatch is used by our data stores and they do not throw errors in the console un such cases.

js/data/array_utils.js
NickMitrokhin: OK.

jtoming830 (dismissed):

MikeVitik (dismissed):

js/data/apply_changes.d.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * @docid Utils.applyChanges
+ * @publicName applyChanges(data, changes, options)
+ * @param1 data:Array<any>
+ * @param2 changes:Array<any>
+ * @param3 options?:any
```
vconst, line 6: describe parameters of options

js/data/array_utils.js
```diff
@@ -87,14 +87,14 @@ function createObjectWithChanges(target, changes) {
     return objectUtils.deepExtendArraySafe(result, changes, true, true);
 }
 
-function applyBatch(keyInfo, array, batchData, groupCount, useInsertIndex, immutable) {
+function applyBatch({ keyInfo, array, batchData, groupCount, useInsertIndex, immutable, enableCache }) {
     const isDataImmutable = immutable === true;
```
jtoming830: This variable seems useless.

js/data/array_utils.js
```diff
@@ -78,21 +79,50 @@ function setDataByKeyMapValue(array, key, data) {
...
-function applyBatch(keyInfo, array, batchData, groupCount, useInsertIndex) {
+function createObjectWithChanges(target, changes) {
+    const result = target ? Object.create(Object.getPrototypeOf(target)) : {};
+    const targetWithoutPrototype = extendFromObject({}, target);
+
+    objectUtils.deepExtendArraySafe(result, targetWithoutPrototype, true, true);
+    return objectUtils.deepExtendArraySafe(result, changes, true, true);
+}
+
+function applyBatch({ keyInfo, array, batchData, groupCount, useInsertIndex, immutable, enableCache }) {
```
vconst: Rename array to data, batchData to changes to be consistent with applyChanges.

jtoming830 (changes requested):

vconst (changes requested):

MikeVitik (dismissed):

js/data/apply_changes.d.ts
NickMitrokhin: ok

js/data/array_utils.js
NickMitrokhin: ok

MikeVitik (approved):

Alyar666 (approved):

jtoming830 (approved):

vconst (approved):

PR #14663, opened 2020-09-22
DataGrid: Fix empty changes after editing start

jtoming830:

### DIFF
--- a/js/ui/grid_core/ui.grid_core.validating.js
+++ b/js/ui/grid_core/ui.grid_core.validating.js
@@ -357,16 +357,16 @@ const ValidatingController = modules.Controller.inherit((function() {
             if(editIndex < 0) {
                 if(!showEditorAlways) {
                     const columnsController = this.getController('columns');
-                    const visibleColumns = columnsController && columnsController.getVisibleColumns() || [];
+                    const visibleColumns = columnsController?.getVisibleColumns() || [];
                     showEditorAlways = visibleColumns.some(function(column) { return column.showEditorAlways; });
-                }
-
-                if(showEditorAlways && editingController.isCellOrBatchEditMode() && editingController.allowUpdating({ row: parameters.row })) {
-                    editIndex = editingController._addEditData({ key: parameters.key, oldData: parameters.data });
                 }
 
                 if(this.option('editing.editRowKey') === parameters.key) {
                     editIndex = 0;
+                }
+
+                if(showEditorAlways && editingController.isCellOrBatchEditMode() && editingController.allowUpdating({ row: parameters.row })) {
+                    editIndex = editingController._addEditData({ key: parameters.key, oldData: parameters.data });
                 }
             }
 
--- a/js/ui/grid_core/ui.grid_core.validating.js
+++ b/js/ui/grid_core/ui.grid_core.validating.js
@@ -344,7 +344,6 @@ const ValidatingController = modules.Controller.inherit((function() {
         },
 
         createValidator: function(parameters, $container) {
-            let editData;
             let editIndex;
             const editingController = this._editingController;
             const column = parameters.column;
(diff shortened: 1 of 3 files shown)


js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -362,6 +364,10 @@ const ValidatingController = modules.Controller.inherit((function() {
                 if(showEditorAlways && editingController.isCellOrBatchEditMode() && editingController.allowUpdating({ row: parameters.row })) {
                     editIndex = editingController._addEditData({ key: parameters.key, oldData: parameters.data });
                 }
+
+                if(this.option('editing.editRowKey') === parameters.key) {
```
```suggestion

                if(editindex < 0 && this.option('editing.editRowKey') === parameters.key) {
```

**_let editData;_** should be removed from **createValidator**. It is better to put **editData** declarations directly to **getValue** and **dataGetter** callbacks.

```js
const getValue = () => {
    const editData = editingController.getEditDataByKey(validationData?.key);
    ...
}
...
dataGetter: function() {
    const editData = editingController.getEditDataByKey(validationData?.key);
     ...
},
```

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -9450,6 +9450,90 @@ QUnit.module('Editing with real dataController', {
...
+                    'name': 'Alex',
+                    'phone': '555555',
+                    'room': 1,
+                    'state': {
+                        'name': 'state 1'
+                    },
+                    'stateId': 0
+                },
+                'type': 'update'
+            }], 'no changes');
```
NickMitrokhin: It is not clear what you mean by 'no changes'.

NickMitrokhin (changes requested):

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -9450,6 +9450,90 @@ QUnit.module('Editing with real dataController', {
...
```
NickMitrokhin: It is not clear what you mean by 'no changes'.

js/ui/grid_core/ui.grid_core.validating.js
jtoming830: >  if(editindex < 0 && this.option('editing.editRowKey') === parameters.key) {

We don't need this because `editIndex < 0` condition is met on the 357 line.

jtoming830: > let editData; should be removed from createValidator.

Ok.

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
jtoming830: Copypaste 👍

testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
jtoming830: Copypaste again 👍

js/ui/grid_core/ui.grid_core.validating.js
NickMitrokhin: >We don't need this because editIndex < 0 condition is met on the 357 line.

You are right, but it may confusing if the following code will set editIndex to a value that is greater or equal to 0:
```js
if(showEditorAlways && editingController.isCellOrBatchEditMode() && editingController.allowUpdating({ row: parameters.row })) {
    editIndex = editingController._addEditData({ key: parameters.key, oldData: parameters.data });
}
```
and then, reset editIndex to 0 under the following condition:

```js
if(this.option('editing.editRowKey') === parameters.key) {
    editIndex = 0;
}
```

jtoming830: Ok, I will move index reset before this condition.

NickMitrokhin (approved):

vconst (approved):

PR #14723, opened 2020-09-28
Update German and Japanese localization

DokaRus:


alexander-kotov-dx (approved):

js/localization/messages/de.json
```diff
@@ -57,7 +57,7 @@
         "dxFileUploader-upload": "Hochladen",
         "dxFileUploader-uploaded": "Hochgeladen",
         "dxFileUploader-readyToUpload": "Bereit zum hochladen",
-        "dxFileUploader-uploadAbortedMessage": "TODO",
+        "dxFileUploader-uploadAbortedMessage": "Upload abbrechen",
```
San4es, line 60: It looks like "Upload abbrechen" in German means "Cancel upload" instead of "Upload canceled":
https://github.com/search?q=Upload+abbrechen&type=code
https://github.com/search?q=Upload+abgebrochen&type=code

Oliver @oliversturm, can you please clarify the correct way?

```suggestion
        "dxFileUploader-uploadAbortedMessage": "Upload abgebrochen",
```

oliversturm: You're exactly right. "Upload abgebrochen" should be used for "Upload cancelled".

DokaRus: Roger that https://github.com/DevExpress/DevExtreme/pull/14732

PR #14736, opened 2020-09-28
applyBatch - Adds the logError parameter

NickMitrokhin:
Errors are displayed in the browser console only when the logErrors parameter is set to true.

### DIFF
--- a/js/data/array_utils.js
+++ b/js/data/array_utils.js
@@ -105,7 +105,7 @@ function applyBatch({ keyInfo, data, changes, groupCount, useInsertIndex, immuta
 }
 
 function getErrorResult(isBatch, logError, errorCode) {
-    return !isBatch ? rejectedPromise(errors.Error(errorCode)) : logError === true && errors.log(errorCode);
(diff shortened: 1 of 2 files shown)


js/data/array_utils.js
```diff
@@ -96,16 +96,16 @@ function applyBatch({ keyInfo, data, changes, groupCount, useInsertIndex, immuta
...
+            case 'remove': remove(keyInfo, items, item.key, true, logError); break;
         }
     });
     return resultItems;
 }
 
-function getErrorResult(isBatch, errorCode) {
-    return !isBatch ? rejectedPromise(errors.Error(errorCode)) : errors.log(errorCode);
+function getErrorResult(isBatch, logError, errorCode) {
+    return !isBatch ? rejectedPromise(errors.Error(errorCode)) : logError === true && errors.log(errorCode);
```
jtoming830: why `logError === true` and not simple `logError`?

NickMitrokhin: Will change to logError:)

vconst (approved):

PR #14740, opened 2020-09-28
Drawer - Avoid drawer content flicks/repositioning if it contains heavy markup (T856247)

IgnatovDan:


IgnatovDan:
The current solution is complex enough. We review new requests and then find other solutions.

PR #14785, opened 2020-10-01
dxScheduler - implement scrollToTime for virtual scrolling, implement scrollTo

AryamnovEugeniy:


js/ui/scheduler/workspaces/view_data_provider.js
```diff
@@ -406,4 +406,53 @@ export default class ViewDataProvider {
...
+                                ? rowIndex - 1
+                                : rowIndex,
+                        },
+                        cellData,
+                    };
+                }
+            }
+        }
+
+        return undefined;
```
AntonSermyazhko: Do we realy need to return 'undefined' ?

AntonSermyazhko (dismissed):

js/ui/scheduler/workspaces/view_data_provider.js
AryamnovEugeniy: Yes. It may return undefined if the cell is not in the viewport - in case we select a lot fo cells in several days. And, as it seems to me, it is better to reflect this in code as `return undefined`

andrewmakarov (dismissed):

js/ui/scheduler/appointmentPopup.js
```diff
@@ -27,6 +27,8 @@ const APPOINTMENT_POPUP_WIDTH_MOBILE = 350;
 const TOOLBAR_ITEM_AFTER_LOCATION = 'after';
 const TOOLBAR_ITEM_BEFORE_LOCATION = 'before';
 
+const DAY_IN_MS = 24 * 3600000;
```
AntonSermyazhko: https://github.com/DevExpress/DevExtreme/blob/4ba417bdca9f39243c0d94e9023a24dbaf3d6a7a/js/ui/scheduler/workspaces/ui.scheduler.timeline_work_week.js#L42

`toMs('day')`

AryamnovEugeniy: Done

AntonSermyazhko (approved):

PR #14787, opened 2020-10-01
PDF Export DataGrid - add doс comments for arguments of the 'exportDataGrid' function

EugeniyKiyashko:


js/exporter/jspdf/export.js
```diff
@@ -110,7 +110,7 @@ export const Export = {
                 jsPDFDocument.autoTable.__autoTableOptions = autoTableOptions;
                 ///#ENDDEBUG
 
-                resolve(jsPDFDocument);
+                resolve(jsPDFDocument); // Do we need to return the Promise with jsPDFDocument param?
```
IgnatovDan: +1
It's in input args and should not be passed as the only result of our primise:

```
    export: function(options) {
        const {
            jsPDFDocument,
```
Change to "resolve()"?

js/pdf_exporter.d.ts
```diff
@@ -0,0 +1,67 @@
+import dxDataGrid from './ui/data_grid';
+
+/**
+* @docid PdfExportBaseProps
+* @namespace DevExpress.pdfExporter
+* @type object
+* @hidden
+*/
+export interface PdfExportBaseProps {
```
IgnatovDan: imho we should not split interface into Base and DataGrid props.
It's easy to split it when we will add Pivot props

ts/dx.all.d.ts
```diff
@@ -2975,6 +2975,42 @@ declare module DevExpress.fx {
      */
     export function stop(element: Element, jumpToEnd: boolean): void;
 }
+declare module DevExpress.pdfExporter {
+    /**
+     * [descr:PdfExportBaseProps]
+     */
+    export interface PdfExportBaseProps {
```
IgnatovDan: Another excessive split to Base and Grid props.

IgnatovDan (changes requested):

js/exporter/jspdf/export.js
EugeniyKiyashko: Yes. I also think that we must to call resolve() with empty value. Fixed in [b50f5c4](https://github.com/DevExpress/DevExtreme/pull/14787/commits/b50f5c4d3384b10050d24ca9d14d4e1295198079)

js/pdf_exporter.d.ts
EugeniyKiyashko: I fully agree with you. We don't want to get ahead of ourselves here. I merged two interfaces into one. https://github.com/DevExpress/DevExtreme/pull/14787/commits/27a4e5e7240e73d918f2db63bf62930dec56dfa3

ts/dx.all.d.ts
EugeniyKiyashko: It is autogenerated part of the merge after update-ts was called.

IgnatovDan (approved):

jsdmitry (approved):

PR #14958, opened 2020-10-16
DataGrid / TreeList: Fix focus of inserted row after scrolling if scrolling mode is virtual (T939311)

Alyar666:
See https://devexpress.com/issue=T939311

### DIFF
--- a/js/ui/grid_core/ui.grid_core.focus.js
+++ b/js/ui/grid_core/ui.grid_core.focus.js
@@ -423,9 +423,13 @@ module.exports = {
                 },
 
                 setFocusedRowIndex: function(rowIndex) {
+                    const dataController = this.getController('data');
+
                     this.callBase(rowIndex);
 
-                    const visibleRow = this.getController('data').getVisibleRows()[rowIndex];
+                    const visibleRowIndex = rowIndex - dataController.getRowIndexOffset();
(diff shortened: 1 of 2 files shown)


NickMitrokhin (dismissed):

js/ui/grid_core/ui.grid_core.focus.js
```diff
@@ -423,9 +423,13 @@ module.exports = {
                 },
 
                 setFocusedRowIndex: function(rowIndex) {
+                    const dataController = this.getController('data');
+
                     this.callBase(rowIndex);
 
-                    const visibleRow = this.getController('data').getVisibleRows()[rowIndex];
+                    const visibleRowIndex = rowIndex - dataController?.getRowIndexOffset();
```
vconst: Why is "?." operator here?
If dataController is null then visibleRowIndex will be NaN.

js/ui/grid_core/ui.grid_core.focus.js
```diff
@@ -423,9 +423,13 @@ module.exports = {
...
+                    const visibleRow = dataController?.getVisibleRows()[visibleRowIndex];
```
vconst: And here

NickMitrokhin (approved):

vconst (approved):

PR #14962, opened 2020-10-16
DataGrid: Fix validation frame is not updated if the repaintChangesOnly option is enabled in Cell editing mode (T914600)

jtoming830:


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -2649,6 +2650,8 @@ const EditingController = modules.ViewController.inherit((function() {
 
         _beforeCancelEditData: function() { },
 
+        _afterCancelEditData: function() { },
+
```
NickMitrokhin: Why this method is needed? Maybe it is better to put this._internalState = []; into it?

js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -1289,10 +1290,13 @@ export default {
                         rowKey: oldRow.key,
                         columnIndex
                     });
+                    const validationData = validatingController._getValidationData(oldRow.key);
                     const newValidationStatus = validationResultIsValid(validationResult) ? validationResult.status : validationResult;
                     const rowIsModified = JSON.stringify(newRow.modifiedValues) !== JSON.stringify(oldRow.modifiedValues);
                     if(oldValidationStatus !== newValidationStatus && rowIsModified) {
                         return true;
+                    } else if(validationData.isValid) {
+                        $(cell?.cellElement).toggleClass(this.addWidgetPrefix(INVALIDATE_CLASS), false);
```
NickMitrokhin: I think that the isCellChanged method should not change anything. 
```suggestion
                    const cellIsMarkedAsInvalid = $(cell?.cellElement).hasClass(this.addWidgetPrefix(INVALIDATE_CLASS));
                    if((oldValidationStatus !== newValidationStatus && rowIsModified) || (validationData.isValid && cellIsMarkedAsInvalid)) {
                        return true;
                    }
```

NickMitrokhin (changes requested):

NickMitrokhin (approved):

vconst (approved):

PR #15010, opened 2020-10-21
Animation position utils should take scrollbar width into account (T939748)

ksercs:
…

### DIFF
--- a/testing/tests/DevExpress.animation/position.tests.js
+++ b/testing/tests/DevExpress.animation/position.tests.js
@@ -966,12 +966,8 @@ const testCollision = (name, fixtureName, params, expectedHorzDist, expectedVert
         const $what = $('#what').height(300);
         const initialInnerHeight = window.innerHeight;
         const initialOuterHeight = window.outerHeight;
-        let heightStub;
-        if(QUnit.urlParams['nojquery']) {
-            heightStub = sinon.stub(renderer.fn, 'height').returns(1000);
-        } else {
-            heightStub = sinon.stub($.prototype, 'height').returns(1000);
-        }
+
+        const heightStub = sinon.stub(renderer.fn, 'height').returns(1000);
 
         try {
             window.innerHeight = 500;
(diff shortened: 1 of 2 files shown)


testing/tests/DevExpress.animation/position.tests.js
```diff
@@ -955,6 +955,34 @@ const testCollision = (name, fixtureName, params, expectedHorzDist, expectedVert
         }
     });
 
+    QUnit.test('position should return window.height() if window.outerHeight == window.innerHeight (T939748)', function(assert) {
+        const isPhone = devices.real().deviceType === 'phone';
+        if(isPhone || browser.safari) {
+            assert.expect(0);
```
DokaRus:
```suggestion
            assert.ok(true, 'tell me why');
```

testing/tests/DevExpress.animation/position.tests.js
```diff
@@ -955,6 +955,34 @@ const testCollision = (name, fixtureName, params, expectedHorzDist, expectedVert
         }
     });
 
+    QUnit.test('position should return window.height() if window.outerHeight == window.innerHeight (T939748)', function(assert) {
+        const isPhone = devices.real().deviceType === 'phone';
```
DokaRus, line 960: Is it actual only for phones? Are the tablets OK?

testing/tests/DevExpress.animation/position.tests.js
ksercs: fixed

testing/tests/DevExpress.animation/position.tests.js
ksercs: There is the same behavior on tablets as on desktops

DokaRus (dismissed):

San4es (dismissed):

alexander-kotov-dx (dismissed):

testing/tests/DevExpress.animation/position.tests.js
```diff
@@ -955,6 +956,64 @@ const testCollision = (name, fixtureName, params, expectedHorzDist, expectedVert
...
+        }
+
+        const $what = $('#what').height(300);
+        const initialInnerHeight = window.innerHeight;
+        const initialOuterHeight = window.outerHeight;
+
+        const heightStub = sinon.stub(renderer.fn, 'height').returns(1000);
+
+        try {
+            window.innerHeight = 500;
```
nightskylark, line 973: I think it's better to use a mock for window instead of modifying a real one.

ksercs: How can we do it? `innerHeight` is property, not method, so we can not use `sinon.stub`

San4es (dismissed):

alexander-kotov-dx (approved):

PR #15025, opened 2020-10-22
Fix dropDownEditors height and dropDownOptions.height issues

ksercs:
1. Added dropDownEditor popup repaint after `height` option runtime change for popup position recalculating
2. Made possible to customize dateBox (with list) popup height using dropDownOptions
3. Added tests for all widgets


alexander-kotov-dx (approved):

San4es (approved):
LGTM

PR #15027, opened 2020-10-23
dxScrollable - Inconsistent native scroll behavior in RTL mode (T936630)

EugeniyKiyashko:


EugeniyKiyashko:
[T936630 DataGrid - Column headers are shifted if rtlEnabled is true and native scrolling is used](https://trello.com/c/j5Ldahkn/1329-t936630-datagrid-column-headers-are-shifted-if-rtlenabled-is-true-and-native-scrolling-is-used)

EugeniyKiyashko:
We transfered this solution to 21.1 branch only - **https://github.com/DevExpress/DevExtreme/pull/15649**

PR #15043, opened 2020-10-26
Removes the KeyboardNavigationController.getFocusedColumnIndex method

NickMitrokhin:
Replaced the getFocusedColumnIndex method with getColumnIndex.

### DIFF
--- a/testing/tests/DevExpress.ui.widgets.dataGrid/focus.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.dataGrid/focus.tests.js
@@ -530,7 +530,7 @@ QUnit.module('Focused row', getModuleConfig(true), () => {
         const keyboardController = this.getController('keyboardNavigation');
         assert.ok(keyboardController.isCellFocusType(), 'Cell focus type');
         assert.equal(keyboardController.getVisibleRowIndex(), 0, 'Focused row index');
-        assert.equal(keyboardController.getFocusedColumnIndex(), 1, 'Focused column index');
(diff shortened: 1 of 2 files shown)


testing/tests/DevExpress.ui.widgets.dataGrid/focus.tests.js
```diff
@@ -530,7 +530,7 @@ QUnit.module('Focused row', getModuleConfig(true), () => {
         const keyboardController = this.getController('keyboardNavigation');
         assert.ok(keyboardController.isCellFocusType(), 'Cell focus type');
         assert.equal(keyboardController.getVisibleRowIndex(), 0, 'Focused row index');
-        assert.equal(keyboardController.getFocusedColumnIndex(), 1, 'Focused column index');
+        assert.equal(keyboardController.getColumnIndex(), 1, 'Focused column index');
```
jtoming830, line 533: Why not `keyboardController.option('focusedColumnIndex')`?

NickMitrokhin: **focusedColumnIndex** option may not have the same value as the **_focusedCellPosition.columnIndex** option that is used under the hood of the getColumnIndex method. We have plans to remove the **_focusedCellPosition** object at all and use the **focusedColumnIndex** option instead. But this is a separate task and I do not think that we should change anything regarding this in the context of these simple changes.

jtoming830 (approved):

vconst (approved):

PR #15045, opened 2020-10-26
Form - Label alignment of SimpleItems is broken when a GroupItem is included (T942517)

novsstation:
Bugfix T942324

### DIFF
--- a/js/ui/form/ui.form.js
+++ b/js/ui/form/ui.form.js
@@ -868,6 +868,7 @@ const Form = Widget.inherit({
                     this._invalidate();
                 }
                 break;
+            case 'alignRootItemLabels':
             case 'readOnly':
                 break;
             case 'width':
--- a/js/ui/form/ui.form.js
+++ b/js/ui/form/ui.form.js
@@ -38,26 +38,26 @@ import {
 import '../validation_summary';
 import '../validation_group';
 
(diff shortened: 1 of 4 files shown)


js/ui/form/ui.form.js
```diff
@@ -364,6 +367,13 @@ const Form = Widget.inherit({
         }
     },
 
+    _alignRootSimpleItems: function($container, colCount, labelsOptions) {
```
IgnatovDan: Declaration: function($container, colCount, labelsOptions) 
Call: this._alignRootSimpleItems($container, colCount, excludeTabbed);

excludeTabbed -> labelsOptions

Can we keep the same name?

IgnatovDan (changes requested):

js/ui/form/ui.form.js
novsstation: Yep, fixed https://github.com/DevExpress/DevExtreme/pull/15045/commits/71209775b78eab2b61bbe037264852167903d785

IgnatovDan (approved):
maxWidth will be updated only for root items labels by a maxWidth of root items labels.
maxWidth from group items labels is not used for root items labels.

jsdmitry (approved):

PR #15064, opened 2020-10-27
SelectBox should raise valueChanged with keydown event as parameter when value is removed using backspace/delete (T940489)

ksercs:
…


San4es (approved):
LGTM

alexander-kotov-dx (approved):

PR #15073, opened 2020-10-27
Fix memory leaks in click and feedback events

nightskylark:
Related to the issue T929051

### DIFF
--- a/js/events/click.js
+++ b/js/events/click.js
@@ -152,6 +152,8 @@ const useNativeClick =
             this.callBase();
 
             eventsEngine.off(this.getElement(), 'click', clickHandler);
+
+            lastFiredEvent = null;
         }
     });
 })();
--- a/js/events/click.js
+++ b/js/events/click.js
@@ -153,7 +153,9 @@ const useNativeClick =
 
             eventsEngine.off(this.getElement(), 'click', clickHandler);
 
-            lastFiredEvent = null;
+            if(lastFiredEvent.path.indexOf(this.getElement().get(0)) > -1) {
+                lastFiredEvent = null;
+            }
         }
     });
 })();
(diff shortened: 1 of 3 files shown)


San4es (dismissed):
LGTM

js/events/click.js
```diff
@@ -96,6 +97,29 @@ const useNativeClick =
...
+
+        return {
+            subscribe: (event, callback) => {
+                eventsEngine.on(nodesByEvent(event), REMOVE_EVENT_NAME, callback);
+            },
+            unsubscribe: (event, callback) => {
+                eventsEngine.off(nodesByEvent(event), REMOVE_EVENT_NAME, callback);
+            }
+        };
+    })();
```
MaximKudriavtsev: I suggest using static class methods or two pure functions instead of make a such strange and complex structure.
The main points are
- get rid of auto called function; (it's our performance...)
- get rid of excess closure.

ovchinnikov (commented):
Is is possible to unit test this changes? I mean write Unit tests

nightskylark:
> Is is possible to unit test this changes? I mean write Unit tests

I'm not sure it's possible to find out that DOM node was not disposed in unit tests. Do you have any ideas?

js/events/utils/event_nodes_disposing.js
```diff
@@ -0,0 +1,20 @@
+import eventsEngine from '../core/events_engine';
+
+const REMOVE_EVENT_NAME = 'dxremove';
```
MaximKudriavtsev, line 3: Import this event name from the [initial file](https://github.com/DevExpress/DevExtreme/blob/20_2/js/core/remove_event.js) be better in this case.

MaximKudriavtsev (approved):

PR #15175, opened 2020-11-09
ScrollView - scrollTo() method does not take into account the scrollbars dimension in Firefox (T947463)

EugeniyKiyashko:
Initially this bug should be occur after update of jquery version. I think that this bug is regression.

### DIFF
--- a/js/ui/scroll_view/ui.scrollable.js
+++ b/js/ui/scroll_view/ui.scrollable.js
@@ -550,7 +550,7 @@ const Scrollable = DOMComponent.inherit({
         const elementPosition = elementPositionRelativeToContent - pushBackOffset;
         const elementSize = $element[isVertical ? 'outerHeight' : 'outerWidth']();
         const scrollLocation = (isVertical ? this.scrollTop() : this.scrollLeft());
-        const clientSize = (isVertical ? this.clientHeight() : this.clientWidth());
+        const clientSize = this._container().get(0)[isVertical ? 'clientHeight' : 'clientWidth'];
 
         const startDistance = scrollLocation - elementPosition + startOffset;
         const endDistance = scrollLocation - elementPosition - elementSize + clientSize - endOffset;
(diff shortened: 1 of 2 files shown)


novsstation (dismissed):

js/ui/scroll_view/ui.scrollable.js
```diff
@@ -550,7 +550,7 @@ const Scrollable = DOMComponent.inherit({
         const elementPosition = elementPositionRelativeToContent - pushBackOffset;
         const elementSize = $element[isVertical ? 'outerHeight' : 'outerWidth']();
         const scrollLocation = (isVertical ? this.scrollTop() : this.scrollLeft());
-        const clientSize = (isVertical ? this.clientHeight() : this.clientWidth());
+        const clientSize = this._container().get(0)[isVertical ? 'clientHeight' : 'clientWidth'];
```
IgnatovDan, line 553: imho we need a new test in this PR: 
https://codepen.io/DanIgnatov/pen/LYZXVoB?editors=1010 - both the horizontal and vertical scroll bar sizes are not counted in FF

IgnatovDan (changes requested):

js/ui/scroll_view/ui.scrollable.js
EugeniyKiyashko: Tests was added in https://github.com/DevExpress/DevExtreme/pull/15175/commits/902b7584e4fe955a79eece101cc58e0704092cdd

novsstation (approved):

EugeniyKiyashko:
[T936630 DataGrid - Column headers are shifted if rtlEnabled is true and native scrolling is used](https://trello.com/c/j5Ldahkn/1329-t936630-datagrid-column-headers-are-shifted-if-rtlenabled-is-true-and-native-scrolling-is-used)

EugeniyKiyashko:
[ScrollView - scrollTo() method does not take into account the scrollbars dimension in Firefox (T947463)](https://trello.com/c/9Aq8OElW/1449-scrollview-scrollto-method-does-not-take-into-account-the-scrollbars-dimension-in-firefox-t947463)

IgnatovDan (approved):

PR #15227, opened 2020-11-12
DataGrid: Fix scroll position on initial load (T945892)

Alyar666:

### DIFF
--- a/js/ui/grid_core/ui.grid_core.virtual_scrolling.js
+++ b/js/ui/grid_core/ui.grid_core.virtual_scrolling.js
@@ -336,7 +336,7 @@ const VirtualScrollingRowsViewExtender = (function() {
                 scrollPosition = itemIndex * itemSize;
 
                 for(const index in itemSizes) {
-                    if(index <= itemIndex) {
+                    if(index < itemIndex) {
                         scrollPosition += itemSizes[index] - itemSize;
                     }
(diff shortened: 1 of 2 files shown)


NickMitrokhin (dismissed):

js/ui/grid_core/ui.grid_core.virtual_scrolling.js
```diff
@@ -335,9 +335,11 @@ const VirtualScrollingRowsViewExtender = (function() {
...
                 scrollPosition = itemIndex * itemSize;
 
-                for(const index in itemSizes) {
-                    if(index <= itemIndex) {
-                        scrollPosition += itemSizes[index] - itemSize;
+                if(itemIndex > 0) {
+                    for(const index in itemSizes) {
+                        if(index <= itemIndex) {
+                            scrollPosition += itemSizes[index] - itemSize;
+                        }
```
NickMitrokhin: Shouldn't we break the loop if index is greater than itemIndex? May be it is better to interrupt it.
```js
for(const index in itemSizes) {
   if(index <= itemIndex) {
      scrollPosition += itemSizes[index] - itemSize;
   } else {
      break;
   }
}
```

vconst (approved):

NickMitrokhin (approved):

PR #15250, opened 2020-11-12
DataGrid: Add duplicated names list to the E1059 error message (T948687)

jtoming830:


Alyar666 (dismissed):

js/ui/grid_core/ui.grid_core.columns_controller.js
```diff
@@ -2322,23 +2322,24 @@ export default {
...
                             usedNames[name] = true;
                         } else if(isEditable) {
                             hasEditableColumnWithoutName = true;
                         }
                     });
 
-                    if(hasDuplicatedNames) {
-                        errors.log('E1059');
+                    if(duplicatedNames.length) {
+                        errors.log('E1059', duplicatedNames.toString());
```
vconst: I think that `duplicatedNames.join(', ')` will be better instead of `duplicatedNames.toString()`

vconst (approved):

pavel-2020 (approved):

PR #15358, opened 2020-11-23
DataGrid - Fixes the data parameter in the validationCallback (T950070)

NickMitrokhin:


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1442,20 +1444,25 @@ const EditingController = modules.ViewController.inherit((function() {
...
 
-            that._pageIndex = that._dataController.pageIndex();
+            this._pageIndex = this._dataController.pageIndex();
+
+            this._setEditRowKey(item.key);
+            this._setEditColumnNameByIndex(editColumnIndex);
 
-            that._setEditRowKey(item.key);
-            that._setEditColumnNameByIndex(editColumnIndex);
+            if(!params.column.showEditorAlways) {
```
jtoming830, line 1460: Why we need this condition?

testing/tests/DevExpress.ui.widgets.dataGrid/editing.integration.tests.js
```diff
@@ -2183,6 +2183,63 @@ QUnit.module('Editing', baseModuleConfig, () => {
...
+                        },
+                        {
+                            dataField: 'name',
+                            validationRules: [
+                                { type: 'custom', validationCallback }
+                            ]
+                        }
+                    ],
+                    editing: {
+                        mode: editMode.toLowerCase(),
```
jtoming830, line 2215: Why using capital letters and then lowering them? Why not simple lower case instead of this? Aesthetics?

NickMitrokhin: >Aesthetics?

Exactly:)

js/ui/grid_core/ui.grid_core.editing.js
NickMitrokhin: This is for the boolean or any showEditorAlways column where the state is added before calling the _prepareEditCell. In this case, mutaded oldData can be stored to the internal state. For the scenario demonstrated in [this](https://github.com/DevExpress/DevExtreme/blob/20_2/testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js#L4719) test.

vconst (approved):

jtoming830 (approved):

PR #15382, opened 2020-11-24
Fix Deferred TS-declaration (T947959)

IlyaKhD:
* `Deferred` is declared as a class but implemented as a function (a factory).
* ~adding `@public` will remove the warning ([/core/utils/deferred.d.ts](https://unpkg.com/browse/devextreme@20.2.3/core/utils/deferred.d.ts), L43-45)~

### DIFF
--- a/js/core/utils/deferred.d.ts
+++ b/js/core/utils/deferred.d.ts
@@ -26,17 +26,7 @@ export interface dxDeferred<T> {
     promise(target?: any): dxPromise<T>;
 }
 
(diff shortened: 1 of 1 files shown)


nightskylark:
@IlyaKhD I didn't get why you've decided to remove warning. Isn't it correct?

IlyaKhD:
> @IlyaKhD I didn't get why you've decided to remove warning. Isn't it correct?

I assumed it is public since it is used in a public example. I'll remove `@public` tag.

nightskylark (approved):

PR #15400, opened 2020-11-25
dxScheduler - show a border between the DateTable and TimePanel when virtual scrolling and crossScrollingEnabled are used

AryamnovEugeniy:

### DIFF
--- a/js/renovation/ui/scheduler/workspaces/base/virtual-row.tsx
+++ b/js/renovation/ui/scheduler/workspaces/base/virtual-row.tsx
@@ -9,11 +9,11 @@ export const viewFunction = (viewModel: VirtualRow): JSX.Element => (
   <Row
     // eslint-disable-next-line react/jsx-props-no-spreading
     {...viewModel.restAttributes}
-    className={viewModel.classes}
     style={viewModel.style}
+    className={viewModel.classes}
   >
-    {viewModel.virtualCells.map(() => (
-      <VirtualCell />
+    {viewModel.virtualCells.map((_, index) => (
+      <VirtualCell key={index.toString()} />
     ))}
   </Row>
 );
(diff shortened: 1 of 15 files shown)


js/renovation/ui/scheduler/workspaces/base/virtual-row.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const { height } = this.props;
+    const { style } = this.restAttributes;
+
+    return addHeightToStyle(height, style);
+  }
+
+  get classes(): string {
+    const { className } = this.props;
+
+    return `dx-scheduler-virtual-row ${className}`;
```
MaximKudriavtsev, line 43: This variable has been declared in `scheduler/workspaces/ui.scheduler.work_space.js`. Can we use it here too?
> const VIRTUAL_ROW_CLASS = 'dx-scheduler-virtual-row';

js/renovation/ui/scheduler/workspaces/base/virtual-row.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  get virtualCells(): number[] {
+    const { cellsCount } = this.props;
+    const virtualCells = [] as number[];
+    for (let cellIndex = 0; cellIndex < cellsCount; cellIndex += 1) {
+      virtualCells.push(cellIndex);
+    }
+
+    return virtualCells;
```
MaximKudriavtsev: Use instead 

```js
return [...Array(4)].map((u, i) => i)
```
OR
```js
return Array.from({ length: 4 }).map((u, i) => i)
```

*simple and short*

js/renovation/ui/scheduler/workspaces/base/virtual-row.tsx
AryamnovEugeniy: Usually, we do not use constants from old code (as far as I know, no one does in renovation). An we were explicitly asked **not** to use constants for class names in the renovated code, so I would not do that, at least for now.

AntonSermyazhko (approved):

andrewmakarov (approved):

PR #15431, opened 2020-11-27
Testing: Update to use qunit package

Krinkle:
The `qunitjs` package was deprecated since 2.4.1, in favour of the `qunit` package name. Apologies for the inconvenience!

Ref <https://qunitjs.com/intro/#package-name-prior-to-241>
Ref <https://github.com/qunitjs/qunit/blob/2.12.0/History.md>

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
     "normalize-path": "^3.0.0",
     "npm-run-all": "^4.1.5",
     "pre-commit": "^1.2.2",
-    "qunitjs": "^2.0.1",
+    "qunit": "^2.12.0",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-test-renderer": "^16.12.0",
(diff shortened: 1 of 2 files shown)


Krinkle:
It looks like some test helpers that rely on QUnit internals don't quite work the same. I'm not familiar with the exact details of these helpers, but let me know if I can be of any help.

If you'd be willing to explain some of the specific use cases, I might be able to recommend alternate approaches that might work better on the current version, and/or to consider providing better hooks in the next release to simplify some of these extensions.

Thanks!

artem-kurchenko:
Hi @Krinkle

You're right. We use QUnit internals and this doesn't allow us to migrate to the next version easily. So, this task is not in our current plans. 

Could you please clarify if there are any specific reasons (except it becomes deprecated) we should update to the next version? Does the current one have severe issues?

Thanks,
Artem

Krinkle:
No, none that I'm aware of 🙂

If and when you have a need to update or switch, feel free to ping me, happy to help smoothen anything.

PR #15545, opened 2020-12-03
dropDownLists - support container limits for popup position calculation (T948335)

alexander-kotov-dx:

### DIFF
--- a/js/animation/position.js
+++ b/js/animation/position.js
@@ -220,6 +220,7 @@ const calculatePosition = function(what, options) {
     const offset = normalizeOffset(options.offset);
     const collision = normalizeCollision(options.collision);
     const boundary = options.boundary;
+    const container = options.container;
     const boundaryOffset = normalizeOffset(options.boundaryOffset);
 
     const h = {
@@ -281,8 +282,10 @@ const calculatePosition = function(what, options) {
         const win = $(window);
         const windowWidth = win.width();
         const windowHeight = win.height();
-        let left = win.scrollLeft();
-        let top = win.scrollTop();
+        let boundaryLeft = win.scrollLeft();
+        let boundaryTop = win.scrollTop();
+        let containerLeft = win.scrollLeft();
+        let containerTop = win.scrollTop();
         const documentElement = domAdapter.getDocumentElement();
         const hZoomLevel = touch ? documentElement.clientWidth / windowWidth : 1;
         const vZoomLevel = touch ? documentElement.clientHeight / windowHeight : 1;
@@ -293,26 +296,39 @@ const calculatePosition = function(what, options) {
 
         let boundaryWidth = windowWidth;
         let boundaryHeight = windowHeight;
+        let containerWidth = windowWidth;
+        let containerHeight = windowHeight;
 
         if(boundary) {
             const $boundary = $(boundary);
             const boundaryPosition = $boundary.offset();
 
-            left = boundaryPosition.left;
-            top = boundaryPosition.top;
+            boundaryLeft = boundaryPosition.left;
+            boundaryTop = boundaryPosition.top;
 
             boundaryWidth = $boundary.width();
             boundaryHeight = $boundary.height();
         }
 
+        if(container) {
+            const $container = $(container);
+            const containerPosition = $container.offset();
+
+            containerLeft = containerPosition.left;
+            containerTop = containerPosition.top;
+
+            containerWidth = $container.width();
+            containerHeight = $container.height();
+        }
+
         return {
             h: {
-                min: left + h.boundaryOffset,
-                max: left + boundaryWidth / hZoomLevel - h.mySize - h.boundaryOffset
+                min: Math.max(boundaryLeft + h.boundaryOffset, containerLeft),
+                max: Math.min(boundaryLeft + boundaryWidth / hZoomLevel - h.mySize - h.boundaryOffset, containerLeft + containerWidth / hZoomLevel - h.mySize)
             },
             v: {
-                min: top + v.boundaryOffset,
-                max: top + boundaryHeight / vZoomLevel - v.mySize - v.boundaryOffset
+                min: Math.max(boundaryTop + v.boundaryOffset, containerTop),
+                max: Math.min(boundaryTop + boundaryHeight / vZoomLevel - v.mySize - v.boundaryOffset, containerTop + containerHeight / vZoomLevel - v.mySize)
             }
         };
     })();
--- a/js/animation/position.js
+++ b/js/animation/position.js
@@ -119,6 +119,34 @@ const initMyLocation = function(data) {
         + data.offset;
 };
 
+const adjustBoundary = function(container, boundaryPreset, positionOptions) {
+    let result;
+
+    if(container && !isWindow(container)) {
+        const $container = $(container);
+        const containerPosition = $container.offset();
+
+        const containerLeft = containerPosition.left;
+        const containerTop = containerPosition.top;
+
+        const containerWidth = $container.width();
+        const containerHeight = $container.height();
+
+        result = {
+            h: {
+                min: Math.max(boundaryPreset.h.min, containerLeft),
+                max: Math.min(boundaryPreset.h.max, containerLeft + containerWidth / positionOptions.hZoomLevel - positionOptions.hMySize)
+            },
+            v: {
+                min: Math.max(boundaryPreset.v.min, containerTop),
+                max: Math.min(boundaryPreset.v.max, containerTop + containerHeight / positionOptions.vZoomLevel - positionOptions.vMySize)
+            }
+        };
+    }
+
+    return result || boundaryPreset;
+};
+
 const collisionResolvers = {
 
     'fit': function(data, bounds) {
@@ -294,8 +322,6 @@ const calculatePosition = function(what, options) {
 
         let boundaryWidth = windowWidth;
         let boundaryHeight = windowHeight;
-        let containerWidth = windowWidth;
-        let containerHeight = windowHeight;
 
         if(boundary) {
             const $boundary = $(boundary);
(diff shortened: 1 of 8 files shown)


js/animation/position.js
```diff
@@ -293,26 +294,50 @@ const calculatePosition = function(what, options) {
...
+            const containerTop = containerPosition.top;
+
+            containerWidth = $container.width();
+            containerHeight = $container.height();
+
+            horizontalMin = Math.max(horizontalMin, containerLeft);
+            horizontalMax = Math.min(horizontalMax, containerLeft + containerWidth / hZoomLevel - h.mySize);
+            verticalMin = Math.max(verticalMin, containerTop);
+            verticalMax = Math.min(verticalMax, containerTop + containerHeight / vZoomLevel - v.mySize);
+        }
```
DokaRus: could we extract this logic? 
```js
const boundaryPreset = {
      h: {
                min: horizontalMin,
                max: horizontalMax
            },
            v: {
                min: verticalMin,
                max: verticalMax
            }
};

return this.adjustBoundary(container, boundaryPreset, {
  hZoomLevel,
  vZoomLevel,
  hMySize: h.mySize,
  vMySize: v.mySize
});
<...>
adjustBoundary: function(container, boundaryPreset, positionOptions)
```

testing/tests/DevExpress.ui.widgets.editors/dropDownList.tests.js
```diff
@@ -1125,6 +1125,39 @@ QUnit.module('popup', moduleConfig, () => {
...
+        for(let i = 0; i < 20; i++) {
+            items.push(`item ${i}`);
+        }
+
+        const windowHeight = $(window).outerHeight();
+        const parentContainer = $('<div>')
+            .attr('id', 'specific-container')
+            .css('overflow', 'hidden')
+            .height(windowHeight / 3)
+            .insertBefore('#qunit-fixture');
```
ksercs: Is it necessary to place the container exactly outside of `#qunit-fixture`? If we place the container inside of it we'll avoid container removing in the test end

testing/tests/DevExpress.ui.widgets.editors/dropDownList.tests.js
```diff
@@ -1125,6 +1125,39 @@ QUnit.module('popup', moduleConfig, () => {
...
+        const $element = $('#dropDownList')
+            .css('margin-top', (windowHeight / 4) + 'px')
```
ksercs: it's not necessary to add `px` prefix

testing/tests/DevExpress.core/utils.position.tests.js
```diff
@@ -66,3 +66,37 @@ testModule('getBoundingRect', {
...
+
+        this.$innerContainer.appendTo(this.$container);
+        this.$container.appendTo($('#qunit-fixture'));
+    }
+}, function() {
+    test('getAvailableBoundsContainer should return undefined if container and its parents has no overflow: hidden styles', function(assert) {
+        assert.strictEqual(getAvailableBoundsContainer(this.$container), undefined);
+    });
+
+    test('getAvailableBoundsContainer should return container if container has overflow: hidden styles', function(assert) {
```
ksercs: This test is equal to the next now. Did u want to use `this.$innerContainer`?

js/core/utils/position.js
```diff
@@ -34,7 +39,29 @@ const getBoundingRect = (element) => {
     return rect;
 };
 
+const getAvailableBoundsContainer = (container) => {
```
ksercs: maybe it's better to name this function like `getCustomBoundaryContainer`?

js/core/utils/position.js
```diff
@@ -34,7 +39,29 @@ const getBoundingRect = (element) => {
...
+    const $container = container && $(container);
```
ksercs, line 43: To my mind, it's good idea to resolve cases when passed `container` :
- is not defined
- is window
- or at all `hasWindow()` is false

at the test beginning to simplify code below

js/core/utils/position.js
```diff
@@ -34,7 +39,29 @@ const getBoundingRect = (element) => {
...
+    let result;
+
+    if($container && $container.length && !isWindow($container.get(0))) {
+        const $containerWithParents = [].slice.call($container.parents());
+        $containerWithParents.unshift($container.get(0));
+
+        each($containerWithParents, function(i, parent) {
+            if(parent === $('body').get(0)) {
```
ksercs: Looks like we can just write condition as (if we accept previous sugge)
```
if(parent !== $('body').get(0) && window.getComputedStyle(parent).overflowY === 'hidden') {
    result = $(parent);
    return false;
}
```

js/ui/overlay/ui.overlay.js
```diff
@@ -488,6 +489,11 @@ const Overlay = Widget.inherit({
     _normalizePosition: function() {
         const position = this.option('position');
         this._position = typeof position === 'function' ? position() : position;
+        const container = this.option('container');
```
ksercs, line 492: are u sure that we should use `this.option('container')` here, not `this._getContainer()`? We don't take into account case when `position.of = #someContainer`. Is it ok?

js/ui/overlay/ui.overlay.js
```diff
@@ -488,6 +489,11 @@ const Overlay = Widget.inherit({
...
+        if(container && !isWindow(container)) {
```
ksercs: is this condition necessary? We check it inside of `getAvailableBoundContainer`

js/animation/position.js
```diff
@@ -293,26 +294,50 @@ const calculatePosition = function(what, options) {
...
             boundaryHeight = $boundary.height();
         }
 
+        let horizontalMin = boundaryLeft + h.boundaryOffset;
+        let horizontalMax = boundaryLeft + boundaryWidth / hZoomLevel - h.mySize - h.boundaryOffset;
+        let verticalMin = boundaryTop + v.boundaryOffset;
+        let verticalMax = boundaryTop + boundaryHeight / vZoomLevel - v.mySize - v.boundaryOffset;
+
+        if(container && !isWindow(container)) {
+            const $container = $(container);
```
ksercs: looks like we can just update `top` and `left` here this way:
```
top = Math.min(top, containerPosition.top);
top = Math.min(left, containerPosition.left);
```

and not to change any calculations

js/ui/overlay/ui.overlay.js
```diff
@@ -488,6 +489,11 @@ const Overlay = Widget.inherit({
...
+            this._position = this._position || {};
+            this._position.container = getAvailableBoundsContainer(container);
```
ksercs: mb we should name it like `boundaryContainer` to avoid confusion boundary container with real container in which overlay is positioned?

testing/tests/DevExpress.ui.widgets.editors/dropDownList.tests.js
alexander-kotov-dx: we need to use the visible part of the window

DokaRus: Will adding [qunit-fixture-visible](https://github.com/DevExpress/DevExtreme/blob/21_1/testing/runner/Views/Main/RunSuite.cshtml#L60-L65) class to element `#qunit-fixture` help in this case?

js/animation/position.js
alexander-kotov-dx: it can be used for 'min' bound calculation, but for 'max' we should use containerWidth/Height. And it is unnecessary after ['adjustBoundary'](https://github.com/DevExpress/DevExtreme/pull/15545#discussion_r536046578) function extraction

js/ui/overlay/ui.overlay.js
alexander-kotov-dx: I think this naming can add confusion with the boundary in the 'calculatePosition' function

js/ui/overlay/ui.overlay.js
alexander-kotov-dx: It is ok, position.of does not control dom element where the popup is rendered

js/animation/position.js
```diff
@@ -119,6 +119,34 @@ const initMyLocation = function(data) {
...
 
+const adjustBoundary = function(container, boundaryPreset, positionOptions) {
+    let result;
+
+    if(container && !isWindow(container)) {
+        const $container = $(container);
+        const containerPosition = $container.offset();
+
+        const containerLeft = containerPosition.left;
+        const containerTop = containerPosition.top;
```
DokaRus:
```suggestion
const {
  left: containerLeft,
  top: containerTop
} = $container.offset();
```

testing/tests/DevExpress.ui.widgets.editors/dropDownList.tests.js
```diff
@@ -1125,6 +1125,40 @@ QUnit.module('popup', moduleConfig, () => {
         parentContainer.remove();
     });
 
+    QUnit.test('popup is rendered at the top of element if container limited place below (T948335)', function(assert) {
+        const items = [];
+        $('#qunit-fixture').addClass('qunit-fixture-visible');
```
ksercs, line 1130: we should remove this class in the test end

testing/tests/DevExpress.ui.widgets.editors/dropDownList.tests.js
```diff
@@ -1125,6 +1125,40 @@ QUnit.module('popup', moduleConfig, () => {
...
+        }).dxDropDownList('instance');
+
+        instance.open();
+        this.clock.tick();
+
+        const $overlay = $(instance.content()).parent();
+        assert.roughEqual($element.offset().top - $overlay.offset().top, $overlay.height(), 2);
+        assert.ok($overlay.offset().top < $element.offset().top);
+
+        parentContainer.remove();
```
ksercs: it's not necessary because if qunit-fixture content update on every test start

js/core/utils/position.js
```diff
@@ -34,7 +39,27 @@ const getBoundingRect = (element) => {
     return rect;
 };
 
+const getCustomBoundaryContainer = (container) => {
+    const $container = container && $(container);
+    let result;
+
+    if($container && $container.length && hasWindow() && !isWindow($(container).get(0))) {
```
ksercs, line 46: to my mind it's better to use `container` native element here to simplify code

js/core/utils/position.js
```diff
@@ -34,7 +39,27 @@ const getBoundingRect = (element) => {
...
+        const $containerWithParents = [].slice.call($container.parents());
+        $containerWithParents.unshift($container.get(0));
+
+        each($containerWithParents, function(i, parent) {
+            if(parent !== $('body').get(0) && window.getComputedStyle(parent).overflowY === 'hidden') {
```
ksercs, line 51: we can use `is('body')` here

nightskylark (approved):
LGTM

js/animation/position.js
```diff
@@ -298,23 +327,30 @@ const calculatePosition = function(what, options) {
             const $boundary = $(boundary);
             const boundaryPosition = $boundary.offset();
 
-            left = boundaryPosition.left;
-            top = boundaryPosition.top;
+            boundaryLeft = boundaryPosition.left;
+            boundaryTop = boundaryPosition.top;
```
DokaRus, line 331:
```suggestion
const {
  left: boundaryLeft,
  top: boundaryTop
} = $boundary.offset();
```

DokaRus (approved):

PR #15590, opened 2020-12-07
DataGrid: Replaced insertKey with key in editing.changes

jtoming830:


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -828,15 +840,25 @@ const EditingController = modules.ViewController.inherit((function() {
...
-                        items.splice(key.dataRowIndex ? loadedRowIndex : 0, 0, item);
-                    }
+                let key = change.key;
+                let insertKey = this._getInternalData(key)?.insertKey;
+                if(isInsert && (!isDefined(change.key) || !isDefined(insertKey))) {
+                    const keys = this._getInsertKey();
+                    change.key = keys.key;
+                    key = keys.key;
+                    insertKey = keys.insertKey;
+                }
```
NickMitrokhin: isInsert is redundant in the following condition:
```suggestion
                if(!isDefined(change.key) || !isDefined(insertKey)) {
                    const keys = this._getInsertKey();
                    change.key = key = keys.key;
                    insertKey = keys.insertKey;
                }
```
Would you please clarify what scenario is covered by this code? 

In addition, insertKey and key that are stored in an internal data item are confusing. Maybe rename insertKey to insertId or insertInfo or something else?

NickMitrokhin (changes requested):

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: User can add his own changes to the init configuration. So we need to process it and add key/insertKey.

NickMitrokhin (dismissed):

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -895,35 +917,47 @@ const EditingController = modules.ViewController.inherit((function() {
...
             }).length;
 
-            if(editMode !== EDIT_MODE_BATCH) {
-                this._setEditRowKey(insertKey, true);
+            insertInfo[INSERT_INDEX] = this._getInsertIndex();
+
+            return insertInfo;
+        },
+
+        _getInsertInfo: function({ parentKey, key } = {}) {
```
vconst: Rename _getInsertInfo to _addInsertInfo because this method create insertInfo and add it to internal data.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -660,6 +660,15 @@ const EditingController = modules.ViewController.inherit((function() {
...
             }
 
+            args.value.forEach(change => {
+                if(change.type === 'insert') {
+                    const { key } = this._getInsertInfo({ key: change.key });
+                    if(!isDefined(change.key)) {
+                        change.key = key;
+                    }
+                }
+            });
```
vconst, line 678: Extract to separate method. For example, _processInsertChanges.

vconst (changes requested):

vconst (approved):

NickMitrokhin (approved):

PR #15649, opened 2020-12-10
Scrollview - native rtl behaviour (T936630)

novsstation:
Native rtl behaviour (T936630)


EugeniyKiyashko:
https://github.com/DevExpress/DevExtreme/pull/15027

PR #15733, opened 2020-12-16
List should have 'pageLoadMode' option default value equal to 'scrollBottom' on Mac

ksercs:
…

### DIFF
--- a/js/ui/list/ui.list.base.js
+++ b/js/ui/list/ui.list.base.js
@@ -246,14 +246,6 @@ export const ListBase = CollectionWidget.inherit({
                     pageLoadingText: '',
                     useInkRipple: true
                 }
-            },
-            {
-                device: function() {
-                    return devices.real().mac;
-                },
-                options: {
-                    pageLoadMode: 'scrollBottom'
-                }
             }
         ]);
     },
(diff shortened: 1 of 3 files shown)


js/ui/list/ui.list.base.js
```diff
@@ -246,6 +246,14 @@ export const ListBase = CollectionWidget.inherit({
                     pageLoadingText: '',
                     useInkRipple: true
                 }
+            },
+            {
+                device: function() {
+                    return devices.real().mac;
+                },
+                options: {
+                    pageLoadMode: 'scrollBottom'
```
DokaRus: This is the default value for the mac as it is.
We need to specify the [default value for the documentation](https://github.com/DevExpress/DevExtreme/blob/21_1/js/ui/list.d.ts#L327-L328) (macbook is a desktop, but primarily a mac)

```js
* @default 'scrollBottom' [for](Mac)
```

DokaRus (approved):

js/ui/list/ui.list.base.js
ksercs: oh thx

AlexJameson (approved):

PR #15762, opened 2020-12-18
DataGrid loses summaries for certain columns when exporting to Excel if columnRenderingMode is 'virtual' (T955954)

EugeniyKiyashko:


IgnatovDan (approved):
New scenario. At the moment, we don't have tests for columnRenderingMode: 'virtual' configs.

KuznetsovVN (approved):

PR #15835, opened 2020-12-23
Viz - tooltip renovation (step #3)

Krijovnick:

### DIFF
--- a/js/renovation/viz/core/__tests__/tooltip.test.tsx
+++ b/js/renovation/viz/core/__tests__/tooltip.test.tsx
@@ -271,6 +271,18 @@ describe('Render', () => {
     expect(tooltip.find('div').at(1).children().props()).toEqual({ valueText: 'Tooltip value text' });
     expect(tooltip.find('.tooltip-template').text()).toBe('Tooltip value text_template');
   });
+
+  it('should set on the div zIndex', () => {
+    const tooltip = shallow(<TooltipComponent {...{
+      ...props,
+      props: { ...props.props, zIndex: 3 },
+    } as any}
+    />);
+
+    expect(tooltip.find('div').at(0).props().style).toMatchObject({
+      zIndex: 3,
+    });
+  });
 });
 
 describe('Effect', () => {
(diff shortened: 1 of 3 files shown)


js/renovation/viz/core/__tests__/tooltip.test.tsx
```diff
@@ -271,11 +272,35 @@ describe('Render', () => {
     expect(tooltip.find('div').at(1).children().props()).toEqual({ valueText: 'Tooltip value text' });
     expect(tooltip.find('.tooltip-template').text()).toBe('Tooltip value text_template');
   });
+
+  it('should set on the div zIndex', () => {
+    const tooltip = shallow(<TooltipComponent {...{
+      ...props,
+      props: { ...props.props, zIndex: 3 },
+    } as any}
+    />);
```
MaximKudriavtsev: The same with previous PR. It's hard to understand what does it mean:
```jsx
<TooltipComponent {...{
      ...props,
      props: { ...props.props, zIndex: 3 },
    } as any}
/>
```

js/renovation/viz/core/tooltip.tsx
```diff
@@ -201,7 +205,17 @@ export class TooltipProps {
...
   @OneWay() interactive = false;
 
+  @OneWay() enabled = true;
+
+  @OneWay() shared = false;
+
+  @OneWay() location = 'center';
+
+  @OneWay() zIndex?: number;
+
```
MaximKudriavtsev: Why you did not specify a property type?

```ts
  @OneWay() enabled: boolean = true;

  @OneWay() shared: boolean = false;

  @OneWay() location: locationType = 'center';

  @OneWay() zIndex?: number;
```

js/renovation/viz/core/tooltip.tsx
```diff
@@ -266,6 +282,21 @@ export class Tooltip extends JSXComponent(TooltipProps) {
...
   }
 
+  @Method()
+  isEnabled(): boolean {
+    return this.props.enabled;
+  }
+
+  @Method()
+  isShared(): boolean {
+    return !!this.props.shared;
```
MaximKudriavtsev: Is not `shared` a boolean type? Why do we use a `!!` operator?

js/renovation/viz/core/tooltip.tsx
Krijovnick: The `shared` and `enabled` options have default values and it is clear what the type. The `location` option, I agree, should have the type

js/renovation/viz/core/tooltip.tsx
Krijovnick: yes, it is copy paste from the original code :)

MaximKudriavtsev (approved):

roman-simionov (approved):

dxBeardedBear (approved):

PR #15918, opened 2020-12-30
Resolve some Or and Nullish coalescing

alexander-kotov-dx:

### DIFF
--- a/js/ui/load_indicator.js
+++ b/js/ui/load_indicator.js
@@ -160,28 +160,6 @@ const LoadIndicator = Widget.inherit({
 
     _renderDimensions: function() {
         this.callBase();
-        this._updateContentSizeForAnimation();
-    },
-
-    _updateContentSizeForAnimation: function() {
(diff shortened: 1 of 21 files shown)


js/ui/load_indicator.js
```diff
@@ -160,28 +160,6 @@ const LoadIndicator = Widget.inherit({
 
     _renderDimensions: function() {
         this.callBase();
-        this._updateContentSizeForAnimation();
```
ksercs: imho, we should not remove this function. It will be called after `width` runtime option change for example

ksercs (changes requested):

ksercs:
Let's replace our [isDefined](https://github.com/DevExpress/DevExtreme/blob/21_1/js/core/utils/type.js#L27) function call with nullish coalescing?

ksercs (approved):

DokaRus (approved):

PR #16005, opened 2021-01-14
fix: elementAttr.class should overwrite previous value (T962695)

churkin:


js/core/dom_component.js
```diff
@@ -59,6 +59,8 @@ const DOMComponent = Component.inherit({
     _visibilityChanged: abstract,
     _dimensionChanged: abstract,
 
+    _customClass: null,
```
vconst: I think that default value for _customClass field should be assigned in the DOMComponent constructor.

vconst (approved):

PR #16123, opened 2021-01-22
DataGrid: extract formBased from editing

jtoming830:


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -751,23 +700,14 @@ const EditingController = modules.ViewController.inherit((function() {
         },
 
         getEditFormRowIndex: function() {
-            const editMode = getEditMode(this);
-
-            return editMode === EDIT_MODE_FORM || editMode === EDIT_MODE_POPUP ? this._getVisibleEditRowIndex() : -1;
+            return -1;
```
NickMitrokhin, line 701: It seems that getEditFormRowIndex should not be in ui.grid_core.editing.js. The same is related to other form and popup methods. From what I gather ui.grid_core.editing is assumed to be a base editing class. Should it know about popup and form?

jtoming830: As I think, base editing _can_ know about declarations and contain dummy implementations. This method can be called when mode is not form/popup, so returning -1 in the base class is OK. 
However, when it comes to some more complicated logic (ex. calling _getVisibleEditRowIndex), this should be moved to the formBased.
Also, in this pull request I did not move all code related to form/popup edit modes, only easy to move and the most obvious parts.

NickMitrokhin: OK.

NickMitrokhin (dismissed):

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -289,6 +247,23 @@ const EditingController = modules.ViewController.inherit((function() {
             this.component._optionsByReference[EDITING_CHANGES_OPTION_NAME] = true;
         },
 
+        getDefaultEditorTemplate: function() {
```
vconst: getDefaultEditorTemplate -> _getDefaultEditorTemplate because it is used in this module only.

vconst (commented):
ui.data_grid.editing_formBased.js -> ui.data_grid.editing_form_based.js 

Also, Why the following variables and methods are not moved to form_based? 
_editForm
_updateEditFormDeferred
getEditFormRowIndex
getEditForm
getEditFormTemplate
getEditFormOptions

vconst (approved):

NickMitrokhin (approved):

PR #16169, opened 2021-01-26
Remove deferRendering styles from dx.common.css

groshenkovamarina:


babich-a (changes requested):
We need to include these styles to the material and generic bundles.

babich-a (approved):

PR #16182, opened 2021-01-27
Makes DataGrid load several pages at once on virtual scrolling

NickMitrokhin:
Added the capability to calculate how many pages should be loaded on a single load operation. This is done by calculating the loadPageCount option at the DataSourceAdapter level using the **skip** and **take** parameters. These params are calculated based on the view port size. Note, that this logic operates only if the "scrolling.newMode" option is set to true. The option is added temporarily for preventing the current behavior from breaking.

### DIFF
--- a/testing/tests/DevExpress.ui.widgets.dataGrid/virtualScrolling.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.dataGrid/virtualScrolling.tests.js
@@ -902,14 +902,6 @@ QUnit.module('VirtualScrollingController. New mode', {
         pageIndex.restore();
     });
 
-    QUnit.test('load does set loading page indexes', function(assert) {
-        // act
-        this.scrollController.load();
-
-        // assert
-        assert.deepEqual(this.scrollController._loadingPageIndexes, {}, '_loadingPageIndexes is empty');
-    });
-
     QUnit.test('Viewport params at the top', function(assert) {
         const viewportSize = 25;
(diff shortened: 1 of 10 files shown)


vconst (commented):
Add tests to virtualScrolling.tests.js for virtualItemsCount, setViewportItemIndex, pageIndex, load methods where logic is changed for newMode.
Also, I did not find tests about loadPageCount modification and take correction.

Alyar666 (dismissed):

testing/tests/DevExpress.ui.widgets.dataGrid/virtualScrolling.tests.js
```diff
@@ -858,6 +858,58 @@ QUnit.module('VirtualScrollingController. New mode', {
...
+
+        pageIndex.restore();
+    });
+
+    QUnit.test('load does set loading page indexes', function(assert) {
+        // act
+        this.scrollController.load();
+
+        // assert
+        assert.deepEqual(this.scrollController._loadingPageIndexes, {}, '_loadingPageIndexes is empty');
```
vconst: Do not test private state.

NickMitrokhin: At the moment, there is no option to test when the load method is called with the new mode. It seems it is better to remove this test.

vconst (approved):

PR #16212, opened 2021-01-28
Remove responsiveBox styles from dx.common.css

groshenkovamarina:


scss/widgets/base/_responsiveBox.scss
```diff
@@ -0,0 +1,2 @@
+
```
babich-a: Do we need this file, if it has no styles and do not included in other files?

babich-a (changes requested):

babich-a (approved):

PR #16260, opened 2021-02-03
Renovated DataGrid: Add beginUpdate, endUpdate and dispose methods

Alyar666:

### DIFF
--- a/js/renovation/preact_wrapper/__tests__/data_grid.test.tsx
+++ b/js/renovation/preact_wrapper/__tests__/data_grid.test.tsx
@@ -1,0 +1,51 @@
+/* eslint-disable class-methods-use-this */
+import DataGridComponent from '../data_grid';
+import BaseComponent from '../component';
+
+jest.mock('../component', () => {
+  class T {
+    static beginUpdateMock = jest.fn();
+
+    static endUpdateMock = jest.fn();
+
+    beginUpdate() {
+      T.beginUpdateMock();
+    }
+
+    endUpdate() {
+      T.endUpdateMock();
+    }
+  }
+
+  return T;
+});
+
+it('beginUpdate', () => {
+  const beginUpdate = jest.fn();
+  const component = new DataGridComponent({} as Element) as any;
+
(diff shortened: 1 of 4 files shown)


js/renovation/preact_wrapper/data_grid.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable */
+import Component from './component';
+
+export default class DataGrid extends Component {
+    beginUpdate() {
+        const gridInstance = (this as any).viewRef?.getComponentInstance();
```
vconst: Try not to use any here.

js/renovation/preact_wrapper/__tests__/data_grid.test.tsx
```diff
@@ -0,0 +1,51 @@
+/* eslint-disable class-methods-use-this */
+import DataGridComponent from '../data_grid';
+import BaseComponent from '../component';
+
+jest.mock('../component', () => {
+  class T {
+    static beginUpdateMock = jest.fn();
+
+    static endUpdateMock = jest.fn();
```
vconst: Move mock functions outside jest.mock. In this case mock functions should have prefix 'mock'. It will be simpler to access the mock functions in the tests and you can use mockImplementation instead of class T.

js/renovation/ui/data_grid/data_grid.tsx
```diff
@@ -421,6 +431,11 @@ export class DataGrid extends JSXComponent(DataGridProps) {
     this.prevProps = this.props;
   }
 
+  @Effect()
+  dispose() {
```
vconst, line 435: Add run: 'once' parameter to Effect. Or dispose will be called before real dispose.

vconst (changes requested):

vconst (dismissed):

MikeVitik (dismissed):

vconst (approved):

MikeVitik (approved):

PR #16276, opened 2021-02-03
TreeList: Fix additional border when header panel is visible (T969977)

jtoming830:


scss/widgets/base/_gridBase.scss
```diff
@@ -82,7 +82,7 @@
       }
     }
 
-    &.dx-#{$widget-name}-borders {
+    &.dx-#{$widget-name}-borders, &> .dx-#{$widget-name}-borders {
```
vconst: Why two locations of .dx-#{$widget-name}-borders class here? May be we can remove &.dx-#{$widget-name}-borders and it was wrong?

Alyar666 (approved):

vconst (approved):

PR #16303, opened 2021-02-08
DataGrid - Fixes resetting tabindex of the expanded cell in the master row (T969832)

NickMitrokhin:

### DIFF
--- a/js/ui/grid_core/ui.grid_core.keyboard_navigation.js
+++ b/js/ui/grid_core/ui.grid_core.keyboard_navigation.js
@@ -144,7 +144,7 @@ const KeyboardNavigationController = core.ViewController.inherit({
             const isRelatedTargetInRowsView = $(event.relatedTarget).closest(rowsView.element()).length;
             const isCommandButton = $element.hasClass('dx-link');
 
-            if(isCommandButton && !isRelatedTargetInRowsView) {
+            if(isCommandButton && !isRelatedTargetInRowsView && this._isEventInCurrentGrid(event)) {
                 let $focusedCell = this._getFocusedCell();
 
                 $focusedCell = !isElementDefined($focusedCell) ? rowsView.getCellElements(0).filter('[tabindex]').eq(0) : $focusedCell;
--- a/js/ui/grid_core/ui.grid_core.keyboard_navigation.js
+++ b/js/ui/grid_core/ui.grid_core.keyboard_navigation.js
@@ -189,6 +189,9 @@ const KeyboardNavigationController = core.ViewController.inherit({
             const columnsResizerController = this.getController('columnsResizer');
             const isColumnResizing = !!columnsResizerController && columnsResizerController.isResizing();
             if(!isCurrentRowsViewClick && !isEditorOverlay && !isColumnResizing) {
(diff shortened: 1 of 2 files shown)


js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -189,6 +189,9 @@ const KeyboardNavigationController = core.ViewController.inherit({
             const columnsResizerController = this.getController('columnsResizer');
             const isColumnResizing = !!columnsResizerController && columnsResizerController.isResizing();
             if(!isCurrentRowsViewClick && !isEditorOverlay && !isColumnResizing) {
+                const targetInsideFocusedView = this._focusedView ? $target.parents().filter(this._focusedView.element()).length > 0 : false;
+
+                !targetInsideFocusedView && this._resetFocusedCell(this._getFocusedCell(), true);
```
jtoming830: Maybe "if" will be more readable?
Also _resetFocusedCell will get focused cell by default if first argument is not defined. Maybe . We also have the same logic (get focused cell and reset it) in the _resetFocusedView method. 
I think that first argument in _resetFocusedCell can be removed at all because it is never used.

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -505,6 +508,7 @@ const KeyboardNavigationController = core.ViewController.inherit({
         if(isOriginalHandlerRequired) {
             this._editorFactory.loseFocus();
             if(this._editingController.isEditing() && !this._isRowEditMode()) {
+                this._resetFocusedCell(this._getFocusedCell(), true);
```
jtoming830: The same

testing/tests/DevExpress.ui.widgets.dataGrid/focus.integration.tests.js
```diff
@@ -3267,6 +3267,95 @@ QUnit.module('View\'s focus', {
...
+        // act
+        $(this.dataGrid.element()).find('.myclass .dx-link-edit:eq(0)').trigger('dxpointerdown').trigger('click');
+        this.clock.tick();
+
+        // assert
+        assert.strictEqual($(this.dataGrid.element()).find('.dx-datagrid-rowsview .dx-command-expand:eq(1)').attr('tabindex'), '0', 'tab index is set to the expanded button');
+        assert.equal($(this.dataGrid.element()).find('.myclass .dx-editor-cell.dx-focused').length, 1, 'focused edit cell');
+    });
+
+    QUnit.testInActiveWindow('Grid should scroll to the top when a command button is clicked in a detail grid (T969832)', function(assert) {
```
jtoming830: As I understand - should not, am I right?

NickMitrokhin: Yes:)

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
NickMitrokhin: I'm not sure that "if" will be more "readable". As for the first parameter of the _resetFocusedCell, you are right, it can be removed.

Alyar666 (approved):

jtoming830 (approved):

PR #16322, opened 2021-02-09
DataGrid - Fixes scrolling position with focused row (T970969)

NickMitrokhin:

### DIFF
--- a/js/ui/grid_core/ui.grid_core.keyboard_navigation.js
+++ b/js/ui/grid_core/ui.grid_core.keyboard_navigation.js
@@ -101,6 +101,11 @@ function isFixedColumnIndexOffsetRequired(that, column) {
     return result;
 }
 
+function shouldPreventScroll(that) {
+    const keyboardController = that.getController('keyboardNavigation');
+    return keyboardController._isVirtualScrolling() ? that.option('focusedRowIndex') === keyboardController.getRowIndex() : false;
+}
+
 const KeyboardNavigationController = core.ViewController.inherit({
     // #region Initialization
     init: function() {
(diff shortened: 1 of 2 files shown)


js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -101,6 +101,11 @@ function isFixedColumnIndexOffsetRequired(that, column) {
     return result;
 }
 
+function shouldPreventScroll(that) {
```
jtoming830, line 104: IMO - if function uses context, it should be a method.

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -1998,13 +2003,15 @@ export default {
                 renderDelayedTemplates: function(change) {
                     this.callBase.apply(this, arguments);
                     if(!change || !change.repaintChangesOnly) {
-                        this.renderFocusState();
+                        const preventScroll = shouldPreventScroll(this);
```
jtoming830, line 2009: Maybe move this logic inside renderFocusState? Something like this:
```
renderFocusState: function(preventScroll) {
  const keyboardController = this.getController('keyboardNavigation');
  preventScroll = preventScroll ?? keyboardController.shouldPreventScroll();
  // with ES2021 can be changed to:
  // preventScroll ??= keyboardController.shouldPreventScroll();
  . . .
}
```
If we will need to render focus state without scroll preventing, we can pass false and it will work correctly. No argument - ask keyboardController for shouldPreventScroll. True - just prevent scroll.

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
NickMitrokhin: This is a private function that is not used outside the module. I think it is better to keep it as is.

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
NickMitrokhin: Technically, the scroll itself is implemented in FocusController._setFocusedRowElementTabIndex method. And the preventScroll parameter should be recalculated there. Since the **renderFocusState** method accepts the **preventScroll** parameter, I think it is better to calculate it before passing it to **renderFocusState**.

jtoming830 (approved):

Alyar666 (approved):

PR #16331, opened 2021-02-10
Fix timeout in angular SSR application on first start (T968400)

vconst:


testing/tests/DevExpress.serverSide/themes.tests.js
```diff
@@ -1,59 +1,17 @@
...
-    }
-});
-
-QUnit.test('theme changing', function(assert) {
-    const done = assert.async();
-    themes.current('generic.light');
-
-    assert.equal(themes.current(), 'generic.light');
-
-    themes.initialized(() => {
```
babich-a, line 27: The reason why these tests were deleted is not clear

vconst: They were wrong because it is not possible to change theme on the server side and there are not theme links on the server side.

nightskylark (approved):

PR #16336, opened 2021-02-10
Remove dx.viz and dx.web bundles from npm package to make package size less 100M

vconst:


churkin (dismissed):

nightskylark (commented):
Why did you left these files
https://github.com/DevExpress/DevExtreme/blob/21_1/js/bundles/dx.viz.js
https://github.com/DevExpress/DevExtreme/blob/21_1/js/bundles/dx.web.js

and these usages?
https://github.com/DevExpress/DevExtreme/blob/7b16996944d8e1db002a05401b6b69be4651cc93/testing/tests/Bundles/dx.viz.tests.js#L4
https://github.com/DevExpress/DevExtreme/blob/7b16996944d8e1db002a05401b6b69be4651cc93/testing/tests/DevExpress.aspnet/aspnet_bundled.tests.js#L14
https://github.com/DevExpress/DevExtreme/blob/7b16996944d8e1db002a05401b6b69be4651cc93/testing/tests/Bundles/dx.web.tests.js#L4

vconst:
> Why did you left these files
> https://github.com/DevExpress/DevExtreme/blob/21_1/js/bundles/dx.viz.js
> https://github.com/DevExpress/DevExtreme/blob/21_1/js/bundles/dx.web.js
> 
> and these usages?
> ...

I don't want to remove these bundles from Installation and from DevExtreme CDN to minimize BC for us and for customers. So, these usages are actual. Also, I have updated my PR because my first commit removed these bundles from everywhere and it was wrong.

nightskylark (approved):

ghost:
I'm having trouble understanding why these files where removed, except making this package size smaller for bundlers.

I'm using a standard .NET Framework MVC app, and I was directly importing the js files I needed from the dist folder, and dx.web.js was one of them.
This was saving me more than a megabyte of script import, since I only was using web components.

image

Do I need to build my own package with this new version to save space ?

PR #16372, opened 2021-02-12
Add default exports to internal cjs modules

churkin:


build/gulp/npm.js
```diff
@@ -100,12 +99,12 @@ if(isEsmPackage) {
 
 const jsonGlobs = ['js/**/*.json', '!js/viz/vector_map.utils/*.*'];
 
+const hasDefaultExport = (content) => /exports\.default\s=/.test(String(content));
```
vconst, line 102: Create regexp outside function because regexp creation take a time.

churkin: image

vconst (dismissed):

build/gulp/npm.js
```diff
@@ -100,12 +99,12 @@ if(isEsmPackage) {
...
+const isCjsModule = (path) => path.indexOf('esm/') !== 0;
```
IlyaKhD:
```suggestion
const isCjsModule = (path) => !path.startsWith('esm/');
```

vconst (approved):

PR #16374, opened 2021-02-12
DataGrid/TreeList: add editing_cell_based module

jtoming830:


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -1039,6 +939,17 @@ const EditingController = modules.ViewController.inherit((function() {
...
 
+        _checkFocus: function() {
+            return true;
+        },
+
+        _checkEditItem: function() {
+            return true;
+        },
+
+        _validateBeforeEditCell: noop,
```
NickMitrokhin: I think editing modules should not know about validation. So, all methods with the "validate" word should be moved to the validating module.

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -2046,20 +2046,22 @@ export default {
         },
         controllers: {
             editing: {
-                editCell: function(rowIndex, columnIndex) {
+                _editCell: function({ isCellEditing }) {
```
NickMitrokhin: Shouldn't this method be called before _editCell in the CellBased module?

js/ui/grid_core/ui.grid_core.keyboard_navigation.js
```diff
@@ -2101,12 +2101,10 @@ export default {
                 closeEditCell: function() {
                     const keyboardNavigation = this.getController('keyboardNavigation');
                     keyboardNavigation._fastEditingStarted = false;
-
-                    const result = this.callBase.apply(this, arguments);
-
+                },
+                _afterCloseEditCell: function() {
+                    const keyboardNavigation = this.getController('keyboardNavigation');
```
NickMitrokhin: It seems that the callBase call is missing here.

js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -644,15 +644,12 @@ export default {
                     return result;
                 },
 
-                _prepareEditCell: function(params) {
-                    const isNotCanceled = this.callBase.apply(this, arguments);
+                _afterPrepareEditCell: function(params) {
                     const validatingController = this.getController('validating');
 
-                    if(isNotCanceled && params.column.showEditorAlways) {
+                    if(params.column.showEditorAlways) {
```
NickMitrokhin: Why is NotCanceled removed from the condition?

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: I think it doesn't matter how are called methods which implementations are placed in extenders. I can rename it and according your point of view it will be OK, but the essence remains the same - editing calls some methods which are implemented in validating/KBN/adaptivity/elsewhere. IMO, we should think simpler - editing must not know about validating _controller_ or contain validating _implementation_.

js/ui/grid_core/ui.grid_core.adaptivity.js
```diff
@@ -883,15 +883,13 @@ export default {
                     return this.callBase(cellOptions, column);
                 },
 
-                _closeEditItem: function($targetElement) {
+                _checkEditItem: function($targetElement) {
```
vconst: Bad name for method. It is not clear what checks here. May be _checkCloseEditItem will be better.

vconst: Also, Why is this method not moved to cellBased file completely? Is it needed in row modes?

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -584,21 +516,12 @@ const EditingController = modules.ViewController.inherit((function() {
                 }
             } else {
                 if(isDefined(args.value) && args.value !== args.previousValue) {
-                    this._editCellFromOptionChanged(columnIndex, columnIndex, oldRowIndex);
+                    this._editCellFromOptionChanged?.(columnIndex, columnIndex, oldRowIndex);
```
vconst: Why is this code not moved to cellBased?

NickMitrokhin (commented):
We should think of how to remove the following remained constants and their usage from ui.grid_core.editing.js to corresponding modules:

const EDIT_MODE_BATCH = 'batch'; //to ui.grid_core.editing_cell_based.js
const EDIT_MODE_CELL = 'cell'; //to ui.grid_core.editing_cell_based.js
const EDIT_MODE_FORM = 'form'; //to ui.grid_core.editing_form_based.js
const EDIT_MODE_POPUP = 'popup'; //to ui.grid_core.editing_form_based.js

jtoming830:
> We should think of how to remove the following remained constants and their usage from ui.grid_core.editing.js to corresponding modules:
> 
> const EDIT_MODE_BATCH = 'batch'; //to ui.grid_core.editing_cell_based.js
> const EDIT_MODE_CELL = 'cell'; //to ui.grid_core.editing_cell_based.js
> const EDIT_MODE_FORM = 'form'; //to ui.grid_core.editing_form_based.js
> const EDIT_MODE_POPUP = 'popup'; //to ui.grid_core.editing_form_based.js

I think that we will move all constants from editing.js to the separate file (editing_consts.js?) to avoid duplication of const definitions. And these constants will be imported from it.

NickMitrokhin:
>I think that we will move all constants from editing.js to the separate file (editing_consts.js?) to avoid duplication of const definitions. And these constants will be imported from it.

Yes, we can move them to a separate file, but one problem remains: we need to remove the usage of these constants from ui.grid_core.editing.js.

NickMitrokhin (approved):

vconst (approved):

PR #16552, opened 2021-03-02
HTML Editor - rename toolbar item options

alexander-kotov-dx:


DokaRus (dismissed):

RomanTsukanov (dismissed):

js/ui/html_editor/modules/toolbar.js
```diff
@@ -472,11 +473,28 @@ if(Quill) {
             return $container;
         }
 
+        _checkItemOptions(item) {
```
ksercs: maybe it's better to name this function as `warnOnRenamedOptions` or smth like this? To my mind it's not clear what we want to check now

js/ui/html_editor/modules/toolbar.js
```diff
@@ -472,11 +473,28 @@ if(Quill) {
...
+            const optionsInfo = [{
+                newName: 'name',
+                oldName: 'formatName'
+            }, {
+                newName: 'acceptedValues',
+                oldName: 'formatValues'
+            }];
+
+            each(optionsInfo, (index, optionName) => {
+                if(isObject(item) && isDefined(item[optionName.oldName])) {
```
ksercs: `isObject(null)` gives true. Seems like `Cannot read property of null` can be thrown here `optionName.oldName`

ksercs: we can also move `isObject` check out of the loop

ksercs: What about case when item old named option set to `null` or `undefined`? User will not get any error. Mb it's better to check if `oldName` exists in `item`?

js/ui/html_editor/modules/toolbar.js
```diff
@@ -510,18 +528,18 @@ if(Quill) {
...
-            const iconName = formatName === 'clear' ? 'clearformat' : formatName;
-            const buttonText = titleize(formatName);
+        _prepareButtonItemConfig(name) {
+            const iconName = name === 'clear' ? 'clearformat' : name;
+            const buttonText = titleize(name);
 
             return {
                 widget: 'dxButton',
-                formatName: formatName,
+                name: name,
```
ksercs: we can see write just `name` here

js/ui/html_editor/modules/toolbar.js
```diff
@@ -531,33 +549,33 @@ if(Quill) {
         _prepareSelectItemConfig(item) {
```
ksercs, line 551: maybe destructuring? I know that this row was not changed by u but it can help make a lot of changed rows below more accurate
`{ name, acceptedValues }`

js/ui/html_editor/modules/toolbar.js
alexander-kotov-dx: our isObject returns false in this case 
console.log(isObject(null));  // false

DokaRus (approved):

PR #16618, opened 2021-03-10
DataGrid - Fixes deleting rows on editing (T978455) (#16617)

NickMitrokhin:

### DIFF
--- a/js/ui/grid_core/ui.grid_core.editing.js
+++ b/js/ui/grid_core/ui.grid_core.editing.js
@@ -1534,9 +1534,11 @@ const EditingController = modules.ViewController.inherit((function() {
         deleteRow: function(rowIndex) {
             if(this.option('editing.mode') === 'cell' && this.isEditing()) {
                 const isNewRow = this._dataController.items()[rowIndex].isNewRow;
+                const rowKey = this._dataController.getKeyByRowIndex(rowIndex);
 
                 // T850905
                 this.closeEditCell(null, isNewRow).always(() => {
+                    rowIndex = this._dataController.getRowIndexByKey(rowKey);
                     this._checkAndDeleteRow(rowIndex);
                 });
             } else {
@@ -1813,11 +1815,11 @@ const EditingController = modules.ViewController.inherit((function() {
                     }
                     this._saving = true;
                     this._saveEditDataInner()
-                        .done(deferred.resolve)
-                        .fail(deferred.reject)
                         .always(() => {
                             this._saving = false;
-                        });
+                        })
+                        .done(deferred.resolve)
+                        .fail(deferred.reject);
                 }).fail(deferred.reject);
             }).fail(deferred.reject);
             return deferred.promise();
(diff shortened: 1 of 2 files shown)


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -2073,42 +2075,50 @@ const EditingController = modules.ViewController.inherit((function() {
...
 
             if(!isRowEditMode(that)) {
-                result = Deferred();
-                this.executeOperation(result, () => {
-                    this._closeEditCellCore(isError, oldEditRowIndex, withoutSaveEditData);
-                    result.resolve();
+                const deferred = new Deferred();
+                result = new Deferred();
+                this.executeOperation(deferred, () => {
+                    this._closeEditCellCore(isError, oldEditRowIndex, withoutSaveEditData).always(result.resolve);
```
jtoming830, line 2081: `closeEditCellCore` can not be rejected, why are you using `always` instead of `done`?

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -2073,42 +2075,50 @@ const EditingController = modules.ViewController.inherit((function() {
...
 
-            if(editMode === EDIT_MODE_CELL && that.hasChanges()) {
+            if(editMode === EDIT_MODE_CELL && this.hasChanges()) {
                 if(!withoutSaveEditData) {
-                    that.saveEditData().done(function(error) {
-                        if(!that.hasChanges()) {
-                            that.closeEditCell(!!error);
+                    this.saveEditData().done((error) => {
+                        if(!this.hasChanges()) {
+                            this.closeEditCell(!!error).always(deferred.resolve);
```
jtoming830, line 2098: `closeEditCell` can not be rejected, why are you using `always` instead of `done`?

js/ui/grid_core/ui.grid_core.editing.js
NickMitrokhin: I think it does not matter in this particular case. This call is analogue of the this.closeEditCell(null, isNewRow).always in the deleteRow method. Especially if we add the case when _closeEditCellCore will reject the returned promise in future.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -2073,42 +2075,50 @@ const EditingController = modules.ViewController.inherit((function() {
...
         },
 
         _closeEditCellCore(isError, oldEditRowIndex, withoutSaveEditData) {
-            const that = this;
-            const editMode = getEditMode(that);
-            const dataController = that._dataController;
+            const editMode = getEditMode(this);
+            const dataController = this._dataController;
+            const deferred = new Deferred();
+            const promise = deferred.promise();
```
jtoming830, line 2092: you can not create this variable and return `deferred.promise()`. It's definition seems like something happens with `promise` in code but it is only returned.

js/ui/grid_core/ui.grid_core.editing.js
NickMitrokhin: The cause is the same as in the previous comment)

js/ui/grid_core/ui.grid_core.editing.js
NickMitrokhin: I did it to prevent from writing redundant lines of code with "deferred.resolve" in different if...else conditions.

jtoming830: Ah, sorry, didn't see 2103 line

js/ui/grid_core/ui.grid_core.editing.js
jtoming830: > Especially if we add the case when _closeEditCellCore will reject the returned promise in future.

Tbh, resolving `closeEditCell`'s deferred when `_closeEditCellCore` rejects it's own looks strange. But if you think it doesn't matter - OK.

jtoming830 (approved):

vconst (approved):

PR #16642, opened 2021-03-11
Checkbox - implementation of setAria method

novsstation:
Temporary implementation of setAria method to maintain backward compatibility

### DIFF
--- a/js/renovation/component_wrapper/check_box.ts
+++ b/js/renovation/component_wrapper/check_box.ts
@@ -27,4 +27,11 @@ export default class CheckBox extends Editor {
 
     this._invalidate();
   }
+
+  setAria(name: string, value: string): void {
+    const attrName = (name === 'role' || name === 'id')
+      ? name
+      : `aria-${name}`;
+    (this.$element() as any).attr(attrName, value);
+  }
 }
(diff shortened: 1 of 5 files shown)


EugeniyKiyashko (dismissed):

js/renovation/component_wrapper/check_box.ts
```diff
@@ -27,4 +27,9 @@ export default class CheckBox extends Editor {
 
     this._invalidate();
   }
+
+  setAria(name: string, value: string): void {
+    const attrName = this._getAriaName(name);
+    (this.$element() as any).attr(attrName, value);
```
MikeVitik: May be (this.$element() as any) -> (this.$element() as HTMLDivElement)

MikeVitik (dismissed):

js/renovation/component_wrapper/check_box.ts
novsstation: $element() returns our wrapper, not native HTMLDivElement.
This wrapper implements JQuery interface.

It's very strange but lint warns about non [consistent type](https://github.com/DevExpress/DevExtreme/pull/16642/checks?check_run_id=2094488713), if I cast it to the JQuery
image

novsstation: Fixed via using of the addAttributes method in our utils..ts

ksercs (approved):

EugeniyKiyashko (approved):

PR #16656, opened 2021-03-12
Fix hands color in timpicker (T978638)

OlegKipchatov:


babich-a (changes requested):
It seems that hands will not change color if you change accent color in themebuilder. Look at diagramm/filemanager approach

babich-a (dismissed):
I cant read all code but this approach looks good to me. TB tests says that cant compile this. Sass don't like variables in paths

babich-a (dismissed):

babich-a (approved):

PR #16703, opened 2021-03-16
ScrollView - Scrolling does not work when nested in another ScrollView that has useNative enabled (T977318)

EugeniyKiyashko:


novsstation (dismissed):

scss/widgets/base/_scrollable.scss
```diff
@@ -15,15 +15,17 @@
   -ms-overflow-style: -ms-autohiding-scrollbar;
   -ms-scroll-snap-type: proximity; // stylelint-disable-line property-no-vendor-prefix
 
-  .dx-scrollable-scrollbar {
+  &.dx-scrollable-wrapper > .dx-scrollable-container > .dx-scrollable-scrollbar,
```
IgnatovDan, line 18: I think we need a new test

IgnatovDan (changes requested):

scss/widgets/base/_scrollable.scss
EugeniyKiyashko: Fixed in https://github.com/DevExpress/DevExtreme/pull/16703/commits/e21dfd0e8d4a9c34439196558c39af09d318efc9

novsstation (dismissed):

novsstation (approved):

IgnatovDan (approved):

PR #16704, opened 2021-03-16
ScrollView - Scrolling does not work when nested in another ScrollView that has useNative enabled (T977318)

EugeniyKiyashko:


scss/widgets/base/_scrollable.scss
```diff
@@ -15,15 +15,17 @@
   -ms-overflow-style: -ms-autohiding-scrollbar;
   -ms-scroll-snap-type: proximity; // stylelint-disable-line property-no-vendor-prefix
 
-  .dx-scrollable-scrollbar {
+  &.dx-scrollable-wrapper > .dx-scrollable-container > .dx-scrollable-scrollbar,
+  > div > .dx-scrollable-wrapper > .dx-scrollable-container > .dx-scrollable-scrollbar {
     display: none;
```
IgnatovDan, line 20: I think we need a new test

IgnatovDan (dismissed):

scss/widgets/base/_scrollable.scss
EugeniyKiyashko: Fixed in 33b9c36

novsstation (approved):

PR #16756, opened 2021-03-19
Renovation: Fix pager style

MikeVitik:


scss/widgets/base/_pager.scss
```diff
@@ -1,4 +1,4 @@
-@use "./mixins" as *;
+@use "./mixins"as *;
```
vconst: It is unnecessary change

vconst (approved):

PR #16758, opened 2021-03-19
Export formatter - fix LGTM warning

novsstation:

### DIFF
--- a/js/exporter/excel_format_converter.js
+++ b/js/exporter/excel_format_converter.js
@@ -23,9 +23,6 @@ const DAY_REGEXP = /E/g;
 const DO_REGEXP = /dE+/g;
 const STANDALONE_MONTH_REGEXP = /L/g;
 const HOUR_REGEXP = /h/g;
-const SLASH_REGEXP = /\//g;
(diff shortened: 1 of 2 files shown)


EugeniyKiyashko (approved):

js/exporter/excel_format_converter.js
```diff
@@ -57,16 +54,16 @@ const excelFormatConverter = {
     },
 
     _convertDateFormatToOpenXml: function(format) {
-        return format.replace(SLASH_REGEXP, '\\/').split('\'').map(function(datePart, index) {
+        return format.split('/').join('\\/').split('\'').map(function(datePart, index) {
```
IgnatovDan, line 57: Here, we hide our code from the ["A string transformer that does not replace or escape all occurrences of a meta-character may be ineffective" check](https://lgtm.com/rules/1506222917439/) because our code converts the source display format string into XLSX(OpenXml) display format string and saves it into a generated XLSX file.
The new string should be safely processed by XLSX readers because anyone can create the same XLSX file without our code and this vulnerability should be handled at the readers level.

IgnatovDan (approved):

PR #16766, opened 2021-03-22
ColorBox - clear button should set empty value if editAlphaChannel is true (T976630)

alexander-kotov-dx:

### DIFF
--- a/js/ui/color_box/color_box.js
+++ b/js/ui/color_box/color_box.js
@@ -202,13 +202,12 @@ const ColorBox = DropDownEditor.inherit({
             },
 
             onValueChanged: function({ event, value, previousValue }) {
-                if(colorUtils.makeRgba(value) === previousValue) {
-                    return;
-                }
-
                 const instantlyMode = that.option('applyValueMode') === 'instantly';
+                const realColorChanged = colorUtils.makeRgba(value) !== previousValue;
+                const changesApplied = instantlyMode || that._colorViewEnterKeyPressed;
+                const valueCleared = that._shouldSaveEmptyValue;
 
-                if(!instantlyMode && !that._colorViewEnterKeyPressed) {
+                if(!realColorChanged || !changesApplied || valueCleared) {
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


js/ui/color_box/color_box.js
```diff
@@ -302,7 +302,8 @@ const ColorBox = DropDownEditor.inherit({
 
     _renderValue: function() {
         const value = this.option('value');
-        this.option('text', this.option('editAlphaChannel') ? colorUtils.makeRgba(value) : value);
+
+        this.option('text', (this.option('editAlphaChannel') && value !== null) ? colorUtils.makeRgba(value) : value);
```
DokaRus: Could you simplify this expression? E.g. 
```suggestion
        const convertToColor = value !== null && this.option('editAlphaChannel');
        const text = convertToColor ? colorUtils.makeRgba(value) : value;
        this.option('text', text);
```

ksercs: What about `undefined`?

alexander-kotov-dx: Now the _undefined_ value sets as "rgba(0, 0, 0, 1)" in the same way as before my fix. The _undefined_ value is unexpected for the widget and i think it is no reasons to change behavior that works since 18.1.6

ksercs: I agree, thx

js/ui/color_box/color_box.js
```diff
@@ -202,13 +202,12 @@ const ColorBox = DropDownEditor.inherit({
...
-                    return;
-                }
-
                 const instantlyMode = that.option('applyValueMode') === 'instantly';
+                const realColorChanged = colorUtils.makeRgba(value) !== previousValue;
+                const changesApplied = instantlyMode || that._colorViewEnterKeyPressed;
+                const valueCleared = that._shouldSaveEmptyValue;
 
-                if(!instantlyMode && !that._colorViewEnterKeyPressed) {
+                if(!realColorChanged || !changesApplied || valueCleared) {
```
ksercs: negation of the negation seems too difficult to me. Maybe this way?
```
const isOldValue = colorUtils.makeRgba(value) === previousValue;
if(isOldValue || ... ) {
```

ksercs (dismissed):

Zedwag (dismissed):

ksercs (approved):

Zedwag (approved):

PR #16795, opened 2021-03-23
Grids: Improve typescript definitions

vconst:


js/ui/data_grid.d.ts
```diff
@@ -69,767 +69,1019 @@ import Widget, {
...
+
+
+/**
+ * @public
+ */
+export interface InitNewRowEvent<T extends GridBase> {
+  readonly component: T;
+  readonly element: TElement;
+  readonly model?: any;
+  data: any;
```
jtoming830, line 130: isn't it readonly? We use it in this way AFAIK:
`e.data.myField = . . .`, so it is only mutation, not assignment. At least, this is shown in docs. However, we pass `param.data` in our internal code (`_addRow` method), so it can be assigned.

js/ui/data_grid.d.ts
```diff
@@ -69,767 +69,1019 @@ import Widget, {
...
+}
+
+/**
+ * @public
+ */
+export interface RowInsertingEvent<T extends GridBase> {
+  readonly component: T;
+  readonly element: TElement;
+  readonly model?: any;
+  readonly data: any;
```
jtoming830: Here data is readonly, but in InitNewRowEvent it isn't. I think these two events (and RowUpdatingEvent's newData which is not readonly) should be consistent.

js/ui/data_grid.d.ts
```diff
@@ -69,767 +69,1019 @@ import Widget, {
...
+}
+
+/**
+ * @public
+ */
+export interface ToolbarPreparingEvent<T extends GridBase> {
+  readonly component: T;
+  readonly element: TElement;
+  readonly model?: any;
+  toolbarOptions: dxToolbarOptions;
```
jtoming830, line 318: Are you sure that toolbarOptions should not be readonly? We only modificate them in docs.

js/ui/data_grid.d.ts
```diff
@@ -69,767 +69,1019 @@ import Widget, {
...
-     * @type_function_param1 e:object
-     * @type_function_param1_field4 changes:Array<any>
-     * @default null
-     * @action
+     * @prevFileNamespace DevExpress.ui
      * @public
+     * @type object
      */
-    onEditCanceled?: ((e: { component?: T, element?: TElement, model?: any, changes?: Array<any> }) => void);
+    editing?: EditingBase;
```
jtoming830, line 832: editing.changes must always be array otherwise editing module will fail with null ref. So editing can not be optional before we support dynamic imports.

jtoming830: I also don't think that other options such as headerFilter, filterRow, etc. can be optional.

js/ui/data_grid.d.ts
```diff
@@ -69,767 +69,1019 @@ import Widget, {
...
-    onRowInserted?: ((e: { component?: T, element?: TElement, model?: any, data?: any, key?: any, error?: Error }) => void);
+    focusedColumnIndex?: number;
     /**
      * @docid
-     * @type_function_param1 e:object
-     * @type_function_param1_field4 data:object
-     * @type_function_param1_field5 cancel:boolean|Promise<void>
-     * @default null
-     * @action
+     * @default false
```
jtoming830, line 894: A little offtop, but still - maybe think about changing this default in future? False looks strange, we use null in such cases.

js/ui/data_grid.d.ts
```diff
@@ -1444,2238 +1303,3296 @@ export interface GridBaseOptions<T = GridBase> extends WidgetOptions<T> {
...
-     * @param2 optionName:string
-     * @param3 optionValue:any
      * @prevFileNamespace DevExpress.ui
-     * @public
+     * @type Enums.GridEnterKeyDirection
+     * @default "none"
      */
-    columnOption(id: number | string, optionName: string, optionValue: any): void;
-    /**
+    enterKeyDirection?: 'none' | 'column' | 'row'
```
jtoming830, line 1660: I am not sure that such properties should be optional. We support only three values and one of them is our default - why this value can be undefined?

js/ui/data_grid.d.ts
```diff
@@ -1444,2238 +1303,3296 @@ export interface GridBaseOptions<T = GridBase> extends WidgetOptions<T> {
...
-     * @param1 key:any
-     * @return boolean
-     * @prevFileNamespace DevExpress.ui
+     * @docid GridBaseOptions.editing.changes
+     * @default []
+     * @fires GridBaseOptions.onOptionChanged
      * @public
      */
-    isAdaptiveDetailRowExpanded(key: any): boolean;
+    changes?: Array<any>;
```
jtoming830, line 1917: Editing will fail if changes is not array and I believe that this is good behavior.

js/ui/data_grid.d.ts
```diff
@@ -1444,2238 +1303,3296 @@ export interface GridBaseOptions<T = GridBase> extends WidgetOptions<T> {
...
-     * @default null
-     * @action
+     * @type Enums.HorizontalAlignment
+     * @default undefined
+     * @acceptValues undefined
      * @prevFileNamespace DevExpress.ui
      * @public
      */
-    onEditingStart?: ((e: { component?: dxDataGrid, element?: TElement, model?: any, data?: any, key?: any, cancel?: boolean, column?: any }) => void);
+    alignment?: 'center' | 'left' | 'right' | undefined;
```
jtoming830: I think `| undefined` can be removed because `alignment` is optional.

js/ui/data_grid.d.ts
```diff
@@ -1444,2238 +1303,3296 @@ export interface GridBaseOptions<T = GridBase> extends WidgetOptions<T> {
...
+}
+
+/**
+ * @public
+ */
+export interface ContextMenuPreparingEvent {
+  readonly component: dxDataGrid;
+  readonly element: TElement;
+  readonly model?: any;
+  items?: Array<any>;
```
jtoming830, line 3335: The same as toolbarPreparing, initNewRow, etc. I think we should discuss it and make one decision for all such cases.

js/ui/data_grid.d.ts
vconst: Some users in the past wanted to assign his custom class:
`e.data.= new MyMagicClass()`
For this case data should not be readonly.

js/ui/data_grid.d.ts
vconst: I will remove readonly here.

js/ui/data_grid.d.ts
vconst: But by code, it is possible to modify it and it will work. We can discuss it.

js/ui/data_grid.d.ts
vconst: They are optional because you can define DataGrid configuration without editing option.

js/ui/data_grid.d.ts
vconst: This option is optional because you can skip definition for this option in the configuration and DataGrid gets default value for it.

js/ui/data_grid.d.ts
vconst: User can assign editing: {.allowUpdating: true } and skip changes definition.

js/ui/data_grid.d.ts
vconst: I will try to fix it.

js/ui/data_grid.d.ts
vconst: I will remove readonly here.

js/ui/data_grid.d.ts
jtoming830: Again, it's look like we need to discuss these cases and make one decision for all of them. Maybe with some restrictions but still.

jtoming830: Or leave as it is and postpone this discussion for a better time)

js/ui/data_grid.d.ts
```diff
@@ -69,767 +69,1019 @@ import Widget, {
...
+/**
+ * @public
+ */
+export interface RowUpdatedEvent<T extends GridBase> {
+  readonly component: T;
+  readonly element: TElement;
+  readonly model?: any;
+  readonly data: any;
+  readonly key: any;
+  readonly error: Error;
```
NickMitrokhin, line 243:
```suggestion
  ...
  readonly error?: Error;
```

error may not be defined

js/ui/data_grid.d.ts
NickMitrokhin: I don't remember that customers change the entire toolbarOptions object. Usually, they modify items. However, there may be cases that customers did it and did not ask any questions, because it worked.

js/ui/data_grid.d.ts
```diff
@@ -1444,2238 +1303,3296 @@ export interface GridBaseOptions<T = GridBase> extends WidgetOptions<T> {
...
+    onRowPrepared?: ((e: RowPreparedEvent) => any);
+    /**
+     * @docid
+     * @type boolean|object|Enums.Mode
+     * @default "auto"
+     * @prevFileNamespace DevExpress.ui
+     * @public
+     */
+    remoteOperations?: boolean | {
+    /**
```
NickMitrokhin, line 3958: Why is the remoteOperations object is not extracted into a certain type?

js/ui/data_grid.d.ts
```diff
@@ -1444,2238 +1303,3296 @@ export interface GridBaseOptions<T = GridBase> extends WidgetOptions<T> {
...
+     * @type_function_param1_field14 groupIndex:number
+     * @type_function_param1_field15 rowElement:dxElement
+     * @type_function_param1_field16 handled:boolean
+     * @default null
+     * @action
+     * @prevFileNamespace DevExpress.ui
+     * @public
+     */
+    onRowClick?: ((e: RowClickEvent) => any) | string;
+    /**
```
NickMitrokhin: Do we really need "string" for onRowClick?

js/ui/data_grid.d.ts
vconst: Because when we discuss improvements with Ilya, we decided that such types we skip now and will extract later.

PR #16797, opened 2021-03-23
TreeView - refactoring of the selection logic

novsstation:
1) Inline the _changeCheckBoxState method
2) Remove recursive code
3) New test cases

### DIFF
--- a/js/ui/tree_view/ui.tree_view.base.js
+++ b/js/ui/tree_view/ui.tree_view.base.js
@@ -1159,7 +1159,7 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         if(!value && this._isLastRequired(node)) {
             if(this._showCheckboxes()) {
                 const $node = this._getNodeElement(node);
-                this._getCheckBoxInstance($node).option('value', value);
+                this._getCheckBoxInstance($node).option('value', true);
             }
             return false;
         }
--- a/js/ui/tree_view/ui.tree_view.base.js
+++ b/js/ui/tree_view/ui.tree_view.base.js
@@ -1166,13 +1166,10 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
             return false;
         }
 
-        if(value === true && this._isSingleSelection()) {
-            const selectedKeys = this.getSelectedNodeKeys();
-            each(selectedKeys, (index, key) => {
-                this._dataAdapter.toggleSelection(key, false);
-                this._updateItemsUI();
-
-                this._fireItemSelectionChanged(this._getNode(key));
+        const selectedNodesKeys = this.getSelectedNodeKeys();
+        if(this._isSingleSelection() && value) {
+            each(selectedNodesKeys, (index, nodeKey) => {
+                this.unselectItem(nodeKey);
             });
         }
 
@@ -1186,7 +1183,7 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
 
     _setCheckboxValue: function($node, value) {
         const checkbox = this._getCheckBoxInstance($node);
-        checkbox && checkbox.option('value', true);
+        checkbox && checkbox.option('value', value);
     },
 
     _fireItemSelectionChanged: function(node, dxEvent) {
--- a/js/ui/tree_view/ui.tree_view.base.js
+++ b/js/ui/tree_view/ui.tree_view.base.js
@@ -101,7 +101,7 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         const checkboxInstance = this._getCheckBoxInstance($element);
         const currentState = checkboxInstance.option('value');
         if(!checkboxInstance.option('disabled')) {
-            this._updateItemSelection(!currentState, $element.find('.' + ITEM_CLASS).get(0), true, $element);
+            this._updateItemSelection(!currentState, $element.find('.' + ITEM_CLASS).get(0), true);
         }
     },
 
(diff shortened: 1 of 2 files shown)


js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -101,7 +101,7 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         const checkboxInstance = this._getCheckBoxInstance($element);
         const currentState = checkboxInstance.option('value');
         if(!checkboxInstance.option('disabled')) {
-            this._updateItemSelection(!currentState, $element.find('.' + ITEM_CLASS).get(0), true, $element);
```
novsstation, line 104: It is not used paramater

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -1158,36 +1158,46 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
             return false;
         }
 
-        const selectedNodesKeys = this.getSelectedNodeKeys();
-        if(this._isSingleSelection() && value) {
-            each(selectedNodesKeys, (index, nodeKey) => {
-                this.unselectItem(nodeKey);
+        if(value === true && this._isSingleSelection()) {
+            const selectedKeys = this.getSelectedNodeKeys();
+            each(selectedKeys, (index, key) => {
```
novsstation, line 1169: There will be only one key in single selection mode. 
Refer to this [test](https://github.com/DevExpress/DevExtreme/pull/16797/files#diff-93b90b0dccd7cf8f9cc76c2579db3ffb418b310975d50d8cfdfc21f7fc88e240R1031)

KuznetsovVN (dismissed):

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -1158,36 +1158,46 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
 
         this._dataAdapter.toggleSelection(node.internalFields.key, value);
         this._updateItemsUI();
 
+        this._fireItemSelectionChanged(node, dxEvent);
+        this._fireSelectionChanged();
+        return true;
+    },
+
+    _setCheckboxValue: function($node, value) {
```
IgnatovDan: There is the "_changeCheckBoxState: function($element) {" function with name like this function name but with different internal code
https://github.com/DevExpress/DevExtreme/blob/21_1/js/ui/tree_view/ui.tree_view.base.js#L100
It's better to have different names.

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -1158,36 +1158,46 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
+        const checkbox = this._getCheckBoxInstance($node);
+        checkbox && checkbox.option('value', value);
```
IgnatovDan: This function can do nothing in some cases
Let' rename it to "_trySetCheckBoxValue"

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -1158,36 +1158,46 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
         }
 
-        if(!value && this._isLastRequired(node)) {
+        if(value === false && this._isLastRequired(node)) {
             if(this._showCheckboxes()) {
                 const $node = this._getNodeElement(node);
-                const checkbox = this._getCheckBoxInstance($node);
-
-                checkbox && checkbox.option('value', true);
+                this._setCheckboxValue($node, true);
```
IgnatovDan: There are many checkboxes, let's rename this function to _setItemCheckboxValue

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -1158,36 +1158,46 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
 
+        this._fireItemSelectionChanged(node, dxEvent);
+        this._fireSelectionChanged();
+        return true;
+    },
+
+    _setCheckboxValue: function($node, value) {
+        const checkbox = this._getCheckBoxInstance($node);
+        checkbox && checkbox.option('value', value);
+    },
```
IgnatovDan: Will we use this function for "this._getCheckBoxInstance($parentNode).option('value', parentValue);" ?

https://github.com/DevExpress/DevExtreme/blob/21_1/js/ui/tree_view/ui.tree_view.base.js#L1227

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -1158,36 +1158,46 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
         }
 
         this._dataAdapter.toggleSelection(node.internalFields.key, value);
         this._updateItemsUI();
 
+        this._fireItemSelectionChanged(node, dxEvent);
+        this._fireSelectionChanged();
+        return true;
+    },
+
```
IgnatovDan, line 1185: Why the new '_setCheckboxValue' function is better than "this._getCheckBoxInstance($node).option('value', nodeSelection)"?

IgnatovDan (changes requested):

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -69,7 +69,13 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
 
         const select = (e) => {
             e.preventDefault();
-            this._changeCheckBoxState($(this.option('focusedElement')));
```
novsstation, line 72: I have inlined this method

js/ui/tree_view/ui.tree_view.base.js
novsstation: Previously, I think, that it is easy to read. Complexity of this method was decreased.

But in th Treeview there is another method "changeCheckBoxState" with different behavior. And as result there are 2 methods with similar name and different behavior. So, summary complexity in the Treeview was increased.

So, I removed my new method.

js/ui/tree_view/ui.tree_view.base.js
novsstation: I have removed the whole new method and inlined it's code.
Just want to note, that checking the checkbox is redundant, because there is similar code in the _updateItemsUI method without any checks. The _updateItemsUI  method is called just after this method

js/ui/tree_view/ui.tree_view.base.js
novsstation: I have removed the whole new method

js/ui/tree_view/ui.tree_view.base.js
novsstation: Yep, their behavior is different. 
So it's easy to confuse. That's why I removed both of them.

js/ui/tree_view/ui.tree_view.base.js
novsstation: Yep, I removed the whole method and rewrote code without any checks as in this sample

EugeniyKiyashko (approved):

IgnatovDan (approved):

KuznetsovVN (approved):

PR #16819, opened 2021-03-25
Html Editor - add some Quill API methods

alexander-kotov-dx:

### DIFF
--- a/js/ui/editor/editor.js
+++ b/js/ui/editor/editor.js
@@ -9,6 +9,8 @@ import ValidationEngine from '../validation_engine';
 import EventsEngine from '../../events/core/events_engine';
 import ValidationMessage from '../validation_message';
 import Guid from '../../core/guid';
+import { noop } from '../../core/utils/common';
+import { resetActiveElement } from '../../core/utils/dom';
 
 const INVALID_MESSAGE_AUTO = 'dx-invalid-message-auto';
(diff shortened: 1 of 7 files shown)


czernous-devexpress (dismissed):

js/ui/html_editor/ui.html_editor.js
```diff
@@ -611,10 +621,22 @@ const HtmlEditor = Editor.inherit({
...
         this.callBase();
 
         this._applyQuillMethod('focus');
+    },
+
+    blur: function() {
+        const activeElement = domAdapter.getActiveElement();
+        if(this.$element().is(activeElement) || this.$element().find(activeElement).length) {
+            resetActiveElement();
+        }
```
DokaRus: Let's unify it? Extract the same method from texteditor -> editor. It is worth separating the condition into a separate function to make it easier to overload

ksercs (dismissed):

czernous-devexpress (dismissed):

js/ui/editor/editor.js
```diff
@@ -359,6 +363,12 @@ const Editor = Widget.inherit({
         }
     },
 
+    blur: function() {
```
DokaRus, line 366: Add a test (for the editor component) that checks that we do not blurring the active item when the `blur` method is called

DokaRus (approved):

ksercs (approved):

PR #16846, opened 2021-03-26
Renovation: use updated generators

LazyLahtak:


js/renovation/ui/scheduler/workspaces/base/__tests__/row.test.tsx
```diff
@@ -64,7 +64,7 @@ describe('RowBase', () => {
           props: {
             leftVirtualCellWidth: 100,
             rightVirtualCellWidth: 200,
-            children: <td className="child" width={50} />,
+            children: <td className="child" style={{ width: 50 }} />,
```
vconst, line 67: Why tests are changed?

package.json
```diff
@@ -29,9 +29,9 @@
   "license": "SEE LICENSE IN README.md",
   "dependencies": {
     "@babel/runtime": "^7.12.1",
+    "@devextreme/vdom": "^1.0.2",
```
vconst, line 32: Do we use semver for new packages? If some breaking change is happened in generator, will version become 2.0.0?

LazyLahtak: Yes, we want to avoid breaking changes in minor or patch. If we has some breaking cahanges, we will release a new major.
New features without BC we will release in minor, if we have fixes without features - it will be patch

js/renovation/ui/scheduler/workspaces/base/__tests__/row.test.tsx
LazyLahtak: For some reason typescript didn't check this test earlier.
Now, typescript thrown an error, that `width` property doesn't exists in `td`.
@AryamnovEugeniy wanted to recheck this test.

vconst (dismissed):

AryamnovEugeniy (approved):

PR #16916, opened 2021-04-01
TypeScript declarations: fix callback return types

viterobk:

### DIFF
--- a/js/animation/fx.d.ts
+++ b/js/animation/fx.d.ts
@@ -19,7 +19,7 @@ export interface animationConfig {
      * @prevFileNamespace DevExpress.animation
      * @public
      */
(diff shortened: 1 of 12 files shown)


DokaRus (approved):
LGTM

nightskylark (approved):

nightskylark:
I'll leave these links here just in case somebody will try to find out why we are sure that these methods do not have to return anything.
beforeSend
https://github.com/DevExpress/DevExtreme/blob/21_1/js/data/odata/utils.js#L162
errorHandler
https://github.com/DevExpress/DevExtreme/blob/21_1/js/data/array_query.js#L497
https://github.com/DevExpress/DevExtreme/blob/21_1/js/data/remote_query.js#L27

PR #16918, opened 2021-04-01
Fix base hover color in Bootstrap theme - T982896

OlegKipchatov:


themebuilder-scss/src/data/bootstrap-metadata/bootstrap-metadata.ts
```diff
@@ -14,7 +14,7 @@ export default {
   'base-danger': '@brand-danger',
 
   'base-link-color': '@link-color',
-  'base-hover-color': '@table-bg-hover',
+  'base-hover-color': '@table-bg-accent',
```
babich-a: Is this right? Should we use table accent bg as hover?

babich-a (changes requested):

themebuilder-scss/src/data/bootstrap-metadata/bootstrap-metadata.ts
OlegKipchatov: No...
I suggest replace on '@text-color', because Bootstrap4 inherit list: $table-hover-color -> $table-color -> $body-color

groshenkovamarina (dismissed):

groshenkovamarina (approved):

babich-a (approved):

PR #16967, opened 2021-04-07
Update ui.scheduler.appointments.strategy.horizontal_month_line.js

fullstacksk:
Edited _getDurationInDays function to return always ZERO_APPOINTMENT_DURATION_IN_DAYS so that appointment width will be limited to only one cell in month view also


andrewmakarov:
Hi @fullstacksk ,
Thank you for taking part in developing our product. We researched your PR and found that it breaks the appointment's rendering when the appointment's duration is more than one day. You can see the expected and actual results in the screenshot below:
Screenshot 2021-04-28 at 14 08 14

Another important requirement for PR is test coverage. All current tests should pass, and a new test for the current use case should be added. As we see, your PR doesn't contain tests, and our current tests are not passed.

As far as I understand, you faced some issues when trying to integrate our Scheduler in your application. If it is so, please submit a ticket to our [Support Center](http://devexpress.com/ask) describing your use case and the limitations you encounter. Our Support Team will do their best to help you.

PR #16990, opened 2021-04-08
Form - fix invalid size of inner widgets, after aligning labels (T986577)

novsstation:
Bugfix T986577

### DIFF
--- a/js/ui/form/ui.form.js
+++ b/js/ui/form/ui.form.js
@@ -9,7 +9,7 @@ import { inArray } from '../../core/utils/array';
 import { extend } from '../../core/utils/extend';
 import { isEmpty } from '../../core/utils/string';
 import browser from '../../core/utils/browser';
-import { triggerResizeEvent, triggerShownEvent } from '../../events/visibility_change';
+import { triggerShownEvent } from '../../events/visibility_change';
 import { getPublicElement } from '../../core/element';
 import messageLocalization from '../../localization/message';
 import Widget from '../widget/ui.widget';
@@ -60,8 +60,6 @@ import {
     FIELD_ITEM_CONTENT_CLASS,
     FORM_VALIDATION_SUMMARY,
     ROOT_SIMPLE_ITEM_CLASS } from './constants';
-
-import { TOOLBAR_CLASS } from '../toolbar/constants';
 
 const WIDGET_CLASS = 'dx-widget';
 const FOCUSED_STATE_CLASS = 'dx-state-focused';
@@ -458,8 +456,6 @@ const Form = Widget.inherit({
             items: this.option('items'),
             inOneColumn
         });
-
-        triggerResizeEvent(this.$element().find(`.${TOOLBAR_CLASS}`));
     },
 
     _clean: function() {
--- a/js/ui/form/ui.form.js
+++ b/js/ui/form/ui.form.js
@@ -9,7 +9,7 @@ import { inArray } from '../../core/utils/array';
 import { extend } from '../../core/utils/extend';
 import { isEmpty } from '../../core/utils/string';
 import browser from '../../core/utils/browser';
-import { triggerShownEvent } from '../../events/visibility_change';
+import { triggerResizeEvent, triggerShownEvent } from '../../events/visibility_change';
 import { getPublicElement } from '../../core/element';
 import messageLocalization from '../../localization/message';
 import Widget from '../widget/ui.widget';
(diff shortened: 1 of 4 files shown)


testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -1407,6 +1411,28 @@ QUnit.test('optional mark aligned', function(assert) {
...
+                }
+            }, {
+                label: { text: 'Very very long text' }
+            } ]
+        });
+
+        const toolbar = $form.find(`.${TOOLBAR_CLASS}`).eq(0).dxToolbar('instance');
+        const menuItems = toolbar._getMenuItems();
+
+        assert.equal(menuItems.length, alignItemLabels ? 3 : 0, 'italic, strike and underline items are located in menu if alignItemLabels is enabled');
```
IgnatovDan: >alignItemLabels ? 3 : 0

It's not clear why it should be '3'? Can HtmlEditor devs remove/redesign these items without failures in this test?
Scenario looks similar to https://github.com/DevExpress/DevExtreme/blob/20_1/testing/tests/DevExpress.ui.widgets/drawer.tests.js#L165

IgnatovDan (changes requested):

js/ui/form/ui.form.js
```diff
@@ -456,6 +458,8 @@ const Form = Widget.inherit({
             items: this.option('items'),
             inOneColumn
         });
+
+        triggerResizeEvent(this.$element().find(`.${TOOLBAR_CLASS}`));
```
IgnatovDan, line 462: Why dxResize is called for TOOLBAR_CLASS elements only?
Why dxResize is called for all nested TOOLBAR_CLASS elements?

IgnatovDan (changes requested):

js/ui/form/ui.form.js
IgnatovDan: Why dxResize is not called for item' DOM element when label size is changed at https://github.com/DevExpress/DevExtreme/blob/21_1/js/ui/form/ui.form.js#L207

> $labelTexts[i].style.width = maxWidth + 'px';

IgnatovDan (changes requested):

testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -1407,6 +1411,28 @@ QUnit.test('optional mark aligned', function(assert) {
     assert.ok($optionalLabel.position().left < $optionalMark.position().left, 'optional mark should be after of the text');
 });
 
+[true, false].forEach(alignItemLabels => {
```
IgnatovDan: imho it's easier to read two tests than one test inside "[true, false].forEach"

IgnatovDan (changes requested):

js/ui/form/ui.form.js
novsstation: > Why dxResize is called for TOOLBAR_CLASS elements only?
> Why dxResize is called for all nested TOOLBAR_CLASS elements?

Initialy, I called for the all dx-widgets - https://github.com/DevExpress/DevExtreme/pull/16990/commits/99493ee654fff7372ae4207742edf6c7fd7070f7

But after that, I reviewed our [documentation](https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxForm/Item_Types/SimpleItem/#editorType) and found, that it is redundant. It seems, that there is only one widget, which has custom resizing logic.

So, I decided to call it for toolbars only, which are placed inside the form widget (even for nested). Because if some widget has inner toolbar, it's necessary to call the dxresize on the specific toolbar container, not the item container. Because this event is not propagated to the inner elements.

novsstation: > Why dxResize is not called for item' DOM element when label size is changed at

Because, calling the dxresize on the DOM node, doesn't refer to the calling _updateDimensions of the toolbar.
Seems, this events is not propagate to the inner elements.  So, calling the dxresize on the parent nodes will not work.

image

That's why I call it only on the elements with dx-toolbar class

image


> when label size is changed at

Because the '_applyLabelsWidthByCol' method  is called multiple times from multiple places

image

So, this method can be called in the middle of the form rendering process.

That's why I call 'dxresize' only once after all labels are aligned at the end of the '_alignLabels' method

testing/tests/DevExpress.ui.widgets.form/form.tests.js
novsstation: Yep, fixed - https://github.com/DevExpress/DevExtreme/pull/16990/commits/318a22fb526f356c174bda8ef4caac529e3ae276

EugeniyKiyashko (dismissed):

EugeniyKiyashko (dismissed):

js/ui/form/ui.form.js
IgnatovDan: ok, we can make a 'spot' change for toolbar only and at that moment only and reinvestigate it after a new ticket will be posted

IgnatovDan (approved):

EugeniyKiyashko (approved):

PR #17023, opened 2021-04-09
CheckBox and Button: onContentReady now strikes after changing options which affects markup

Zedwag:

### DIFF
--- a/js/renovation/ui/button.tsx
+++ b/js/renovation/ui/button.tsx
@@ -164,9 +164,7 @@ export class Button extends JSXComponent(ButtonProps) {
     const {
       onContentReady, text, icon, template,
     } = this.props;
-    ((x): unknown => x)(text);
-    ((y): unknown => y)(icon);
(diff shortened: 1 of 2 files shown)


js/renovation/ui/button.tsx
```diff
@@ -161,8 +161,12 @@ export class Button extends JSXComponent(ButtonProps) {
...
     //       property upon which onContentReady depends
     //       (for example, text, icon, etc)
-    const { onContentReady } = this.props;
-
+    const {
+      onContentReady, text, icon, template,
+    } = this.props;
+    ((x): unknown => x)(text);
+    ((y): unknown => y)(icon);
+    ((z): unknown => z)(template);
```
ksercs: Let's refactor this way:

`((...args): unknown => args)(text, icon, template);`

ksercs (approved):

ksercs:
We're gonna get rid of contentReady in declaration. So we'll implement another solution later

PR #17028, opened 2021-04-12
ContextMenu - A browser's context menu appears above the ContextMenu on desktop touch screens (T983617)

KuznetsovVN:


testing/tests/DevExpress.ui.events/contextmenu.tests.js
```diff
@@ -29,6 +29,23 @@ QUnit.test('contextmenu should be fired on hold event', function(assert) {
     $element.trigger($.Event(holdEvent.name, { pointerType: 'touch' }));
 });
 
+QUnit.test('default\'s contextmenu should be prevented on hold event', function(assert) {
+    if(!support.touch) {
+        assert.ok(true);
+        return;
+    }
+    const $element = $('#element');
```
IgnatovDan: assert.expect( 2 ); ?
Event handlers may be called or may be not so assert may not be called.
Or, we can rearrange asserts: introduce locals and move all asserts at the end of function.

testing/tests/DevExpress.ui.events/contextmenu.tests.js
```diff
@@ -96,3 +113,35 @@ QUnit.test('contextmenu should be unsubscribed on unsubscribing contextmenu', fu
 
     assert.equal(holdEvents.length, 0, 'contextmenu event handler was removed');
 });
+
+[
+    { isSimulator: true, description: 'default\'s contextmenu should be prevented on dxhold in simulator' },
+    { isSimulator: false, description: 'default\'s contextmenu should not be prevented on dxhold' }
+].forEach(function(config) {
```
IgnatovDan: imho it's easier to read two tests than one test with conditions inside "[].foreach"

testing/tests/DevExpress.ui.events/contextmenu.tests.js
```diff
@@ -96,3 +113,35 @@ QUnit.test('contextmenu should be unsubscribed on unsubscribing contextmenu', fu
 
     assert.equal(holdEvents.length, 0, 'contextmenu event handler was removed');
 });
+
+[
+    { isSimulator: true, description: 'default\'s contextmenu should be prevented on dxhold in simulator' },
+    { isSimulator: false, description: 'default\'s contextmenu should not be prevented on dxhold' }
```
IgnatovDan: >default\'s contextmenu should not be prevented on dxhold

why? 
"DevExtreme menu appears while I hold my finger on the screen. As soon as I remove my finger - the browser menu appears" - it looks like a "dxhold" occurs (HOLD_TIMEOUT = 750;) and dxMenu was shown and system menu was shown somewhen after that moment (on_up?)

IgnatovDan (changes requested):

js/events/contextmenu.js
```diff
@@ -36,6 +36,9 @@ const ContextMenu = Class.inherit({
     },
 
     _contextMenuHandler: function(e) {
+        if(touch || devices.isSimulator()) {
+            e.preventDefault();
```
IgnatovDan, line 41: I am not sure that we should prevent each 'contextMenu' event on entire element that was targeted by our dxcontextmenu event because our event can be cancelled ("e.cancel = true;") and our code will block browser' context menu without a reason.

IgnatovDan (changes requested):

js/events/contextmenu.js
KuznetsovVN: I checked this code "e.cancel = true" instead of "e.preventDefault ()" but it didn't canceled the browser menu appears.

EugeniyKiyashko (dismissed):

novsstation (dismissed):

IgnatovDan (approved):

novsstation (approved):

EugeniyKiyashko (approved):

PR #17056, opened 2021-04-13
CheckBox: add getter for viewModel.target.current for correct angular component generating

Zedwag:


js/renovation/ui/check_box.tsx
```diff
@@ -266,6 +266,10 @@ export class CheckBox extends JSXComponent(CheckBoxProps) {
     return allValidationErrors;
   }
 
+  get targetCurrent(): HTMLDivElement | null {
```
ksercs: We should also add test for this getter

ksercs (dismissed):

pavelgruba (approved):

ksercs (approved):

PR #17086, opened 2021-04-15
Renovated component: options set to undefined should fall it back to default options considering defaultOptionRules

ksercs:
…


js/renovation/component_wrapper/component.ts
```diff
@@ -158,23 +151,53 @@ export default class ComponentWrapper extends DOMComponent {
     return this._elementAttr;
   }
 
+  _setDefaultOptionValue(options, defaultValueGetter) {
```
pavelgruba: This method does not use `this`, so why is it an instance method now?

ksercs: just forgot to move it out after some changes, thx

ksercs:
waits for generator improving

ksercs:
will be merged only in 21.2

PR #17154, opened 2021-04-20
Overlay: remove hacks for Android 4.3-

Zedwag:


DokaRus (commented):
Android - OK
About Safari - let's discuss it in the channel with the POs (perhaps we can reconsider supported iOS and Android versions)

ksercs:
We should also remove appropriate tests in popup.test

ksercs (approved):

alexander-kotov-dx (approved):

PR #17167, opened 2021-04-20
VirtualScrolling - Disables loading logic for _virtualScrollController

NickMitrokhin:

### DIFF
--- a/js/ui/grid_core/ui.grid_core.virtual_scrolling.js
+++ b/js/ui/grid_core/ui.grid_core.virtual_scrolling.js
@@ -164,15 +164,15 @@ const VirtualScrollingDataSourceAdapterExtender = (function() {
             }
         },
         _handleLoadError: function() {
-            this._isLoading = false;
-            this.loadingChanged.fire(false);
+            if(!this.option(NEW_SCROLLING_MODE)) {
+                this._isLoading = false;
+                this.loadingChanged.fire(false);
+            }
 
             this.callBase.apply(this, arguments);
         },
         _handleDataChanged: function(e) {
             if(this.option(NEW_SCROLLING_MODE)) {
-                this._items = this._dataSource.items();
-                this.endPageIndex(this._dataSource.pageIndex());
                 this.callBase.apply(this, arguments);
                 return;
             }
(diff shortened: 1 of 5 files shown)


js/ui/grid_core/ui.grid_core.virtual_scrolling.js
```diff
@@ -671,16 +683,35 @@ const VirtualScrollingRowsViewExtender = (function() {
...
+            const hasEnabledOperationType = function(types) {
+                let result = false;
+                if(types) {
+                    for(const t in types) {
+                        if(types[t]) {
+                            result = true;
+                            break;
+                        }
+                    }
+                }
```
vconst: types = types || {}
return Object.keys(types).some(name => types[name]);

js/ui/grid_core/ui.grid_core.virtual_scrolling.js
```diff
@@ -671,16 +683,35 @@ const VirtualScrollingRowsViewExtender = (function() {
...
+
+                return result;
+            };
+
+            if(this.option(NEW_SCROLLING_MODE) && isLoading && operationTypes) {
+                const loadingOperationTypes = dataController.dataSource()?.loadingOperationTypes?.();
+                if(loadingOperationTypes?.paging || loadingOperationTypes?.fullReload || !hasEnabledOperationType(loadingOperationTypes)) {
```
vconst, line 710: fullReload should be removed in the future because it exists on refresh

js/ui/grid_core/ui.grid_core.virtual_scrolling.js
```diff
@@ -161,14 +164,19 @@ const VirtualScrollingDataSourceAdapterExtender = (function() {
             }
         },
         _handleLoadError: function() {
-            const that = this;
+            this._isLoading = false;
```
vconst: Do not modify _isLoading and do not call loadingChanged in new mode

js/ui/grid_core/ui.grid_core.virtual_scrolling.js
```diff
@@ -161,14 +164,19 @@ const VirtualScrollingDataSourceAdapterExtender = (function() {
...
 
-            that._isLoading = false;
-            that.loadingChanged.fire(false);
-
-            that.callBase.apply(that, arguments);
+            this.callBase.apply(this, arguments);
         },
         _handleDataChanged: function(e) {
+            if(this.option(NEW_SCROLLING_MODE)) {
+                this._items = this._dataSource.items();
```
vconst: Try change items method to return _dataSource.items() directly instead of assign _items here.

js/ui/grid_core/ui.grid_core.virtual_scrolling.js
```diff
@@ -161,14 +164,19 @@ const VirtualScrollingDataSourceAdapterExtender = (function() {
...
+                this.endPageIndex(this._dataSource.pageIndex());
```
vconst: Why is _lastPageIndex assigned here?

js/ui/grid_core/ui.grid_core.virtual_scrolling.js
NickMitrokhin: Yes, I'm planning to reconsider this condition after improving loading logic.

js/ui/grid_core/ui.grid_core.virtual_scrolling.js
NickMitrokhin: This already implemented. The code is redundant here.

js/ui/grid_core/ui.grid_core.virtual_scrolling.js
NickMitrokhin: endPageIndex is used in the editing module. So, the endPageIndex should return dataSource.pageIndex() in a new scrolling mode.

vconst (dismissed):

Alyar666 (dismissed):

PR #17227, opened 2021-04-26
Replace all references to the version with 21.2.0, get rid of shippable.yml

churkin:


js/bundles/modules/data.js
```diff
@@ -9,7 +9,7 @@ Object.defineProperty(DevExpress.data, 'errorHandler', {
         return require('../../data/errors').errorHandler;
     },
     set: function(value) {
-        errors.log('W0003', 'DevExpress.data', 'errorHandler', '21.1', 'Use the \'setErrorHandler\' method instead');
+        errors.log('W0003', 'DevExpress.data', 'errorHandler', '21.2', 'Use the \'setErrorHandler\' method instead');
```
babich-a: Is this a version when a method/property was deprecated? Should we change it?

testing/testcafe/tests/scheduler/dragAndDrop/outlookDragging/base.ts
```diff
@@ -124,24 +124,24 @@ test('Basic drag-n-drop movements from tooltip in month view', async (t) => {
 }).before(() => createWidget('dxScheduler', {
   dataSource: [{
     text: 'Appointment 1',
-    startDate: new Date(2021, 4, 26, 9, 30),
-    endDate: new Date(2021, 4, 27, 12, 0),
+    startDate: new Date(2021, 2, 31, 9, 30),
```
babich-a, line 127: Why we need to change tests?

js/bundles/modules/data.js
churkin: fixed

babich-a (approved):

PR #17230, opened 2021-04-26
Form - changing visibility option inside beginUpdate/endUpdate throws error (T990701)

novsstation:
Bugfix T990701

### DIFF
--- a/js/ui/form/ui.form.js
+++ b/js/ui/form/ui.form.js
@@ -996,7 +996,9 @@ const Form = Widget.inherit({
                         const layoutManagerItems = layoutManager.option('items');
                         formItems.forEach((item, index) => {
                             const layoutItem = layoutManagerItems[index];
-                            layoutItem.visibleIndex = item.visibleIndex;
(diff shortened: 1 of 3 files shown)


js/ui/form/ui.form.js
```diff
@@ -991,6 +991,10 @@ const Form = Widget.inherit({
                     }
                 }
                 if(optionName === 'visible') { // T874843
+                    if(layoutManager.option(fullOptionName) === value) {
```
IgnatovDan: no
see https://hg/mobile/rev/85c90979eb7c

IgnatovDan (changes requested):

js/ui/form/ui.form.js
novsstation: Yep, you are absolutely right.
Prepared a new variant of fix - more simple

EugeniyKiyashko (approved):

IgnatovDan (approved):

PR #17245, opened 2021-04-26
Replace DxPromise type with PromiseLike where possible

IlyaKhD:
We support JQuery v2.1 - v2.2, but `JQueryPromise` got `catch` method in v3.0.

If a project uses JQuery v2.1 - v2.2. promises created outside DevExtreme don't meet `Promise` interface, so we should use `PromiseLike` that is less strict.

Values of both `Promise` and `JQueryPromise` types are assignable to `PromiseLike` props.


alexanderPolosatov (commented):
LGTM

nightskylark (approved):

PR #17338, opened 2021-04-30
Overlays: refactor accessors

Zedwag:
Now overlay have public method `$wrapper()` that returns `this._$wrapper` to use from outside, instead of code like $(overlay.content()).parent() in tests. Also method `_wrapper` was deleted, because there is no need of private accessor, and it was used only in couple lines.

Method `overlayContent` is replaced with `$overlayContent` for Popup for following the same method name rules as other accessors.

Method `_container` was the same as `$content` for Overlay and `$overlayContent` for Popup. So it was repleced with these two methods.

Also there is method `content` that is equals to `$content().get(0)`, but it using for overlays `content` option, so it cannot be deleted. However, strange code like `$(overlay.content())` in tests was replaced with `overlay.$content()`


js/ui/action_sheet.js
```diff
@@ -191,7 +191,7 @@ const ActionSheet = CollectionWidget.inherit({
 
     _renderPopupTitle: function() {
         this._mapPopupOption('showTitle');
-        this._popup && this._popup._wrapper().toggleClass(ACTION_SHEET_WITHOUT_TITLE_CLASS, !this.option('showTitle'));
+        this._popup && this._popup._$wrapper.toggleClass(ACTION_SHEET_WITHOUT_TITLE_CLASS, !this.option('showTitle'));
```
ksercs: y do we use private data here, not public method as [here](https://github.com/DevExpress/DevExtreme/pull/17338/files#diff-12d9304a47fe06bdb4ab862be69b15c48007d6ad6e44fb2d3d82407969a9b7e6R411)?

Zedwag: actually i'm still not sure should we use `_$wrapper` or `$wrapper()` in files that inherits overlays. There are both ways in different files.

js/ui/popup.js
```diff
@@ -233,7 +233,7 @@ const Popup = Overlay.inherit({
         this.callBase();
 
         this.$element().addClass(POPUP_CLASS);
-        this._wrapper().addClass(POPUP_WRAPPER_CLASS);
+        this._$wrapper.addClass(POPUP_WRAPPER_CLASS);
```
ksercs, line 236: Is it overlay wrapper? We should use public `$wrapper` method in this case

js/ui/action_sheet.js
ksercs: Let's use `$wrapper()` everywhere to resolve this inconsistent and decrease probability to broke something in the future if `$wrapper` method changes

Zedwag: Ok

Zedwag: Done

js/ui/date_box/ui.date_view.js
```diff
@@ -137,7 +137,7 @@ const DateView = Editor.inherit({
                 }
             });
         });
-        that._$rollersContainer.appendTo(that._wrapper());
+        that._$rollersContainer.appendTo(that._$wrapper);
```
ksercs: why no public method?

Zedwag: that is 'date_view' own method. Actually i changed it accidentally and forget to return as it was:) but there is no difference

ksercs: let's retrieve it

Zedwag: Done

ksercs:
Let's get rid of `_$content` and `_$popupContent` using in all overlay-based widgets to avoid misunderstanding in the future. For example, `_$content` is used in popover now. But it's not a content, it's an overlay content and it should be replaced with `$overlayContent()`

ksercs:
Seems like we should also rename `_$contentWrapper` as `_$loadPanelContentWrapper` in LoadPanel to improve a bit this strange code: `this._$contentWrapper.appendTo(this._$content);`

js/ui/overlay/ui.overlay.js
```diff
@@ -996,7 +992,7 @@ const Overlay = Widget.inherit({
     },
 
     _getDragTarget: function() {
-        return this.$content();
+        return this._$content;
```
ksercs: Seems like this change can break drag in popup

ksercs:
What about 538 row in context menu?

ksercs:
Skipped files to refactor:
- ui.context_menu
- dropDownButton.markup.tests
- dropDownButton.tests
- dropDownMenu.tests

js/ui/overlay/ui.overlay.js
Zedwag: Done

AryamnovEugeniy (approved):

vladkovl (approved):

DokaRus (approved):

ksercs (approved):

alexander-kotov-dx (approved):

novsstation (approved):

vconst (approved):

PR #17351, opened 2021-05-04
ResponsiveBox - Render Box widget

novsstation:


js/renovation/ui/box/box.tsx
```diff
@@ -3,8 +3,14 @@ import {
 } from '@devextreme-generator/declarations';
 import { Widget } from '../common/widget';
 import { BoxProps } from './box_props';
+import { combineClasses } from '../../utils/combine_classes';
 
-export const viewFunction = (): JSX.Element => (<Widget />);
+export const viewFunction = (): JSX.Element => {
+  const cssClasses = combineClasses({ 'dx-box dx-box-flex': true });
```
EugeniyKiyashko: viewFunction should be clear as possible. All operation that should be execute we should transfer to getters.

js/renovation/ui/responsive_box/_toDelete/responsive_box.js
```diff
@@ -112,18 +111,6 @@ const ResponsiveBox = CollectionWidget.inherit({
...
     _initMarkup: function() {
         this.callBase();
-        // NOTE: Fallback box strategy
-        this._updateRootBox();
-    },
-
-    _updateRootBox: function() {
-        clearTimeout(this._updateTimer);
-
-        this._updateTimer = setTimeout((function() {
```
EugeniyKiyashko, line 122: Why we don't need timeout anymore?

novsstation: This timer is necessary only for Fallback strategy ('dxupdate' event is used only in Fallback strategy), which will be removed soon.
I remove timer in this request, because it uses the root property, which I removed in this PR

js/renovation/ui/box/box.tsx
novsstation: Yep, fixed

EugeniyKiyashko (approved):

KuznetsovVN (approved):

PR #17408, opened 2021-05-07
TreeView - fix selectAll checkbox doesn't not work (T996410)

novsstation:
Bugfix (T996410)

### DIFF
--- a/js/ui/tree_view/ui.tree_view.base.js
+++ b/js/ui/tree_view/ui.tree_view.base.js
@@ -1028,11 +1028,11 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         this._$selectAllItem = $('<div>').addClass(SELECT_ALL_ITEM_CLASS);
 
         const value = this._dataAdapter.isAllSelected();
-        const checkBox = this._createComponent(this._$selectAllItem, CheckBox, {
+        this._createComponent(this._$selectAllItem, CheckBox, {
             value: value,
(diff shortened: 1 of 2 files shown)


js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -1028,11 +1028,11 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
 
         const value = this._dataAdapter.isAllSelected();
-        this._createComponent(this._$selectAllItem, CheckBox, {
+        const checkBox = this._createComponent(this._$selectAllItem, CheckBox, {
             value: value,
-            text: this.option('selectAllText'),
-            onValueChanged: this._onSelectAllCheckboxValueChanged.bind(this)
+            text: this.option('selectAllText')
         });
+        checkBox.on('valueChanged', this._onSelectAllCheckboxValueChanged.bind(this));
```
IgnatovDan: It's not clear why we have to change to this approach.

IgnatovDan (changes requested):

js/ui/tree_view/ui.tree_view.base.js
novsstation: Replaced with old approach with using of the option method

KuznetsovVN (approved):

IgnatovDan (approved):

EugeniyKiyashko (approved):

PR #17452, opened 2021-05-12
datagrid - fixes focusing cells in fixed banded columns (T996394)

NickMitrokhin:


vconst (commented):
Why is test added to focus.integration tests instead of columnFixing.integration?
Usually test should be written for module which code is modified.

Alyar666 (approved):

vconst (approved):

PR #17460, opened 2021-05-12
Renovation fix button Qunit

MikeVitik:


churkin (dismissed):

testing/tests/Renovation/compatibility.test.js
```diff
@@ -67,7 +67,7 @@ QUnit.module('Mandatory component setup', {
             + '<root {...viewModel.restAttributes} />';
 
             $('#component')[`dx${meta.name}`]({
-                'data-custom-option': 'custom-value',
+                elementAttr: { 'data-custom-option': 'custom-value' },
```
pavelgruba: ElementAttr is the widget's standard option, but this test should ensure that any other non-standard option is passed as-is to underlying component.

pavelgruba (approved):

PR #17559, opened 2021-05-18
scrollview: public update() should call recalculate element sizes

EugeniyKiyashko:


KuznetsovVN (dismissed):

js/renovation/ui/scroll_view/scrollable_native.tsx
```diff
@@ -894,7 +892,9 @@ export class ScrollableNative extends JSXComponent<ScrollableNativePropsType>()
       return false;
     }
 
-    this.update();
+    if (!this.props.updateManually) {
```
IgnatovDan: Condition seems to be incorrect:
- the 'validate' method should do validation without additional actions, additional actions should be out of this method
- tryGetAllowedDirection/isScrollingOutOfBound are called after 'update' is completed and their result can be changed if 'update' was not called.

IgnatovDan: Can we remove it?
hg/mobile/rev/c52851107602

IgnatovDan: Can we remove updateManually?
hg/mobile/rev/c52851107602

IgnatovDan (dismissed):

js/renovation/ui/scroll_view/scrollable_native.tsx
EugeniyKiyashko: fixed in c4f2bdc

novsstation (dismissed):

KuznetsovVN (dismissed):

novsstation (approved):

PR #17601, opened 2021-05-21
Renovation: Make `children` render in right order if placed with other nodes

LazyLahtak:


testing/tests/Renovation/button.tests.js
```diff
@@ -114,6 +114,19 @@ QUnit.test('should render submit input with custom template', function(assert) {
     assert.strictEqual(element.find('.dx-button-submit-input').length, 1, 'render submit input');
 });
 
+QUnit.test('should not fall if template returned parent node', function(assert) {
```
pavelgruba:
```suggestion
QUnit.test('should not fail if template returned parent node', function(assert) {
```

pavelgruba (dismissed):

pavelgruba (approved):

PR #17691, opened 2021-05-29
Fix obsolete style rules (T1002219)

OlegKipchatov:


babich-a (changes requested):
https://sass-lang.com/documentation/operators/numeric#division
math.div is supported starting from 1.33.0
So we need to update sass in these places:
https://github.com/DevExpress/DevExtreme/blob/21_1/package.json#L161
https://github.com/DevExpress/DevExtreme/blob/21_1/themebuilder-scss/package.json#L21
https://github.com/DevExpress/DevExtreme/blob/21_1/themebuilder-scss/dart-compiler/pubspec.yaml#L5

But! The latest place is very dangerous. 1.33 is not working in dart compiler. We should research why and update.

babich-a (approved):

PR #17699, opened 2021-05-31
Add explicit `any` in type definitions

alexserov:
This fixes the T1001991

### DIFF
--- a/ts/dx.all.d.ts
+++ b/ts/dx.all.d.ts
@@ -674,7 +674,7 @@ declare module DevExpress {
      */
     resetOption(optionName: string): void;
 
-    _options: { silent(path, value): void };
+    _options: { silent(path: any, value: any): void };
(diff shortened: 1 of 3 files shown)


ts/dx.all.d.ts
```diff
@@ -674,7 +674,7 @@ declare module DevExpress {
      */
     resetOption(optionName: string): void;
 
-    _options: { silent(path, value): void };
+    _options: { silent(path: any, value: any): void };
```
nightskylark, line 677: Not sure `dx.all` is allowed to be modified manually. @IlyaKhD, is it OK?

IlyaKhD: Yep, there is a check for TS-bundle being up-to-date - no difference how it was updated, by tool, or by hand

IlyaKhD (approved):

PR #17737, opened 2021-06-01
Overlay.baseZIndex: fix ES module named export

ksercs:


DokaRus (dismissed):

Zedwag (dismissed):

js/ui/overlay.js
```diff
@@ -1,2 +1,7 @@
 import Overlay from './overlay/ui.overlay';
-export default Overlay;
+
+const baseZIndex = Overlay.baseZIndex;
+export {
+    Overlay,
```
nightskylark, line 5: Do we need this export? Overlay is private and all internal imports uses the `overlay/ui.overlay' module now. So who is it for?

js/ui/drawer/ui.drawer.rendering.strategy.overlap.js
```diff
@@ -2,7 +2,7 @@ import { animation } from './ui.drawer.animation';
 import DrawerStrategy from './ui.drawer.rendering.strategy';
 import $ from '../../core/renderer';
 import { move } from '../../animation/translator';
-import Overlay from '../overlay';
+import Overlay from '../overlay/ui.overlay';
```
vconst: Why is not `import { Overlay } from '../overlay'`? When we renovate Overlay only '../overlay' will be renovated in tests. The '../overlay/ui.overlay' import will return old jquery implementation.

Zedwag (dismissed):

js/ui/overlay.js
ksercs: It was left for consistence with another editors to support flow with auto-replace to renovated component. But I forgot to use it in some places. fixed

js/ui/drawer/ui.drawer.rendering.strategy.overlap.js
ksercs: It does not work in renovated wrappers. But I agree that we should use this approach in other cases

Zedwag (approved):

DokaRus (approved):

vconst (approved):

PR #17750, opened 2021-06-02
Extract private augmentation for Component

IlyaKhD:


js/core/component.js
```diff
@@ -381,4 +381,11 @@ const Component = Class.inherit({
     }
 });
 
-export default Component;
```
alexserov, line 384: Should we announce a breaking change for this?

IlyaKhD: I guess no, `Component` is not a public class, there shouldn't be imports in the user code

MikeVitik (approved):

PR #17773, opened 2021-06-03
Add Xor type

IlyaKhD:
This is an implementation for exact props check, which is currently not supported: https://github.com/Microsoft/TypeScript/issues/12936. Using this `Xor` type will prevent users from this:

```ts
// we expect either prop1 + prop2 or type3+prop4, not both sets at a time
type Expected =
| {
  prop1: number;
  prop2: number;
} 
| {
  prop3: string;
  prop4: string;
}
```

```ts
// user defined value contains all the props
const value: Expected = {
  prop1: 123,
  prop2: 456,
  prop3: 'abc',
  prop4: 'def',
}
```


nightskylark:
How about to add a little description? :)

IlyaKhD:
> How about to add a little description? :)

Sure, done

alexserov (dismissed):

nightskylark (dismissed):

nightskylark (approved):

PR #17798, opened 2021-06-07
Renovation: enable 'prefer-nullish-coalescing' eslint rule

ksercs:


AryamnovEugeniy (approved):

DokaRus (approved):

churkin (approved):

alexanderPolosatov (approved):

js/renovation/ui/button.tsx
```diff
@@ -242,7 +242,11 @@ export class Button extends JSXComponent(ButtonProps) {
   get iconSource(): string {
     const { icon, type } = this.props;
 
-    return icon || type === 'back' ? icon || 'back' : '';
+    if (icon || type === 'back') {
+      return (icon ?? '') || 'back';
```
LazyLahtak, line 246: It seems that we can return just `(icon ?? 'back')`

Zedwag (approved):

js/renovation/ui/button.tsx
ksercs: Nope because icon can be an empty string

MikeVitik (approved):

PR #17870, opened 2021-06-10
DataGrid.Pdf - Indents for grouped rows with summaries

KuznetsovVN:


novsstation (dismissed):

EugeniyKiyashko (dismissed):

novsstation (dismissed):

js/exporter/jspdf/export_data_grid_2.js
```diff
@@ -48,7 +48,28 @@ function exportDataGrid(doc, dataGrid, options) {
                             pdfCell.colSpan = cellMerging.colspan;
                         }
                     } else if(rowType === 'group') {
-                        pdfCell.colSpan = columns.length - 1;
+                        if(cellData.cellSourceData.groupSummaryItems?.length > 0) {
+                            pdfCell._summaryItem = true;
```
IgnatovDan: '_summaryItem' looks strange: there is no 'summary' in 'pdfCell'
Can we rename it? 'isEmpty' ?

js/exporter/jspdf/export_data_grid_2.js
```diff
@@ -48,7 +48,28 @@ function exportDataGrid(doc, dataGrid, options) {
...
+                        }
+
+                        if(cellIndex > 0) {
+                            const prevCellsHaveSummaryItems = currentRow.slice(1).filter(pdfCell => pdfCell._summaryItem).length > 0;
+                            if(!pdfCell._summaryItem && !prevCellsHaveSummaryItems) {
```
IgnatovDan: '_summaryItem' is used in this method only but declared in a long living object.
Can we change it to a local variable?

IgnatovDan (changes requested):

js/exporter/jspdf/export_data_grid_2.js
KuznetsovVN: Ok, we can remove the '_summaryItem' field and use the 'isDefined(pdfCell.text)' expression instead. In this case we will assume that if the text is defined in pdfCell of the group row, then this is a summary item aligned by column. Except the first cell.

KuznetsovVN: Done

IgnatovDan (approved):

EugeniyKiyashko (approved):

PR #17880, opened 2021-06-11
Renovated Editors: fix 'strict-boolean-expressions' lint rule

Zedwag:


js/renovation/component_wrapper/editors/editor.ts
```diff
@@ -127,4 +127,4 @@ export default class Editor extends Component {
 
 const prevIsEditor = (OldEditor as unknown as { isEditor }).isEditor;
```
ksercs: Can not we specify the type here more strict? e.g `(OldEditor as unknown as  { isEditor: (instance: Component) => boolean })` or smth like this?

It will neutralize the warning you're trying to fix

js/renovation/ui/button.tsx
```diff
@@ -34,10 +34,10 @@ const getCssClasses = (model: ButtonProps): string => {
   const isValidStylingMode = stylingMode && stylingModes.includes(stylingMode);
   const classesMap = {
     'dx-button': true,
-    [`dx-button-mode-${isValidStylingMode ? stylingMode : 'contained'}`]: true,
+    [`dx-button-mode-${isValidStylingMode ?? false ? stylingMode : 'contained'}`]: true,
```
ksercs: let's move nullish coalescing to `isValidStylingMode` initialization

js/renovation/ui/button.tsx
```diff
@@ -235,13 +235,13 @@ export class Button extends JSXComponent(ButtonProps) {
```
ksercs, line 234: maybe it's better to get the props this way:
```
   const  text = this.props.text ?? '';
   const icon = this.props.icon ?? '';
```
to avoid changes below?

We can use this approach everywhere now and remove this additional type check after the question about possible `undefined` assignment is solved.

@churkin  what do u think?

ksercs (dismissed):

js/renovation/ui/button.tsx
Zedwag: ok

js/renovation/component_wrapper/editors/editor.ts
Zedwag: done

Zedwag:
Implemented in another PR

PR #17899, opened 2021-06-15
DataGrid - fix updatePropsImmutable when option is nested array

pomahtri:
Function `updatePropsImmutable` didn't work when option is nested array, for example calling `grid.option("toolbar.items[0].disabled", true)`. I added `getPathParts` to parse indices correctly and used it in several files


js/renovation/component_wrapper/utils/update_props_immutable.ts
```diff
@@ -1,10 +1,38 @@
...
+    const match = /(.+?)\[\s*(\d+)\s*\]/g.exec(part);
+
+    if (match) {
+      parts.push(match[1], parseInt(match[2], 10));
+    } else {
+      parts.push(part);
+    }
+  });
+
+  // if any of parts contains ']' or '[' even after parsing, then it's incorrect
```
pomahtri: I need to check whether it's correct because of "wrong array index" test

js/renovation/component_wrapper/utils/__tests__/update-props-immutable.test.ts
```diff
@@ -100,3 +100,12 @@ it('change object item in nested array option', () => {
   expect(props.nestedArray).not.toBe(nestedArray);
   expect(props.nestedArray[1]).toBe('item2 modified');
 });
+
+it('change object item in double nested array option', () => {
```
MikeVitik, line 94: Check more complex scenario { item: { objectProp: {}, items: [{ value: 'item1' }, { value: 'item2' }] } } and check that only data related to 'item1' mutated

js/renovation/component_wrapper/utils/update_props_immutable.ts
```diff
@@ -1,10 +1,38 @@
...
+  const isIncorrect = parts.some((part) => {
```
MikeVitik: Where is a test for incorrect prop name?

js/renovation/component_wrapper/utils/update_props_immutable.ts
```diff
@@ -1,10 +1,38 @@
...
+    if (isString(part)) {
```
MikeVitik: Why do you do testing index correction in the separated block?

MikeVitik (approved):

MikeVitik (changes requested):
Fix request

js/renovation/component_wrapper/utils/update_props_immutable.ts
pomahtri: There is test named `wrong array index`, this check is required only because of that test

MikeVitik (dismissed):

vconst (approved):

MikeVitik (approved):

PR #17910, opened 2021-06-15
toolbar: remove redundant handleToolbarVisibilityChange() method

EugeniyKiyashko:


IgnatovDan (dismissed):

js/ui/toolbar.js
```diff
@@ -294,7 +294,6 @@ const Toolbar = ToolbarBase.inherit({
                 break;
             case 'visible':
```
DokaRus: You can remove this `case`. It duplicates the default case now

EugeniyKiyashko: You are right. I missed it. Fixed in https://github.com/DevExpress/DevExtreme/pull/17910/commits/231815965c48049375d3af177441bc9462af5fe1

novsstation (approved):

PR #17911, opened 2021-06-16
DataGrid - add toolbar customization

pomahtri:
Demo - https://pomahtri.github.io/DataGridToolbarCustomization/

---

I added `toolbar.items` option to customize buttons in toolbar


pavel-2020 (dismissed):

js/ui/data_grid.d.ts
```diff
@@ -4108,6 +4114,10 @@ export interface SummaryTexts {
     sumOtherColumn?: string
 }
 
+export interface Toolbar {
+  items?: (string | dxToolbarItem)[];
```
vconst: Add doc comments for Toolbar and Toolbar.items. 
Also, declare possible string values and define ToolbarItem for DataGrid because dxToolbarItem does not contain name field.
See HtmlEditor as an example - https://github.com/DevExpress/DevExtreme/blob/c6a11a3c27afde0bd2b44ef65c6f1fb4304b0490/js/ui/html_editor.d.ts#L464.

js/ui/grid_core/ui.grid_core.header_panel.js
```diff
@@ -120,6 +148,10 @@ const HeaderPanel = ColumnsView.inherit({
             this._invalidate();
             args.handled = true;
         }
+        if(args.name === 'toolbar') {
+            const toolbarOptionName = args.fullName.split('.').slice(1).join('.');
+            this._toolbar.option(toolbarOptionName, args.value);
```
vconst: Add merging with default items here. Add tests about modification "toolbar", "toolbar.items", "toolbar.items[i]" options.

vconst (dismissed):

js/renovation/ui/grids/data_grid/utils/get_updated_options.ts
```diff
@@ -21,17 +21,17 @@ function compare(resultPaths: ResultItem[], item1, item2, key: string): void {
       resultPaths.push(getDiffItem(key, item2, item1));
     } else {
     // eslint-disable-next-line @typescript-eslint/no-use-before-define
-      const diffPaths = objectDiffs(item1, item2);
+      const diffPaths = objectDiffs(item1, item2, fullPropName);
       resultPaths.push(...diffPaths.map((item) => ({ ...item, path: `${key}.${item.path}` })));
     }
   } else if (type1 === 'array') {
-    if (key !== 'columns' && item1 !== item2) {
+    if (key !== 'columns' && !(key === 'items' && fullPropName.includes('toolbar')) && item1 !== item2) {
```
vconst: Why is this check necessary here? Where is jest tests for this change?

js/renovation/component_wrapper/utils/update_props_immutable.ts
```diff
@@ -24,7 +25,11 @@ function cloneObjectProp(
       fullNameParts.slice(1),
     );
   } else if (name) {
-    result[name] = value[name];
+    if (isPlainObject(value[name])) {
+      result[name] = extend(true, {}, value[name]);
```
vconst: Deep extend is bad. Why is it necessary? Where is jest tests for it? May be one level extend is enough?
```javascript
result[name] =  { ...value[name] };
```

js/ui/grid_core/ui.grid_core.header_panel.js
```diff
@@ -116,10 +152,28 @@ const HeaderPanel = ColumnsView.inherit({
...
         }
+        if(args.name === 'toolbar') {
+            if(parts.length <= 2) {
+                // toolbar and toolbar.items case
+                const toolbarOptions = this._getToolbarOptions();
+                this._toolbar.option(toolbarOptions);
+            } else if(parts.length === 3) {
+                // toolbar.items[i] case
+                const normalizedItem = this._normalizeToolbarItems(this._getToolbarItems(), args.value);
+                const optionName = args.fullName.replace(/^toolbar./, '');
```
vconst: The optionName calculation is  duplicated in two cases (parts.length === 3 and parts.length >= 4). May be combine this two cases in one condition? 
```javascript
if(parts.length >= 3) {
    const value = parts.length === 3 ? this._normalizeToolbarItems(this._getToolbarItems(), args.value) : args.value;
    const optionName = args.fullName.replace(/^toolbar./, '');
    this._toolbar.option(optionName, value);
}
```

js/ui/tree_list.d.ts
```diff
@@ -1184,6 +1192,36 @@ export default class dxTreeList extends Widget<dxTreeListOptions> implements Gri
...
+ * @type object
+ * @namespace DevExpress.ui
+ */
+export interface dxTreeListToolbarItem extends dxToolbarItem {
+  /**
+   * @docid
+   * @type Enums.TreeListToolbarItemName|string
+   * @public
+  */
+  name?: dxTreeListDefaultToolbarItemName | string
```
vconst, line 1208: Why is string type necessary here? DataGrid and TreeList have limited build-in items.

pomahtri: It might be useful in case user wants to set custom name to identificate it later, for example, in `onToolbarPreparing` event. I wrote it following example of htmlEditor's toolbar: https://github.com/DevExpress/DevExtreme/blob/21_2/js/ui/html_editor.d.ts#L495

js/renovation/component_wrapper/utils/update_props_immutable.ts
pomahtri: In my case one level extend is enough, but it might be not enough in some other cases.

The problem was in this case (it's test called `Change toolbar.items[i] in runtime` in `dataGrid.tests.js`):

- set debugger to `js\renovation\ui\grids\data_grid\utils\get_updated_options.ts:28` (after `else if (type1 === 'array')`)
- I have toolbar with toolbar with two buttons, both are `before`
```
toolbar: { items: [  {
    name: 'columnChooserButton',
    location: 'before'
}, {
    name: 'addRowButton',
    location: 'before'
} ] }
```
- I'm changing second button: `grid.option('toolbar.items[1]', { name: 'addRowButton', location: 'after' });`
- In debugger `item1 !== item2` is true, but their content is equal: both have `location: 'after'` for second button, so `optionChanged` won't be triggered. It happens because `item1[1] === item2[1]`

I'm not sure if one level extend will work in other cases, but i can't give an example yet

js/renovation/ui/grids/data_grid/utils/get_updated_options.ts
pomahtri: This check is necessary. Before that function `compare` went deep comparing changes only if it's some of column options. In toolbar I also need to deeply check what option was changed. 

Without this check when I call `grid.option('toolbar.items[0].disabled', true)`, in `optionChanged` argument `fullName` would be `toolbar.items`, not `toolbar.items[0].disabled`

js/ui/grid_core/ui.grid_core.header_panel.js
```diff
@@ -116,10 +152,28 @@ const HeaderPanel = ColumnsView.inherit({
     },
 
     optionChanged: function(args) {
+        const parts = getPathParts(args.fullName);
+
```
NickMitrokhin: Since "parts" is used only  under the following condition:

```js
if(args.name === 'toolbar')
```
move this variable declaration there.

testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -2204,6 +2203,250 @@ QUnit.module('Assign options', baseModuleConfig, () => {
         // assert
         assert.strictEqual(onOptionChanged.callCount, 1, 'onOptionChanged is called once');
     });
+
+    QUnit.test('Change toolbar.items[i].prop in runtime', function(assert) {
```
NickMitrokhin: "in runtime" -> "at runtime"

testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -2204,6 +2203,250 @@ QUnit.module('Assign options', baseModuleConfig, () => {
...
+
+        // assert
+        assert.equal($buttonBefore.length, 1, 'count button');
+        assert.equal($buttonAfter.length, 1, 'count button');
+
+        assert.ok($buttonBefore.hasClass('dx-datagrid-column-chooser-button'), 'has column chooser button');
+        assert.ok($buttonAfter.hasClass('dx-datagrid-addrow-button'), 'has add button');
+    });
+
+    QUnit.test('Change toolbar.items[i] in runtime', function(assert) {
```
NickMitrokhin: "in runtime" -> "at runtime"

testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -2204,6 +2203,250 @@ QUnit.module('Assign options', baseModuleConfig, () => {
...
+
+        // assert
+        assert.equal($buttonBefore.length, 1, 'count button');
+        assert.equal($buttonAfter.length, 1, 'count button');
+
+        assert.ok($buttonBefore.hasClass('dx-datagrid-column-chooser-button'), 'has column chooser button');
+        assert.ok($buttonAfter.hasClass('dx-datagrid-addrow-button'), 'has add button');
+    });
+
+    QUnit.test('Change toolbar.items in runtime', function(assert) {
```
NickMitrokhin: "in runtime" -> "at runtime"

testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -2204,6 +2203,250 @@ QUnit.module('Assign options', baseModuleConfig, () => {
...
+
+        // assert
+        assert.equal($buttonBefore.length, 1, 'count button');
+        assert.equal($buttonAfter.length, 1, 'count button');
+
+        assert.ok($buttonBefore.hasClass('dx-datagrid-column-chooser-button'), 'has column chooser button');
+        assert.ok($buttonAfter.hasClass('dx-datagrid-addrow-button'), 'has add button');
+    });
+
+    QUnit.test('Change toolbar in runtime', function(assert) {
```
NickMitrokhin: "in runtime" -> "at runtime"

MikeVitik (approved):

NickMitrokhin (approved):

PR #17988, opened 2021-06-18
Configure `webpack` to resolve `cjs` modules

jacksonrr3:
Fix for [T1002756](https://supportcenter.devexpress.com/internal/ticket/details/T1002756)


alexserov:
Is it possible to add tests for this feature so that we could protect ourselves from the same issues in future?

nightskylark (approved):

PR #18015, opened 2021-06-21
pivotgrid: change default value for export.loadPanel.height property

EugeniyKiyashko:


novsstation (approved):

EugeniyKiyashko:
We desided that we don't need to set export.loadpanel.height value according to pivot.loadpanel.height property value equal 70px. Actual value is 90px and has already described in documentation. This value looks suit for us too.

PR #18016, opened 2021-06-21
DropDownEditors: fix issue when scrolling does not work if dropDownOptions container is root element

Zedwag:


Zedwag:
replaced with this https://github.com/DevExpress/DevExtreme/pull/18039

PR #18064, opened 2021-06-23
Validation Engine: validation group config should get actual information about integrated validators state (T1006292, T1006667)

ksercs:


js/ui/validation_engine.js
```diff
@@ -451,7 +451,7 @@ const GroupConfig = Class.inherit({
             this._validationInfo.result.isValid = this._validationInfo.result.status === STATUS.valid;
             const res = extend({}, this._validationInfo.result, { complete: null });
             const deferred = this._validationInfo.deferred;
-            this._resetValidationInfo();
+            this._validationInfo.deferred = null;
```
NickMitrokhin, line 454: Why did you replace the this._resetValidationInfo(); method?

ksercs: `_validationInfo` contains info about group validation (several validators, many rules). We should not remove all this info on only single validator removing (this code is called from `removeRefisteredValidator`).

Particularly, it breaks the case when we dispose several integrated editors one by one

NickMitrokhin (approved):

js/ui/validation_engine.js
```diff
@@ -592,6 +592,8 @@ const ValidationEngine = {
             status: STATUS.valid,
             complete: null
         };
+        const validator = rules?.[0]?.validator;
```
ksercs, line 595: NOTE: this optional chaining needs for tests passing. Rule and validator inside of it always exist in real scenario

DokaRus (approved):

Zedwag (approved):

alexander-kotov-dx (approved):

PR #18069, opened 2021-06-24
Get rid of default export in the data/utils module

jacksonrr3:


js/bundles/modules/data.js
```diff
@@ -46,16 +46,14 @@ DevExpress.EndpointSelector = require('../../data/endpoint_selector');
 DevExpress.data.queryImpl = require('../../data/query_implementation').queryImpl;
 DevExpress.data.queryAdapters = require('../../data/query_adapters');
 
-const dataUtils = require('../../data/utils');
```
alexserov: Are these changes necessary? Seems that we can keep an old version of this file

js/data/odata/query_adapter.js
```diff
@@ -3,9 +3,20 @@ import { each } from '../../core/utils/iterator';
 import config from '../../core/config';
 import { extend } from '../../core/utils/extend';
 import queryAdapters from '../query_adapters';
-import { sendRequest, generateSelect, generateExpand, serializeValue, convertPrimitiveValue, serializePropName } from './utils';
```
alexserov, line 6: @nightskylark We prefer single-line imports or multi-line ones?

js/bundles/modules/data.js
jacksonrr3: Not necessary, but explicit named import looks better.

alexserov (approved):

PR #18103, opened 2021-06-27
SelectBox: fix bug with excess datasource load when search non-existant item (T1001808)

Zedwag:


Zedwag:
decided not to rework it now, but keep it in mind in future interaction with dataSource rework

PR #18117, opened 2021-06-28
LDML Date Parser: getRegExpInfo should return correct regular expression for unambiguous not separated `formats` and throw warning message in another case. (T1008667)

jacksonrr3:
RegExp `y` pattern quantifier was greedy, so it captures more than necessary digits in some cases when `displayFormat` does not have separators.
Also fixed two-digits patterns and add checking the format for unambiguous.


js/localization/ldml/date.parser.js
```diff
@@ -40,7 +41,7 @@ const PATTERN_REGEXPS = {
         return dateParts.getPeriodNames(FORMAT_TYPES[count < 3 ? 3 : count], 'format').join('|');
     },
     d: function(count) {
-        return '0?[1-9]|[12][0-9]|3[01]';
+        return count === 2 ? '3[01]|[12][0-9]|0[1-9]' : '0?[1-9]|[12][0-9]|3[01]';
```
nightskylark: There is only one sign difference, but it is not obious because of different parts oder. I recommend to apply the condition for only one that sign.

nightskylark: Why did you make more strict regexp only for two of date parts: month and day. What about the rest of them?

testing/tests/DevExpress.localization/ldml.tests.js
```diff
@@ -412,4 +413,142 @@ QUnit.module('number formatter', () => {
...
+        ].forEach(({ format, tests }) => {
+            const regExpInfo = getRegExpInfo(format);
+            tests.forEach(({ date, result }) => {
+                const regExpGroupsResult = regExpInfo.regexp.exec(date).slice(1);
+                assert.deepEqual(regExpGroupsResult, result, `${format} - format ok`);
+            });
+        });
+    });
+
+    QUnit.test('getRegExpInfo should write warning message if there are not separated single simbols in the `format`!', function(assert) {
```
nightskylark:
```suggestion
    QUnit.test('getRegExpInfo should throw warning message if there are no separated single simbols in the `format`!', function(assert) {
```

js/localization/ldml/date.parser.js
```diff
@@ -218,6 +219,22 @@ export const getRegExpInfo = function(format, dateParts) {
 
     addPreviousStub();
 
+    patterns.every((pattern, index) => {
+        const char = pattern[0];
+        const prevPatternChar = index ? patterns[index - 1][0] : null;
```
nightskylark:
```suggestion
        const prevPatternChar = index && patterns[index - 1][0];
```

js/localization/ldml/date.parser.js
```diff
@@ -218,6 +219,22 @@ export const getRegExpInfo = function(format, dateParts) {
...
+        const nextPatternChar = index < patterns.length - 1 ? patterns[index + 1][0] : null;
```
nightskylark:
```suggestion
        const nextPatternChar = index < patterns.length - 1 && patterns[index + 1][0];
```

js/localization/ldml/date.parser.js
```diff
@@ -218,6 +219,22 @@ export const getRegExpInfo = function(format, dateParts) {
...
+        const isSingle = pattern.length === 1;
+        const isFirst = index === 0;
+        const isRegexpPart = (char) => PATTERN_REGEXPS[char] && char !== ':';
+
+        if(isSingle && isRegexpPart(char) && (isRegexpPart(prevPatternChar) || (isFirst && isRegexpPart(nextPatternChar)))) {
+            logger.warn(`Date-time formats without separators may produce unexpected results. Please add separators to the following format: ${format}.`);
```
nightskylark: I suppose the warning text that Alexey Serov suggested describes the possible solutions more completely
>The ${invalidPattern} may lead to undefined parsing behavior. Consider adding separators, or replace minimum digits symbols with the fixed digits notation

js/localization/ldml/date.parser.js
```diff
@@ -218,6 +219,22 @@ export const getRegExpInfo = function(format, dateParts) {
...
 
+    patterns.every((pattern, index) => {
+        const char = pattern[0];
+        const prevPatternChar = index ? patterns[index - 1][0] : null;
+        const nextPatternChar = index < patterns.length - 1 ? patterns[index + 1][0] : null;
+        const isSingle = pattern.length === 1;
+        const isFirst = index === 0;
+        const isRegexpPart = (char) => PATTERN_REGEXPS[char] && char !== ':';
+
+        if(isSingle && isRegexpPart(char) && (isRegexpPart(prevPatternChar) || (isFirst && isRegexpPart(nextPatternChar)))) {
```
nightskylark: I prefer to extract such long conditions to the separate variable. It's name should make the code to be easier to read.

js/localization/ldml/date.parser.js
```diff
@@ -218,6 +219,22 @@ export const getRegExpInfo = function(format, dateParts) {
 
     addPreviousStub();
 
+    patterns.every((pattern, index) => {
```
nightskylark: Actually, it's better to extract this code to the separate function. The `getRegExpInfo` method becomes too big and it's hard to read the code.

nightskylark (changes requested):

nightskylark (approved):

PR #18166, opened 2021-06-30
DateBox should not raise any errors when useMaskBehavior is enabled, locale digits are different to arabic digits and Fractional Seconds in the "displayFormat"

jacksonrr3:
`_prepareRegExpInfo()` converts quantifies digits at the regExp, which provokes wrong regExp behavior.


js/ui/date_box/ui.date_box.mask.js
```diff
@@ -325,10 +325,18 @@ const DateBoxMask = DateBoxBase.inherit({
 
     _prepareRegExpInfo() {
         this._regExpInfo = getRegExpInfo(this._getFormatPattern(), dateLocalization);
-        const regExp = this._regExpInfo.regexp;
-        const flags = regExp.flags;
-        const convertedRegExp = numberLocalization.convertDigits(this._regExpInfo.regexp.source, false);
-        this._regExpInfo.regexp = RegExp(convertedRegExp, flags);
+        const source = this._regExpInfo.regexp.source;
+        const flags = this._regExpInfo.regexp.flags;
+        const quantifiers = new RegExp(/(\{[0-9],?[0-9]?\})/);
```
ksercs: Let's move this regex to `QUANTIFIER_REGEX` constant?

ksercs: Shoudn't we also take into account the scenario when 2+ digits number is used in quantifier?

For example, current regex will not work for string with `{1, 10}` quantifier

ksercs: Shouldn't we use `g` flag here?

js/ui/date_box/ui.date_box.mask.js
```diff
@@ -325,10 +325,18 @@ const DateBoxMask = DateBoxBase.inherit({
 
     _prepareRegExpInfo() {
         this._regExpInfo = getRegExpInfo(this._getFormatPattern(), dateLocalization);
-        const regExp = this._regExpInfo.regexp;
-        const flags = regExp.flags;
-        const convertedRegExp = numberLocalization.convertDigits(this._regExpInfo.regexp.source, false);
-        this._regExpInfo.regexp = RegExp(convertedRegExp, flags);
+        const source = this._regExpInfo.regexp.source;
+        const flags = this._regExpInfo.regexp.flags;
```
ksercs: let's move `this._regExpInfo.regexp` to a temporary variable to avoid copypaste?

js/ui/date_box/ui.date_box.mask.js
ksercs: Let's rename this variable to `quantifierRegExp` to improve readability of the following code?

testing/tests/DevExpress.localization/localization.globalize.widgets.tests.js
```diff
@@ -150,6 +150,26 @@ QUnit.module('DateBox', commonEnvironment, () => {
         }
     });
 
+    QUnit.test('DateBox should not raise error when digits are not default arabic digits and Fractional Seconds in the "displayFormat"', function(assert) {
+        const originalCulture = Globalize.locale().locale;
+
+        try {
+            Globalize.locale('ar');
+
+            const $dateBoxInstance = $('#dateBox').dxDateBox({
```
ksercs: we should use `$` prefix only for jquery elements

ksercs (approved):

nightskylark:
I still don't like the modifying of the regular expressions. It meens that datebox don't only uses them as the pure functions. It knows about what structure of the regular expressions the ldml module provides. It's more than the ldml module API provides.  As a result we can get a bug in a future if someone will chage regular expressions structure that does not considered by this code.

DokaRus (approved):

PR #18175, opened 2021-07-01
Pointer events: "dxpointercancel" should be triggered on "pointercancel" native event raise (T1005143)

ksercs:


nightskylark:
Did you find out why it was not added initially?

ksercs:
> Did you find out why it was not added initially?

it was made this way in the commit `26771 (46cd112569d7) B239957` with event strategies separation

ksercs:
We should think about more global solution

PR #18226, opened 2021-07-06
DataGrid - Field is not updating correctly with cell.setValue method (T1009772)

pomahtri:
Ticket:  https://devexpress.com/issue=T1009772

Cherry picks:

- https://github.com/DevExpress/DevExtreme/pull/18228
- https://github.com/DevExpress/DevExtreme/pull/18227

---

In method `_addChange` new changes were added using [`createObjectWithChanges`], which led to object values to be merged instead of overwritten. 

Using of [`createObjectWithChanges`] seems to be important because of using safe extending, so I copied its code as `createObjectWithChangesOnOneLevel` and added deleting of previous values. Unfortunately, I can't just delete keys before calling this func: object needs to be copied before that, and it's important to save prototype

[`createObjectWithChanges`]: https://github.com/DevExpress/DevExtreme/blob/21_2/js/data/array_utils.js#L100-L104

### DIFF
--- a/js/data/array_utils.js
+++ b/js/data/array_utils.js
@@ -246,6 +246,7 @@ function indexByKey(keyInfo, array, key) {
 export {
     applyBatch,
     createObjectWithChanges,
+    cloneInstance,
     update,
     insert,
     remove,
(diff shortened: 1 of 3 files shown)


NickMitrokhin (approved):

pomahtri:
We decided to leave the previous behavior. `cell.setValue` method is connected to `store.update`, where it's supposed to be merge-like changing

PR #18227, opened 2021-07-06
DataGrid - Field is not updating correctly with cell.setValue method (T1009772)

pomahtri:
Cherry pick from https://github.com/DevExpress/DevExtreme/pull/18226


NickMitrokhin (approved):

pomahtri:
We decided to leave the previous behavior. `cell.setValue` method is connected to `store.update`, where it's supposed to be merge-like changing

PR #18265, opened 2021-07-08
EventType should extend JQueryEventObject

Bykiev:
EventType should extend JQueryEventObject


Bykiev:
Can't make it building, help is appreciated!

PR #18302, opened 2021-07-12
Build: add new themebuilder-npm-devextreme script to build devextreme npm only

vconst:
Often, it is necessary to build devextreme npm package only. But build task contains themebuilder-npm task that usually is not necessary and takes 2 minutes.

In this PR I add build-npm-devextreme npm task that does not build themebuilder npm package.


babich-a (dismissed):

package.json
```diff
@@ -229,4 +230,4 @@
     ]
   },
   "pre-commit": "lint-staged"
-}
+}
```
nightskylark: No New Line

vconst: fixed

nightskylark (approved):

PR #18309, opened 2021-07-13
ESLint: enable no-non-null-assertion rule

churkin:


js/renovation/ui/scroll_view/scrollable_simulated.tsx
```diff
@@ -1189,10 +1193,10 @@ export class ScrollableSimulated extends JSXComponent<ScrollableSimulatedPropsTy
     }
 
     // T320141
-    const isOverflowHidden = getElementComputedStyle(this.contentRef.current)!.overflowY === 'hidden';
+    const computedStyle = getElementComputedStyle(this.contentRef.current);
```
MikeVitik: // ignore no-null-assertion
    const computedStyle = getElementComputedStyle(this.contentRef.current)!;
    if (computedStyle.overflowY === 'hidden') {
      return this.contentClientHeight;
    }

js/renovation/ui/scroll_view/scrollable_simulated.tsx
```diff
@@ -344,18 +344,22 @@ export class ScrollableSimulated extends JSXComponent<ScrollableSimulatedPropsTy
     }
 
     this.updateHandler();
```
MikeVitik: I suppose remove disectionfrom restoreLocation and inline direction part here.

MikeVitik (changes requested):

js/renovation/ui/scroll_view/scrollable_simulated.tsx
```diff
@@ -344,18 +344,22 @@ export class ScrollableSimulated extends JSXComponent<ScrollableSimulatedPropsTy
...
+    const isTopDefined = location.top !== undefined;
+    const isLeftDefined = location.left !== undefined;
```
IgnatovDan: These "!== undefined" are not necessary because there is the code few line above:
```
    if (!location.top && !location.left) {
      return;
    }
```
How do you suggest to manage this case?

IgnatovDan (changes requested):

js/renovation/ui/scroll_view/scrollable_simulated.tsx
churkin: One of these can be undefined after the "if"

js/renovation/ui/scroll_view/scrollable_simulated.tsx
```diff
@@ -344,18 +344,22 @@ export class ScrollableSimulated extends JSXComponent<ScrollableSimulatedPropsTy
...
 
     // TODO: try to simplify it
-    if (this.direction.isVertical) {
+    if (this.direction.isVertical && isTopDefined) {
       const scrollbar = this.vScrollbarRef.current!;
+
       location.top = scrollbar.getLocationWithinRange(
-        location.top! + this.vScrollLocation,
+        location.top as number + this.vScrollLocation,
```
IgnatovDan: The 'location' variable value is returned by the 'restoreLocation' function.
The result of this function is declared as 
```
export interface ScrollOffset {
  top: number;
  left: number;
}
```
Why do we need to write another 'as number' cast at these lines?

IgnatovDan (changes requested):

js/renovation/ui/scroll_view/scrollable_simulated.tsx
```diff
@@ -1171,10 +1175,10 @@ export class ScrollableSimulated extends JSXComponent<ScrollableSimulatedPropsTy
     }
 
     // T320141
-    const isOverflowHidden = getElementComputedStyle(this.contentRef.current)!.overflowX === 'hidden';
+    const computedStyle = getElementComputedStyle(this.contentRef.current);
```
IgnatovDan: The 'getElementComputedStyle' result is not null/undefined if its argument is not null/undefined.
The 'this.contentRef.current' variable is passed as its argument and there is this code few lines above:
```
    if (!isDefined(this.contentRef?.current)) {
      return 0;
    }
```
This member will not null/undefined and the getElementComputedStyle function result will not be null/undefined.
Why do you want to change this line of code?

IgnatovDan (changes requested):

js/renovation/ui/scroll_view/scrollable_simulated.tsx
IgnatovDan: Internally, the 'getElementComputedStyle' uses the [Window/getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) that doesn't return null/undefined.
Why this line requires 'ignore no-null-assertion' comment?

IgnatovDan (changes requested):

js/renovation/ui/scroll_view/scrollable_simulated.tsx
churkin: restoreLocation returns only part of ScrollOffset:
```typescript
xport function restoreLocation(
  location: number | Partial<{ x: number; y: number; top: number; left: number }>,
  direction?: ScrollableDirection,
): Partial<ScrollOffset> {
```

js/renovation/ui/scroll_view/utils/get_element_padding.ts
```diff
@@ -1,11 +1,11 @@
-import getElementComputedStyle from '../../../utils/get_computed_style';
+import { getElementComputedStyle } from '../../../utils/get_computed_style';
 import { toNumber } from '../../../utils/type_conversion';
 
 export function getElementStyle(
   name: keyof CSSStyleDeclaration, element: Element | null | undefined,
 ): number {
-  const computedStyle = getElementComputedStyle(element) ?? {};
-  return toNumber(computedStyle[name]);
+  const computedStyle = getElementComputedStyle(element as Element) ?? {};
```
IgnatovDan: No.
It should be 
```
export function getElementPaddingBottom(element: Element): number {
  return toNumber(getElementComputedStyle(element).paddingBottom);
}
```
The 'getElementStyle' function is imported nowhere and we can delete it.

IgnatovDan (changes requested):

js/renovation/ui/scroll_view/scrollable_simulated.tsx
```diff
@@ -1189,10 +1193,10 @@ export class ScrollableSimulated extends JSXComponent<ScrollableSimulatedPropsTy
     }
 
     // T320141
-    const isOverflowHidden = getElementComputedStyle(this.contentRef.current)!.overflowY === 'hidden';
+    const computedStyle = getElementComputedStyle(this.contentRef.current);
 
     /* istanbul ignore next */
-    if (isOverflowHidden) {
+    if (computedStyle.overflowY === 'hidden') {
```
IgnatovDan: Can we change this declaration:
```
export default function getElementComputedStyle(el: Element | undefined | null):
CSSStyleDeclaration | null {
```
to this:
```
export default function getElementComputedStyle(el: Element):
CSSStyleDeclaration {
```
?
Or, even use 'window.getComputedStyle(el)' directly without wrappers?
It will resolve such monkey changes with "ignore no-null-assertion" comments.

IgnatovDan (changes requested):

js/renovation/ui/scroll_view/scrollable_simulated.tsx
IgnatovDan: I see.
Then it requires another change because "undefined + 1 = NaN"
It will be ready soon.

IgnatovDan: "Partial<ScrollOffset>" is used by a mistake.
Actual result is a "{ left, top }" object.
Both the 'left' ad 'top' can be undefined:
```
    return {
      left: isDefined(left) ? -left : undefined,
      top: isDefined(top) ? -top : undefined,
    };
```

We need this change:
1. at line '342':
if (!location.top && !location.left) -> **if (!isDefined(location.top) || !isDefined(location.left))** 

2. in restore_location.ts:
**export interface Location {
  top: number | undefined;
  left: number | undefined;
}**
export function restoreLocation(
  location: number | Partial<{ x: number; y: number; top: number; left: number }>,
  direction?: ScrollableDirection,
): **Location** 

Some tests can fail, lets extract this change to a new PR.

js/renovation/ui/scroll_view/utils/get_element_padding.ts
IgnatovDan: Let's extract this change to a new PR with change to window.getComputedStyle direct call.

js/renovation/ui/scroll_view/scrollable_simulated.tsx
IgnatovDan: Let's extract this change to a new PR.

js/renovation/ui/scroll_view/scrollable_simulated.tsx
IgnatovDan: Let's extract this change to a new PR with change to window.getComputedStyle direct call.

js/renovation/ui/scroll_view/scrollable_simulated.tsx
IgnatovDan: Let's extract this change to a new PR with change to window.getComputedStyle direct call.

IgnatovDan (dismissed):

IgnatovDan (approved):

MikeVitik (approved):

PR #18323, opened 2021-07-13
Deprecate function value for width and height

MikeVitik:


churkin (dismissed):

js/core/dom_component.js
```diff
@@ -58,11 +58,23 @@ const DOMComponent = Component.inherit({
...
     },
 
+    _checkFunctionValueDeprecation: function(optionNames) {
+        if(!this.option('_ignoreFunctionValueDeprecation')) {
+            optionNames.forEach(optionName => {
+                if(isFunction(this.option(optionName))) {
+                    errors.log('W0017', optionName);
+                }
+            });
+
```
ksercs:
```suggestion

```

js/ui/lookup.js
```diff
@@ -689,7 +690,7 @@ const Lookup = DropDownList.inherit({
             closeOnTargetScroll: false,
             onPositioned: null,
 
-            maxHeight: function() { return $(window).height(); },
+            maxHeight: '100hv',
```
ksercs: ?
```suggestion
            maxHeight: '100vh',
```

ksercs (commented):
What about using this options in inheritors? 

Try to search by `this._getOptionValue('width',` or something like this

andrewmakarov (dismissed):

ksercs (dismissed):

ksercs:
Shouldn't we also change overlay default width and height?

ksercs (approved):

PR #18362, opened 2021-07-16
DataGrid: Fix saving text editor value on enter key if mask is defined (T1013643)

vconst:


testing/testcafe/tests/dataGrid/editing_matrix.ts
```diff
@@ -411,7 +418,7 @@ editingModes.forEach((mode) => {
               repaintChangesOnly,
             };
 
-            test(`Update cell value ${JSON.stringify({
+            test.only(`Update cell value ${JSON.stringify({
```
pomahtri: forgot debug

NickMitrokhin (approved):

pomahtri (approved):

PR #18368, opened 2021-07-19
Attach CSS device classes if viewport already exist (T1010826)

jacksonrr3:


js/core/devices.js
```diff
@@ -348,6 +348,11 @@ class Devices {
 
 const devices = new Devices();
 
+const existViewPort = viewPort();
+if(existViewPort) {
+    devices.attachCssClasses(existViewPort);
+}
```
nightskylark:
```suggestion
const viewPortElement = viewPort();
if(viewPortElement) {
    devices.attachCssClasses(viewPortElement);
}
```

nightskylark (approved):

PR #18385, opened 2021-07-20
Overlays: rework private options

alexanderPolosatov:
Meaningful changes:
1. Popover and Tooltip default position value is not string anymore, it's an object
2. `_position` cache is calculated in the single function. It's used everywhere instead of direct option value getting
3. Elements which scroll triggers overlay hiding are `wrapper` element parents now. So position does not affect `closeOnTargetScroll` option behaviour anymore


js/ui/overlay/ui.overlay.js
```diff
@@ -502,12 +472,44 @@ const Overlay = Widget.inherit({
 
     _normalizePosition: function() {
         const position = this.option('position');
-        this._position = typeof position === 'function' ? position() : position;
+        const defaultPositionOptions = {
+            of: this.option('target'),
+            boundaryOffset: DEFAULT_BOUNDARY_OFFSET
+        };
+        if(isFunction(position)) {
+            this._position = extend(true, {}, defaultPositionOptions, position());
```
ksercs: function passed to `position` property can return alias string. Seems like this case is broken.

Let's use `getOptionValue`?

js/ui/overlay/ui.overlay.js
```diff
@@ -502,12 +472,44 @@ const Overlay = Widget.inherit({
 
     _normalizePosition: function() {
         const position = this.option('position');
```
ksercs: const { position, target } = this.option();

js/ui/overlay/ui.overlay.js
```diff
@@ -502,12 +472,44 @@ const Overlay = Widget.inherit({
...
+            }
+        }
     },
 
     _getAnimationConfig: function() {
         let animation = this.option('animation');
         if(isFunction(animation)) animation = animation.call(this);
+        const target = this.option('target');
+
+        if(isDefined(target)) {
```
ksercs: oh, I don't think it's a good idea to move `_initTarget` function code to this function.

`_initTarget` was called only in `initOptions` and `target` option changed handler.
But this function is called every time overlay shows or hides

js/ui/overlay/ui.overlay.js
```diff
@@ -564,7 +566,6 @@ const Overlay = Widget.inherit({
 
         this._normalizePosition();
```
ksercs, line 586: do we really need this call now?

js/ui/overlay/ui.overlay.js
```diff
@@ -849,27 +850,18 @@ const Overlay = Widget.inherit({
...
 
-        if(devices.real().deviceType === 'desktop') {
-            $parents = $parents.add(window);
-        }
-
-        this._proxiedTargetParentsScrollHandler = this._proxiedTargetParentsScrollHandler
-            || (e => { this._targetParentsScrollHandler(e); });
-
-        eventsEngine.off($().add(this._$prevTargetParents), scrollEvent, this._proxiedTargetParentsScrollHandler);
+        eventsEngine.off($().add(this._$prevTargetParents), scrollEvent, this._scrollHandler);
```
ksercs: why we cannot get rid of `$().add`. It'll be a problem during renovation

js/ui/overlay/ui.overlay.js
```diff
@@ -290,11 +289,11 @@ const Overlay = Widget.inherit({
 
         this._toggleViewPortSubscription(true);
         this._initHideTopOverlayHandler(this.option('hideTopOverlayHandler'));
+        this._scrollHandler = (e => { this._targetParentsScrollHandler(e); });
```
ksercs: Can we name this variable more specific?

ksercs: If we use it only inside of `_toggleParentsScrollSubscription` shouldn't we cache both handler and "parents" there as single object?

Something like
```
this._parentScrollSubsciptionInfo= {
    prevTargets: ...,
    targets: ...,
    handler: ...
}

```

js/ui/overlay/ui.overlay.js
```diff
@@ -1292,7 +1284,7 @@ const Overlay = Widget.inherit({
         const container = this.option('container');
         let positionOf = null;
 
-        if(!container && position) {
+        if(!container) {
```
ksercs: Why do we change it?

js/ui/overlay/ui.overlay.js
```diff
@@ -1416,7 +1405,7 @@ const Overlay = Widget.inherit({
         this._updateZIndexStackPosition(false);
         this._toggleTabTerminator(false);
 
-        this._actions = null;
+        this._actions = this._scrollHandler = this._$prevTargetParents = null;
```
ksercs: it's better to use `undefined` to not show this cache variables in instance object

js/ui/overlay/ui.overlay.js
```diff
@@ -1321,28 +1313,25 @@ const Overlay = Widget.inherit({
...
-            const position = this._transformStringPosition(this._position, POSITION_ALIASES);
-            const resultPosition = positionUtils.setup(this._$content, position);
-
-            return resultPosition;
+            return positionUtils.setup(this._$content, position);
         }
     },
 
-    _transformStringPosition: function(position, positionAliases) {
+    _getPositionOptions: function(positionAliases) {
```
ksercs: mb `_getPositionOptionAsObject` or smth like this?

testing/tests/DevExpress.ui.widgets.editors/validationMessage.tests.js
```diff
@@ -33,14 +33,19 @@ QUnit.module('options', moduleSetup, () => {
...
+        const onShowingSpy = sinon.spy();
 
         try {
             $element.appendTo('#qunit-fixture');
+            this._validationMessage.on('showing', onShowingSpy);
 
             this._validationMessage.option('target', $element);
 
-            assert.strictEqual(this._validationMessage.option('position').of, $element, 'position is recalculated');
+            assert.strictEqual(onShowingSpy.callCount, 1, 'position is recalculated');
```
ksercs: Why did u change this test? I think its initial meaning is lost

testing/tests/DevExpress.ui.widgets/dropDownMenu.tests.js
```diff
@@ -482,12 +482,17 @@ QUnit.module('position', {
         $(element.dxDropDownMenu('instance')._button.$element()).trigger('dxclick');
 
         positionConfig = instance._popup.option('position');
-        assert.equal(positionConfig, instance.option('popupPosition'));
```
ksercs, line 485: cannot we use `deepEqual` here?

ksercs (commented):
Do we need `normalizePosition` call in `renderGeometryImpl` now? Seems like it should be called only on init and appropriate options change

testing/tests/DevExpress.ui.widgets.editors/validationMessage.tests.js
alexanderPolosatov: because at present, options does not mobified by 'target'. target sets to _position field.
onShowingSpy.callCount is equal to one, because I subscribed on 'showing' event after initial show.

js/ui/overlay/ui.overlay.js
alexanderPolosatov: because _position already normalized on _init and has 'of' field

js/ui/overlay/ui.overlay.js
alexanderPolosatov: fixed

js/ui/overlay/ui.overlay.js
alexanderPolosatov: fixed

js/ui/overlay/ui.overlay.js
alexanderPolosatov: reworkerd

js/ui/overlay/ui.overlay.js
```diff
@@ -878,28 +883,20 @@ const Overlay = Widget.inherit({
...
+        eventsEngine.off(parentsScrollSubscriptionInfo.prevTargets, scrollEvent, parentsScrollSubscriptionInfo.handler);
 
-        eventsEngine.off($().add(this._$prevTargetParents), scrollEvent, this._proxiedTargetParentsScrollHandler);
-
-        if(subscribe && closeOnScroll) {
-            eventsEngine.on($parents, scrollEvent, this._proxiedTargetParentsScrollHandler);
-            this._$prevTargetParents = $parents;
+        const closeOnScroll = this.option('closeOnTargetScroll');
+        if(needSubscribe && closeOnScroll) {
+            let $parents = getElement(this._position.of).parents();
```
ksercs: why do we get parents from `position.of`,  not `container`? `position.of` affect visual position but not a markup

testing/tests/DevExpress.ui.widgets.editors/validationMessage.tests.js
ksercs: Why does `showing` event raise guarantee us position is recalculated? Seems like this test should be written another way. For example, check visual position using `getBoundingClientRect`

js/ui/overlay/ui.overlay.js
alexanderPolosatov: reworked

js/ui/overlay/ui.overlay.js
alexanderPolosatov: reworked. at present, we get parents of wrapper element

testing/tests/DevExpress.ui.widgets.editors/validationMessage.tests.js
alexanderPolosatov: rewirked

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -2032,26 +2050,10 @@ testModule('close on target scroll', moduleConfig, () => {
         assert.strictEqual(overlay.option('visible'), true, 'overlay should not be hidden if show animation is not completed');
     });
 
-    test('overlay should be hidden if any of jQuery Event target\'s parents were scrolled', function(assert) {
```
ksercs, line 2035: Why do we delete this test?

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -2154,44 +2159,21 @@ testModule('close on target scroll', moduleConfig, () => {
...
-        const targetParent = $target.parent().get(0);
+        const containerParent = $container.parent().get(0);
         $target.detach();
         overlay.hide();
-        const parentEvents = $._data(targetParent).events || {};
+        const parentEvents = $._data(containerParent).events || {};
         assert.strictEqual('scroll' in parentEvents, false, 'scroll unsubscribed');
     });
 
-    test('all opened overlays should be closed on scroll', function(assert) {
```
ksercs, line 2164: Why do we delete this test?

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -2562,22 +2544,68 @@ testModule('target', moduleConfig, () => {
...
+            left: 100
+        });
+
+        const overlay = new Overlay($('#overlay'), {
+            target: $target
         });
+
+        overlay.show();
+
+        const $wrapper = overlay.$content().closest(toSelector(OVERLAY_WRAPPER_CLASS));
```
ksercs:
```suggestion
        const $wrapper = overlay.$wrapper();
```

js/ui/overlay/ui.overlay.js
```diff
@@ -1441,7 +1437,7 @@ const Overlay = Widget.inherit({
         this._updateZIndexStackPosition(false);
         this._toggleTabTerminator(false);
 
-        this._actions = null;
+        this._actions = this._scrollHandler = this._$prevTargetParents = null;
```
ksercs:
```suggestion
        this._actions = null;
        this._parentsScrollSubscriptionInfo = null;
```

js/ui/overlay/ui.overlay.js
```diff
@@ -276,11 +274,15 @@ const Overlay = Widget.inherit({
 
         this._toggleViewPortSubscription(true);
         this._initHideTopOverlayHandler(this.option('hideTopOverlayHandler'));
+        this._parentsScrollSubscriptionInfo = {
+            handler: (e => { this._targetParentsScrollHandler(e); })
```
ksercs: do we really need this brackets`(...)`?

ksercs (dismissed):

js/ui/overlay/ui.overlay.js
alexanderPolosatov: can be removed

testing/tests/DevExpress.ui.widgets/overlay.tests.js
alexanderPolosatov: I think, this test looks like 'funcional', not looks like 'utin-test'. My goal in the next: 'do not have tests on no existing code'. Our exists code is cover by other unit-tests and his test looks like 'extra'.

testing/tests/DevExpress.ui.widgets/overlay.tests.js
alexanderPolosatov: see comment about https://github.com/DevExpress/DevExtreme/pull/18385/files#r682836440

ksercs: I returned it back

testing/tests/DevExpress.ui.widgets/overlay.tests.js
ksercs: I returned it back

ksercs (approved):

DokaRus (approved):

Zedwag (approved):

PR #18434, opened 2021-07-22
DataGrid - Multiple Selection - The row becomes blue except for the column with checkboxes (T1016005)

pomahtri:
Ticket: https://devexpress.com/issue=T1016005

Cherry picks:

- https://github.com/DevExpress/DevExtreme/pull/18435
- https://github.com/DevExpress/DevExtreme/pull/18436

---

When focusing, cell was checked not to have class `dx-cell-focus-disabled`: https://github.com/DevExpress/DevExtreme/blob/21_2/js/ui/grid_core/ui.grid_core.editor_factory.js#L55

But that class was added to row, not to cell: https://github.com/DevExpress/DevExtreme/blob/21_2/js/ui/grid_core/ui.grid_core.keyboard_navigation.js#L1097

Replacing `$focusElement` to `$cell` led tests to fail, so I left both of them

---

Also I had to fix `'An input that resides in a group row template should be focused on click (T931756)'` test

It tested that `td` with `input` should be focused, but according to [this code](https://github.com/DevExpress/DevExtreme/blob/21_2/js/ui/grid_core/ui.grid_core.keyboard_navigation.js#L951) correct behavior is to disable focus on interactive elements (such as `input`)

### DIFF
--- a/testing/tests/DevExpress.ui.widgets.dataGrid/focus.integration.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.dataGrid/focus.integration.tests.js
@@ -3551,9 +3551,32 @@ QUnit.module('View\'s focus', {
         this.clock.tick();
 
         // assert
-        assert.ok($inputElement.closest('td').hasClass('dx-focused'), 'cell is marked as focused');
         assert.ok($inputElement.is(':focus'), 'input is focused');
     });
+
+    QUnit.testInActiveWindow('Cell with checkbox should be focused with other row (T1016005)', function(assert) {
+        // arrange
+        this.dataGrid.option({
+            dataSource: [{ id: 1 }],
+            keyExpr: 'id',
+            columns: ['id'],
+            selection: {
+                mode: 'multiple'
+            },
+            focusedRowEnabled: true,
+        });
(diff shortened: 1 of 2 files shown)


testing/tests/DevExpress.ui.widgets.dataGrid/focus.integration.tests.js
```diff
@@ -3551,10 +3551,34 @@ QUnit.module('View\'s focus', {
...
+
+        const $checkbox = $(this.dataGrid.element()).find('.dx-datagrid-rowsview .dx-checkbox');
+
+        // act
+        $checkbox.trigger('dxpointerdown').trigger('dxclick');
+        this.clock.tick();
+
+        // assert
+        assert.ok($checkbox.parents('tr').hasClass('dx-row-focused'), 'row is focused');
+        assert.ok(!$checkbox.parent('td').hasClass('dx-focused'), 'cell is not focused');
```
vconst, line 3578: check that td element has dx-cell-focus-disabled class

vconst (approved):

Alyar666 (approved):

PR #18452, opened 2021-07-23
Add linting for *.d.ts files

Igggr:
Linting for *.d.ts files
```
npm run lint-dts
```

Note: 
- 'import/named': 'off': `import type {SomeType} from 'file.js'` is more correct way then `import {SomeType} from 'file.js'`, but this will break gulp building task. So, I disabled this rule.
- '@typescript-eslint/no-unnecessary-type-arguments': 'off',  - allow write DxPromise<void>, not just DxPromise (where void is default option)
-'@typescript-eslint/ban-types': 'off'
    - plus: will ban Function, object
    - minus: it look like in few place we need to pass constructors, not type
- '@typescript-eslint/prefer-interface': 'off', - we need types in public d.ts


IlyaKhD (dismissed):

js/core/dom_component.d.ts
```diff
@@ -74,7 +75,22 @@ export interface DOMComponentOptions<TComponent> extends ComponentOptions<TCompo
  * @hidden
  */
 export default class DOMComponent<TProperties = Properties> extends Component<TProperties> {
+    // eslint-disable-next-line no-underscore-dangle
```
pavelgruba:
```suggestion
```

pavelgruba (dismissed):

pavelgruba (dismissed):

pavelgruba (dismissed):

js/localization.d.ts
```diff
@@ -66,15 +66,15 @@ export function locale(): string;
...
+ * @publicName locale(localeToSet)
  * @param1 locale:string
  * @static
  * @module localization
  * @export locale
  * @namespace DevExpress.localization
  * @public
  */
-export function locale(locale: string): void;
+export function locale(localeToSet: string): void;
```
IlyaKhD: This change affects topcis generation
```suggestion
export function locale(locale: string): void;
```

js/core/config.d.ts
```diff
@@ -15,14 +15,14 @@ declare function config(): globalConfig;
 
 /**
  * @docid
- * @publicName config(config)
+ * @publicName config(configuration)
```
IlyaKhD: This change affects topics generation

```suggestion
 * @publicName config(config)
```

IlyaKhD (approved):

PR #18486, opened 2021-07-27
Form - Pass element from DOM tree as a component target element

IgnatovDan:


IgnatovDan:
https://github.com/DevExpress/DevExtreme/pull/18488/files is better

PR #18523, opened 2021-07-29
DataGrid: Fix exception when adding a change object for an invisible row to the editing.changes option (T1016329)

Alyar666:
See https://devexpress.com/issue=T1016329


pomahtri (dismissed):
- [x] Fixes sample from ticket
- [x] Understood fix
- [x] No obvious mistakes
- [ ] Sure it's the best solution
- [x] Has test that failed before fix and pass after

vconst (approved):

PR #18525, opened 2021-07-29
DataGrid: Fix exception when adding a change object for an invisible row to the editing.changes option (T1016329)

Alyar666:
See https://devexpress.com/issue=T1016329


pomahtri (changes requested):
- [ ] Fixes sample from ticket
- [x] Understood fix
- [x] No obvious mistakes
- [ ] Sure it's the best solution
- [x] Has test that failed before fix and pass after 

---

In 21.1 and 20.2 it works correctly, but in 21.2 I faced problem with sample from ticket: when I click "Select all" nothing happens: neither selecting nor throwing exception

vconst (approved):

pomahtri (approved):

PR #18526, opened 2021-07-29
Move to webpack 5

MikeVitik:


webpack.config.js
```diff
@@ -1,8 +1,15 @@
 /* eslint-env node */
 const path = require('path');
-const webpack = require('webpack');
+const TerserPlugin = require('terser-webpack-plugin');
```
nightskylark, line 3: Why you've added this plugin?
https://webpack.js.org/plugins/terser-webpack-plugin/
> If you are using webpack v5 or above you do not need to install this plugin. Webpack v5 comes with the latest terser-webpack-plugin out of the box.

nightskylark (approved):

pavelgruba (approved):

PR #18560, opened 2021-08-02
scrollable: use container() method instead of _$container property

EugeniyKiyashko:


js/ui/scroll_view/ui.scrollable.js
```diff
@@ -515,7 +511,7 @@ const Scrollable = DOMComponent.inherit({
                 : leftPosition;
 
             if(this._isRtlNativeStrategy()) {
-                scrollPosition.left += this._container().get(0).offsetWidth - this._container().get(0).clientWidth;
+                scrollPosition.left += this.container().offsetWidth - this.container.clientWidth;
```
IgnatovDan: this.container.clientWidth -> this.container().clientWidth ?

IgnatovDan (changes requested):

js/ui/scroll_view/ui.scrollable.js
EugeniyKiyashko: fixed in https://github.com/DevExpress/DevExtreme/pull/18560/commits/8704fb4a882b93bf28bbeb6506630887366c1c79

IgnatovDan (approved):

PR #18610, opened 2021-08-04
scrollable: not update scrollbar location when container doesn't have sizes

EugeniyKiyashko:


KuznetsovVN (dismissed):

js/renovation/ui/scroll_view/scrollbar.tsx
```diff
@@ -411,7 +411,10 @@ export class Scrollbar extends JSXComponent<ScrollbarPropsType>() {
     const contentSizeChanged = this.props.contentSize !== this.prevContentSize;
     const containerSizeChanged = this.props.containerSize !== this.prevContainerSize;
 
-    if (contentSizeChanged || containerSizeChanged) {
+    if (this.props.containerDimensionsIsNotZero
```
IgnatovDan: Let's make it the first line of method:
```
if (!this.props.containerDimensionsIsNotZero){
  return;
}
```
Or rename it:
```
if (this.props.containerHasSize){
  return;
}
```
It's very easy to understand what this condition means and why it is applied here.
It's not a beautiful English but it's very easy to understand it.

Do we need the same condition in other methods?

IgnatovDan (dismissed):

js/renovation/ui/scroll_view/scrollbar.tsx
EugeniyKiyashko: I renamed 'containerDimensionsIsNotZero' property to 'containerHasSize**s**' because it is important that the container's  **height & width** dimensions was not zero value simeltainiously. 

As for other methods if it's become a necessity we will add these condition in future PRs. I can't tell yet.

KuznetsovVN (approved):

novsstation (approved):

PR #18729, opened 2021-08-11
PDF Export - Implement word wrap

novsstation:


EugeniyKiyashko (dismissed):

js/exporter/jspdf/export_data_grid_2.js
```diff
@@ -16,13 +16,33 @@ function _getFullOptions(options) {
     return fullOptions;
 }
 
+function calculateColumnWidthsByColSpanAndSplitInfo(pdfGrid, currentRowInfo) {
```
KuznetsovVN, line 19: Maybe move "calculateColumnWidthsByColSpanAndSplitInfo" method to the separated file?
Something like "grid_calculations.js"

novsstation: I think, it's better to keep it in this file, because it's private method and will be used only in one place.
This method logically relates to the building of the rows info, so I place it in the 'export_data_grid_2.js' file.

PR #18803, opened 2021-08-17
ButtonGroup and Toolbar - pass elementAttr to Button widget

novsstation:
- Tests and ts definition for elementAttr in buttonGrpup item
- Tests for passing elementAttr from toolbar to button group item


js/ui/button_group.d.ts
```diff
@@ -174,6 +174,12 @@ export interface dxButtonGroupItem extends CollectionWidgetItem {
      * @public
      */
     type?: 'back' | 'danger' | 'default' | 'normal' | 'success';
+
+    /**
+     * @docid
+     * @public
+     */
+    elementAttr?: { [key: string]: any };
```
EugeniyKiyashko, line 182: This description inherits from dom component. Does we need it?

novsstation: No, collectionwidgetitem doesn't have this property. So, I think this property will not be inherit from DomComponent

image

EugeniyKiyashko (approved):

KuznetsovVN (approved):

PR #18806, opened 2021-08-17
Extract the lightweight core module from the List component

nightskylark:
### Extracted 1 core and 5 additional modules. See the result sizes:
| Bundle | Size |
|-|-|
| ui/list | 637 Kb |
| ui/list_light (Module name will be discussed) | 495 Kb |
| ui/list_light + ui/list/modules/deleting | 542 Kb |
| ui/list_light + ui/list/modules/search | 541 Kb |
| ui/list_light + ui/list/modules/dragging | 526 Kb |
| ui/list_light + ui/list/modules/selection | 525 Kb |
| ui/list_light + ui/list/modules/context | 499 Kb |

### The `deleting` module was splitted on 6 separate modules:
- ui/list/modules/deleting.context
- ui/list/modules/deleting.slideButton
- ui/list/modules/deleting.slideItem
- ui/list/modules/deleting.static
- ui/list/modules/deleting.swipe
- ui/list/modules/deleting.toggle

### Prepared a `modular` page in playground that demonstrates a usage of modules in VanillaJS
For example:
```JS
import ListBase from 'devextreme/ui/list_light';
import 'devextreme/ui/list/modules/search';
new ListBase(div, {
    dataSource: [ 1, 2, 3 ],
    searchEnabled: true
});
```

## What is going to be done next
1. Wrap these modules into the nested comonents for plaforms
2. Throw correct Errors if necessary module found for the set option
3. Prepare testing environment for testing splitted and full components
4. Use lightweight List where possible in our Components


js/ui/list.js
```diff
@@ -1,10 +1,12 @@
-import ListEdit from './list/ui.list.edit.search';
-import registerComponent from '../core/component_registrator';
-// STYLE list
-
-registerComponent('dxList', ListEdit);
+import List from './list_light';
```
alexserov, line 1: Discuss naming with techwriters.

What about: 
`./list/lightweight`
`./list/core`

js/ui/list/modules/deleting.js
```diff
@@ -0,0 +1,6 @@
+import './deleting.context';
+import './deleting.slideButton';
+import './deleting.slideItem';
+import './deleting.static';
+import './deleting.swipe';
+import './deleting.toggle';
```
alexserov, line 6: Discuss naming with TW:
We can use `deletion` or `delete`  instead of `deleting`

js/ui/list.js
```diff
@@ -1,10 +1,12 @@
...
+import './list/modules/selection';
+import './list/modules/deleting';
+import './list/modules/dragging';
+import './list/modules/context';
+import './list/modules/search';
```
alexserov, line 6: I don't really like an idea of implicit plugin usage since this may be not intuitive and the code may be accidentally removed during refactoring
```suggestion
import selection from './list/modules/selection';
import deleting from './list/modules/deleting';
import dragging from './list/modules/dragging';
import context from './list/modules/context';
import search from './list/modules/search';

const ListEdit = List
    .extend(selection)
    .extend(deleting)
    .extend(dragging)
    .extend(context)
    .extend(search)
    
export default ListEdit;
```

js/core/modules_registry.js
```diff
@@ -0,0 +1,21 @@
+const registry = {};
+
+export function setModule(name, module) {
+    registry[name] = module;
+}
```
alexserov: I suggest you use name registration function in the same manner we did it for ComponentRegistrator:
```suggestion
export function registerModule(name, module) {
    registry[name] = module;
}
```

js/core/modules_registry.js
```diff
@@ -0,0 +1,21 @@
...
+export function getModule(name) {
+    return registry[name];
+}
```
alexserov: Exporting `getModule` and `componentGetter` in the same file may lead to incorrect API usage.
```suggestion
function getModule(name) {
    return registry[name];
}
```

js/core/modules_registry.js
```diff
@@ -0,0 +1,21 @@
...
+
+export function componentGetter(path) {
```
alexserov: I suggest you rename the function so that it would describe an action.

```suggestion
export function getComponent(path) {
```

```suggestion
export function queryComponent(path) {
```

Or in the following way:
```suggestion
export function componentFactory(path) {
```

Also, consider replacing `component` with `module`

playground/modular/entry.js
```diff
@@ -0,0 +1,102 @@
...
+    itemDeleteMode: 'swipe'
+});
+
+import 'devextreme/ui/list/modules/deleting.slideItem';
+new ListBase(newElement('itemDeleteMode slideItem'), {
+    dataSource: [ 1, 2, 3 ],
+    allowItemDeleting: true,
+    itemDeleteMode: 'slideItem'
+});
+
```
alexserov, line 46: I think explicit syntax will look better

js/ui/list.js
nightskylark: I don't want to put it to the `list` folder. All API modules are located in the 'ui' and folders are for internal modules.
The name is still discussable. Maybe `list_base` or `list_core`.

js/ui/list.js
nightskylark: I still think that simple is better. Will discuss it separately.

js/core/modules_registry.js
nightskylark: The `getModule` is a base functionality. We'll use it for not-component modules. And the `componentGetter ` is an additional helper.

js/core/modules_registry.js
nightskylark: I like `componentFactory`.

playground/modular/entry.js
nightskylark: Don't forget that the Vanilla JS is not basic scenario of usage. We'll put this dependency to the corresponding nested components for Angular, React and Vue.

js/ui/list.js
alexserov: This is up to you but: for the public API it may be like
`devextreme/ui/list` <--contains everything
`devextreme/ui/list/core` <--logically located inside the list, but only the core part (this is the `list_light` in this PR). So a user can understand that this is a sub-part of a list
`devextreme/ui/list/modules/selection` <-- modules. We can even move it to the outer scope like `devextreme/ui/list/selection` (if the list and other widgets do not contain anything inside)

alexserov (approved):

DokaRus (approved):

PR #18821, opened 2021-08-18
Renovation: default key handlers refactoring

churkin:


js/renovation/component_wrapper/button.ts
```diff
@@ -16,6 +16,10 @@ export default class ButtonWrapper extends Component {
     return ['content'];
   }
 
+  getSupportedKeyNames(): string[] {
```
ksercs, line 19: shouldn't we also specify supported key names in `check_box.ts`?

ksercs (approved):

PR #18862, opened 2021-08-20
DataGrid.Pdf - Group rendering of tables by pages

KuznetsovVN:


js/exporter/jspdf/pdf_grid.js
```diff
@@ -37,13 +40,9 @@ export class PdfGrid {
                 this._currentHorizontalTables.push(
                     new PdfTable(drawTableBorder, this._splitByColumns[i].tableTopLeft, this._columnWidths.slice(beginColumnIndex, endColumnIndex))
                 );
-                if(this._splitByColumns[i].drawOnNewPage) {
-                    this._addLastTableToNewPages();
-                }
+                this._addLastTableToPage(this._splitByColumns[i].drawOnNewPage);
```
IgnatovDan: Can we change it in this way?

const table = new PdfTable(...);
this._pages.push(new PdfPage(table));

IgnatovDan (changes requested):

js/exporter/jspdf/pdf_grid.js
KuznetsovVN: In our tests, we assume that the splitting of the table by columns can be performed both on the current page and with a transfer to a new page.
https://github.com/DevExpress/DevExtreme/blob/21_1/testing/tests/DevExpress.exporter/jspdfParts/jspdf_v2.dataGrid.tests.js#L1457

Although there are no tests for the case when the new table remains on the current page.

In your proposed version, we explicitly remove the ability to add tables for a page that already exists and we will explicitly add a new page and add a table for it.

I think we need to add a new test for the missing case, when "drawOnNewPage: false", like here:
https://github.com/DevExpress/DevExtreme/blob/21_1/testing/tests/DevExpress.exporter/jspdfParts/jspdf_v2.dataGrid.tests.js#L1457
and with this test, check the current logic for grouping tables by pages.

What you think about it?

IgnatovDan (approved):

novsstation (approved):

EugeniyKiyashko (approved):

PR #19003, opened 2021-08-31
ToolbarWrapper - pass rtlEnabled from toolbar to all items

novsstation:


js/renovation/ui/toolbar/toolbar.tsx
```diff
@@ -8,18 +8,31 @@ import LegacyToolbar from '../../../ui/toolbar';
...
-    componentType={LegacyToolbar}
-    componentProps={props}
-  // eslint-disable-next-line react/jsx-props-no-spreading
-    {...restAttributes}
-  />
-);
+}: Toolbar): JSX.Element => {
+  const { rtlEnabled, items } = props;
+
+  items?.forEach((item) => {
```
MikeVitik: Use map it's immutable and  allow you to remove /* eslint-disable no-param-reassign */

js/renovation/ui/toolbar/toolbar_props.tsx
```diff
@@ -114,6 +114,8 @@ export class ToolbarTextBoxProps {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   @Event() onValueChanged?: EventCallback<any>;
 
+  @OneWay()
```
MikeVitik: Try to extract rtlEnabled in to the base ToolbarItemProps (add it if you doesn't have it) and reuse it in other item class using inheritance
@ComponentBindings()
export class ToolbarItemProps {
rtlEnabled?: boolean | undefined;
}
@ComponentBindings()
export class ToolbarCheckBoxProps inherits ToolbarItemProps{
}

js/renovation/ui/toolbar/toolbar.tsx
```diff
@@ -8,18 +8,31 @@ import LegacyToolbar from '../../../ui/toolbar';
...
+    if (typeof item !== 'string') {
```
MikeVitik: We have helper method isString?

MikeVitik: Why do you ignore string defined item? It's also should implement the same rtlEnabled behaviour. Isn't it?

js/renovation/ui/toolbar/toolbar.tsx
```diff
@@ -8,18 +8,31 @@ import LegacyToolbar from '../../../ui/toolbar';
...
+      if (!isDefined(item.options)) {
+        item.options = {};
+      }
+      item.options.rtlEnabled = rtlEnabled ?? false;
```
MikeVitik: We shouldn't rewrite item.options.rtlEnabled if it's defined

MikeVitik: And may be we should get some portion of logic regarding getting rtlEnable from context

MikeVitik: Because props.rtlEnabled it's only one way to get real value of rtlEnabled

js/renovation/ui/toolbar/toolbar.tsx
novsstation: No, if item is string, then it will be rendered as pure string inside the Toolbar.
Because Toolbar has rtl class on it's container, string will be rendered correctly (from right to left).
It's not necessary to add rtlEnabled props to such item

js/renovation/ui/toolbar/toolbar.tsx
novsstation: Yep, fixed via 'for' cycle.
Because even map method will throw errors about 'no-param-reassign'

js/renovation/ui/toolbar/toolbar_props.tsx
novsstation: Yep, great idea. Fixed

js/renovation/ui/toolbar/toolbar.tsx
novsstation: Yep, my mistake.
Fixed

js/renovation/ui/toolbar/toolbar.tsx
novsstation: isString is fixed. Now I use isObject method instead

js/renovation/ui/toolbar/toolbar.tsx
```diff
@@ -1,29 +1,51 @@
...
+import { resolveRtlEnabled } from '../../utils/resolve_rtl';
+
+export const viewFunction = ({ props, rtl, restAttributes }: Toolbar): JSX.Element => {
+  const { items } = props;
+  if (isDefined(items)) {
+    // eslint-disable-next-line @typescript-eslint/prefer-for-of
+    for (let index = 0; index < items.length; index += 1) {
+      const item = items[index];
+      if (!isString(item)) {
+        if (!isDefined(item.options)) {
```
EugeniyKiyashko:
```suggestion
       item.options = item.options ?? {};
```

js/renovation/ui/toolbar/toolbar.tsx
```diff
@@ -1,29 +1,51 @@
...
+import { ConfigContext, ConfigContextValue } from '../../common/config_context';
+import { resolveRtlEnabled } from '../../utils/resolve_rtl';
+
+export const viewFunction = ({ props, rtl, restAttributes }: Toolbar): JSX.Element => {
+  const { items } = props;
+  if (isDefined(items)) {
+    // eslint-disable-next-line @typescript-eslint/prefer-for-of
+    for (let index = 0; index < items.length; index += 1) {
+      const item = items[index];
+      if (!isString(item)) {
```
EugeniyKiyashko: May be we need to expect isPlainObject(item) instead of !isString?

js/renovation/ui/toolbar/toolbar.tsx
```diff
@@ -1,29 +1,51 @@
...
+        if (!isDefined(item.options)) {
+          item.options = {};
+        }
+        item.options.rtlEnabled = item.options.rtlEnabled ?? rtl;
```
EugeniyKiyashko:
```suggestion
       if(isPlainObject(item)) {
           if(isPlainObject(item.options)) {
               item.options.rtlEnabled = item.options.rtlEnabled ?? rtl;
           } else {
               item.options = { rtlEnabled: rtl };
           }
       }
```

js/renovation/ui/toolbar/toolbar_props.tsx
```diff
@@ -105,9 +105,15 @@ export class CollectionWidgetItem {
   ];
   <Toolbar items={toolbarItems}></Toolbar>
 */
+
+@ComponentBindings()
+export class ToolbarItemOptionProps {
```
EugeniyKiyashko:
```suggestion
export class BaseToolbarItemOptionProps {
```

js/renovation/ui/toolbar/toolbar.tsx
```diff
@@ -1,29 +1,51 @@
...
   defaultOptionRules: null,
   view: viewFunction,
 })
 
-export class Toolbar extends JSXComponent<ToolbarProps>() {}
+export class Toolbar extends JSXComponent<ToolbarProps>() {
+  @Consumer(ConfigContext)
+  config?: ConfigContextValue;
+
+  get rtl(): boolean {
```
EugeniyKiyashko: Shall we rename it to rtlEnabled?

novsstation: I think, it may confuse.
Because Toolbar already has prop with name rtlEnabled. Meaning is the same, but it is different code (getter and prop).
So, I prefer original variant

js/renovation/ui/toolbar/toolbar.tsx
novsstation: Yep, IsObject (not isPlainObject) successfully meets my requirements. Thank you

js/renovation/ui/toolbar/toolbar.tsx
novsstation: It,s not necessary to pass rtlEnabled, if item is string

js/renovation/ui/toolbar/toolbar.tsx
```diff
@@ -26,4 +24,27 @@ export const viewFunction = ({
   view: viewFunction,
 })
 
-export class Toolbar extends JSXComponent<ToolbarProps>() {}
+export class Toolbar extends JSXComponent<ToolbarProps>() {
+  @Consumer(ConfigContext)
+  config?: ConfigContextValue;
+
+  get fixedProps(): ToolbarProps {
```
IgnatovDan: What does 'fixed' word mean here?
Why it is easier to understand what does it means than with the 'props1' name?

js/renovation/ui/toolbar/toolbar.tsx
```diff
@@ -26,4 +24,27 @@ export const viewFunction = ({
...
+    const { items } = this.props;
+    const fixedItems = items?.map((item) => {
```
IgnatovDan: Another unclear 'fixed'.
I think it it will be much easier to understand this code with the 'items1' name: I will not guess what 'fixed' means and will immediately investigate the code.

js/renovation/ui/toolbar/__tests__/toolbar.test.tsx
```diff
@@ -11,35 +12,95 @@ import LegacyToolbar from '../../../../ui/toolbar';
...
+    each([false, true, undefined]).describe('rtlEnabled: %o', (isRtlEnabled) => {
+      it('correctly pass rtlEnabled', () => {
+        const buttonItem = new ToolbarItem();
+        const buttonGroupItem = new ToolbarItem();
+        const dropDownButtonItem = new ToolbarItem();
+        const checkBoxItem = new ToolbarItem();
+        const textBoxItem = new ToolbarItem();
+        const toolbarProps = new ToolbarProps();
+        toolbarProps.items = [buttonItem, buttonGroupItem,
+          dropDownButtonItem, checkBoxItem, textBoxItem];
```
IgnatovDan: It's easier to read a flat array:
[new ToolbarItem(), new ToolbarItem(), new ToolbarItem(), new ToolbarItem(), new ToolbarItem()]

IgnatovDan (changes requested):

js/renovation/ui/toolbar/toolbar.tsx
IgnatovDan: Everywhere we use 'rtlEnabled' to specify whether rtl should be enabled or not.
This function calculates new 'rtlEnabled' from several sources.
It's better to keep 'rtlEnabled' in its name (similar to other our rtlEnabled properties) and add a prefix/suffix that easily describes "what is the difference between props.rtlEnabled and this property"
Actual/effective/resulting/calculated/final/techwriters?

IgnatovDan (dismissed):

js/renovation/ui/toolbar/__tests__/toolbar.test.tsx
novsstation: Yep, great idea. Fixed

js/renovation/ui/toolbar/toolbar.tsx
novsstation: Yep, renamed it to toolbarItems

js/renovation/ui/toolbar/toolbar.tsx
novsstation: Yep, current name is difficult to understand. Renamed it to componentProps

js/renovation/ui/toolbar/__tests__/toolbar.test.tsx
novsstation: Sorry, and regarding this question https://github.com/DevExpress/DevExtreme/pull/19003#discussion_r701236622
I cant reply directly to it, so I reply here:
- fixed. Rename rtl to resolvedRtlEnabled, because it resolved from multiple sources

js/renovation/ui/toolbar/toolbar_props.tsx
```diff
@@ -105,9 +105,15 @@ export class CollectionWidgetItem {
   ];
   <Toolbar items={toolbarItems}></Toolbar>
 */
+
+@ComponentBindings()
+export class BaseToolbarItemOptionProps {
```
EugeniyKiyashko:
```suggestion
export class BaseToolbarItemProps {
```

novsstation: Fixed

MikeVitik (approved):

PR #19028, opened 2021-09-02
TreeView - fix keyboard navigation after search (T1022031)

novsstation:


testing/tests/DevExpress.ui.widgets/treeViewParts/searching.js
```diff
@@ -497,3 +497,38 @@ QUnit.test('apply search after searchTimeout', function(assert) {
...
+
+['none', 'selectAll', 'normal'].forEach(selectionMode => {
+    QUnit.test(`focusIn -> search -> focusIn -> clearSearch -> focusIn. selectionMode: ${selectionMode}`, function(assert) {
+        const treeView = initTree({
+            items: [{ id: 1, text: 'item1' }, { id: 2, text: 'item2' }],
+            searchEnabled: true
+        }).dxTreeView('instance');
+
+        const isNodeFocused = (id) => treeView.$element().find(`[data-item-id="${id}"]`).hasClass('dx-state-focused');
+        const triggerFocus = () => treeView._focusInHandler({});
```
IgnatovDan: _focusInHandler is not used in other tests

$(this.$element).trigger('focusin'); ?
$('#input1').focus(); ?
QUnit.testInActiveWindow ?
$parentItem.trigger('dxpointerdown'); ?
wrapper.instance._moveFocus(config.direction, {}); ?
$treeView.focusin(); ?
https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxTreeView/Methods/#focus ?
First list item should be focused on the \'tab\' key press when the search editor is focused ?

testing/tests/DevExpress.ui.widgets/treeViewParts/searching.js
```diff
@@ -497,3 +497,38 @@ QUnit.test('apply search after searchTimeout', function(assert) {
...
+
+        assert.equal(isNodeFocused(1), false, 'item1 is not focused after initialization');
+        assert.equal(isNodeFocused(2), false, 'item2 is not focused after initialization');
```
IgnatovDan, line 524: Will we check 'focusedElement'? It's an independent reference to a focused element.
assert.equal($(this.instance.option('focusedElement')).text(), 'item3'); ?

IgnatovDan (changes requested):

testing/tests/DevExpress.ui.widgets/treeViewParts/searching.js
novsstation: Yep, great idea - fixed

testing/tests/DevExpress.ui.widgets/treeViewParts/searching.js
novsstation: > _focusInHandler is not used in other tests

Yep, my mistake... I use internal method....

> First list item should be focused on the 'tab' key press when the search editor is focused ?

Yep, it's default behavior for default state (when search is not applied).
When customer use search, then the whole TreeView is re-builded. 
So, I think it's normal to send focus to first item (clear focusedElement after search value is changed).

When customer press 'tab', focus moves to next active target (from textBox to treeView nodes container).
In next commit I simulate it via  [$('treeview-nodes-container').trigger('focusin') ](https://github.com/DevExpress/DevExtreme/pull/19028/commits/a56b5a2e28d6375b6abcda6ac214f5532d212b28#diff-3f5aa2a335b4f895ae3794b42d3898ac3145fee4b03edeaaa70023c45f088b3dR511)

I use it because https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxTreeView/Methods/#focus  method sends focus to textEditor instead of TreeView element https://github.com/DevExpress/DevExtreme/blob/21_2/js/ui/widget/ui.search_box_mixin.js#L157

IgnatovDan (dismissed):

IgnatovDan (approved):

PR #19074, opened 2021-09-06
pivotgrid: determine and set correct scrollable strategy on the first render

EugeniyKiyashko:


novsstation (dismissed):

js/ui/pivot_grid/ui.pivot_grid.js
```diff
@@ -1167,6 +1167,14 @@ const PivotGrid = Widget.inherit({
         return this.callBase() && !this._dataController.isLoading();
     },
 
+    getUseNativeOptionValue: function() {
```
IgnatovDan: getUseNativeScrollBarInDataArea ?
(seems to be correct inglish)

js/ui/pivot_grid/ui.pivot_grid.js
```diff
@@ -1177,9 +1185,9 @@ const PivotGrid = Widget.inherit({
         let rowsAreaWidth = 0;
         let hasRowsScroll;
         let hasColumnsScroll;
-        const scrollingOptions = that.option('scrolling') || {};
-        const scrollBarInfo = getScrollBarInfo(scrollingOptions.useNative);
-        const scrollBarWidth = scrollBarInfo.scrollBarWidth;
+
+        const scrollBarUseNative = this.getUseNativeOptionValue();
```
IgnatovDan: const useNativeScrollBarInDataArea?

js/ui/pivot_grid/utils/calculate_scrollbar_width.js
```diff
@@ -0,0 +1,18 @@
...
+import { getScrollbarWidth } from './get_scrollbar_width';
+
+const calculateScrollbarWidth = callOnce(function() {
+    const document = domAdapter.getDocument();
+
+    document.body.insertAdjacentHTML('beforeend',
+        '<div style=\'position: absolute; overflow: scroll; width: 100px; height: 100px; top: -9999;\'></div>');
+
+    const scrollbar = document.body.lastElementChild;
+    const scrollbarWidth = getScrollbarWidth(scrollbar);
```
IgnatovDan, line 12: ok
We don't change scrollbar size in our css and imho we can use a standalone scrollbar instead of creating '$dummyScrollable.dxScrollable(options).dxScrollable('instance');'

js/ui/pivot_grid/ui.pivot_grid.js
```diff
@@ -1349,14 +1360,15 @@ const PivotGrid = Widget.inherit({
...
                 if(scrollingOptions.mode === 'virtual') {
                     that._setVirtualContentParams(scrollingOptions, resultWidths, resultHeights, groupWidth, groupHeight, that._hasHeight, rowsAreaWidth);
                 }
 
                 const updateScrollableResults = [];
                 that._dataArea.updateScrollableOptions({
-                    useNative: !!scrollBarInfo.scrollBarUseNative,
-                    useSimulatedScrollbar: !scrollBarInfo.scrollBarUseNative,
+                    useNative: scrollBarUseNative,
+                    useSimulatedScrollbar: !scrollBarUseNative,
```
IgnatovDan: Can we pass 'useNative' only?
(updateScrollableOptions will manually add 'useSimulatedScrollbar' to its 'restOptions')

IgnatovDan (changes requested):

js/ui/pivot_grid/ui.pivot_grid.js
EugeniyKiyashko: ok, I simplified this logic and passed it to dataArea component

js/ui/pivot_grid/utils/calculate_scrollbar_width.js
EugeniyKiyashko: The browser calculate scrollbar size for varios zoom values. So we need calculate its size for initial setting of the current page. 
We cant use any constant value for this purpose.

js/ui/pivot_grid/ui.pivot_grid.js
EugeniyKiyashko: I may set 'false' value to 'useSimulatedScrollbar' option directly on first initialization. It needs to override calculated default value inside scrollable if we didn't set custom value.

Current behavior:
- useSimulatedScrollbar don't use in **simulated** strategy.
- we use useSimulatedScrollbar: !useNative setting to set false value for native strategy only
- we wont use simulated scrollbars in any strategy

js/ui/pivot_grid/ui.pivot_grid.js
EugeniyKiyashko: I transfer this part direct to dataArea component

IgnatovDan (dismissed):

novsstation (approved):

PR #19075, opened 2021-09-06
Add DefaultOptionsRule type

IlyaKhD:

### DIFF
--- a/js/core/dom_component.d.ts
+++ b/js/core/dom_component.d.ts
@@ -10,7 +10,7 @@ import {
 
 import { TemplateManager } from './template_manager';
 import { FunctionTemplate } from './templates/function_template';
-import { Rule } from './options/utils';
+import { DefaultsRule } from './options';
 
 /* eslint-disable no-underscore-dangle */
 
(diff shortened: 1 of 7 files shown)


js/core/dom_component.d.ts
```diff
@@ -102,7 +102,7 @@ export default class DOMComponent<TProperties = Properties> extends Component<TP
      * @param1_field2 options:Object
      * @public
      */
-    static defaultOptions<TProperties = Properties>(rule: Partial<Rule<TProperties>>): void;
+    static defaultOptions<TProperties = Properties>(rule: Partial<DefaultsRule<TProperties>>): void;
```
IlyaKhD: @alexserov, how do you think, is it really `Partial` ?
```suggestion
    static defaultOptions<TProperties = Properties>(rule: DefaultsRule<TProperties>): void;
```

Same for BaseWidget's [defaultOptions](
https://github.com/DevExpress/DevExtreme/pull/19075/files#diff-6743ea59386f27ddbc3e125c2937a369f0ae545cb5018ab8759bd76a8858c72cR594).

alexserov: A developer should be able to specify a `rule` without necessity of specifying the `DefaultsRule.device` field. However, the `options` field seems to be required.

alexserov (dismissed):

RomanTsukanov (dismissed):

RomanTsukanov (approved):

alexserov (approved):

PR #19085, opened 2021-09-07
Use size.js instead of jQuery width\height helpers

alexserov:


js/core/utils/size.js
```diff
@@ -140,24 +144,55 @@ export const getVisibleHeight = function(element) {
...
+    getHeight: function(el, value) { return elementSizeHelper('height', ...arguments); },
+    setHeight: function(el, value) { return elementSizeHelper('height', ...arguments); },
+    getOuterWidth: function(el, value) { return elementSizeHelper('outerWidth', ...arguments); },
+    setOuterWidth: function(el, value) { return elementSizeHelper('outerWidth', ...arguments); },
+    getOuterHeight: function(el, value) { return elementSizeHelper('outerHeight', ...arguments); },
+    setOuterHeight: function(el, value) { return elementSizeHelper('outerHeight', ...arguments); },
+    getInnerWidth: function(el, value) { return elementSizeHelper('innerWidth', ...arguments); },
+    setInnerWidth: function(el, value) { return elementSizeHelper('innerWidth', ...arguments); },
+    getInnerHeight: function(el, value) { return elementSizeHelper('innerHeight', ...arguments); },
+    setInnerHeight: function(el, value) { return elementSizeHelper('innerHeight', ...arguments); },
```
nightskylark:
```suggestion
    getWidth: (...args) => elementSizeHelper('width', ...args),
    setWidth: (...args) => elementSizeHelper('width', ...args),
    getHeight: (...args) => elementSizeHelper('height', ...args),
    setHeight: (...args) => elementSizeHelper('height', ...args),
    getOuterWidth: (...args) => elementSizeHelper('outerWidth', ...args),
    setOuterWidth: (...args) => elementSizeHelper('outerWidth', ...args),
    getOuterHeight: (...args) => elementSizeHelper('outerHeight', ...args),
    setOuterHeight: (...args) => elementSizeHelper('outerHeight', ...args),
    getInnerWidth: (...args) => elementSizeHelper('innerWidth', ...args),
    setInnerWidth: (...args) => elementSizeHelper('innerWidth', ...args),
    getInnerHeight: (...args) => elementSizeHelper('innerHeight', ...args),
    setInnerHeight: (...args) => elementSizeHelper('innerHeight', ...args),
```

js/core/utils/size.js
```diff
@@ -140,24 +144,55 @@ export const getVisibleHeight = function(element) {
...
     const propName = partialName.toLowerCase();
     const isOuter = sizeProperty.indexOf('outer') === 0;
     const isInner = sizeProperty.indexOf('inner') === 0;
+    const isGetter = arguments.length === 2 || typeof value === 'boolean';
+
+    if(isRenderer(el)) {
+        if(el.length > 1 && !isGetter) {
+            for(let innerIndex = 0; innerIndex < el.length; innerIndex++) {
+                elementSize(el[innerIndex], sizeProperty, value);
+            }
```
nightskylark:
```suggestion
            for(let i= 0; i< el.length; i++) {
                elementSize(el[i], sizeProperty, value);
            }
```

js/core/utils/size.js
```diff
@@ -36,14 +36,18 @@ const getBoxSizingOffset = function(name, elementStyles, boxParams) {
 
     return 0;
 };
+const getComputedStyle = function(element) {
```
nightskylark: The name is the same as window method (https://developer.mozilla.org/ru/docs/Web/API/Window/getComputedStyle).
I recommend to rename it to avoid confusion.

js/core/utils/type.js
```diff
@@ -74,7 +74,7 @@ const isWindow = function(object) {
 };
 
 const isRenderer = function(object) {
-    return !!(object.jquery || object.dxRenderer);
+    return object && !!(object.jquery || object.dxRenderer);
```
nightskylark: The method should return `boolean` but in case the `object` is `null`, it will not be custed.
```suggestion
    return !!(object?.jquery || object?.dxRenderer);
```

js/ui/collection/ui.collection_widget.base.js
```diff
@@ -1044,7 +1045,7 @@ const CollectionWidget = Widget.inherit({
 
         if(items) {
             each(items, function(_, item) {
-                result += $(item).outerWidth(includeMargin || false);
+                result += getOuterWidth($(item), includeMargin || false);
```
nightskylark: Let's try to get rid of wrapping in $ in such cases.

js/ui/gantt/ui.gantt.size_helper.js
```diff
@@ -6,22 +7,26 @@ export class GanttSizeHelper {
     }
 
     _setTreeListDimension(dimension, value) {
-        this._gantt._$treeListWrapper[dimension](value);
-        this._gantt._ganttTreeList?.setOption(dimension, this._gantt._$treeListWrapper[dimension]());
+        const setter = dimension === 'width' ? setWidth : setHeight;
```
nightskylark, line 10: Maybe it will be beter to use a map object like this?
```
sizeMethods: {
  get: {
    width: getWidth,
    height: getHeight
  },
  set: {
    width: setWidth,
    height: setHeight
  }
}
```

js/core/utils/size.js
```diff
@@ -140,24 +144,55 @@ export const getVisibleHeight = function(element) {
...
+export const getHeight = function(el, value) { return commonCallbacks.getHeight(...arguments); };
+export const setHeight = function(el, value) { return commonCallbacks.setHeight(...arguments); };
+export const getOuterWidth = function(el, value) { return commonCallbacks.getOuterWidth(...arguments); };
+export const setOuterWidth = function(el, value) { return commonCallbacks.setOuterWidth(...arguments); };
+export const getOuterHeight = function(el, value) { return commonCallbacks.getOuterHeight(...arguments); };
+export const setOuterHeight = function(el, value) { return commonCallbacks.setOuterHeight(...arguments); };
+export const getInnerWidth = function(el, value) { return commonCallbacks.getInnerWidth(...arguments); };
+export const setInnerWidth = function(el, value) { return commonCallbacks.setInnerWidth(...arguments); };
+export const getInnerHeight = function(el, value) { return commonCallbacks.getInnerHeight(...arguments); };
+export const setInnerHeight = function(el, value) { return commonCallbacks.setInnerHeight(...arguments); };
```
nightskylark: It looks like `commonCallbacks` usage breaks Tree Shaking for these methods.

js/core/utils/size.js
```diff
@@ -140,24 +144,55 @@ export const getVisibleHeight = function(element) {
...
-export const setHeight = (el, value) => elementSize(el, 'height', value);
-export const getOuterWidth = (el) => elementSize(el, 'outerWidth');
-export const setOuterWidth = (el, value) => elementSize(el, 'outerWidth', value);
-export const getOuterHeight = (el) => elementSize(el, 'outerHeight');
-export const setOuterHeight = (el, value) => elementSize(el, 'outerHeight', value);
-export const getInnerWidth = (el) => elementSize(el, 'innerWidth');
-export const setInnerWidth = (el, value) => elementSize(el, 'innerWidth', value);
-export const getInnerHeight = (el) => elementSize(el, 'innerHeight');
-export const setInnerHeight = (el, value) => elementSize(el, 'innerHeight', value);
+export const commonCallbacks = {
```
nightskylark: Don't we need to wrap it into the DEBUG / ENDDEBUG?

js/core/utils/size.js
alexserov: All of these methods share the same implementation (the `elementSize` function). So the `commonCallbacks` costs us only 10 LOCs

js/core/utils/size.js
alexserov: No, in this case we need to create a separate implementation of `getWidth` and other methods. It is impossible since `#debug` and `#enddebug` directives are not enabled now, moreover, I don't see if we have an `#else` directive

js/ui/gantt/ui.gantt.size_helper.js
alexserov: I would prefer to leave refactoring to code owners. 

For example, in current case, defining the `sizeMethods` creates a new member in the `GanttSizeHelper` class. An author may want to keep this member private. Or it may be prefer to declare a helper function (or a map) in the global scope.

js/core/utils/size.js
alexserov: rename to `implementationsMap`, Add the following TODO:
"// TODO: remove when we'll start mocking named exports"

nightskylark (approved):

nightskylark (dismissed):

nightskylark (approved):

PR #19086, opened 2021-09-07
DataGrid - toolbar.disabled and toolbar.visible options

pomahtri:
I added options `toolbar.disabled` and `toolbar.visible`, they are similar to DxToolbar's one.

- `toolbar.disabled` by default is `undefined`, cause it has default definition in DxToolbar. But I wrote to doccomments that default is `false`, cause behavior is the same.
- `toolbar.visible` by default is `undefined`
  - if it's `undefined`, visibility depends on predefined or custom buttons existance. 
  - if it's `true`, it's always visible (even if there's no buttons in toolbar)
  - if it's `false`, it's always invisible (even if there are some buttons in toolbar)

### DIFF
--- a/js/ui/grid_core/ui.grid_core.header_panel.js
+++ b/js/ui/grid_core/ui.grid_core.header_panel.js
@@ -35,8 +35,8 @@ const HeaderPanel = ColumnsView.inherit({
         const options = {
             toolbarOptions: {
                 items: this._getToolbarItems(),
-                visible: userToolbarOptions && userToolbarOptions.visible,
-                disabled: userToolbarOptions && userToolbarOptions.disabled,
+                visible: userToolbarOptions?.visible,
+                disabled: userToolbarOptions?.disabled,
                 onItemRendered: function(e) {
                     const itemRenderedCallback = e.itemData.onItemRendered;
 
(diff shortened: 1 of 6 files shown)


pavel-2020 (dismissed):

js/ui/grid_core/ui.grid_core.header_panel.js
```diff
@@ -48,9 +50,17 @@ const HeaderPanel = ColumnsView.inherit({
 
         this.executeAction('onToolbarPreparing', options);
 
-        if(options.toolbarOptions && !isDefined(options.toolbarOptions.visible)) {
+        if(options.toolbarOptions) {
+            if(!isDefined(options.toolbarOptions.visible)) {
+                options.toolbarOptions.visible = true;
+            }
```
pomahtri: This `if` is needed cause in unittests `defaultOptions` isn't called.  Even if I set `initDefaultOptions: true` in `setupDataGridModules`, this value will be overriden cause in most tests options are set with `this.options = {...}`, not `this.option({...})`

js/ui/grid_core/ui.grid_core.header_panel.js
```diff
@@ -48,9 +50,17 @@ const HeaderPanel = ColumnsView.inherit({
...
+
             const toolbarItems = options.toolbarOptions.items;
-            options.toolbarOptions.visible = !!(toolbarItems && toolbarItems.length);
+            options.toolbarOptions.visible = !!(
+                options.toolbarOptions.visible &&
+                toolbarItems &&
+                toolbarItems.length
```
vconst: I think that only if user does not assign visible property it should be calculated by toolbarItems.length.

js/ui/grid_core/ui.grid_core.header_panel.js
```diff
@@ -187,6 +209,10 @@ const HeaderPanel = ColumnsView.inherit({
 export const headerPanelModule = {
     defaultOptions: function() {
         return {
+            toolbar: {
+                visible: true,
+                disabled: false
```
vconst: I think that visible and disabled default values should not be assigned here. In this case visible will be calculated by toolbar.tems.length

vconst (dismissed):

NickMitrokhin (dismissed):

vconst (approved):

PR #19091, opened 2021-09-07
Form: fix helpText overlapping for validationMessageMode: "always" (T1026923)

babich-a:

### DIFF
--- a/testing/tests/DevExpress.ui.widgets.form/form.materialTheme.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.form/form.materialTheme.tests.js
@@ -520,6 +520,29 @@ QUnit.module('dx-invalid class on dx-field-item-content-wrapper (T949269)', {
         this.clock.tick();
         assert.ok(wrapper.hasClass(invalidClass));
     });
+
+    QUnit.testInActiveWindow('dx-invalid class is added for invalid editor if validationMessageMode: "always" (T1026923)', function(assert) {
+        const formInstance = $('#form').dxForm({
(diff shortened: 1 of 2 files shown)


DokaRus (approved):

novsstation (approved):

testing/tests/DevExpress.ui.widgets.form/form.materialTheme.tests.js
```diff
@@ -520,6 +520,29 @@ QUnit.module('dx-invalid class on dx-field-item-content-wrapper (T949269)', {
...
+                isRequired: true
+            }]
+        }).dxForm('instance');
+
+        formInstance.validate();
+
+        const editorInstance = formInstance.getEditor('field1');
+        const wrapper = $(editorInstance.element()).closest(`.${FIELD_ITEM_CONTENT_WRAPPER_CLASS}`);
+
+        assert.ok(wrapper.hasClass(invalidClass));
```
IgnatovDan, line 544: For this scenario, I would prefer a screenshot test because some CSS class on some DOM element doesn't guarantee that our component works as expected in the client scenario from T1026923 ticket:

assert.ok(wrapper.hasClass(invalidClass));

For me, that class on that element means nothing without deep researching the generated DOM tree and all CSS classes with all its CSS values in each our theme: when it fails, I will have to get know everything about it just to understand what does it mean.

And, the client scenario from T1026923 ticket can fail again (after changes in DOM tree or in CSS settings) even though this test will be passed without errors.

PR #19097, opened 2021-09-07
DataGrid.PDF - Comments with available steps for next development

KuznetsovVN:


novsstation (dismissed):

js/exporter/jspdf/v3/export_data_grid_3.js
```diff
@@ -43,22 +44,36 @@ function exportDataGrid(doc, dataGrid, options) {
...
 
+            // recalculate for grouped rows
+            // TODO: applyGroupIndents()
+
+            // set/update/initBorders(rows);
+            // TODO: initBorders(rows);
+
+            // splitting to pages
+            // TODO: applySplitting()
+            // ??? maybe after [][] => [], and cellInfo must contain rowIndex + colIndex ( for sorting, filtering )
```
IgnatovDan: Here, we can 'split' a cell with 'colSpan > 0' (for example a long 'band' cell) into two cells and place the first cell on the first page and second cell on the second page. The 'split rectangles by pages' algorithms will not split rectangles in this way.
And, we can show initial 'text' in the both new cells.
The same approach we can apply to cells with 'rowSpan > 0' (row headers in Pivot).
Let's add this comment into PR.

IgnatovDan (changes requested):

IgnatovDan (approved):

EugeniyKiyashko (approved):

novsstation (approved):

PR #19111, opened 2021-09-07
PDF - bands implemantation (without wordwrap)

novsstation:


js/exporter/jspdf/v3/draw_utils.js
```diff
@@ -7,7 +7,9 @@ const defaultBorderLineWidth = 1;
 function drawPdfCells(doc, cellsArray) {
     const docStyles = getDocumentStyles(doc);
     cellsArray.forEach(cell => {
-        drawCell(doc, cell, docStyles);
+        if(!cell.isMerged) {
```
KuznetsovVN: As I think, the **isMerged** property should mean that this cell is linked with other cell(s), but that shouldn't mean that it doesn't need to be drawn.
The previous version of the "export to PDF _v2" approach had a **skip** option.
Let's get back this option ?!

novsstation: I think skip is not very informative option name.
In context of drawing it's not clear what reasons refer to skipping of the drawing.
IsMerged adds aditional information, that this cell is merged with another cells and that's why it's necessary to skip it.

IgnatovDan (approved):

KuznetsovVN (approved):

PR #19170, opened 2021-09-10
PDF - implement grouping and summaries

novsstation:


KuznetsovVN (dismissed):

js/exporter/jspdf/v3/row_utils.js
```diff
@@ -67,6 +67,12 @@ function applyRowSpans(rows) {
     }
 }
 
+function resizeColumnsByIndentLevel(rows, options) {
```
IgnatovDan: resizeFirstColumnByIndentLevel ?

js/exporter/jspdf/v3/row_utils.js
```diff
@@ -102,8 +108,9 @@ function calculateCoordinates(doc, rows, options) {
     let y = options?.topLeft?.y ?? 0;
     rows.forEach(row => {
         let x = options?.topLeft?.x ?? 0;
+        const intend = row.indentLevel * options.indent;
         row.cells.forEach(cell => {
-            cell.pdfCell._rect.x = x;
+            cell.pdfCell._rect.x = x + intend;
```
IgnatovDan, line 113: Can we update '_rect.w' near this line?

row.cells[0].pdfCell._rect.w -= row.indentLevel * options.indent;

js/exporter/jspdf/v3/rows_generator.js
```diff
@@ -75,14 +75,15 @@ function generateRowCells({ dataProvider, rowIndex, wordWrapEnabled, colCount, r
                 cellInfo.colSpan = cellMerging.colspan;
             }
         } else if(rowType === 'group') {
-            cellInfo.drawLeftBorder = false;
-            cellInfo.drawRightBorder = false;
+            cellInfo.pdfCell.drawLeftBorder = cellIndex === 0;
+            cellInfo.pdfCell.drawRightBorder = cellIndex === colCount - 1;
```
IgnatovDan, line 79: Why there was 'cellInfo.drawLeftBorder = false;' ?
I have not found a test for it.

IgnatovDan (changes requested):

js/exporter/jspdf/v3/row_utils.js
novsstation: Unfortunately, nope.
Reduced column width is involved in calculating the height of the text of cells.
Decreasing the width may result in a new line of text. So, it's necessary to reduce widths before height calculation.

I also check the approach with virtual cells. Some thing like this:
```js
    rows.forEach(row => {
         if(row.rowType === 'group') {
                row.cells[0].pdfCell._rect.w -= row.indentLevel * options.indent;
                row.cells.splice(0, 0, {
                      pdfCell: {
                          _rect: {
                             w: row.indentLevel * options.indent,
                             h: 0,
                          }
                        }
        });
    });
```

But in this case rowSpan will be broken, as in the image below
image

js/exporter/jspdf/v3/rows_generator.js
novsstation: In the previous version of the export, all cells in the group has no left and right border.
But visually it was not noticeable, because we always draw the table borders

image

In the new version of the export we don't draw the table borders. So, it's necessary to draw the left border for the first cell and the right border for the last cell. (Or first, if it has collSpan)
image

novsstation: Currently, there is a lot of test for checking the left and right borders.

Any [test](https://github.com/DevExpress/DevExtreme/pull/19170/files#diff-31e0e57f3c31cf384714db7acee32571faf50d2f66126e1d1eb01d397783bda8R218) for summaries will check it.

image

In the old version it looks like this 
image

js/exporter/jspdf/v3/row_utils.js
novsstation: Yep, great idea. Fixed https://github.com/DevExpress/DevExtreme/pull/19170/commits/19c6b9d76ceb53a624bc437ac92e2eeb04598169

js/exporter/jspdf/v3/row_utils.js
novsstation: I have check more complex approach with always adding virtual cells to all columns
RowsSpan calculation in this case is ok. 
```js
    rows.forEach(row => {
        row.cells[0].pdfCell._rect.w -= row.indentLevel * options.indent;
        row.cells.splice(0, 0, {
            pdfCell: {
                isVirtual: true,
                _rect: {
                    w: row.indentLevel * options.indent,
                    h: 0,
                }
            }
        });
    });
```

image


But in this case it's necessary to filter those virtal cells in the steps with applyingBorders config and drawing. 
After that,I found that all tests resulting table width grows up by 2 pixels. So, I returned to the first solution

KuznetsovVN (approved):

IgnatovDan (approved):

PR #19224, opened 2021-09-14
Text Editors - Add animated labels 

groshenkovamarina:


js/ui/text_box/text_box.js
```diff
@@ -94,6 +94,20 @@ const TextBox = TextEditor.inherit({
         this._$searchIcon = $searchIcon;
     },
 
+
+    _renderLabel: function() {
+        this.callBase();
+
+        if(this._$searchIcon && this._$label) {
+            const labelBeforeElement = this._$label.find('.dx-label-before');
+            const labelBeforeWidth = labelBeforeElement.width() + this._$searchIcon.outerWidth();
```
babich-a: we need to merge with 21_2 and use getWidth method

js/ui/text_box/text_box.js
```diff
@@ -94,6 +94,20 @@ const TextBox = TextEditor.inherit({
...
+
+            labelBeforeElement.css('width', labelBeforeWidth);
+
+            this._$label.find('.dx-label').css('max-width', this._input().parent().width() - labelBeforeWidth - this._$label.find('.dx-label-after').width());
```
babich-a: I know now why 'max-len' eslint property exists :). I think it a good idea to split this line and add some named variables to know what these calculations do.

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -140,7 +147,9 @@ const TextEditorBase = Editor.inherit({
                     return isMaterial(themeName);
                 },
                 options: {
-                    stylingMode: config().editorStylingMode || 'filled'
+                    stylingMode: config().editorStylingMode || 'filled',
+                    label: '',
```
babich-a: Default value for `label` option is ''. Why do we need to redefine it for material?

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -439,6 +450,42 @@ const TextEditorBase = Editor.inherit({
         this._input().prop('spellcheck', this.option('spellcheck'));
     },
 
+    _renderLabel: function() {
+        const TEXTEDITOR_WITH_BEFORE_BUTTONS_CLASS = 'dx-texteditor-with-before-buttons';
+
+        if(!this.label && this.$element().find('.' + TEXTEDITOR_LABEL_CLASS).length === 1 || this.$element().find('.' + TEXTEDITOR_LABEL_CLASS).length === 2) {
```
babich-a: It seems we need to extract `this.$element().find('.' + TEXTEDITOR_LABEL_CLASS)` as variable

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -439,6 +450,42 @@ const TextEditorBase = Editor.inherit({
...
+            .removeClass(TEXTEDITOR_WITH_BEFORE_BUTTONS_CLASS);
+
+        if(!this.option('label') || this.option('labelMode') === 'hidden') return;
+
+        this.$element().addClass(this.option('labelMode') === 'floating' ? TEXTEDITOR_WITH_FLOATING_LABEL_CLASS : TEXTEDITOR_WITH_LABEL_CLASS);
+
+        const labelText = this.option('label');
+        const $label = this._$label = $('<div>')
+            .addClass(TEXTEDITOR_LABEL_CLASS)
+            .html('<div class="dx-label-before"></div><div class="dx-label"><span>' + labelText + '</span></div><div class="dx-label-after"></div>');
```
babich-a:
```suggestion
            .html(`<div class="dx-label-before"></div><div class="dx-label"><span>${labelText}</span></div><div class="dx-label-after"></div>`);
```

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -439,6 +450,42 @@ const TextEditorBase = Editor.inherit({
...
+
+        $label.appendTo(this.$element());
+
+        if(this._$beforeButtonsContainer) {
+            this.$element().addClass(TEXTEDITOR_WITH_BEFORE_BUTTONS_CLASS);
+            this._$label.find('.dx-label-before').css('width', this._$beforeButtonsContainer.width());
+        }
+
+        this._$label.find('.dx-label').css('max-width', (this._$field ? this._$field.width() : (this._$tagsContainer ? this._$tagsContainer.width() : this._input().width()))) - this._$label.find('.dx-label-after').width();
```
babich-a: too long for me :)

js/ui/text_box/text_box.js
```diff
@@ -94,6 +94,20 @@ const TextBox = TextEditor.inherit({
         this._$searchIcon = $searchIcon;
     },
 
+
+    _renderLabel: function() {
```
DokaRus, line 99: So, why not to extract an animated label as a separate component? 
We will be able to collect the logic of the label in one place, and specific widgets, if necessary, will switch its individual options

groshenkovamarina: What is the difference between placeholder and label? Placeholder is not the separate component. But I agree to use renderLabel method with arguments for lookup and tagbox widgets.

babich-a (approved):

PR #19229, opened 2021-09-15
Add support for plain object in the animationConfig.to field; Fix d.ts

alexserov:

### DIFF
--- a/js/animation/fx.js
+++ b/js/animation/fx.js
@@ -492,8 +492,13 @@ const SlideAnimationConfigurator = {
 const FadeAnimationConfigurator = {
     setup: function($element, config) {
         const from = config.from;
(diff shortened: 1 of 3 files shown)


dxBeardedBear (approved):

js/animation/fx.js
```diff
@@ -492,8 +492,13 @@ const SlideAnimationConfigurator = {
 const FadeAnimationConfigurator = {
     setup: function($element, config) {
         const from = config.from;
-        const fromOpacity = isPlainObject(from) ? (config.skipElementInitialStyles ? 0 : $element.css('opacity')) : String(from);
-        let toOpacity;
+        const to = config.to;
```
nightskylark, line 495: Why don't you use the force of ES6? :)
```
        const { from, to } = config;
```

PR #19231, opened 2021-09-15
Fix DataSource.store option declaration

alexserov:

### DIFF
--- a/js/data/custom_store.d.ts
+++ b/js/data/custom_store.d.ts
@@ -29,7 +29,7 @@ export interface CustomStoreOptions<TKey = any, TValue = any> extends StoreOptio
      * @type_function_return Promise<any>|Array<any>
      * @public
      */
-    load?: ((options: LoadOptions<TValue>) => PromiseLike<TValue> | Array<TValue>);
+    load: ((options: LoadOptions<TValue>) => PromiseLike<TValue> | Array<TValue>);
     /**
      * @docid
      * @default 'processed'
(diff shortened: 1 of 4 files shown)


js/data/custom_store.d.ts
```diff
@@ -29,7 +29,7 @@ export interface CustomStoreOptions<TKey = any, TValue = any> extends StoreOptio
      * @type_function_return Promise<any>|Array<any>
      * @public
      */
-    load?: ((options: LoadOptions<TValue>) => PromiseLike<TValue> | Array<TValue>);
+    load: ((options: LoadOptions<TValue>) => PromiseLike<TValue> | Array<TValue>);
```
alexserov, line 32: @RomanTsukanov We need to introduce a BC after merging this PR, something like follows:
'The `CustomStore.load` option is required now'

RomanTsukanov (dismissed):

js/data/custom_store.d.ts
nightskylark: Not sure it's a BC. It looks like we help customer not to make a mistake in CustomStore configuring.

nightskylark (dismissed):

js/data/data_source.d.ts
```diff
@@ -401,4 +409,5 @@ export default class DataSource<TKey = any, TValue = any> {
     totalCount(): number;
 }
 
+type StorageType = 'array' | 'local' | 'odata';
```
IlyaKhD: This can be removed now

nightskylark (dismissed):

IlyaKhD (approved):

PR #19235, opened 2021-09-15
DataGrid.PDF - Customize paddings for cells

KuznetsovVN:


js/exporter/jspdf/v3/pdf_utils_v3.js
```diff
@@ -5,23 +5,24 @@ function round(value) {
...
-function calculateTextHeight(doc, text, font, { wordWrapEnabled, columnWidth }) {
+function calculateTextHeight(doc, text, padding, font, { wordWrapEnabled, columnWidth }) {
     const height = doc.getTextDimensions(text, {
         fontSize: font?.size || doc.getFontSize()
     }).h;
 
-    const linesCount = getTextLines(doc, text, font, { wordWrapEnabled, columnWidth }).length;
-    return height * linesCount * doc.getLineHeightFactor();
+    const linesCount = getTextLines(doc, text, padding, font, { wordWrapEnabled, columnWidth }).length;
+    return height * linesCount * doc.getLineHeightFactor() + (isDefined(padding) ? padding * 2 : 0);
```
novsstation: I think it's will be read  better with splitting on2 lines

```js
const cellPadding = padding ?? 0;
return (height * linesCount * doc.getLineHeightFactor()) + cellPadding * 2;
```

js/exporter/jspdf/v3/pdf_utils_v3.js
```diff
@@ -61,19 +63,27 @@ function drawRect(doc, x, y, width, height, style) {
...
 
-function drawTextInRect(doc, text, rect, wordWrapEnabled, jsPdfTextOptions) {
-    const textArray = getTextLines(doc, text, doc.getFont(), { wordWrapEnabled, columnWidth: rect.w });
+function drawTextInRect(doc, text, padding, rect, wordWrapEnabled, jsPdfTextOptions) {
+    const textArray = getTextLines(doc, text, padding, doc.getFont(), { wordWrapEnabled, columnWidth: rect.w });
     const linesCount = textArray.length;
 
-    const heightOfOneLine = calculateTextHeight(doc, textArray[0], doc.getFont(), { wordWrapEnabled: false });
+    const heightOfOneLine = calculateTextHeight(doc, textArray[0], 0, doc.getFont(), { wordWrapEnabled: false });
+    const contentRect = isDefined(padding)
```
novsstation: I think it's better to introduce cell padding variable as in this comment https://github.com/DevExpress/DevExtreme/pull/19235/files#r709812183

Something like this:

```js
const cellPadding = padding ?? 0;
const textRect = {
            x: rect.x + cellPadding ,
            y: rect.y + cellPadding ,
            w: rect.w - cellPadding  * 2,
            h: rect.h - cellPadding  * 2
        }
```

And then use textRect variable

js/exporter/jspdf/v3/options.js
```diff
@@ -0,0 +1,32 @@
+import { isDefined, isNumeric } from '../../../core/utils/type';
+
+function normalizeOptions(rows) {
```
novsstation: Is it better to rename this method to normalizePaddings?
Currently it normalize paddings only. But it is discussable

js/exporter/jspdf/v3/options.js
```diff
@@ -0,0 +1,32 @@
...
+    rows.forEach(row => {
+        row.cells.forEach(({ pdfCell }) => {
+            normalizePadding(pdfCell);
+        });
+    });
+}
+
+function normalizePadding(pdfCell) {
```
novsstation: Is it better to rename it to normalizeCellPadding? But it is discussable

js/exporter/jspdf/v3/options.js
```diff
@@ -0,0 +1,32 @@
...
+    if(!isDefined(pdfCell.padding)) {
+        pdfCell.padding = { top: 0, right: 0, bottom: 0, left: 0 };
```
novsstation: Also, I think, we can introduce padding in this method

```js
function generateRowCells({ dataProvider, rowIndex, wordWrapEnabled, colCount, rowType, backgroundColor }) {
    const result = [];
    for(let cellIndex = 0; cellIndex < colCount; cellIndex++) {
        const cellData = dataProvider.getCellData(rowIndex, cellIndex, true);
        const cellInfo = {
            gridCell: cellData.cellSourceData,
            pdfCell: {
                text: cellData.value,
                verticalAlign: 'middle',
                padding: { top: 0, botton: 0, left: 0, fight: 0 }, <--------------------------- where 0 is default value
                wordWrapEnabled,
                backgroundColor,
                _rect: {}
            }
        };
```

Because currently onCustomizeCell callback is executed after rows are generated and before normzlizeOptions is called.
So, in this callback there will be no padding object.

But if add paddings imediatly in the generateCells method (as in the snippet above), then customer can see paddings and modify it in the onCustomizCell.
Also, all styles will be located in the same place. I think, it will be more clear.

We will have such interface of the IRow
```js
// Returns IPdfRowInfo[]
// [
//    {
//      rowType, - readonly
//      rowIndex, - readonly
//      indentLevel, - readonly
//      cells: [ - readonly
//        {
//          colSpan, - readonly (for internal use/hide from api/useless???)
//          rowSpan, - readonly (for internal use/hide from api/useless???)
//          gridCell, - readonly
//          pdfCell : {
//              text,
//              textColor: '#0000ff', // TODO: specify color format for docs
//              backgroundColor: '#0000ff', // TODO: specify color format for docs
//              verticalAlign: 'top' | 'bottom' | 'middle | undefined. Default value is middle
//              wordWrapEnabled, // true, false. Default value is inherited from grid props,
//              padings  <-----------------------------
//              drawRightBorder,
//              drawLeftBorder
//              drawTopBorder
//              drawBottomBorder
//          }
//        }
//      ],
//    }
// ]
```

js/exporter/jspdf/v3/options.js
KuznetsovVN: It is assumed that this method will validate not only padding object but also other parameters ( in future ), for example, color:

**types**  - 'string' | number | [number,number,number]
**values** - '#ff0000' | 128 | [128,128,128]

http://raw.githack.com/MrRio/jsPDF/master/docs/jsPDF.html#setTextColor
https://github.com/simonbengtsson/jsPDF-AutoTable#styling-options

This is the reason why I named the methods like this:

```
function normalizeOptions(rows) {
    rows.forEach(row => {
        row.cells.forEach(({ pdfCell }) => {
            normalizePadding(pdfCell);
            // TODO: normalizeColor(pdfCell);
            // TODO: normalize some other parameter in the cell
        });
    });
}
```

js/exporter/jspdf/v3/options.js
KuznetsovVN: that's a good idea.

js/exporter/jspdf/v3/rows_generator.js
```diff
@@ -43,7 +43,8 @@ function generateRowsInfo(dataProvider, dataGrid, headerBackgroundColor) {
                 columns,
                 rowType,
                 colCount: columns.length,
-                backgroundColor: (rowType === 'header') ? headerBackgroundColor : undefined
+                backgroundColor: (rowType === 'header') ? headerBackgroundColor : undefined,
+                padding: 0
```
novsstation: I think we can omit passing this parameter, since cell creation does not depend on the logic of this method.
Something like this:
```js
function generateRowCells({ dataProvider, rowIndex, wordWrapEnabled, colCount, backgroundColor }) {
    const result = [];
    for(let cellIndex = 0; cellIndex < colCount; cellIndex++) {
        const cellData = dataProvider.getCellData(rowIndex, cellIndex, true);
        const cellInfo = {
            gridCell: cellData.cellSourceData,
            pdfCell: {
                text: cellData.value,
                wordWrapEnabled,
                backgroundColor,
                padding: 0,  <-------- default value
                _rect: {}
            }
        };
```

It's more clear, and other parameters like verticalAlign/horizontalAlign will be placed there too.
Creating of the cell and it's default parameters are placed together.
All the variables on which the creation of the cell not depends are there (default values)
All the variables on which the creation of the cell depends are passed as parameters.

novsstation (dismissed):

js/exporter/jspdf/v3/draw_utils.js
```diff
@@ -22,7 +22,7 @@ function drawCellText(doc, cell, docStyles) {
     if(isDefined(cell.text) && cell.text !== '') { // TODO: use cell.text.trim() ?
         const { textColor, font } = cell;
         setTextStyles(doc, { textColor, font }, docStyles);
-        drawTextInRect(doc, cell.text, cell._rect, cell.verticalAlign, cell.wordWrapEnabled, cell.jsPdfTextOptions);
+        drawTextInRect(doc, cell.text, cell.padding, cell._rect, cell.verticalAlign, cell.wordWrapEnabled, cell.jsPdfTextOptions);
```
novsstation: I think we can calculate text rect in this scope. And pass it to the drawTextInRect function.
Something like:

```js
    const textRect = {
        x: rect.x + padding.left,
        y: rect.y + padding.top,
        w: rect.w - (padding.left + padding.right),
        h: rect.h - (padding.top + padding.bottom)
    };
    
    
    drawTextInRect(doc, cell.text, cell.padding, textRect, cell.verticalAlign, cell.wordWrapEnabled, cell.jsPdfTextOptions);
```

KuznetsovVN: Done

js/exporter/jspdf/v3/options.js
```diff
@@ -0,0 +1,31 @@
...
+        };
+    } else {
+        pdfCell.padding.top = pdfCell.padding.top ?? 0;
+        pdfCell.padding.right = pdfCell.padding.right ?? 0;
+        pdfCell.padding.bottom = pdfCell.padding.bottom ?? 0;
+        pdfCell.padding.left = pdfCell.padding.left ?? 0;
+    }
+}
+
+export { normalizeOptions };
```
IgnatovDan: Rename options.js to normalizeOptions.js ?

js/exporter/jspdf/v3/pdf_utils_v3.js
```diff
@@ -5,22 +5,26 @@ function round(value) {
     return Math.round(value * 1000) / 1000; // checked with browser zoom - 500%
 }
 
-function getTextLines(doc, text, font, { wordWrapEnabled, columnWidth }) {
+function getTextLines(doc, text, font, { wordWrapEnabled, textWidth }) {
```
IgnatovDan: textWidth -> targetAreaWidth ?
(similar to https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Introduction_to_the_CSS_box_model)
Or, targetRectWidth?
(we used 'rect' and don't use 'area')

novsstation (dismissed):

js/exporter/jspdf/v3/pdf_utils_v3.js
```diff
@@ -5,22 +5,26 @@ function round(value) {
...
-        return doc.splitTextToSize(text, columnWidth, {
+        return doc.splitTextToSize(text, textWidth, {
             fontSize: font?.size || doc.getFontSize()
         });
     }
     return text.split('\n');
 }
 
-function calculateTextHeight(doc, text, font, { wordWrapEnabled, columnWidth }) {
+function calculateTextWidth(columnWidth, padding) {
```
IgnatovDan: It doesn't calculate 'width of text'
It calculates 'content area width' (similar to https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Introduction_to_the_CSS_box_model)
calculateContentAreaWidth ?

Or, calculateTargetRectWidth ?
(we used 'rect' and don't use 'area')

js/exporter/jspdf/v3/pdf_utils_v3.js
```diff
@@ -5,22 +5,26 @@ function round(value) {
...
+    return columnWidth - (padding.left + padding.right);
+}
+
+function calculateTextHeight(doc, text, font, { wordWrapEnabled, textWidth }) {
```
IgnatovDan: textWidth -> targetAreaWidth/targetRectWidth?

js/exporter/jspdf/v3/pdf_utils_v3.js
```diff
@@ -68,21 +74,21 @@ function getLineHeightShift(doc) {
     return (doc.getLineHeightFactor() - DEFAULT_LINE_HEIGHT) * doc.getFontSize();
 }
 
-function drawTextInRect(doc, text, rect, verticalAlign, wordWrapEnabled, jsPdfTextOptions) {
-    const textArray = getTextLines(doc, text, doc.getFont(), { wordWrapEnabled, columnWidth: rect.w });
+function drawTextInRect(doc, text, textRect, verticalAlign, wordWrapEnabled, jsPdfTextOptions) {
```
IgnatovDan: imho it's obvious that 'rect' is the 'target rect to paint the text'
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-drawtext - 'rect'
https://developer.apple.com/documentation/uikit/uilabel/1620527-drawtext - 'rect'
It can be 'targetRect'
Why we need to rename it to 'textRect'?
'textRect' means 'rect of text itself', while here it is the 'target rect to show text in'

js/exporter/jspdf/v3/pdf_utils_v3.js
```diff
@@ -68,21 +74,21 @@ function getLineHeightShift(doc) {
...
+    const textArray = getTextLines(doc, text, doc.getFont(), { wordWrapEnabled, textWidth: textRect.w });
```
IgnatovDan: textWidth -> targetAreaWidth/targetRectWidth?

testing/tests/DevExpress.exporter/jspdfParts/v3/jspdf_v3.dataGrid.paddings.tests.js
```diff
@@ -0,0 +1,1443 @@
+import { exportDataGrid } from 'exporter/jspdf/v3/export_data_grid_3';
```
IgnatovDan: 'padding' seems to be an additional config to existing configs.
I would prefer to review tests as 'additional' test near the source test in our common, grouping, summary, bands, and other tests.

IgnatovDan (changes requested):

novsstation (approved):

IgnatovDan (approved):

PR #19258, opened 2021-09-17
Get rid of IE hacks: core utils dom

jacksonrr3:


js/core/utils/dom.js
```diff
@@ -6,21 +6,16 @@ import { getWindow } from './window';
 
 const window = getWindow();
 
-export const resetActiveElement = function() {
+export const resetActiveElement = () => {
     const activeElement = domAdapter.getActiveElement();
     const body = domAdapter.getBody();
 
-    // TODO: remove this hack after msie 11 support stopped
     if(activeElement && activeElement !== body && activeElement.blur) {
```
nightskylark: Why don't you use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)?

nightskylark (changes requested):

nightskylark (approved):

PR #19260, opened 2021-09-17
Get rid of IE hacks: testing helpers

jacksonrr3:


testing/helpers/ariaAccessibilityTestHelper.js
```diff
@@ -36,19 +36,7 @@ class ariaAccessibilityTestHelper {
     }
 
     _getAttributeNames(element) {
```
nightskylark: Once again. Why we need to keep this helper now? What it helps to do?

nightskylark (approved):

PR #19272, opened 2021-09-17
Renovated Editor: fix validationErrors getter to get values not references

Zedwag:


js/renovation/ui/editors/internal/editor.tsx
```diff
@@ -187,9 +187,14 @@ export class Editor extends JSXComponent<EditorPropsType>() {
 
   get validationErrors(): Record<string, unknown>[] | null | undefined {
     const { validationErrors, validationError } = this.props;
-    let allValidationErrors = validationErrors;
```
ksercs, line 190: if we need just to change `validationErrors` object reference, we can only change this row to:
`let allValidationErrors = validationErrors && [...validationErrors];`

Zedwag: right

DokaRus (dismissed):

ksercs:
@Zedwag   Let's add a Unit test for editor to describe the broken scenario. Smth like "validation message text should be update after validationErrors option change"

alexanderPolosatov (dismissed):

ksercs (approved):

DokaRus (approved):

PR #19301, opened 2021-09-20
DomComponent: width/height option change to undefined/null should restore appropriate inline style

ksercs:


Zedwag (dismissed):

nightskylark (commented):
LGTM

ksercs:
We decide to leave the behaviour as it is

PR #19338, opened 2021-09-22
Improve width\height calculation in size helpers

alexserov:
This PR introduces the new `getSize` method realization that: 
- inspired by current jQuery realization;
- does not contain workarounds for non-supported browsers
- rewritten for better readability (ha-ha)
- fixes jQuery bug (Incorrect return value when calling $.outerWidth(..., true) when element shows scrollbar, and display is set to `content-box`)

https://github.com/DevExpress/devextreme-demos/pull/1349


js/core/utils/size.js
```diff
@@ -27,49 +27,106 @@ export const getElementBoxParams = function(name, elementStyles) {
...
 };
+const getCSSProperty = function(element, styles, name, defaultValue) {
+    let result = styles && styles[name];
+    if(result === '' && element.style) {
+        result = element.style[name];
+    }
+    if(result === '' || result === undefined) {
+        return defaultValue;
+    }
+    return result;
```
nightskylark: We can simlify method in case the style value is always string.
```suggestion
    return styles?.[name] || element.style?.[name] || defaultValue
```

nightskylark (changes requested):

nightskylark (dismissed):

nightskylark (dismissed):

nightskylark (dismissed):

nightskylark (approved):

PR #19349, opened 2021-09-23
Get rid of IE hacks: Utils - ReadyCallbacks

jacksonrr3:
Internet Explorer 9 and 10 have bugs where the 'interactive' state can be fired too early before the document has finished parsing.
Since we no longer support Internet Explorer,  we need only checking that `getReadyState()` returned value is not `loading`.

### DIFF
--- a/js/core/utils/ready_callbacks.js
+++ b/js/core/utils/ready_callbacks.js
@@ -14,8 +14,7 @@ const subscribeReady = callOnce(() => {
 const readyCallbacks = {
     add: callback => {
         const windowExists = hasWindow();
-        const isReady = domAdapter.getReadyState() !== 'loading';
-        if(windowExists && isReady) {
+        if(windowExists && domAdapter.getReadyState() !== 'loading') {
             callback();
         } else {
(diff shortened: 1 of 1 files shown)


js/core/utils/ready_callbacks.js
```diff
@@ -19,8 +14,7 @@ const subscribeReady = callOnce(() => {
 const readyCallbacks = {
     add: callback => {
         const windowExists = hasWindow();
-
-        if(windowExists && isReady()) {
+        if(windowExists && domAdapter.getReadyState() !== 'loading') {
```
nightskylark, line 17: So what about the 'interactive' value mentioned in the Note?

nightskylark: It's not obvious why did you made such changes, If everything is correct it need an explanation in the description.

jacksonrr3: We don't need extra checking when IE returns 'interactive' value too early.
So It need only checking that value is not 'loading'

nightskylark (approved):

PR #19352, opened 2021-09-23
Editors: get rid of deprecated in 20.x options

Zedwag:


js/ui/drop_down_box.js
```diff
@@ -290,7 +290,7 @@ const DropDownBox = DropDownEditor.inherit({
             position: extend(this.option('popupPosition'), {
                 of: this.$element(),
             }),
-            onKeyboardHandled: opts => this.option('focusStateEnabled') && this._popupElementTabHandler(opts),
+            onKeyboardHandled: opts => focusStateEnabled && this._popupElementTabHandler(opts),
```
DokaRus: Will this function work correctly after changing the `focusStateEnabled` option at runtime?

js/ui/lookup.js
```diff
@@ -652,7 +644,10 @@ const Lookup = DropDownList.inherit({
                 _fixWrapperPosition: false,
                 width: this._isInitialOptionValue('dropDownOptions.width')
                     ? (function() { return getOuterWidth(this.$element()); }).bind(this)
-                    : this._popupConfig().width
+                    : this._popupConfig().width,
+                height: this._isInitialOptionValue('dropDownOptions.height')
```
ksercs: why? default height should not be equal to lookup input height

js/ui/lookup.js
```diff
@@ -1036,28 +1031,10 @@ const Lookup = DropDownList.inherit({
...
-        });
-    },
-
     _optionChanged: function(args) {
-        const name = args.name;
-        const value = args.value;
+        const { name, fullName, value } = args;
 
-        switch(name) {
+        switch(fullName) {
```
ksercs: I don't think it's a good idea: `name` is used in all other components

js/ui/lookup.js
```diff
@@ -1129,6 +1090,12 @@ const Lookup = DropDownList.inherit({
             case 'cleanSearchOnOpening':
             case '_scrollToSelectedItemEnabled':
                 break;
+            case 'dropDownOptions.width':
```
ksercs: why? When user set option to `auto` he expects option to get 'auto' value, not lookup element width

ksercs:
@Zedwag 
let's also get rid of `popupHeight` in `dom_component.js`

js/ui/lookup.js
Zedwag: fixed

js/ui/lookup.js
Zedwag: https://github.com/DevExpress/DevExtreme/blob/21_2/testing/tests/DevExpress.ui.widgets.editors/lookup.tests.js#L1461

js/ui/lookup.js
Zedwag: ok

js/ui/drop_down_box.js
Zedwag: fixed

Zedwag:
> @Zedwag
> let's also get rid of `popupHeight` in `dom_component.js`

DropDownMenu still has `popupHeight` option

js/ui/lookup.js
ksercs: oh.. ok, let's leave it as it is..

js/ui/lookup.js
```diff
@@ -1129,6 +1087,28 @@ const Lookup = DropDownList.inherit({
...
+                        });
+                        this._options.cache('dropDownOptions', this.option('dropDownOptions'));
+                        break;
+                    case 'dropDownOptions.height':
+                        this._popupOptionChanged({
+                            name,
+                            fullName,
+                            value: value === 'auto' ? this.initialOption('dropDownOptions').height : value
+                        });
+                        this._options.cache('dropDownOptions', this.option('dropDownOptions'));
```
DokaRus:
```suggestion
                    case 'dropDownOptions.width':
                    case 'dropDownOptions.height':
                        this._popupOptionChanged({
                            name,
                            fullName,
                            value: value === 'auto' ? this.initialOption('dropDownOptions')[name] : value
                        });
                        this._options.cache('dropDownOptions', this.option('dropDownOptions'));
                        break;
```

Zedwag: `name` equals 'dropDownOptions' here. I had idea to use something like `this.initialOption('dropDownOptions')[fullName.split('.')[1]]` but this looks realy weird :)

Zedwag: implemented wia `getFieldName` utill

DokaRus (approved):

alexander-kotov-dx (approved):

PR #19359, opened 2021-09-23
HtmlEditor - support new table modules API

DokaRus:


js/ui/html_editor/modules/toolbar.js
```diff
@@ -52,20 +52,14 @@ if(Quill) {
...
-        'insertColumnRight',
-        'deleteColumn',
-        'deleteRow',
-        'deleteTable'
-    ];
-
     const USER_ACTION = 'user';
     const SILENT_ACTION = 'silent';
 
+    const ICON_MAPPER = {
```
ksercs: why `mapper`, not `map`?

js/ui/html_editor/utils/table_helper.js
```diff
@@ -0,0 +1,35 @@
...
+    'insertColumnLeft',
+    'insertColumnRight',
+    'deleteColumn',
+    'deleteRow',
+    'deleteTable'
+];
+
+function getTableFormats(quill) {
+    const tableModule = quill.getModule('table');
+    return tableModule?.tableFormats ? tableModule.tableFormats() : TABLE_FORMATS; // backward compatibility with previous devextreme-quill versions
```
ksercs: seems like this row length is out of 80 characters. Mb we should move comment to the previous row?

ksercs (dismissed):

js/ui/html_editor/modules/toolbar.js
DokaRus: I was thinking of something more global)

Zedwag (approved):

PR #19365, opened 2021-09-23
DataGrid: Improve virtual scrolling performance by setting prerenderedRowChunkSize to 1

vconst:


js/ui/grid_core/ui.grid_core.virtual_scrolling_core.js
```diff
@@ -357,10 +357,13 @@ export const VirtualScrollController = Class.inherit((function() {
                 ? Math.min(this._viewportItemIndex, Math.max(0, totalItemsCount - this._viewportSize))
                 : this._viewportItemIndex;
             const bottomIndex = this._viewportSize + topIndex;
-            const maxGap = this.pageSize();
+            const maxGap = this.option('scrolling.prerenderedRowChunkSize') || 1;
```
NickMitrokhin, line 360: Isn't it better to set prerenderedRowChunkSize to1 in the defaultOptions method?

vconst: I write so to avoid division by zero error.

NickMitrokhin (approved):

PR #19428, opened 2021-09-28
disable @typescript-eslint/no-type-alias for generics

alexserov:

### DIFF
--- a/js/renovation/ui/editors/check_box/check_box.tsx
+++ b/js/renovation/ui/editors/check_box/check_box.tsx
@@ -111,6 +111,7 @@ export class CheckBoxProps extends EditorProps {
 }
 
 export type CheckBoxPropsType = CheckBoxProps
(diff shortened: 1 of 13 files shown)


dxBeardedBear (approved):

hakimio:
@alexserov maybe a better solution would have been to [allow aliases in `in-unions-and-intersections`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-type-alias.md#allowaliases)?

alexserov:
@hakimio I thought about such solution or about [allowing generics (link)](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-type-alias.md#allowgenerics) too, but we decided to keep this rule enabled globally for a while, since we need to discuss it a bit more detailed with the team

PR #19461, opened 2021-09-30
Set label size with editor changing

babich-a:

### DIFF
--- a/js/ui/text_box/ui.text_editor.base.js
+++ b/js/ui/text_box/ui.text_editor.base.js
@@ -75,6 +75,7 @@ const TextEditorBase = Editor.inherit({
 
         this._$beforeButtonsContainer = null;
         this._$afterButtonsContainer = null;
+        this._$labelObservableTarget = null;
 
         this.callBase.apply(this, arguments);
     },
@@ -291,6 +292,8 @@ const TextEditorBase = Editor.inherit({
     _clean() {
         this._buttonCollection.clean();
         this._disposePendingIndicator();
+        this._cleanLabelObservable();
+        this._$labelObservableTarget = null;
         this._$beforeButtonsContainer = null;
         this._$afterButtonsContainer = null;
         this._$textEditorContainer = null;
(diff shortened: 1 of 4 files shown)


js/ui/text_box/ui.text_editor.base.js
```diff
@@ -442,6 +446,26 @@ const TextEditorBase = Editor.inherit({
...
     },
 
+    _cleanLabelObservable: function() {
+        if(this._$labelObservableTarget) {
+            resizeObserverSingleton.unobserve(this._$labelObservableTarget);
+        }
+    },
+
+    _setLabelWidth: function() {
+        this._$labelObservableTarget = $(this._$field ? this._$field : this._$tagsContainer ? this._$tagsContainer : this._input()).get(0);
```
ksercs, line 456: 1. Name. I don't understand why `target`. Maybe `_$labelContainer`? I'm also not sure `observable` word gives us important information
2. Oh, what is `_$field` and `_$tagsContainer`? Imho, TextEditor should not know about its inheritors' private variables. Maybe we can implement `_getLabelContainer` getter and override it down the inheritance tree?

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -442,6 +446,26 @@ const TextEditorBase = Editor.inherit({
...
+
+        const setWidth = () => {
+            const labelWidth = getWidth(this._$labelObservableTarget);
```
ksercs, line 459: We should not get width manually because it's passed as the parameter property

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -442,6 +446,26 @@ const TextEditorBase = Editor.inherit({
...
+            this._$label.find('.dx-label').css('maxWidth', labelWidth);
+        };
+
+        setWidth();
```
ksercs, line 463: `resizeObserverSingleton.observe` call immediatelly plans callback call, so there is no necessity to update width before observe

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -442,6 +446,26 @@ const TextEditorBase = Editor.inherit({
         this._input().prop('spellcheck', this.option('spellcheck'));
     },
 
+    _cleanLabelObservable: function() {
+        if(this._$labelObservableTarget) {
+            resizeObserverSingleton.unobserve(this._$labelObservableTarget);
+        }
+    },
+
+    _setLabelWidth: function() {
```
ksercs, line 455: This function is named as `_setLabelWidth` but it makes several things:
- sets label width
- updates `_$labelObservableTarget` cached value
- initialized resizeObserver subscription

I think we should redevide a responsibility.

Maybe `_updateLabelContainer` (or `_getLabelContainer`) and `_observeLabelResize(shouldObserver)`? This will also get rid of [this method](https://github.com/DevExpress/DevExtreme/pull/19461/files#diff-d5f8ba33de2fd615f123442b1375acee5e5112c3357d545e514ca1838796c197R449)

js/ui/text_box/ui.text_editor.base.js
```diff
@@ -442,6 +446,26 @@ const TextEditorBase = Editor.inherit({
...
+        this._$labelObservableTarget = $(this._$field ? this._$field : this._$tagsContainer ? this._$tagsContainer : this._input()).get(0);
+
+        const setWidth = () => {
+            const labelWidth = getWidth(this._$labelObservableTarget);
+            this._$label.find('.dx-label').css('maxWidth', labelWidth);
+        };
+
+        setWidth();
+
+        this._cleanLabelObservable();
```
ksercs, line 465: we already do it [here](https://github.com/DevExpress/DevExtreme/pull/19461/files#diff-d5f8ba33de2fd615f123442b1375acee5e5112c3357d545e514ca1838796c197R477)

testing/tests/DevExpress.ui.widgets.editors/textEditorParts/common.tests.js
```diff
@@ -462,6 +463,30 @@ QUnit.module('general', {}, () => {
 
         themes.isMaterial = realIsMaterial;
     });
+
+    QUnit.test('label max-width should be changed with editor size (without option change)', function(assert) {
```
ksercs, line 467: I don't think it's a good idea:
1. test is not isolated. `resizeObserverSingleton.observe` is restored inside of a callback which can not be called
2. weird timeout (test can be unstable). Did u test it in different browsers?
3. Does this tests improves a test coverage? We have similar testcafe test

babich-a:
Close now. Code with comments will be used after label component will be extracted

PR #19463, opened 2021-09-30
Editor's label - Extract in a separate component

groshenkovamarina:


js/ui/text_box/text_box.js
```diff
@@ -97,16 +97,13 @@ const TextBox = TextEditor.inherit({
...
-        this.callBase();
-
-        if(this._$searchIcon && this._$label) {
-            const labelBeforeElement = this._$label.find('.dx-label-before');
-            const labelBeforeWidth = getWidth(labelBeforeElement) + getOuterWidth(this._$searchIcon);
+        const options = {
+            editor: this,
+            container: this._input(),
+            icon: this._$searchIcon
+        };
```
DokaRus:
```suggestion
        const options = {
            editor: this,
            container: this._input(),
            icon: this._$searchIcon
        };
```

Let's extract it to a separate method

js/ui/tag_box.js
```diff
@@ -224,6 +225,15 @@ const TagBox = SelectBox.inherit({
...
 
+    _renderLabel: function() {
+        const options = {
+            editor: this,
+            container: this._$tagsContainer
+        };
+
+        renderLabel(options);
+    },
+
```
DokaRus:
```suggestion
    _getLabelOptions: function() {
        return {
            editor: this,
            container: this._$tagsContainer
        };
    },

```
After extraction we may override option getter only

js/ui/lookup.js
```diff
@@ -413,6 +413,13 @@ const Lookup = DropDownList.inherit({
         this.$element().toggleClass(LOOKUP_EMPTY_CLASS, !this.option('selectedItem'));
     },
 
+    _getLabelOptions: function() {
```
ksercs: we can extend `this.callBase()` to reduce copypaste

js/ui/text_box/ui.text_editor.label.js
```diff
@@ -0,0 +1,61 @@
+import { getWidth, getOuterWidth } from '../../core/utils/size';
+import $ from '../../core/renderer';
+
+const TEXTEDITOR_WITH_LABEL_CLASS = 'dx-texteditor-with-label';
+const TEXTEDITOR_WITH_FLOATING_LABEL_CLASS = 'dx-texteditor-with-floating-label';
+const TEXTEDITOR_LABEL_CLASS = 'dx-texteditor-label';
+const TEXTEDITOR_LABEL_SELECTOR = '.' + TEXTEDITOR_LABEL_CLASS;
```
ksercs: we use CLASS constants everywhere, but selector here. Let's unify the approach?

js/ui/text_box/ui.text_editor.label.js
```diff
@@ -0,0 +1,61 @@
...
+const TEXTEDITOR_WITH_BEFORE_BUTTONS_CLASS = 'dx-texteditor-with-before-buttons';
+
+export function renderLabel({ editor, container, icon = null }) {
+    const labelElement = editor.$element().find(TEXTEDITOR_LABEL_SELECTOR);
+    const labelText = editor.option('label');
```
ksercs:
```suggestion
    const { text as labelText, labelMark, labelMode } = editor.option();
```

js/ui/text_box/ui.text_editor.label.js
```diff
@@ -0,0 +1,61 @@
...
+    const labelMark = editor.option('labelMark');
+    const labelMode = editor.option('labelMode');
+
+    if(!editor.label && labelElement.length === 1 || labelElement.length === 2) {
```
ksercs: what is `editor.label`? I can't find `label` property in editor

js/ui/text_box/ui.text_editor.label.js
```diff
@@ -0,0 +1,61 @@
...
+        labelElement.first().remove();
```
ksercs: Looks strange to me. Why do we remove it? What is this element? Why this conditions?

js/ui/text_box/ui.text_editor.label.js
```diff
@@ -0,0 +1,61 @@
...
+        editor._$label.remove();
+        editor._$label = null;
+    }
+
+    editor.$element()
+        .removeClass(TEXTEDITOR_WITH_LABEL_CLASS)
+        .removeClass(TEXTEDITOR_WITH_FLOATING_LABEL_CLASS)
+        .removeClass(TEXTEDITOR_WITH_BEFORE_BUTTONS_CLASS);
+
+    if(!labelText || labelMode === 'hidden') return;
```
ksercs: maybe it's better to write `if (labelText && labelMode !== 'hidden') {` to make this confition more readable and reduce exit points count?

js/ui/text_box/ui.text_editor.label.js
```diff
@@ -0,0 +1,61 @@
...
+
+    setLabelWidth({ editor, container, icon });
+}
+
+
+function setLabelWidth({ editor, container, icon = null }) {
+    const labelBeforeElement = editor._$label.find('.dx-label-before');
+
+    if(editor._$beforeButtonsContainer) {
+        editor.$element().addClass(TEXTEDITOR_WITH_BEFORE_BUTTONS_CLASS);
```
ksercs: Is it really refers to the width setting?

js/ui/text_box/ui.text_editor.label.js
```diff
@@ -0,0 +1,61 @@
...
+        .removeClass(TEXTEDITOR_WITH_FLOATING_LABEL_CLASS)
+        .removeClass(TEXTEDITOR_WITH_BEFORE_BUTTONS_CLASS);
+
+    if(!labelText || labelMode === 'hidden') return;
+
+    editor.$element().addClass(labelMode === 'floating' ? TEXTEDITOR_WITH_FLOATING_LABEL_CLASS : TEXTEDITOR_WITH_LABEL_CLASS);
+
+    const $label = editor._$label = $('<div>')
+        .addClass(TEXTEDITOR_LABEL_CLASS)
+        .html(`<div class="dx-label-before"></div><div class="dx-label"><span data-mark="${labelMark}">${labelText}</span></div><div class="dx-label-after"></div>`);
```
ksercs: I see several reasons to create a new `TextEditorLabel` class, not functions module:
1. Abstraction.
TextEditor calls some magic `renderLabel` public function and use `_$label` private variable after this. Which is this label life circle? What data is it based on? When `_$label` cache is cleared? How we will integrate resize observer?
Imho, label is a new ideological abstraction: it has own markup, props which describe component state, life circle.
2. Incapsulation.
- we mutate the `editor` parameter here :(
- `_$label` is created and configurated by this functions, but it's also used in ddEditor
3. Data passing.
There are 2 functions with the same parameter, which we pass one to another. It's a kind of a copypaste
4. Code style.
It's a consequence of the previous point. We have common data which is used in almost each operation => we cannot create one-purpose functions to improve readability because we don't wanna pass the same data to each new function => we have big functions with not strict responsibility.
5. Testing.
It's easier to test separate component than separate function (btw, what about tests?)

js/ui/text_box/text_box.js
```diff
@@ -95,25 +95,35 @@ const TextBox = TextEditor.inherit({
...
+        return labelBeforeWidth;
     },
 
     _optionChanged: function(args) {
         switch(args.name) {
             case 'maxLength':
                 this._toggleMaxLengthProp();
                 break;
+            case 'mode':
+                this.callBase();
```
ksercs:
```suggestion
                this.callBase(args);
```

sorry :)

ksercs (approved):

ksercs (approved):

DokaRus (approved):

PR #19577, opened 2021-10-07
Collection widgets - add data-based typings

pavelgruba:

### DIFF
--- a/js/ui/validation_summary.d.ts
+++ b/js/ui/validation_summary.d.ts
@@ -7,29 +7,48 @@ import {
 } from '../events/index';
 
 import CollectionWidget, {
+    CollectionWidgetItem,
     CollectionWidgetOptions,
 } from './collection/ui.collection_widget.base';
 
 /** @public */
-export type ContentReadyEvent = EventInfo<dxValidationSummary>;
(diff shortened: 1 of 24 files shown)


js/ui/validation_summary.d.ts
```diff
@@ -43,13 +62,38 @@ export interface dxValidationSummaryOptions extends CollectionWidgetOptions<dxVa
...
+    InitializedEvent: InitializedEvent<TItem, TKey>;
+    ItemClickEvent: ItemClickEvent<TItem, TKey>;
+    OptionChangedEvent: OptionChangedEvent<TItem, TKey>;
+};
+
+/** @public */
+export type Properties<
+    TItem extends string | CollectionWidgetItem<any> | any = any,
+    TKey = any,
+> = dxValidationSummaryOptions<TItem, TKey>;
```
Igggr, line 87: As we must always specify generic constraints, when creating a type alias we have a risk that they start to "differ". I suggest to swap Properties and dxValidationSummaryOptions name - so deprecated option would be defined througth new one, not vise versa. That way - if their type constraints diverge at least that error wouldn't affect thoose, who use modern version.

PR #19623, opened 2021-10-11
DataGrid - Remove index and pageIndex in editing.changes

Alyar666:

### DIFF
--- a/js/ui/grid_core/ui.grid_core.editing.js
+++ b/js/ui/grid_core/ui.grid_core.editing.js
@@ -603,10 +603,6 @@ const EditingController = modules.ViewController.inherit((function() {
                             }
                         } else if(equalByValue(dataController.keyOf(item), key)) {
                             result = index;
-                        } else {
-                            if(equalByValue(item, key)) {
-                                result = index;
-                            }
                         }
                     }
 
--- a/js/ui/grid_core/ui.grid_core.editing.js
+++ b/js/ui/grid_core/ui.grid_core.editing.js
@@ -594,10 +594,10 @@ const EditingController = modules.ViewController.inherit((function() {
                 result = 0;
             } else if(isDefined(key)) {
                 items.some((item, index) => {
-                    isProcessedItems = isProcessedItems || this._isProcessedItem(item);
+                    const isProcessedItem = isProcessedItems || this._isProcessedItem(item);
 
                     if(isObject(item)) {
-                        if(isProcessedItems || isDefined(item[INSERT_INDEX])) {
+                        if(isProcessedItem || isDefined(item[INSERT_INDEX])) {
                             if(equalByValue(item.key, key)) {
                                 result = index;
                             }
@@ -640,8 +640,8 @@ const EditingController = modules.ViewController.inherit((function() {
 
             return item;
         },
-        _getLoadedRowIndex: function(items, change) {
-            let loadedRowIndex = this._getInsertRowIndex(items, change);
+        _getLoadedRowIndex: function(items, change, isProcessedItems) {
+            let loadedRowIndex = this._getInsertRowIndex(items, change, isProcessedItems);
             const dataController = this._dataController;
 
             if(loadedRowIndex < 0) {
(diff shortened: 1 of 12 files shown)


js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -582,40 +581,62 @@ const EditingController = modules.ViewController.inherit((function() {
...
+                result = 0;
+            } else if(isDefined(key)) {
+                items.some((item, index) => {
+                    if(isObject(item)) {
+                        if(this._isProcessedItem(item) || isDefined(item[INSERT_INDEX])) {
+                            if(equalByValue(item.key, key)) {
+                                result = index;
+                            }
+                        } else if(isDefined(keyExpr)) {
+                            if(equalByValue(item[keyExpr], key)) {
```
vconst: Use dataController.keyOf to get key by data. Also you can remove` isDefined(keyExpr) ` condition in this case.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -582,40 +581,62 @@ const EditingController = modules.ViewController.inherit((function() {
...
-        _needInsertItem: function(change, changeType) {
-            const dataController = this._dataController;
-            const dataSource = dataController.dataSource();
-            const scrollingMode = this.option('scrolling.mode');
-            const pageIndex = dataSource.pageIndex();
-            const beginPageIndex = dataSource.beginPageIndex ? dataSource.beginPageIndex() : pageIndex;
-            const endPageIndex = dataSource.endPageIndex ? dataSource.endPageIndex() : pageIndex;
-            const isLastPage = endPageIndex === dataSource.pageCount() - 1;
+        _isProcessedItem: function(item) {
+            return isDefined(item.rowType);
```
vconst: Use more complex condition from ui.data_grid.grouping.js because data item can contains rowType field:
```
isDefined(item.groupIndex) && isString(item.rowType) && item.rowType.indexOf('group') === 0
```
Use can return false in this file and override this method (_IsProcessedItem) in ui.data_grid.grouping.js.

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -795,81 +770,52 @@ const EditingController = modules.ViewController.inherit((function() {
...
                 case PAGE_BOTTOM_NEW_ROW_POSITION:
-                    change.index = scrollingMode === 'standard' ? pageSize : (pageIndex * pageSize) + pageSize;
+                    change.insertAfterKey = allItems[allItems.length - 1].key;
                     break;
-                default:
-                    change.index = change.index ?? this._calculateIndex(rowIndex);
-            }
+                default: {
+                    const isViewportBottom = newRowPosition === VIEWPORT_BOTTOM_NEW_ROW_POSITION;
+                    let visibleItemIndex = isViewportBottom ? rowsView?.getBottomVisibleItemIndex(isViewportBottom) : rowsView?.getTopVisibleItemIndex(isViewportBottom);
```
vconst: It is strange that getBottomVisibleItemIndex and getTopVisibleItemIndex have different parameters here and in the _getInsertRowIndex method.

testing/tests/DevExpress.ui.widgets.dataGrid/dataController.tests.js
```diff
@@ -3589,6 +3589,11 @@ const setupVirtualRenderingModule = function() {
     this.dataController.changed.add(function(e) {
         that.changedArgs.push(e);
     });
+
+    this._views.rowsView = {
+        getTopVisibleItemIndex: () => 0,
+        _getCellElement: () => {}
+    };
```
vconst: Extract this code to helper method to avoid code duplication.

vconst (changes requested):

js/ui/grid_core/ui.grid_core.editing.js
```diff
@@ -582,40 +581,51 @@ const EditingController = modules.ViewController.inherit((function() {
...
-            const scrollingMode = this.option('scrolling.mode');
-            const pageIndex = dataSource.pageIndex();
-            const beginPageIndex = dataSource.beginPageIndex ? dataSource.beginPageIndex() : pageIndex;
-            const endPageIndex = dataSource.endPageIndex ? dataSource.endPageIndex() : pageIndex;
-            const isLastPage = endPageIndex === dataSource.pageCount() - 1;
+            const key = this._getInsertAfterOrBeforeKey(change);
+
+            if(!isDefined(key) && items.length === 0) {
+                result = 0;
+            } else if(isDefined(key)) {
```
NickMitrokhin, line 595: The key may not be defined if you add a row using the addRow method and you are on the first or last page. In case if the newRowPosition is 'first'/'last' this method returns -1 and updateItems will not cause new row rendering. It is necessary to consider the case when the key is undefined and the result should be calculated based on the newRowPosition value.

NickMitrokhin (changes requested):

NickMitrokhin (dismissed):

vconst (approved):

PR #19645, opened 2021-10-12
Menus - add data-based typings

pavelgruba:


js/ui/context_menu.d.ts
```diff
@@ -39,51 +39,55 @@ import {
 } from './collection/ui.collection_widget.base';
 
 /** @public */
-export type ContentReadyEvent = EventInfo<dxContextMenu>;
+export type ContentReadyEvent<TItem extends Item<any> = Item<any>, TKey = any> = EventInfo<dxContextMenu<TItem, TKey>>;
```
novsstation: Could you please clarify this sentence?
```
TItem extends Item<any> = Item<any>
```

If I understood it correctly, it mean, that generic type TItem extends another generic type Item<any>
But what does assignment mean when expanding of the another type

```
TItem extends Item<any> = Item<any>
````

pavelgruba: `TItem` is a generic parameter that should extend `Item`
` = Item` is default value for this generic parameter.
So, if one does not specify any parameter then `Item` is used.
`const myEvent: ContentReadyEvent` is equal to `const myEvent: ContentReadyEvent<Item>`

IlyaKhD (approved):

PR #19677, opened 2021-10-13
Overlay - Deprecate position as function

alexanderPolosatov:


Zedwag (dismissed):

js/ui/overlay/ui.overlay.js
```diff
@@ -259,6 +259,13 @@ const Overlay = Widget.inherit({
         };
 
         this._updateResizeCallbackSkipCondition();
+        this._checkPositionValues();
```
ksercs: why value**s**? :) A lot of them?

Maybe smth like `warnPositionAsFunction`? The original name is too common for me. The comment below suggests that for you too

Zedwag (approved):

DokaRus (approved):

PR #19704, opened 2021-10-15
DataGrid: Do not rerender fixed table during horizontal virtual scrolling

vconst:


js/ui/grid_core/ui.grid_core.column_fixing.js
```diff
@@ -98,21 +124,26 @@ const baseFixedColumns = {
...
 
-            that._correctColumnIndicesForFixedColumns(fixedColumns, change);
+                that._correctColumnIndicesForFixedColumns(fixedColumns, change);
 
-            $fixedTable = that._createTable(fixedColumns);
-            that._renderRows($fixedTable, extend({}, options, { columns: fixedColumns }));
-            that._updateContent($fixedTable, change);
+                $fixedTable = that._createTable(fixedColumns);
+                that._renderRows($fixedTable, extend({}, options, { columns: fixedColumns }));
+                that._updateContent($fixedTable, change);
```
NickMitrokhin: Let's use "this" instead of "that" in all parts of the _renderTable method.

NickMitrokhin (changes requested):

NickMitrokhin (approved):

PR #19712, opened 2021-10-18
DataGrid - typing - fix groupRow key

pomahtri:
Group rows have different typing for `TKey` and `TRowData`. For example see `data` arg in [groupCellComponent](https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/columns/#groupCellComponent); or [in this demo](https://js.devexpress.com/Demos/WidgetsGallery/Demo/DataGrid/RecordGrouping/jQuery/Light/) first group row's key will be `['California']`

I used union here, so user's will need to use `if`s to check current type if they want to access it (see https://github.com/DevExpress/DevExtreme/pull/19724)

I added internal types `TGroupKey` and `TGroupData`

@vconst, please check that:

- [x] all changed types may be group-like
- [x] there's no need to add them to `RowInsertedInfo`, `RowValidatedInfo` etc
- [x] group row can be focused by design
- [x] group row can **not** be selected by design

@IlyaKhD, please check that:

- [x] there's no common mistakes in typing
- [x] I can use **uninon**, not intersection there.
- [x] there's no mistakes in added doccoments (`@type any` and `@type_function_param1 key:any`)
- [x] types that I added need to be internal


js/ui/data_grid.d.ts
```diff
@@ -99,8 +99,14 @@ export interface KeyDownInfo {
   handled: boolean;
 }
 
+type TGroupKey = any[];
+type TGroupData<TRowData> = {
+  key: any;
+  items: Array<TRowData>;
```
vconst: Items can be Array<TGroupData<TRowData>> if several columns are grouped. Also, items can be null if group is collapsed. So, the correct type is
`items: Array<TRowData> | Array<TGroupData<TRowData>> | null;`

Also, TGroupData contains undocumented prop collapsedItems with the same type as items.
Also, TGroupData can contains the following field if local summary is defined:
`aggregates?: Array<any>;`
and the following if remote summary is defined:
`summary?: Array<any>;`
Also, TGroupData can contains the following undocumented parameters:
`isContinuation?: boolean;`
`isContinuationOnNextPage?: boolean;`

js/ui/data_grid.d.ts
```diff
@@ -99,8 +99,19 @@ export interface KeyDownInfo {
   handled: boolean;
 }
 
+type TGroupKey = any[];
+type TGroupData<TRowData> = {
+  key: any;
+  items: Array<TRowData> | Array<TGroupData<TRowData>> | null;
+  collapsedItems: Array<TRowData> | Array<TGroupData<TRowData>> | null;
```
vconst: I checked collapsedItems in the code and the correct type is:
`collapsedItems?: Array<TRowData> | Array<TGroupData<TRowData>>`
So, it is can be not defined, but it cannot be null.

IlyaKhD (approved):

PR #19464, opened 2021-09-30
DataGrid - getting rid of any - TKey, TRowData

pomahtri:

PR #19724, opened 2021-10-19
DataGrid - typing - revert intersect (T1037095)

pomahtri:
Reverting some changes from https://github.com/DevExpress/DevExtreme/pull/19464

---

I thought that intersection was typo, but it was by design. If there was union, users would need to check current type everytime. It's wrong conceptually, but preferable for users, cause they know in compile time whether it's promise or array.

### DIFF
--- a/js/ui/data_grid.d.ts
+++ b/js/ui/data_grid.d.ts
@@ -4353,14 +4353,14 @@ declare class dxDataGrid<TRowData = any, TKey = any> extends Widget<dxDataGridOp
     /**
      * @docid
      * @publicName getSelectedRowKeys()
-     * @return Array<any> & Promise<any>
+     * @return Array<any> | Promise<any>
      * @public
      */
     getSelectedRowKeys(): Array<TKey> & DxPromise<Array<TKey>>;
     /**
      * @docid
      * @publicName getSelectedRowsData()
-     * @return Array<any> & Promise<any>
+     * @return Array<any> | Promise<any>
      * @public
      */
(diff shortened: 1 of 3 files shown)


js/ui/data_grid.d.ts
```diff
@@ -4353,17 +4353,17 @@ declare class dxDataGrid<TRowData = any, TKey = any> extends Widget<dxDataGridOp
     /**
      * @docid
      * @publicName getSelectedRowKeys()
-     * @return Array<any> | Promise<any>
+     * @return Array<any> & Promise<any>
```
IlyaKhD: Actually it's a union, right? We do not provide an object that can used as an `Array` and `Promise` at the same time
```suggestion
     * @return Array<any> | Promise<any>
```

js/ui/data_grid.d.ts
```diff
@@ -4353,17 +4353,17 @@ declare class dxDataGrid<TRowData = any, TKey = any> extends Widget<dxDataGridOp
...
      * @public
      */
-    getSelectedRowKeys(): Array<TKey> | DxPromise<Array<TKey>>;
+    getSelectedRowKeys(): Array<TKey> & DxPromise<Array<TKey>>;
     /**
      * @docid
      * @publicName getSelectedRowsData()
-     * @return Array<any> | Promise<any>
+     * @return Array<any> & Promise<any>
```
IlyaKhD:
```suggestion
     * @return Array<any> | Promise<any>
```

vconst (dismissed):

IlyaKhD (approved):

PR #19736, opened 2021-10-19
Refactor DataSource and Stores types

IlyaKhD:
There're some renamings here:
1. `TValue` was renamed to `TItem` to remove false similarity to "key-value" words combination
2. `TSourceItem` was renamed to `TStoreItem` as it better describes actual type: DataSource accepts a store, not a source
3. I changed DataSource params order: `TStoreItem - TItem - TMappedItem` -> `TStoreItem - TMappedItem - TItem`. This matches to data-flow: it is taken from the store, can be mapped and post-processed after all.

The other changes are:
1. We decided to remove `TKeyExpr` as it has a little meaning but reduces readability a lot
2. I renamed remaining `LocalStore` and `ODataStore` Options types

### DIFF
--- a/js/data/data_source.d.ts
+++ b/js/data/data_source.d.ts
@@ -10,19 +10,20 @@ import { Options as ODataStoreOptions } from './odata/store';
 
 /** @public */
 export type Options<
-    TSourceValue = any,
-    TValue = TSourceValue,
-    TMappedValue = TValue,
+    TSourceItem = any,
+    TItem = TSourceItem,
+    TMappedItem = TItem,
     TKey = any,
-> = DataSourceOptions<TSourceValue, TValue, TMappedValue, TKey>;
+> = DataSourceOptions<TSourceItem, TItem, TMappedItem, TKey>;
 
 /**
  * @namespace DevExpress.data
  * @deprecated Use Options instead
  */
 export interface DataSourceOptions<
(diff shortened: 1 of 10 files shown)


js/data/data_source.d.ts
```diff
@@ -434,6 +431,11 @@ export default class DataSource
  * @docid
  * @type Store|DataSource|DataSourceOptions|string|Array<any>
  * */
-export type DataSourceLike<TItem, TKey = any> = string | Array<TItem> | Store<TItem, any, TKey> | Options<any, TItem, any, any, TKey> | DataSource<TItem, any, TKey>;
+export type DataSourceLike<TItem, TKey = any> =
+    string |
+    Array<TItem> |
+    Store<TItem, TKey> |
+    Options<any, TItem, any, TKey> |
```
alexserov: Since the TMappedItem is the second generic arg, you should change order here:
```suggestion
    Options<any, any, TItem, TKey> |
```

alexserov (changes requested):

js/data/data_source.d.ts
IlyaKhD: Thanks, fixed

RomanTsukanov (approved):

alexserov (approved):

PR #19737, opened 2021-10-19
DateBox: onInput event handler should be called even when useMaskBehavior option is true (T1023540)

Zedwag:


DokaRus (dismissed):

js/ui/date_box/ui.date_box.mask.js
```diff
@@ -155,6 +155,7 @@ const DateBoxMask = DateBoxBase.inherit({
         } else if(this._isSingleCharKey(e)) {
             this._processInputKey(key);
             e.originalEvent.preventDefault();
+            this.option('onInput') && this.option('onInput').call();
```
DokaRus:
```suggestion
            if(this.option('onInput')) {
                const action = this._createActionByOption('onInput', { excludeValidators: ['readOnly'] });
                action({ event: e });
            }
```

Zedwag: implemented

ksercs (approved):

DokaRus (approved):

PR #19822, opened 2021-10-23
Make DataGrid specific types public

IlyaKhD:
It's better to review commit by commit. 
There is a couple of small fixes, and the main commit that consists of same-type changes:

```diff
+/** @public */
+export type Foo = {
-export interface Foo {
...
+};
-}
```

`@public` tag added to remove warning message in the production build. `interface` is changed to `type` to follow our typing conventions.

Note that `DataChange`, `Toolbar` and `ToolbarItem` were added in v21.2, so we can move\rename them before v21.2.3 published

### DIFF
--- a/js/ui/data_grid.d.ts
+++ b/js/ui/data_grid.d.ts
@@ -1638,7 +1638,7 @@ export interface EditingBase<TRowData = any, TKey = any> {
 
 /**
  * @docid
- * @namespace DevExpress.ui
+ * @namespace DevExpress.ui.dxDataGrid
  */
 export interface DataChange<TRowData = any, TKey = any> {
     /**
--- a/js/ui/data_grid.d.ts
+++ b/js/ui/data_grid.d.ts
@@ -4180,19 +4180,18 @@ type DefaultToolbarItemName = 'addRowButton' | 'applyFilterButton' | 'columnChoo
 
 /**
  * @public
- * @docid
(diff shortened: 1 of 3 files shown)


pomahtri:
I changed `defaultToolbarItem` -> `defaultToolbarItemName` and make the similar [to this commit](https://github.com/DevExpress/DevExtreme/pull/19822/commits/c5052d909e6dd4b81df779c8a167b9987ca085d6) for treelist

UPD: I removed commit with TreeList, cause it led to errors in 'TS Declarations' GA task

pomahtri (dismissed):

pavelgruba (dismissed):

pomahtri (dismissed):

js/ui/data_grid.d.ts
```diff
@@ -1638,7 +1638,7 @@ export interface EditingBase<TRowData = any, TKey = any> {
 
 /**
  * @docid
- * @namespace DevExpress.ui
+ * @namespace DevExpress.ui.dxDataGrid
```
pavelgruba, line 1641: I think we can't use such a namespace if the member has a `docid` tag.

IlyaKhD: Suddenly, this is required for ng metadata:

image

pavelgruba (approved):

PR #19828, opened 2021-10-25
SelectBox: fix bug with excess selections when selected item is not on the first page (T1025148)

Zedwag:


testing/tests/DevExpress.ui.widgets.editors/selectBox.tests.js
```diff
@@ -3974,6 +3974,41 @@ QUnit.module('Async tests', {}, () => {
...
+        const selectBox = $selectBox.dxSelectBox('instance');
+        selectBox.open();
+        const $list = $(selectBox.content()).find('.dx-list');
+
+        setTimeout(() => {
+            const $selectedItems = $list.find(toSelector(LIST_ITEM_SELECTED_CLASS));
+
+            assert.strictEqual($selectedItems.length, 0, 'no items are selected');
+            done();
+        }, 0);
```
ksercs: I think it's better to use more readable settimeout and sinon.useFakeTimers for sure test is stable

js/ui/selection/selection.strategy.standard.js
```diff
@@ -390,7 +390,7 @@ export default SelectionStrategy.inherit({
     _isItemSelectionInProgress: function(key, checkPending) {
         const shouldCheckPending = checkPending && this._lastRequestData && this._requestInProgress();
         if(shouldCheckPending) {
-            return this._lastRequestData.addedItems?.indexOf(key) !== -1;
+            return this._lastRequestData.addedItems && this._lastRequestData.addedItems.indexOf(key) !== -1;
```
DokaRus:
```suggestion
            const addedItems = this._lastRequestData.addedItems ?? [];
            return addedItems.includes(key);
```

testing/tests/DevExpress.ui.widgets.editors/selectBox.tests.js
Zedwag: done

js/ui/selection/selection.strategy.standard.js
Zedwag: done

DokaRus (dismissed):

ksercs:
Let's name the PR more specific. Which bug with selection?

The suffix with ticket id does not comply with the our PR naming convention

testing/tests/DevExpress.ui.widgets.editors/selectBox.tests.js
```diff
@@ -3974,6 +3974,46 @@ QUnit.module('Async tests', {}, () => {
...
+            }).dxSelectBox('instance');
+            const $list = $(selectBox.content()).find('.dx-list');
+
+            this.clock.tick(50);
+
+            const $selectedItems = $list.find(toSelector(LIST_ITEM_SELECTED_CLASS));
+
+            assert.strictEqual($selectedItems.length, 0, 'no items are selected');
+        } catch(e) {
+            assert.ok(false, 'Exception: ' + e);
```
ksercs: do we expect some kind of error in this test?

According to the test name, we should check only items selection. If we wanna to check that no errors is raised in some scenario, it's a good idea to write a separate test with a specific name

Zedwag: done

alexanderPolosatov (approved):

alexander-kotov-dx (approved):

PR #19847, opened 2021-10-26
Sortable - The drop indicator is displayed incorrectly after scrolling when dragging an item from another Sortable (T1035958)

pomahtri:
`_subscribeToSourceScroll` was called only in `_dragStartHandler`. That means that scrollable was updating only for one-component dragging. In order to update `itemPoints` in cross-component dragging, I added `_subscribeToSourceScroll` in `_dragEnterHandler`

### DIFF
--- a/js/ui/sortable.js
+++ b/js/ui/sortable.js
@@ -764,7 +764,7 @@ const Sortable = Draggable.inherit({
         return left;
     },
 
-    _movePlaceholder: function(_isFromHandler) {
+    _movePlaceholder: function() {
         const that = this;
         const $placeholderElement = that._$placeholderElement || that._createPlaceholder();
         if(!$placeholderElement) {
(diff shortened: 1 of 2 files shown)


js/ui/sortable.js
```diff
@@ -755,9 +764,13 @@ const Sortable = Draggable.inherit({
         return left;
     },
 
-    _movePlaceholder: function() {
+    _movePlaceholder: function(_isFromHandler) {
```
NickMitrokhin: What for "_isFromHandler"?

NickMitrokhin (changes requested):

js/ui/sortable.js
pomahtri: forgot to remove, fixed

NickMitrokhin (dismissed):

NickMitrokhin (approved):

PR #19922, opened 2021-10-29
SelectBox: input should be displayed from the beggining when searching (T1035997)

Zedwag:


Zedwag:
We decided to leave current behavior in our code and provided a WA for Firefox

PR #19953, opened 2021-11-02
Add native events types to Navigtion widgets

pavelgruba:


IlyaKhD (dismissed):

IlyaKhD (dismissed):

js/ui/scroll_view/ui.scrollable.d.ts
```diff
@@ -15,7 +15,7 @@ import {
     NativeEventInfo,
 } from '../../events/index';
 
-export interface ScrollEventInfo<T> extends NativeEventInfo<T> {
```
EugeniyKiyashko, line 18: + WheelEvent

IgnatovDan (approved):
Our doc describes the 'event' property as "The event that caused the function to execute. It is an EventObject or a jQuery.Event when you use jQuery." without a reference to specific event source.
List of specific event types is a new enhancement to the existing documentation and we are not ready to start working on it.
You can add any type if it was passed in some scenario, simply add a comment 'how you have checked it'.
In this PR I would prefer to provide a very short list of specific event types with manual checks and avoid any additional checks/tests.
We will add new specific types later if someone asks us about that.

EugeniyKiyashko (approved):

PR #19963, opened 2021-11-02
Update component.d.ts to permit cast of option value

ZeBobo5:
Permit cast of value


IlyaKhD (approved):
@ZeBobo5, hello!

I added a commit to your branch to keep TS bundle up-to-date. Thanks for contributing!

PR #20257, opened 2021-11-24
DataGrid - state storing - fix clearing `searchPanel.text` (T1046195)

pomahtri:
I reverted changes from https://github.com/DevExpress/DevExtreme/pull/13079. If `searchPanel.text` was empty, it was ingored even in case it needed to be cleared.

In order to fix previous bug I added condition not to update state on loading if it wasn't saved before

---

I added `filterSync` module in tests to initialize `filterValue` option's default value


js/ui/grid_core/ui.grid_core.state_storing_core.js
```diff
@@ -108,7 +108,9 @@ export const StateStoringController = modules.ViewController.inherit((function()
                 this._isLoaded = true;
                 this._isLoading = false;
             }).done((state) => {
-                this.state(state);
+                if(state !== null) {
```
NickMitrokhin: why?

NickMitrokhin (approved):

PR #20328, opened 2021-11-29
scrollable: refactor list code to take into account perticularities specific for renovated widget

EugeniyKiyashko:


DokaRus (approved):
Perhaps, it would be better to replace multuple `this.option` calls with the single call

```js
const {
  width,
  height,
  scrollingEnabled,
  // more options
} = this.option()
```

novsstation (approved):

PR #20330, opened 2021-11-29
Events - Fix isTrusted property (T1046369)

dxBeardedBear:


js/events/core/events_engine.js
```diff
@@ -595,6 +595,7 @@ initEvent(normalizeEventArguments(function(src, config) {
         that.originalEvent = src;
         that.type = src.type;
         that.currentTarget = undefined;
+        Object.prototype.hasOwnProperty.call(src, 'isTrusted') && (that.isTrusted = src.isTrusted);
```
nightskylark: 1. Make code readable
```suggestion
        if(Object.prototype.hasOwnProperty.call(src, 'isTrusted')) {
            that.isTrusted = src.isTrusted;
        }
```

2. Let's get rid of `that` variable in this function.

nightskylark (changes requested):

js/events/core/events_engine.js
dxBeardedBear: done

nightskylark (approved):

PR #20339, opened 2021-11-29
Renovated Npm - Make it possible to exclude modules from entrypoint

dxBeardedBear:


build/gulp/renovation-npm/index.js
```diff
@@ -139,6 +139,7 @@ addCompilationTask({
     switches: {
         transpile: true
     },
+    components: 'all',
```
alexserov: Consider `undefined` as default value
```suggestion
```

build/gulp/renovation-npm/index.js
```diff
@@ -155,6 +156,7 @@ addCompilationTask({
     switches: {
         installPackages: true,
     },
+    components: ['Button'],
```
alexserov: what if we use glob-style patterns ([minimatch](https://www.npmjs.com/package/minimatch)) here?

dxBeardedBear: done

build/gulp/renovation-npm/index.js
dxBeardedBear: done

alexserov (approved):

PR #20359, opened 2021-11-30
Update internal-tools to 7.5.0

Igggr:
Should fix failing CI


package.json
```diff
@@ -33,13 +33,13 @@
     "devexpress-diagram": "2.1.39",
     "devexpress-gantt": "3.0.19",
     "devextreme-quill": "~1.4.2",
+    "devextreme-showdown": "^1.0.1",
```
IlyaKhD, line 36: Is this required for fixing CI? Or how does it connected with internal-tools bump?

IlyaKhD (approved):

PR #20379, opened 2021-12-01
DataGrid: Fix virtual scrolling performance decrease after virtual scrolling using (T1047506)

vconst:


js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -866,7 +878,9 @@ export const validatingModule = {
                         }
 
                         if(shouldResetValidationState) {
-                            this.getController('validating')._validationState = [];
+                            const validatingController = this.getController('validating');
+                            validatingController._validationState = [];
+                            validatingController._validationStateCache = {};
```
```suggestion
                            validatingController.resetValidationState();
```
I would put these two lines in a separate method and use it instead.

NickMitrokhin (changes requested):

NickMitrokhin (approved):

PR #20406, opened 2021-12-02
TextBox - Fix label overlapping search icon on Material theme if mode is "search" 

groshenkovamarina:

### DIFF
--- a/scss/widgets/material/textEditor/_index.scss
+++ b/scss/widgets/material/textEditor/_index.scss
@@ -30,22 +30,6 @@ $material-editor-label-transition: transform 0.2s cubic-bezier(0, 0, 0.2, 1) 0ms
   @include dx-icon-sizing($material-texteditor-icon-container-size);
 }
 
-@mixin dx-texteditor-with-label-icon($texteditor-padding) {
-  position: absolute;
-  top: 50%;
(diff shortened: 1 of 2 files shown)


scss/widgets/material/textEditor/_index.scss
```diff
@@ -30,6 +30,22 @@ $material-editor-label-transition: transform 0.2s cubic-bezier(0, 0, 0.2, 1) 0ms
...
+@mixin dx-texteditor-with-label-icon($texteditor-padding) {
+  position: absolute;
+  top: 50%;
+  margin-top: math.div(-($material-texteditor-icon-container-size + $texteditor-padding), 2);
+
+  @include dx-icon-sizing($material-texteditor-icon-container-size, $material-texteditor-icon-container-size + $texteditor-padding);
+
+  left: 0;
+
+  @at-root #{selector-append(".dx-rtl", &)},
```
babich-a: can we use `&.dx-rtl` instead of this construction?

groshenkovamarina: I can't use this class for the icon, I can move this code to texteditor styles, but it will be the same block  @at-root #{selector-append(".dx-rtl", &)}

ksercs (approved):

DokaRus (approved):

PR #20435, opened 2021-12-03
TreeList - state(null) does not restore the initial state of the following options: expandedRowKeys, focusedRowKey, and selectedRowKeys (T1049850)

pomahtri:
Changes similar to https://github.com/DevExpress/DevExtreme/pull/20257 for options:

- `expandedRowKeys`
- `focusedRowKey`
- `selectedRowKeys`

### DIFF
--- a/js/ui/grid_core/ui.grid_core.state_storing.js
+++ b/js/ui/grid_core/ui.grid_core.state_storing.js
@@ -172,9 +172,7 @@ export const stateStoringModule = {
                         exportController.selectionOnly(state.exportSelectionOnly);
                     }
 
-                    if(selectedRowKeys) {
-                        that.option('selectedRowKeys', selectedRowKeys);
-                    }
+                    that.option('selectedRowKeys', selectedRowKeys || []);
 
                     that.option('selectionFilter', selectionFilter);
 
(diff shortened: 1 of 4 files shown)


js/ui/grid_core/ui.grid_core.state_storing.js
```diff
@@ -172,8 +172,8 @@ export const stateStoringModule = {
                         exportController.selectionOnly(state.exportSelectionOnly);
                     }
 
-                    if(selectedRowKeys) {
-                        that.option('selectedRowKeys', selectedRowKeys);
+                    if(!that.option('selection.deferred')) {
```
pomahtri, line 175: In case `selection.deferred` is `true` and `applyState` is called when page is just reloaded, calling `that.option('selectedRowKeys', ...)` leads to error [E1042](https://js.devexpress.com/Documentation/ApiReference/UI_Components/Errors_and_Warnings/#E1042) "`keyExpr` is not set". 

It happens because state is applied before dataSource is initialized (See [`_refreshDataSource` overloading in state storing](_refreshDataSource), but changing `selectedRowKeys` in deferred mode requires `keyExpr`

It seems that `selectionFilter` replicates current selected nodes, so I added `!that.option('selection.deferred')` condition

vconst (approved):

PR #20463, opened 2021-12-08
Renovation packages: production mode and dependencies fixes

alexserov:

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "license": "SEE LICENSE IN README.md",
   "dependencies": {
     "@babel/runtime": "^7.12.1",
-    "@devextreme/runtime": "~3.0.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
   "license": "SEE LICENSE IN README.md",
   "dependencies": {
     "@babel/runtime": "^7.12.1",
-    "@devextreme/runtime": "3.0.0",
+    "@devextreme/runtime": "~3.0.0",
```
alexserov: @dxvladislavvolkov I wonder if this change is okay: can I use non-strict version here?

dxvladislavvolkov: I think it is ok

dxBeardedBear (approved):

PR #20464, opened 2021-12-08
datagrid - fixes rendering error row in virtual scrolling mode (T1048220)

NickMitrokhin:


NickMitrokhin:
[T1048220: DataGrid - The error row is duplicated many times when 'virtual' scrolling is used](https://trello.com/c/86LC5DvG/6184-t1048220-datagrid-the-error-row-is-duplicated-many-times-when-virtual-scrolling-is-used)

vconst (approved):

PR #20466, opened 2021-12-08
Change ExternalFormat type to `never` ( fixes #20356 )

### LINKED ISSUE #20356: [TypeScript] `Format` type is `any`
devextreme@21.2.3

https://github.com/DevExpress/DevExtreme/blob/af935aaa8297b1fd57bfa065323d0e4aece667c8/js/localization.d.ts#L109-L120

`Format` type is union type with `any`, so `Format` is `any`.

image

---

As I see it was broken in #17641

alexserov:
This PR fixes #20356

### DIFF
--- a/js/localization.d.ts
+++ b/js/localization.d.ts
@@ -106,7 +106,7 @@ export interface FormatObject {
     */
    type?: PredefinedFormat | string;
 }
-type ExternalFormat = any;
+type ExternalFormat = never;
 
 /**
  * @docid
--- a/js/localization.d.ts
+++ b/js/localization.d.ts
@@ -117,4 +117,11 @@ type ExternalFormat = never;
  * @namespace DevExpress.ui
  * @public
  */
-export type Format = FormatObject | PredefinedFormat | string | ((value: number | Date) => string) | ExternalFormat;
+export type Format =
+  FormatObject |
+  PredefinedFormat |
+  string |
+  ((value: number | Date) => string) |
+  ((value: Date) => string) |
+  ((value: number) => string) |
+  ExternalFormat;
--- a/js/format_helper.d.ts
+++ b/js/format_helper.d.ts
@@ -1,7 +1,11 @@
+import {
+  Format,
+} from './localization';
+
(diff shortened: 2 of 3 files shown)


js/localization.d.ts
```diff
@@ -117,4 +117,11 @@ type ExternalFormat = any;
...
  * @public
  */
-export type Format = FormatObject | PredefinedFormat | string | ((value: number | Date) => string) | ExternalFormat;
+export type Format =
+  FormatObject |
+  PredefinedFormat |
+  string |
+  ((value: number | Date) => string) |
+  ((value: Date) => string) |
+  ((value: number) => string) |
```
Igggr, line 126: I suggest to get rid of 1 or (2 and 3), according to [docs](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html) union should be preffered
```diff
 ((value: number | Date) => string) |
-  ((value: Date) => string) |
-  ((value: number) => string) |
```
Or at least put the version with union last

js/localization.d.ts
```diff
@@ -117,4 +117,11 @@ type ExternalFormat = any;
...
+  ExternalFormat;
```
Igggr, line 127: Wait. You just said, that ExternalFormat is never. Never doesn't do anything in unions - it's like empty set. I suggest to just don't write it in the union.
```diff
((value: Date) => string) |
+ ((value: number) => string);
-  ((value: number) => string) |
-  ExternalFormat;
```

js/localization.d.ts
alexserov: Removing these signatures breaks renovation compilation: I cannot pass a delegate with the `(value: Date) => string` signature as an argument to a function that expects the following one: `(value: number | Date) => string`

js/localization.d.ts
alexserov: `ExternalFormat` is a stub for external format types that we're going to add in future.

Igggr (approved):

PR #20495, opened 2021-12-10
Validator: fix editor accessibility attributes render for renovation and framework wrappers (T1048128, T1049206)

ksercs:


DokaRus (commented):
The target element of the `setAria` method is not always the same as the root element. Has the screen reader been checked to see if it works correctly?

Zedwag (approved):

PR #20512, opened 2021-12-13
DataGrid - Editor is not rendered in a popup form if the infinite scrolling is used and when the column's renderAsync is set to true (T1050488)

pomahtri:
In `ui.grid_core.editing_form_based.js` there were no calls to `_rowsView.renderDelayedTemplates()`, so editors with `columns[].asyncRender` were skipped

I added calling `this._rowsView.renderDelayedTemplates()` to `renderFormEditTemplate` method`


js/ui/grid_core/ui.grid_core.editing_form_based.js
```diff
@@ -289,6 +289,7 @@ export const editingFormBasedModule = {
                     this._rowsView.renderTemplate($container, template, cellOptions, !!$container.closest(getWindow().document).length).done(() => {
                         this._rowsView._updateCell($container, cellOptions);
                     });
+                    this._rowsView.renderDelayedTemplates();
```
vconst: In this place renderDelayedTemplates is called in form editing mode. But the problem is actual for popup editing mode only. Also, renderDelayedTemplates should be called once after render, but now it is called after rendering each column with renderAsync.

The better place to call renderDelayedTemplates is here: https://github.com/DevExpress/DevExtreme/blob/22_1/js/ui/grid_core/ui.grid_core.editing_form_based.js#L203

vconst (approved):

PR #20523, opened 2021-12-13
DropDownButton: Remove template value from filter (T1021257)

dxvladislavvolkov:

### DIFF
--- a/js/ui/selection/selection.strategy.js
+++ b/js/ui/selection/selection.strategy.js
@@ -69,12 +69,19 @@ export default Class.inherit({
         return this.selectedItemKeys(keys, preserve, isDeselect, isSelectAll);
     },
 
+    _removeTemplateProp: function(filterItem) {
+        if(isObject(filterItem) && Object.prototype.hasOwnProperty.call(filterItem, 'template')) {
+            delete filterItem.template;
+        }
+    },
+
     _prepareFilterValue: function(remoteFilter) {
         if(Array.isArray(remoteFilter)) {
-            remoteFilter.forEach((filterItem, index) => {
-                const filterValue = filterItem[2];
-                if(isObject(filterValue) && Object.prototype.hasOwnProperty.call(filterValue, 'template')) {
-                    delete filterValue.template;
+            remoteFilter.forEach((filterItem) => {
+                if(Array.isArray(filterItem)) {
+                    this._prepareFilterValue(filterItem);
+                } else {
+                    this._removeTemplateProp(filterItem);
                 }
             });
         }
--- a/js/ui/selection/selection.strategy.js
+++ b/js/ui/selection/selection.strategy.js
@@ -69,28 +69,20 @@ export default Class.inherit({
         return this.selectedItemKeys(keys, preserve, isDeselect, isSelectAll);
     },
 
-    _removeTemplateProp: function(filterItem) {
-        if(isObject(filterItem) && Object.prototype.hasOwnProperty.call(filterItem, 'template')) {
-            delete filterItem.template;
-        }
-    },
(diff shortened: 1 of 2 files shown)


js/ui/selection/selection.strategy.js
```diff
@@ -69,8 +69,28 @@ export default Class.inherit({
         return this.selectedItemKeys(keys, preserve, isDeselect, isSelectAll);
     },
 
+    _removeTemplateProp: function(filterItem) {
+        if(isObject(filterItem) && Object.prototype.hasOwnProperty.call(filterItem, 'template')) {
```
ksercs: do we really need to check if `template` property exists in the item? Think, we can always delete it

ksercs: shouldn't we return new object, not mutate the parameter?

js/ui/selection/selection.strategy.js
```diff
@@ -69,8 +69,28 @@ export default Class.inherit({
...
+            delete filterItem.template;
+        }
+    },
+
+    _prepareFilterValue: function(remoteFilter) {
```
ksercs: don't we can simplify this recursion? Smth like:
```
_prepareFilterValue: function(remoteFilter) {
       if(Array.isArray(remoteFilter)) {
            return remoteFilter.map(this._prepareFilterValue);
        }

      return this._removeTemplateProp(filterItem);   
 },
```

testing/tests/DevExpress.ui.widgets/selection.test.js
```diff
@@ -2226,6 +2226,17 @@ QUnit.test('Pass filter to load when selection filter', function(assert) {
     assert.deepEqual(selectedKeys, [1, 2, 6], 'selected keys');
 });
 
+QUnit.test('Remove template property from filter value', function(assert) {
+    const selection = this.createDeferredSelection(this.data, {
+        selectionFilter: [['id', '=', { id: 2, template: 'content' }]]
+    });
+
+    // act
```
ksercs:
```suggestion

```

testing/tests/DevExpress.ui.widgets/selection.test.js
```diff
@@ -2226,6 +2226,17 @@ QUnit.test('Pass filter to load when selection filter', function(assert) {
...
+    selection.getSelectedItemKeys().done();
+    // assert
```
ksercs:
```suggestion

```

testing/tests/DevExpress.ui.widgets/selection.test.js
```diff
@@ -2226,6 +2226,17 @@ QUnit.test('Pass filter to load when selection filter', function(assert) {
...
+    assert.equal(this.load.lastCall.args[0].filter[0][2].template, undefined);
```
ksercs:
```suggestion
    assert.strictEqual(this.load.lastCall.args[0].filter[0][2].template, undefined);
```

testing/tests/DevExpress.ui.widgets/selection.test.js
```diff
@@ -2226,6 +2226,17 @@ QUnit.test('Pass filter to load when selection filter', function(assert) {
...
+});
+
```
ksercs, line 2254: Let's write tests for cases when array nesting > 1 and = 0?

js/ui/selection/selection.strategy.js
ksercs: I think we should name this function more concrete. For example, we can name it `_removeTemplateProperty` and union this 2 functions

ksercs (approved):

PR #20528, opened 2021-12-14
Sinon: bump version to 12.0.1

ksercs:
Migration points:
- stub.callsFake replaces stub(obj, ‘meth’, fn)
- replace spy.reset with spy.resetHistory
- change default content-type header parameter when `useFakeXMLHttpRequest` is used ([nise issue](https://github.com/sinonjs/nise/issues/33), sinon [default value specifying code](https://chromium.googlesource.com/chromium/src/third_party/+/refs/heads/main/sinonjs/src/sinon.js?autodive=0%2F%2F#3795) )
- get rid of sinon.test (sinon is a SandBox by default now)
- some timers are increased because of UseFakeTimers re-implementation: nullish timers do not run on the current iteration any more
- get rid of some sinon private API using: `callTimer`, `firstTimerInRange`, etc


ksercs:
I close this PR, because after update we have big problems with timers. 

In the previous sinon versions scarcely planned nullish settimeout was ticked on the current iteration (see [this code](https://github.com/sinonjs/sinon/blob/3dafc8b25e0fd54bb784f16d72334da35cd0bb74/docs/releases/sinon-timers-1.4.2.js#L186)). But now nullish settimeout is planned for a one ms from now: [proof](https://github.com/sinonjs/fake-timers/blob/master/src/fake-timers-src.js#L577).

We have some nullish settimeouts (expecially, in grids), so tests are in a big trouble.

Sample of broken scenarios:
```
QUnit.test('broken scenario', function(assert) {
        setTimeout(() => {
            console.log('main');
            setTimeout(() => {
                console.log('empty');
            });
        }, 1000);
        this.clock.tick(1000);
        // "empty" is not logged presently
    });
```

PR #20576, opened 2021-12-16
Form - Keep validation summary div in nested Forms

IgnatovDan:


andrewmakarov (dismissed):

js/ui/scheduler/recurrence_editor.js
```diff
@@ -431,7 +431,7 @@ class RecurrenceEditor extends Editor {
     _renderEditors($container) {
         this._recurrenceForm = this._createComponent($container, Form, {
             items: this._editors,
-            showValidationSummary: true,
+            showValidationSummary: false,
```
novsstation, line 434: Do we need render validation summary in Scheduler by default?

js/ui/form/ui.form.js
```diff
@@ -314,18 +318,15 @@ const Form = Widget.inherit({
     },
 
     _renderValidationSummary: function() {
-        const $validationSummary = this.$element().find('.' + FORM_VALIDATION_SUMMARY);
-
-        if($validationSummary.length > 0) {
-            $validationSummary.remove();
-        }
+        this._$validationSummary?.remove();
+        this._$validationSummary = undefined;
```
novsstation, line 321: I think it is better to move 2 lines above to a new method disposeValidationSummary.
Something like this:

``` js
 function disposeValidationSummary() {
        this._$validationSummary?.remove();
        this._$validationSummary = undefined;
}

Or maybee it is unnecessary?

js/ui/scheduler/recurrence_editor.js
IgnatovDan: In dxForm, there will be a div with fixed size that reserves empty space when validationSUmmary is enabled.
The root Appointment form have this option enabled.
But the nested Recurrence editor, that internally uses the dxForm, doesn't declare validation rules and should not reserve that space for summary.

novsstation (approved):

js/ui/form/ui.form.js
IgnatovDan: You are right and I have fixed this code duplication.

EugeniyKiyashko (approved):

PR #20615, opened 2021-12-20
HtmlEditor - Update toolbar buttons state on format change by hotkey (T1027453)

alexander-kotov-dx:
see the Devextreme-Quill part of solution here: https://github.com/DevExpress/devextreme-quill/pull/30


js/ui/html_editor/modules/toolbar.js
```diff
@@ -181,6 +194,37 @@ if(Quill) {
...
+        _detectKeyboardFormatChanging() {
+            const namespace = addNamespace('keydown', this.editorInstance.NAME);
+
+            eventsEngine.off(this.editorInstance._getQuillContainer(), namespace);
+            eventsEngine.on(this.editorInstance._getQuillContainer(), namespace, ({ originalEvent }) => {
+                const pressedKey = originalEvent && normalizeKeyName(originalEvent);
+
+                if(pressedKey) {
+                    each(FORMAT_HOTKEYS, (index, { key, formatName }) => {
+                        if(pressedKey.toLowerCase() === key && isCommandKeyPressed(originalEvent)) {
```
ksercs: we can use Object/Map to store FORMAT_KEYS instead of list to find a necessary format with O(1) instead of O(formats count).

It'll be smth like:

```
const FORMAT_HOTKEYS_MAP = { b" 'bold' };

...

const formatName = FORMAT_HOTKEYS_MAP[pressedKey];

if (formatName) {
    this._updateButtonState(formatName);

    return false;
}
...


```

js/ui/html_editor/modules/toolbar.js
```diff
@@ -115,6 +126,8 @@ if(Quill) {
                 e.preventDefault();
             });
 
+            this._detectKeyboardFormatChanging();
```
ksercs: It's not a suitable name for me. 
1. We do not **detect** changes, we just subscribe to a keydown event to find out about some specific key is down later.
2. We detect not a format changing but a key press. Format is changed by ourself.

Mb smth like `_subscribeFormatHotKeys`?

js/ui/html_editor/modules/toolbar.js
```diff
@@ -181,6 +189,41 @@ if(Quill) {
             }
         }
 
+        _subscribeFormatHotKeys() {
+            this.quill.keyboard.addBinding({
+                which: 66,
```
DokaRus: Don't use "magic numbers"

```js
const KEY_CODES = {
  b: 66
};
```

```suggestion
                which: KEY_CODES.B,
```

DokaRus (approved):

alexanderPolosatov (approved):

PR #20621, opened 2021-12-20
List: selectAll should not unselect disabled items (T1050340)

Zedwag:


js/ui/selection/selection.strategy.js
```diff
@@ -137,19 +137,20 @@ export default Class.inherit({
...
         const dataFilter = this.options.filter();
-        let selectedItems = this.options.selectedItems;
+        let selectedItems = this.options.ignoreDisabledItems ? this.options.selectedItems : this.options.selectedItems.filter(item => !item?.disabled);
 
         if(dataFilter) {
             selectedItems = dataQuery(selectedItems).filter(dataFilter).toArray();
         }
 
         const selectedItemsLength = selectedItems.length;
+        const disabledItemsLength = items.length - this.getSelectableItems(items).length;
```
ksercs, line 147: shouldn't it be disabled**Selected**ItemsLength?

testing/tests/DevExpress.ui.widgets/listParts/editingTests.js
```diff
@@ -408,6 +408,24 @@ QUnit.test('selectAll/unselectAll for \'allPages\' selectAllMode', function(asse
     assert.equal(loading.callCount, 2, 'no load during unselect all');
 });
 
+QUnit.test('unselectAll method should not unselect disabled items', function(assert) {
```
ksercs:
```suggestion
QUnit.test('unselectAll method should not unselect disabled items (T1050340)', function(assert) {
```

testing/tests/DevExpress.ui.widgets/listParts/editingTests.js
```diff
@@ -408,6 +408,24 @@ QUnit.test('selectAll/unselectAll for \'allPages\' selectAllMode', function(asse
...
+    const instance = $('#list').dxList({
+        dataSource: items,
+        selectedItemKeys: '1',
+        selectionMode: 'all',
+        keyExpr: 'text',
+    }).dxList('instance');
+
+    instance.selectAll();
+
+    assert.deepEqual(instance.option('selectedItems'), items, 'all item are selected');
```
ksercs: Why all items should be selected here? Shouldn't it be ignored as in the test below?

testing/tests/DevExpress.ui.widgets/listParts/editingTests.js
Zedwag: fixed

testing/tests/DevExpress.ui.widgets/listParts/editingTests.js
Zedwag: `disabled` item is selected on init, so `selectAll()` here just selects second item (not disabled).
actually there is no need in this, I have refactored this test - now they are both selected on init and then we call `unselectAll()`

js/ui/selection/selection.strategy.js
Zedwag: nope, we expect `true` when all non disabled items are selected
all non disabled items = all items - all disabled items

ksercs (approved):

DokaRus (approved):

PR #20634, opened 2021-12-21
Fix display left border for las cell (T1053931)

OlegKipchatov:


babich-a (changes requested):
Where is test screenshot?

babich-a (approved):

PR #20648, opened 2021-12-22
Overlay: add alis for copyRootClassesToWrapper

Zedwag:


Zedwag:
Decided to add info wia Note in docs

PR #20670, opened 2021-12-23
Bump devextreme-internal-tools to 7.7.0

Igggr:
Should fix farm failure on StrongMetaData.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "cssom": "^0.4.4",
     "del": "^2.2.2",
     "devextreme-cldr-data": "^1.0.2",
-    "devextreme-internal-tools": "^7.7.0",
+    "devextreme-internal-tools": "^7",
     "devextreme-screenshot-comparer": "^2.0.9",
     "enzyme": "^3.11.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -80,7 +80,7 @@
     "cssom": "^0.4.4",
     "del": "^2.2.2",
     "devextreme-cldr-data": "^1.0.2",
-    "devextreme-internal-tools": "^7",
+    "devextreme-internal-tools": "^7.7.0",
```
IlyaKhD: Hm... it seems that we agreed to keep "^7" version before.
Better leave it as is then, and update only package-lock

Igggr: In that case - shouldn't I also revert this change in **package-lock.json**
```diff
-        "devextreme-internal-tools": "^7",
+        "devextreme-internal-tools": "^7.7.0",
```
and keep only that:

```diff
 "node_modules/devextreme-internal-tools": {
-      "version": "7.6.0",
+      "version": "7.7.0",
```

IlyaKhD: I guess, yes

PR #20674, opened 2021-12-24
DataGrid.PDF - Split grid into pages byColumns with Bands and Grouping

KuznetsovVN:


js/exporter/jspdf/v3/rows_splitting.js
```diff
@@ -29,6 +30,42 @@ function splitRectsHorizontalByPages(rects, topLeft, maxBottomRight) {
             }
         });
 
+        const currentPageCanBeSplitRects = rectsToSplit.filter(rect => {
+            // Check cells that have 'rect.x' less than 'currentPageMaxRectRight'
+            const currentRectLeft = round(rect.x);
+            const currentRectRight = round(rect.x + rect.w);
+            if(currentRectLeft < currentPageMaxRectRight && currentPageMaxRectRight < currentRectRight) {
```
novsstation, line 37: This algo is executed after calculation of all heights and widths.
It will produce invalid heights for the text

image

Initially it was rendered in all columns
image

js/exporter/jspdf/v3/rows_splitting.js
```diff
@@ -29,6 +30,42 @@ function splitRectsHorizontalByPages(rects, topLeft, maxBottomRight) {
...
+                w: rect.w - (currentPageMaxRectRight - rect.x),
+                h: rect.h
+            }, { sourceCellInfo: Object.assign({}, rect.sourceCellInfo) });
+            rightRect.sourceCellInfo.text = '';
+
+            rectsToSplit.push(rightRect);
+
+            const index = rectsToSplit.indexOf(rect);
+            if(index !== -1) {
+                rectsToSplit.splice(index, 1);
```
novsstation, line 60: It seems there are small mistake in this line
https://github.com/DevExpress/DevExtreme/pull/20674/files#diff-68c01aa7191ff3d2eb5219afa7a97e01b4fea20258ae93127de52fb8b0719bdeR77



It doesn't took in account margin value.

It's necessary to rewrite in such way
```js
rect.x = (currentPageMaxRectRight !== undefined) ? (rect.x - currentPageMaxRectRight + topLeft.x + margin.left) : rect.x;
```

KuznetsovVN: Yes, we missed the case with margins.
I have added the tests for this case. And used this property in calculation.
Done.

js/exporter/jspdf/v3/rows_splitting.js
KuznetsovVN: Let's implement the logic for recalculating the cells by content after splitting into a separate PR. Because this requires special consideration.

novsstation (dismissed):

js/exporter/jspdf/v3/rows_splitting.js
```diff
@@ -29,6 +30,42 @@ function splitRectsHorizontalByPages(rects, topLeft, maxBottomRight) {
...
+            }, { sourceCellInfo: rect.sourceCellInfo });
+
+            currentPageRects.push(leftRect);
+
+            const rightRect = Object.assign({}, {
+                x: currentPageMaxRectRight,
+                y: rect.y,
+                w: rect.w - (currentPageMaxRectRight - rect.x),
+                h: rect.h
+            }, { sourceCellInfo: Object.assign({}, rect.sourceCellInfo) });
```
IgnatovDan: It is not clear why we need to clone the 'sourceCellInfo' object when a new rightRect object is created.

js/exporter/jspdf/v3/rows_splitting.js
```diff
@@ -1,24 +1,25 @@
...
-function splitRectsHorizontalByPages(rects, topLeft, maxBottomRight) {
+function splitRectsHorizontalByPages(rects, margin, topLeft, maxBottomRight) {
     const pages = [];
     const rectsToSplit = [...rects];
 
     while(rectsToSplit.length > 0) {
         let currentPageMaxRectRight = 0;
         const currentPageRects = rectsToSplit.filter(rect => {
-            const currentRectRight = Math.round((rect.x + rect.w) * 1000) / 1000;
+            const currentRectRight = round(rect.x + rect.w);
```
IgnatovDan: It's not clear why we don't need math.round now, it was before:

const currentRectRight = Math.round((rect.x + rect.w) * 1000) / 1000;

IgnatovDan (changes requested):

js/exporter/jspdf/v3/rows_splitting.js
KuznetsovVN: We have identical code in 'draw_utils.js' file:
https://github.com/DevExpress/DevExtreme/blob/22_1/js/exporter/jspdf/v3/draw_utils.js#L8
In this place, I got rid of duplicate code.

js/exporter/jspdf/v3/rows_splitting.js
KuznetsovVN: Cloning will provide partial differences from the original (content, styles, borders)

If we want to use one object for two divided cells, then we need to modify the 'sourceCellInfo' object for using arrays of rects and texts

What you think about this?

js/exporter/jspdf/v3/draw_utils.js
```diff
@@ -4,7 +4,7 @@ import { calculateTextHeight, getTextLines } from './pdf_utils_v3';
 
 const defaultBorderLineWidth = 1;
 
-function round(value) {
+function roundScaledValue(value) {
```
IgnatovDan: roundToThreeDecimals ?

js/exporter/jspdf/v3/rows_splitting.js
```diff
@@ -29,6 +30,38 @@ function splitRectsHorizontalByPages(rects, topLeft, maxBottomRight) {
             }
         });
 
+        const currentPageCanBeSplitRects = rectsToSplit.filter(rect => {
```
IgnatovDan: It's a pity that the 'split' word is arranged for the argument name.
Can we rename that argument?
Or this variable?

rectsToBisect ? rectsToCutInHalf ? rectsToHalve?

onSplitRectHorizontally - should be synced (onHalveRectHorisontally?)

js/exporter/jspdf/v3/rows_splitting.js
```diff
@@ -1,24 +1,25 @@
 import { isDefined } from '../../../core/utils/type';
+import { roundScaledValue } from './draw_utils';
 
-function splitRectsByPages(rects, topLeft, maxBottomRight) {
+function splitRectsByPages(rects, margin, topLeft, maxBottomRight, onSplitRectHorizontally) {
```
IgnatovDan: allocateRectsByPages? arrangeRectsByPages? sortRectsByPages? Vlada?

IgnatovDan (changes requested):

js/exporter/jspdf/v3/draw_utils.js
KuznetsovVN: This method is located in the file "draw_utils.js" and was originally intended to round off coordinates taking into account the use of zoom. For this, it is also used in the "row_splitting.js" file.

We discussed the method name with Sergey Novikov and wanted to keep this idea in the name.

@novsstation What you think?

novsstation: yep, roundToThreeDecimals is a good name

IgnatovDan (approved):

js/exporter/jspdf/v3/rows_splitting.js
IgnatovDan: My comment is outdated, now the cloning and its explanation is on the same line of code: assign(... { text: ''})

novsstation (approved):

PR #20677, opened 2021-12-24
DataGridLight: Improve plugins

vconst:
I made the following changes:
- Spited PagingPluginData interface to several plugin values.
- Added createSelector that calculate value depends on another plugin values.
- Added deps parameter to PlaceHolderExtender to avoid using state in Pager.
- Added ValueSetter and GetterExtedner components to simplify passing values to plugins.


js/renovation/ui/grids/data_grid_light/widgets/paging.tsx
```diff
@@ -82,33 +72,13 @@ export class Paging extends JSXComponent(PagingProps) {
...
   calculateVisibleItems(dataSource: RowData[]): RowData[] {
-    if (!this.props.enabled || this.pageSize === 'all') {
+    if (!this.props.enabled || this.props.pageSize === 0 || this.props.pageSize === 'all') {
       return dataSource;
     }
 
-    const start = (this.props.pageIndex as number) * (this.pageSize as number);
-    const end = start + (this.pageSize as number);
+    const start = (this.props.pageIndex as number) * (this.props.pageSize as number);
+    const end = start + (this.props.pageSize as number);
```
pomahtri: Why `this.pageSize` -> `this.props.pageSize`?

js/renovation/ui/grids/data_grid_light/widgets/paging.tsx
```diff
@@ -82,33 +72,13 @@ export class Paging extends JSXComponent(PagingProps) {
   }
 
   calculateVisibleItems(dataSource: RowData[]): RowData[] {
-    if (!this.props.enabled || this.pageSize === 'all') {
+    if (!this.props.enabled || this.props.pageSize === 0 || this.props.pageSize === 'all') {
```
pomahtri: It's not necessary to check this value for zero if using `pageSize` getter, or you should remove `get pageSize()`

I think we should keep this getter and avoid zero-checks in other parts of the code, cause it's easy to forget about them

js/renovation/utils/plugin/context.ts
```diff
@@ -36,10 +36,46 @@ export class PluginGetter<T> extends PluginEntity<T, GetterStoreValue<T>> {
...
+//   func: (a: A, b: B, c: C) => R
+// ): PluginSelector<R>;
+
+export function createSelector<R>(
+  deps: PluginEntity<unknown, unknown>[],
+  func: SelectorFunc<R>,
+): PluginSelector<R> {
+  return new PluginSelector<R>(deps, func);
+}
+
```
pomahtri, line 78: There is ts playground with variadic generics, but it doesn't infer types of func's params:

[link](https://www.typescriptlang.org/play?#code/LAKAlgdgLgpgTgMwIYGMYAIAKAbArgc0gFFowoBPAHgEEA+dAb1HRfQGcB7AWxgDUk8MAFzpqoAL6hIsRKgw4CkAMoxsMFFA5wa9BpJChQAE3XYkcDAlwQNYDhHQoLSWCrUatlACroYAD1gIIzZ0JAhyAG0AXQAadAAlWgAKZlYTAAc2EQZ0CIBpdEh0AGsYcg4EdC8okQVCCBIoMiovfKj6cRjUlisbESSAOiHzfCyqgEp0AF56eNBx2rx6t3VNbUSAbkMQNSgsJHwYAGEOayhFxQbSCkoIXC4AI3h6KcZOHn5BEQBGcS2dmB7JRQOCQfCfXDCLBLYjXKhsEFgl5vbh8ASQkQAciQbCMmL+2wA9IT0N8BugAOpgbDYRzcdLUjAAdzIAAt0KD8Ky9hR0jA2KAnDAXDAVh44CkQKxcsDQRBweiYHFMAdjqdoLFuugkgi5QqvuxEfK4ulVScziI7o94JMZoxttL9OMiSSAEzkin2TF7FD0xnoFlQdlMuD2fDoXn8wXOVyqVZaSXSiKysEQpX7Q7mjVdKWsHVG-UY9BWp5wE1m9XnYv3Uu23QO1hOl3oADMHuptOQNIeqGKEY46GsdgcYEqkZC5gwbGKYHSfKM0eFsfca0TrGTBbTyorZ01uZY+b1W-Qpszlbr9v30vQWqbBhAQA)

js/renovation/utils/plugin/placeholder_item.tsx
```diff
@@ -46,8 +51,39 @@ export class PlaceholderItemProps {
...
+
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   get currentComponent(): any {
-    return this.props.componentTypes[this.props.index];
+    if (this.componentDeps.every((entity) => this.plugins.hasValue(entity))) {
+      return this.props.componentTypes[this.props.index];
+    }
+
+    return null;
+  }
```
pomahtri, line 84: It won't render next `PlaceholderItem`s if current doesn't have one of deps. Is it by design?

js/renovation/utils/plugin/placeholder_item.tsx
```diff
@@ -46,8 +51,39 @@ export class PlaceholderItemProps {
...
+  plugins!: Plugins;
+
+  @InternalState()
+  updateCounter = 0;
+
+  @Effect()
+  updateArgs(): () => void {
+    const disposers = this.componentDeps.map((entity) => this.plugins.watch(entity, () => {
+      this.updateCounter += 1;
+    }));
```
pomahtri: Maybe it's better to use something like

```ts
@InternalState()
args: Record<number, unknown> = {}

@Effect()
updateArgs(): () => void {
  const disposers = this.componentDeps.map((entity) => this.plugins.watch(entity, (value) => {
    this.args = {...this.args, [entity.id]: value}
  }));
}

get args(): unknown[]
  return this.componentDeps.map(
    (entity) => this.args[entity.id],
  );
}
```

The problem with memorized getter can be resolved using it

js/renovation/utils/plugin/context.ts
pomahtri: another one, with infering, but with ugly params order: [link](https://www.typescriptlang.org/play?#code/LAKAlgdgLgpgTgMwIYGMYAIAKAbArgc0gFFowoBPAHgEEA+dAb1HRfQGcB7AWxgDUk8MAFzpqoAL6hIsRKgw4CkAMoxsMFFA5wa9BpJChQAE3XYkcDAlwQNYDhHQoLSWCrUatlACroYAD1gIIzZ0JAhyAG0AXQAadAAlWgAKZlYrGxEkgDoc83w2ES8ASnQAXnp4mNSWHKyTAAcCxnQIgGl0SHQAaxhyDgR0LyiRBUIIEigyKi82qPpxKpAikbwxt3VNbUSAbkMQNSgsJHwYAGEOaygVxXHSCkoIXC4AI3h6UsZOHn5BEQBGcS7fYwQ5KKBwSD4H64YRYVbEO5UNjgyHvT7cPgCGEiADkSDYRhxgL2AHoSeg-ll0AB1MDYbCObj1OkYADuZAAFugIfgOYcKPUYGxQE4YC4YOsPHAUiBWOgksiIRAoVjYYrIXF6sczhdoCJHi94CVyow9nKFtV0GClSrBHFMNrzpdFkVSeSAExU6n2HGHFBMlnodlQLmsuD2fDoAVCkXOVyqDZaGVyhUo5XQ2EG15wTWO3VXdBZo1lXRm1gW2Wsa2QjP2vPO0CugwgMnoADMVOjHQgCHgkNjYvj7k2ydYqZttfQWpOTugxtLlblLEtFbl1fTqrrM-zLr2QA)

js/renovation/utils/plugin/placeholder_item.tsx
vconst: It is necessary to avoid several renders on start because each dependency initialisation cause rerender without this check.

js/renovation/utils/plugin/context.ts
vconst: The both suggested solutions are not perfect. So, I keep current approach.

pomahtri (approved):

PR #20685, opened 2021-12-27
TS - Add ts testing for regression cases (T1052277, T1051918)

dxBeardedBear:
Fixed issues T1052277, T1051918 and added tests for checking TS regression


.github/workflows/ts_declarations.yml
```diff
@@ -74,10 +74,35 @@ jobs:
...
+      with:
+        path: ./node_modules
+        key: ${{ runner.os }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: ${{ runner.os }}-node-modules
+
+    - name: Run npm install
+      run: npm install --no-audit --no-fund
+
+    - name: Run TS tests
+      run: npm run test-typescript
```
IlyaKhD:
```suggestion
    - name: Check sources compilation
      run: npm run validate-ts

    - name: Check test cases
      run: npm run test-typescript
```
And remove `npm run validate-ts` from the `'validate-declarations'` job

.github/workflows/ts_declarations.yml
```diff
@@ -74,10 +74,35 @@ jobs:
         npm run validate-ts
         npm run validate-declarations
 
+  check-ts-regression-cases:
```
IlyaKhD:
```suggestion
  check-ts-compilation:
```

build/gulp/test_typescript.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
```
IlyaKhD: Why not use the existing 'js.ts' ?

testing/typescript/data/custom_store.ts
```diff
@@ -0,0 +1,57 @@
+import $ from 'jquery';
```
IlyaKhD, line 7: Is all this stuff really required? What exactly checks the test?

testing/typescript/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "private": true,
+  "name": "tests",
+  "version": "1.1.0",
+  "scripts": {
+    "test-ts": "node ../../node_modules/typescript/bin/tsc -p ./tsconfig.json"
```
IlyaKhD: Shouldn't is use `typescript-min` instead?

testing/typescript/data/custom_store.ts
dxBeardedBear: I agree that this is not rational - we only need ajax for data request. However, jquery is exported as a whole chunk. We can use, for example, http://projects.jga.me/jquery-builder or https://github.com/jquery/jquery#how-to-build-your-own-jquery. Do we really need to make our own special cropped custom JQuery bundle?

IlyaKhD: Sry, I was not enough specific, by 'all stuff' I mean creating a class, defining several methods, etc. Can't it be reduced to 3-5 lines?

.github/workflows/lint.yml
```diff
@@ -60,6 +60,13 @@ jobs:
     - name: Run npm install
       run: npm install --no-audit --no-fund
 
+    - name: Run npm install TS testing
+      run: |
+        cd testing/typescript
+        npm install --no-audit --no-fund
+        cd ../../
```
IlyaKhD:
```suggestion
        pushd testing/typescript
        npm install --no-audit --no-fund
        popd
```

js/core/utils/dependency_injector.d.ts
```diff
@@ -0,0 +1,4 @@
+export interface DependencyInjector {
```
IlyaKhD: Ang this - not public too?

js/core/utils/ajax.d.ts
```diff
@@ -0,0 +1,12 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface, @typescript-eslint/no-unused-vars
+export interface JQueryXHR { }
+
+interface Ajax extends DependencyInjector {
+  sendRequest(any): DxPromise<any> | JQueryXHR;
+}
+
+declare const ajax: Ajax;
+export default ajax;
```
IlyaKhD: This is now public, right?

js/core/utils/ajax.d.ts
```diff
@@ -0,0 +1,12 @@
+import { DxPromise } from './deferred';
+import { DependencyInjector } from './dependency_injector';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface, @typescript-eslint/no-unused-vars
+export interface JQueryXHR { }
```
IlyaKhD: Better avoid using the word 'JQuery', it may be confusing

build/gulp/ts.js
```diff
@@ -157,6 +158,35 @@ gulp.task('ts-check-public-modules', gulp.series('ts-copy-modules', function() {
         .pipe(compileTS({ allowSyntheticDefaultImports: true }));
 }));
 
+gulp.task('test-ts', gulp.series(
+  (callback) => {
+    exec(`cd testing/typescript && npm i --no-audit --no-fund`, (e, out, err) => {
+      console.log(out, err);
```
nightskylark: It's better to use `spawn` instead of `exec`. It allows controlling stdout so we'll get the real-time output instead of the `console.log` on the done callback. See the example of usage [here](https://github.com/DevExpress/devextreme-cli/blob/0aef326559c4168120ccacbdaddfd0e2f15e2c77/packages/devextreme-cli/src/utility/run-command.js).

nightskylark (changes requested):

build/gulp/ts.js
dxBeardedBear: done

testing/typescript/data/custom_store.ts
dxBeardedBear: Reduced by 30%

testing/typescript/data/custom_store.ts
```diff
@@ -0,0 +1,39 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies
+import $ from 'jquery';
+import { HttpClient } from '@angular/common/http';
+import CustomStore from '../../../js/data/custom_store';
+
+export function loadByFetch(url: string): CustomStore {
+  return new CustomStore({
+    load: () => fetch(url).then((response) => response.json()),
+  });
+}
```
alexserov: Since `fetch(...).then(x=>x.json())` is `Promise<any>`, testing fetch API is excess.
```suggestion
```

testing/typescript/data/custom_store.ts
```diff
@@ -0,0 +1,39 @@
...
+    load: () => fetch(url)
+      .then((response) => response.json())
+      .then((data) => ({
+        data: data.data,
+        totalCount: data.totalCount,
+        summary: data.summary,
+        groupCount: data.groupCount,
+      })),
+  });
+}
```
alexserov: This test is also seems to be useless, since the `load` accepts `() => Promise<object>` as a value.
However, we can test that type inference works fine when the `load` fields contains a lambda that returns one of the known types:
```suggestion
export async function infersTItemFromComplexLoadResult() {
  const store = new CustomStore({
    load: async () => ({
      data: ['1', '2', '3'],
      totalCount: 10,
      summary: [10],
      groupCount: 3,
    })
  });
  let value: string;
  //if TItem inferred incorrectly, the assignment below will result to compilation error.
  //However, if TItem inferred as any, it still will be broken, so we may try adding some logic to avoid it
  value = await store.byKey('stub');
}
```

testing/typescript/data/custom_store.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function loadByAjax(url: string, data: string): CustomStore {
+  return new CustomStore({
+    load: (): JQueryXHR => $.ajax({
+      url, cache: false, dataType: 'json', data,
+    }),
+  });
+}
```
alexserov: Function should be **parameterless**.  
Function name should be a sentence that describes the entity we test.
```suggestion
export function loadAcceptsAjaxResult() {
  new CustomStore({
    load: (): JQueryXHR => $.ajax({
      'url', cache: false, dataType: 'json', 'data',
    }),
  });
}
```

testing/typescript/data/custom_store.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export function loadByHttpClientAngular(http: HttpClient, url: string): CustomStore {
+  return new CustomStore({
+    load: () => http.get(url).toPromise(),
+  });
+}
```
alexserov: We can and **should** write integrational tests so that we could track if third-parties developers introduce BC's to our API. But:

1.  The syntax should be always the same (no return type, no args, informative name)
2.  We should add a unit test in addition to integrational test to avoid BC's in our code. In the case below:
    - the integrational test adds a context for behavior
    - the unit test fixes the behavior)

```suggestion
export function loadAcceptsAngularHttpClient() {
  const http: HttpClient;

  new CustomStore({
    load: () => http.get('url').toPromise(),
  });
}

export function loadAcceptsPromiseOfObject() {
  new CustomStore({
    load: () => new Promise(r => r({} as object))
  });
}
```

testing/typescript/data/custom_store.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
alexserov:
```suggestion
/* eslint-disable no-new */
/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```

alexserov (changes requested):

testing/typescript/data/custom_store.ts
IlyaKhD: one-liner?
```ts
const value: string = await store.byKey('stub');
```

IlyaKhD: Maybe it's better to use more complex type instead of built-in:

```
      data: [ { a: 123 } ],
...
  const value: { a: number } = await store.byKey('stub');
```

testing/typescript/data/custom_store.ts
IlyaKhD: Or
```ts
load: () => new Promise<object>(undefined)
```

alexserov (approved):

nightskylark (approved):

PR #20690, opened 2021-12-27
Button: button height should be equal if content is text either icon (T1015242)

ksercs:


Zedwag (dismissed):

ksercs:
I gonna close this pull request because this minor MacOs bug fix demands extensive button height changes

ksercs:
**The last problem:** flex item is horizontally centered, but it's expected that it should be on the left size if flex item width is more that flex container width. Sample: set appointment width to `20px` on this demo: https://js.devexpress.com/Demos/WidgetsGallery/Demo/Scheduler/LimitAppointmentCountPerCell/jQuery/Light/. We want to see `1 m` text, but see `mo`.

I close this pull request because this fix increases default button height a little bit. It can be non-expected for some users, so it's better to plan this changed for a major release.

PR #20770, opened 2022-01-11
Editors Label - Fix editor doesn't provide accessibility information about it's label

groshenkovamarina:


js/ui/text_box/ui.text_editor.base.js
```diff
@@ -444,8 +444,7 @@ const TextEditorBase = Editor.inherit({
         this._input().prop('spellcheck', this.option('spellcheck'));
     },
 
-
-    _cleanLabelObservable: function() {
+    _unobserveLabelContainerResize: function() {
         if(this._labelContainerElement) {
```
ksercs, line 448: RO does nothing if we call `unobserve` with not subscribed element as a parameter. So this condition is not required

ksercs (dismissed):

js/ui/form/components/field_item.js
```diff
@@ -121,7 +121,7 @@ export function renderFieldItem({
         try {
             widgetInstance = createComponentCallback($div, item.editorType, editorOptions);
             widgetInstance.setAria('describedby', helpID);
-            widgetInstance.setAria('labelledby', labelID);
+            if(labelID) widgetInstance.setAria('labelledby', labelID);
```
novsstation: If I understood correctly this line was updated, because if editor options contains a labelledby attr, then form will remove this attr?

novsstation (dismissed):

js/ui/form/components/field_item.js
IgnatovDan: why it is not this?
```
if(widgetInstance.getAria('labelledby')) {
   widgetInstance.setAria('labelledby', labelID);
}
```

Imho, the low level value should win.

IgnatovDan (dismissed):

novsstation (approved):

PR #20876, opened 2022-01-19
TagBox: add possibility to change caret position by mousedown (T1046705)

Zedwag:


alexander-kotov-dx (dismissed):

js/ui/drop_down_editor/ui.drop_down_list.js
```diff
@@ -487,12 +487,14 @@ const DropDownList = DropDownEditor.inherit({
         this._setListFocusedElementOptionChange();
     },
 
-    _renderPreventBlur: function($target) {
+    _renderPreventBlur: function($target, exclusionElements = []) {
```
ksercs: Isn't it better to implement 2 different functions: one for ddList and another for  TagBox?

Why I think it's a good idea:
- second parameter is passed from inheritor to parent component (and it's useful only for 1/4 inheritors)
- new specific case is actual only for the TagBox
- we create event for TagBox on ddList namespace now

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -5846,6 +5847,19 @@ QUnit.module('single line mode', {
 
         $inputWrapper.trigger('mousedown');
     });
+
+    QUnit.test('mousedown should not be prevented when input field clicked (T1046705)', function(assert) {
+        assert.expect(1);
```
ksercs: Why should ew use `expect`? Is there really some async code execution?

Zedwag: Nope 
Just copypasted previus test whithout check as always, sry

js/ui/drop_down_editor/ui.drop_down_list.js
Zedwag: done

ksercs (approved):

PR #20881, opened 2022-01-20
DataGrid - Incorrect results for a lookup column when searching in the search panel (T1059631)

pomahtri:
When it's only one column which is lookup, in case there are no search results, DataSource sends `[]` filter in load request, but `[]` filter means "all rows"

I added special filter value `['!']` which means "no rows" and added its handling in data controller


js/ui/grid_core/ui.grid_core.data_controller.js
```diff
@@ -252,6 +252,10 @@ export const dataControllerModule = {
 
                     storeLoadOptions.filter = this.combinedFilter(storeLoadOptions.filter);
 
+                    if(storeLoadOptions.filter?.length === 1 && storeLoadOptions.filter[0] === '!') {
+                        e.data = { data: [], totalCount: 0 };
```
NickMitrokhin: why object instead of an empty array?

js/ui/grid_core/ui.grid_core.data_source_adapter.js
```diff
@@ -449,7 +449,9 @@ export default gridCore.Controller.inherit((function() {
         },
         _handleCustomizeStoreLoadOptions(options) {
             this._handleDataLoading(options);
-            options.data = getPageDataFromCache(options, true) || options.cachedStoreData;
+            if(!(options.data?.data?.length === 0)) {
+                options.data = getPageDataFromCache(options, true) || options.cachedStoreData;
+            }
```
NickMitrokhin: getPageDataFromCache returns an array of items. Why the options.data property should be an object?

vconst: Can we simplify condition to `if(!options.data)`?

js/ui/grid_core/ui.grid_core.utils.js
```diff
@@ -234,7 +234,16 @@ export default {
...
         for(let i = 0; i < filters.length; i++) {
-            if(!filters[i]) continue;
+            if(!filters[i]) {
+                continue;
+            }
+            if(filters[i]?.length === 1 && filters[i][0] === '!') {
+                if(operation === 'and') {
+                    return ['!'];
+                } else if(operation === 'or') {
+                    continue;
```
vconst, line 244: Why is this continue necessary?

js/ui/grid_core/ui.grid_core.data_controller.js
pomahtri: In case `remoteOperations: true` datagrid raise error 'no totalCount' if I pass just empty array

js/ui/grid_core/ui.grid_core.data_source_adapter.js
pomahtri: @NickMitrokhin, it seems that `getPageDataFromCache` correctly resolves when `totalCount` is required (or is called only when it is not required)

@vconst, several tests in `dataController` failed if I use `if(!options.data)`. I didn't research the reasons

js/ui/grid_core/ui.grid_core.utils.js
pomahtri: In order to resolve this cases:

- ['and', ['!'], someFilter] -> ['!']
- ['or', ['!'], someFilter] -> someFilter

vconst (approved):

PR #20947, opened 2022-01-25
TS - fix promise resolving with multiple arguments (T1055063)

dxBeardedBear:
Allowed to use an additional argument when resolving promises (method then) for data  store methods (load, insert, update). Based on PR 'Remove JQuery types away from modules (#16638)' by @IlyaKhD


js/core/utils/deferred.d.ts
```diff
@@ -31,3 +31,14 @@ export interface PromiseType<T> { }
  */
 // eslint-disable-next-line @typescript-eslint/no-invalid-void-type
 export type DxPromise<T = void> = {} extends PromiseType<T> ? Promise<T> : PromiseType<T>;
+
+/** @namespace DevExpress.core.utils */
+export interface PromiseExtension<T> {
```
IlyaKhD: Since we no have an alias (`DxPromiseLike`) it's better make the extension a part of it:

```ts
export type DxPromiseLike<T> = DxPromise<T> & {
  then...
};
```

js/core/utils/deferred.d.ts
```diff
@@ -31,3 +31,14 @@ export interface PromiseType<T> { }
...
+    then<TResult1 = T, TResult2 = never>(
+        onFulfilled?: ((value: T, extraParameters?: any) => TResult1 | PromiseLike<TResult1>) | undefined | null,
+        onRejected?: ((reason: any) => TResult2 | PromiseLike<TResult2>) | undefined | null
+    ): PromiseLike<TResult1 | TResult2>;
+}
+
+/** @namespace DevExpress.core.utils */
+export type DxPromiseLike<T> = DxPromise<T> & PromiseExtension<T>;
```
IlyaKhD:
```suggestion
export type DxExtendedPromise<T> = DxPromise<T> & PromiseExtension<T>;
```

`-Like` means joining multiple types which differences are not important from processing point-of-view. E.g. `falsy` values can be named `FalseLike` (empty array, `undefined`, etc).

Here we have a type, describing values that have extra overload for the `then` method. This is important, as it changes the way the values processed.

testing/typescript/data/custom_store.ts
```diff
@@ -46,3 +49,22 @@ export function loadAcceptsPromiseOfObject() {
...
+      resolve((v1, v2: { totalCount: number; groupCount: number }) => {});
+    }),
+  });
+  store.insert(value).then((data, key: number) => ({ data, key }));
+  store.update(1, value).then((data, key: number) => ({ data, key }));
+
+  const dataSource = new DataSource(store);
+  dataSource.load().then(
+    (data, extra: { totalCount: number; groupCount: number }) => ({ data, extra }),
+  );
```
IlyaKhD: A test should be as short as possible.

I think it can be reduced to this:

```ts
/* eslint-disable @typescript-eslint/no-unnecessary-type-arguments */
/* eslint-disable @typescript-eslint/init-declarations */
/* eslint-disable @typescript-eslint/no-explicit-any */
...

const ANY = undefined as any;
...

export function promiseResolveAcceptsMultipleArguments() {
  const callback: (a: any, b: any) => void = ANY;

  const store: CustomStore<any, any> = ANY;
  store.load().then(callback);
  store.load(ANY).then(callback);
  store.insert(ANY).then(callback);
  store.update(ANY, ANY).then(callback);
}

```

IlyaKhD: It seems that same compile cases should be checked for stores of other kinds. DataSource's cases should be moved to a separate file as well (or rename `custom_store.ts` to something more generic).

js/core/utils/deferred.d.ts
dxBeardedBear: done

js/core/utils/deferred.d.ts
dxBeardedBear: done

testing/typescript/data/custom_store.ts
dxBeardedBear: done

testing/typescript/data/stores.ts
```diff
@@ -1,12 +1,12 @@
...
-/* eslint-disable @typescript-eslint/no-unused-vars */
-/* eslint-disable @typescript-eslint/no-unused-vars-experimental */
-/* eslint-disable @typescript-eslint/explicit-function-return-type */
-/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-unsafe-return */
 // eslint-disable-next-line import/no-extraneous-dependencies
 import $ from 'jquery';
 import { HttpClient, HttpHandler } from '@angular/common/http';
 import CustomStore from '../../../js/data/custom_store';
+import Store from '../../../js/data/abstract_store';
```
IlyaKhD:
```suggestion
import { Store } from '../../../js/data';
```

This required updating to the 22_1 branch, I've just merged #20807

testing/typescript/data/stores.ts
```diff
@@ -46,3 +46,13 @@ export function loadAcceptsPromiseOfObject() {
     load: () => new Promise<object>(undefined),
   });
 }
+
+export function promiseThenAcceptsMultipleArgumentsForCustomStoreMethods() {
```
IlyaKhD: Not only CustomStore, right?

testing/typescript/data/data_source.ts
```diff
@@ -0,0 +1,8 @@
+import DataSource from '../../../js/data/data_source';
+
+export function promiseResolveAcceptsMultipleArguments() {
+  const callback: (a: any, b: any) => void = undefined;
+  const dataSource = new DataSource({});
```
IlyaKhD:
```suggestion
  const callback: (a: any, b: any) => void = ANY;
  const dataSource: DataSource<any, any> = ANY;
```
We have to share this `ANY` between files

IlyaKhD (approved):

alexserov (approved):

PR #21067, opened 2022-02-08
dxScheduler - Fix broken layout after triggerResizeEvent when scheduler was invisible(T1063540)

DendyGrobovshik:


js/ui/scheduler/workspaces/ui.scheduler.work_space.js
```diff
@@ -585,15 +585,17 @@ class SchedulerWorkSpace extends WidgetObserver {
     }
 
     _dimensionChanged() {
-        if(this.option('crossScrollingEnabled')) {
-            this._setTableSizes();
-        }
+        if(this._isVisible()) {
```
AntonSermyazhko: I suggest replace built-in condition with return from the method. But it's up to you. )
```
if(!this._isVisible()) return;
```

AntonSermyazhko (dismissed):

andrewmakarov (approved):

PR #21140, opened 2022-02-16
DataGridLight - improvements

pomahtri:
Better review this PR commit-by-commit

Changes:

### moved class names constants to `classes.ts`.

Maybe in future we will separate this file to `<module>/classes.ts`

### added `getKey` function

It will throw correct errors if `keyExpr` doesn't exist. In future it will be useful to get complex keys

This function needs to be able to differ when keyExpr wasn't yet initialized (cause plugins set keyExpr in effect), so I added `null` type to keyExpr

- `null` means that user didn't specified keyExpr in config and widget should throw error
- `undefined` means that key wasn't initialized yet

### used row key as react key

removed `// eslint-disable-next-line react/no-array-index-key`

I created function `getReactRowKey`, it combines row's type and key to create react key (because master-detail row and data row have same key)

### fixed using `getValue()` in pager

changed to `.watch()`

### DIFF
--- a/js/renovation/ui/grids/data_grid_light/types.ts
+++ b/js/renovation/ui/grids/data_grid_light/types.ts
@@ -4,9 +4,9 @@
 import { JSXTemplate } from '@devextreme-generator/declarations';
 
 export type RowData = Record<string, unknown>;
-export type ColumnUserConfig = string;
-export type KeyExprUserConfig = string;
-export type KeyExpr = KeyExprUserConfig | null;
+export type Column = string;
+export type KeyExpr = string;
+export type KeyExprInternal = KeyExpr | null;
 export type Key = unknown;
 export interface Row {
   key?: Key;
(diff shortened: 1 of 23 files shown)


js/renovation/ui/grids/data_grid_light/types.ts
```diff
@@ -5,7 +5,8 @@ import { JSXTemplate } from '@devextreme-generator/declarations';
 
 export type RowData = Record<string, unknown>;
 export type ColumnUserConfig = string;
-export type KeyExpr = string;
+export type KeyExprUserConfig = string;
+export type KeyExpr = KeyExprUserConfig | null;
```
vconst: I think that it will be better to have KeyExpr and KeyExprInternal types because KeyExpr type name is better than KeyExprUserConfig for type that customers can use.

vconst (approved):

PR #21147, opened 2022-02-17
Add retry functionality to qunit tests

alexserov:
This PR attempts to fix blinking tests by adding retry functionality to a runner. 

When we execute tests, runner writes passed fixtures to the `testing/CompletedSuites.txt`
If any of the tests fail, the runner starts one more time (up to **three** times), ignoring already passed fixtures, so it runs only failed fixtures.
In the worst case, this may slow down task execution time in case there's a real failing test after a dev commit, but it may increase the time for about 3-6 minutes (on my expectations)
In the best case, retries will hide blinking tests and allow developers to stop focusing on issues that do not relate directly to their work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ project.lock.json
 yarn.lock
 RawLog.txt
 .DS_Store
+testing/CompletedSuites.txt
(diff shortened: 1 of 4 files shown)


nightskylark:
@alexserov please leave a description for the PR. Why do we need these changes? What exactly did you change?

alexserov:
@nightskylark done

dxBeardedBear (approved):

PR #21168, opened 2022-02-18
Plugins: Rework GetterExtender to use selector

vconst:
The GetterExtender component reworked for the following reasons: 
1) To fix flickering items during paging in angular and double creating visible items in react.
2) To avoid recreating methods on dependencies change in angular (It was the second solutions for the first problem).
3) To separate logic from platform specific code. Defining selectors outside component helps to avoid bugs in platform specific generated code.


js/renovation/ui/grids/data_grid_light/master_detail/master_detail.tsx
```diff
@@ -26,9 +26,33 @@ import { ExpandColumn } from './expand_column';
 import { SetExpanded, IsExpanded, MasterDetailTemplate } from './plugins';
 import { MasterDetailRow } from './master_detail_row';
 
-export const viewFunction = (viewModel: MasterDetail): JSX.Element => (
+export const AddMasterDetailRows = createSelector(
```
pomahtri, line 30: Let's move it to `plugins.ts`

pomahtri (dismissed):

js/renovation/ui/grids/data_grid_light/master_detail/master_detail.tsx
vconst: This selector uses imported MasterDetailRow widget. Now, plugins is lightweight file that can be imported by another modules. So, I will think about location for such selectors and make another PR for moving this function to another place but not to plugins.ts.

pomahtri (approved):

PR #21208, opened 2022-02-28
DataGrid - Lookup - Display in a filter row/header filter only existing items in a column

pomahtri:
[jquery.html](https://gist.github.com/pomahtri/ceff2c735d1343511680e700473d5d15)

### DIFF
--- a/js/ui/grid_core/ui.grid_core.filter_row.js
+++ b/js/ui/grid_core/ui.grid_core.filter_row.js
@@ -666,7 +666,7 @@ const ColumnHeadersViewFilterRowExtender = (function() {
         _handleDataChanged: function(e) {
             this.callBase.apply(this, arguments);
 
-            if(e.operationTypes?.filtering || e.operationTypes?.reload) {
+            if(e.operationTypes?.filtering) {
                 this.updateLookupDataSource();
             }
         },
@@ -872,6 +872,27 @@ export const filterRowModule = {
                     }
 
                     that.callBase.apply(that, arguments);
+                }
+            },
+            editing: {
+                updateFieldValue(options) {
+                    if(options.column.lookup) {
+                        this._needUpdateLookupDataSource = true;
+                    }
+
+                    return this.callBase.apply(this, arguments);
+                },
+                _afterSaveEditData(cancel) {
+                    if(this._needUpdateLookupDataSource && !cancel) {
+                        this.getView('columnHeadersView')?.updateLookupDataSource();
+                    }
+                    this._needUpdateLookupDataSource = false;
+
+                    return this.callBase.apply(this, arguments);
+                },
+                _afterCancelEditData() {
+                    this._needUpdateLookupDataSource = false;
+                    return this.callBase.apply(this, arguments);
                 }
             }
         },
--- a/js/ui/grid_core/ui.grid_core.filter_row.js
+++ b/js/ui/grid_core/ui.grid_core.filter_row.js
@@ -666,7 +666,7 @@ const ColumnHeadersViewFilterRowExtender = (function() {
         _handleDataChanged: function(e) {
             this.callBase.apply(this, arguments);
 
-            if(e.operationTypes?.filtering) {
+            if(e.operationTypes?.filtering || e.operationTypes?.reload) {
                 this.updateLookupDataSource();
             }
         },
--- a/js/ui/grid_core/ui.grid_core.filter_row.js
+++ b/js/ui/grid_core/ui.grid_core.filter_row.js
@@ -666,7 +666,7 @@ const ColumnHeadersViewFilterRowExtender = (function() {
         _handleDataChanged: function(e) {
             this.callBase.apply(this, arguments);
 
-            if(e.operationTypes.filtering) {
+            if(e.operationTypes?.filtering) {
                 this.updateLookupDataSource();
             }
         },
--- a/js/ui/grid_core/ui.grid_core.filter_row.js
+++ b/js/ui/grid_core/ui.grid_core.filter_row.js
@@ -944,6 +944,9 @@ export const filterRowModule = {
                     if(args.name === 'filterRow') {
                         this._invalidate();
                         args.handled = true;
+                    } else if(args.name === 'syncLookupFilterValues') {
+                        this.component.getView('columnHeadersView')?.updateLookupDataSource();
+                        args.handled = true;
                     } else {
                         this.callBase(args);
                     }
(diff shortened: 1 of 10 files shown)


js/ui/grid_core/ui.grid_core.filter_row.js
```diff
@@ -670,17 +720,23 @@ const DataControllerFilterRowExtender = {
...
             }
         });
 
         return gridCoreUtils.combineFilters(filters);
-    }
+    },
+
+    _applyFilter: function() {
+        this.callBase.apply(this, arguments);
+        this.component._views['columnHeadersView']?.updateLookupDataSource();
```
vconst: Use public method getView instead of _views field.

js/ui/grid_core/ui.grid_core.filter_row.js
```diff
@@ -670,17 +720,23 @@ const DataControllerFilterRowExtender = {
 
         const filters = [this.callBase()];
         const columns = this._columnsController.getVisibleColumns(null, true);
+        const filterRowController = this.getController('applyFilter');
 
         each(columns, function() {
-
-            if(this.allowFiltering && this.calculateFilterExpression && isDefined(this.filterValue)) {
+            const shouldSkip = filterRowController._currentColumn?.index === this.index;
```
vconst: It is strange that _currentColumn is assigned to applyFilter controller that does not use this value.
Add public methods to applyFilter controller to get and set _currentColumn and use this public methods. Or _currentFilterColumn can be defined in dataController and can be the same for filterRow and headerFilter. We can add helper method getCombinedFilterWithoutColumn(column) to dataController to avoid code duplication for setting and clearing _currentColumn in several places.

js/ui/grid_core/ui.grid_core.filter_row.js
```diff
@@ -476,7 +477,25 @@ const ColumnHeadersViewFilterRowExtender = (function() {
 
         _renderEditor: function($editorContainer, options) {
             $editorContainer.empty();
-            return this.getController('editorFactory').createEditor($('<div>').appendTo($editorContainer), options);
+            const $element = $('<div>').appendTo($editorContainer);
+            const editorController = this.getController('editorFactory');
+            const dataSource = this.getController('data').dataSource();
+
+            if(options.lookup && this.option('filterRow.showRelevantValues')) {
+                const filter = this.getController('data').getCombinedFilter();
```
vconst, line 487: Why full combined filter is calculated here?

js/ui/grid_core/ui.grid_core.utils.js
```diff
@@ -544,5 +547,53 @@ export default {
...
+
+    getWrappedLookupDataSource(column, dataSource, filter) {
+        const lookupDataSourceOptions = this.normalizeLookupDataSource(column.lookup);
+
+        const lookupDataSource = {
+            load: (loadOptions) => {
+                const d = new Deferred();
+                dataSource.load({
+                    ...loadOptions,
+                    filter: this.combineFilters([filter, loadOptions.filter], 'and'),
```
vconst: Add lookupDataSourceOptions.filter here

js/ui/grid_core/ui.grid_core.utils.js
```diff
@@ -544,5 +547,53 @@ export default {
...
+                    const filter = this.combineFilters(
+                        items.map((data => [
+                            column.lookup.valueExpr, data.key,
+                        ])),
+                        'or'
+                    );
+
+                    (new DataSource({
+                        ...lookupDataSourceOptions,
+                        filter: this.combineFilters([filter, lookupDataSourceOptions.filter], 'and'),
```
vconst: Move filters combination to single place.

js/ui/grid_core/ui.grid_core.utils.js
```diff
@@ -544,5 +547,53 @@ export default {
...
+                }).done((items) => {
+                    const filter = this.combineFilters(
+                        items.map((data => [
+                            column.lookup.valueExpr, data.key,
+                        ])),
+                        'or'
+                    );
+
+                    (new DataSource({
+                        ...lookupDataSourceOptions,
```
vconst: It is possible to use column.lookup.items here if it is defined to avoid additional request:
`...(column.lookup.items ? {.store: column.lookup.items } : lookupDataSourceOptions)`
If lookup optimisation is not applied (https://js.devexpress.com/Documentation/Guide/UI_Components/DataGrid/Enhance_Performance_on_Large_Datasets/#Lookup_Optimization) DataGrid loads lookup on first render and loaded lookup items are assigned to lookup.items.

js/ui/grid_core/ui.grid_core.header_filter.js
```diff
@@ -511,6 +510,7 @@ export const headerFilterModule = {
                 height: 325,
                 allowSearch: false,
                 searchTimeout: 500,
+                showRelevantValues: true,
```
vconst: We should think about option names. The showRelevantValues option name does not say that it is applied to lookup columns only. In the PivotGrid this option is applied to all fields.

js/ui/grid_core/ui.grid_core.utils.js
vconst: If filter is big server can return get request length limit error. For lookup column with optimisation when calculateDisplayValue field is defined we can make filtering by this display column to avoid additional request with big filter. We should discuss this. Or showRelevantValues should be disabled by default. It will be bad if after upgrading some customers see request length limit error.

js/ui/grid_core/ui.grid_core.utils.js
```diff
@@ -544,5 +547,53 @@ export default {
...
+    getWrappedLookupDataSource(column, dataSource, filter) {
+        const lookupDataSourceOptions = this.normalizeLookupDataSource(column.lookup);
+
+        const lookupDataSource = {
+            load: (loadOptions) => {
+                const d = new Deferred();
+                dataSource.load({
+                    ...loadOptions,
+                    filter: this.combineFilters([filter, loadOptions.filter], 'and'),
+                    group: column.dataField || column.name
```
vconst: Use isExpanded false parameter to avoid all data items loading: https://js.devexpress.com/Documentation/ApiReference/Data_Layer/CustomStore/LoadOptions/#group.

NickMitrokhin (dismissed):

NickMitrokhin (dismissed):

Alyar666 (approved):

PR #21234, opened 2022-03-02
Property type extractor

Igggr:
Write extractor for property type and compile-check some examples of extracted properties

UPD: We have to set TestCafe's `customCompilerModulePath` in order to use TypeScript 4.2


testing/helpers/type_helpers/type_helpers.ts
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable @typescript-eslint/no-type-alias */
+export type PropertyType<T, TProp extends PropertyKey> =
+  T extends (infer T2)[]
+    ? PropertyType<T2, TProp>
+    : T extends Partial<Record<TProp, infer TValue>>
+      ? TValue
+      : never;
+
+export type DeepPropertyType<T, TProps> =
```
IlyaKhD: Consider this renaming:
```suggestion
export type RecursivePropertyType<T, TProps> =
```

IlyaKhD (commented):
1) Compilation cases should be moved to testing/typescrit dir
2) `PropertyType` should be added to the main package. e.g. core/index.d.ts
2) There should be a single type: `PropertyType` or `DeepPropertyType`
3) Currently another type equality check is used, we should not add one more. Check `assertType` in the testing/typescript/consts.ts module

testing/helpers/type_helpers/type_helpers.ts
Igggr: Done.

Igggr: ~~Also changed tuples to string templates. Think `'a.b.c'` is prettier than ['a', 'b', 'c']~~ No. It require ts 4.1

js/core/index.d.ts
```diff
@@ -25,3 +25,29 @@ export type DeepPartial<T> = T extends object ? {
...
+type Property<T, P> =
+T extends (infer El)[]
+  ? P extends PropertyKey
+    ? El extends Partial<Record<P, infer TValue>> ? TValue : never
+    : never
+  : P extends PropertyKey
+    ? T extends Partial<Record<P, infer TValue>>
+      ? TValue
+      : never
+    : never;
```
IlyaKhD:
```suggestion
type Property<T, P extends PropertyKey> = P extends keyof T ? T[P] : never;

type PropertyType<T, P> =
  P extends PropertyKey
  ? T extends (infer TItem)[]
    ? Property<TItem, P>
    : Property<T, P>
  : never;
```

testing/typescript/compile-check-property-extractor.ts
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
IlyaKhD: All files within the 'testing/typescript' folder are compilation checks, you can name this one simply 'property-extractor.ts'
But it should be moved to the 'core' folder

Igggr: In that case - maybe `RecursivePropertyType` itself should be in that file, rather than in index.d.ts?

js/core/index.d.ts
```diff
@@ -25,3 +25,16 @@ export type DeepPartial<T> = T extends object ? {
 
 // Omit does not exist in TS < 3.5.1
 export type Skip<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
+export type RecursivePropertyType<T, TProps extends string> =
```
IlyaKhD: Consider splitting this into simpler types:

```ts
type ItemType<T> = T extends (infer TItem)[] ? TItem : T;
type Property<T, TPropName extends string> = T extends Partial<Record<TPropName, infer TValue>> ? TValue : never;
type OwnPropertyType<T, TPropName extends string> = Property<ItemType<T>, TPropName>;

export type PropertyType<T, TProp extends string> =
  TProp extends `${infer TOwnProp}.${infer TNestedProps}`
    ? PropertyType<OwnPropertyType<T, TOwnProp>, TNestedProps>
    : OwnPropertyType<T, TProp>;
```

IlyaKhD (dismissed):

js/core/index.d.ts
Igggr: Done. But I think you doesn't want helper types:

> There should be a single type: PropertyType or DeepPropertyType
https://github.com/DevExpress/DevExtreme/pull/21234#pullrequestreview-899225071

IlyaKhD: It depends on situation, our aim is to make code more readable.

testing/typescript/compile-check-property-extractor.ts
IlyaKhD: >  maybe RecursivePropertyType itself should be in that file

What file do you mean?

Igggr: > What file do you mean?

property_extractor.ts

IlyaKhD: Public `PropertyType` should be located in '/core/index.d.ts' - OK
Compilation cases should be moved to '/testing/typescript/core.ts'

IlyaKhD (approved):

PR #21278, opened 2022-03-22
Remove tags that can be autogenerated

Igggr:
1) some tags has wrong field number - changed
2) remove tags that can be auto generated

Depend upon commit in devextreme-internal-tools


IlyaKhD (dismissed):

IlyaKhD:
I've created another one: #21657

PR #21295, opened 2022-03-24
DataGrid - Columns are overlapped when the grouping feature and fixed columns are used if the type for one column is set to 'groupExpand' (T1075560)

pomahtri:
If column with `fixedPosition: 'right'` was grouped, expand column also had `fixedPosition: 'right'`. Because of it in [_getFixedColumnsCore](https://github.com/pomahtri/devextreme/blob/DataGrid_fixed_position_with_grouping/21_1/js/ui/grid_core/ui.grid_core.columns_controller.js#L1321) beginning of the transparent column was wrong.

I added setting `fixedPosition: left` to expand columns


NickMitrokhin (changes requested):
I think we should add a test for enabled RTL.

NickMitrokhin (approved):

PR #21318, opened 2022-03-25
DataGrid: Fix keyboard navigation after adding a new row (T1069848)

Alyar666:
See https://devexpress.com/issue=T1069848


NickMitrokhin:
It would be great to add tests for 'cell' and 'row' editing modes.

NickMitrokhin (changes requested):
It would be great to add tests for 'cell' and 'row' editing modes.

NickMitrokhin (approved):

PR #21578, opened 2022-04-25
pdf export: make PdfCell type public

EugeniyKiyashko:


novsstation (dismissed):

js/pdf_exporter.d.ts
```diff
@@ -3,20 +3,25 @@ import dxDataGrid from './ui/data_grid';
...
+  */
+export interface PdfDataGridCell extends ExcelCell {}
 
 /**
+ * @public
  * @docid
  * @namespace DevExpress.pdfExporter
- * @type object
  */
 export interface PdfCell {
```
IlyaKhD: If we make it public, we should check its name. It seems that we should rename it to just `Cell`.

EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/21578/commits/e1d152c47234a271f8ee7e1c34da76d9f2ab82f1

IlyaKhD (approved):

PR #21616, opened 2022-04-28
RangeSlider: add previousValue field to onValueChanged parameter

Zedwag:


js/ui/range_slider.js
```diff
@@ -318,7 +318,8 @@ const RangeSlider = Slider.inherit({
                     start: start,
                     end: end,
                     value: [start, end],
-                    event: this._valueChangeEventInstance
+                    event: this._valueChangeEventInstance,
+                    previousValue: args.previousValue
```
DokaRus, line 322: Could you test it?

Zedwag: modified one of old tests

DokaRus (approved):

PR #21705, opened 2022-05-16
22 1 localization

GoodDayForSurf:


js/localization/globalize/core.js
```diff
@@ -1,26 +1,12 @@
 // eslint-disable-next-line no-restricted-imports
 import Globalize from 'globalize';
 import coreLocalization from '../core';
+import { likelySubtags } from '../cldr-data/likely_subtags';
```
nightskylark, line 4: I did not find this file in the localization sources generation task. Are you sure it will exist on other machines?

PR #21742, opened 2022-05-19
Add stricter default for collections TItem generic param

IlyaKhD:


IlyaKhD:
Will merge this to v22.2

PR #21774, opened 2022-05-26
T1064953 Modify type of InputEvent for Inputs Components

GoodDayForSurf:
Specify EventObject TS types for certain events

### DIFF
--- a/js/events/index.d.ts
+++ b/js/events/index.d.ts
@@ -34,6 +34,11 @@ export interface NativeEventInfo<TComponent, TNativeEvent = Event> {
 }
 
 /** @public */
+export interface NativeInputEventInfo<TComponent, TNativeEvent = Event> extends Omit<NativeEventInfo<TComponent, TNativeEvent>, 'event'> {
+    readonly event?: DxEvent<Omit<TNativeEvent, 'target'> & { target: HTMLInputElement }> ;
+}
+
+/** @public */
 export interface ChangedOptionInfo {
     readonly name: string;
     readonly fullName: string;
--- a/js/ui/autocomplete.d.ts
+++ b/js/ui/autocomplete.d.ts
@@ -1,9 +1,9 @@
 import {
-    EventInfo,
-    NativeEventInfo,
-    InitializedEventInfo,
-    ChangedOptionInfo,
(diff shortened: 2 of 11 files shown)


js/events/index.d.ts
```diff
@@ -33,6 +33,11 @@ export interface NativeEventInfo<TComponent, TNativeEvent = Event> {
     readonly event?: DxEvent<TNativeEvent>;
 }
 
+/** @public */
+export interface NativeInputEventInfo<TComponent, TNativeEvent = Event> extends Omit<NativeEventInfo<TComponent, TNativeEvent>, 'event'> {
```
IlyaKhD, line 37: I would make it more verbose for better readability:
1) Extract a helper type for replacing `target` prop
2) Do not use `omit` for `NativeEventInfo`

I believe that keeping plain fields structure in `EventInfo`, `NativeEventInfo` and this new `NativeInputEventInfo` makes these types easier to read.

IlyaKhD: Smth like this:

```ts
/** @public */
export interface NativeInputEventInfo<TComponent, TNativeEvent extends Event = Event> {
    readonly component: TComponent;
    readonly element: DxElement;
    readonly model?: any;
    readonly event?: DxEvent<SpecifyTarget<TNativeEvent, HTMLInputElement>>;
}

type SpecifyTarget<TNativeEvent extends Event, TEventTarget extends EventTarget> = {
    target: TEventTarget;
} & Omit<TNativeEvent, 'target'>;
```

js/ui/autocomplete.d.ts
```diff
@@ -1,9 +1,9 @@
...
-    EventInfo,
-    NativeEventInfo,
-    InitializedEventInfo,
-    ChangedOptionInfo,
-    ItemInfo,
+  EventInfo,
+  NativeEventInfo,
+  InitializedEventInfo,
+  ChangedOptionInfo,
+  ItemInfo, NativeInputEventInfo,
```
IlyaKhD, line 6:
```suggestion
  ItemInfo,
  NativeInputEventInfo,
```
To keep consistency. Same for other cases

PR #21806, opened 2022-06-06
NumberBox: The caret jumps to the beginning of the input after the minus sign is typed if the format is '#' (T1093688)

EugeniyKiyashko:


js/localization/number.js
```diff
@@ -257,6 +257,10 @@ const numberLocalization = dependencyInjector({
             return 1;
         }
 
+        if(text === '-') {
```
Zedwag: can we fix this in numberBox files? It doesn't seem good to change core squad files t fix such a little bug

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
```diff
@@ -2093,15 +2093,19 @@ QUnit.module('stubs', moduleConfig, function() {
...
+    ].forEach(({ format, expectedText, typedText, expectedValue }) => {
         QUnit.test(`widget should correctly apply format="${format}", value="${expectedValue}"`, function(assert) {
             this.instance.option({
                 value: null,
                 format
             });
 
             this.keyboard
-                .type(expectedValue.toString())
+                .type(typedText || expectedValue.toString())
```
ksercs, line 2119:
```suggestion
                .type(typedText ?? expectedValue.toString())
```

js/localization/number.js
ksercs: Hm, it works only for simple formats, it will not work for `#;-#` for example, or `0;<0`. Moreover it's not the best practice to add such corner case. Maybe it's better to update etalon calculation? Seems like we can just modify 247 row this way:
`negativeEtalon = negativeEtalon.replaceAll('1', '.*');`

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
EugeniyKiyashko: ?? doesn't supported in our test environment

js/localization/number.js
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/21806/commits/5501331453ded0b91d021e49fa43b142b22afce6

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
```diff
@@ -2093,15 +2093,30 @@ QUnit.module('stubs', moduleConfig, function() {
...
+        { format: '\'--\'#', expectedText: '---1', typedText: '-1', expectedValue: -1 },
+        { format: '#;-#', expectedText: '-1', typedText: '1-', expectedValue: -1 },
+        { format: '#;-#', expectedText: '1', typedText: '1--', expectedValue: 1 },
+        { format: '0;<0', expectedText: '<1', typedText: '-1', expectedValue: -1 },
+        { format: '0;<0', expectedText: '<1', typedText: '1-', expectedValue: -1 },
+        { format: '0;<0', expectedText: '1', typedText: '1--', expectedValue: 1 },
+        { format: '\'plus\' 0;\'minus\' 0', expectedText: 'minus 1', typedText: '1-', expectedValue: -1 },
+        { format: '\'plus\' 0;\'minus\' 0', expectedText: 'plus 1', typedText: '1--', expectedValue: 1 },
+    ].forEach(({ format, expectedText, typedText, expectedValue }) => {
+        QUnit.test(`widget should correctly apply format="${format}", value="${typedText || expectedValue}"`, function(assert) {
```
ksercs, line 2112: nice work!

ksercs (approved):

Zedwag (approved):

js/localization/number.js
EugeniyKiyashko: Before fix the single char '-' doesn't determined as possible negative value, but before regression it was.

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
EugeniyKiyashko: you are welcome!

PR #21911, opened 2022-06-16
CI - Prevent generate angular

AntonSermyazhko:


build/gulp/ts.js
```diff
@@ -59,12 +59,6 @@ gulp.task('ts-copy-bundle', gulp.series(
             .pipe(replace(/(interface JQuery\b[\s\S]*?{)[\s\S]+?(})/gm, '$1$2'))
             .pipe(gulp.dest(packageBundlesPath));
     },
-
-    function writeAngularHack() {
```
MikeVitik, line 63: Looks like we shouldnt remove this function. It's related to the devextere-angular not native angular

package.json
```diff
@@ -212,16 +211,11 @@
     "build:react": "gulp generate-react",
     "build:react:watch": "gulp generate-react-watch",
     "build:react:typescript": "gulp generate-react-typescript",
-    "build:angular": "gulp generate-angular",
-    "build:angular:watch": "gulp generate-angular-watch",
-    "build:angular:typescript": "gulp generate-angular-typescript",
     "build:vue": "gulp generate-vue",
```
MikeVitik, line 214: please remove vue task as angular because you have already remove gulp task

PR #21937, opened 2022-06-19
Editors: validation message position is not correct if a parent container was hidden (T1095900)

EugeniyKiyashko:


testing/testcafe/tests/editors/validationMessage.ts
```diff
@@ -20,3 +27,55 @@ test('ValidationMessage integrated in editor should not raise any errors when it
...
+    .ok()
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+
+  await ClientFunction(() => {
+    (document.querySelector('#container') as HTMLElement).setAttribute('hidden', 'true');
+  })();
+
+  await ClientFunction(() => {
+    document.querySelector('#container')?.removeAttribute('hidden');
```
ksercs, line 53: cannot we do this in the previous ClientFunction?

testing/testcafe/tests/editors/validationMessage.ts
```diff
@@ -20,3 +27,55 @@ test('ValidationMessage integrated in editor should not raise any errors when it
     .click(checkBox1.element)
     .expect(true).ok();
 });
+
+fixture`ValidationMessage_TextBox`
+  .page(url(__dirname, '../container.html'));
+
+test('Validation Message position should be correct after change visibility of parent container', async (t) => {
```
ksercs:
```suggestion
test('Validation Message position should be correct after change visibility of parent container (T1095900)', async (t) => {
```

testing/testcafe/tests/editors/validationMessage.ts
```diff
@@ -20,3 +27,55 @@ test('ValidationMessage integrated in editor should not raise any errors when it
...
+    document.querySelector('#container')?.removeAttribute('hidden');
+  })();
+
+  await t
+    .expect(await takeScreenshot('Textbox_validation_message.png'))
+    .ok()
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+}).before(async (t) => {
+  await t.resizeWindow(300, 200);
```
ksercs, line 62: why do we need this window resize?

testing/testcafe/tests/editors/validationMessage.ts
```diff
@@ -1,9 +1,16 @@
...
+import { restoreBrowserSize } from '../../helpers/restoreBrowserSize';
+import createWidget from '../../helpers/createWidget';
+import { appendElementTo } from '../navigation/helpers/domUtils';
+
+const TEXTEDITOR_INPUT_CLASS = 'dx-texteditor-input';
 
 fixture`ValidationMessage`
   .page(url(__dirname, './pages/T941581.html'))
-  .beforeEach(async (t) => { await t.wait(10000); });
+  .beforeEach(async (t) => { await t.wait(1000); });
```
ksercs: why is it necessary?

js/ui/overlay/ui.overlay.js
```diff
@@ -334,7 +334,7 @@ const Overlay = Widget.inherit({
     },
 
     _contentResizeHandler: function(entry) {
-        if(!this._shouldSkipContentResize(entry)) {
+        if(!this._shouldSkipContentResize(entry) && this._$content.is(':visible')) {
```
ksercs, line 337: Isn't it better to use our api to check if popup is opened?

testing/testcafe/tests/editors/validationMessage.ts
EugeniyKiyashko: I do setAttribute and removeAttribute logic in various clientFunction willfully. It helps me separate important parts of scenario logic separate each other and debug it.

testing/testcafe/tests/editors/validationMessage.ts
EugeniyKiyashko: I need to do screenshot which contains the Texteditor and the ValidationMessage simultaneously. There is no container element that include both of them. So, I get minimal screenshot dimensions for my purpose through resize windows.

js/ui/overlay/ui.overlay.js
EugeniyKiyashko: No, the parent container is hiding. So, visibilityChange for nested component doesn't call and ValidationMessage stay visible. It's important check visibility this way.

ksercs (approved):

PR #22014, opened 2022-06-27
SelectBox: data is loaded infinitely if the visible option of items in subsequent pages are set to false (T1092746)

EugeniyKiyashko:


js/ui/list/ui.list.base.js
```diff
@@ -543,11 +540,18 @@ export const ListBase = CollectionWidget.inherit({
         }
     },
 
+    _shouldContinueLoading: function(shouldLoadNextPage) {
+        const isReachBottom = getHeight(this._scrollView.content()) - getHeight(this._scrollView.container()) < this._scrollView.scrollOffset().top;
```
ksercs:
```suggestion
        const isBottomReached = getHeight(this._scrollView.content()) - getHeight(this._scrollView.container()) >= this._scrollView.scrollOffset().top;
```

and `return shouldLoadNextPage && (!this._scrollViewIsFull() || !isBottomReached);`

EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/22014/commits/7ba70d78face4fd9da0889c9d58d62caf79392d9

ksercs (approved):

PR #22019, opened 2022-06-28
Collection, List, DDList, SelectBox, TagBox: replace all ds requests with DataController requests

Zedwag:


js/core/data_controller.js
```diff
@@ -0,0 +1,173 @@
...
+    loadOptions() {
+        return this._dataSource?.loadOptions();
+    }
+
+    userData() {
+        return this._dataSource?._userData;
+    }
+
+    cancel(operationId) {
+        this._dataSource?.cancel(operationId);
```
ksercs: let's use [null object pattern](https://sourcemaking.com/design_patterns/null_object) to get rid of all this optional chainings and conditions. You can find an example here: https://github.com/DevExpress/DevExtreme/blob/22_2/js/core/resize_observer.js

Zedwag: done

js/ui/collection/ui.collection_widget.edit.js
```diff
@@ -215,37 +214,34 @@ const CollectionWidget = BaseCollectionWidget.inherit({
             filter: that._getCombinedFilter.bind(that),
             totalCount: function() {
                 const items = that.option('items');
-                const dataSource = that._dataSource;
-                return dataSource && dataSource.totalCount() >= 0
-                    ? dataSource.totalCount()
+                const dataController = that._dataController;
+                return dataController.totalCount() >= 0
```
ksercs: let's remember the result of `dataController.totalCount()` to not call function two times

js/ui/editor/ui.data_expression.js
```diff
@@ -41,6 +42,7 @@ const DataExpressionMixin = extend({}, DataHelperMixin, {
                 store: new ArrayStore(this.option('items')),
                 pageSize: 0
             });
+            this._dataController = new DataController(this._dataSource);
```
ksercs: why do we create dataSource here manually, not use `initDataController` function?

Zedwag: done

js/ui/collection/ui.collection_widget.edit.js
Zedwag: done

ksercs (dismissed):

EugeniyKiyashko (approved):

ksercs (approved):

PR #22030, opened 2022-06-28
ShadomDOM support: experimental css inside shadow DOM, shadowDOM mode for QUnit tests

AlexKamaev:


js/core/dom_component.js
```diff
@@ -5,6 +5,7 @@ import windowResizeCallbacks from '../core/utils/resize_callbacks';
 import { Component } from './component';
 import { TemplateManager } from './template_manager';
 import { attachInstanceToElement, getInstanceByElement } from './utils/public_component';
+import { addShadowDOMStyles } from './utils/shadow-dom';
```
ksercs: We use CamelCase for abbreviations
```suggestion
import { addShadowDomStyles } from './utils/shadow-dom';
```

js/core/utils/shadow-dom.js
```diff
@@ -0,0 +1,59 @@
+let constructedStyleSheet = null;
+
+function createConstructedStyleSheet(rootNode) {
+    try {
+        // eslint-disable-next-line no-undef
+        return new CSSStyleSheet();
```
ksercs: seems like we can write it this way to get rid of lint disable:
```suggestion
        return new window.CSSStyleSheet();
```

js/core/utils/shadow-dom.js
```diff
@@ -0,0 +1,59 @@
...
+    } catch(err) {
+        const styleEl = rootNode.ownerDocument.createElement('style');
```
ksercs:
```suggestion
        const styleElement = rootNode.ownerDocument.createElement('style');
```

js/core/utils/shadow-dom.js
```diff
@@ -0,0 +1,59 @@
+let constructedStyleSheet = null;
+
+function createConstructedStyleSheet(rootNode) {
+    try {
+        // eslint-disable-next-line no-undef
+        return new CSSStyleSheet();
+    } catch(err) {
```
ksercs: In which cases we get an error here? According to [canIUse](https://caniuse.com/?search=CssStyleSheet) CssStyleSheet is supported in all browsers

js/core/utils/shadow-dom.js
```diff
@@ -0,0 +1,59 @@
+let constructedStyleSheet = null;
```
ksercs: we name files with `_` underscore, not dash. `js/core/utils/shadow_dom.js`

js/core/utils/shadow-dom.js
```diff
@@ -0,0 +1,59 @@
...
+        const styleEl = rootNode.ownerDocument.createElement('style');
+
+        rootNode.appendChild(styleEl);
+
+        return styleEl.sheet;
+    }
+}
+
+function processRules(styleSheets, needApplyAllStyles) {
+    [...styleSheets].forEach(sheet => {
```
ksercs: According to the [StyleSheets documentation](https://developer.mozilla.org/en-US/docs/Web/API/StyleSheetList#examples) we can iterate through the StyleSheets using usual for loop and its indexes. It means it's not necessary to convert the Array-like styleSheets object to an array with O(n) time complexity.
```javascript
    for (let i = 0; i < styleSheets.length; ++i) {
        const sheet = styleSheets[i];
        // ...
    }
```

js/core/utils/shadow-dom.js
```diff
@@ -0,0 +1,59 @@
...
+        // NOTE: need try/catch block for not-supported cross-domain css
```
ksercs: let's move this note to `catch` block to make it not empty. Like "Error can occur because of cross-domain css restriction"

js/core/utils/shadow-dom.js
```diff
@@ -0,0 +1,59 @@
...
+                insertRule(sheet.cssRules[j], needApplyAllStyles);
+            }
+        } catch(err) {
+        }
+    });
+}
+
+function insertRule(rule, needApplyAllStyles) {
+    const isDxRule = needApplyAllStyles ||
+                     rule.selectorText?.includes('.dx-') ||
```
ksercs: Seems like 'dx-' substring occurrence is enough, it's not necessary to transform it to selector, am I right? 

We can move `'dx-'` string to a constant `DX_PREFIX` to reduce string recreation

js/core/utils/shadow-dom.js
```diff
@@ -0,0 +1,59 @@
...
+                     rule.cssRules?.[0]?.selectorText?.includes('.dx-') ||
+                     rule.name?.startsWith('dx-') ||
```
ksercs: We can sort this checks by "light weighting" to improve performance. For example, this one should be first after `needApplyCllStyles`

testing/helpers/shadowDOM.js
```diff
@@ -1,13 +1,29 @@
...
-    this.root = root;
-    this.container = container;
-    this.control = control;
+import $ from 'jquery';
+
+export function ensureShadowRootIfNeeded() {
+    let container = document.querySelector('#qunit-fixture');
+
+    if(QUnit.urlParams['shadowDom'] && QUnit.urlParams['nojquery']) {
+        const root = document.querySelector('#qunit-fixture');
```
ksercs: Why we make this search the second time? We already have the result saved in `root` variable.

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/common.tests.js
```diff
@@ -5,7 +5,7 @@ import devices from 'core/devices';
 import eventsEngine from 'events/core/events_engine';
 import keyboardMock from '../../../helpers/keyboardMock.js';
 import pointerMock from '../../../helpers/pointerMock.js';
-import { appendShadowRoot } from '../../../helpers/shadowDOM.js';
+import { get } from '../../../helpers/shadowDOM.js';
```
ksercs: As i understood, now we need to refactor all our tests using `get` and `ensureShadowRootIfNeeded` to be able to run tests in ShadowDom (and start writing tests only using this api). Can we design a solution which help us not to change current testing approach?

As an option, we can add a specific `QUnit.testStart` in `qunitExtension.js` when `Use ShadowSom` checkbox is enabled. Example as a start point:
```javascript
            const initjQuery = jQuery.fn.init;
            $.fn.init = function(s, c, r) {
                const result = new initjQuery(s, c, r);
                return new initjQuery(result?.get(0).shadowRoot || result, c, r);
            };
``` 
to get shadowRoot instead of usual dom element + similar approach for shadowDom attach

js/core/utils/shadow-dom.js
AlexKamaev: Safari does not support a constructor CssStyleSheet, though the CssStyleSheet is supported

js/core/utils/shadow-dom.js
AlexKamaev: We can. However, in this case n is too small I believe

js/core/utils/shadow-dom.js
AlexKamaev: > Seems like 'dx-' substring occurrence is enough, it's not necessary to transform it to selector, am I right?

Not sure I understand this. Do you mean that we need to remove the `.` char?

js/core/utils/shadow-dom.js
AlexKamaev: I think this will not have effect. The biggest percent of rules are covered with `rule.selectorText?.includes('.dx-')`.
If I remember correctly the `name` rule is applied only for a few rules like `@keyframes` or smth like this.

testing/helpers/shadowDOM.js
AlexKamaev: OMG. Thank you

js/core/utils/shadow-dom.js
AlexKamaev: ESLint does not allow to use `window` here

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/common.tests.js
AlexKamaev: I'll think about this approach

AlexKamaev: done

js/core/utils/shadow-dom.js
ksercs: yeap, a suggest to move `dx-` string to a constant and use it in every rule here

js/core/utils/shadow-dom.js
ksercs: oh, ok

js/core/dom_component.js
```diff
@@ -122,6 +123,8 @@ const DOMComponent = Component.inherit({
 
     _render() {
         this._attachVisibilityChangeHandlers();
+
+        addShadowDomStyles(this._$element);
```
ksercs:
```suggestion
        addShadowDomStyles(this.$element());
```

testing/helpers/qunitExtensions.js
```diff
@@ -153,6 +153,82 @@
...
+        const shadowContainer = document.createElement('div');
+        shadowContainer.part = 'shadow';
+
+        root.shadowRoot.appendChild(shadowContainer);
+    }
+
+    function clearShadowRootTree() {
+        const container = get('div');
+
+        jQuery(container).empty();
```
ksercs, line 185: Is removing really not enough?

testing/helpers/qunitExtensions.js
```diff
@@ -153,6 +153,82 @@
...
+    QUnit.testStart(function() {
+        if(!isInShadowDomMode()) {
+            return;
+        }
+
+        addShadowRootTree();
+
+        jQueryInit = jQuery.fn.init;
+
+        jQuery.fn.init = function(s, c, r) {
```
ksercs: let's use more meaningful names instead of `s`, `c`, `r`

testing/helpers/qunitExtensions.js
```diff
@@ -153,6 +153,82 @@
...
+            const result = new jQueryInit(s, c, r);
+            const resultElement = result.get(0);
+
+            if(!resultElement) {
+                return new jQueryInit(get(s), c, r);
+            }
+
+            if(resultElement === getRoot().host) {
+                return new jQueryInit(get('div'), c, r);
```
ksercs: why we search for some "div", not selector? Could you please explain this conditions?

testing/helpers/qunitExtensions.js
AlexKamaev: It was related to jQuery and how jQuery clears the elemens inside the selector. However, after some refactoring this line is redundant.

testing/helpers/qunitExtensions.js
AlexKamaev: This is an edge case when we use `$('qunit-fixture')` as a selector.
When we do not use the shadow DOM we just put markup inside the `$(qunit-fixture)` element.
However, if we use shadow DOM we need to place markup to the element Shadow Root, which is that div.

AlexKamaev: We also need this `div` because we clear it insede the TestDone hook. JQuery cannot clear/remove ShadowRoot itself

ksercs (approved):
Great job!

Zedwag (approved):

PR #22046, opened 2022-06-30
Overlay: add "_ignoreElementAttrDeprecation" option

AlekseyMartynov:
This will allow to correctly re-fix https://devexpress.com/issue=T1084114 and resolve https://devexpress.com/issue=T1097600

### DIFF
--- a/js/ui/overlay/ui.overlay.js
+++ b/js/ui/overlay/ui.overlay.js
@@ -144,6 +144,8 @@ const Overlay = Widget.inherit({
 
             _ignoreCopyRootClassesToWrapperDeprecation: false,
 
+            _ignoreElementAttrDeprecation: false,
+
             onShowing: null,
 
             onShown: null,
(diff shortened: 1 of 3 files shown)


js/ui/overlay/ui.overlay.js
```diff
@@ -213,18 +215,23 @@ const Overlay = Widget.inherit({
...
-            'elementAttr': { since: '21.2', message: 'Use the "wrapperAttr" option instead' }
-        });
-    },
-
     ctor: function(element, options) {
         this.callBase(element, options);
 
-        if(options && options.copyRootClassesToWrapper && !options._ignoreCopyRootClassesToWrapperDeprecation) {
-            errors.log('W0001', this.NAME, 'copyRootClassesToWrapper', '21.2', 'Use the "wrapperAttr" option instead');
+        function createWrapperAttrDeprecationInfo() {
```
ksercs, line 221: Why should we create a function, not just once create `wrapperAttrDeprecationInfo` object? I don't think function creation is more light-weight operation than object creation.

ksercs:
We usually start name with `Component: ` prefix and select an assign. It makes navigation by history simpler.

js/ui/overlay/ui.overlay.js
AlekseyMartynov: I'm also not sure which way is more lightweight, if there is any difference at all. Do you suggest to declare a local const instead? Or, alternatively, move the function to the module top level?

ksercs (approved):

Zedwag (approved):

PR #22050, opened 2022-06-30
Scheduler - Timeline view should have Cross-Scrolling enabled by default (T1099189)

AntonSermyazhko:


testing/tests/DevExpress.ui.widgets.scheduler/integration.workSpace.tests.js
```diff
@@ -1721,11 +1721,15 @@ module('Resource Cell Template', () => {
                         ],
                     }],
                     resourceCellTemplate: function(cellData, cellIndex, cellElement) {
-                        if(!cellIndex) {
-                            const $cell = $(cellElement);
-                            assert.equal(getOuterWidth($cell), 100, 'Resource cell width is OK');
-                            assert.roughEqual(getOuterHeight($cell), 276, 1.001, 'Resource cell height is OK');
-                        }
+                        const done = assert.async();
+                        setTimeout(() => {
```
williamvinogradov, line 1725: It's a comment to discuss.

Maybe, we can use a fake setTimeout in the old QUnit tests?
Something like:
https://www.npmjs.com/package/@sinonjs/fake-timers

But, I'm not sure that this package is suitable for us, because we have a very old version of Sinon 
(we use 1.10.3, the actual version is 14.0.0)

js/renovation/ui/scheduler/view_model/to_test/views/utils/base.ts
```diff
@@ -147,6 +148,10 @@ export const isSupportMultiDayAppointments = (
   VIEWS.TIMELINE_MONTH,
 ].includes(viewType);
 
+export const isTimelineView = (
+  viewType: ViewType,
+): boolean => !!TIMELINE_VIEWS.find((type) => viewType === type);
```
williamvinogradov: Minor comment.

If we will use Set / Record instead of the array, we can make this function easier.
See next comment.

js/renovation/ui/scheduler/view_model/to_test/views/utils/const.ts
```diff
@@ -0,0 +1,20 @@
...
+  WORK_WEEK: 'workWeek',
+  MONTH: 'month',
+  TIMELINE_DAY: 'timelineDay',
+  TIMELINE_WEEK: 'timelineWeek',
+  TIMELINE_WORK_WEEK: 'timelineWorkWeek',
+  TIMELINE_MONTH: 'timelineMonth',
+  AGENDA: 'agenda',
+};
+
+export const TIMELINE_VIEWS: ViewType[] = [
```
williamvinogradov: Minor comment.

Maybe, we shall use Set or Record<ViewType, boolean> instead of the array here?
Connected with the previous comment.

williamvinogradov (dismissed):

testing/tests/DevExpress.ui.widgets.scheduler/integration.workSpace.tests.js
AntonSermyazhko: This approach doesn't work for us because our templating engine needs to render asynchronously.

js/renovation/ui/scheduler/view_model/to_test/views/utils/const.ts
AntonSermyazhko: Done.
Also, I made some refactoring for removing a duplicate.

js/renovation/ui/scheduler/view_model/to_test/views/utils/base.ts
AntonSermyazhko: Done

williamvinogradov (approved):

PR #22064, opened 2022-07-03
Lookup: Placeholder shouldn't be removed when the items option is changed (T1099804)

EugeniyKiyashko:


ksercs:
Reading this pull request name I cannot understand: it is a problem that placeholder is removed? Or it should be removed and it is an expected behavior?

testing/testcafe/tests/editors/autocomplete/common.ts
```diff
@@ -0,0 +1,31 @@
...
+import createWidget from '../../../helpers/createWidget';
+import { restoreBrowserSize } from '../../../helpers/restoreBrowserSize';
+
+fixture`Autocomplete_placeholder`
+  .page(url(__dirname, '../../container.html'))
+  .afterEach(async (t) => {
+    await restoreBrowserSize(t);
+  });
+
+test('The placeholder is visible after changing of items option when value is not choosen', async (t) => {
```
ksercs:
```suggestion
test('Placeholder is visible after items option change when value is not chosen (T1099804)', async (t) => {
```

testing/testcafe/tests/editors/autocomplete/common.ts
```diff
@@ -0,0 +1,31 @@
+import { createScreenshotsComparer } from 'devextreme-screenshot-comparer';
+import url from '../../../helpers/getPageUrl';
+import Autocomplete from '../../../model/autocomplete';
+import createWidget from '../../../helpers/createWidget';
+import { restoreBrowserSize } from '../../../helpers/restoreBrowserSize';
+
+fixture`Autocomplete_placeholder`
+  .page(url(__dirname, '../../container.html'))
+  .afterEach(async (t) => {
+    await restoreBrowserSize(t);
```
ksercs: Do we really need a functional test for this scenario? Seems like unit test can be written easy as well

ksercs (approved):

testing/testcafe/tests/editors/autocomplete/common.ts
EugeniyKiyashko: Yes, there are no testcafe tests for placeholder.

Zedwag (approved):

PR #22101, opened 2022-07-08
Shadow DOM Support: attempt to add shadow dom tasks

AlexKamaev:

### DIFF
--- a/js/core/dom_adapter.js
+++ b/js/core/dom_adapter.js
@@ -1,6 +1,7 @@
 /* global document */
 import injector from './utils/dependency_injector';
 import { noop } from './utils/common';
+import { getShadowElementsFromPoint } from './utils/shadow_dom';
 
 const ELEMENT_NODE = 1;
 const TEXT_NODE = 3;
@@ -122,9 +123,13 @@ const nativeDOMAdapterStrategy = {
     },
 
     getActiveElement(element) {
-        const activeElementHolder = element?.getRootNode?.() ?? this._document;
+        const activeElementHolder = this.getRootNode(element);
 
         return activeElementHolder.activeElement;
+    },
+
+    getRootNode(element) {
+        return element?.getRootNode?.() ?? this.getDocument();
     },
 
     getBody() {
@@ -171,8 +176,14 @@ const nativeDOMAdapterStrategy = {
         };
     },
 
-    elementsFromPoint(x, y) {
-        return this._document.elementsFromPoint(x, y);
+    elementsFromPoint(x, y, element) {
+        const activeElementHolder = this.getRootNode(element);
+
+        if(activeElementHolder.host) {
+            return getShadowElementsFromPoint(x, y, activeElementHolder);
+        }
+
+        return activeElementHolder.elementsFromPoint(x, y);
     }
 };
 
--- a/js/core/dom_adapter.js
+++ b/js/core/dom_adapter.js
@@ -129,7 +129,7 @@ const nativeDOMAdapterStrategy = {
     },
 
     getRootNode(element) {
-        return element?.getRootNode?.() ?? this.getDocument();
+        return element?.getRootNode?.() ?? this._document;
     },
 
     getBody() {
--- a/js/core/utils/shadow_dom.js
+++ b/js/core/utils/shadow_dom.js
@@ -61,32 +61,29 @@ export function addShadowDomStyles($element) {
     root.adoptedStyleSheets = [constructedStyleSheet];
 }
 
-function getShadowDomElementsInViewPort(x, y, el) {
-    let result = [];
+function isPositionInElementRectangle(element, x, y) {
+    const rect = element.getBoundingClientRect();
 
-    const rect = el.getBoundingClientRect();
-
-    if(rect) {
-        if(x >= rect.left && x <= rect.right - 1 && y <= rect.bottom - 1 && y >= rect.top) {
-            result = [el];
-
-            for(let i = 0; i < el.children.length; i++) {
-                result.unshift(...getShadowDomElementsInViewPort(x, y, el.children[i]));
-            }
-        }
-    }
-
-    return result;
+    return rect && x >= rect.left && x < rect.right && y >= rect.top && y < rect.bottom;
 }
 
 export function getShadowElementsFromPoint(x, y, root) {
     const result = [];
-    // eslint-disable-next-line no-undef
-    const childNodes = [...root.childNodes].filter(node => node.nodeType === Node.ELEMENT_NODE);
+    const elementQueue = [root];
 
-    for(let i = 0; i < childNodes.length; i++) {
-        result.push(...getShadowDomElementsInViewPort(x, y, childNodes[i]));
+    while(elementQueue.length) {
+        const el = elementQueue.shift();
+
+        for(let i = 0; i < el.childNodes.length; i++) {
+            const childNode = el.childNodes[i];
+
+            // eslint-disable-next-line no-undef
+            if(childNode.nodeType === Node.ELEMENT_NODE && isPositionInElementRectangle(childNode, x, y)) {
+                elementQueue.push(childNode);
+                result.push(childNode);
+            }
+        }
     }
 
-    return result;
+    return result.reverse();
 }
--- a/js/core/utils/shadow_dom.js
+++ b/js/core/utils/shadow_dom.js
@@ -60,3 +60,33 @@ export function addShadowDomStyles($element) {
 
     root.adoptedStyleSheets = [constructedStyleSheet];
 }
+
+function getShadowDomElementsInViewPort(x, y, el) {
+    let result = [];
+
+    const rect = el.getBoundingClientRect();
+
+    if(rect) {
+        if(x >= rect.left && x <= rect.right - 1 && y <= rect.bottom - 1 && y >= rect.top) {
+            result = [el];
+
+            for(let i = 0; i < el.children.length; i++) {
+                result.unshift(...getShadowDomElementsInViewPort(x, y, el.children[i]));
+            }
+        }
+    }
+
+    return result;
+}
+
+export function getShadowElementsFromPoint(x, y, root) {
+    const result = [];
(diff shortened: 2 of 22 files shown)


js/core/utils/shadow_dom.js
```diff
@@ -60,3 +60,33 @@ export function addShadowDomStyles($element) {
...
+        }
+    }
+
+    return result;
+}
+
+export function getShadowElementsFromPoint(x, y, root) {
+    const result = [];
+    // eslint-disable-next-line no-undef
+    const childNodes = [...root.childNodes].filter(node => node.nodeType === Node.ELEMENT_NODE);
```
ksercs: It would be faster if we just check this condition in `for` loop

js/core/utils/shadow_dom.js
```diff
@@ -60,3 +60,33 @@ export function addShadowDomStyles($element) {
...
+    let result = [];
+
+    const rect = el.getBoundingClientRect();
+
+    if(rect) {
+        if(x >= rect.left && x <= rect.right - 1 && y <= rect.bottom - 1 && y >= rect.top) {
+            result = [el];
+
+            for(let i = 0; i < el.children.length; i++) {
+                result.unshift(...getShadowDomElementsInViewPort(x, y, el.children[i]));
```
ksercs: do we really need use exactly `unshift` with O(n) time instead of `push` with O(1)?

js/core/utils/shadow_dom.js
```diff
@@ -60,3 +60,33 @@ export function addShadowDomStyles($element) {
...
     root.adoptedStyleSheets = [constructedStyleSheet];
 }
+
+function getShadowDomElementsInViewPort(x, y, el) {
+    let result = [];
+
+    const rect = el.getBoundingClientRect();
+
+    if(rect) {
+        if(x >= rect.left && x <= rect.right - 1 && y <= rect.bottom - 1 && y >= rect.top) {
```
ksercs: hm, we check that `y` coordinate is out of rect, is it correct?

ksercs: why do we use `-1` delta for right and bottom?

js/core/utils/shadow_dom.js
ksercs: this function differs with `getShadowDomElementInViewPort` only with this nodeType check. Cannot we merge this 2 functions?

js/core/dom_adapter.js
```diff
@@ -171,8 +176,14 @@ const nativeDOMAdapterStrategy = {
         };
     },
 
-    elementsFromPoint(x, y) {
-        return this._document.elementsFromPoint(x, y);
+    elementsFromPoint(x, y, element) {
+        const activeElementHolder = this.getRootNode(element);
+
+        if(activeElementHolder.host) {
+            return getShadowElementsFromPoint(x, y, activeElementHolder);
```
ksercs, line 183: do we really need implement it by ourself? Seems like shadowRoot should have such api: https://docs.w3cub.com/dom/documentorshadowroot

js/core/dom_adapter.js
```diff
@@ -171,8 +176,14 @@ const nativeDOMAdapterStrategy = {
         };
     },
 
-    elementsFromPoint(x, y) {
-        return this._document.elementsFromPoint(x, y);
+    elementsFromPoint(x, y, element) {
```
ksercs, line 179: Can you explain please, why should this function get `element` as a parameter?

js/ui/popup/popup_drag.js
```diff
@@ -131,7 +131,7 @@ class PopupDrag {
     }
 
     _getContainerDimensions() {
-        const document = domAdapter.getDocument();
+        const document = domAdapter.getRootNode(this._draggableElement);
```
ksercs: Why do we need this change? Will this code work `rootNode.body.clientHeight`?

testing/helpers/fileManagerHelpers.js
```diff
@@ -219,7 +219,8 @@ export class FileManagerWrapper {
     }
 
     getToolbarElements() {
-        return this._$element.find(`.${Consts.TOOLBAR_CLASS} .${Consts.BUTTON_TEXT_CLASS}:visible, .${Consts.TOOLBAR_CLASS} .${Consts.NATIVE_TOOLBAR_CLASS}:visible .${Consts.DROP_DOWN_BUTTON_CLASS}`);
+        return this._$element.find(`.${Consts.TOOLBAR_CLASS} .${Consts.BUTTON_TEXT_CLASS}, .${Consts.TOOLBAR_CLASS} .${Consts.NATIVE_TOOLBAR_CLASS} .${Consts.DROP_DOWN_BUTTON_CLASS}`)
```
ksercs, line 222: This code checked some concrete parent elements' visibility, but now it gets all final elements and filters its by visibility. Is it expected?

testing/tests/DevExpress.ui.widgets/ganttParts/events.tests.js
```diff
@@ -116,14 +116,14 @@ QUnit.module('Events', moduleConfig, () => {
         this.createInstance(my_options);
         this.clock.tick();
 
-        const $top_item = $('body').find('.gst_item');
+        const $top_item = $('#qunit-fixture').find('.gst_item');
```
ksercs, line 119: Why is it necessary?

testing/tests/DevExpress.ui.widgets/sortable.tests.js
```diff
@@ -2615,6 +2618,8 @@ QUnit.module('With scroll', getModuleConfigForTestsWithScroll('#itemsWithScroll'
     });
 
     QUnit.test('Placeholder should be visible after page scrolling (T871213)', function(assert) {
+        const fixtureRoot = document.querySelector('#qunit-fixture');
```
ksercs, line 2621: Why do we need this change?

testing/tests/DevExpress.ui.widgets/toolbarParts/toolbar.themes.sharedTests.js
```diff
@@ -45,7 +45,7 @@ export const runThemesSharedTests = function(moduleNamePostfix) {
                     }]
                 };
 
-                const toolbarElement = document.getElementById('toolbar');
+                const toolbarElement = $('#toolbar')[0];
```
ksercs:
```suggestion
                const toolbarElement = $('#toolbar').get(0);
```

js/core/utils/shadow_dom.js
AlexKamaev: We cannot use `push` here since we need to have reversed order. I'll check if it will be enough to make one `reverse` call in the end.

js/core/utils/shadow_dom.js
AlexKamaev: I did not get it. Why? `y > top && and y < bottom` is correct way to check if the point inside the rect

AlexKamaev: Because `right` and `bottom` values are not in the rect. You can check it here: https://yari-demos.prod.mdn.mozit.cloud/en-US/docs/Web/API/Element/getBoundingClientRect/_sample_.basic.html
The `elementsFromPoint` will return incorrect element

js/core/utils/shadow_dom.js
AlexKamaev: We can, but I would prefer to leave this as is, because the `root` element does not have the `children` property, while I'm using the `children` property inside the `getShadowDomElementsInViewPort` function.

js/core/dom_adapter.js
AlexKamaev: Unfortunately, this does not work as I expected. It does not consider shadow elements

js/core/dom_adapter.js
AlexKamaev: Because element can be inside of shadowRoot. In this case we cannot use the `document.elementsFromPoint` method.

js/ui/popup/popup_drag.js
AlexKamaev: Let me check it please

testing/helpers/fileManagerHelpers.js
AlexKamaev: This may be confusing, but unfortunately we cannot use the `:visible` selector in the `document.querySelectorAll` method, so I moved this visibility check outside of the selector. It should work in the same manner.

testing/tests/DevExpress.ui.widgets/ganttParts/events.tests.js
AlexKamaev: because shadowDom does not have the `body` selector

testing/tests/DevExpress.ui.widgets/sortable.tests.js
AlexKamaev: This test was the most difficult to adjust to shadow Dom. We need to modify the `#qunit-fixture` element itself, but not it's shadow container. This change is specific only for this test

js/core/utils/shadow_dom.js
ksercs: doesn't not strict comparison help?

testing/tests/DevExpress.ui.widgets/sortable.tests.js
ksercs: Let's add a note

testing/helpers/fileManagerHelpers.js
ksercs: @EugeniyKiyashko  how do u think, can this change make test work another way?

js/ui/popup/popup_drag.js
AlexKamaev: I removed this change

js/core/utils/shadow_dom.js
AlexKamaev: Ok, I merget it, and removed recursion

AlexKamaev:
see https://github.com/DevExpress/DevExtreme/pull/22128

PR #22120, opened 2022-07-11
TransitionExecutor - fix: api requires jqury but it should not (T1101319)

GoodDayForSurf:


js/animation/transition_executor/transition_executor.js
```diff
@@ -29,6 +29,9 @@ export const TransitionExecutor = Class.inherit({
     _createAnimations: function($elements, initialConfig, configModifier, type) {
         const that = this;
         const result = [];
+        if(!$elements.jquery && !$elements.dxRenderer) {
+            $elements = $($elements);
```
dxvladislavvolkov: I think it is not good when you change the argument value

MikeVitik (approved):

PR #22128, opened 2022-07-12
Shadow DOM support: widgets task + tests

AlexKamaev:


js/core/utils/shadow_dom.js
```diff
@@ -60,3 +60,34 @@ export function addShadowDomStyles($element) {
...
+
+    return rect && x >= rect.left && x < rect.right && y >= rect.top && y < rect.bottom;
+}
+
+export function getShadowElementsFromPoint(x, y, root) {
+    const result = [];
+    const elementQueue = [root];
+
+    while(elementQueue.length) {
+        const el = elementQueue.shift();
```
ksercs, line 99: This code gives us `O(n^2)` asymptotic. If we need to implement BFS it's good idea to implement queue data structure (using Object foe example)

js/core/utils/shadow_dom.js
```diff
@@ -60,3 +60,34 @@ export function addShadowDomStyles($element) {
...
+
+        for(let i = 0; i < el.childNodes.length; i++) {
+            const childNode = el.childNodes[i];
+
+            // eslint-disable-next-line no-undef
+            if(childNode.nodeType === Node.ELEMENT_NODE &&
+               isPositionInElementRectangle(childNode, x, y) &&
+               // eslint-disable-next-line no-undef
+               getComputedStyle(childNode).pointerEvents !== 'none'
```
ksercs, line 108: This code will trigger relayout, which is very expensive operation. Is it really necessary for our use cases?

js/ui/draggable.js
```diff
@@ -866,7 +866,7 @@ const Draggable = DOMComponent.inherit({
         const $targetDraggableElement = this.$element();
 
         const mousePosition = getMousePosition(e);
-        const elements = domAdapter.elementsFromPoint(mousePosition.x, mousePosition.y);
+        const elements = domAdapter.elementsFromPoint(mousePosition.x, mousePosition.y, this.element());
```
ksercs, line 869: Does this work? Seem like it gonna be
```suggestion
        const elements = domAdapter.elementsFromPoint(mousePosition.x, mousePosition.y, this.$element().get(0));
```

testing/tests/DevExpress.ui.widgets/deferRendering.animation.tests.js
```diff
@@ -23,7 +16,16 @@ QUnit.testStart(function() {
...
+    const style = document.createElement('style');
+    style.innerHTML =
+        '.test-staggering-item,\
+        test-no-staggering-item {\
+            height: 10px;\
+            width: 10px;\
+        }';
+
     $('#qunit-fixture').html(markup);
+    document.body.appendChild(style);
```
ksercs, line 29: why is this change necessary?

testing/tests/DevExpress.ui.widgets/diagramParts/toolbox.tests.js
```diff
@@ -27,24 +27,24 @@ QUnit.module('Toolbox', {
 }, () => {
     test('should render if toolbox.visibility is "visible"', function(assert) {
         this.instance.option('toolbox.visibility', 'visible');
-        const $accordion = $('body').find(Consts.TOOLBOX_ACCORDION_SELECTOR);
+        const $accordion = $('#qunit-fixture').find(Consts.TOOLBOX_ACCORDION_SELECTOR);
```
ksercs, line 30: why cannot we use body here?

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -127,6 +128,8 @@ const toSelector = (cssClass) => `.${cssClass}`;
 
 const moduleConfig = {
     beforeEach: function() {
+        viewPort($('#qunit-fixture').addClass(VIEWPORT_CLASS));
```
ksercs: shouldn't we restore viewport value in afterEach?

testing/tests/DevExpress.ui.widgets/tabPanel.width.tests.js
```diff
@@ -15,6 +16,8 @@ QUnit.module('Tabs width', () => {
             this.$tabPanel = $('<div>');
             this.assert = assert;
             this.setWidthApproach = setWidthApproach;
+
+            addShadowDomStyles($('#qunit-fixture'));
```
ksercs: why should we do this? I thought that shadow dom styles are added on component render

js/core/utils/shadow_dom.js
AlexKamaev: ok

js/core/utils/shadow_dom.js
AlexKamaev: unfortunately we need to check this for overlays.
I am not sure if it's needed only for overlays, but if so, we can skip just overlay elements

js/ui/draggable.js
AlexKamaev: both these approaches work

testing/tests/DevExpress.ui.widgets/deferRendering.animation.tests.js
AlexKamaev: we need to style for the `.test-staggering-item` element, which is in root document, not in shadow root, so we need to attach styles to the body element

testing/tests/DevExpress.ui.widgets/diagramParts/toolbox.tests.js
AlexKamaev: because we do not have the `body` element inside the shadow root

testing/tests/DevExpress.ui.widgets/overlay.tests.js
AlexKamaev: well. It was not being restored in the previous version. However, using the `beforeEach` is redundant. I moved this code to QUnit.testStart. Thank you for pointing me to this

testing/tests/DevExpress.ui.widgets/tabPanel.width.tests.js
AlexKamaev: I'll check it. Probably it's a fix that is not actual anymore

js/core/utils/shadow_dom.js
```diff
@@ -60,3 +60,55 @@ export function addShadowDomStyles($element) {
...
+            ) {
+                elementQueue.push(childNode);
+                result.push(childNode);
+            }
+        }
+    }
+
+    return result.reverse();
```
ksercs: seems like we don't need `result` array anymore - we can just get `items` array from our queue

testing/tests/DevExpress.ui.widgets/deferRendering.animation.tests.js
ksercs: i think it's good idea to add a comment for such specific case
`// NOTE: staggering item is placed in body outside of shadow root

testing/tests/DevExpress.ui.widgets/tabPanel.width.tests.js
AlexKamaev: not needed, thank you for pointing out

testing/tests/DevExpress.ui.widgets/deferRendering.animation.tests.js
AlexKamaev: ok

js/core/utils/shadow_dom.js
AlexKamaev: we can use items, but we still need to reverse them and exclude the shadow root element

ksercs: My idea is that we may not add an additional memory. After your last change we still have an additional array, but also make additional linear iteration.

Smth like this:
```javascript
const result = queue.getItems();
result.pop();

return result;
```

ksercs (approved):

Zedwag (approved):

PR #22160, opened 2022-07-14
ValidationEngine: validator should be detected as a part of a validation group if they are created on the same element (T1102012)

GoodDayForSurf:
Validation does not work correctly when using validator|adapter in a certain scenario

### DIFF
--- a/testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
@@ -57,6 +57,23 @@ QUnit.module('General', {
         assert.ok(validator.validate.calledOnce, 'Validator should be validated as part of group');
     });
 
+    QUnit.test('validator should work a part of a validation group if they are created on the same element (T1102012)', function(assert) {
+        const $groupContainer = $('#dxValidationGroup');
+        const group = this.fixture.createGroup($groupContainer);
+        const $validator = $groupContainer.dxValidator({
+            adapter: sinon.createStubInstance(DefaultAdapter)
+        });
+        const validator = $validator.dxValidator('instance');
+        validator.validate = sinon.spy(validator.validate);
+
+        // act
+        triggerShownEvent($groupContainer);
+        ValidationEngine.validateGroup(group);
+
+        // assert
+        assert.ok(validator.validate.calledOnce, 'Validator should be validated as part of group');
+    });
+
     QUnit.test('group should be validated positively (async)', function(assert) {
         const $container = $('#dxValidationGroup');
         const group = this.fixture.createGroup($container);
--- a/testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
@@ -57,7 +57,7 @@ QUnit.module('General', {
         assert.ok(validator.validate.calledOnce, 'Validator should be validated as part of group');
     });
 
-    QUnit.test('validator should work a part of a validation group if they are created on the same element (T1102012)', function(assert) {
+    QUnit.test('validator should be validated as a part of a validation group if they are created on the same element (T1102012)', function(assert) {
         const $groupContainer = $('#dxValidationGroup');
         const group = this.fixture.createGroup($groupContainer);
         const $validator = $groupContainer.dxValidator({
--- a/js/ui/validation_engine.js
+++ b/js/ui/validation_engine.js
@@ -522,6 +522,12 @@ const ValidationEngine = {
     },
 
     findGroup($element, model) {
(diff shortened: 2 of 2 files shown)


MikeVitik (dismissed):

testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
```diff
@@ -57,6 +57,22 @@ QUnit.module('General', {
         assert.ok(validator.validate.calledOnce, 'Validator should be validated as part of group');
     });
 
+    QUnit.test('should find group if validator is child of validationGroup ', function(assert) {
```
ksercs: Actual test name is too common to me. Who should find? Why we check it with this test? Besides it, this test checks the scenario when validator and validationGroup are created on the same element, not when validator is a child of a ValidationGroup.

Also it's a good idea to add a ticked ID as a suffix of a test name. It can make future debug easier.

```suggestion
    QUnit.test('validator should work a part of a validation group if they are created on the same element (T1102012)', function(assert) {
```

testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
```diff
@@ -57,6 +57,22 @@ QUnit.module('General', {
...
+        const $group_container = $('#dxValidationGroup');
```
ksercs:
```suggestion
        const $groupContainer = $('#dxValidationGroup');
```

testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
```diff
@@ -57,6 +57,22 @@ QUnit.module('General', {
...
+        const group = this.fixture.createGroup($group_container);
+        const $validator = $group_container.dxValidator({
+            adapter: sinon.createStubInstance(DefaultAdapter)
+        });
+        const validator = $validator.dxValidator('instance');
+        validator.validate = sinon.spy(validator.validate);
+        $validator.appendTo($group_container);
```
ksercs: Seems like this operation is excess because we already bound the Validator to `#validationGroup` element in 63th row

testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
```diff
@@ -57,6 +57,22 @@ QUnit.module('General', {
...
+        triggerShownEvent($group_container);
+        ValidationEngine.validateGroup(group);
```
ksercs, line 71: Let's add an empty row before this to separate test "action" part from "arrange" part. [AAA](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/)

testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
```diff
@@ -57,6 +57,22 @@ QUnit.module('General', {
...
+
+        // assert
```
ksercs, line 73:
```suggestion
```

testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
```diff
@@ -57,6 +57,22 @@ QUnit.module('General', {
...
+        assert.ok(validator.validate.calledOnce, 'Validator should be validated as part of group');
```
ksercs, line 73:
```suggestion
        assert.ok(validator.validate.calledOnce, 'Validator should be validated as a part of the group');
```

testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
GoodDayForSurf: yes..fixed

testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
GoodDayForSurf: done

testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
```diff
@@ -57,6 +57,23 @@ QUnit.module('General', {
         assert.ok(validator.validate.calledOnce, 'Validator should be validated as part of group');
     });
 
+    QUnit.test('validator should work a part of a validation group if they are created on the same element (T1102012)', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('validator should be validated as a part of a validation group if they are created on the same element (T1102012)', function(assert) {
```

ksercs (approved):

EugeniyKiyashko (approved):

PR #22163, opened 2022-07-15
Scheduler: Fix disappearing recurrence appointment in DST time issue (T1102713)

williamvinogradov:
See ticket: https://supportcenter.devexpress.com/internal/ticket/details/T1102713


js/ui/scheduler/recurrence.js
```diff
@@ -34,24 +34,69 @@ class RecurrenceProcessor {
...
+
+        return {
+            startIntervalDate,
+            minViewTime,
+            minViewDate,
+            maxViewDate,
+            startIntervalDateDSTShift: switchToSummerTime ? 0 : startDateDSTDifferenceMs,
+            appointmentDuration: duration,
+        };
+    }
```
AntonSermyazhko, line 81: I suggest using separate methods for these conversions because otherwise `_createRruleIntervalParams` would be too large.

AntonSermyazhko (approved):

PR #22171, opened 2022-07-15
Toolbar: replace devextreme inherits to ES6 class

EugeniyKiyashko:


js/ui/toolbar/ui.toolbar.base.js
```diff
@@ -32,42 +30,44 @@ const DEFAULT_DROPDOWNBUTTON_STYLING_MODE = 'contained';
 
 const TOOLBAR_ITEM_DATA_KEY = 'dxToolbarItemDataKey';
 
-const ToolbarBase = AsyncCollectionWidget.inherit({
-    compactMode: false,
+class ToolbarBase extends AsyncCollectionWidget {
```
ksercs, line 34: why did u decide to use `extends` instead of our usual `inherit`?

js/ui/toolbar/ui.toolbar.utils.js
```diff
@@ -30,14 +35,57 @@ export function toggleItemFocusableElementTabIndex(context, item) {
...
                 $focusTarget.attr('tabIndex', -1);
             } else {
-                $focusTarget.attr('tabIndex', tabIndex ? tabIndex : 0);
+                $focusTarget.attr('tabIndex', tabIndex ?? 0);
             }
         }
     }
 }
+
+export function updateDimensionsInMaterial() {
```
ksercs: why did u decided to move it to utils? Seems strange to me because component state is used here

js/ui/toolbar/ui.toolbar.base.js
EugeniyKiyashko: It is a build in and documented functionality to create a child of parent class. In debug mode, we immediately get into the parent function. Also works a quick transition to the implementation of the methods by `ctrl + click`. We wide use `extends` for new components Gantt, FileManager, Drawer, TextEditors and so on.

js/ui/toolbar/ui.toolbar.utils.js
EugeniyKiyashko: There are no strong arguments. I will return it as component's method - https://github.com/DevExpress/DevExtreme/pull/22171/commits/660b8dc639f4e1d1e94a239c82db90d5dc9ca9ef

ksercs (approved):

Zedwag (approved):

PR #22184, opened 2022-07-18
DateBox: disable blinking ValidationMessagePosition tests for material

Zedwag:


testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js
```diff
@@ -5500,6 +5502,43 @@ QUnit.module('datebox validation', {}, () => {
...
+                const dateBox = $dateBox.dxDateBox('instance');
+
+                dateBox.option('value', new Date());
+
+                const inputPosition = $dateBox.find(`.${TEXTEDITOR_INPUT_CLASS}`).get(0).getBoundingClientRect();
+                const messagePosition = $dateBox.find(`.${VALIDATION_MESSAGE_CONTENT}`).get(0).getBoundingClientRect();
+
+                assert.ok(checkFunction(messagePosition, inputPosition), 'position is correct');
+
+                themes.isMaterial = origIsMaterial;
```
ksercs: It seems like there is no specific code for material theme. Why should we write specific test for it?

js/ui/validation_message.js
```diff
@@ -158,7 +158,7 @@ const ValidationMessage = Overlay.inherit({
             boundary,
             my: positions[0],
             at: positions[1],
-            collision: 'none flip'
```
ksercs: Why do u think we should change it? I think that, otherwise, collision should be specified for both directions

testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js
Zedwag: fair enough

js/ui/validation_message.js
Zedwag: >> Why do u think we should change it
Yes
image

ksercs (dismissed):

ksercs (approved):

PR #22191, opened 2022-07-19
Popup: try to use full version of toolbar

EugeniyKiyashko:


ksercs (approved):
LGTM. I gonna trust u about all places using popup or popover are refactored

Zedwag (approved):

PR #22197, opened 2022-07-20
Overlays: refactor testcafe tests

EugeniyKiyashko:


ksercs:
Great job!

I only suggest to rerun these tests several times in GA to be sure that they are stable.

Zedwag (approved):

PR #22284, opened 2022-08-03
Slider: add new option 'callValueChange' to specify when value should be changed

dayanch-n:


js/common.d.ts
```diff
@@ -262,6 +262,13 @@ export type ValidationMessageMode = 'always' | 'auto';
  */
 export type ValidationStatus = 'valid' | 'invalid' | 'pending';
 
+/**
+ * @public
+ * @namespace DevExpress.common
+ */
+
+export type ValueChangeMode = 'instant' | 'eventual';
```
ksercs:
```suggestion
export type SliderValueChangeMode = 'instant' | 'eventual';
```

Actual name seems to be too common

testing/tests/DevExpress.ui.widgets.editors/slider.tests.js
```diff
@@ -1761,3 +1764,63 @@ module('if only the single value is possible', moduleOptions, () => {
...
+
+        assert.notOk(this.valueChangedHandler.called, 'the onValueChanged is not called');
+
+        this.pointer.swipeEnd(40);
+
+        assert.ok(this.valueChangedHandler.called, 'the onValueChanged is called');
+        assert.strictEqual(this.instance.option('value'), 90);
+
+    });
+    test('slider tooltip value should change with "eventual" valueChangeMode', function(assert) {
```
ksercs: Let's separate tests with an empty line.
Please, refactor this in all such places.

js/ui/range_slider.js
```diff
@@ -195,8 +195,9 @@ const RangeSlider = Slider.inherit({
 
         this._changeValueOnSwipe(newRatio);
 
-        const startValue = this.option('start');
-        const endValue = this.option('end');
+        const startValue = this._currentValue && this._currentValue[0] || this.option('start');
```
Zedwag: let's move getting value from `_currentValue` (or mb better name this something like `_actualValue`) in new function to avoid this strange construction

testing/tests/DevExpress.ui.widgets.editors/rangeSlider.tests.js
```diff
@@ -1343,3 +1343,81 @@ QUnit.module('valueChanged handler should receive correct event parameter', {
...
+        this.rightHandleTooltip = this.rightHandle.find(`.${TOOLTIP_CLASS}`);
+
+        this.range = this.$element.find('.' + SLIDER_RANGE_CLASS);
+
+        this.getLeftTooltipText = () => $.trim(this.leftHandleTooltip.text());
+        this.getRightTooltipText = () => $.trim(this.rightHandleTooltip.text());
+    }
+}, () => {
+    QUnit.test('left slider handle value should not change on swipe with "eventual" valueChangeMode', function(assert) {
+
```
Zedwag:
```suggestion
```

testing/tests/DevExpress.ui.widgets.editors/rangeSlider.tests.js
```diff
@@ -1343,3 +1343,81 @@ QUnit.module('valueChanged handler should receive correct event parameter', {
...
+        const pointer = pointerMock(this.leftHandle);
+        pointer.start().move(this.range.offset().left).down().move(200);
+        assert.notOk(this.valueChangedHandler.called, 'the onValueChanged is not called');
```
Zedwag, line 1379:
```suggestion
       
         assert.notOk(this.valueChangedHandler.called, 'the onValueChanged is not called');
```

testing/tests/DevExpress.ui.widgets.editors/rangeSlider.tests.js
```diff
@@ -1343,3 +1343,81 @@ QUnit.module('valueChanged handler should receive correct event parameter', {
...
+        this.leftHandleTooltip = this.leftHandle.find(`.${TOOLTIP_CLASS}`);
+        this.rightHandleTooltip = this.rightHandle.find(`.${TOOLTIP_CLASS}`);
+
+        this.range = this.$element.find('.' + SLIDER_RANGE_CLASS);
+
+        this.getLeftTooltipText = () => $.trim(this.leftHandleTooltip.text());
+        this.getRightTooltipText = () => $.trim(this.rightHandleTooltip.text());
+    }
+}, () => {
+    QUnit.test('left slider handle value should not change on swipe with "eventual" valueChangeMode', function(assert) {
```
Zedwag: there is no such thing like left slider handle value, there is only option value, so let's name it something like "value option should not change on left handle swipe with "eventual" valueChangeMode'

testing/tests/DevExpress.ui.widgets.editors/rangeSlider.tests.js
```diff
@@ -1343,3 +1343,81 @@ QUnit.module('valueChanged handler should receive correct event parameter', {
...
+        pointer.start().move(this.range.offset().left).down().move(200);
+        assert.notOk(this.valueChangedHandler.called, 'the onValueChanged is not called');
+
+        pointer.up();
+
+        assert.equal(this.instance.option('start'), 40);
+        assert.ok(this.valueChangedHandler.called, 'the onValueChanged is called');
+    });
+
+    QUnit.test('right slider handle value should not change on swipe with "eventual" valueChangeMode', function(assert) {
```
Zedwag: same as previous

testing/tests/DevExpress.ui.widgets.editors/rangeSlider.tests.js
```diff
@@ -1343,3 +1343,81 @@ QUnit.module('valueChanged handler should receive correct event parameter', {
...
+        const pointer = pointerMock(this.rightHandle);
+        pointer.start().down(this.rightHandle.offset().left + CONTAINER_MARGIN).move(200);
+
+        assert.strictEqual(this.getRightTooltipText(), '80');
+
+        pointer.move(100).up();
+
+        assert.strictEqual(this.getRightTooltipText(), '90');
+    });
+});
```
Zedwag, line 1445: we should add test on scenario that was broken: when we swipe right handle through left handle

Zedwag (dismissed):

testing/tests/DevExpress.ui.widgets.editors/rangeSlider.tests.js
dayanch-n: Do you think it is also relevant to write a test when is clicked (instead of swipe) for this scenario?

Zedwag: Looks like it's already done in this test https://github.com/DevExpress/DevExtreme/pull/22284/files#diff-4494ecb2395df1b3d5abda0b7e429e04da5c693a3480bb59f92cb02077b82307L309

js/ui/range_slider.js
```diff
@@ -195,8 +195,9 @@ const RangeSlider = Slider.inherit({
 
         this._changeValueOnSwipe(newRatio);
 
-        const startValue = this.option('start');
-        const endValue = this.option('end');
+        const startValue = this._getActualStartValue() || this.option('start');
```
ksercs: Shouldn't we use nullish coalescing operator `??` instead of `||` in all such places? Seems like this code does not work if `_getActualStartValue` returns `0`.

If it does not work, let's add a test first of all

js/ui/range_slider.js
```diff
@@ -297,6 +307,20 @@ const RangeSlider = Slider.inherit({
         this.option('value', [start, end]);
     },
 
+    _getActualStartValue: function() {
+        if(!this._actualValue) {
+            return undefined;
```
ksercs: let's return `this.option('start')` here to reduce copypaste in all such places.

js/ui/range_slider.js
```diff
@@ -297,6 +307,20 @@ const RangeSlider = Slider.inherit({
...
+        }
+        return this._actualValue[0];
+    },
+
+    _getActualEndValue: function() {
```
ksercs: I think it's a good idea to implement only one function `_getActualValue` instead `_getActualStartValue` and `_getActualEndValue` to reduce copypaste and simplify the code.

js/ui/slider.d.ts
```diff
@@ -159,6 +160,12 @@ export interface dxSliderBaseOptions<TComponent> extends dxTrackBarOptions<TComp
        */
       showMode?: TooltipShowMode;
     };
+    /**
+     * @docid
+     * @default 'instant'
```
ksercs: default value should be renamed to `onMoving`

js/ui/slider.d.ts
```diff
@@ -159,6 +160,12 @@ export interface dxSliderBaseOptions<TComponent> extends dxTrackBarOptions<TComp
...
+     * @public
+     */
+     SliderCallValueChange: SliderCallValueChange;
```
ksercs: option name is not `SliderCallValueChange`, it's `callValueChange`.
`SliderCallValueChange` is just a type of this option value.

js/ui/slider/ui.slider.js
```diff
@@ -380,16 +381,34 @@ const Slider = TrackBar.inherit({
 
     _renderStartHandler: function() {
         const pointerDownEventName = addNamespace(pointerEvents.down, this.NAME);
+        const pointerUpEventName = addNamespace(pointerEvents.up, this.NAME);
         const clickEventName = addNamespace(clickName, this.NAME);
         const startAction = this._createAction(this._startHandler.bind(this));
         const $element = this.$element();
 
+        const processOnMovingComplete = () => {
```
ksercs:
```suggestion
        const changeValueOnMovingComplete = () => {
```

js/ui/slider/ui.slider.js
```diff
@@ -380,16 +381,34 @@ const Slider = TrackBar.inherit({
...
         eventsEngine.off($element, pointerDownEventName);
         eventsEngine.on($element, pointerDownEventName, e => {
             if(isMouseEvent(e)) {
                 startAction({ event: e });
             }
         });
+
+        eventsEngine.off($element, pointerUpEventName);
+        eventsEngine.on($element, pointerUpEventName, e => {
+            if(isMouseEvent(e)) {
```
ksercs: why is this condition necessary? Seems like it will not work on mobile devices, because pointerup is `touch` event on mobiles.

js/ui/range_slider.js
```diff
@@ -297,6 +307,20 @@ const RangeSlider = Slider.inherit({
         this.option('value', [start, end]);
     },
 
+    _getActualStartValue: function() {
```
ksercs: I think it's better to implement getter and setter as a one function. Like it's done for [isTextCompositionInProgress](https://github.com/DevExpress/DevExtreme/blob/22_2/js/ui/drop_down_editor/ui.drop_down_list.js#L677) in ddList, for example.

It can simplify future debug because we have only one place where`_actualValue` prop changes

js/ui/slider/ui.slider.js
```diff
@@ -520,10 +548,24 @@ const Slider = TrackBar.inherit({
     },
 
     _setValueOnSwipe: function(value) {
+        this._actualValue = value;
+
+        if(this.option('callValueChange') === 'onMovingComplete') {
+            SliderHandle.getInstance(this._activeHandle()).option('value', value);
+            return;
```
ksercs: it's better to use `if-else` instead of such unexpected `return` to have only one function end point and simplify future debug

testing/tests/DevExpress.ui.widgets.editors/rangeSlider.tests.js
```diff
@@ -1343,3 +1343,96 @@ QUnit.module('valueChanged handler should receive correct event parameter', {
...
+    });
+
+    QUnit.test('value should be correctly updated when right handle is moved through left handle', function(assert) {
+        this.instance.option('start', 40);
+        const pointer = pointerMock(this.rightHandle);
+        pointer.start().down(this.rightHandle.offset().left + CONTAINER_MARGIN).move(-400).up();
+
+        assert.strictEqual(this.instance.option('start'), 20);
+    });
+    QUnit.test('value should be correctly updated when left handle is moved through right handle', function(assert) {
```
ksercs, line 1431: let's add an empty line to separate test cases

testing/tests/DevExpress.ui.widgets.editors/slider.tests.js
```diff
@@ -1761,3 +1764,65 @@ module('if only the single value is possible', moduleOptions, () => {
...
+        this.pointer.swipeStart();
+        this.pointer.swipe(20);
+
+        assert.notOk(this.valueChangedHandler.called, 'the onValueChanged is not called');
+
+        this.pointer.swipeEnd(40);
+
+        assert.ok(this.valueChangedHandler.called, 'the onValueChanged is called');
+        assert.strictEqual(this.instance.option('value'), 90);
+
```
ksercs:
```suggestion
```

testing/tests/DevExpress.ui.widgets.editors/slider.tests.js
```diff
@@ -1761,3 +1764,65 @@ module('if only the single value is possible', moduleOptions, () => {
...
+    });
+
+    test('slider tooltip value should change with "onMovingComplete" callValueChange', function(assert) {
```
ksercs:
```suggestion
    test('slider tooltip value should change on swipe if callValueChange = onMovingComplete', function(assert) {
```

testing/tests/DevExpress.ui.widgets.editors/slider.tests.js
```diff
@@ -1761,3 +1764,65 @@ module('if only the single value is possible', moduleOptions, () => {
...
+        this.pointer.swipe(20);
+
+        assert.strictEqual(this.getTooltipText(), '70');
+
+        this.pointer.swipeEnd(40);
+
+        assert.strictEqual(this.getTooltipText(), '90');
+    });
+
+    test('slider should change on every step after runtime change callValueChange to onMoving', function(assert) {
```
ksercs:
```suggestion
    test('slider should change its value on every step after runtime change callValueChange to onMoving', function(assert) {
```

ksercs: what about invert scenario? When we change `cellValueChange` to `onMovingComplete` at runtime?

ksercs (changes requested):

ksercs:
@dayanch-n I found a broken scenario: tooltip text is not updated on value change by keyboard navigation (left/right arrows) in `onMoving` mode. Let's add tests and fix this scenario.

js/ui/slider/ui.slider.js
```diff
@@ -380,16 +381,34 @@ const Slider = TrackBar.inherit({
...
     _renderStartHandler: function() {
         const pointerDownEventName = addNamespace(pointerEvents.down, this.NAME);
+        const pointerUpEventName = addNamespace(pointerEvents.up, this.NAME);
         const clickEventName = addNamespace(clickName, this.NAME);
         const startAction = this._createAction(this._startHandler.bind(this));
         const $element = this.$element();
 
+        const processOnMovingComplete = () => {
+            if(this.option('callValueChange') === 'onMovingComplete') {
+                if(this._getActualValue()) {
```
ksercs: seems like this condition is always `true` now

js/ui/slider/ui.slider.js
ksercs: Checked on iphone - works ok

ksercs (approved):

EugeniyKiyashko (approved):

PR #22319, opened 2022-08-09
Overlay: support promise in onShowing/onHiding e.cancel

Zedwag:


js/ui/overlay/ui.overlay.js
```diff
@@ -537,23 +560,32 @@ const Overlay = Widget.inherit({
...
+            const applyHide = () => {
                 this._forceFocusLost();
                 this._toggleShading(false);
                 this._toggleSubscriptions(false);
                 this._stopShowTimer();
 
                 this._animateHiding();
+            };
+
+            if(hidingArgs.cancel.then) {
```
ksercs: let's use `isPromise` function from `core/utils/type`

js/ui/overlay/ui.overlay.js
```diff
@@ -537,23 +560,32 @@ const Overlay = Widget.inherit({
...
+                hidingArgs.cancel.then(value => value ? cancelHide() : applyHide(), () => applyHide());
```
ksercs: we can use `catch` instead of the second parameter here to make the code more readable

ksercs: Why so strange name, `value`? Let's make it more meaningful, `shouldCancel`, for example.

We should not use ternary operator if return value is not interesting to us.

js/ui/overlay/ui.overlay.js
ksercs: this code is very similar for showing and hiding, cannot we unify it?

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -780,31 +780,156 @@ testModule('visibility', moduleConfig, () => {
...
+            assert.ok(this.$overlay.is(':hidden')), 'overlay is hidden';
+            this.checkIsNotShown();
         });
-        const overlay = $overlay.dxOverlay('instance');
-        overlay.show();
 
-        assert.ok($overlay.is(':hidden')), 'overlay is hidden';
-        assert.notOk(onShown.called, 'onShown should not be called');
-    });
+        test('overlay should not be hidden if e.cancel == true in the onHidinging event handler', function(assert) {
```
ksercs:
```suggestion
        test('overlay should not be hidden if e.cancel == true in the onHiding event handler', function(assert) {
```

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -780,31 +780,156 @@ testModule('visibility', moduleConfig, () => {
...
+                assert.ok(this.$overlay.is(':hidden')), 'overlay is hidden';
+                assert.ok(this.onHidden.called, 'onHidden should be called');
+            };
+        }
+    }, () => {
+        test('overlay should not be shown if e.cancel == true in the onShowing event handler (T825865)', function(assert) {
+            this.overlay.option({ onShowing: e => e.cancel = true });
+            this.overlay.show();
+
+            assert.ok(this.$overlay.is(':hidden')), 'overlay is hidden';
```
ksercs: Here's a misstype with brackets.

Besides, why this assert is not inside `checkIsNotShown` function how it's done for all another similar functions?

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -780,31 +780,156 @@ testModule('visibility', moduleConfig, () => {
...
-            visible: true,
-            onHiding: e => e.cancel = true,
-            onHidden
+            this.checkIsNotHidden();
+        });
+
+        test('overlay should be shown after timeout if e.cancel == promise resolving false after timeout', function(assert) {
+            const done = assert.async();
+            this.overlay.option({
+                onShowing: e => e.cancel = new Promise((resolve) => { setTimeout(() => resolve(false), 0); }),
```
ksercs: please refactor all such places. It's hard to read this when it's written in a single line

js/ui/overlay/ui.overlay.js
```diff
@@ -1200,10 +1211,18 @@ const Overlay = Widget.inherit({
         this._animateDeferred = animateDeferred;
         this.option('visible', showing);
 
-        animateDeferred.promise().done(() => {
-            delete this._animateDeferred;
-            result.resolveWith(this, [this.option('visible')]);
-        });
+        animateDeferred.promise()
+            .done(
+                () => {
```
ksercs: why did u decide to start a function from a new line?

js/ui/overlay/ui.overlay.js
Zedwag: done

Zedwag: added `_processShowingHidingCancel()` function

js/ui/overlay/ui.overlay.js
Zedwag: ok

Zedwag: ok

testing/tests/DevExpress.ui.widgets/overlay.tests.js
Zedwag: because we actualy show the Overlay (but hide it via `visibility: hidden`) before we resolve cancel

js/ui/overlay/ui.overlay.js
Zedwag: I am an artist, i see it in this way!

testing/tests/DevExpress.ui.widgets/overlay.tests.js
Zedwag: done

js/ui/overlay/ui.overlay.js
```diff
@@ -1200,10 +1219,16 @@ const Overlay = Widget.inherit({
...
+        animateDeferred.promise()
+            .done(() => {
+                delete this._animateDeferred;
+                result.resolveWith(this, [this.option('visible')]);
+            })
+            .fail(() => {
+                delete this._animateDeferred;
+                result.reject();
+            }
+            );
```
ksercs: Please move this to the previous line

ksercs (approved):

EugeniyKiyashko (approved):

PR #22358, opened 2022-08-12
Fix argument type for viz/export() (T1108890)

GoodDayForSurf:

### DIFF
--- a/js/viz/export.d.ts
+++ b/js/viz/export.d.ts
@@ -1,10 +1,9 @@
-import DOMComponent from '../core/dom_component';
-
 import {
-    ExportFormat,
-    HorizontalAlignment,
-    VerticalAlignment,
+  ExportFormat,
+  HorizontalAlignment,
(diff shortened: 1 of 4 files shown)


js/viz/export.d.ts
```diff
@@ -22,7 +21,7 @@ export function exportFromMarkup(markup: string, options: { fileName?: string; f
  * @static
  * @public
  */
-export function exportWidgets(widgetInstances: Array<Array<DOMComponent>>): void;
+export function exportWidgets(widgetInstances: BaseWidget | Array<BaseWidget> | Array<Array<BaseWidget>>): void;
```
IlyaKhD: `BaseWidget` is an internal type and should be used in public declaration
We have to define a union, smth like this
```ts
/** @public */
export type ChartWidget = dxChart | dxBullet | dxBarGauge| ... ;
```

I'm really not sure about the name, it should be 'any charting widget'. Or "component".

PR #22366, opened 2022-08-15
ShadowDOM: grid task

AlexKamaev:


ksercs:
Ping me after tests are green plz

js/core/utils/dom.js
```diff
@@ -152,3 +152,11 @@ export const replaceWith = (element, newElement) => {
 
     return newElement;
 };
+
+export const isElementInDom = $element => {
+    const element = $element?.get(0);
+    const shadowHost = element?.getRootNode().host;
+    const elementOrShadowHost = shadowHost || element;
+
+    return !!$(elementOrShadowHost).closest(getWindow().document).length;
```
ksercs: this variable looks strange for me - its name repeats the operation.
```suggestion
    return !!$(shadowHost || element).closest(getWindow().document).length;
```

js/ui/grid_core/ui.grid_core.editor_factory.js
```diff
@@ -191,11 +191,25 @@ const EditorFactory = modules.ViewController.inherit({
...
 
         this._updateFocusHandler = this._updateFocusHandler || this.createAction(this._updateFocus.bind(this));
-        eventsEngine.on(domAdapter.getDocument(), UPDATE_FOCUS_EVENTS, this._updateFocusHandler);
+        eventsEngine.on(this._getContainerRoot(), UPDATE_FOCUS_EVENTS, this._updateFocusHandler);
 
         this._attachContainerEventHandlers();
     },
 
+    _getContainerRoot: function() {
+        const $container = this.component && this.component.$element();
```
ksercs:
```suggestion
        const $container = this.component?.$element();
```

testing/tests/DevExpress.ui.widgets.dataGrid/filterRow.tests.js
```diff
@@ -2625,9 +2626,9 @@ QUnit.module('Filter Row with real dataController and columnsController', {
                 assert.strictEqual(loadSpy.getCall(0).args[0].skip, undefined);
             }
 
-            const dropDownList1 = $('.dx-list:eq(0)');
+            const dropDownList1 = $('.dx-list').eq(0);
             assert.strictEqual(dropDownList1.find('.dx-item').length, 91); // 90 rows + (All)
-            assert.strictEqual(dropDownList1.find('.dx-item:eq(1)').text(), 'value10');
+            assert.strictEqual(dropDownList1.find('.dx-item').eq(1).text(), 'value10');
             assert.strictEqual(dropDownList1.find('.dx-item:eq(-1)').text(), 'value99');
```
ksercs, line 2648: Why are not all such selectors refactored?

testing/tests/DevExpress.ui.widgets.treeList/selection.tests.js
```diff
@@ -8,6 +10,8 @@ QUnit.testStart(function() {
 ';
 
     $('#qunit-fixture').html(markup);
+
+    addShadowDomStyles($('#qunit-fixture'));
```
ksercs, line 14: Why do we need additionally attach shadow dom styles? I thought it's made on component level

testing/tests/DevExpress.ui.widgets.dataGrid/filterRow.tests.js
AlexKamaev: We need to discuss this moment. Here and in some other places we use not components themselves, but this thing:
`this.rowsView.render($testElement);`
so, the `render` function is not called. Thats my I added styles manually, because I thought that the styles should be added only inside the `render` function of real component

testing/tests/DevExpress.ui.widgets.treeList/selection.tests.js
AlexKamaev: We need to discuss this moment. Here and in some other places we use not components themselves, but this thing:
`this.rowsView.render($testElement);`
so, the `render` function is not called. Thats my I added styles manually, because I thought that the styles should be added only inside the `render` function of real component

testing/tests/DevExpress.ui.widgets.dataGrid/filterRow.tests.js
AlexKamaev: Exactly in this case it looks like auto-replace. I can revert changes for the 2631 line.
However, I need  to replace only root selectors like this one:
instead of $('div:eq(0)')
I write $('div').eq(0) and so on.
It's because of implementation of my custom JQuery init function, which I do not want to make more complex

ksercs (dismissed):

pomahtri (approved):

ksercs (approved):

Alyar666 (approved):

PR #22367, opened 2022-08-15
Fix export argument type (T1108890)

GoodDayForSurf:

### DIFF
--- a/js/viz/core/base_widget.d.ts
+++ b/js/viz/core/base_widget.d.ts
@@ -1,6 +1,5 @@
 import DOMComponent, {
   DOMComponentOptions,
-  Properties,
 } from '../../core/dom_component';
 
 import {
@@ -558,7 +557,7 @@ export interface BaseWidgetTooltip {
  * @inherits DOMComponent
  * @namespace DevExpress.viz
  */
-export default class BaseWidget<TProperties = Properties> extends DOMComponent<TProperties> {
+export default class BaseWidget<TProperties> extends DOMComponent<TProperties> {
     /**
      * @docid
      * @static
--- a/js/viz/core/base_widget.d.ts
+++ b/js/viz/core/base_widget.d.ts
@@ -1,5 +1,5 @@
 import DOMComponent, {
-  DOMComponentOptions,
+    DOMComponentOptions,
 } from '../../core/dom_component';
 
 import {
--- a/js/viz/core/base_widget.d.ts
+++ b/js/viz/core/base_widget.d.ts
@@ -1,5 +1,6 @@
 import DOMComponent, {
-    DOMComponentOptions,
+  DOMComponentOptions,
+  Properties,
 } from '../../core/dom_component';
 
 import {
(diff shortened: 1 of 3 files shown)


js/viz/core/base_widget.d.ts
```diff
@@ -557,7 +558,7 @@ export interface BaseWidgetTooltip {
  * @inherits DOMComponent
  * @namespace DevExpress.viz
  */
-export default class BaseWidget<TProperties> extends DOMComponent<TProperties> {
+export default class BaseWidget<TProperties = Properties> extends DOMComponent<TProperties> {
```
IlyaKhD: Is this change still required?
`BaseWidget` is not a public type and is used on rare occasions, it could have no default parameter

js/viz/export.d.ts
```diff
@@ -1,11 +1,24 @@
...
+import dxSankey from './sankey';
+import dxTreeMap from './tree_map';
+import dxVectorMap from './vector_map';
+import dxSparkline from './sparkline';
+import dxBullet from './bullet';
+import dxBarGauge from './bar_gauge';
+
+/**
+ * @docid
+ * @public */
```
IlyaKhD: Format:
```suggestion
/**
 * @docid
 * @public
 */
```

IlyaKhD: About this '@docid' - actually we could remove it. 
It depends whether we want to have a special topic for `VizWidget`, or use a kind of inline description
/cc @vladaskorohodova

GoodDayForSurf: without @docid validate-declarations does not pass

js/viz/core/base_widget.d.ts
```diff
@@ -1,5 +1,5 @@
 import DOMComponent, {
-    DOMComponentOptions,
+  DOMComponentOptions,
```
IlyaKhD: Unnecessary change

IlyaKhD (dismissed):

js/viz/export.d.ts
```diff
@@ -1,11 +1,24 @@
...
 
+import dxChart from './chart';
+import dxPieChart from './pie_chart';
+import dxFunnel from './funnel';
+import dxSankey from './sankey';
+import dxTreeMap from './tree_map';
+import dxVectorMap from './vector_map';
+import dxSparkline from './sparkline';
+import dxBullet from './bullet';
+import dxBarGauge from './bar_gauge';
```
IlyaKhD: Use `import type` to remove imports in run-time:

```suggestion
import type dxChart from './chart';
import type dxPieChart from './pie_chart';
import type dxFunnel from './funnel';
import type dxSankey from './sankey';
import type dxTreeMap from './tree_map';
import type dxVectorMap from './vector_map';
import type dxSparkline from './sparkline';
import type dxBullet from './bullet';
import type dxBarGauge from './bar_gauge';
```

IlyaKhD (approved):

PR #22383, opened 2022-08-17
LoadPanel: set propagateOutsideClick: true by default

dayanch-n:


js/ui/load_panel.js
```diff
@@ -54,28 +54,9 @@ const LoadPanel = Overlay.inherit({
...
-            * @hidden
-            */
-            resizeEnabled: false,
-
-            focusStateEnabled: false
-
-            /**
-            * @name dxLoadPanelOptions.dragEnabled
-            * @hidden
-            */
```
ksercs: Why did u remove this default values and doccoments? This changes seem to be excess

ksercs (changes requested):

js/ui/load_panel.js
dayanch-n: I have compared with upstream 21_2 and those were not there. So I thought it was outdated.

dayanch-n: Should I add them back?

Zedwag: https://github.com/DevExpress/DevExtreme/blob/21_2/js/ui/load_panel.js#L66 it exsist in 21_2

Zedwag (approved):

ksercs (approved):

PR #22392, opened 2022-08-18
dxScheduler - fixes agenda throws a TypeError when changing the data source of resources(T1108755)

xSev07:


js/ui/scheduler/ui.scheduler.js
```diff
@@ -566,7 +566,7 @@ class Scheduler extends Widget {
                 break;
             case 'resources':
                 this._dataAccessors.resources = createExpressions(this.option('resources'));
-                this.agendaResourceProcessor.initializeState(value);
```
AntonSermyazhko, line 569: The `resources` option is requested twice in this block. So, I suggest using a constant for it.

AntonSermyazhko (approved):

PR #22411, opened 2022-08-22
Scheduler: hides time zones for allday appt

xSev07:
if allowed to edit time zones they hides in all day's appointments when form will update.
Refactor appointment popup test model and tests


xSev07:
wrong branch

PR #22423, opened 2022-08-23
Scheduler: hides time zones for allday appt

xSev07:
if allowed to edit time zones they hides in all day's appointments when form will update.
Refactor appointment popup test model and tests

### DIFF
--- a/js/ui/scheduler/appointmentPopup/popup.js
+++ b/js/ui/scheduler/appointmentPopup/popup.js
@@ -13,6 +13,7 @@ import {
     getMaxWidth,
     getPopupToolbarItems,
 } from '../../../renovation/ui/scheduler/appointment_edit_form/popup_config';
+import { APPOINTMENT_FORM_GROUP_NAMES } from './form';
 
 const toMs = dateUtils.dateToMilliseconds;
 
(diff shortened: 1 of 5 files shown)


testing/testcafe/model/scheduler/appointment/popup.ts
```diff
@@ -13,6 +13,24 @@ export const CLASS = {
...
+interface ISchedulerPopupExpr {
+  subject?: string;
+  startDate?: string;
+  startDateTimeZone?: string;
+  endDate?: string;
+  endDateTimeZone?: string;
+  description?: string;
+}
+
+const defaultExpr = {
```
AntonSermyazhko: I think that `defaultExpr` should implement the `ISchedulerPopupExpr` interface.

js/ui/scheduler/appointmentPopup/popup.js
```diff
@@ -176,6 +176,14 @@ export class AppointmentPopup {
         );
     }
 
+    _setTimeZonesVisibility() {
+        if(this.scheduler.getEditingConfig().allowTimeZoneEditing) {
```
AntonSermyazhko, line 181: `this.scheduler` is a group of options for the Scheduler appointment popup and form. It'll be cool to rename this variable to the `this.config` e.g.
We can make this change in a separate PR.

js/ui/scheduler/appointmentPopup/popup.js
```diff
@@ -176,6 +176,14 @@ export class AppointmentPopup {
...
+            const isVisible = !this.state.appointment.data.allDay;
+            this.form.form.itemOption('undefined.startDateTimeZone', 'visible', isVisible);
+            this.form.form.itemOption('undefined.endDateTimeZone', 'visible', isVisible);
```
AntonSermyazhko: It's better to use this.form.[dxForm](https://github.com/DevExpress/DevExtreme/blob/22_2/js/ui/scheduler/appointmentPopup/form.js#L73) instead of this.form.**form**

PR #22625, opened 2022-09-13
Toolbar: get the same appearance for buttonGroup item when item has been initialized via template

EugeniyKiyashko:


marker-dao (approved):

scss/widgets/base/_toolbar.scss
```diff
@@ -182,35 +210,3 @@
 .dx-toolbar-item-invisible {
   display: none;
 }
-
-.dx-toolbar-hidden-button-group {
-  &.dx-state-hover {
-    background-color: transparent;
-  }
-
-  .dx-list-item-content {
```
ksercs, line 191: is it ok that this styles are removed at all?

Zedwag (approved):

scss/widgets/base/_toolbar.scss
EugeniyKiyashko: It is ok. The test-cafe screenshots approve it. If I right remember this styles are redundant for new hierarchy

ksercs (approved):

PR #22626, opened 2022-09-13
DropDownButton: The list of items doesn't appear in the Toolbar's menu if the toolbar item's 'locateInMenu' is 'auto' (T1102197)

EugeniyKiyashko:


Zedwag (approved):

marker-dao (approved):

ksercs (approved):
I gonna trust u that change of text editor's default width in menu is too small change to regard it as a breaking.

PR #22676, opened 2022-09-16
DataGrid - Show editing popup on init if editRowKey is specified (T1113974)

Tucchhaa:


js/ui/grid_core/ui.grid_core.editing_form_based.js
```diff
@@ -57,6 +57,14 @@ export const editingFormBasedModule = {
                     this._updateEditFormDeferred = null;
 
                     this.callBase.apply(this, arguments);
+
+                    this._rowsView.renderCompleted.add(() => {
+                        const rowIndex = this.getEditRowIndex();
+
+                        if(rowIndex !== -1 && this.isEditing() && this.isPopupEditMode()) {
```
pomahtri: you also need to add flag like `isFirstRender` to prevent calling `_showEditPopup` in case editRow is changed

js/ui/grid_core/ui.grid_core.editing_form_based.js
```diff
@@ -57,6 +57,14 @@ export const editingFormBasedModule = {
                     this._updateEditFormDeferred = null;
 
                     this.callBase.apply(this, arguments);
+
+                    this._rowsView.renderCompleted.add(() => {
```
pomahtri:
```suggestion
                    this._rowsView?.renderCompleted.add(() => {
```

tests are failed with `TypeError: Cannot read properties of undefined (reading 'renderCompleted')`, it seems that `_rowsView` isn't initialized in tests, so just add `?`

pomahtri (changes requested):

pomahtri (approved):

Alyar666 (approved):

PR #22710, opened 2022-09-23
Scheduler: Change etalons after chrome update.

williamvinogradov:
Replace the reference screenshots with a new ones after the chrome update.


ksercs (approved):
I approve skip of the test. We will fix it in the nearest days.

PR #22831, opened 2022-10-06
fix LGTM warnings

iBat:
1. What did you do? - fixed some LGTM warnings 
3. How did you do it? - removed unwanted code, accept some LGTM suggestions, ask other tribes help.
4. How should we verify this? - all test green, number of LGTM warnings reduced.

### DIFF
--- a/js/data/odata/utils.js
+++ b/js/data/odata/utils.js
@@ -324,8 +324,8 @@ const interpretJsonFormat = (obj, textStatus, transformOptions, ajaxOptions) => 
     }
 
     const value = 'd' in obj && (Array.isArray(obj.d) || isObject(obj.d))
-        ? interpretVerboseJsonFormat(obj, textStatus)
-        : interpretLightJsonFormat(obj, textStatus);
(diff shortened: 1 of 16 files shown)


js/data/odata/utils.js
```diff
@@ -324,8 +324,8 @@ const interpretJsonFormat = (obj, textStatus, transformOptions, ajaxOptions) =>
     }
 
     const value = 'd' in obj && (Array.isArray(obj.d) || isObject(obj.d))
-        ? interpretVerboseJsonFormat(obj, textStatus)
-        : interpretLightJsonFormat(obj, textStatus);
```
ivanblinov2k17, line 328: why textStatus is removed here? is behaviour the same without it?

iBat: Yes. Both functions interpretVerboseJsonFormat() and interpretLightJsonFormat() accepts only 1 argument.

dxvladislavvolkov (approved):

ivanblinov2k17 (approved):

PR #22833, opened 2022-10-06
Add types for permutations

IlyaKhD:
Cherry-pick of #22805


IlyaKhD:
I forget that v21_2 uses TS 3.1, so we're unable to use this solution. 
(Template Literal Types were added in TS 4.1)

PR #22871, opened 2022-10-11
dxScheduler - prevent trigger events on end dragging in corner cases(T1118059)

andrewmakarov:


js/ui/scheduler/compactAppointmentsHelper.js
```diff
@@ -119,7 +119,7 @@ export class CompactAppointmentsHelper {
                 getItemSettings,
             };
 
-            workSpace._createDragBehaviorBase($element, options);
+            workSpace._createDragBehaviorBase($element, $(this.instance.element()), options);
```
williamvinogradov: Maybe, we should move this parameter initialization upward.
Like this:
```js
const $schedulerElement = $(this.instance.element());
workSpace._createDragBehaviorBase($element, $schedulerElement, options);
```

js/ui/scheduler/workspaces/ui.scheduler.work_space.js
```diff
@@ -2037,13 +2037,15 @@ class SchedulerWorkSpace extends WidgetObserver {
         if(!this.dragBehavior && scheduler) {
             this.dragBehavior = new AppointmentDragBehavior(scheduler);
 
-            this._createDragBehavior(this.getWorkArea());
-            this._createDragBehavior(this.getAllDayContainer());
-            this._createDragBehavior(this._$allDayPanel);
+            const element = $(scheduler.element());
```
williamvinogradov: Following our code style, do we need to add the $ symbol to jQuery instances?
In my point of view, we need to add the $ symbol :)
Like this:
```js
const $element = $(scheduler.element());
```

testing/testcafe/tests/scheduler/dragAndDrop/T1118059.ts
```diff
@@ -0,0 +1,110 @@
...
+  await t
+    .dragToElement(scheduler.getAppointment('Regular test app').element, Selector('#drag-container'));
+
+  await t
+    .expect(Selector('#event-text-box').innerText)
+    .eql('onAppointmentDeleting');
+}).before(async () => createWidget('dxScheduler', {
+  dataSource: [{
+    text: 'All day test app 1',
+    startDate: new Date(2021, 3, 26),
```
williamvinogradov, line 24: Maybe, we should use the UTC format of the new Date in our tests.
If we do this, we've less trouble with timezones in the future.
Like this:
```js
startDate: new Date('2021-04-26T00:00:00Z'),
// or
startDate: '2021-04-26T00:00:00Z',
```

testing/testcafe/tests/scheduler/dragAndDrop/T1118059.ts
```diff
@@ -0,0 +1,110 @@
...
+  },
+}, true));
+
+test('After drag over component area, shouldn\'t called onAppointment* data events and appointment shouldn\'t change position', async (t) => {
+  const scheduler = new Scheduler('#container');
+
+  await t
+    .dragToElement(scheduler.getAppointment('All day test app 2').element, Selector('#space-right'));
+  await t
+    .expect(Selector('#event-text-box').innerText)
```
williamvinogradov: Let's move all string constants (like CSS selectors) into separate variables :)
E.g:
```js
const SCHEDULER_SELECTOR = '#container';
const EVENT_BOX_SELECTOR = '#event-text-box';

// other code.

const scheduler = new Scheduler(SCHEDULER_SELECTOR);

// other code.

await t.expect(Selector(EVENT_BOX_SELECTOR).innerText)

// other code.
```

williamvinogradov (dismissed):

williamvinogradov (approved):

PR #22938, opened 2022-10-19
Localization - Fix formatNumber throws maxFractionDigits value is out of range (T1122179)

ivanblinov2k17:


js/localization/intl/number.js
```diff
@@ -35,10 +36,12 @@ export default {
         let config;
 
         if(format === 'decimal') {
+            const fractionDigits = String(value).split('.')[1];
+            const maxFD = fractionDigits ? fractionDigits.length : undefined;
             config = {
                 minimumIntegerDigits: formatConfig.precision || undefined,
                 useGrouping: false,
-                maximumFractionDigits: String(value).length,
+                maximumFractionDigits: maxFD > MAX_FRACTION_DIGITS ? MAX_FRACTION_DIGITS : maxFD,
```
dxvladislavvolkov: Maybe something like this fractionDigits && fractionDigits.length?

dxvladislavvolkov (dismissed):

js/localization/intl/number.js
```diff
@@ -5,6 +5,7 @@ import openXmlCurrencyFormat from '../open_xml_currency_format';
 import accountingFormats from '../cldr-data/accounting_formats';
 
 const CURRENCY_STYLES = ['standard', 'accounting'];
+const MAX_FRACTION_DIGITS = 20;
```
MikeVitik, line 8: Why do we use two constants with the same name and value?

testing/tests/DevExpress.localization/sharedParts/localization.shared.js
```diff
@@ -624,6 +624,7 @@ export default function() {
             assert.equal(localization.formatNumber(1, { type: 'decimal', precision: 2 }), '01');
             assert.equal(localization.formatNumber(1, { type: 'decimal', precision: 3 }), '001');
             assert.equal(localization.formatNumber(1.23456, { type: 'decimal' }), '1.23456');
+            assert.equal(localization.formatNumber(-0.004768895486559899, { type: 'decimal' }), '-0.004768895486559899');
```
MikeVitik: replace: -0.004768895486559899
to: -0.0123456789012345678

MikeVitik (changes requested):

testing/tests/DevExpress.localization/sharedParts/localization.shared.js
ivanblinov2k17: Had the same idea, but this number gets auto rounded to 
-0.01...568

js/localization/intl/number.js
ivanblinov2k17: Because it's the same spec boundary
Tried to import it from 1 file to another - we have some legacy imports in this code and many tests got failing

MikeVitik (approved):

dxvladislavvolkov (approved):

PR #23044, opened 2022-10-31
Slider, RangeSlider: rename callValueChange with valueChangeMode

ksercs:


EugeniyKiyashko (dismissed):

js/common.d.ts
```diff
@@ -288,6 +288,12 @@ export type ValidationStatus = 'valid' | 'invalid' | 'pending';
  */
 export type ValueChangedCallMode = 'onMoving' | 'onMovingComplete';
 
+/**
+ * @public
+ * @namespace DevExpress.common
+ */
+ export type ValueChangeMode = 'onHandleMove' | 'onHandleRelease';
```
IlyaKhD: We should bring this name into line with recently added `ValueChangedCallMode` (we can rename it as well)
/cc @vladaskorohodova could you take a look?

IlyaKhD (changes requested):

js/common.d.ts
ksercs: This types cannot be merged because they are used in different components. We will rename `ValueChangedCallMode` in RangeSelector later.

Discussed it in a voice.

js/ui/range_slider.d.ts
```diff
@@ -11,10 +11,15 @@ import {
...
     dxSliderBaseOptions,
+    ValueChangeMode,
 } from './slider';
 
 import dxTrackBar from './track_bar';
 
+export {
+    ValueChangeMode,
+};
+
```
IlyaKhD:
```suggestion
export {
    ValueChangeMode,
} from './slider';

```

js/ui/range_slider.d.ts
```diff
@@ -11,10 +11,15 @@ import {
 
 import {
     dxSliderBaseOptions,
+    ValueChangeMode,
```
IlyaKhD:
```suggestion
```

IlyaKhD (approved):

EugeniyKiyashko (approved):

PR #23063, opened 2022-11-02
BarGauge: make resolveLabelOverlapping to support "shift" mode

EvgeniiMorozovDevSoft:


js/viz/gauges/bar_gauge.js
```diff
@@ -252,26 +254,26 @@ export const dxBarGauge = BaseGauge.inherit({
...
+            return getVerticallyShiftedAngularCoords(box, -length, that._context);
+        }
+
         if(overlapStrategy === 'none') {
             return;
         }
+        if(overlapStrategy === 'shift') {
+            const titleCoords = that._title.getLayoutOptions() ? that._title.getLayoutOptions() : { x: 0, y: 0, height: 0, width: 0 };
+            const newBars = dividePoints(bars);
+            let overlapped = false;
```
Krijovnick: Is `overlapped` variable is needed?

js/viz/gauges/bar_gauge.js
```diff
@@ -252,26 +254,26 @@ export const dxBarGauge = BaseGauge.inherit({
...
 
+        function shiftFunction(box, length) {
+            return getVerticallyShiftedAngularCoords(box, -length, that._context);
+        }
+
         if(overlapStrategy === 'none') {
             return;
         }
+        if(overlapStrategy === 'shift') {
+            const titleCoords = that._title.getLayoutOptions() ? that._title.getLayoutOptions() : { x: 0, y: 0, height: 0, width: 0 };
```
Krijovnick: getLayoutOptions calls two times. And title is always exist?

js/viz/core/renderers/renderer.js
```diff
@@ -1992,6 +1938,7 @@ Renderer.prototype = {
     // TODO animate end angle special case
     arc: function(x, y, innerRadius, outerRadius, startAngle, endAngle) {
         const elem = new ArcSvgElement(this);
+
```
Krijovnick: extra line

js/viz/gauges/bar_gauge.js
```diff
@@ -252,26 +254,26 @@ export const dxBarGauge = BaseGauge.inherit({
...
+            drawConnector(bars);
 
-            if(current.checkIntersect(next)) {
-                next.hideLabel();
-                nextIndex++;
-            } else {
-                currentIndex = nextIndex;
-                nextIndex = currentIndex + 1;
-            }
+            return;
```
Krijovnick: Remove return and add block else after that block. Otherwise it is hard to read the code and `return` word can be overlooked

js/viz/gauges/bar_gauge.js
```diff
@@ -252,26 +254,26 @@ export const dxBarGauge = BaseGauge.inherit({
...
         }
+        cleanLabels(bars);
```
Krijovnick: I think name 'cleanLabels' is not correct. We dont remove all labels, we hide some of them

js/viz/gauges/bar_guage/resolve_label_overlapping.js
```diff
@@ -0,0 +1,107 @@
...
+        const box = bar._text.getBBox();
+        const lastCoords = bar._text._lastCoords ? bar._text._lastCoords : { x: 0, y: 0 };
+        if(minY > box.y + lastCoords.y) {
+            bar.hideLabel();
+            return;
+        }
+    });
+};
+
+const ARC_COORD_PREC = 5;
```
Krijovnick: const variabels should be on the top of file

js/viz/gauges/bar_guage/resolve_label_overlapping.js
```diff
@@ -0,0 +1,107 @@
...
+        const y = bar._bar.attr('y');
+        const innerRadius = bar._bar.attr('innerRadius');
+        const outerRadius = bar._bar.attr('outerRadius');
+        const startAngle = bar._bar.attr('startAngle');
+        const endAngle = bar._bar.attr('endAngle');
+        const coordStart = getStartCoordsArc.apply(null, normalizeArcParams(x, y, innerRadius, outerRadius, startAngle, endAngle));
+        const xStart = Number(coordStart.x);
+        const yStart = Number(coordStart.y);
+        const box = bar._text.getBBox();
+        const lastCoords = bar._text._lastCoords ? bar._text._lastCoords : { x: 0, y: 0 };
```
Krijovnick: Сhecking for the existing of lastCoords is in many places. On initialization we can set they to { x: 0, y: 0 }

js/viz/gauges/bar_guage/resolve_label_overlapping.js
```diff
@@ -0,0 +1,107 @@
...
+        if(bar._angle > 90) {
+            bar._line.attr({ points: [
+                xStart + 1, yStart + 1, box.x + box.width + lastCoords.x, box.y + box.height / 2 + lastCoords.y
```
Krijovnick: Why 1 in this, what it is? Why not the half of the connector width?

js/viz/gauges/bar_guage/resolve_label_overlapping.js
```diff
@@ -0,0 +1,107 @@
...
+
+            ], });
+        } else if(bar._angle < 90) {
+            bar._line.attr({ points: [
+                xStart - 1, yStart - 1, box.x + lastCoords.x, box.y + box.height / 2 + lastCoords.y
+
+            ], });
```
Krijovnick: unnecessary comma

js/viz/gauges/bar_guage/resolve_label_overlapping.js
```diff
@@ -0,0 +1,107 @@
...
+        const lastCoords = bar._text._lastCoords ? bar._text._lastCoords : { x: 0, y: 0 };
+        if(bar._angle > 90) {
+            bar._line.attr({ points: [
+                xStart + 1, yStart + 1, box.x + box.width + lastCoords.x, box.y + box.height / 2 + lastCoords.y
+
+            ], });
+        } else if(bar._angle < 90) {
+            bar._line.attr({ points: [
+                xStart - 1, yStart - 1, box.x + lastCoords.x, box.y + box.height / 2 + lastCoords.y
+
```
Krijovnick: unnecessary line

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
```diff
@@ -574,6 +574,29 @@ QUnit.test('Some values are not changed', function(assert) {
     }, this);
 });
 
+QUnit.test('resolveLabelOverlapping - shift', function(assert) {
```
Krijovnick: No tests on the follow cases:
1. Hide labels that cross the title, also if title is not set
2. Hide labels that not fit to the canvas
3. Check labels from left and right ( < 90 and > 90 degrees)

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
```diff
@@ -574,6 +574,29 @@ QUnit.test('Some values are not changed', function(assert) {
...
+        resolveLabelOverlapping: 'shift'
+    });
+    const group = this.getBarsGroup();
+
+    group.animationComplete = $.proxy(function() {
+        assert.deepEqual(group.children[2]._stored_settings.points, [64.04786, 219.49845, 21, 7], 'line 1 coords');
+        assert.equal(group.children[3]._stored_settings.x.toFixed(3), 	34.027, 'text 1 is coord x');
+        assert.equal(group.children[3]._stored_settings.y.toFixed(3), 222.679, 'text 1 is coord y');
+        assert.deepEqual(group.children[6]._stored_settings.points, [95.86324, 203.76216, 12, 17], 'line 2 coords');
+        assert.equal(group.children[7]._stored_settings.x.toFixed(3), 31.813, 'text 1 is coord x');
```
Krijovnick: `text 1` is used on every check. There is should be `text 2`

js/viz/gauges/bar_guage/resolve_label_overlapping.js
```diff
@@ -0,0 +1,107 @@
...
+        return r;
+    }, { left: [], right: [] });
+}
+
+export const cleanLabelsForShift = (bars, minY)=> {
+    bars.forEach(bar => {
+        const box = bar._text.getBBox();
+        const lastCoords = bar._text._lastCoords ? bar._text._lastCoords : { x: 0, y: 0 };
+        if(minY > box.y + lastCoords.y) {
+            bar.hideLabel();
```
alexanderPolosatov: remove this return

alexanderPolosatov (changes requested):

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,105 @@
...
+        const startAngle = bar._bar.attr('startAngle');
+        const endAngle = bar._bar.attr('endAngle');
+        const coordStart = getStartCoordsArc.apply(null, normalizeArcParams(x, y, innerRadius, outerRadius, startAngle, endAngle));
+        const xStart = Number(coordStart.x);
+        const yStart = Number(coordStart.y);
+        const box = bar._text.getBBox();
+        const shiftConnector = connectorWidth / 2;
+        const lastCoords = bar._text._lastCoords;
+
+
```
EvgeniiMorozovDevSoft: delite line

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
```diff
@@ -574,6 +575,106 @@ QUnit.test('Some values are not changed', function(assert) {
...
+        assert.equal(group.children[7]._stored_settings.x.toFixed(3), 370.593, 'text 2 is coord x');
+        assert.equal(group.children[7]._stored_settings.y.toFixed(3), 128.743, 'text 2 is coord y');
+        assert.deepEqual(group.children[10]._stored_settings.points, [ 270.67515, 156.59695, -16, 27], 'line 3 coords');
+        assert.equal(group.children[11]._stored_settings.x.toFixed(3), 368.848, 'text 3 is coord x');
+        assert.equal(group.children[11]._stored_settings.y.toFixed(3), 121.215, 'text 3 is coord y');
+        done();
+    }, this);
+});
+
+
```
EvgeniiMorozovDevSoft: delite line

alexanderPolosatov (approved):

alexanderPolosatov (dismissed):

alexanderPolosatov (dismissed):

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,104 @@
+const ARC_COORD_PREC = 5;
+
+import { normalizeAngle, normalizeArcParams } from '../../core/utils';
+
+export function dividePoints(bars) {
+    return bars.reduce(function(r, bar) {
```
ksercs: Le't rename `r` to make the name more meaningful

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,104 @@
...
+        const angle = normalizeAngle(bar._angle);
+        bar._text._lastCoords = { x: 0, y: 0 };
+        (angle <= 90 || angle >= 270 ? r.right : r.left)
```
ksercs: Let's save this value to a separate variable to make the code clearer

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,104 @@
...
+            .push({
+                series: {
+                    isStackedSeries: ()=> false,
```
ksercs: Are lint rules different for charts' code? I think should look like this everywhere
```suggestion
                    isStackedSeries: () => false,
```

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,104 @@
...
+                getLabels: ()=> [{
+                    isVisible: ()=> true,
+                    getBoundingRect: () => {
+                        const { height, width, x, y } = bar._text.getBBox();
+                        const lastCoords = bar._text._lastCoords;
+
+                        return {
+                            x: x + lastCoords.x,
+                            y: y + lastCoords.y,
+                            width: width,
```
ksercs:
```suggestion
                            width,
```

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,104 @@
...
+        const lastCoords = bar._text._lastCoords;
+        if(minY > box.y + lastCoords.y) {
+
+            bar.hideLabel();
+        }
+    });
+};
+
+const getStartCoordsArc = function(x, y, innerR, outerR, startAngleCos, startAngleSin, endAngleCos, endAngleSin, isCircle, longFlag) {
+    return { x: (x + outerR * startAngleCos).toFixed(ARC_COORD_PREC), y: (y - outerR * startAngleSin).toFixed(ARC_COORD_PREC) };
```
ksercs: It's better to write it in several rows but make the code easier to understand

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,104 @@
...
+        const box = bar._text.getBBox();
+        const lastCoords = bar._text._lastCoords;
+        if(minY > box.y + lastCoords.y) {
+
+            bar.hideLabel();
+        }
+    });
+};
+
+const getStartCoordsArc = function(x, y, innerR, outerR, startAngleCos, startAngleSin, endAngleCos, endAngleSin, isCircle, longFlag) {
```
ksercs: Oh, so many parameters. Shouldn't we pass a configuration object instead of this? https://refactoring.guru/introduce-parameter-object

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const drawConnector = (bars, connectorWidth) => {
+    bars.forEach((bar) => {
+        const x = bar._bar.attr('x');
+        const y = bar._bar.attr('y');
+        const innerRadius = bar._bar.attr('innerRadius');
+        const outerRadius = bar._bar.attr('outerRadius');
+        const startAngle = bar._bar.attr('startAngle');
+        const endAngle = bar._bar.attr('endAngle');
```
ksercs, line 66: I think we can use distructurization here instead of so many rows

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,104 @@
...
+        const coordStart = getStartCoordsArc.apply(null, normalizeArcParams(x, y, innerRadius, outerRadius, startAngle, endAngle));
+        const xStart = Number(coordStart.x);
+        const yStart = Number(coordStart.y);
+        const box = bar._text.getBBox();
+        const shiftConnector = connectorWidth / 2;
+        const lastCoords = bar._text._lastCoords;
+
+        if(bar._angle > 90) {
+            bar._line.attr({ points: [
+                xStart + shiftConnector, yStart + shiftConnector, box.x + box.width + lastCoords.x, box.y + box.height / 2 + lastCoords.y
```
ksercs: 1. It's not necessary to do it on the single row.
2. Seems like this points and points on the `else` part differs only by the first point. Can we add all other after this confition?

js/viz/bar_gauge.d.ts
```diff
@@ -89,6 +89,8 @@ export interface TooltipInfo {
     target?: any;
 }
 
+/** @public */
+export type BarGaugeLabelOverlap = 'hide' | 'none' | 'shift';
```
ksercs: Do we really need introduce a new separate type for the BarGaude? Seems like this type should be used also in another components (PieChart, for example).

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
```diff
@@ -574,6 +575,104 @@ QUnit.test('Some values are not changed', function(assert) {
...
+QUnit.test('label location if the angle is < 90 and resolveLabelOverlapping - shift', function(assert) {
+    const done = assert.async();
+    this.$container.dxBarGauge({
+        values: [10, 11, 12],
+        resolveLabelOverlapping: 'shift'
+    });
+    const group = this.getBarsGroup();
+
+    group.animationComplete = $.proxy(function() {
+        assert.deepEqual(group.children[2]._stored_settings.points, [64.04786, 219.49845, 21, 7], 'line 1 coords');
```
ksercs, line 587: cannot we test it without a private api?
Then a suggest you to introduce a separate module for this tests and create a getter for this private api in `beforeEach` section to simplify possible refactoring if private api is changed

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
```diff
@@ -574,6 +575,104 @@ QUnit.test('Some values are not changed', function(assert) {
...
+        assert.deepEqual(group.children[6]._stored_settings.points, [305.73529, 152.89618, -8, 17], 'line 2 coords');
+        assert.equal(group.children[7]._stored_settings.x.toFixed(3), 370.593, 'text 2 is coord x');
+        assert.equal(group.children[7]._stored_settings.y.toFixed(3), 128.743, 'text 2 is coord y');
+        assert.deepEqual(group.children[10]._stored_settings.points, [ 270.67515, 156.59695, -16, 27], 'line 3 coords');
+        assert.equal(group.children[11]._stored_settings.x.toFixed(3), 368.848, 'text 3 is coord x');
+        assert.equal(group.children[11]._stored_settings.y.toFixed(3), 121.215, 'text 3 is coord y');
+        done();
+    }, this);
+});
+
```
ksercs, line 674: @EvgeniiMorozovDevSoft  it's a UI feature so I believe we should add a screenshot tests

js/viz/bar_gauge.d.ts
IlyaKhD: Let's try to generalize it.

common/charts module contains these:

```ts
export type ChartsAxisLabelOverlap = 'rotate' | 'stagger' | 'none' | 'hide';
export type ChartsLabelOverlap = 'hide' | 'none' | 'stack';
export type LabelOverlap = 'hide' | 'none';
```

Besides that there are:
```ts
...
export type FunnelLabelOverlap = 'hide' | 'none' | 'shift';
...
export type PieChartLabelOverlap = 'hide' | 'none' | 'shift';
...
export type CircularGaugeLabelOverlap = 'first' | 'last';
```

IlyaKhD: @vladaskorohodova, could you try to figure out how these types can be renamed so that we can move `'hide' | 'none' | 'shift'` to the common/charts module?

vladaskorohodova: Maybe `ChartsStackLabelOverlap` and `ChartsShiftLabelOverlap`? We could use something like `CommonChartsLabelOverlap` and `ExtraChartsLabelOverlap` but I think it doesn't really convey the difference. Anyway it's really hard to group Funnel, PieChart and BarGauge.

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
```diff
@@ -574,6 +575,104 @@ QUnit.test('Some values are not changed', function(assert) {
...
+    const done = assert.async();
+    this.$container.dxBarGauge({
+        values: [10, 11, 12],
+        resolveLabelOverlapping: 'shift'
+    });
+    const group = this.getBarsGroup();
+
+    group.animationComplete = $.proxy(function() {
+        assert.deepEqual(group.children[2]._stored_settings.points, [64.04786, 219.49845, 21, 7], 'line 1 coords');
+        assert.equal(group.children[3]._stored_settings.x.toFixed(3), 34.027, 'text 1 is coord x');
```
ksercs: Please, let's use `strictEqual` everywhere

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: No, they differ because of the calculation 
I made line breaks, now it is clear that only the "y" coordinate of the second point is the same

js/viz/gauges/bar_gauge.js
EvgeniiMorozovDevSoft: No, I removed it

js/viz/gauges/bar_gauge.js
EvgeniiMorozovDevSoft: removed the double call

js/viz/core/renderers/renderer.js
EvgeniiMorozovDevSoft: removed

js/viz/gauges/bar_gauge.js
EvgeniiMorozovDevSoft: renamed

js/viz/gauges/bar_guage/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: Fixed, shift is now half the thickness of the connector

js/viz/gauges/bar_guage/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: removed

js/viz/gauges/bar_guage/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: removed

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
EvgeniiMorozovDevSoft: Unit tests for these cases are written

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
EvgeniiMorozovDevSoft: fixed

js/viz/gauges/bar_guage/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: fixed

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: fixed

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
EvgeniiMorozovDevSoft: fixed

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: имя переменой изменено

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: I put the value in a variable

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: fixed

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: fixed

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: fixed

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
EvgeniiMorozovDevSoft: I partially agree with this, but here let's create a refactoring task and fix the tests in it, where a private api is used

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
EvgeniiMorozovDevSoft: At the moment, we have discussed that screenshot tests will not help us much if all the alleged cases are covered in the unit test

testing/tests/DevExpress.viz.gauges/barGauge.tests.js
EvgeniiMorozovDevSoft: Fixed for stricter verification

js/viz/bar_gauge.d.ts
EvgeniiMorozovDevSoft: if we allocate some separate type for Funnel, Pie Chart and Bar Gauge, we will only mislead people.
We already have a dedicated type in common/charts.d.ts Label Overlap, which is already used in several places .
Secondly, I don't see much point in combining their types into one constant due to the fact that within themselves these three components use different implementations for displaying labels

Krijovnick (dismissed):

js/viz/bar_gauge.d.ts
ksercs: > We already have a dedicated type in common/charts.d.ts Label Overlap

We don't have questions to `LabelOverlap` type, we just don't what maintain several different types describing the single enum.

> we allocate some separate type for Funnel, Pie Chart and Bar Gauge, we will only mislead people.

I don't understand why concrete type, for example like Vlada suggested - ChartsShiftLabelOverlap, will misslead users

>  I don't see much point in combining their types into one constant due to the fact that within themselves these three components use different implementations for displaying labels

Does not matter, there is a single implementation or several. 
Why several types for the single enum is bad:
1. users cannot save option value of the type and share it between Funnel/PieChart/BarGauge.
2. we wanna have consistent api. it's simpler to maintain it when u have only 1 type - u can change it and the change will affect all related components. U can think about this as about reduce of copypaste.

I think new common `export type ChartsShiftLabelOverlap = 'hide' | 'none' | 'shift';` is ok. But I think @IlyaKhD  should make a final decision.

js/viz/core/utils.js
```diff
@@ -444,6 +444,50 @@ export function getVizRangeObject(value) {
     }
 }
 
+export function normalizeArcParams(x, y, innerR, outerR, startAngle, endAngle) {
```
ksercs: What is `r`? Radius? Can we write a full name?

js/viz/core/utils.js
```diff
@@ -444,6 +444,50 @@ export function getVizRangeObject(value) {
...
+            endAngle = endAngle % 360;
+        }
+
+        if(startAngle > endAngle) {
+            startAngle -= 360;
+        }
+        noArc = false;
+    }
+
+    startAngle = startAngle * PI_DIV_180;
```
ksercs:
```suggestion
    startAngle *= PI_DIV_180;
```

js/viz/core/utils.js
```diff
@@ -444,6 +444,50 @@ export function getVizRangeObject(value) {
...
+    const angleDiff = roundValue(endAngle, 3) - roundValue(startAngle, 3);
+    if(angleDiff) {
+        if((abs(angleDiff) % 360) === 0) {
+            startAngle = 0;
+            endAngle = 360;
+            isCircle = true;
+            endAngle -= 0.01;
+        }
+
+        if(startAngle > 360) {
```
ksercs, line 459: why do we need this if? Seem like we can just write `startAngle %= 360;`

js/viz/core/utils.js
```diff
@@ -444,6 +444,50 @@ export function getVizRangeObject(value) {
...
+            startAngle = startAngle % 360;
```
ksercs, line 460: Let's move 360 to a constant?

js/viz/core/utils.js
```diff
@@ -444,6 +444,50 @@ export function getVizRangeObject(value) {
     }
 }
 
+export function normalizeArcParams(x, y, innerR, outerR, startAngle, endAngle) {
+    let isCircle;
+    let noArc = true;
+    const angleDiff = roundValue(endAngle, 3) - roundValue(startAngle, 3);
```
ksercs, line 450: seems like we can get `abs` here to make the following code clearer

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,115 @@
+const ARC_COORD_PREC = 5;
+
+import { normalizeAngle, normalizeArcParams } from '../../core/utils';
+
+export function dividePoints(bars) {
+    return bars.reduce(function(stackBars, bar) {
+        const angle = normalizeAngle(bar._angle);
```
ksercs: it's not ok that we use a private api in utils.
I believe we should pass only the necessary data to utils, not full instances because we wanna isolate the code that can mutate our instance's state.

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,115 @@
...
+        const isRightSide = angle <= 90 || angle >= 270;
+        bar._text._lastCoords = { x: 0, y: 0 };
+
+        (isRightSide ? stackBars.right : stackBars.left)
```
ksercs: let's move it to a variable `const barToExtend = isRightSide ? stackBars.right : stackBars.left;`

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,115 @@
...
+                            height,
+                        };
+                    },
+                    shift: (x, y) => {
+                        const box = bar._text.getBBox();
+
+                        bar._text._lastCoords = { x: x - box.x, y: y - box.y };
+                        bar._text.attr({ translateX: x - box.x, translateY: y - box.y });
+                    },
+                    draw: ()=> bar.hideLabel(),
```
ksercs: still strange style of arrow functions)
```suggestion
                    draw: () => bar.hideLabel(),
```

js/viz/gauges/utils/resolve_label_overlapping.js
```diff
@@ -0,0 +1,115 @@
...
+export const drawConnector = (bars, connectorWidth) => {
+    bars.forEach((bar) => {
+        const x = bar._bar.attr('x');
+        const y = bar._bar.attr('y');
+        const innerRadius = bar._bar.attr('innerRadius');
+        const outerRadius = bar._bar.attr('outerRadius');
+        const startAngle = bar._bar.attr('startAngle');
+        const endAngle = bar._bar.attr('endAngle');
+        const coordStart = getStartCoordsArc.apply(null, normalizeArcParams(x, y, innerRadius, outerRadius, startAngle, endAngle));
+        const xStart = Number(coordStart.x);
```
ksercs: what is this for?

EvgeniiMorozovDevSoft: If you're talking about type casting, it's because it happened historically and we return string format,

if why are we collecting a mountain of constants, then first it is necessary to calculate the origin of coordinates for drawing the connector in barGauge

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: An interesting point
Of course I'll fix it, but I'm wondering why lint didn't work automatically

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: Explain the moment of the problem, as I understand you, is that unclean functions are used that change the initial state of the object and because they have a separate export it becomes unsafe?

js/viz/core/utils.js
EvgeniiMorozovDevSoft: Apart from the fact that it was a migration of the old working code to another location, then I have a question.

What do you understand when you see the numbers 180, 360, 720 and so on? I understand this in the context of any circles, that this is an angle, if such things are taken out in a separate explanation in terms of a variable, then it seems to me it can confuse more

js/viz/core/utils.js
EvgeniiMorozovDevSoft: As for readability, I disagree.
what do we say that the angle is greater than the full rotation (360 degrees), or does the angle have the remainder of the division by 360

in the second case, they make me think that I need another remainder

js/viz/gauges/utils/resolve_label_overlapping.js
ksercs: Smth like this, yes. Utils should be a pure functions - they gets some data, calculates smth and returns a result. They shouldn't get another class instance as a prop and mutate it. If some function need access to other class' members this function should mostly be a method of this class :) We should not mutate instance's private members outside of the class implementation.

Please, call me if u need further assistance. U think it'll be faster to discuss it in voice

js/viz/core/utils.js
ksercs: > What do you understand when you see the numbers 180, 360, 720 and so on?

I don't wanna say it's impossible to understand the meaning of this number now. But i need to waste a little brain resource to realise what this variable mean.

180 is not only a half on a full angle, it's also number of Fahrenheit degree between freezing point and boiling point, angry number from "Final destination", max result in darts and so on.

See https://eslint.org/docs/latest/rules/no-magic-numbers

But I agree it's not so important now. We also do not want import/export/reinitialize this constant while it's used in many different files.

U can close this conversation, thanks.

js/viz/core/utils.js
ksercs: sorry i didn't catch. Text me directly plz

js/viz/core/utils.js
EvgeniiMorozovDevSoft: Renaming

js/viz/core/utils.js
EvgeniiMorozovDevSoft: Fixed

js/viz/bar_gauge.d.ts
EvgeniiMorozovDevSoft: Agreed with Vlada on the ChartsShiftLabelOverlap option

js/viz/core/utils.js
EvgeniiMorozovDevSoft: "roundValue" is also used for exponential numbers, and since this is the output of the old code, I don't want to change it in this place

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: fixed

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: Finished the discussion at the meeting

js/viz/gauges/utils/resolve_label_overlapping.js
EvgeniiMorozovDevSoft: Fixed

js/viz/funnel.d.ts
```diff
@@ -57,13 +58,14 @@ export {
...
     TextOverflow,
+    ShiftLabelOverlap,
     WordWrap,
 };
 
 /** @public */
 export type FunnelAlgorithm = 'dynamicHeight' | 'dynamicSlope';
 /** @public */
-export type FunnelLabelOverlap = 'hide' | 'none' | 'shift';
+export type FunnelLabelOverlap = ShiftLabelOverlap;
```
ksercs, line 68: I think we should just deprecate `FunnelLabelOverlap` and recommend using of `ShiftLabelOverlap`

js/viz/funnel.d.ts
```diff
@@ -57,13 +58,14 @@ export {
     Palette,
     PaletteExtensionMode,
     TextOverflow,
+    ShiftLabelOverlap,
     WordWrap,
 };
 
 /** @public */
 export type FunnelAlgorithm = 'dynamicHeight' | 'dynamicSlope';
 /** @public */
```
ksercs:
```suggestion
/**
 * @deprecated Use ShiftLabelOverlap from 'devextreme/common/charts' instead 
 */
```

We should deprecate old types `FunnelLabelOverlap`, `PieChartLabelOverlap` to notify users about change and recommend using of `ShiftLabelOverlap`

Do it for PieChart also please.

Let's also add a card for deprecation remove in 2 releases.

js/viz/funnel.d.ts
```diff
@@ -57,13 +58,14 @@ export {
...
-export type FunnelLabelOverlap = 'hide' | 'none' | 'shift';
+export type FunnelLabelOverlap = ShiftLabelOverlap;
```
ksercs, line 71: We should update "resolveLabelOverlapping" option's type to new ShiftLabelOverlap type. It will repaint the tests.

Do it for PieChart also please.

ts/dx.all.d.ts
```diff
@@ -31421,7 +31422,8 @@ declare module DevExpress.viz {
     export type OptionChangedEvent = DevExpress.events.EventInfo<dxPieChart> &
       DevExpress.events.ChangedOptionInfo;
     export type PieChartAnnotationLocation = 'center' | 'edge';
-    export type PieChartLabelOverlap = 'hide' | 'none' | 'shift';
+    export type PieChartLabelOverlap =
+      DevExpress.common.charts.ShiftLabelOverlap;
```
ksercs, line 31426: After apply of the previous suggestions we should add

```javascript
declare module DevExpress.common.charts {
  /** @deprecated Use DevExpress.common.charts.ShiftLabelOverlap instead */
  export type FunnelLabelOverlap = DevExpress.common.charts.ShiftLabelOverlap;
  /** @deprecated Use DevExpress.common.charts.ShiftLabelOverlap instead */
  export type PieChartLabelOverlap = DevExpress.common.charts.ShiftLabelOverlap;
}
```
to `/ts/aliases.d.ts` file
(Note that `module DevExpress.common.charts` already exists - we should only extend it)

EvgeniiMorozovDevSoft: done

js/viz/funnel.d.ts
EvgeniiMorozovDevSoft: done

js/viz/funnel.d.ts
```diff
@@ -57,13 +58,16 @@ export {
...
+    ShiftLabelOverlap,
     WordWrap,
 };
 
 /** @public */
 export type FunnelAlgorithm = 'dynamicHeight' | 'dynamicSlope';
-/** @public */
-export type FunnelLabelOverlap = 'hide' | 'none' | 'shift';
+/**
+ * @deprecated Use ShiftLabelOverlap instead
```
ksercs: please, make it exactly how it was suggested in the appropriate review note

js/viz/core/utils.js
ksercs: What does it mean "renaming"? I still see a strange abbreviations `innerR`, `outerR` although you close the conversation

Let's rename it to `innerRadius` and `outerRadius`

Krijovnick (dismissed):

js/viz/funnel.d.ts
EvgeniiMorozovDevSoft: done

js/viz/funnel.d.ts
EvgeniiMorozovDevSoft: done

js/common/charts.d.ts
```diff
@@ -64,6 +64,12 @@ export type HatchDirection = 'left' | 'none' | 'right';
  */
 export type LabelOverlap = 'hide' | 'none';
 
+/**
+ * @public
+ * @namespace DevExpress.common.charts
+ */
+export type ShiftLabelOverlap = 'hide' | 'none' | 'shift';
```
IlyaKhD, line 71: @vladaskorohodova are we OK with this name?
As for me there is something confusing in how the word 'shift' used: maybe 'shiftable' or 'shifted' fit better?
Smth like `ShiftableLabelOverlap` or `LabelShiftedOverlap`

ts/aliases.d.ts
```diff
@@ -421,6 +421,10 @@ declare module DevExpress.common.charts {
     export type StateStoreType = DevExpress.common.grids.StateStoreType;
     /** @deprecated Use DevExpress.common.grids.SummaryType instead */
     export type SummaryType = DevExpress.common.grids.SummaryType;
+    /** @deprecated Use DevExpress.common.charts.ShiftLabelOverlap instead */
+    export type FunnelLabelOverlap = DevExpress.common.charts.ShiftLabelOverlap;
+    /** @deprecated Use DevExpress.common.charts.ShiftLabelOverlap instead */
+    export type PieChartLabelOverlap = DevExpress.common.charts.ShiftLabelOverlap;
```
IlyaKhD: Parent namespace is incorrect:

```suggestion

declare module DevExpress.viz.dxFunnel {
    /** @deprecated Use DevExpress.common.charts.ShiftLabelOverlap instead */
    export type FunnelLabelOverlap = DevExpress.common.charts.ShiftLabelOverlap;
}

declare module DevExpress.viz.dxPieChart {
    /** @deprecated Use DevExpress.common.charts.ShiftLabelOverlap instead */
    export type PieChartLabelOverlap = DevExpress.common.charts.ShiftLabelOverlap;
}
```

IlyaKhD: One more thing: move these lines down under the `// v23.1` header:

```ts
// v23.1

declare module DevExpress.viz.dxFunnel {
...
```

IlyaKhD (changes requested):

ts/aliases.d.ts
EvgeniiMorozovDevSoft: Made changes

js/common/charts.d.ts
EvgeniiMorozovDevSoft: After discussing with Vlada, we left the name ShiftLabelOverlap

IlyaKhD (approved):

vladaskorohodova (approved):

Krijovnick (approved):

alexanderPolosatov (approved):

PR #23082, opened 2022-11-03
Tabs: improve the logic of adding 'dx-tabs-stretched' class in case when tabs as template is rendered asynch

EugeniyKiyashko:


const314 (approved):

ksercs (approved):
I gonna trust u about this

PR #23182, opened 2022-11-11
Overlay: remove warning for elementAttr

marker-dao:


EugeniyKiyashko:
We decided to think about this in v23.1

PR #23287, opened 2022-11-25
fix return type for CustomStore.load() (T1128989)

GoodDayForSurf:
### Cherry-picks
- #23461

### DIFF
--- a/js/data/abstract_store.d.ts
+++ b/js/data/abstract_store.d.ts
@@ -1,6 +1,14 @@
-import { DxPromise, DxExtendedPromise } from '../core/utils/deferred';
+// eslint-disable-next-line max-classes-per-file
+import {
+  DxPromise,
+  DxExtendedPromise,
+} from '../core/utils/deferred';
 import { DeepPartial } from '../core/index';
-import { FilterDescriptor, GroupDescriptor, LoadOptions } from './index';
+import {
+  FilterDescriptor,
+  GroupDescriptor,
+  LoadOptions,
+} from './index';
 
 export type Options<
     TItem = any,
@@ -107,11 +115,9 @@ export interface StoreOptions<
 type EventName = 'loaded' | 'loading' | 'inserted' | 'inserting' | 'updated' | 'updating' | 'push' | 'removed' | 'removing' | 'modified' | 'modifying';
 
 /**
- * @docid
- * @hidden
  * @namespace DevExpress.data
  */
-export default class Store<
+export class Store<
     TItem = any,
     TKey = any,
 > {
@@ -147,21 +153,6 @@ export default class Store<
      * @public
      */
     keyOf(obj: TItem): TKey;
-    /**
-     * @docid
-     * @publicName load()
-     * @return Promise<any>
-     * @public
-     */
-    load(): DxExtendedPromise<Array<TItem>>;
-    /**
-     * @docid
-     * @publicName load(options)
-     * @param1 options:LoadOptions
-     * @return Promise<any>
-     * @public
-     */
-    load(options: LoadOptions<TItem>): DxExtendedPromise<Array<TItem>>;
     /**
      * @docid
      * @publicName off(eventName)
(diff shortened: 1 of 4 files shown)


js/data/abstract_store.d.ts
```diff
@@ -114,6 +114,7 @@ type EventName = 'loaded' | 'loading' | 'inserted' | 'inserting' | 'updated' | '
 export default class Store<
     TItem = any,
     TKey = any,
+    TLoadRet = Array<TItem>,
```
IlyaKhD:
```suggestion
    TLoadResult = Array<TItem>,
```
or
```suggestion
    TLoaded = Array<TItem>,
```

IlyaKhD: ping @vladaskorohodova

js/data/abstract_store.d.ts
```diff
@@ -1,230 +1,243 @@
...
-    onUpdated?: ((key: TKey, values: TItem) => void);
-    /**
-     * @docid
-     * @type_function_param1 key:object|string|number
-     * @type_function_param2 values:object
-     * @action
-     * @public
-     */
-    onUpdating?: ((key: TKey, values: TItem) => void);
+  TItem = any,
```
IlyaKhD: Please remove unnecessary changes

testing/typescript/data/stores.ts
```diff
@@ -20,6 +21,17 @@ export async function infersTItemFromComplexLoadResult() {
...
 
+export async function loadReturnRightTypeIfLoadDefinedInOptions() {
+  const store = new CustomStore({
+    // eslint-disable-next-line @typescript-eslint/require-await
+    load: async () => ({
+      data: [{ a: 1 }],
+    }),
+  });
+
+  const value: ResolvedData<{ a: number }> = await store.load();
```
IlyaKhD: Do not use explicit type specification, it seems that it checks nothing in this case.

I think we should check  the item type here, smth like this:
```suggestion
  const loadResult = await store.load();

  const actualItem: (typeof loadResult) extends ResolvedData<infer Item> ? Item : never = ANY;

  notAny(actualItem);
  notNever(actualItem);
  assertType<{a: number}>(toAssertion(actualItem));
```

testing/typescript/data/stores.ts
```diff
@@ -20,6 +21,17 @@ export async function infersTItemFromComplexLoadResult() {
   const value: { a: number } = result;
 }
 
+export async function loadReturnRightTypeIfLoadDefinedInOptions() {
+  const store = new CustomStore({
+    // eslint-disable-next-line @typescript-eslint/require-await
+    load: async () => ({
+      data: [{ a: 1 }],
```
IlyaKhD: A test for custom loaded type required:
```suggestion
      data: { data: [{ ab: 1 }], extraField: 123},
```

E.g. such test would fail, because `result` is `any`:

```ts
  const store = new CustomStore({
    // eslint-disable-next-line @typescript-eslint/require-await
    load: async () => ({
      data: { data: [{ ab: 1 }], extraField: 123},
    }),
  });

  const result = await store.byKey(ANY);
  notAny(result);
```

testing/typescript/data/stores.ts
```diff
@@ -20,6 +27,24 @@ export async function infersTItemFromComplexLoadResult() {
...
+      totalCount: 1,
+    }),
+  });
+
+  const loadResult = await store.load();
+  const actualItem: (typeof loadResult) extends ResolvedData<infer Item> ? Item : never = ANY;
+
+  notAny(actualItem);
+  notNever(actualItem);
+  assertType<typeof dataItem>(toAssertion(actualItem));
```
IlyaKhD, line 45: Let's add one more case:

```suggestion
  notAny(loadItem);
  notNever(loadItem);
  assertType<typeof dataItem>(toAssertion(loadItem));

  const byKeyItem = await store.byKey(ANY);
  notAny(byKeyItem);
  notNever(byKeyItem);
  assertType<typeof dataItem>(toAssertion(byKeyItem));
```

IlyaKhD (approved):

PR #23333, opened 2022-12-01
Renovation: Fix case with defaultProps = undefined.

williamvinogradov:


const314 (dismissed):

js/renovation/component_wrapper/common/component.ts
```diff
@@ -566,6 +568,11 @@ export default class ComponentWrapper extends DOMComponent<ComponentWrapperProps
     this._aria[name] = value;
     this._initMarkup();
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  _getViewComponentDefaultProps(): any {
```
MikeVitik: Do you try to use Object instead of any? And if yes what is going wrong?

williamvinogradov: Agree with you
Changed it to:
```ts
_getViewComponentDefaultProps(): Record<PropertyKey, unknown> {
```

const314 (approved):

MikeVitik (approved):

PR #23360, opened 2022-12-05
DateBox: enter key should submit form when useMaskBehavior is enabled (T1131035)

Zedwag:


testing/tests/DevExpress.ui.widgets.editors/datebox.mask.tests.js
```diff
@@ -676,12 +672,26 @@ module('Keyboard navigation', setupModule, () => {
 
     test('enter should clear search value', function(assert) {
         this.keyboard.type('1');
+
         assert.strictEqual(this.instance.option('text'), 'January 10 2012', 'text has been changed');
+        assert.strictEqual(this.instance._searchValue, '1', 'there is searchValue');
```
ksercs: cannot we use input val instead of private property?

Zedwag: thought that better use private property if it make test more obvious
ok then, I returned the test like it was except of section moving part on enter key

ksercs (approved):

marker-dao (approved):

PR #23390, opened 2022-12-08
React changes for jquery

MikeVitik:


const314 (approved):

williamvinogradov (approved):

packages/react/src/internal/hooks/use-id.ts
```diff
@@ -0,0 +1,15 @@
...
+  const charactersLength = characters.length;
+  for (let i = 0; i < length; i += 1) {
+    result += characters.charAt(Math.floor(Math.random() * charactersLength));
+  }
+  return result;
+}
+export function useId() {
+  const id = useMemo(() => makeid(), []);
+  return id;
+}
```
IlyaKhD: How do we guarantee these IDs to be unique?


```suggestion
function createIdGenerator() {
  let counter = BigInt(0); 
  return () => (counter++).toString(36);
}

const  makeId = createIdGenerator();

export function useId() {
  const id = useMemo(() => makeId(), []);
  return id;
}
```

Btw, 62^5 is 916132832.

const314 (approved):

IlyaKhD (approved):

PR #23422, opened 2022-12-12
Renovation: Fix onChange call for RadioButton inside RadioGroup

const314:


packages/jquery/patch/radio-button-hocs.jsx
```diff
@@ -26,6 +26,7 @@ function withRadioGroup(RadioButton) {
         const handleChange = (event) => {
             event.preventDefault();
             event.stopPropagation();
```
williamvinogradov: Let's not cancel the native event here:
```suggestion
```

packages/react/src/components/radio-button/radio-button-hocs.tsx
```diff
@@ -46,6 +46,7 @@ function withRadioGroup(RadioButton: RadioButtonRenderType) {
     const handleChange: ChangeEventHandler<HTMLInputElement> = (event) => {
       event.preventDefault();
       event.stopPropagation();
```
williamvinogradov: Let's not cancel the native event here:
```suggestion
```

packages/jquery/patch/radio-button-hocs.jsx
const314: fixed

packages/react/src/components/radio-button/radio-button-hocs.tsx
const314: fixed

williamvinogradov (approved):

IlyaKhD (approved):

PR #23305, opened 2022-11-29
TestCafe: add material theme testing

EugeniyKiyashko:

PR #23431, opened 2022-12-13
Scheduler: try to accelerate testcafe tests

EugeniyKiyashko:


testing/testcafe/tests/htmlEditor/format.ts
```diff
@@ -1,13 +1,17 @@
 import SelectBox from '../../model/selectBox';
 import url from '../../helpers/getPageUrl';
 import createWidget from '../../helpers/createWidget';
+import { clearTestPage } from '../../helpers/clearPage';
 
-fixture`HtmlEditor - formats`
-  .page(url(__dirname, '../container.html'));
+fixture.disablePageReloads`HtmlEditor - formats`
+  .page(url(__dirname, '../containerQuill.html'))
+  .afterEach(async () => clearTestPage());
```
williamvinogradov, line 8: Question.

Does this ``` afterEach ``` in the fixture work properly?
I had some troubles with it when developing the ```safeSizeTest``` helper.
It had not been executed in some test cases.

But maybe I missed something.

williamvinogradov:
I've reviewed ```HtmlEditor``` / ```Scheduler``` test cases by corner of my eye

It looks ok, but here are two points:
- In some test cases we change ```safeSizeTest``` back to ```test```
The scheduler team wraps all tests in safeSizeTest for increasing the stability of the scheduler blinking tests.
Maybe it's overhead, but tests stopped blinking :)

- In some test cases, we remove the ```SPEED``` option
As far as I know, this option was added for increasing stability, and removing it may cause some blinking tests issue.

I agree with you that our Testcafe's test pipeline should be speeded up.
Maybe some of these points in edited test scenarios were added by accident.
And it only slows down our tests and is redundant.
But maybe not.

**Therefore let's try to speed up our pipeline and merge it.**
But keeping eye on the blinking test cases carefully to fix them as we discover them.

testing/testcafe/tests/htmlEditor/format.ts
EugeniyKiyashko: Yes, afterEach works properly in this case

williamvinogradov (dismissed):

testing/testcafe/runner.js
```diff
@@ -115,3 +132,24 @@ function getArgs() {
         }
     });
 }
+
+function clearTestPage() {
```
williamvinogradov, line 136: Let's move this function before ```createTestCafe```

Because this function is used before it's declaration :)

williamvinogradov (dismissed):

testing/testcafe/helpers/domUtils.ts
```diff
@@ -0,0 +1,51 @@
+import { ClientFunction } from 'testcafe';
+
+function createElement(
+  tagName: string,
+  id: string,
+  style: Partial<CSSStyleDeclaration>,
+): HTMLElement {
+  const el = document.createElement(tagName);
```
ksercs:
```suggestion
  const element = document.createElement(tagName);
```

testing/testcafe/runner.js
```diff
@@ -24,10 +25,11 @@ createTestCafe('localhost', 1437, 1438)
 
         componentFolder = componentFolder ? `${componentFolder}/**` : '**';
         if(fs.existsSync('./testing/testcafe/screenshots')) {
-            fs.rmdirSync('./testing/testcafe/screenshots', { recursive: true });
+            fs.rmSync('./testing/testcafe/screenshots', { recursive: true });
         }
 
-        const browsers = args.browsers.split(' ').map(expandBrowserAlias);
+        const browsers = args.browsers.split(' ')
+            .map((browser) => `${expandBrowserAlias(browser)}${args.componentFolder.trim() === 'scheduler' ? ' --window-size=1200,800' : ''}`);
```
ksercs, line 32: Why is this window size selected only for scheduler?

testing/testcafe/runner.js
```diff
@@ -77,9 +80,22 @@ createTestCafe('localhost', 1437, 1438)
...
+
+        const runOptions = {
+            quarantineMode: args.quarantineMode,
+        };
+
+        if(args.componentFolder.trim() === 'scheduler') {
+            runOptions.hooks = {
+                test: {
+                    after: async() => {
+                        await clearTestPage();
```
ksercs, line 92: Why only sheduler?

testing/testcafe/runner.js
```diff
@@ -115,3 +132,24 @@ function getArgs() {
         }
     });
 }
+
+function clearTestPage() {
+    return createTestCafe.ClientFunction(() => {
```
ksercs, line 137: Why do we reimplement this function?
https://github.com/DevExpress/DevExtreme/pull/23431/files#diff-521e8c2c541410a6f64a734a893d78e2e8b03a83138c34e8d6033b7cb6780a59R3

testing/testcafe/tests/navigation/form/form.simpleItem.ts
```diff
@@ -37,7 +38,7 @@ fixture`Form`
 
 ['left', 'right', 'top'].forEach((labelLocation) => {
   ['generic.light', 'material.blue.light'].forEach((theme) => {
-    test('widget alignment (T1086611)', async (t) => {
+    safeSizeTest('widget alignment (T1086611)', async (t) => {
```
ksercs, line 41: Why do we add `safeSizeTest` while remove it in Sheduler?
Why theme testing is implemented here in old way?

testing/testcafe/tests/scheduler/api/deleteRecurrence.ts
```diff
@@ -106,10 +105,10 @@ safeSizeTest('should exclude from recurrence if mode is "occurrence"', async (t)
       apptColor: 1,
       recurrenceRule: 'FREQ=DAILY;COUNT=4',
     }],
-  },
+  }, true,
 ));
 
-safeSizeTest('should show delete recurrence dialog if mode is "dialog"', async (t) => {
+test.skip('should show delete recurrence dialog if mode is "dialog"', async (t) => {
```
ksercs, line 111: Should this test be skipped?

testing/testcafe/tests/scheduler/layout/appointments/dataSource.ts
```diff
@@ -20,4 +20,32 @@ test.skip('Appointment key should be deleted when removing an appointment from s
 
     .expect(compareResults.isValid())
     .ok(compareResults.errorMessages());
+}).before(async () => {
+  const initScheduler = ClientFunction(() => {
+    const devExpress = (window as any).DevExpress;
+
+    ($('#container') as any).dxScheduler({
+      dataSource: new devExpress.data.DataSource({
```
ksercs, line 28: don't we have another way to get access to `DataSource`?

testing/testcafe/tests/scheduler/api/deleteRecurrence.ts
williamvinogradov: Skipped this, I think we shouldn't skip this test

Let's bring it back :)

testing/testcafe/runner.js
williamvinogradov: Unfortunately, we cannot import *ts in *js files directly.

The best solution is to compile existing ```clearTestPage``` with ```tsc``` and import it.
But reimplementing this function in js is a fast solution and we agreed with @EugeniyKiyashko that it will be replaced via compiled from ts version in the next PR.

testing/testcafe/runner.js
EugeniyKiyashko: Tests of other components is not ready to use this default browser size. I am working on it. When all will ready the all tests will use this browser size.

testing/testcafe/runner.js
EugeniyKiyashko: The same reason that is in previous answer

testing/testcafe/runner.js
EugeniyKiyashko: Yes it will be leaded to a general view in the future requests

testing/testcafe/tests/navigation/form/form.simpleItem.ts
EugeniyKiyashko: A new way will be introduce in the follow PR - https://github.com/DevExpress/DevExtreme/pull/23305/files#top

testing/testcafe/tests/scheduler/api/deleteRecurrence.ts
EugeniyKiyashko: It should be skipped. The test creates two instances of dialog instead of one. I discussed this test with @andrewmakarov and we decided to skip it. Andrew promises to investigate the issue and delete skip after that.

testing/testcafe/tests/scheduler/layout/appointments/dataSource.ts
EugeniyKiyashko: No, I don't know another way. This approach uses wide in our tests.

testing/testcafe/runner.js
EugeniyKiyashko: There are other functions declared the same way - setTestingPlatform(), expandBrowserAlias()... I think that is not a big problem. This function will be move to separate js file soon. So, let's wait future changes. I am going to request your review for this change.

ksercs (approved):

williamvinogradov (approved):

PR #23443, opened 2022-12-14
Ci/watch hot reload

iBat:


packages/react/project.json
```diff
@@ -8,7 +8,6 @@
             "executor": "nx:run-commands",
             "options": {
                 "commands": [
-                    "tsc -b -w ./tsconfig.package.json",
```
williamvinogradov, line 11: If I remember correctly, the only reason why the React package has a project.json is that we need a complicated command for React here.
We simplified the command for the React package.

Maybe we can delete project.json for this package now?

playgrounds/react/package.json
```diff
@@ -29,6 +29,7 @@
     "build": "react-scripts build",
     "start": "react-scripts start",
     "serve": "npm start",
```
williamvinogradov: Let's remove the duplicate npm command

```suggestion
```

playgrounds/angular/package.json
```diff
@@ -3,9 +3,10 @@
   "version": "1.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve",
+    "start": "nx serve @playgrounds/angular",
     "serve": "npm start",
```
williamvinogradov: Let's remove the duplicate npm command

```suggestion
```

playgrounds/angular/project.json
```diff
@@ -0,0 +1,83 @@
+{
```
williamvinogradov, line 1: Minor commment / Quiestion.

Why is the angular config file new? 
What happened with the old version?

iBat: With the old version, the application was not updated when the libraries changed.

williamvinogradov (approved):

MikeVitik (changes requested):
Add jquery watch

alexslavr (approved):

tsconfig.base.json
```diff
@@ -19,7 +19,7 @@
     "noPropertyAccessFromIndexSignature": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
-    "noEmitOnError": true,
+    "noEmitOnError": false,
```
MikeVitik: Move this noEmitOnError into the watch rollup task only

MikeVitik (changes requested):

MikeVitik (approved):

const314 (approved):

PR #23445, opened 2022-12-14
React: Remove dependencies between the radio-button and the radio-group component

const314:


MikeVitik (approved):
May be rename it to
packages/react/src/compatible-components/radio-group/radio-group.tsx

packages/react/src/components/radio-common/types.ts
```diff
@@ -0,0 +1 @@
+export type RadioGroupValue = string | number;
```
williamvinogradov: Looks like we don't need RadioGroupValue in the React package anymore.
 Because we already move it to the 'components' package in [this PR](https://github.com/DevExpress/DevExtreme/pull/23450)
 
 So, let's delete this file :)

packages/react/src/components/radio-common/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './types';
```
williamvinogradov:
```suggestion
```

packages/react/src/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from './components/radio-common';
```
williamvinogradov: Should we export the RadioGroup context from the React library?
Anyway, our users cannot use it.

Let's leave radio-common private.

```suggestion
```

packages/react/src/components/radio-common/types.ts
const314: You're right, I forgot to delete it. Fixed.

packages/react/src/components/radio-common/index.ts
const314: fixed

packages/react/src/index.ts
const314: fixed

williamvinogradov (approved):

PR #23452, opened 2022-12-15
Renovation Angular: Implement first version of the dynamic templates.

williamvinogradov:


const314 (approved):

packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts
```diff
@@ -0,0 +1,99 @@
...
+  Input,
+  TemplateRef,
+  ViewContainerRef,
+} from '@angular/core';
+import { UnknownRecord } from '@devextreme/core';
+import { AngularTemplate } from '../types';
+import { ViewComponent } from './view.component';
+
+interface AngularTemplateContext<T> {
+  $implicit: T;
```
MikeVitik, line 15: Question: What is '$implicit' (as I undestand it props for push data to template)? Is it standard Angular template props? Can you provide some doc ref? (I try but without success)

packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+
+  private setDataToComponent(): void {
+    const { instance } = this.componentRef!;
+
+    (Object.keys(this.dataValue!) as (keyof TData)[])
+      // TODO: Think about optimization here.
+      // .filter((key) => Object.prototype.hasOwnProperty.call(instance, key))
+      .forEach((key) => {
```
MikeVitik, line 94: Minor: Alternative for forEach here is for ... of.

packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  private templateRef?: EmbeddedViewRef<AngularTemplateContext<TData>>;
+
+  constructor(
+    private viewContainerRef: ViewContainerRef,
+  ) {
+  }
+
+  private setTemplate(template: AngularTemplate<TComponent>): void {
+    if (!template) {
```
MikeVitik, line 58: Question: Is it OK that we doesn't reset templateValue it template is null or undefined?

packages/angular/projects/angular/src/components/radio-button/radio-button.component.ts
```diff
@@ -62,15 +73,28 @@ implements OnDestroy {
...
+  @Input() labelTemplate: AngularTemplate<LabelViewComponent> = LabelViewComponent;
+
+  // TODO: Add this template to core prop types.
+  @Input() radioTemplate: AngularTemplate<RadioViewComponent> = RadioViewComponent;
+
   @Output() onChange = new EventEmitter<Event>();
 
   @Output() onClick = new EventEmitter<MouseEvent>();
 
   checked$ = this.strategy.checked$;
```
MikeVitik, line 88: Suggestion: May be we can return boolean from this.strategy.checked$ and remove map here?

packages/angular/projects/angular/src/components/radio-button/radio-button.strategies.ts
```diff
@@ -50,12 +43,25 @@ function createRadioGroupStrategy(
 
   const setChecked = () => {};
 
+  const onInit = () => {
+    radioGroupCore$.pipe(waitContextAndDo())
```
MikeVitik, line 46: Suggestion: Looks like radioGroupCore has already has context value. May be we should simplify code and remove waitContextAndDo at all.

playgrounds/angular/src/app/examples/radio-group/radio-group-example.component.ts
```diff
@@ -1,32 +1,13 @@
...
-
-    <div>
-      <span>Selected value: </span>
-      <span>{{value$ | async}}</span>
-    </div>
+    <dx-radio-group-simple></dx-radio-group-simple>
+    <dx-radio-group-customization-components></dx-radio-group-customization-components>
+    <dx-radio-group-customization-templates></dx-radio-group-customization-templates>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush,
```
MikeVitik, line 10: Question: Can you explain why do we force OnPush strategies here and is it related to how our user should use our component? (For example: this code is obligate for the radio-group component properly working)

MikeVitik (commented):
Please give me explanation of my questions.

packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts
williamvinogradov: Unfortunately, the Angular templates are badly documented.

```
Using the key $implicit in the context object will set its value as default.
```
[Here is the link to the docs with $implicit](https://angular.io/api/common/NgTemplateOutlet#description)

Our clients may don't know what name of the variable we pass to the template.
And with ```$implicit``` usage they don't need to think about it, they can just declare the template variable.

```html
<!-- template without $implicit -->
<ng-template let-data="variable_name"> <!-- <-- need to know what to write here -->
  {{ data }}
</ng-template>

<!-- template with $implicit -->
<ng-template let-data> <!-- <-- just declare template variable -->
  {{ data }}
</ng-template>

File: packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts (hunk 3 above)
Comment (williamvinogradov) [reply packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts:58]: We don't need the dynamic template component without a template.

If we want to destroy a template we can just destroy the dynamic-template component :)

E.g
```html
<dx-dynamic-template
        <!-- component will be destroyed if the 'labelTemplate' will be 'null or 'undefined' -->
        *ngIf="labelTemplate" 
        [template]="labelTemplate"
        [data]="{ label: label }">
</dx-dynamic-template>
```

I will add this check to the radio-button component. Thanks!

packages/angular/projects/angular/src/components/radio-button/radio-button.component.ts
williamvinogradov: We already return a boolean as stream in ``` this.strategy.checked$ ``` ( ```  Observable<boolean> ```)
We need to wrap ```checked``` in the object for the ```dx-dynamic-template``` component.

packages/angular/projects/angular/src/components/radio-button/radio-button.strategies.ts
williamvinogradov: The core has context here only if we still init it in in the ngOnInit of the radio-group component.

Without helper operator ```waitContextAndDo``` it looks like:
```ts
radioGroupCore$.pipe(
  filter((context): context is RadioGroupCore<RadioGroupValue> => !!context),
  take(1),
).subscribe(...);
```

If we remove the filter, we win just one line, but lose in safety and future changes:
```ts
radioGroupCore$.pipe(
  take(1),
).subscribe(...);
```

So, let's leave it as is :)

playgrounds/angular/src/app/examples/radio-group/radio-group-example.component.ts
williamvinogradov: **Why we must use ``` changeDetection.onPush ```?**

Angular  ``` changeDetection.onPush ``` has one specific feature.
If the parent component has ``` onPush``` strategy, all child components will live with ``` onPush ``` strategy too.
And it doesn't matter what kind of strategy set in the ```@Component``` decorator
And if the component works correctly in ``` onPush ``` strategy, it'll work correctly in the default strategy, but not vice versa.

In other words, if we develop Angular components with onPush it will be suitable for all clients, but if we will choose the default strategy, clients with ``` onPush ``` may have troubles with our components.
This is the reason why we must have ``` onPush ``` in all our Angular components :)

**Regarding using it at this place**
- I automatically use define ``` onPush ``` everythere.
Also, we need to add it to ``` schematic ``` (generation of the component template to set ``` onPush ``` by default) and eslint rule to make it required.
- We must use ``` onPush ``` in playground because if the component works in ``` onPush ``` it will work in default strategy, but not vice versa (see previous paragraph )

packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts
williamvinogradov: I think you mean ```for ... in ``` not ``` for ... of```

If we use ```for ... in``` here, it will look like this:
```ts
    for (const key in this.dataValue!) {
      // It's for the eslint rule 'guard-for-in'
      if (Object.prototype.hasOwnProperty.call(instance, key)) {
        (instance as UnknownRecord)[key] = this.dataValue![key];
      }
    }
```    

But the ```hasOwnProperty``` check is not suitable for us, because the view component may have the ```@Input()``` field in ts, but will not have it in runtime.

E.g. components like this
```ts
@Component({...})
export class MyComponent {
  @Input() myField?: string;
}
```
in runtime will not have ```myField``` until it will be initialized :)

If we implement your suggestion, we will need to disable the eslint rule 'guard-for-in' :(
So, let's leave it as is

MikeVitik: Yes. It's my mistake. Object.entries or Object.getOwnPropertyNames is more sutable suggestion for here.

packages/angular/projects/angular/src/components/radio-button/radio-button.strategies.ts
MikeVitik: I suppose without waitContextAndDo it's simplified to

```
radioGroupCore$.subscribe(...);
```

MikeVitik (approved):

packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts
MikeVitik: Minor: Another worlds/ After your update template has never will null and if (!template) should raise assertion or debug warning about missing *ngIf="labelTemplate"?

playgrounds/angular/src/app/examples/radio-group/radio-group-example.component.ts
MikeVitik: If I understand you right, it is not necessary or optional for our customers to use the onPush strategy. It is up to their decision.

williamvinogradov: Yeah, our customers can use the change detection strategy that they like :)

IlyaKhD: Can we add auto-check or a sort of testing to ensure that `onPush` is enabled in all the components?

packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  private setDataToComponent(): void {
+    const { instance } = this.componentRef!;
+
+    (Object.keys(this.dataValue!) as (keyof TData)[])
+      // TODO: Think about optimization here.
+      // .filter((key) => Object.prototype.hasOwnProperty.call(instance, key))
+      .forEach((key) => {
+        (instance as UnknownRecord)[key] = this.dataValue![key];
```
IlyaKhD: **Minor**: can you avoid using `as`?

If not - I suggest moving it to variable declaration:
```ts
const instance = this.componentRef?.instance as UnknownRecord;
```

IlyaKhD (approved):

playgrounds/angular/src/app/examples/radio-group/radio-group-example.component.ts
williamvinogradov: Yes, I will add it via separate PR soon

packages/angular/projects/angular/src/internal/dynamic-template/dynamic-template.component.ts
williamvinogradov: Unfortunately, we cannot avoid ```as``` here because it works like reflection.

I will move it to a separate variable as you suggest.

PR #23487, opened 2022-12-20
dxScheduler - removed header top margin

gagharutyunyan1993:


gagharutyunyan1993:
[The scheduler header cell indent needs to be corrected](https://trello.com/c/Nqf7OnCI/1788-the-scheduler-header-cell-indent-needs-to-be-corrected)

PR #23488, opened 2022-12-20
Renovation: Implement Angular compatible RG.

williamvinogradov:
Added the Angular RG compatible component.
Also, some minor changes in the existing Angular RG for the compatible component support.

### DIFF
--- a/packages/angular/projects/angular/src/compatible-components/radio-group/radio-group-compatible.component.ts
+++ b/packages/angular/projects/angular/src/compatible-components/radio-group/radio-group-compatible.component.ts
@@ -1,0 +1,64 @@
+import {
+  ChangeDetectionStrategy, Component, ContentChild, Input,
+} from '@angular/core';
+import { RadioGroupValue } from '@devextreme/components/src';
+import { compileGetter, ItemLike } from '@devextreme/interim';
+import { TemplateCompatibleDirective } from '../../compatible-directives/template';
+import { RadioGroupBaseComponent } from '../../components/radio-common';
+
+// TODO: Move these to components package
+//  because these types equal to similar types in React RG compatible component.
+type ValueGetter = <T extends RadioGroupValue>(item: ItemLike) => T;
+type LabelGetter = (item: ItemLike) => string;
+
+@Component({
+  selector: 'dx-radio-group-compat',
+  template: `
+    <dx-radio-group
+      [value]="value"
+      (valueChange)="valueChange.emit($event)">
+        <dx-radio-button
+          *ngFor="let item of items"
+          [label]="labelGetter | apply:item"
+          [value]="valueGetter | apply:item"
+          [labelTemplate]="labelTemplateDirective?.templateRef"
+          >
+        </dx-radio-button>
+    </dx-radio-group>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class RadioGroupCompatibleComponent<T extends RadioGroupValue>
+  extends RadioGroupBaseComponent<T> {
+  @Input() value?: T;
+
+  @Input() items: Array<ItemLike> = [];
+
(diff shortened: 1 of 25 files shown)


playgrounds/angular/src/app/examples/radio-group-compat/radio-group-compat-simple.component.ts
```diff
@@ -0,0 +1,41 @@
...
+@Component({
+  selector: 'dx-radio-group-compat-simple',
+  template: `
+    <div class="example">
+      <div class="example__title">
+        RadioGroup simple example
+      </div>
+      <div class="example__control">
+        <dx-radio-group-compat
+          *ngIf="value$ | async as value"
```
MikeVitik, line 13: Question: Is it possible that value$ is undefined or null?

packages/angular/projects/angular/src/components/radio-group/radio-group.component.ts
```diff
@@ -31,16 +29,16 @@ export type RadioGroupInputs<T extends RadioGroupValue> =
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class RadioGroupComponent<T extends RadioGroupValue>
-implements OnInit, RadioGroupInputs<T> {
+  extends RadioGroupBaseComponent<T>
+  implements OnInit, RadioGroupInputs<T> {
   private inputValue?: T;
 
-  @Input() set value(value: T) {
+  @Input() set value(value: T | undefined) {
```
MikeVitik, line 36: Minor
```suggestion
  @Input() set value(value?: T) {
```

packages/angular/projects/angular/src/compatible-components/radio-group/radio-group-compatible.component.ts
```diff
@@ -0,0 +1,64 @@
+import {
+  ChangeDetectionStrategy, Component, ContentChild, Input,
+} from '@angular/core';
+import { RadioGroupValue } from '@devextreme/components/src';
+import { compileGetter, ItemLike } from '@devextreme/interim';
+import { TemplateCompatibleDirective } from '../../compatible-directives/template';
+import { RadioGroupBaseComponent } from '../../components/radio-common';
+
+// TODO: Move these to components package
```
MikeVitik, line 9:
```suggestion
// TODO @williamvinogradov: Move these to components package
```

MikeVitik (approved):

playgrounds/angular/src/app/examples/radio-group-compat/radio-group-compat-simple.component.ts
williamvinogradov: Yes, the first returning value in the data stream from the async pipe is 'null'

I didn't find anything about it in the official Angular documentation.
Looks like this behavior is badly documented.
But [here is the link](https://github.com/angular/angular/blob/bbeac0727b8f267a47aba1ff1bcfc8cc5ca15b61/packages/common/src/pipes/async_pipe.ts#L73) to the async pipe source code that can prove it :)

packages/angular/projects/angular/src/components/radio-group/radio-group.component.ts
williamvinogradov: No, we cannot make ```value``` optional because if we do it we got the next ts error:
```
TS1051: A 'set' accessor cannot have an optional parameter.
```

const314 (approved):

packages/angular/projects/angular/src/compatible-components/radio-group/radio-group-compatible.component.ts
williamvinogradov: I especially don't mention myself in this TODO.
Because if we start moving these common types from React it can be done in Angular in the same PR.
And this task may be assigned to another developer :)

PR #23523, opened 2022-12-22
Renovation: Make the Angular RG fully generic.

williamvinogradov:
Angular radio group became fully generic.
Added new playground examples for the radio button.
Also, small fixes in naming and etc.

### DIFF
--- a/packages/angular/projects/angular/src/components/radio-button/radio-button.component.ts
+++ b/packages/angular/projects/angular/src/components/radio-button/radio-button.component.ts
@@ -8,8 +8,12 @@ import {
   Optional,
   Output,
 } from '@angular/core';
-import { RadioGroupValue } from '@devextreme/components';
-import { map } from 'rxjs';
+import {
+  filter,
+  map,
+  Subject,
+  takeUntil,
+} from 'rxjs';
 import { AngularTemplate } from '../../internal';
 import { RadioGroupService } from '../radio-common';
 
@@ -52,7 +56,7 @@ let nextUniqueId = 0;
   `],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class RadioButtonComponent<T extends RadioGroupValue>
+export class RadioButtonComponent<T>
 implements OnInit, OnDestroy {
   private strategy = createRadioButtonStrategy(
     this.radioGroupService?.context$,
@@ -61,6 +65,8 @@ implements OnInit, OnDestroy {
 
   // eslint-disable-next-line no-plusplus
   private uniqueId = `dx-radio-button-${++nextUniqueId}`;
+
+  private destroy$ = new Subject<void>();
 
   @Input() id = this.uniqueId;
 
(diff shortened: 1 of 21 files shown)


packages/angular/projects/angular/src/components/radio-button/radio-button.component.ts
```diff
@@ -90,6 +96,9 @@ implements OnInit, OnDestroy {
   // eslint-disable-next-line @angular-eslint/no-output-on-prefix
   @Output() onClick = new EventEmitter<MouseEvent>();
 
+  // eslint-disable-next-line @angular-eslint/no-output-on-prefix
+  @Output() onSelected = new EventEmitter<T>();
```
MikeVitik, line 90: Do we need both onChange and onSelected event?

MikeVitik (changes requested):
Do we need both onChange and onSelected event?

packages/angular/projects/angular/src/components/radio-button/radio-button.strategies.ts
```diff
@@ -74,8 +74,8 @@ export interface RadioButtonStrategy {
 }
 
 export function createRadioButtonStrategy(
-  radioGroupCore$: Observable<RadioGroupCore<RadioGroupValue> | undefined> | undefined,
-  getRadioButtonValue: () => RadioGroupValue,
+  radioGroupCore$: Observable<RadioGroupCore<unknown> | undefined> | undefined,
```
const314, line 77: Can we make createRadioButtonStrategy generic as well?

const314 (approved):

packages/angular/projects/angular/src/components/radio-button/radio-button.strategies.ts
williamvinogradov: Yes, we can, but here are some nuances about it.

We must also complete the radio group service generic to make the radio-button strategy generic.
Because strategy uses context from the service.

```ts
// create function
export function createRadioButtonStrategy<T>(
  radioGroupCore$: Observable<RadioGroupCore<T> | undefined> | undefined, // <-- here
  getRadioButtonValue: () => unknown,
): RadioButtonStrategy {
// ...
}

// call of this function
private strategy = createRadioButtonStrategy<T>(
    this.radioGroupService?.context$, // <-- context$ must be generic with T too
    () => this.value,
  );
```

And here is the nuance comes to play :)
We know the exact T type only at runtime when the client provides value to the radio-group component.
So, in other words, we know the exact T type in the radio group at the runtime.
To provide a generic service we must know the exact T type at the compilation time.

```ts
// generic service
export const INJECTION_TOKEN = new InjectionToken('MyGenericService');

export class MyGenericService<T> {}

// providing service (register it in the DI)
@Component({
  selector: 'app-my-component',
  template: '...'
  providers: [{
    provide: INJECTION_TOKEN,
    useValue: new MyGenericService<T>(), // <-- Error: T is not defined here, we must know the exact type.
  }]
})
export class MyComponent {}
```

Some workarounds exist to solve this issue but in my opinion, the game's not worth playing.
Because we don't work with type in strategy/service and the unknown type is okay here :)

MikeVitik (approved):

PR #23534, opened 2022-12-23
Gallery - Fix loop slide show when itemTemplate is used in React  (T1132935)

ivanblinov2k17:


js/ui/gallery.js
```diff
@@ -90,6 +90,7 @@ const GalleryNavButton = Widget.inherit({
 const Gallery = CollectionWidget.inherit({
 
     _activeStateUnit: GALLERY_ITEM_SELECTOR,
+    _resizeCalled: false,
```
ksercs:
```suggestion
    _wasAnyItemTemplateRendered: false,
```

js/ui/gallery.js
```diff
@@ -337,6 +338,20 @@ const Gallery = CollectionWidget.inherit({
         this._loadNextPageIfNeeded();
     },
 
+    _createItemByTemplate(itemTemplate, renderArgs) {
+        return itemTemplate.render({
+            model: renderArgs.itemData,
+            container: renderArgs.container,
+            index: renderArgs.index,
+            onRendered: () => {
+                if(!this._resizeCalled) {
```
ksercs: Hm, can we just add an empty `onItemTemplateRendered` handler on collection level and override if here to reduce the copypaste?

js/ui/gallery.js
ksercs: We should also delete this cache on component dispose

js/ui/gallery.js
```diff
@@ -337,6 +338,20 @@ const Gallery = CollectionWidget.inherit({
...
+                    this._resizeCalled = true;
+                    this._dimensionChanged();
+                }
+            }
+        });
+    },
+
```
ksercs, line 350: Can we add a test? For example, to check that resizeCallback is called after template rendering

js/ui/gallery.js
```diff
@@ -337,6 +338,20 @@ const Gallery = CollectionWidget.inherit({
...
 
+    _createItemByTemplate(itemTemplate, renderArgs) {
+        return itemTemplate.render({
+            model: renderArgs.itemData,
+            container: renderArgs.container,
+            index: renderArgs.index,
+            onRendered: () => {
+                if(!this._resizeCalled) {
+                    this._resizeCalled = true;
+                    this._dimensionChanged();
```
ksercs:
```suggestion
                    this._dimensionChanged(); // NOTE: T1132935
```

js/ui/collection/ui.collection_widget.base.js
```diff
@@ -973,10 +973,15 @@ const CollectionWidget = Widget.inherit({
         return itemTemplate.render({
             model: renderArgs.itemData,
             container: renderArgs.container,
-            index: renderArgs.index
+            index: renderArgs.index,
+            onRendered: this._onItemTemplateRendered(itemTemplate, renderArgs)
         });
     },
 
+    _onItemTemplateRendered: function() {
```
ksercs, line 981: Seems like it should looks like 
```suggestion
    _onItemTemplateRendered: noop,
```

js/ui/gallery.js
```diff
@@ -988,6 +998,7 @@ const Gallery = CollectionWidget.inherit({
     },
 
     _dispose: function() {
+        this._wasAnyItemTemplateRendered = undefined;
```
ksercs:
```suggestion
        this._wasAnyItemTemplateRendered = null;
```

js/ui/gallery.js
```diff
@@ -90,6 +90,7 @@ const GalleryNavButton = Widget.inherit({
 const Gallery = CollectionWidget.inherit({
 
     _activeStateUnit: GALLERY_ITEM_SELECTOR,
+    _wasAnyItemTemplateRendered: undefined,
```
ksercs:
```suggestion
    _wasAnyItemTemplateRendered: false,
```

testing/tests/DevExpress.ui.widgets/gallery.tests.js
```diff
@@ -740,6 +740,13 @@ QUnit.module('behavior', {
 
         assert.equal(instance._pagesCount(), 2, 'pages count is correct');
     });
+
+    QUnit.test('resizeCallback is called after item is rendered (T1132935)', function(assert) {
+        const gallery = this.$element.dxGallery({
+            items: [0, 1, 2, 3],
+        }).dxGallery('instance');
+        assert.equal(gallery._wasAnyItemTemplateRendered, true);
```
ksercs: can we do it without a private api?

ivanblinov2k17: i tried the Callbacks.fired, but it was not fired
i tried to add callback that sets cache variable to resizeCallbacks and it didnt work too and i havent found the similar cases in test code, so i decided to leave it as it is

js/ui/collection/ui.collection_widget.base.js
ivanblinov2k17: it needs to return noop to be callable cause we have to pass render args to closure in widgets async

testing/tests/DevExpress.ui.widgets/gallery.tests.js
ksercs: probably if we use `triggerResizeEvent` instead of direct `_dimensionChanged` call it will be easier to test

ksercs (approved):

EugeniyKiyashko (approved):

PR #23602, opened 2022-12-30
Add examples for using theme in react

MikeVitik:
Add examples:
* using the default theme and scoped theme
* using a dynamic theme switching
For supporting scoped theme scenario split on theme css file to *-scoped.css (has .dx-theme-* selector) and *.css (has only :root selector)

### DIFF
--- a/playgrounds/react/src/examples/themes/dynamic-theme-example.tsx
+++ b/playgrounds/react/src/examples/themes/dynamic-theme-example.tsx
@@ -1,0 +1,27 @@
+import { RadioButton, RadioGroup } from '@devextreme/react';
+
+import '@devextreme/styles/lib/themes/dark-scoped.css';
+
+const OPTIONS = ['light', 'dark'];
+const handleChange = async (newValue?: string) => {
+  // In reaal demo we can replace static import to fetch and link injection:
+  // const cssText = await fetch('@devextreme/styles/lib/themes/dark-scoped.css');
+  // const link = document.getElementById('theme-style') || document.createElement('style');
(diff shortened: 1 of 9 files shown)


playgrounds/react/src/examples/themes/dynamic-theme-example.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import '@devextreme/styles/lib/themes/dark-scoped.css';
+
+const OPTIONS = ['light', 'dark'];
+const handleChange = async (newValue?: string) => {
+  // In reaal demo we can replace static import to fetch and link injection:
+  // const cssText = await fetch('@devextreme/styles/lib/themes/dark-scoped.css');
+  // const link = document.getElementById('theme-style') || document.createElement('style');
+  // style.innerText = cssText;
+  document.getElementById('dynamic-theme')!.className = `dx-theme-${newValue}`;
```
const314, line 11: Do we need direct DOM manipulation? What if we save chosen theme to the state in handleChange and set className directly in div on render?
<div id="dynamic-theme" className=`dx-theme-${chosenTheme}`>

const314 (approved):

playgrounds/react/src/examples/themes/dynamic-theme-example.tsx
MikeVitik: Currently, it's a proof of concept. I suppose in the future it can be a multi-platform helper method, so I don't want to write platform-dependent implementation.

williamvinogradov (approved):

PR #23615, opened 2023-01-04
Sync with 23.1 (5611024)

IlyaKhD:

### DIFF
--- a/packages/interim/src/renovation/component_wrapper/common/component.ts
+++ b/packages/interim/src/renovation/component_wrapper/common/component.ts
@@ -97,16 +97,14 @@ export default class ComponentWrapper extends DOMComponent<ComponentWrapperProps
   }
 
   validateKeyDownHandler(): void {
-    /* const supportedKeyNames = this.getSupportedKeyNames();
+    const supportedKeyNames = this.getSupportedKeyNames();
     const hasComponentDefaultKeyHandlers = supportedKeyNames.length > 0;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-     TODO
-    
     const hasComponentKeyDownMethod = typeof (this._viewComponent.prototype as any).keyDown === 'function';
 
     if (hasComponentDefaultKeyHandlers && !hasComponentKeyDownMethod) {
       throw Error('Component\'s declaration must have \'keyDown\' method.');
-    }*/ 
+    }
   }
 
   public get viewRef(): unknown {
--- a/packages/interim/src/renovation/component_wrapper/common/mutations_recording.ts
+++ b/packages/interim/src/renovation/component_wrapper/common/mutations_recording.ts
@@ -1,27 +1,26 @@
 function isChildNode(node: Node): node is ChildNode {
-    return typeof (node as Partial<ChildNode>).remove === 'function';
+  return typeof (node as Partial<ChildNode>).remove === 'function';
+}
+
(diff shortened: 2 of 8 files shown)


packages/interim/src/renovation/component_wrapper/common/component.ts
```diff
@@ -97,16 +97,14 @@ export default class ComponentWrapper extends DOMComponent<ComponentWrapperProps
   }
 
   validateKeyDownHandler(): void {
-    /* const supportedKeyNames = this.getSupportedKeyNames();
+    const supportedKeyNames = this.getSupportedKeyNames();
```
IlyaKhD, line 100: @MikeVitik, is it OK that I uncommented these lines?

const314 (approved):

packages/interim/src/renovation/component_wrapper/common/mutations_recording.ts
```diff
@@ -1,27 +1,26 @@
...
-    const mutations = observer.takeRecords();
-    observer.disconnect();
-  
-    return () => mutations.forEach(revertMutation);
-  }
-  
\ No newline at end of file
+}
+
+export function recordMutations<TReturn>(target: Node, func: () => TReturn): () => void {
```
williamvinogradov, line 15: Minor comment.

Maybe we should change the generic ```TReturn``` to the ```unknown``` type?
Something like this:
```suggestion
export function recordMutations(target: Node, func: () => unknown): () => void {
```

Because we lost the ```func``` return value and it doesn't matter what this function returns :)

packages/interim/src/renovation/component_wrapper/common/mutations_recording.ts
```diff
@@ -1,27 +1,26 @@
...
+}
+
+function revertMutation({ type, addedNodes }: MutationRecord): void {
+  switch (type) {
+    case 'childList':
+      addedNodes.forEach((n) => isChildNode(n) && n.remove());
+      break;
+    default:
+      break;
   }
```
williamvinogradov, line 12: Minor suggestion.

Let's don't use the ```switch``` with one possible option:
```suggestion
function revertMutation({ type, addedNodes }: MutationRecord): void {
  if (type === 'childList') {
    addedNodes.forEach((n) => isChildNode(n) && n.remove());
  }
}
```

williamvinogradov (approved):

packages/interim/src/renovation/component_wrapper/common/component.ts
MikeVitik: Yes. It's OK

MikeVitik (approved):

packages/interim/src/renovation/component_wrapper/common/mutations_recording.ts
IlyaKhD: I like using `switch` statement in such cases because it highlights that execution flow is divided into branches. `if` statements creates branches that can be combined.

packages/interim/src/renovation/component_wrapper/common/mutations_recording.ts
IlyaKhD: Yep, we should clean up it. Since this PR is for synchronisation I've included proposed changes to this one: https://github.com/DevExpress/DevExtreme/pull/23637

PR #23632, opened 2023-01-09
DataGrid - The type information is incorrect when the customizeText cellInfo value is an object (T1128631)

pomahtri:


sergepilipchuk (approved):

sergepilipchuk (approved):

js/common/grids.d.ts
```diff
@@ -519,7 +519,7 @@ export type ColumnChooserMode = 'dragAndDrop' | 'select';
  * @namespace DevExpress.common.grids
  */
 export type ColumnCustomizeTextArg = {
-  value?: string | number | Date;
+  value?: any;
```
IlyaKhD, line 522: Better add generic parameter instead:
```suggestion
  value?: TValue;
```
Is it correct to pass in `ColumnBase`'s `TRowData` ?

IlyaKhD (approved):

js/common/grids.d.ts
IlyaKhD: Discussed in private, we gonna try fix all such cases in the future

Alyar666 (approved):

PR #23635, opened 2023-01-09
JQuery theme example

MikeVitik:
Demonstrate global theming and scoping theming in jQuery.

### DIFF
--- a/playgrounds/jquery/theme-example.html
+++ b/playgrounds/jquery/theme-example.html
@@ -1,0 +1,32 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <title>jQuery app</title>
+    <link rel="stylesheet" href="../../packages/styles/lib/themes/dark-scoped.css" />
+    <link rel="stylesheet" href="../../packages/styles/lib/themes/light.css" />
+</head>
+
+<body>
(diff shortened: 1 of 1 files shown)


playgrounds/jquery/theme-example.html
```diff
@@ -0,0 +1,32 @@
...
+    <link rel="stylesheet" href="../../packages/styles/lib/themes/light.css" />
+</head>
+
+<body>
+    <script src="../../packages/jquery/lib/bundle.debug.js" charset="utf-8"></script>
+    <span>Component styled by global theme (the last css link to the styles/lib/themes/*.css)</span>
+    <div id="cmp">
+        <script>
+            (function() {
+                new window.DevExpress.ui.dxRadioButton(document.getElementById('cmp'), { value: false, label: 'Check me' })
```
const314, line 16: Minor: Since this is an example, it is better to also pass checked or defaultChecked props and set value to some string instead of boolean. Otherwise, it looks a bit confusing, one can think that value sets checked state.

const314 (approved):

playgrounds/jquery/theme-example.html
IlyaKhD: I support the agreement

IlyaKhD (approved):

williamvinogradov (approved):

PR #23649, opened 2023-01-10
TreeView: fix height calculations when searchEnabled is true in Material theme (T1138605)

marker-dao:


scss/widgets/material/treeView/_index.scss
```diff
@@ -18,7 +18,16 @@ $material-treeview-container-padding:
   $material-treeview-border-padding
   $material-treeview-container-with-border-offset;
 
-$material-treeview-search-editor-height: 33px;
+$material-treeview-search-editor-height: null !default;
+
+@if $size == "default" {
+  $material-treeview-search-editor-height: 48px;
```
ksercs: i think we should move it to "sizes" file

testing/testcafe/tests/navigation/treeView/common.ts
```diff
@@ -5,10 +5,40 @@ import createWidget from '../../../helpers/createWidget';
 // eslint-disable-next-line import/extensions
 import { employees } from './data.js';
 import { setAttribute } from '../../../helpers/domUtils';
+import TreeView from '../../../model/treeView';
 
-fixture.disablePageReloads`TreeView_selectAll`
+fixture.disablePageReloads`TreeView`
   .page(url(__dirname, '../../container.html'));
 
+test('TreeView: the height calculates incorrectly when searchEnabled is true (T1138605)', async (t) => {
```
ksercs:
```suggestion
test('TreeView: height should be calculated correctly when searchEnabled is true (T1138605)', async (t) => {
```

scss/widgets/material/treeView/_index.scss
```diff
@@ -18,7 +18,16 @@ $material-treeview-container-padding:
   $material-treeview-border-padding
   $material-treeview-container-with-border-offset;
 
-$material-treeview-search-editor-height: 33px;
+$material-treeview-search-editor-height: null !default;
```
EugeniyKiyashko: I think we shouldn't use the constant sizes for this purpose. Use display-flex for parent container that includes searchEditor & scrollable container. After that remove all constants that store this magic height sizes. This approach must work for all themes and sizes well.

marker-dao: Your choice is a great solution. I saw it in another ticket. But it is worth considering the established principles of development in the current version of the design system. In this case, we do not use automatic height calculations in any of the themes (generic for example). The task you are talking about is refactoring. This problem must be solved when developing a new design system.

In this pull request, I tried not to make drastic changes.

If you still vote to remove the constants and rewrite the component to display-flex, let me know.

@EugeniyKiyashko, @ksercs

ksercs: I think using of constants is ok at the moment

EugeniyKiyashko: It allows us to make our code clear and remove a magic numbers and excess styles. But... well. It is advice only to chose the better way.

EugeniyKiyashko (dismissed):

scss/widgets/material/treeView/_index.scss
marker-dao: https://github.com/DevExpress/DevExtreme/pull/23649/commits/99f3c58b1d0e9057a7f93892f817ef25839b119c

ksercs (approved):

PR #23662, opened 2023-01-11
Renovation: Basic React form spike

const314:
Basic React form implemented in the playground, two different approaches: with HOC and without it.
Supports FormItem children layout rendering, calls validation on submit and valueChange.
Also includes separate hooks and context for working with validation.


packages/react/src/types.ts
```diff
@@ -0,0 +1,6 @@
+import { ReadonlyProps, TemplateProps, ValueProps } from '@devextreme/components';
+import { Props } from './internal/props';
+
+export interface EditorProps<T> extends Props<ValueProps<T>, ReadonlyProps, TemplateProps> {
```
williamvinogradov: Let's move this type to ```packages/react/src/internal/props.ts```
Because this type will be used by other editor components and in my point of view, we should add it to the internal part of our React codebase :)

playgrounds/react/src/examples/form/components/hooks/use-form-item-layout.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const {
+    label, hint, editor, rules,
+  } = useMemo<FormItemChildrenInfo>(() => {
+    const childrenArray = Children.toArray(children);
+
+    return {
+      label: findNodeByTypes(childrenArray, [FormItemLabel]),
+      editor: findNodeByTypes(childrenArray, editorTypes),
+      hint: findNodeByTypes(childrenArray, [FormItemHint]),
+      rules: filterNodesByTypes(childrenArray, [CustomRule, CustomRule1]) || [],
```
williamvinogradov: Question.

Maybe one common component for the rules is more suitable here?
Because as far as I understand it clearly, the form item can work only with the rule components types mentioned here.
So, if our client wants to create a custom rule by themself, it must use one of the supported rule types.

playgrounds/react/src/examples/form/components/form.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const result: string[] = [];
+    if (rules?.length) {
+      rules.forEach((rule) => {
+        if (!rule.validate(value)) {
+          result.push(rule.message);
+        }
+      });
+    }
+    return result;
+  };
```
williamvinogradov: Minor suggestion.

Let's make it shorter:
```suggestion
  const validateFormItemValue: FormItemValidator = (
    value, rules,
  ) => rules?.filter(({ validate }) => !validate(value)).map(({ message }) => message);
```

williamvinogradov (approved):

playgrounds/react/src/examples/form/components/form-context.ts
```diff
@@ -0,0 +1,12 @@
+import { createContext } from 'react';
+import { FormValidationResult, Rule } from './types';
+
+interface FormContextValue {
```
MikeVitik:
```suggestion
/*
The validation-related props should move to the future validation context because the validation can be used separately from the form. Of course, the form should create validation context implicitly. 
*/
interface FormContextValue {
```

playgrounds/react/src/examples/form/components/form.tsx
```diff
@@ -0,0 +1,73 @@
+import {
+  FormEventHandler, useMemo, useRef, useState,
+} from 'react';
+import { FormContext } from './form-context';
+import {
+  FormItemValidator, FormProps, FormValidationResult, Rule,
+} from './types';
+
+export function Form({ children, onSubmit }: FormProps) {
```
MikeVitik, line 17:
```suggestion
/*
Vitik: The previous form implements parts:
- formData collection
- item layout: a label for the editor, groups, columns, and tab panels,
- editor factory base on dataType 
- editors validation
The previous form doesn't allow the use of them separately the next form should allow this.
*/

export function Form({ children, onSubmit }: FormProps) {
```

MikeVitik (approved):

playgrounds/react/src/examples/form/components/form-item-hoc-approach/with-form-context.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { FormContext } from '../form-context';
+
+export function withFormContext<T>(Editor: ComponentType<EditorProps<T>>) {
+  function FormEditor(props: EditorProps<T>) {
+    const formContext = useContext(FormContext);
+    useEffect(() => {
+      const initialValue = props.value || props.defaultValue;
+      if (props.name) {
+        formContext?.onValueChanged(props.name, initialValue);
+      }
```
IlyaKhD: **Minor**:
```suggestion
      if (props.name) {
        formContext?.onValueChanged(props.name, props.value || props.defaultValue);
      }
```

playgrounds/react/src/examples/form/components/hooks/use-form-item-layout.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function useFormItemLayout(children: ReactNode, editorTypes: JSXElementConstructor<any>[]) {
+  const {
+    label, hint, editor, rules,
+  } = useMemo<FormItemChildrenInfo>(() => {
+    const childrenArray = Children.toArray(children);
+
+    return {
+      label: findNodeByTypes(childrenArray, [FormItemLabel]),
```
IlyaKhD, line 22: **Minor**:
It seems that we could extract props from components here instead of doing it in the `useFormItemRulesInitialization` function

packages/react/src/components/radio-group/radio-group.tsx
```diff
@@ -30,14 +32,23 @@ function RadioGroupInternal<T>(props: RadioGroupProps<T>) {
...
     <RadioGroupContext.Provider value={radioGroupCore}>
       <div>
-        {props.children}
+        {props.name
+          ? Children.map(
+            props.children,
+            child => (
+              isValidElement<EditorProps<T>>(child)
+                ? cloneElement(child, { name: props.name })
+                : child
```
IlyaKhD, line 39: **Question**:
Could we use context-based approach here? So that children get `name` from RadioGroupContext?

IlyaKhD (approved):

packages/react/src/types.ts
const314: fixed

playgrounds/react/src/examples/form/components/form.tsx
const314: Good point. Fixed.

packages/react/src/components/radio-group/radio-group.tsx
const314: Yes, but in this case, we'll need to store the name in the core as well

playgrounds/react/src/examples/form/components/hooks/use-form-item-layout.tsx
const314: I think it is even better to move the rule component filtration along with the rule extraction logic into a separate validation-related hook.

playgrounds/react/src/examples/form/components/hooks/use-form-item-layout.tsx
const314: Ideally, it should be filtered by the base type of the child's props, which should the same for all validation rules. I'll try to implement it in a separate PR

playgrounds/react/src/examples/form/components/hooks/use-form-item-layout.tsx
const314: Done

playgrounds/react/src/examples/form/components/form.tsx
```diff
@@ -65,10 +45,12 @@ export function Form({ children, onSubmit }: FormProps) {
 
   return (
     <FormContext.Provider value={formContextValue}>
-      <form onSubmit={handleSubmit}>
-        {children}
-        <input type="submit" value="Submit" />
-      </form>
+      <ValidationContext.Provider value={validationContextValue}>
```
MikeVitik, line 48: Suggestion to the next version of validation.
Move ValidationContext.Provider into new Validator component and use it here.

MikeVitik (approved):

PR #23679, opened 2023-01-12
DataGrid - Columns have incorrect width on the first render  after updating from 22.1.4 to 22.1.6 if renderAsync and columnAutoWidth are set to "true" (T1126234)

pomahtri:


js/ui/grid_core/ui.grid_core.grid_view.js
```diff
@@ -64,8 +64,7 @@ const ResizingController = modules.ViewController.inherit({
             this._refreshSizesHandler = (e) => {
                 dataController.changed.remove(this._refreshSizesHandler);
 
-                const templateDeferreds = e && e.templateDeferreds || [];
-                when.apply(this, templateDeferreds).done(() => {
+                this.getView('rowsView')._waitAsyncTemplates(e || {}).done(() => {
```
Alyar666: I think it's better to use a declared property: this._rowsView.
And make the `_waitAsyncTemplates` method public (remove prefix`_`)

Alyar666 (changes requested):

Alyar666 (approved):

PR #23717, opened 2023-01-16
Renovation: Replace view models with selectors + simplify the core 

williamvinogradov:
Simplified the view model part of the core package.
Got rid of unused and complicated from the architectural point of view code.

In details:
 - Renamed  the ```StateManager``` entity into the ```Store``` entity and merged it with ```State``` entity.
- Made changes connected with core package changes in all dependent packages (```components``` / ```react``` / ```angular```)


packages/angular/projects/angular/src/internal/rxjs/use-store-selector.ts
```diff
@@ -0,0 +1,20 @@
+import { Selector, StoreState, UnknownRecord } from '@devextreme/core';
+import { finalize, Observable, startWith } from 'rxjs';
+
+export function useStoreSelector<
```
MikeVitik, line 4: Minor suggestion, Mind sync, Not required to response
useStoreSelector is just a transparent wrapper of core observable into the rx observable
May be we can directly use rx observale as a parameter in the createStore and doesnt need such useStoreSelector

MikeVitik (approved):

const314 (approved):

packages/angular/projects/angular/src/internal/rxjs/use-store-selector.ts
williamvinogradov: Yeah, it's just a converter from core observable to rxjs observable.

But, this way is the best practice to do it because it's similar to creating the new rxjs observable from callback :)

PR #23718, opened 2023-01-16
Renovation:  Replace the reducer via universal addUpdate

williamvinogradov:
Replaced reducer via the ```addUpdate``` universal method.

In details:
 - Merged ```Dispatcher``` with ```addUpdate``` method into one ```addUpdate``` method with different ```commit update``` strategies.
Also, got rid of ```Reducer``` because it's become redundant.
- Made changes connected with core package changes in all dependent packages (```components``` / ```react``` / ```angular```)
- Improved typings of the Angular base context service.


packages/core/src/store.ts
```diff
@@ -1,44 +1,33 @@
...
   subscribe: SubscribeFunc<TState>;
-}
-
-export interface Store<
-  TState extends UnknownRecord,
-  THandlers extends Handlers<TState>,
-  > extends StoreState<TState> {
-  addUpdate(statePart: Partial<TState>): void;
+  addUpdate(updateAction: UpdateStateAction<TState>): void;
+  commitPropsUpdates(): void;
```
const314, line 13: Minor: Two commit methods still look a bit confusing to me, it is not clear when to call each of them. Maybe we should think about how to make usage/naming more obvious.

const314 (approved):

packages/core/src/store.ts
williamvinogradov: Partially agree with you, but let's think about it separately

packages/core/src/store.ts
```diff
@@ -65,41 +54,33 @@ export function createStore<
...
       ...nextState,
-      ...statePart,
+      ...updateAction(nextState),
     };
   };
 
-  const commitUpdates = () => {
+  const commitPropsUpdates = () => {
     currentState = nextState;
     baseCommitUpdate(validator(currentState));
```
MikeVitik, line 67: **Clarification, possible code changes**
Should we really call baseCommitUpdate?
I suppose it's called then we get new props and should sync it with a store. Isn't it? So if yes why do we call baseCommitUpdate what related to changes store by action and implement controlled mode?

williamvinogradov: We discussed it in a live conversation.
And we decided to discuss this store updating mechanism in the planned meeting about the controlled mode.

MikeVitik (approved):

PR #23719, opened 2023-01-16
[Blocked by #23718] Renovation: Refactor the core package tests + add new ones

williamvinogradov:
Added unit tests for new added entities (```selector``` & ```store```)
And unified test case names + added the coverage report to jest.config & .gitignore

### DIFF
--- a/packages/core/src/__test__/selector.test.ts
+++ b/packages/core/src/__test__/selector.test.ts
@@ -1,0 +1,66 @@
+import { createSelector } from '../selector';
+import { memoize, shallowComparer } from '../utils';
+
+jest.mock('../utils');
+
+const memoizeMock = jest.mocked(memoize);
+memoizeMock.mockReturnValue(jest.fn);
+
+describe('core', () => {
+  describe('selector', () => {
+    describe('createSelector', () => {
+      it('passes arguments to memoize', () => {
+        const func = jest.fn();
+        const comparer = jest.fn();
+
+        createSelector(func, jest.fn(), comparer);
+
+        expect(memoizeMock).toBeCalledTimes(1);
+        expect(memoizeMock).toBeCalledWith(func, comparer);
+      });
+
+      it('uses shallowComparer by default', () => {
+        createSelector(jest.fn(), jest.fn());
+
+        expect(memoizeMock).toBeCalledWith(expect.anything(), shallowComparer);
+      });
+
+      it('returns memoized', () => {
+        const cachedValue = {};
+        const cached = jest.fn().mockReturnValue(cachedValue);
+        memoizeMock.mockReturnValue(cached);
+
+        const selector = createSelector(jest.fn(), jest.fn(), jest.fn());
+
+        expect(memoizeMock).toBeCalledTimes(1);
+        expect(selector({})).toBe(cachedValue);
+      });
+
+      it('passes params to memoized', () => {
+        const param = {};
+        const getParams = jest.fn().mockReturnValue(param);
+        const cached = jest.fn();
+        memoizeMock.mockReturnValue(cached);
+
+        const selector = createSelector(jest.fn(), getParams, jest.fn());
+        selector({});
+
+        expect(cached).toBeCalledTimes(1);
+        expect(cached).toBeCalledWith(param);
+      });
+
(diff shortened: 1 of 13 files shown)


const314 (approved):

packages/core/src/__test__/selector.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        expect(memoizeMock).toBeCalledTimes(1);
+        expect(memoizeMock).toBeCalledWith(func, comparer);
+      });
+
+      it('uses shallowComparer by default', () => {
+        createSelector(jest.fn(), jest.fn());
+
+        expect(memoizeMock).toBeCalledWith(expect.anything(), shallowComparer);
+      });
```
IlyaKhD: This test seems to be redundant, as it checks memoize implementation

packages/core/src/__test__/selector.test.ts
```diff
@@ -0,0 +1,66 @@
+import { createSelector } from '../selector';
+import { memoize, shallowComparer } from '../utils';
+
+jest.mock('../utils');
+
+const memoizeMock = jest.mocked(memoize);
+memoizeMock.mockReturnValue(jest.fn);
+
+describe('core', () => {
```
IlyaKhD: I suggest trying to rewrite this test suit: it should consider `createSelector` as a black box checking it's output. Instead it gets `createSelector` implementation and tests how functions calls each other

packages/core/src/middlewares/__test__/callbacks-middleware.test.ts
```diff
@@ -4,44 +4,48 @@ import { callbacksMiddleware } from '../callbacks-middleware';
...
-    const next = { a: 4, b: 5, c: 6 };
-    const config = {};
-    getChangedKeysMock.mockReturnValue(['a', 'b', 'c']);
-
-    const funcArray = callbacksMiddleware(prev, next, config);
-
-    expect(funcArray.length).toEqual(0);
+describe('core', () => {
+  describe('middlewares', () => {
+    describe('callbacksMiddleware', () => {
```
IlyaKhD: **Question**
What for do we need this explicit nesting?

packages/core/src/__test__/selector.test.ts
IlyaKhD: I mean, the whole suit seems to be useless, like we don't have tests for `createSelector` at all

packages/core/src/__test__/selector.test.ts
williamvinogradov: I don't agree with you
This test checks that we pass shallow compare to memorize inside tested function by default.
And we don't check the memoize implementation here because memoize function is mocked :)

packages/core/src/__test__/selector.test.ts
williamvinogradov: I just moved the selector code and its tests from the existing ViewModel code.
So, this test isn't new and just was moved. (I added the last one only, which checks that we pass to memoize shallowComparer by default).
I'm confused about your comment because you developed these test cases.
[Link to your commit](https://github.com/DevExpress/DevExtreme/commit/74bef39f3398616b7f89357a44067f8c5a58c574#diff-b42bc75648481eb41f48c321dbdfdac6f7034d2e30c71c93fe0686174003457bR138)

I'm not against of change these test cases, but can you please provide an example of what you want?

IlyaKhD: image

IlyaKhD: Okay then, if these are not new tests we should keep them as is.

packages/core/src/__test__/selector.test.ts
IlyaKhD: I see, it's for default parameter value. It's OK, you're right.

IlyaKhD (approved):

packages/core/src/middlewares/__test__/callbacks-middleware.test.ts
williamvinogradov: Oh, it's my mistake
I thought that this describe nesting will help organize how IDE shows test cases.
But I was wrong, It's redundant.

Thanks for the comment!

PR #23750, opened 2023-01-20
DataGrid - The `TypeError: this._contentReadyAction is not a function` is thrown after widget disposed if renderAsync property is set to `true` (T1138639)

MikeVitik:
Add clearTimeout into the dispose method for removing scheduled template processing
OnRender can be called then component disposed so add the _disposing flag check to prevent error.


MikeVitik:
Add clearTimeout into dispose for disposing template timeout
OnRender also possible can be called then component disposed so add _disposing flag check to prevent error.

pomahtri (approved):

Alyar666 (approved):

PR #23761, opened 2023-01-24
Add R2 Vue Wrappers

IlyaKhD:


MikeVitik (approved):

playgrounds/vue/.gitignore
```diff
@@ -0,0 +1,28 @@
+# Logs
```
williamvinogradov: Minor suggestion.

Do we need separate ```.gitignore``` in the Vue playground project?
It looks like it was auto-generated
Maybe we should delete it?

playgrounds/vue/README.md
```diff
@@ -0,0 +1,40 @@
+# vue
```
williamvinogradov: Minor suggestion.

It looks like the auto-generated readme.
It doesn't useful in our current mono repo structure.
Maybe we should change or delete it? :)

playgrounds/vue/src/assets/base.css
```diff
@@ -0,0 +1,74 @@
+/* color palette from <https://github.com/vuejs/theme> */
```
williamvinogradov: Minor suggestion.

This CSS was auto-generated by default.
We don't need it and it may confuse someone in the future.
Let's delete it.

playgrounds/vue/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
williamvinogradov, line 1: Question.

Why do we've two ```tsconfig``` files here?

williamvinogradov (approved):

playgrounds/vue/.gitignore
IlyaKhD: done

playgrounds/vue/README.md
IlyaKhD: done

playgrounds/vue/src/assets/base.css
IlyaKhD: done

packages/vue/src/radio-button.ts
```diff
@@ -0,0 +1,58 @@
+import RadioButton, { Properties } from "../../jquery/lib/esm/radio-button";
+import { createComponent } from "devextreme-vue/core/index";
+
+type AccessibleOptions = Pick<Properties,
```
const314, line 4: Minor: The type name makes me think that it is somehow related to accessibility (a11y), which can be a bit confusing.

playgrounds/vue/tsconfig.json
IlyaKhD: Vue CLI has created them. As I see, one extends `@vue/tsconfig/tsconfig.node.json` and the other - `@vue/tsconfig/tsconfig.web.json`

I think it's better to leave it as is for our dev-ops guys.

const314 (approved):

packages/vue/src/radio-button.ts
IlyaKhD: I agree that the naming is a bit confusing.
This file was generated by the vue-generator tool, we can modify it, but later.

PR #23774, opened 2023-01-25
DropDownBox, TagBox: add keyboard navigation tests

AlexanderMoiseev:


testing/tests/DevExpress.ui.widgets.editors/dropDownBox.tests.js
```diff
@@ -27,6 +27,8 @@ QUnit.testStart(() => {
     $('#qunit-fixture').addClass('qunit-fixture-visible');
 });
 
+const OPENED_OPTION_NAME = 'opened';
```
ksercs: i don't think this constant simplifies the code, vice versa, worsens the readability. Let's get rid of this

testing/tests/DevExpress.ui.widgets.editors/dropDownBox.tests.js
```diff
@@ -765,6 +767,19 @@ QUnit.module('popup options', moduleConfig, () => {
 });
 
 QUnit.module('keyboard navigation', moduleConfig, () => {
+    QUnit.test('alt+down should open dropDownBox', function(assert) {
```
ksercs, line 768: Unit test should be atomic: one test = 1 behavior check. U named this test as "alt+down should open dropDownBox", but u test 3 different points inside of it)

Let's leave only the third assert, but create a separate test if we need to test that dropDownBox is not opened after down key is pressed

testing/tests/DevExpress.ui.widgets.editors/dropDownBox.tests.js
```diff
@@ -765,6 +767,19 @@ QUnit.module('popup options', moduleConfig, () => {
...
+        const instance = new DropDownBox(this.$element);
+        assert.notOk(instance.option(OPENED_OPTION_NAME), 'dropDownBox closed by default');
+
+        const $input = this.$element.find(`.${TEXTEDITOR_INPUT_CLASS}`);
+
+        keyboardMock($input).keyDown('down', { altKey: false });
+        assert.notOk(instance.option(OPENED_OPTION_NAME), 'dropDownBox not opened after down key pressed');
+
+        keyboardMock($input).keyDown('down', { altKey: true });
+        assert.ok(instance.option(OPENED_OPTION_NAME), 'dropDownBox opened after alt+down pressed');
```
ksercs:
```suggestion
        assert.ok(instance.option(OPENED_OPTION_NAME), 'dropDownBox is opened after alt+down is pressed');
```

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -2180,6 +2180,35 @@ QUnit.module('keyboard navigation', {
         fx.off = false;
     }
 }, () => {
+    QUnit.test('navigation keys should focus items', function(assert) {
```
ksercs: 1. I don't see a reason to write several asserts for the single key. 
2. This test name is too common for me: home/end, arrowLeft/arrowRight are "navigation keys" too. This is common because this test is not atomic. Let's create 2 tests: "up/down key press should move tag focus" and "pagedown/pageup key press should move tag focus"

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -2180,6 +2180,35 @@ QUnit.module('keyboard navigation', {
...
+
+        assertFocusedItem(assert, this.keyboard, 'down', $listItems, 1);
+        assertFocusedItem(assert, this.keyboard, 'down', $listItems, 2);
+        assertFocusedItem(assert, this.keyboard, 'up', $listItems, 1);
+        assertFocusedItem(assert, this.keyboard, 'up', $listItems, 0);
+        assertFocusedItem(assert, this.keyboard, 'pagedown', $listItems, 2);
+        assertFocusedItem(assert, this.keyboard, 'pageup', $listItems, 0);
+    });
+
+    function assertFocusedItem(assert, keyboard, key, listItems, expectedFocusedItemIndex) {
```
ksercs: 1. We can simplify signature if this function If we move it to "beforeEach" and save as an instance's method. Then we don't need to pass `keyboard` and `listItems` props
2. This function's name is not complete. We don't only assert focused item, we also press a key here. So it should be named as `pressKeyAndAssertFocusedItem`. Here u can already see that there is an other issue - this function is not single responsible. It's better to make an action and an assert separately.

So, i suggest u get rid of this function and write this test imperatively. Yeap, there would be some copypaste, but the tests would be easier to read (especially, after splitting).

```javascript
QUnit.test('down/up key press should move item focus', function(assert) {
       const $listItems = getListItems(this.instance);
       const $firstItem = $listItems.eq(0);
       const $secondItem = $listItems.eq(1);
       
       this.keyboard.press('down');
       assert.strictEqual($secondItem.hasClass(FOCUSED_CLASS), 'second item is focused');

       this.keyboard.press('up');
       assert.strictEqual($firstItem.hasClass(FOCUSED_CLASS), 'first item is focused');
});
```

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -2297,10 +2326,13 @@ QUnit.module('keyboard navigation', {
 
         prevented = 0;
         this.instance.option('opened', true);
+        assert.strictEqual(this.instance.option('opened'), true);
```
ksercs: Why did u add this asserts? It seems to be excess here

AlexanderMoiseev: i was thinking it'd be nice to check if popup is opened before closing it, but yeah, don't think we should check if 'opened' option works correctly, sure it is already tested

testing/tests/DevExpress.ui.widgets.editors/dropDownBox.tests.js
AlexanderMoiseev: i dunno, checking if dropdown is closed before opening it and then checking if it's opened after some action seems natural to me, both assertions still stay in the context of the test as far as i'm concerned. I'm talking about the first assertion, don't see any harm in it. But ok, single responsibility it is then)

testing/tests/DevExpress.ui.widgets.editors/dropDownBox.tests.js
AlexanderMoiseev: yeah, actually since option name doesn't change often it's ok to leave it as it is

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
AlexanderMoiseev: yep, ok, thought about it

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
AlexanderMoiseev: I added separate test for testing esc button -  'escape key should close popup'

Zedwag (approved):

ksercs (approved):

PR #23775, opened 2023-01-26
R2 React: base component functionality in the new RG.

williamvinogradov:
Implemented the necessary functionality of the base widget for the new RG component.
And made this functionality as shared as possible for future uses in the new components.


packages/components/src/radio-group/index.ts
```diff
@@ -32,15 +38,60 @@ function updateValueAction<T>(
   });
 }
 
+function updateUiAction<T>(
```
MikeVitik: Question: Suggestion: Minor: Should we rename updateUiAction to updateUIAction or maybe to updateRootContainerAction because UI is very common and doesn't provide necessary specifics?

packages/components/src/root-container/props.ts
```diff
@@ -0,0 +1,232 @@
...
+
+// --- hint ---
+export type HintDomOptions = {
+  hint?: string;
+};
+
+export function buildHintProp(
+  { hint }: HintDomOptions,
+): PipeFunc<RootContainerProps> {
+  return buildAttributesProp(!!hint, {
```
MikeVitik: I set a breakpoint here and it's called many times then any prop is changed for example I change disable in the react playground

packages/react/src/compatible-components/radio-group/radio-group.tsx
```diff
@@ -43,19 +44,20 @@ export function RadioGroupCompatible<T>({
...
-        return (
-          <RadioButton
-            key={key}
-            value={value}
-            label={renderLabel(item, index)}
-          />
-        );
-      })}
-    </RadioGroup>
+    <div>
```
MikeVitik: Question: Why do we need a new div here?

MikeVitik (changes requested):

packages/react/src/internal/components/root-container.tsx
```diff
@@ -0,0 +1,85 @@
...
+export type FocusableHocProps =
+  PropsWithChildren<RootContainerComponentProps & FocusableProps>;
+
+export type RootContainerComponentProps = {
+  domProps: RootContainerProps,
+  className?: string,
+  attributes?: AnyRecord,
+  children?: ReactNode,
+};
+export const RootContainer = forwardRef(RootContainerRender);
```
const314: Minor: The RootContainer name looks a bit misleading to me, it is like there is a single root container for the whole page. Since it is used to wrap editors, maybe we should name them EditorContainer and FocusableEditorContainer or something like that?

packages/react/src/components/radio-group/radio-group.tsx
```diff
@@ -1,35 +1,87 @@
...
+    if (isFocusControlled) {
+      store.addUpdate(RADIO_GROUP_ACTIONS.updateFocus(getFocusValue(props.focus, props.disabled)));
+    }
+
+    const updateUiValue = getRootContainerDomOptionsFromProps(props);
+    store.addUpdate(RADIO_GROUP_ACTIONS.updateUi(updateUiValue));
 
     store.commitPropsUpdates();
-  }, [props.value]);
+  }, [props]);
```
const314: The effect will be called on every prop change, so, even if the value has not changed, the updateValue will be called. It is better to split it into several separate useEffect hooks, one for value, one for focus and one for UI props.

packages/components/src/root-container/props.ts
williamvinogradov: Yes, unfortunately we've issue with cached selectors (they don't work as expected because React recreate them on each render)

I fixed it in the [separate PR](https://github.com/DevExpress/DevExtreme/pull/23809)

packages/components/src/common/index.ts
```diff
@@ -0,0 +1,3 @@
+export type FocusValueProps = {
```
IlyaKhD: **Minor**
```suggestion
export type FocusProps = {
```
or smth like

packages/react/src/components/radio-group/radio-group.tsx
```diff
@@ -1,35 +1,87 @@
...
+import { useCallbackRef, useSecondEffect, useStoreSelector } from '../../internal/hooks';
+import { EditorProps, RootContainerProps } from '../../internal/props';
 import { RadioGroupStoreContext } from '../radio-common';
 
+import '@devextreme/styles/src/radio-group/radio-group.scss';
+
 function RadioGroupInternal<T>(props: RadioGroupProps<T>) {
-  const controlledMode = useMemo(() => Object.hasOwnProperty.call(props, 'value'), []);
+  const isValueControlled = useMemo(() => Object.hasOwnProperty.call(props, 'value'), []);
+  const isFocusControlled = useMemo(() => Object.hasOwnProperty.call(props, 'focus'), []);
```
IlyaKhD: **Question**
What if `value` or `focus` is added after component was initialised? These props should become controlled, shouldn't they?

packages/components/src/root-container/props.ts
```diff
@@ -0,0 +1,232 @@
...
+}
+
+export function getAccessKeyFromProps(
+  props: Partial<AccessKeyDomOptions> = {},
+): AccessKeyDomOptions {
+  return { shortcutKey: props?.shortcutKey };
+}
+
+// --- active-state ---
+export type ActiveStateDomOptions = {
```
IlyaKhD: **Minor**
If we do not use this type, I would better replace it with `RootContainerDomOptions['activeStateEnabled']`

Same for other types here, it's more convenient to have all theses properties in a single 'bag'

packages/components/src/radio-group/index.ts
williamvinogradov: Renamed this action :)

packages/components/src/root-container/props.ts
williamvinogradov: Ok, in the latest version of the PR, all types exist inside one parent type.

packages/react/src/components/radio-group/radio-group.tsx
williamvinogradov: No, we check that the prop is controlled or uncontrolled only once on the component mounting (first render).

After that prop mode cannot be changed.

packages/components/src/common/index.ts
williamvinogradov: This code is gone :)

packages/react/src/internal/components/root-container.tsx
williamvinogradov: This code is gone :)

packages/react/src/compatible-components/radio-group/radio-group.tsx
williamvinogradov: Removed

packages/react/src/components/radio-group/radio-group.tsx
williamvinogradov: I tried to fix it in the latest version of the PR.
Please take a look :)

packages/react/src/internal/props.ts
```diff
@@ -20,3 +24,8 @@ export type Props<
 export interface EditorProps<T> extends Props<ValueProps<T>, ReadonlyProps, TemplateProps> {
   name?: string
 }
+
+export type FocusableComponent = {
```
MikeVitik, line 29: If onFocus and onBlur is a dom like event can we use typing from DOM?

packages/components/src/radio-group/index.ts
```diff
@@ -33,17 +55,41 @@ function updateValueAction<T>(
...
+}
+
+export function createContainerPropsSelector<T>()
+: Selector<RadioGroupState<T>, RootContainerProps> {
+  return createSelector(
+    (state) => state.readonly,
+    (
+      readonly: ReadonlyProps,
+    ) => RADIO_GROUP_CONTAINER_PROPS_BUILDER.build(readonly)(DEFAULT_ROOT_CONTAINER_PROPS),
   );
```
MikeVitik: Can we rewrite it:
```
  const propsMapper = RADIO_GROUP_CONTAINER_PROPS_BUILDER.build(DEFAULT_ROOT_CONTAINER_PROPS)
  return createSelector(
    (state) => state.readonly,
    (
      readonly: ReadonlyProps,
    ) => propsMapper(readonly),
  );
```

const314 (approved):

packages/components/src/root-container/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+        cssClass: [
+          ...props.cssClass,
+          className,
+        ],
+      }
+      : props
+  );
+}
+
+export function buildAttributes(
```
MikeVitik: Question: Why it named build? It looks like conditional map?

packages/components/src/root-container/prop-builders.ts
```diff
@@ -0,0 +1,57 @@
...
+      { attributes }: DomOptions['attributes'],
+    ) => buildAttributes(true, attributes),
+  ),
+  disabled: createBuilder(
+    DEFAULT_DOM_OPTIONS.disabled,
+    (
+      { disabled }: DomOptions['disabled'],
+    ) => buildAttributes(disabled, {
+      [DOM_ATTRIBUTES.disabled]: ' ',
+      [DOM_ATTRIBUTES.tabIndex]: undefined,
```
MikeVitik: Question: What is happen if we set tabIndex and disabled simultaniosly? Should we add some additional logic here?

packages/components/src/radio-group/index.ts
```diff
@@ -5,22 +5,44 @@ import {
...
-import { UpdateStateAction } from '@devextreme/core/src/store';
+import {
+  DEFAULT_ROOT_CONTAINER_PROPS,
+  ROOT_CONTAINER_PROP_BUILDERS,
+  RootContainerDomOptions,
+  RootContainerProps,
+} from '../root-container';
+
+// === base component ===
+export type RadioGroupContainerDomOptions = RootContainerDomOptions['accessKey']
```
MikeVitik: Suggestion:
More type safer variant (changing in chain directly affected to the type)
```
export type RadioGroupContainerDomOptions = typeof RADIO_GROUP_CONTAINER_PROPS_BUILDER
```
Remove export because we have already aliase this type to ReadonlyProps

packages/components/src/root-container/types.ts
```diff
@@ -0,0 +1,50 @@
+import { PipeFunc, UnknownRecord } from '@devextreme/core';
+
+export type DomAttributes = Record<string, unknown>;
+
+export type Props = {
+  cssClass: string[];
+  attributes: DomAttributes,
+};
+
+export type DomOptions = {
```
MikeVitik, line 10: Question: Is it propsToDomAttributesMap?

packages/components/src/root-container/utils.ts
```diff
@@ -0,0 +1,101 @@
...
+        },
+      }
+      : props
+  );
+}
+
+export function createGetDomOptionsMethod<T extends UnknownRecord>(
+  defaultValue: T,
+): (params: Partial<T>) => T {
+  return (params) => getKeys(defaultValue).reduce(
```
MikeVitik, line 39: Question: Looks like we can replace it to the:
```
return (params) => ({...defaultValue, ...params})
```
Isn't it?

packages/react/src/components/radio-group/radio-group.tsx
```diff
@@ -1,35 +1,65 @@
...
-    if (controlledMode) {
+    if (isValueControlled) {
       store.addUpdate(RADIO_GROUP_ACTIONS.updateValue(props.value));
     }
 
+    store.addUpdate(RADIO_GROUP_ACTIONS.updateReadonly(readonlyValues));
+
     store.commitPropsUpdates();
-  }, [props.value]);
+  }, [props.value, ...Object.values(readonlyValues)]);
```
MikeVitik: Question: Do we really need useEffect here? May be it's a premature optimization? We have already have memozed selector that can solve problem of unnecessary mutations and already now the deps between state and ViewModel. Why we dublication knolodge about deps here?

playgrounds/react/src/examples/radio-group/radio-group-controlled-playground.tsx
```diff
@@ -0,0 +1,284 @@
...
+}
+
+interface PlaygroundState {
+  groupValue?: string;
+  buttons: PlaygroundBtnData[];
+  baseSettings: PlaygroundBaseDomSettings;
+  focused: boolean;
+}
+
+type Actions =
```
MikeVitik, line 26: Add setTabIndex for test it.

MikeVitik (approved):

packages/react/src/internal/props.ts
williamvinogradov: I suggest next typings:

```typescript
export type FocusableComponent = {
  onFocus?: FocusEventHandler<HTMLElement>;
  onBlur?: FocusEventHandler<HTMLDivElement>;
};
```

packages/components/src/root-container/utils.ts
williamvinogradov: Renamed to map

packages/components/src/radio-group/index.ts
williamvinogradov: Discussed and refactored

packages/react/src/components/radio-group/radio-group.tsx
williamvinogradov: Yes, we really need ```useEffect``` here :)

packages/components/src/radio-group/index.ts
williamvinogradov: Fixed

playgrounds/react/src/examples/radio-group/radio-group-controlled-playground.tsx
williamvinogradov: 👌

const314 (approved):

MikeVitik (approved):

PR #23789, opened 2023-01-27
HtmlEditor: ValueChanged event should be triggered once when value contains non optimized markup(T1137577)

alexanderPolosatov:


testing/tests/DevExpress.ui.widgets.form/formLayoutManager.markup.tests.js
```diff
@@ -3019,9 +3019,11 @@ QUnit.module('Supported editors', () => {
         const editor = layoutManager.getEditor('description');
         editor.option('value', 'new <b>value</b>');
 
-        assert.deepEqual(layoutManager.option('layoutData'), { description: 'new <b>value</b>' }, 'layoutData');
         if(windowUtils.hasWindow()) {
             assert.equal($('.dx-htmleditor-content').html(), '<p>new <strong>value</strong></p>', 'HtmlEditor content');
+            assert.deepEqual(layoutManager.option('layoutData'), { description: '<p>new <strong>value</strong></p>' }, 'layoutData');
```
ksercs, line 3024: why is this test changed?

testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/api.tests.js
```diff
@@ -348,11 +348,23 @@ testModule('API', moduleConfig, () => {
...
         assert.strictEqual(updateContentSpy.callCount, 2, 'value changed twice -> update content two times');
         assert.strictEqual(updateContentSpy.lastCall.args[0], 'New text', 'Update content with the new value');
     });
 
+    test('ValueChanged event should be triggered once when value contains non optimized markup(T1137577)', function(assert) {
+        this.createEditor();
+        const valueChangeStub = sinon.stub();
+
+        this.instance.on('valueChanged', valueChangeStub);
+        this.instance.option('value', '<p>after click</p><p></p>');
```
ksercs: let's make markup more meaningful

testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/api.tests.js
```diff
@@ -348,11 +348,23 @@ testModule('API', moduleConfig, () => {
...
+        this.clock.tick();
+
+        assert.strictEqual(valueChangeStub.callCount, 1);
+        assert.strictEqual(valueChangeStub.getCall(0).args[0].value, '<p>after click</p>');
+    });
+
```
ksercs, line 379: Let's also add a test for case when we change value to another one, but new value's optimized version is the same as the previous one's. For example, `<p>before click</p>` -> `<p>before click</p><p></p>`.

Maybe there are some other possible cases?

js/ui/html_editor/ui.html_editor.js
```diff
@@ -509,22 +509,39 @@ const HtmlEditor = Editor.inherit({
...
                 }
 
                 this._setSubmitValue(args.value);
 
-                this.callBase(args);
+                if(Boolean(needCallBase) === false && this._valueChangeCalled) {
+                    this._valueChangeCalled = false;
+                } else {
+                    this.callBase(callBaseArgs);
+                }
```
ksercs, line 533: cannot we simplify the solution? 2 private caches in instance + 2 new variables seems to be too much for such a problem.

Maybe we can write smth like this
```javascript
                        this._suppressValueChangeAction();
                        this._updateHtmlContent(updatedValue);
                        this._resumeValueChangeAction();
                        ...
                        
                        
                // NOTE: value can be optimized by Quill
                const value = this.option('value');
                if(value !== args.previousValue) {
                    this._setSubmitValue(value);
                    this.callBase({ ...args, value });
                }
```

testing/tests/DevExpress.ui.widgets.form/formLayoutManager.markup.tests.js
alexanderPolosatov: When we call `editor.option('value', 'new <b>value</b>');` Quill interpretate it as `<p>new <strong>value</strong></p>`. And in this case he had two calls **valueChanged** event. Firts call contains optimized markup, in last call markup non optimized. it leads to passing non actual value to layoutManager

ksercs: ok, thx

js/ui/html_editor/ui.html_editor.js
```diff
@@ -509,22 +509,30 @@ const HtmlEditor = Editor.inherit({
     },
 
     _optionChanged: function(args) {
+        let value;
```
ksercs: Why do we need define it here?

alexanderPolosatov: because we have [no-case-declarations](https://eslint.org/docs/latest/rules/no-case-declarations) rule. 
Fixed by wrapping into blocks using brackets.

ksercs (approved):

PR #23794, opened 2023-01-27
TagBox: fix tags selection when items are objects, valueExpr="this" and applyValueMode='useButtons' (T1141799)

AlexanderMoiseev:


ksercs:
@AlexanderMoiseev  
1) remove excess spaces in ID brackets plz. It'll simplify future searches
2) name is too common for me. I read it and it seems like tags selection does not work at all.. But it works, we fix only very specific case. Can we describe it in details?

js/core/utils/array.js
```diff
@@ -3,10 +3,16 @@ import { orderEach } from './object';
 import config from '../config';
 
 function createOccurrenceMap(array) {
-    return array.reduce((map, value) => {
-        map[value] = (map[value] ?? 0) + 1;
-        return map;
-    }, {});
+    return array.reduce((map, value) => map.set(value, (map.get(value) ?? 0) + 1), new Map());
```
ksercs: It is really difficult to read. I think that readability is more important than shortness
```suggestion
    return array.reduce((map, value) => {
        const count = (map.get(value) ?? 0) + 1;
        map.set(value, count);
        return map;
    }, new Map());
```

js/core/utils/array.js
```diff
@@ -3,10 +3,16 @@ import { orderEach } from './object';
...
+}
+
+function getFilteredArray(from, toRemove, getUniqueItems) {
+    const toRemoveMap = createOccurrenceMap(toRemove);
+    return from.filter(value => {
+        const occurrencesCount = toRemoveMap.get(value);
+        occurrencesCount && toRemoveMap.set(value, occurrencesCount - 1);
+        return getUniqueItems ^ occurrencesCount;
```
ksercs: To me this function only worsens a readability. Why did u decide to separate it?
- it increases count of code rows
- it needs to pass a non-intuitive true/false parameter
- it uses binary operator
- we don't see simple filtering in parent functions

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -5543,6 +5543,52 @@ QUnit.module('applyValueMode = \'useButtons\'', {
         assert.deepEqual(this.instance.option('value'), [items[2], items[0]], 'tags order is correct');
     });
 
+    QUnit.test('Object value should be updated correctly when item is added', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('value should be updated correctly after item is added if valueExpr="this" (T1141799)', function(assert) {
```

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -5543,6 +5543,52 @@ QUnit.module('applyValueMode = \'useButtons\'', {
...
+        $(this.$popupWrapper.find(`.${POPUP_DONE_BUTTON_CLASS}`)).trigger('dxclick');
+
+        const items = this.instance.option('value');
+        assert.equal(items.length, 3);
+        assert.deepEqual(items[0], firstValue);
+        assert.deepEqual(items[1], thirdValue);
+        assert.deepEqual(items[2], secondValue);
+    });
+
+    QUnit.test('Object value should be updated correctly when item is removed', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('value should be updated correctly after item is removed if valueExpr="this" (T1141799)', function(assert) {
```

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -5543,6 +5543,52 @@ QUnit.module('applyValueMode = \'useButtons\'', {
...
+            applyValueMode: 'useButtons',
+            displayExpr: 'description',
+            opened: true
+        });
+
+        $(this.$listItems.eq(1)).trigger('dxclick');
+        $(this.$popupWrapper.find(`.${POPUP_DONE_BUTTON_CLASS}`)).trigger('dxclick');
+
+        const items = this.instance.option('value');
+        assert.equal(items.length, 3);
```
ksercs:
```suggestion
        assert.strictEqual(items.length, 3);
```

testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -5543,6 +5543,52 @@ QUnit.module('applyValueMode = \'useButtons\'', {
...
+            applyValueMode: 'useButtons',
+            displayExpr: 'description',
+            opened: true
+        });
+
+        $(this.$listItems.eq(1)).trigger('dxclick');
+        $(this.$popupWrapper.find(`.${POPUP_DONE_BUTTON_CLASS}`)).trigger('dxclick');
+
+        const items = this.instance.option('value');
+        assert.equal(items.length, 2);
```
ksercs:
```suggestion
        assert.strictEqual(items.length, 2);
```

js/core/utils/array.js
AlexanderMoiseev: 1) "Why did u decide to separate it?" - i extracted this function to reuse logic. ( look at the 2-nd note below )
2) "it increases count of code rows" -  Not sure i follow this  argument. Actually it does the opposite ) or did you mean something else? but in regard to the parameter and operator- yes that way its more readable. i changed it, please take a look

ksercs: oh, Alex, i didn't note that we use Map instead of object => code copypaste is much wider that was before, sorry

ksercs: But i think actual variant is ok anyway

ksercs (approved):

marker-dao (approved):

PR #23809, opened 2023-01-31
R2 React: Fix issue with cached selectors.

williamvinogradov:
Fixed issue with recreated each time cached selectors.
[Look to the origin comment about this issue](https://github.com/DevExpress/DevExtreme/pull/23775#discussion_r1090514168)

### DIFF
--- a/packages/core/src/utils/__test__/memoize.test.ts
+++ b/packages/core/src/utils/__test__/memoize.test.ts
@@ -5,7 +5,7 @@ describe('memoize', () => {
     const spyFunc = jest.fn();
     const cachedFunc = memoize(spyFunc, () => true);
 
-    cachedFunc();
+    cachedFunc({});
 
     expect(spyFunc).toHaveBeenCalled();
   });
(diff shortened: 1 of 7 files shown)


const314 (approved):

IlyaKhD (approved):

packages/core/src/utils/__test__/memoize.test.ts
```diff
@@ -26,9 +26,9 @@ describe('memoize', () => {
     const spyComparer = jest.fn().mockReturnValue(true);
     const cachedFunc = memoize(spyFunc, spyComparer);
 
-    cachedFunc(...args);
-    cachedFunc(...args);
-    cachedFunc(...args);
+    cachedFunc(args);
```
MikeVitik, line 29: Question, Suggestion: Do we really need so many mock here?
If we remove mock it makes the code clear:
```
    const cachedFunc = memoize(spyFunc, () => true);

    cachedFunc(2, 2);
```

PR #23810, opened 2023-02-01
DataGrid - small ts improvements

pomahtri:


williamvinogradov (dismissed):

js/core/utils/callbacks.d.ts
```diff
@@ -0,0 +1,21 @@
...
+  empty(): this;
+
+  fireWith(context: TContext, args: TArgs): this;
+
+  fire(...args: TArgs): this;
+
+  fired(): boolean;
+}
+
+export default function Callbacks<TArgs extends any[]>(): Callback<TArgs>;
```
IlyaKhD: **Minor**
Let's rename it to `createCallback` (both in  .ts and in .js). This is a default export, nothing should be broken.

```suggestion
export default function createCallback<TArgs extends any[]>(): Callback<TArgs>;
```

IlyaKhD (dismissed):

williamvinogradov (approved):

IlyaKhD (approved):

PR #23813, opened 2023-02-01
ColorView: Use black color when multiple colors are passed to value(T1127428)

alexanderPolosatov:


Zedwag (dismissed):

testing/tests/DevExpress.ui.widgets.editors/colorView.tests.js
```diff
@@ -443,6 +443,14 @@ QUnit.module('ColorView', {
         this.checkColor({ r: 0, g: 100, b: 50, hex: '#006432', alpha: 1 }, assert);
     });
 
+    QUnit.test('ColorView sets black color when value passed as array(T1127428)', function(assert) {
```
ksercs: Actually, we don't support value as an array, we just wanna test a behavior when invalid value is passed. While tests are also our internal documentation, I suggest u name this test like:
```suggestion
    QUnit.test('ColorView should apply a black color when an invalid value is passed (T1127428)', function(assert) {
```

ksercs:
@alexanderPolosatov plz update post's name according to the updates PR's title

ksercs (dismissed):

PR #23820, opened 2023-02-02
 R2 React: Add CSS class & styles forwarding

williamvinogradov:
Forwarding of CSS className and inline styles added.
Because in the origin widgets, we have the ```cssClass``` option that forwards class names.

### DIFF
--- a/packages/react/src/compatible-components/radio-group/radio-group.tsx
+++ b/packages/react/src/compatible-components/radio-group/radio-group.tsx
@@ -42,6 +42,8 @@ export function RadioGroupCompatible<T>({
   itemComponent: ItemComponent,
   valueExpr,
   displayExpr,
+  className,
+  style,
   ...otherProps
 }: CompatibleRadioGroupProps<T>) {
   useCompatibleLifecycle(otherProps);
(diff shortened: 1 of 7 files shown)


packages/react/src/compatible-components/radio-group/radio-group.tsx
```diff
@@ -43,19 +60,37 @@ export function RadioGroupCompatible<T>({
...
-            value={value}
-            label={renderLabel(item, index)}
-          />
-        );
-      })}
-    </RadioGroup>
+    (otherProps.visible ?? true)
+      ? (
+        <RadioGroup<T>
+          {...otherProps}
```
MikeVitik: Suggestion: Introduce new restProps and use it here instead of otherProps, and replace all props.* to local destructed variable.
```
const { width, height, accessKey, onFosusIn, onFocusOut, ...restProps} = otherProps;
```

MikeVitik (approved):

packages/react/src/compatible-components/radio-group/radio-group.tsx
williamvinogradov: Yes, agree with you, I will fix it

const314 (approved):

PR #23822, opened 2023-02-03
R2 Styles: Add Angular styles example + modify React styles example.

williamvinogradov:
Added new Angular CSS variables examples.
Refactored a few existing React examples to make them look the same as Angular examples.


const314 (approved):

playgrounds/angular/src/index.html
```diff
@@ -7,7 +7,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link rel="icon" type="image/x-icon" href="favicon.ico">
 </head>
-<body>
+<body class="dx-material-blue-light">
```
MikeVitik: Question: Do we really need this class, because react works without it?

packages/angular/projects/angular/src/components/radio-button/radio-button.component.scss
```diff
@@ -0,0 +1,10 @@
+@import '@devextreme/styles/src/radio-button/radio-button.scss';
```
MikeVitik, line 1: Question: Do I understand right? It is Angular-specific CSS rules because Angular renders additional DOM elements and common styles don't work.

packages/angular/projects/angular/src/components/radio-button/radio-button.component.scss
```diff
@@ -0,0 +1,10 @@
...
+
+:host {
+  display: block;
+}
+
+.radio-input {
```
MikeVitik: Question, Suggestion: Maybe we should move this to the style package and use it in [react](
https://github.com/DevExpress/DevExtreme/blob/renovation_2_0_monorepo/packages/react/src/components/radio-button/radio-button-internal.tsx#L36)?

MikeVitik (changes requested):
Please read my suggestion about .radio-input.

playgrounds/angular/src/index.html
williamvinogradov: Discussed.

Maybe we should leave only the scoped version of the CSS styles.
But it needs to be discussed in the future, so I fixed this and made it consistent with current React examples.

packages/angular/projects/angular/src/components/radio-button/radio-button.component.scss
williamvinogradov: Eup, we need additional styles for additional Angular component DOM elements.

packages/angular/projects/angular/src/components/radio-button/radio-button.component.scss
williamvinogradov: Yes, agree with you.
Fixed.

MikeVitik (approved):

PR #23833, opened 2023-02-06
Viz: add support of custom styles (patterns, gradients)

Krijovnick:


js/exporter/image_creator.js
```diff
@@ -419,31 +419,44 @@ function drawElement(element, context, parentOptions, shared) {
...
+    applyGradient(context, options, shared, element, 'radial');
 
     context.restore();
 
     return promise;
 }
 
-function applyGradient(context, options, { gradients }, element) {
-    if(gradients.length === 0) {
+function applyGradient(context, options, { linearGradients, radialGradients }, element, type) {
```
EvgeniiMorozovDevSoft, line 436: did I understand correctly that the function is not pure?
since we are changing contexts ?

js/exporter/image_creator.js
```diff
@@ -644,33 +660,50 @@ function strokeElement(context, options, isText) {
...
 }
 
-function getPattern(context, pattern, shared) {
+function getPattern(context, pattern, shared, parentOptions) {
     const options = getElementOptions(pattern, shared.rootAppended);
     const patternCanvas = imageCreator._createCanvas(options.width, options.height, 0);
     const patternContext = patternCanvas.getContext('2d');
 
-    drawCanvasElements(pattern.childNodes, patternContext, options, shared);
+    const d = drawCanvasElements(pattern.childNodes, patternContext, options, shared);
```
EvgeniiMorozovDevSoft: Please can you change the name of the variable "d", When I look at line 677, I have to keep an understanding of what "d"

js/exporter/image_creator.js
```diff
@@ -419,31 +419,44 @@ function drawElement(element, context, parentOptions, shared) {
...
     }
     const id = parseUrl(options.fill);
     if(id && gradients[id]) {
         const box = element.getBBox();
-        const gradient = context.createLinearGradient(box.x, 0, box.x + box.width, 0);
+        const horizontalCenter = box.x + box.width / 2;
+        const verticalCenter = box.y + box.height / 2;
+        const gradient = type === 'linear' ?
+            context.createLinearGradient(box.x, 0, box.x + box.width, 0) :
+            context.createRadialGradient(horizontalCenter, verticalCenter, 0, horizontalCenter, verticalCenter, Math.max(box.height / 2, box.width / 2));
```
EvgeniiMorozovDevSoft: Why do we transmit the `horizontalCenter` twice ?

EvgeniiMorozovDevSoft: `Math.max(box.height / 2, box.width / 2)`, can you put this line in a separate variable ? 

At the moment I don't understand why we need the maximum radius

Krijovnick: Because coordinates of start circle and end circle the same

js/exporter/image_creator.js
```diff
@@ -419,31 +419,44 @@ function drawElement(element, context, parentOptions, shared) {
...
         return;
     }
     const id = parseUrl(options.fill);
     if(id && gradients[id]) {
         const box = element.getBBox();
-        const gradient = context.createLinearGradient(box.x, 0, box.x + box.width, 0);
+        const horizontalCenter = box.x + box.width / 2;
+        const verticalCenter = box.y + box.height / 2;
+        const gradient = type === 'linear' ?
+            context.createLinearGradient(box.x, 0, box.x + box.width, 0) :
```
EvgeniiMorozovDevSoft, line 448: The third parameter is, the final coordinate or the width of something?

(I saw that this is an old code)

js/exporter/image_creator.js
```diff
@@ -419,31 +419,44 @@ function drawElement(element, context, parentOptions, shared) {
...
 
-        gradients[id].forEach(opt => {
+        gradients[id].colors.forEach(opt => {
             const offset = parseInt(opt.offset.replace(/%/, ''));
             gradient.addColorStop(offset / 100, opt.stopColor);
         });
 
+        if(type === 'linear') {
+            context.translate(horizontalCenter, verticalCenter);
+            context.rotate(gradients[id].transform ? gradients[id].transform.replace(/\D/g, '') * Math.PI / 180 : 0);
```
EvgeniiMorozovDevSoft: `gradients[id].transform.replace(/\D/g, '') * Math.PI / 180` Unfortunately, it is very difficult for me to understand what this value is, we can also send it to a variable?

js/exporter/image_creator.js
Krijovnick: Yes, canvas mutation difficult made with pure functions.

js/exporter/image_creator.js
```diff
@@ -419,31 +419,44 @@ function drawElement(element, context, parentOptions, shared) {
...
+            context.translate(-horizontalCenter, -verticalCenter);
```
EvgeniiMorozovDevSoft, line 460: Why do we translate twice?

js/exporter/image_creator.js
Krijovnick: >At the moment I don't understand why we need the maximum radius

If we choose minimum radius, then gradient on the narrow element will be small. In the svg label gradient fill all the space, thus I choose maximum radius

js/exporter/image_creator.js
Krijovnick: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createLinearGradient

js/exporter/image_creator.js
Krijovnick: because of the canvas specifics after rotation

js/exporter/image_creator.js
EvgeniiMorozovDevSoft: Since it has a change effect on the object, maybe it's worth putting it inside the object, or is it currently impossible to do it in a simple way ?

js/exporter/image_creator.js
EvgeniiMorozovDevSoft: That's why I suggested putting it in a separate variable, so that there would be no such question.

js/viz/chart.d.ts
```diff
@@ -2015,9 +2016,10 @@ export interface dxChartCommonPaneSettings {
     /**
      * @docid dxChartOptions.commonPaneSettings.backgroundColor
      * @default 'none'
+     * @type string|object
      * @public
      */
-    backgroundColor?: string;
+    backgroundColor?: Color | string;
```
alexanderPolosatov: let's change order accepted values to _string | Color_ like in other places

js/viz/core/base_widget.d.ts
```diff
@@ -865,3 +865,23 @@ export interface BaseWidgetAnnotationConfig {
      */
     y?: number;
 }
+
+/**
+ * @docid
+ * @type object
+ * @namespace DevExpress.viz
+ */
+export interface Color {
```
alexanderPolosatov: Can we make name more concretive? I think name of type is too common. But Ilya more experienced in this question

js/viz/core/base_widget.js
```diff
@@ -720,6 +731,28 @@ const baseWidget = isServerSide ? getEmptyComponent() : DOMComponent.inherit({
         this.isReady = getFalse;
     },
 
+    _renderGraphicObjects: function() {
+        const graphics = getGraphicObjects();
+        for(const id in graphics) {
+            if(!this._graphicObjects[id]) {
+                const { element, colors, rotationAngle, template, width, height } = graphics[id];
+                const renderer = this._renderer;
```
alexanderPolosatov: we can move this outside of cycle

js/viz/core/renderers/renderer.js
```diff
@@ -1946,19 +1947,38 @@ Renderer.prototype = {
         return elem.attr({ text: text, x: x || 0, y: y || 0 });
     },
 
-    linearGradient: function(stops) {
-        const id = getNextDefsSvgId();
-        const that = this;
-        const gradient = that._createElement('linearGradient', { id: id }).append(that._defs);
+    linearGradient: function(stops, _id, rotationAngle) {
```
alexanderPolosatov: can we write like

_linearGradient: function(stops, id = getNextDefsSvgId(), rotationAngle) {_

?

js/viz/core/renderers/renderer.js
```diff
@@ -1946,19 +1947,38 @@ Renderer.prototype = {
...
+    radialGradient: function(stops, id) {
+        const gradient = this._createElement('radialGradient', { id: id }).append(this._defs);
+
+        this._createGradientStops(stops, gradient);
+
+        return gradient;
+    },
+
+    _createGradientStops: function(stops, group) {
+        const that = this;
```
alexanderPolosatov: extra _that_ variable

js/viz/core/renderers/renderer.js
```diff
@@ -2109,8 +2144,24 @@ Renderer.prototype = {
...
-    initHatching: function() {
-        const storage = this._hatchingStorage = this._hatchingStorage || { byHash: {}, baseId: getNextDefsSvgId() };
+    lightenFilter: function(id) {
+        const that = this;
+        const coef = 1.3;
+        const filter = that._createElement('filter', { id: id }).append(that._defs);
+
+        that._createElement('feColorMatrix')
+            .attr({
+                type: 'matrix', values: `${coef} 0 0 0 0 0 ${coef} 0 0 0 0 0 ${coef} 0 0 0 0 0 1 0`
```
alexanderPolosatov: may be pass attributes like second argument of _createElement?

js/viz/core/renderers/renderer.js
```diff
@@ -2154,13 +2214,17 @@ Renderer.prototype = {
...
+    },
 };
 
-function getHatchingHash(color, hatching) {
+function getHatchingHash({ color, hatching }) {
     return '@' + color + '::' + hatching.step + ':' + hatching.width + ':' + hatching.opacity + ':' + hatching.direction;
 }
 
+function getLighteningHash() {
+    return '@filter::lightening';
```
alexanderPolosatov: may be remove this method and convert returnValue to constant?

js/viz/series/scatter_series.js
```diff
@@ -157,11 +157,15 @@ const baseScatterMethods = {
 
     _createLegendState: function(styleOptions, defaultColor) {
         return {
-            fill: styleOptions.color || defaultColor,
+            fill: extractColor(styleOptions.color, true) || defaultColor,
             hatching: styleOptions.hatching ? _extend({}, styleOptions.hatching, { direction: 'right' }) : undefined
         };
     },
 
+    _getColorId: function() {
```
alexanderPolosatov: replace to noop

testing/tests/DevExpress.viz.core/graphic_objects.tests.js
```diff
@@ -0,0 +1,38 @@
+import { registerPattern, registerGradient, getGraphicObjects } from 'viz/graphic_objects';
+import utils from 'viz/core/utils';
+
+
+QUnit.module('Graphic objects', {
+    beforeEach: function() {
+        this.getNextDefsStub = sinon.stub();
```
alexanderPolosatov: let's add restore to afterEach

testing/tests/DevExpress.viz.renderers/Renderer.tests.js
```diff
@@ -964,62 +965,240 @@ QUnit.test('text with params. text argument is null', function(assert) {
...
-    this.renderer.lockHatching('blue', { direction: 'left' });
-    this.renderer.initHatching();
+    this.renderer.lockDefsElements({ color: 'red', hatching: { direction: 'left' } }, undefined, 'pattern');
+    this.renderer.lockDefsElements({ color: 'blue', hatching: { direction: 'left' } }, undefined, 'pattern');
+    this.renderer.initDefsElements();
     assert.strictEqual(renderers.SvgElement.returnValues[2].dispose.callCount, 1, 'pattern 1');
     assert.strictEqual(renderers.SvgElement.returnValues[3].dispose.callCount, 1, 'pattern 2');
 });
 
 QUnit.test('release after init', function(assert) {
```
alexanderPolosatov: let's make name of tets more informativity

alexanderPolosatov (changes requested):

js/viz/core/renderers/renderer.js
Krijovnick: yes, we can

js/viz/core/base_widget.d.ts
Krijovnick: @IlyaKhD Can you help us and answer if this 'Color' name is too common? Can we name it like `VizColor` then for example?

js/exporter/image_creator.js
```diff
@@ -419,31 +419,46 @@ function drawElement(element, context, parentOptions, shared) {
...
+            context.createLinearGradient(box.x, 0, box.x + box.width, 0) :
+            context.createRadialGradient(horizontalCenter, verticalCenter, 0, horizontalCenter, verticalCenter, maxRadius);
+
+        gradients[id].colors.forEach(opt => {
             const offset = parseInt(opt.offset.replace(/%/, ''));
             gradient.addColorStop(offset / 100, opt.stopColor);
         });
 
+        if(type === 'linear') {
+            const angle = gradients[id].transform ? gradients[id].transform.replace(/\D/g, '') * Math.PI / 180 : 0;
```
ksercs:
```suggestion
            const angle = gradients[id].transform?.replace(/\D/g, '') * Math.PI / 180 ?? 0;
```

js/exporter/image_creator.js
```diff
@@ -419,31 +419,46 @@ function drawElement(element, context, parentOptions, shared) {
...
     const id = parseUrl(options.fill);
     if(id && gradients[id]) {
         const box = element.getBBox();
-        const gradient = context.createLinearGradient(box.x, 0, box.x + box.width, 0);
-
-        gradients[id].forEach(opt => {
+        const horizontalCenter = box.x + box.width / 2;
+        const verticalCenter = box.y + box.height / 2;
+        const maxRadius = Math.max(box.height / 2, box.width / 2);
+        const gradient = type === 'linear' ?
```
ksercs, line 447: Syntax like
```javascript
    const variable = condition
        ? a
        : b;
```
is more intuitive for me because it separates condition/a/b more explicitly. 
Here is a discussion in [airbnb](https://github.com/airbnb/javascript/issues/1281).
I believe we will configure our linter to fix it soon.

js/exporter/image_creator.js
```diff
@@ -644,33 +662,50 @@ function strokeElement(context, options, isText) {
...
-    drawCanvasElements(pattern.childNodes, patternContext, options, shared);
+    const promise = drawCanvasElements(pattern.childNodes, patternContext, options, shared);
+    const onDone = () => {
+        context.fillStyle = context.createPattern(patternCanvas, 'repeat');
+        context.globalAlpha = parentOptions.fillOpacity;
+        context.fill();
+        context.globalAlpha = 1;
+    };
+
+    if(isPromise(promise)) {
```
ksercs, line 678: y do we need such conditions (this is not a single place where we use it)?
I see that `drawCanvasElements` always returns `asyncEach` which always return deferred.
Am i missing smth?

It's also strange to see `isPromise(promise)`. If `promise` can be not a promise, maybe it's better to choose another name for it?

js/viz/chart_components/tracker.js
```diff
@@ -117,7 +117,7 @@ const baseTrackerPrototype = {
 
         if(noHoveredSeries) {
             that._clean();
-            that._renderer.initHatching();
+            that._renderer.initDefsElements();
```
ksercs, line 120: I don't understand what is "defs elements". Is it a special term i didn't meet? Or an abbreviation?

js/viz/core/base_widget.d.ts
ksercs: think @vladaskorohodova  can also consult us

alexanderPolosatov (dismissed):
lgtm

js/viz/core/renderers/renderer.js
```diff
@@ -1946,19 +1949,36 @@ Renderer.prototype = {
...
+        this._createGradientStops(stops, gradient);
 
         return gradient;
     },
 
+    _createGradientStops: function(stops, group) {
+        stops.forEach((stop) => {
+            this._createElement('stop', {
+                offset: stop.offset,
+                'stop-color': stop['stop-color'] || stop.color,
```
ksercs:
```suggestion
                'stop-color': stop['stop-color'] ?? stop.color,
```
Cannot `stop['stop-color']` be a `0`?

EvgeniiMorozovDevSoft (dismissed):

js/viz/core/renderers/renderer.js
```diff
@@ -1987,6 +2007,31 @@ Renderer.prototype = {
         return pattern;
     },
 
+    customPattern: function(id, templateFunction, width, height) {
```
ksercs: It's not necessary to specify variable's type in its name. `template` is meaningful enough
```suggestion
    customPattern: function(id, template, width, height) {
```

js/viz/core/renderers/renderer.js
```diff
@@ -1987,6 +2007,31 @@ Renderer.prototype = {
...
+        const opt = {
```
ksercs: What is `opt`? If it's an abbreviation, let's use a full name

js/viz/core/renderers/renderer.js
```diff
@@ -1987,6 +2007,31 @@ Renderer.prototype = {
...
+            id: id,
```
ksercs:
```suggestion
            id,
```

js/viz/core/renderers/renderer.js
```diff
@@ -1987,6 +2007,31 @@ Renderer.prototype = {
...
+        templateFunction.render({ container: pattern.element });
+
+        return pattern;
+    },
+
+    _getPatternUnits: function(width, height) {
+        const widthInNumber = Number(width);
+        const heightInNumber = Number(height);
+
+        if(widthInNumber && heightInNumber) {
```
ksercs: 1. Do we use casts above because `width/height` can be a string? Is strange for me to see `widthInNumber` name. We can just write 
2. `undefined` is returned by default, it's not necessary to return it expecially
```javascript
_getPatternUnits: function(width, height) {
    if (+width && +height) {
        return 'userSpaceOnUse';
    }
},
```

js/viz/core/renderers/renderer.js
```diff
@@ -2109,8 +2154,22 @@ Renderer.prototype = {
         return filter;
     },
 
-    initHatching: function() {
-        const storage = this._hatchingStorage = this._hatchingStorage || { byHash: {}, baseId: getNextDefsSvgId() };
+    lightenFilter: function(id) {
+        const that = this;
```
ksercs: I don't see y we need `that` here

js/viz/core/renderers/renderer.js
```diff
@@ -2109,8 +2154,22 @@ Renderer.prototype = {
...
+        const coef = 1.3;
+        const filter = that._createElement('filter', { id: id }).append(that._defs);
```
ksercs:
```suggestion
        const filter = that._createElement('filter', { id }).append(that._defs);
```

js/viz/core/renderers/renderer.js
```diff
@@ -2109,8 +2154,22 @@ Renderer.prototype = {
...
+
+        that._createElement('feColorMatrix', {
+            type: 'matrix', values: `${coef} 0 0 0 0 0 ${coef} 0 0 0 0 0 ${coef} 0 0 0 0 0 1 0`
```
ksercs:
```suggestion
            type: 'matrix',
            values: `${coef} 0 0 0 0 0 ${coef} 0 0 0 0 0 ${coef} 0 0 0 0 0 1 0`
```

js/viz/graphic_objects.js
```diff
@@ -0,0 +1,19 @@
+import { getNextDefsSvgId } from './core/utils';
+
+const graphicObjects = {};
+
+export const registerPattern = (options) => {
+    const id = getNextDefsSvgId();
+    graphicObjects[id] = { element: 'pattern', ...options };
```
ksercs: It's strange for me to see `element: 'pattern' | 'gradient'`, can we rename it to `type`, for example? 
(same for `registerGradient`)
```suggestion
    graphicObjects[id] = { type: 'pattern', ...options };
```

js/viz/series/area_series.js
```diff
@@ -84,9 +89,10 @@ const baseAreaMethods = {
             border: borderStyle,
             elements: {
                 stroke: 'none',
-                fill: options.color || defaultColor,
+                fill: extractColor(options.color) || defaultColor,
                 hatching: options.hatching,
-                opacity: options.opacity
+                opacity: options.opacity,
+                filter: options.lightening || null
```
ksercs: I'll not leave similar notes more, but i recommend checking all places when u use `||` for fallback - it's just not safe
```suggestion
                filter: options.lightening ?? null
```

testing/tests/DevExpress.viz.charts/chart.part7.tests.js
```diff
@@ -2086,4 +2087,82 @@ $('<div id="chartContainer">').appendTo('#qunit-fixture');
...
+        seriesMockData.series.push(stubSeries);
+
+        // act
+        const chart = this.createChart({
+            series: {}
+        });
+
+        this.$container.remove();
+
+        assert.equal(chart._graphicObjects, null);
```
ksercs: it's better to use `strictEqual` everywhere, especially in such asserts. Don't u use `==` instead of `===`, right?

testing/tests/DevExpress.viz.charts/chart.part7.tests.js
```diff
@@ -2086,4 +2087,82 @@ $('<div id="chartContainer">').appendTo('#qunit-fixture');
...
+            commons.environment.afterEach.call(this);
+            this.fakeGraphicObjects.restore();
+        },
+    }));
+
+    QUnit.test('Should create graphic objects', function(assert) {
+        const stubSeries = new MockSeries({});
+        seriesMockData.series.push(stubSeries);
+
+        // act
```
ksercs: Why do u leave such comments? I think u separate "arrange-act-assert" with empty rows explicitly enough

testing/tests/DevExpress.viz.charts/chart.part7.tests.js
```diff
@@ -2086,4 +2087,82 @@ $('<div id="chartContainer">').appendTo('#qunit-fixture');
...
+        chart.render({
+            force: true
+        });
+
+        assert.equal(chart._renderer.linearGradient.callCount, 1);
+        assert.equal(chart._renderer.radialGradient.callCount, 1);
+        assert.equal(chart._renderer.customPattern.callCount, 1);
+    });
+
+    QUnit.test('Should dispose graphic objects', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('Should dispose graphic objects on container remove', function(assert) {
```

testing/tests/DevExpress.viz.charts/chart.part7.tests.js
```diff
@@ -2086,4 +2087,82 @@ $('<div id="chartContainer">').appendTo('#qunit-fixture');
...
+            });
+        },
+        afterEach: function() {
+            this.clock.restore();
+            commons.environment.afterEach.call(this);
+            this.fakeGraphicObjects.restore();
+        },
+    }));
+
+    QUnit.test('Should create graphic objects', function(assert) {
```
ksercs: let's name tests more concrete. When should it create graphic objects?
```suggestion
    QUnit.test('Should create graphic objects on charts creating', function(assert) {
```

testing/tests/DevExpress.viz.charts/pieChart.tests.js
```diff
@@ -2431,4 +2430,59 @@ const overlappingEnvironment = $.extend({}, environment, {
         this.checkLabelPosition(assert, points2[0].getLabels()[0], [155, 75]);
         this.checkLabelPosition(assert, points2[1].getLabels()[0], [150, 85]);
     });
+
+    QUnit.module('Render graphic objects', $.extend({}, environment, {
```
ksercs:
```suggestion
    QUnit.module('Graphic objects render', $.extend({}, environment, {
```

testing/tests/DevExpress.viz.charts/pieChart.tests.js
```diff
@@ -2431,4 +2430,59 @@ const overlappingEnvironment = $.extend({}, environment, {
...
+                };
+            });
+        },
+        afterEach: function() {
+            environment.afterEach.call(this);
+            this.fakeGraphicObjects.restore();
+        },
+    }));
+
+    QUnit.test('Should create graphic objects', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('Should create graphic objects on pieChart creating', function(assert) {
```

testing/tests/DevExpress.viz.core.series/areaSeries.tests.js
```diff
@@ -1068,17 +1163,56 @@ function setDiscreteType(series) {
...
+                color: { customId: 'id_color_1' }
+            },
+            selectionStyle: {
+                hatching: { direction: 'left' },
+                color: { customId: 'id_color_2' }
+            }
+        });
+
+        assert.deepEqual(series.getLegendStyles(), {
+            'hover': {
```
ksercs: brackets are excess everywhere

testing/tests/DevExpress.viz.core.series/areaSeries.tests.js
```diff
@@ -1068,17 +1163,56 @@ function setDiscreteType(series) {
...
                 'fill': 's-color',
                 'hatching': 's-hatching',
-                opacity: undefined
+                opacity: undefined,
+                'filter': true,
+            }
+        });
+    });
+
+    QUnit.test('custom styles are defined', function(assert) {
```
ksercs: really common. Maybe like this? Please, review all tests. 
Test name should describe the full case and answer the question "how should it work"?
Maybe like this?
```suggestion
    QUnit.test('series custom hover and selection styles should be passed to legend styles', function(assert) {
```

testing/tests/DevExpress.viz.core.series/financialSeries.tests.js
```diff
@@ -821,6 +823,99 @@ const checkGroups = function(assert, series) {
         });
     });
 
+    QUnit.test('Custom styles should not apply', function(assert) {
```
ksercs, line 826: Never?
```suggestion
    QUnit.test('Custom styles should not apply', function(assert) {
```

js/exporter/image_creator.js
Krijovnick: We say a lot of words about better code readability. But does that make it easier to read? It's hard for me to read and understand this kind of code.

js/exporter/image_creator.js
Krijovnick: drawCanvasElements can return promise or `undefined`. I think about another name but I can't come up with.

js/viz/chart_components/tracker.js
Krijovnick: [defs element](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/defs) - this is for store graphical elements such as gradients and patterns

js/exporter/image_creator.js
Krijovnick: Ok, I think we should to discuss that, because for me more intuitive is my variant. Or make lint to fix this.

js/viz/core/renderers/renderer.js
Krijovnick: +width and +height is it informative? Looks like a hack

testing/tests/DevExpress.viz.charts/chart.part7.tests.js
Krijovnick: Sometimes it's more informative. But in this case it is not necessary.

testing/tests/DevExpress.viz.core.series/areaSeries.tests.js
Krijovnick: The module has already name that all tests about legend styles. And the tests in this module check not only hover and selection. It check all styles: normal, hover and selection. Should it has more detailed describing?

testing/tests/DevExpress.viz.core.series/financialSeries.tests.js
Krijovnick: For financial series should not apply

js/exporter/image_creator.js
ksercs: I'm pretty sure modern syntax does not worse the readability. I believe it's a kind of habit - we just need to get used to new approaches to slow down the aging of the codebase. 
But I actually agree that this code is still hard to understand. Maybe smth like this?
```javascript
    const transform = gradients[id].transform?.replace(/\D/g, '') ?? 0;
    const angle =  transform * Math.PI / 180;
```

js/exporter/image_creator.js
ksercs: Nevermind, I agree that it should be fixed by eslint. I only thought it would be better to start sync now.
I especially referred to airbnb to not rely only on my feelings - it's a standard we're trying to use (not very successfully). Here is one more [mention](https://github.com/airbnb/javascript#comparison--nested-ternaries) right in airbnb guide, code sample on mdn [conditionals](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/conditionals) page, code convention from [Douglas Crockford](https://www.crockford.com/), sample in [JavaScript Standart Style](https://standardjs.com/rules.html).

js/exporter/image_creator.js
ksercs: okay, i believe this function should return promise always, but it's not a change from your code, so let's do not touch it now

js/viz/chart_components/tracker.js
ksercs: thanks!

js/viz/core/renderers/renderer.js
ksercs: Yeap, u can write `Number(width) && Number(height)` if u want - my point was about strange variables

testing/tests/DevExpress.viz.core.series/areaSeries.tests.js
ksercs: About legend series - got it. I'm just not used to old modules when tests are not inside of a module physically, so didn't catch it during review.

About detailed describing - yes. Now when u're trying to imagine a name for a test which tests hover, selection and normal styles, u can see another problem - this test is not atomic, it checks a lot of things at one time.

testing/tests/DevExpress.viz.core.series/financialSeries.tests.js
ksercs: Oh, it's for financial, didn't notice

js/exporter/image_creator.js
```diff
@@ -419,31 +419,46 @@ function drawElement(element, context, parentOptions, shared) {
...
+        const horizontalCenter = box.x + box.width / 2;
+        const verticalCenter = box.y + box.height / 2;
+        const maxRadius = Math.max(box.height / 2, box.width / 2);
+        const gradient = type === 'linear' ?
+            context.createLinearGradient(box.x, 0, box.x + box.width, 0) :
+            context.createRadialGradient(horizontalCenter, verticalCenter, 0, horizontalCenter, verticalCenter, maxRadius);
+
+        gradients[id].colors.forEach(opt => {
             const offset = parseInt(opt.offset.replace(/%/, ''));
             gradient.addColorStop(offset / 100, opt.stopColor);
```
EugeniyKiyashko, line 453: function toDecimal(percent) {
  return parseFloat(percent) / 100;
}

offset = toDecimal(opt.offset)

Krijovnick: Could you explain what this comment mean?

alexanderPolosatov (dismissed):

js/viz/core/base_widget.d.ts
IlyaKhD: Are there other "colors" between out types?
Maybe `ChartsColor` is better?

Krijovnick: >Are there other "colors" between out types?

Currently there are no "colors" in the types. But they can be in the future.

>Maybe ChartsColor is better?

I think this name ok, @vladaskorohodova  what do you think?

js/exporter/image_creator.js
ksercs: I guess it's a suggestion to make offset calculation simpler to read.
But i dont think it's required while it's a single place where u calc this. Moreover, it's not your change

js/viz/core/renderers/renderer.js
```diff
@@ -1946,19 +1949,36 @@ Renderer.prototype = {
         return elem.attr({ text: text, x: x || 0, y: y || 0 });
     },
 
-    linearGradient: function(stops) {
-        const id = getNextDefsSvgId();
-        const that = this;
-        const gradient = that._createElement('linearGradient', { id: id }).append(that._defs);
+    linearGradient: function(stops, id = getNextDefsSvgId(), rotationAngle) {
+        const gradient = this._createElement('linearGradient', {
+            id: id,
```
ksercs:
```suggestion
            id,
```

js/viz/core/renderers/renderer.js
```diff
@@ -1946,19 +1949,36 @@ Renderer.prototype = {
...
-        stops.forEach((stop) => {
-            that._createElement('stop', { offset: stop.offset, 'stop-color': stop['stop-color'] }).append(gradient);
-        });
+        this._createGradientStops(stops, gradient);
 
         return gradient;
     },
 
+    radialGradient: function(stops, id) {
+        const gradient = this._createElement('radialGradient', { id: id }).append(this._defs);
```
ksercs:
```suggestion
        const gradient = this._createElement('radialGradient', { id }).append(this._defs);
```

js/viz/core/renderers/renderer.js
```diff
@@ -1987,6 +2007,25 @@ Renderer.prototype = {
...
+        };
+        const pattern = this._createElement('pattern', option).append(this._defs);
+
+        template.render({ container: pattern.element });
+
+        return pattern;
+    },
+
+    _getPatternUnits: function(width, height) {
+        return Number(width) && Number(height) ? 'userSpaceOnUse' : undefined;
```
ksercs: It's strange to see explicit return of `undefined`.
I don't think ternary operator is the most suitable for this case until u want to choose from 2 options.
Maybe just a usual if?
```javascript
    if(Number(width) && Number(height)) {
        return 'userSpaceOnUse';
    }
```

js/viz/core/utils.js
```diff
@@ -632,3 +632,18 @@ export function valueOf(value) {
 export function pointInCanvas(canvas, x, y) {
     return x >= canvas.left && x <= canvas.right && y >= canvas.top && y <= canvas.bottom;
 }
+
+export const getNextDefsSvgId = (function() {
+    let numDefsSvgElements = 1;
+    return function() { return 'DevExpress_' + numDefsSvgElements++; };
+})();
```
ksercs: Do we really need IIFE here?

ES6 modules implement Singletone by default => we can just specify a variable in a module and a function to get it.
For example, this approach is already used in overlays [z_index](https://github.com/DevExpress/DevExtreme/blob/23_1/js/ui/overlay/z_index.js).

```javascript
let numDefsSvgElements = 1;
export const getNextDefsSvgId = () => {
    return `DevExpress_${numDefsSvgElements++}`;
};
```

Krijovnick: I don't write this IIFE, I only move it from renderer. And I agree that strange code and long time ago (I think 6 years ago maybe or more) I wrote this function like you write it, but then it was writing like you see now. Why? I don't know. I did not touch it because there was a sense in such writing. Maybe @alexanderPolosatov can remember why it was write like this.

js/viz/core/base_widget.d.ts
vladaskorohodova: I also think it's okay

testing/tests/DevExpress.viz.core.series/lineSeries.tests.js
```diff
@@ -1284,6 +1290,44 @@ function setDiscreteType(series) {
         });
     });
 
+    QUnit.test('Custom styles should not apply, normal state', function(assert) {
+        const series = this.createSeries({ ...this.options, ...{ color: { fillId: 'id_color', base: 'n-color' } } });
```
ksercs, line 1294: It's not required to create a new object and distructurize it, we can just specify new props => code is simpler and faster (desctucturization = O(n))
```suggestion
        const series = this.createSeries({ ...this.options, color: { fillId: 'id_color', base: 'n-color' } });
```

ksercs (approved):

ksercs (dismissed):

alexanderPolosatov (dismissed):

EugeniyKiyashko (dismissed):

js/viz/chart.d.ts
```diff
@@ -2015,9 +2016,10 @@ export interface dxChartCommonPaneSettings {
     /**
      * @docid dxChartOptions.commonPaneSettings.backgroundColor
      * @default 'none'
+     * @type string|object
```
IlyaKhD: I think there is no need to add this tag anywhere, let's try removing it.

js/viz/core/base_widget.d.ts
```diff
@@ -865,3 +865,23 @@ export interface BaseWidgetAnnotationConfig {
      */
     y?: number;
 }
+
+/**
+ * @docid
+ * @type object
+ * @namespace DevExpress.viz
+ */
+export interface ChartsColor {
```
IlyaKhD: If this type is shared among  several charts types:
1) move it to common/charts.d.ts
2) make it public (add  `@public` tag
3) make it `type` not `interface` - all our public types are `type`'s

js/viz/graphic_objects.d.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @docid viz.registerPattern
+ * @publicName registerPattern(options)
+ * @static
+ * @public
+ */
+export function registerPattern(options: { width: number | string; height: number | string; template: template }): string;
+/**
+ * @namespace DevExpress.viz
+ */
+export interface GradientColor {
```
IlyaKhD: If this type is public:
1) add the `@public` tag
2) make it `type`

js/viz/graphic_objects.d.ts
```diff
@@ -0,0 +1,35 @@
+import {
```
IlyaKhD: If these new functions are used with all charts, they should be moved to the `common/charts` module

IlyaKhD (changes requested):

ksercs (dismissed):

js/common/charts.d.ts
```diff
@@ -277,3 +281,57 @@ export type WordWrap = 'normal' | 'breakWord' | 'none';
...
+export function registerGradient(type: string, options: { rotationAngle?: number; colors: Array<GradientColor> }): string;
+/**
+ * @docid viz.registerPattern
+ * @publicName registerPattern(options)
+ * @static
+ * @public
+ */
+export function registerPattern(options: { width: number | string; height: number | string; template: template }): string;
+/**
+ * @namespace DevExpress.viz
```
IlyaKhD: All things here should have the same namespace
```suggestion
 * @namespace DevExpress.common.charts
```

js/common/charts.d.ts
```diff
@@ -277,3 +281,57 @@ export type WordWrap = 'normal' | 'breakWord' | 'none';
...
+     * @public
+     */
+    fillId?: string;
+};
+
+/**
+ * @docid viz.registerGradient
+ * @publicName registerGradient(type, options)
+ * @static
+ * @public
```
IlyaKhD, line 311:
```suggestion
 * @public
 * @namespace DevExpress.common.charts
```

js/common/charts.d.ts
```diff
@@ -277,3 +281,57 @@ export type WordWrap = 'normal' | 'breakWord' | 'none';
...
+ */
+export function registerGradient(type: string, options: { rotationAngle?: number; colors: Array<GradientColor> }): string;
+/**
+ * @docid viz.registerPattern
+ * @publicName registerPattern(options)
+ * @static
+ * @public
```
IlyaKhD, line 319:
```suggestion
 * @public
 * @namespace DevExpress.common.charts
```

js/common/charts.d.ts
```diff
@@ -277,3 +281,57 @@ export type WordWrap = 'normal' | 'breakWord' | 'none';
...
+    fillId?: string;
+};
+
+/**
+ * @docid viz.registerGradient
+ * @publicName registerGradient(type, options)
+ * @static
+ * @public
+ */
+export function registerGradient(type: string, options: { rotationAngle?: number; colors: Array<GradientColor> }): string;
```
IlyaKhD, line 313: Since `function` is a value, runtime that is in that `graphic_objects.js` must also be moved here

IlyaKhD (changes requested):

EugeniyKiyashko (dismissed):

IlyaKhD (approved):

vladaskorohodova (approved):

alexanderPolosatov (approved):

iBat (approved):

PR #23835, opened 2023-02-06
Calendar: add selectionMode and values options

Zedwag:


js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -274,15 +274,16 @@ const BaseView = Widget.inherit({
             return;
         }
 
-        const value = this.option('value');
-        const selectedCell = this._getCellByDate(value);
-
-        if(this._selectedCell) {
-            this._selectedCell.removeClass(CALENDAR_SELECTED_DATE_CLASS);
+        if(this._selectedCells) {
```
EugeniyKiyashko: What about tests? It doesn't clear which scenarios you support without them.

EugeniyKiyashko: I think the right way is write tests first according specification and requirement functionality. After that reanimate them step by step.

Zedwag: Added tests, while implementation I still wasn't sure what exactly tests I'll need to add

ksercs: 1. we can just write instead of this condition
2. please, make arrow function's formatting more readable
`this._selectedCells?.forEach(cell => { cell.removeClass(CALENDAR_SELECTED_DATE_CLASS); });`

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -274,15 +274,16 @@ const BaseView = Widget.inherit({
...
+            this._selectedCells.forEach((cell) => (cell.removeClass(CALENDAR_SELECTED_DATE_CLASS)));
         }
+        const multiselect = this.option('multiselect');
```
ksercs: We can write it this way:
```javascript
        const { multiselect, value, values } = this.option();
        const selectedCells = multiselect
            ? values.map(value =>  this._getCellByDate(value))
            : [this._getCellByDate(value)];
```

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -274,15 +274,16 @@ const BaseView = Widget.inherit({
...
+        const selectedCells = multiselect ?
+            this.option('values').map((value) => this._getCellByDate(value)) :
+            [this._getCellByDate(this.option('value'))];
 
-        selectedCell.addClass(CALENDAR_SELECTED_DATE_CLASS);
-        this._selectedCell = selectedCell;
+        this._selectedCells = selectedCells;
+        selectedCells.forEach((cell) => (cell.addClass(CALENDAR_SELECTED_DATE_CLASS)));
```
ksercs:
```suggestion
        selectedCells.forEach(cell => { cell.addClass(CALENDAR_SELECTED_DATE_CLASS); });
```

js/ui/calendar/ui.calendar.js
```diff
@@ -243,28 +248,64 @@ const Calendar = Editor.inherit({
...
     _dateValue: function(value, event) {
+        const optionName = Array.isArray(value) ? 'values' : 'value';
         if(event) {
             if(event.type === 'keydown') {
                 const cellElement = this._view._getContouredCell().get(0);
                 event.target = cellElement;
             }
-            this._saveValueChangeEvent(event);
+            if(optionName === 'value') {
+                this._saveValueChangeEvent(event);
```
ksercs: Do we really need a separate event?

js/ui/calendar/ui.calendar.js
```diff
@@ -243,28 +248,64 @@ const Calendar = Editor.inherit({
...
+            }
         }
-        this._dateOption('value', value);
+        this._dateOption(optionName, value);
     },
 
     _dateOption: function(optionName, optionValue) {
         if(arguments.length === 1) {
-            return this._convertToDate(this.option(optionName), optionName);
+            return optionName === 'values' ?
```
ksercs:
```javascript
            return optionName === 'values'
                ? optionValue.map(value => this._convertToDate(value))
                : this._convertToDate(optionValue);
```

js/ui/calendar/ui.calendar.js
```diff
@@ -243,28 +248,64 @@ const Calendar = Editor.inherit({
...
+                this.option('values').map((value) => this._convertToDate(value)) :
+                this._convertToDate(this.option(optionName));
         }
 
         const serializationFormat = this._getSerializationFormat(optionName);
-        this.option(optionName, dateSerialization.serializeDate(optionValue, serializationFormat));
+        const serializedValue = optionName === 'values' ?
```
ksercs:
```javascript
        const serializedValue = optionName === 'values'
            ? optionValue.map(value => dateSerialization.serializeDate(value, serializationFormat))
            : dateSerialization.serializeDate(optionValue, serializationFormat);
```

js/ui/calendar/ui.calendar.js
```diff
@@ -417,10 +458,22 @@ const Calendar = Editor.inherit({
     },
 
     _initCurrentDate: function() {
-        const currentDate = this._getNormalizedDate(this._dateOption('value')) || this._getNormalizedDate(this.option('currentDate'));
+        const multiselect = this.option('multiselect');
+        const currentDate = (multiselect ?
```
ksercs:
```javascript
        const currentDate = multiselect 
            ? this._getLowestDateInArray(this._dateOption('values'))
            : this._dateOption('value');
        const normalizedCurrentDate = this._getNormalizedDate(currentDate) ?? this.option('currentDate');

        this.option('currentDate', normalizedCurrentDate);
```

js/ui/calendar/ui.calendar.js
```diff
@@ -417,10 +458,22 @@ const Calendar = Editor.inherit({
...
+            this._getNormalizedDate(this._getLowestDateInArray(this._dateOption('values'))) :
+            this._getNormalizedDate(this._dateOption('value'))
+        ) || this._getNormalizedDate(this.option('currentDate'));
+
         this.option('currentDate', currentDate);
     },
 
+    _getLowestDateInArray: function(dates) {
+        if(!dates.length) {
+            return;
```
ksercs: does `this._getNormalizedDate` work in expected way with `undefined` paremeter?

js/ui/calendar/ui.calendar.js
```diff
@@ -417,10 +458,22 @@ const Calendar = Editor.inherit({
...
+        const multiselect = this.option('multiselect');
+        const currentDate = (multiselect ?
+            this._getNormalizedDate(this._getLowestDateInArray(this._dateOption('values'))) :
+            this._getNormalizedDate(this._dateOption('value'))
+        ) || this._getNormalizedDate(this.option('currentDate'));
+
         this.option('currentDate', currentDate);
     },
 
+    _getLowestDateInArray: function(dates) {
```
ksercs:
```javascript
    _getLowestDateInArray: function(dates) {
        if(dates.length) {
            return new Date(Math.min(...dates));
        }
    },
```

js/ui/calendar/ui.calendar.js
```diff
@@ -709,11 +764,30 @@ const Calendar = Editor.inherit({
...
-            this._dateValue(newValue, e.event);
+            this._selectValue(newValue, e.event);
             this._cellClickAction(e);
         }
     },
 
+    _selectValue: function(selectedValue, e) {
+        if(this.option('multiselect')) {
+            const values = [...this._dateOption('values')];
+            const alreadySelectedIndex = values.findIndex(date => date.toDateString() === selectedValue.toDateString());
```
ksercs, line 775: This findIndex and `splice` below will cause performance issues

js/ui/calendar/ui.calendar.js
```diff
@@ -709,11 +764,30 @@ const Calendar = Editor.inherit({
...
+
+            if(alreadySelectedIndex > -1) {
+                values.splice(alreadySelectedIndex, 1);
+            } else {
+                values.push(selectedValue);
+            }
+
+            this._currentDateChanged = true;
+            this.option('currentDate', isDefined(selectedValue) ? selectedValue : new Date());
```
ksercs: We can refactor this way:
```suggestion
            this.option('currentDate', selectedValue ?? new Date();
```
But y do we need set `new Date()`?

js/ui/calendar/ui.calendar.js
```diff
@@ -1139,24 +1213,39 @@ const Calendar = Editor.inherit({
     },
 
     _updateViewsValue: function(value) {
-        const newValue = value ? new Date(value) : null;
+        const multiselect = this.option('multiselect');
+
+        if(multiselect) {
+            this._updateViewsOption('values', value);
+        } else {
+            const newValue = value[0] ? new Date(value[0]) : null;
```
ksercs: 1. I didn't catch - y do we get `value[0]` if multiselect is disabled? intuitively it should be the other way around
2. u can see that both `this._updateViewsOption` calls are very similar - we can get rid of this `if-else`

js/ui/calendar/ui.calendar.js
```diff
@@ -1139,24 +1213,39 @@ const Calendar = Editor.inherit({
...
+        }
 
-        this._view.option('value', newValue);
-        this._beforeView && this._beforeView.option('value', newValue);
-        this._afterView && this._afterView.option('value', newValue);
+    },
+
+    _updateViewsOption: function(optionName, newValue) {
+        this._view.option(optionName, newValue);
+        this._beforeView && this._beforeView.option(optionName, newValue);
```
ksercs:
```suggestion
        this._beforeView?.option(optionName, newValue);
```

js/ui/calendar/ui.calendar.js
```diff
@@ -1139,24 +1213,39 @@ const Calendar = Editor.inherit({
...
+        this._afterView && this._afterView.option(optionName, newValue);
```
ksercs, line 1230:
```suggestion
        this._afterView?.option(optionName, newValue);
```

js/ui/calendar/ui.calendar.js
```diff
@@ -1139,24 +1213,39 @@ const Calendar = Editor.inherit({
...
     },
 
     _updateAriaSelected: function(value, previousValue) {
-        value = value ?? this._dateOption('value');
+        const multiselect = this.option('multiselect');
 
-        const $prevSelectedCell = this._view._getCellByDate(previousValue);
-        const $selectedCell = this._view._getCellByDate(value);
+        if(!multiselect) {
+            value = value ? value : [this._dateOption('value')];
```
ksercs: Please, simplify all constructions
```suggestion
            value ??= [this._dateOption('value')];
```

js/ui/calendar/ui.calendar.js
```diff
@@ -1178,9 +1267,18 @@ const Calendar = Editor.inherit({
         delete this._suppressNavigation;
     },
 
+    _processValueChanged: function(value, previousValue) {
+        const multiselect = this.option('multiselect');
+        value = multiselect ? value.map((item) => this._convertToDate(item)) : [this._convertToDate(value)];
```
ksercs: There are a lot of `if-else` and ternary operators because of `multiselect`. I think we should think about using of [Strategy design pattern](https://refactoring.guru/design-patterns/strategy)

testing/tests/DevExpress.ui.widgets.editors/calendar.tests.js
```diff
@@ -4016,75 +4213,83 @@ QUnit.module('dxCalendar number and string value support', {
...
 });
 
-QUnit.module('valueChanged handler should receive correct event', {
-    beforeEach: function() {
-        fx.off = true;
-        this.clock = sinon.useFakeTimers();
-        this.valueChangedHandler = sinon.stub();
-        this.$element = $('<div>')
-            .dxCalendar({
+['onValuesChanged', 'onValueChanged'].forEach((name) => {
```
ksercs, line 4216: this test is too difficult to read for me now. Maybe it's better to keep it separate?

js/ui/calendar/ui.calendar.js
Zedwag: now onValueChanged is used for both `value` and `values`

js/ui/calendar/ui.calendar.js
Zedwag: it was before my changes. Why? Because some cell should be focused even if there is no selected value, today's date seems good to focused

js/ui/calendar/ui.calendar.js
Zedwag: I used that to avoid more if-elses, so I pass array in both cases. With single select it will be [value], with multi - [...values]

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -253,36 +277,44 @@ const BaseView = Widget.inherit({
...
 
-            this.option('value', newValue);
-        } else {
-            this.option('value', null);
+        let value = this.option('value');
+        if(!Array.isArray(value)) {
+            value = [value];
         }
+
+        this._selectedCells?.forEach((cell) => (cell.removeClass(CALENDAR_SELECTED_DATE_CLASS)));
```
ksercs: [Airbnb guide - arrow functions](https://github.com/airbnb/javascript#arrow-functions)
Let's fix it everywhere
```suggestion
        this._selectedCells?.forEach((cell) => { cell.removeClass(CALENDAR_SELECTED_DATE_CLASS); });
```

js/ui/calendar/ui.calendar.js
```diff
@@ -55,6 +60,12 @@ const ZOOM_LEVEL = {
     CENTURY: 'century'
 };
 
+const SELECTION_STRATEGY_CLASSES = {
```
ksercs:
```suggestion
const SELECTION_STRATEGIES = {
```

js/ui/calendar/ui.calendar.js
```diff
@@ -243,28 +258,35 @@ const Calendar = Editor.inherit({
...
         }
-        this._dateOption('value', value);
+        this._dateOption(optionName, value);
     },
 
     _dateOption: function(optionName, optionValue) {
         if(arguments.length === 1) {
-            return this._convertToDate(this.option(optionName), optionName);
+            return optionName === 'values' ?
+                (this.option('values') || []).map((value) => this._convertToDate(value)) :
```
ksercs: Let's create a separate variable to simplify this expression and the expression below.
```javascript
                const values = this.option('values') ?? [];
```

js/ui/calendar/ui.calendar.js
```diff
@@ -243,28 +258,35 @@ const Calendar = Editor.inherit({
...
+                this._convertToDate(this.option(optionName));
```
ksercs:
```suggestion
                this._convertToDate(this.option('value'));
```

js/ui/calendar/ui.calendar.js
```diff
@@ -243,28 +258,35 @@ const Calendar = Editor.inherit({
...
             this._saveValueChangeEvent(event);
         }
-        this._dateOption('value', value);
+        this._dateOption(optionName, value);
     },
 
     _dateOption: function(optionName, optionValue) {
         if(arguments.length === 1) {
-            return this._convertToDate(this.option(optionName), optionName);
+            return optionName === 'values' ?
```
ksercs: Shouldn't we move it to an appropriate strategy?

js/ui/calendar/ui.calendar.js
```diff
@@ -395,11 +417,39 @@ const Calendar = Editor.inherit({
...
         this._correctZoomLevel();
         this._initCurrentDate();
         this._initActions();
     },
 
+    _initSelectionStrategy: function() {
+        const strategyName = this._getSelectionStrategyName();
+        const strategy = SELECTION_STRATEGY_CLASSES[strategyName];
+
+        if(!(this._selectionStrategy && this._selectionStrategy.NAME === strategyName)) {
```
ksercs: Isn't it more readable, how do u think?
```suggestion
        if(!this._selectionStrategy || this._selectionStrategy.NAME !== strategyName)) {
```

js/ui/calendar/ui.calendar.js
```diff
@@ -417,7 +467,9 @@ const Calendar = Editor.inherit({
     },
 
     _initCurrentDate: function() {
-        const currentDate = this._getNormalizedDate(this._dateOption('value')) || this._getNormalizedDate(this.option('currentDate'));
+        const currentDate = this._getNormalizedDate(this._selectionStrategy.getDefaultCurrentDate())
+            || this._getNormalizedDate(this.option('currentDate'));
```
ksercs: Nope?
```suggestion
            ?? this._getNormalizedDate(this.option('currentDate'));
```

js/ui/calendar/ui.calendar.js
```diff
@@ -1139,26 +1190,15 @@ const Calendar = Editor.inherit({
...
 
-    _updateViewsValue: function(value) {
-        const newValue = value ? new Date(value) : null;
-
-        this._view.option('value', newValue);
-        this._beforeView && this._beforeView.option('value', newValue);
-        this._afterView && this._afterView.option('value', newValue);
+    _updateViewsOption: function(optionName, newValue) {
+        this._view.option(optionName, newValue);
+        this._beforeView && this._beforeView.option(optionName, newValue);
```
ksercs:
```suggestion
        this._beforeView?.option(optionName, newValue);
```

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,227 @@
+import { noop } from '../../core/utils/common';
+import Class from '../../core/class';
+import dateUtils from '../../core/utils/date';
+
+const abstract = Class.abstract;
```
ksercs: y do we need this constant?

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,227 @@
...
+
+const DAY_INTERVAL = 86400000;
+
+const CalendarSelectionStrategy = Class.inherit({
```
ksercs: please, let's move different strategies to different files how it's already done in
[selection strategies](https://github.com/DevExpress/DevExtreme/tree/23_1/js/ui/selection)
[datebox strategies](https://github.com/DevExpress/DevExtreme/tree/23_1/js/ui/date_box)
and others.
It's very difficult to search for an overided method when all strategies are places in a single file.

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,227 @@
...
+    updateAriaSelected: function(value, previousValue) {
+        this.calendar._updateAriaSelected(value, previousValue);
+
+        if(value[0] && this.calendar.option('currentDate').getTime() === value[0].getTime()) {
+            this.calendar._updateAriaId(value[0]);
+        }
+    },
+
+    processValueChanged: function(value, previousValue) {
+        value = (value || []).map((item) => this._convertToDate(item));
```
ksercs:
```suggestion
        value = value?.map((item) => this._convertToDate(item)) ?? [];
```

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,227 @@
...
+        previousValue = (previousValue || []).map((item) => this._convertToDate(item));
+
+        this._updateViewsValue(value);
+        this.updateAriaSelected(value, previousValue);
+
+        !this._currentDateChanged && this.calendar._initCurrentDate();
```
ksercs: It's not a good practice to use logical operator instead of usual statement
https://github.com/airbnb/javascript#control-statements--value-selection
```suggestion
        if(!this._currentDateChanged) {
            this.calendar._initCurrentDate();
        }
```

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,227 @@
...
+    _updateViewsOption: function(optionName, optionValue) {
+        this.calendar._updateViewsOption(optionName, optionValue);
+    },
+
+    _updateViewsValue: function(value) {
+        this._updateViewsOption('value', value);
+    },
+
+    _updateCurrentDate: function(value) {
+        this.calendar.option('currentDate', value || new Date());
```
ksercs: nope?
```suggestion
        this.calendar.option('currentDate', value ?? new Date());
```

js/ui/calendar/ui.calendar.selection.strategy.js
ksercs: Shouldn't we use usual ES6 classes instead of out custom `Class.inherit`?

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,227 @@
...
+            values.push(selectedValue);
+        }
+
+        this._updateCurrentDate(selectedValue);
+        this._currentDateChanged = true;
+        this.dateValue(values, e);
+    },
+
+    updateAriaSelected: function(value, previousValue) {
+        value = value ?? this.dateOption('values');
```
ksercs:
```suggestion
        value ??= this.dateOption('values');
```

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,227 @@
...
+        previousValue = previousValue ?? [];
+
+        this.callBase(value, previousValue);
+    },
+
+    getDefaultCurrentDate: function() {
+        return this._getLowestDateInArray(this.dateOption('values').filter(value => value));
```
ksercs: Why do we need this `filter` here?

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,227 @@
...
+    },
+});
+
+const CalendarRangeSelectionStrategy = CalendarSelectionStrategy.inherit({
```
ksercs: shouldn't we inherit it from MultiSelectionStrategy to reduce copypaste?

scss/widgets/material/calendar/_index.scss
```diff
@@ -233,6 +233,23 @@ $material-calendar-height: $material-calendar-body-height + $material-calendar-n
...
+
+  &.dx-calendar-range-start-date {
+    border-top-left-radius: 50%;
+    border-bottom-left-radius: 50%;
+  }
+
+  &.dx-calendar-range-end-date {
+    border-top-right-radius: 50%;
+    border-bottom-right-radius: 50%;
+  }
```
ksercs, line 252: What about screenshot tests?

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,227 @@
...
+    },
+
+    _getRange: function() {
+        const [startDate, endDate] = this._getValues();
+        return this._getDaysInRange(startDate, endDate);
+    },
+
+    _getDaysInRange: function(startDate, endDate) {
+        return startDate && endDate
+            ? [...dateUtils.getDatesOfInterval(startDate, endDate, DAY_INTERVAL), endDate]
```
ksercs: Cannot we implement a range strategy without settings all range days as a value? I think it's a main reason of a performance issue.
It would be perfect if we rerender only visible rows (and prepare data only for this) when a range is set.

js/ui/calendar/ui.calendar.js
Zedwag: done

js/ui/calendar/ui.calendar.base_view.js
Zedwag: done

testing/tests/DevExpress.ui.widgets.editors/calendar.tests.js
```diff
@@ -1819,6 +1866,295 @@ QUnit.module('Options', {
...
+                    selectionMode: 'range'
+                });
+                const $cell = $(getCurrentViewInstance(this.calendar).$element().find('*[data-value="2023/01/15"]'));
+
+                $cell.trigger('mouseenter');
+
+                assert.ok($cell.hasClass(CALENDAR_RANGE_DATE_CLASS));
+            });
+        });
+    });
```
ksercs, line 2168: shouldn't we add a couple of tests for a runtime change?

js/ui/calendar/ui.calendar.js
Zedwag: a bit, fixed

js/ui/calendar/ui.calendar.js
Zedwag: done

js/ui/calendar/ui.calendar.js
Zedwag: done

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: done

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: done

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: done

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: done

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: no need anymore

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: yeah, I saw how it's done there, but in this case there are only 200 lines for all strategies, as for me it's more handfull to keep them in one file. But ok, I mooved them to separate files

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: done

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: I don't think it's will reduce copypaste, range and multi methods are pretty different

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: It's clear that's is the main reason, but I'm not sure how to do it in another way. I'll try couple ideas

testing/tests/DevExpress.ui.widgets.editors/calendar.tests.js
Zedwag: what tests do u mean, there are tests for runtime switching from single to multi/range and vice verse

scss/widgets/material/calendar/_index.scss
Zedwag: added

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: to avoid passing nulls there

js/ui/calendar/ui.calendar.js
Zedwag: it can also be `min`/`max` so we shouldn't change this

js/ui/calendar/ui.calendar.js
Zedwag: done

js/ui/calendar/ui.calendar.js
```diff
@@ -1180,8 +1218,8 @@ const Calendar = Editor.inherit({
     },
 
     _optionChanged: function(args) {
-        let value = args.value;
-        let previousValue = args.previousValue;
+        const value = args.value;
+        const previousValue = args.previousValue;
```
EugeniyKiyashko: May we use destructuring assignment here?
const { value, previousValue } = args;

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: they will be converted to January 1st 1970 and win this comparing

js/ui/calendar/ui.calendar.js
Zedwag: changed this

just to note: it's actually was not my change, I just removed value/previousValue changing below and lint replaced let with const on ctrl+s

js/ui/calendar/ui.calendar.selection.strategy.js
ksercs: can we make it more explicit? To not refactor erroneously in the future. For example,
```javascript
const dates = this.dateOption('values').filter(value => value);
return this._getLowestDateInArray(dates);
```
Do we have a test for this case?

Zedwag: Yep https://github.com/DevExpress/DevExtreme/pull/23835/files#diff-a48a13dcd405bca6bfcaf3fa67e23b94ab790fff3eebd31fa89f5a446305088eR1945

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: Now range is reduced, it can not be more than +-days from currentDate

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -198,12 +205,29 @@ const BaseView = Widget.inherit({
                 });
             }
         }));
+
+        if(this.option('selectionMode') === 'range') {
+            this._createCellHoverAction();
+            eventsEngine.off(this._$table, CALENDAR_DXHOVER_EVENT_NAME);
```
EugeniyKiyashko: Do we may concrete what hover event using? 

CALENDAR_DXHOVERSTART_EVENT_NAME
I think it gives us more information what happen. If we decide to use subscription on `onhoverend` event we must do this renaming in any case.

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -198,12 +205,29 @@ const BaseView = Widget.inherit({
...
+            eventsEngine.on(this._$table, CALENDAR_DXHOVER_EVENT_NAME, `td:not(.${CALENDAR_WEEK_NUMBER_CELL_CLASS})`, ((e) => {
```
EugeniyKiyashko: What do you think if we extract common selector `td:not(.${CALENDAR_WEEK_NUMBER_CELL_CLASS})` to const variable, it allows us to avoid duplication

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -198,12 +205,29 @@ const BaseView = Widget.inherit({
                 });
             }
         }));
+
```
EugeniyKiyashko, line 209: I undarstand that you don't create _renderEvents() function, but it name seems inappropriate. I would expect something like `initActions` or `createActions` here. What do you think to change this function name? In any case it a note only.

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -253,36 +277,44 @@ const BaseView = Widget.inherit({
...
-        } else {
-            this.option('value', null);
+        let value = this.option('value');
+        if(!Array.isArray(value)) {
+            value = [value];
         }
+
+        this._selectedCells?.forEach((cell) => { cell.removeClass(CALENDAR_SELECTED_DATE_CLASS); });
+
+        const selectedCells = value.map((value) => this._getCellByDate(value));
```
EugeniyKiyashko: Do we need to create additional constant variable to keep `selectedCells`, or we may to use `this._selectedCells` directly here?

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -253,36 +277,44 @@ const BaseView = Widget.inherit({
...
+
+        this._selectedCells = selectedCells;
+        selectedCells.forEach((cell) => { cell.addClass(CALENDAR_SELECTED_DATE_CLASS); });
```
EugeniyKiyashko: If `this._selectedCells` is array of jquery objects we shouldn't use iterator for adding class to each object. We may call `this._selectedCells.addClass(CALENDAR_SELECTED_DATE_CLASS)` only. Maybe I am missing something. Let me know if it's not acceptable here.

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -253,36 +277,44 @@ const BaseView = Widget.inherit({
...
-        if(!this.option('allowValueSelection')) {
+    _renderRange: function() {
+        const { allowValueSelection, selectionMode } = this.option();
+        if(!allowValueSelection || selectionMode !== 'range') {
             return;
         }
 
-        const value = this.option('value');
-        const selectedCell = this._getCellByDate(value);
+        this._rangeCells?.forEach((cell) => { cell.removeClass(CALENDAR_RANGE_DATE_CLASS); });
```
EugeniyKiyashko: We may declare `cache variables` this._rangeCells, this._rangeStartDateCell, this._rangeEndDateCell and other on initialization for modes that require them. It allows us to avoid using ?. and forEach() wrappers and make code clear and more predictable.

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -253,36 +277,44 @@ const BaseView = Widget.inherit({
...
+        this._rangeStartDateCell?.removeClass(CALENDAR_RANGE_START_DATE_CLASS);
+        this._rangeEndDateCell?.removeClass(CALENDAR_RANGE_END_DATE_CLASS);
 
-        if(this._selectedCell) {
-            this._selectedCell.removeClass(CALENDAR_SELECTED_DATE_CLASS);
-        }
+        const rangeCells = this.option('range').map((value) => this._getCellByDate(value));
+        const [rangeStartDateCell, rangeEndDateCell] = this.option('value').map((value) => this._getCellByDate(value));
+
+        this._rangeCells = rangeCells;
```
EugeniyKiyashko: Look like that `_rangeCells` and other similar properties store array of jquery objects. I think it should be name as `this._$rangeCells` for better understanding what property stores.

js/ui/calendar/ui.calendar.js
```diff
@@ -243,28 +256,36 @@ const Calendar = Editor.inherit({
...
         }
-        this._dateOption('value', value);
+        this._dateOption(optionName, value);
     },
 
     _dateOption: function(optionName, optionValue) {
         if(arguments.length === 1) {
-            return this._convertToDate(this.option(optionName), optionName);
+            const values = this.option('values') ?? [];
+            return optionName === 'values'
```
EugeniyKiyashko, line 278: Looks like this condition and others should be processed on the 'strategy' level. What do you think, it's possible?

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,62 @@
+class CalendarSelectionStrategy {
+    constructor(component) {
+        this.calendar = component;
```
EugeniyKiyashko, line 3: I am believe that strategy object shouldn't know anything about component and must not manipulate state of the component. It should be a set of the service functions otherwise, the code connectivity is very high.

js/ui/calendar/ui.calendar.js
```diff
@@ -21,6 +21,9 @@ import messageLocalization from '../../localization/message';
 import dateLocalization from '../../localization/date';
 import { FunctionTemplate } from '../../core/templates/function_template';
 import { isCommandKeyPressed } from '../../events/utils/index';
+import CalendarSingleSelectionStrategy from './ui.calendar.single.selection.strategy';
```
EugeniyKiyashko, line 24: Why you decided that selection strategies should be part of calendar, not calendar.base part?

scss/widgets/generic/calendar/_index.scss
```diff
@@ -226,6 +226,19 @@ $generic-calendar-height: $generic-calendar-body-height + $generic-calendar-navi
       }
     }
   }
+
+  &.dx-calendar-range-date:not(.dx-calendar-selected-date) {
+    &,
+    &.dx-calendar-today {
+      color: $base-inverted-text-color;
+      box-shadow: inset 0 0 0 1000px lighten($calendar-cell-selected-bg, 30%);
```
EugeniyKiyashko, line 234: Why we use 1000px here?

scss/widgets/generic/calendar/_index.scss
```diff
@@ -226,6 +226,19 @@ $generic-calendar-height: $generic-calendar-body-height + $generic-calendar-navi
...
+      font-weight: normal;
+
+      &.dx-calendar-contoured-date {
+        box-shadow: inset 0 0 0 1px $calendar-cell-contoured-border-color, inset 0 0 0 1000px lighten($calendar-cell-selected-bg, 30%);
```
EugeniyKiyashko, line 238: What do you think about extraction these values to separate variable?

js/ui/calendar/ui.calendar.selection.strategy.js
ksercs: I agree with u, but the actual approach is already used in all out strategies and it's not a fast task to separate a responsibility between calendar and its strategies now.

I suggest to note this in spec / card, and postpone this improvement.

js/ui/calendar/ui.calendar.selection.strategy.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    dateOption(optionName) {
```
ksercs, line 6: I see that some methods are not overided in other strategies, so y do we need to implement it?

js/ui/calendar/ui.calendar.js
EugeniyKiyashko: Yes, but you change let to const and use them in you code. The destructuring assignment `const { value, previousValue, name } = args;` looks appropriate here. It's recomendation only and you should to take a final decision to do or not to do.

js/ui/calendar/ui.calendar.base_view.js
Zedwag: I agree that name `renderEvents` sounds pretty inappropriate, but I checked our other components and it seems like we use `initActions` function only for creating actions and `renderEvents` when we also add some events listeners to elements. This function name is common for other editors, so I prefer to keep this as it is
image

js/ui/calendar/ui.calendar.base_view.js
Zedwag: Sure we do not need this, it just makes lines with addClass a bit more clear (but costs exstra lines to create additional constant). I changed this, though I am not sure which way is better.

js/ui/calendar/ui.calendar.base_view.js
Zedwag: Not sure I understand u correctly. We cannot call jQuery methods on js array, even if it is array of jQuery objects: https://codepen.io/zedwag/pen/rNZedGd?editors=0010

js/ui/calendar/ui.calendar.selection.strategy.js
Zedwag: I added some methods like this one in base strategy just to make code in strategies a bit more clear, so we can call `this.dateOption()` instead of calling `this.calendar._dateOption()`.

js/ui/calendar/ui.calendar.base_view.js
Zedwag: fair point, updated this

js/ui/calendar/ui.calendar.base_view.js
Zedwag: done

js/ui/calendar/ui.calendar.base_view.js
Zedwag: fair point, renamed

js/ui/calendar/ui.calendar.js
Zedwag: Well we do not have calendar.base component. There is only base view, but as far as stratigy in needed to deside what options we should pass to view, I do not understand how strategies can be part of view

scss/widgets/generic/calendar/_index.scss
Zedwag: I have no idea actually, for now generic range styles are copypaste of generic selected styles, but with a lightened color. As far as generic design of DateRangeBox is not ready, I don't see any point of concentrating attention on temporary generic styles

js/ui/calendar/ui.calendar.base_view.js
ksercs: I agree, this question is not for this PR

js/ui/calendar/ui.calendar.selection.strategy.js
ksercs: We discussed it in voice. We agree with u, but decided to postpone this refactoring. It was implemented this way because all strategies in DevExtreme uses parent component instance.

js/ui/calendar/ui.calendar.js
Zedwag: Discussed this Anton, we agree that as far as strategies define how to render ranges and values, it's better to place them on View's level. But we have some functions (like for instance setting aria attributes) on ui.calendar level that also depends on strategies. So, moving strategies to view need more calendar refactoring and now we don't have time for this

js/ui/calendar/ui.calendar.js
Zedwag: It's sure possible but requeires to rework this function, cause it's too "wide" now.  For example it's also used to get `min/max` dates, so we will also need to check if optionName is `values` or not on multiselect strategies. Also it's used in `_updateTimeComponent` function that set hours/minutes/seconds of current date depending on `value`.
Without refactoring this function on strategies will be still pretty ugly, so I desided to not move it there

js/ui/calendar/ui.calendar.base_view.js
EugeniyKiyashko: There are troubles with this approach. Let's skip it.

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -187,23 +194,41 @@ const BaseView = Widget.inherit({
     },
 
     _renderEvents: function() {
+        const excludeWeekCellsSelector = `td:not(.${CALENDAR_WEEK_NUMBER_CELL_CLASS})`;
```
ksercs: maybe it's better to declare this constant on file's top nest other constants?
```javascript
const NOT_WEEK_CELL_SELECTOR = `td:not(.${CALENDAR_WEEK_NUMBER_CELL_CLASS})`
```

js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -253,36 +278,38 @@ const BaseView = Widget.inherit({
...
-        } else {
-            this.option('value', null);
+        let value = this.option('value');
+        if(!Array.isArray(value)) {
+            value = [value];
         }
+
+        this._$selectedCells?.forEach((cell) => { cell.removeClass(CALENDAR_SELECTED_DATE_CLASS); });
+        this._$selectedCells = value.map((value) => this._getCellByDate(value));
+        this._$selectedCells.forEach((cell) => { cell.addClass(CALENDAR_SELECTED_DATE_CLASS); });
```
ksercs: We can use a function to reduce a copypaste
 ```javascript
        _toggleSelectedClass($cell, isSelected) {
            $cell?.toggleClass(CALENDAR_SELECTED_DATE_CLASS, isSelected);
        }
```
We can use it in ~8 rows, check all places plz

File: js/ui/calendar/ui.calendar.range.selection.strategy.js [hunk 53]
```diff
@@ -0,0 +1,100 @@
...
+        const [startDate, endDate] = this._getValues();
+        return this._getDaysInRange(startDate, endDate);
+    }
+
+    _getDaysInRange(startDate, endDate) {
+        if(!startDate || !endDate) {
+            return [];
+        }
+
+        // TODO: rework this range redusing algorythm if we desided to puplic multiselect feature
```
ksercs:
```suggestion
        // TODO: Rework this range reducing algorithm to support different multi views 
                 and optimise single views.
```

js/ui/calendar/ui.calendar.base_view.js
Zedwag: there 4 different classes, so I added `_toggleClass($cell, className, isSelected)

ksercs (approved):

PR #23902, opened 2023-02-20
DataGrid: Fix the 'valueChangeEvent' issue when using dxSelectBox/dxLookup in the editing form (T1145047)

markallenramirez:


pomahtri (changes requested):
transferred from dm:

- check case when `dataType: number`, `editorType: dxTextBox` & add test for it
- merge tests for SelectBox & Lookup into one using forEach

pomahtri (approved):

williamvinogradov (approved):

PR #23907, opened 2023-02-20
TreeView: add expander icon customisation

AlexanderMoiseev:


IlyaKhD (dismissed):

js/ui/tree_view.d.ts
```diff
@@ -314,6 +314,18 @@ export interface dxTreeViewOptions<TKey = any>
      * @public
      */
     showCheckBoxesMode?: TreeViewCheckBoxMode;
+    /**
+     * @docid
+     * @default null
+     * @public
+     */
+    collapseButtonIcon?: string;
```
ksercs: if we use `null` as a default value we should specify a type as `string | null`.
(I understand that there are a lot of places when we specify `null` as a default value but don't extend a type, but it's not a good approach)

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -154,6 +160,8 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
             selectNodesRecursive: true,
             expandNodesRecursive: true,
             showCheckBoxesMode: 'none',
+            expandButtonIcon: null,
```
ksercs: @vladaskorohodova  don't we wont to simplify names to `expandIcon` / `collapseIcon`?

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -311,6 +319,8 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
             case 'animationEnabled':
             case 'virtualModeEnabled':
             case 'selectByClick':
+            case 'expandButtonIcon':
+            case 'collapseButtonIcon':
```
ksercs: Hm, I think we should support runtime property change.
Sample:
```javascript
const treeView = new TreeView(element, { expandIcon: 'minus' });
treeView.option('expandIcon', null);
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -672,6 +686,8 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
 
         $node.addClass(showCheckBox ? ITEM_WITH_CHECKBOX_CLASS : ITEM_WITHOUT_CHECKBOX_CLASS);
         $node.toggleClass(INVISIBLE_STATE_CLASS, nodeData.item.visible === false);
+
+        this._showCustomExpanderIcons() && $node.addClass(ITEM_WITH_CUSTOM_EXPANDER_ICON_CLASS);
```
ksercs: It's not a good practice to use logical operator instead of usual statement
https://github.com/airbnb/javascript#control-statements--value-selection
```suggestion
        if(this._showCustomExpanderIcons()) {
            $node.addClass(ITEM_WITH_CUSTOM_EXPANDER_ICON_CLASS);
        }
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -861,21 +878,53 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         $icon.hide();
     },
 
-    _renderToggleItemVisibilityIcon: function($node, node) {
-        const $icon = $('<div>')
-            .addClass(TOGGLE_ITEM_VISIBILITY_CLASS)
-            .appendTo($node);
+    _renderToggleItemVisibilityCustomIcon: function($node, node) {
+        const $expandIcon = getImageContainer(this.option('expandButtonIcon') ?? this.option('collapseButtonIcon'));
```
ksercs: We can simplify this part
```javascript
    const { expandButtonIcon, collapseButtonIcon } = this.option();
    const $expandIcon = getImageContainer(expandButtonIcon ?? collapseButtonIcon);
    const $collapseIcon = getImageContainer(collapseButtonIcon ?? expandButtonIcon);
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -861,21 +878,53 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
+        const $collapseIcon = getImageContainer(this.option('collapseButtonIcon') ?? this.option('expandButtonIcon'));
+
+        $expandIcon.appendTo($node).addClass('dx-custom-expander-icon');
```
ksercs: 1. Let's move `'dx-custom-expander-icon'` to a constant. This name is also too common for me. Maybe `dx-treeview-custom-expander-icon`, or like this?
2. I think it's better to separate `appendTo` from classes adding because it's really important moment which can be not noted now.
3. `addClass` accepts an array as a parameter, we can write
```javascript
    $expandIcon.addClass([
        class1,
        class2,
        class3,
       ...
    ]);
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -861,21 +878,53 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
+        $expandIcon.addClass(CUSTOM_ICON_TOGGLE_ITEM_VISIBILITY_CLASS);
+        $expandIcon.addClass(CUSTOM_EXPAND_ICON_CLASS);
+
+        $collapseIcon.addClass(CUSTOM_ICON_TOGGLE_ITEM_VISIBILITY_CLASS);
+        $collapseIcon.addClass(CUSTOM_COLLAPSE_ICON_CLASS);
+
+        this._toggleCustomExpanderBtnIcons($expandIcon, $collapseIcon, node.internalFields.expanded);
 
         if(node.internalFields.expanded) {
-            $icon.addClass(TOGGLE_ITEM_VISIBILITY_OPENED_CLASS);
```
ksercs: hm, don't we need some analogue here?

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -653,6 +663,10 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         return this.option('showCheckBoxesMode') !== 'none';
     },
 
+    _showCustomExpanderIcons: function() {
```
ksercs: This function is named as `show`, but we show nothing here :)
```suggestion
    _hasCustomExpanderIcons: function() {
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -861,21 +878,53 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
+            $collapseIcon.addClass(DISABLED_STATE_CLASS);
         }
 
-        this._renderToggleItemVisibilityIconClick($icon, node);
+        this._renderToggleItemVisibilityIconClick($collapseIcon, node);
+        this._renderToggleItemVisibilityIconClick($expandIcon, node);
+    },
+
+    _renderToggleItemVisibilityIcon: function($node, node) {
+        if(this._showCustomExpanderIcons()) {
```
ksercs: I think we should move this condition to `_renderChildren`.
According to the name, this function should only render toggle icon, it should not decide which elements should be rendered

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -899,9 +958,14 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
             }
         }
 
-        const $icon = $node.children('.' + TOGGLE_ITEM_VISIBILITY_CLASS);
+        const $icon = $node.children(`.${TOGGLE_ITEM_VISIBILITY_CLASS},.${CUSTOM_ICON_TOGGLE_ITEM_VISIBILITY_CLASS}`);
```
ksercs: I see this selector second time, maybe we should move it to a constant?

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -33,10 +34,15 @@ const IS_LEAF = `${NODE_CLASS}-is-leaf`;
 
 const ITEM_CLASS = `${WIDGET_CLASS}-item`;
 const ITEM_WITH_CHECKBOX_CLASS = `${ITEM_CLASS}-with-checkbox`;
+const ITEM_WITH_CUSTOM_EXPANDER_ICON_CLASS = `${ITEM_CLASS}-with-custom-expander-icon`;
 const ITEM_WITHOUT_CHECKBOX_CLASS = `${ITEM_CLASS}-without-checkbox`;
 const ITEM_DATA_KEY = `${ITEM_CLASS}-data`;
 
 const TOGGLE_ITEM_VISIBILITY_CLASS = `${WIDGET_CLASS}-toggle-item-visibility`;
+const CUSTOM_ICON_TOGGLE_ITEM_VISIBILITY_CLASS = `${WIDGET_CLASS}-custom-icon-toggle-item-visibility`;
```
ksercs: I don't see y we need this class.
Seems like all selectors are build based on `TOGGLE_ITEM_VISIBILITY_CLASS`.
Am i missing smth?

testing/testcafe/tests/navigation/treeView/common.ts
```diff
@@ -72,3 +72,30 @@ test('TreeView: height should be calculated correctly when searchEnabled is true
...
+
+[true, false].forEach((rtlEnabled) => {
+  ['normal', 'none'].forEach((showCheckBoxesMode) => {
+    test(`TreeView with custom expander icons,showCheckBoxesMode=${showCheckBoxesMode}`, async (t) => {
+      const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+      const screenshotName = `Treeview with custom expander icons cbm=${showCheckBoxesMode},rtl=${rtlEnabled}.png`;
+
+      await testScreenshot(t, takeScreenshot, screenshotName, { element: '#container', shouldTestInCompact: true });
+      await testScreenshot(t, takeScreenshot, screenshotName, { element: '#container', theme: 'material.blue.light', shouldTestInCompact: true });
```
ksercs: U don't need to write these - this tests package is automatically run both in generic and material themes

testing/tests/DevExpress.ui.widgets/treeViewParts/expandedItems.js
```diff
@@ -187,6 +187,22 @@ module('Expanded items', {
...
     });
 
+    test('item with custom expander icons should expand on click', function(assert) {
+        const data = $.extend(true, [], DATA[5]);
+        const $treeView = initTree({
+            items: data,
+            expandButtonIcon: 'add',
+            collapseButtonIcon: 'add',
+        });
+        const treeViewInstance = $treeView.dxTreeView('instance');
```
ksercs: it's ok to name it just a `treeView`. It's our standard practice

testing/tests/DevExpress.ui.widgets/treeViewParts/expandedItems.js
```diff
@@ -187,6 +187,22 @@ module('Expanded items', {
...
+        const $firstItem = $treeView.find('.' + internals.ITEM_CLASS).eq(0);
+        const $icon = $firstItem.parent().find('> .' + internals.CUSTOM_EXPAND_ICON_CLASS);
```
ksercs: 1. Cannot we simplify this icon getter?
As far as i see, u have a special class for custom icons, cannot we use it here?
2. We usually use template literals instead of a concatenation - it's just easier to read

Let's update everywhere in your tests

testing/tests/DevExpress.ui.widgets/treeViewParts/expandedItems.js
```diff
@@ -218,6 +251,48 @@ module('Expanded items', {
         assert.ok(treeView.option('items')[0].expanded, 'disabled item was not expanded');
     });
 
+
+    test('expand and collapse icons should change visibility onClick', function(assert) {
```
ksercs:
```suggestion
    test('expand and collapse icons should change visibility on click', function(assert) {
```

testing/tests/DevExpress.ui.widgets/treeViewParts/expandedItems.js
```diff
@@ -264,6 +339,20 @@ module('Expanded items', {
...
+        const data = $.extend(true, [], DATA[5]);
+
+        const $treeView = initTree({
+            items: data,
+            expandButtonIcon: 'add',
+            collapseButtonIcon: 'add',
+        });
+        const treeView = $treeView.dxTreeView('instance');
+        const $toggleExpandIcon = $($treeView.find('.' + internals.CUSTOM_EXPAND_ICON_CLASS).eq(0));
+        treeView.on('itemExpanded', () => assert.ok(true, 'itemExpanded was fired'));
```
ksercs: 1. We can use `sinon.spy` here to make this test clearer.
2. Let's separate [arrange, action, assert](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/) parts with an empty row.

testing/tests/DevExpress.ui.widgets/treeViewParts/virtualMode.js
```diff
@@ -199,6 +199,39 @@ QUnit.test('Remove toggle icon after expand childless item', function(assert) {
...
+
+    assert.ok($expandIconsAfterExpandChildlessNode.eq(0).is(':visible'));
+    assert.ok($expandIconsAfterExpandChildlessNode.eq(1).is(':visible'));
+    assert.notOk($expandIconsAfterExpandChildlessNode.eq(2).is(':visible'));
+
+    const $collapseIconsAfterExpandChildlessNode = this.$element.find(`.${internals.CUSTOM_COLLAPSE_ICON_CLASS}`);
+    assert.notOk($collapseIconsAfterExpandChildlessNode.eq(0).is(':visible'));
+    assert.notOk($collapseIconsAfterExpandChildlessNode.eq(1).is(':visible'));
+    assert.notOk($collapseIconsAfterExpandChildlessNode.eq(2).is(':visible'));
+});
```
ksercs, line 214: Cannot we simplify this test?
Seems like it tests several things at one time. It's not an easy task for me to understand what is going here

Moreover, it's not clear for me how we can "expand childness node")

testing/tests/DevExpress.ui.widgets/treeViewParts/virtualMode.js
```diff
@@ -199,6 +199,39 @@ QUnit.test('Remove toggle icon after expand childless item', function(assert) {
...
+
 QUnit.test('Remove loadindicator after expand childless item', function(assert) {
     new TreeView(this.$element, {
```
ksercs, line 217: I also suggest u to add tests in `treeView.markup.tests` file to check that necessary icons are rendered if an appropriate property is passed in initial config or property value is changed in runtime, check that all necessary classes are added / toggles and so on.

jfyi: `*.markup.tests` files are also run in SSR mode

testing/tests/DevExpress.ui.widgets/treeViewParts/expandedItems.js
```diff
@@ -264,6 +339,20 @@ module('Expanded items', {
...
+        $toggleExpandIcon.trigger('dxclick');
+    });
+
```
ksercs, line 378: I think we should add a test for a case when we use expandIcon instead of collapseIcon and vice versa

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -33,10 +34,15 @@ const IS_LEAF = `${NODE_CLASS}-is-leaf`;
...
+const CUSTOM_EXPAND_ICON_CLASS = `${CUSTOM_ICON_TOGGLE_ITEM_VISIBILITY_CLASS}-expand-button-icon`;
```
ksercs: These names are too long)
I also didn't catch y we need to add this classes because we didn't add such classes for usual icons

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -899,9 +958,14 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
             }
         }
 
-        const $icon = $node.children('.' + TOGGLE_ITEM_VISIBILITY_CLASS);
+        const $icon = $node.children(`.${TOGGLE_ITEM_VISIBILITY_CLASS},.${CUSTOM_ICON_TOGGLE_ITEM_VISIBILITY_CLASS}`);
+
         const $nodeContainer = $node.children(`.${NODE_CONTAINER_CLASS}`);
 
+        if(this._showCustomExpanderIcons() && $nodeContainer.not(':empty').length) {
+            this._toggleCustomExpanderBtnIcons($node.children(`.${CUSTOM_EXPAND_ICON_CLASS}`), $node.children(`.${CUSTOM_COLLAPSE_ICON_CLASS}`), state);
```
ksercs: I removed this code, run test and found them green :)

Why do we need this?

testing/tests/DevExpress.ui.widgets/treeViewParts/virtualMode.js
AlexanderMoiseev: Ok i'll try)
Moreover, it's not clear for me how we can "expand childness node") - this is a virtual mode, so in some cases we dont know if parent contains childrem. we have similar test for default icons already

scss/widgets/material/treeView/_index.scss
```diff
@@ -95,6 +103,10 @@ $material-treeview-item-with-checkbox-offset: $material-treeview-checkbox-offset
 }
 
 .dx-treeview-item-without-checkbox {
+  &.dx-treeview-item-with-custom-expander-icon > .dx-treeview-item {
+    padding-left: $material-treeview-item-side-padding;
```
ksercs, line 113: Why don't we want add this padding in generic theme?

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: i've managed to get rid of some classes

scss/widgets/material/treeView/_index.scss
AlexanderMoiseev: lets ask Pavel tomorrow)

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: it works a little different, i switch visibility, and i do it with _toggleCustomExpanderIcons function. But here, for default icons, we add class that helps to set content with :before pseudoclass. I hoped we will rewrite this later and use my approach, then we set visibility instead of switching content with :before. BTW i left a comment on your anothere note regarding TOGGLE_ITEM_VISIBILITY_CLASS

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: good catch )

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: you mean what inside brackets? but there are 2 selectors, separated with a comma. or you meant move these 2 selectors as constant?)

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: it toggles icon, but before this we check if icon needs to be toggled, since in virtual mode childless icons should not be toggled (test - No custom expander icons should be visible after expand childless item ). Anyway i'll look into this, you right that some test is needed here for this toggle line too

js/ui/tree_view.d.ts
AlexanderMoiseev: do you mean for readability? isn't '?' implies that this is a nullable type ?

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: i changed code a lot, please take a look again. But TOGGLE_ITEM_VISIBILITY_CLASS is only for default icons, we cant use them for new custom icons because of the styles they have, like these on the screenshot below and others
image

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -861,21 +878,53 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
-            .addClass(TOGGLE_ITEM_VISIBILITY_CLASS)
-            .appendTo($node);
+    _renderToggleItemVisibilityCustomIcon: function($node, node) {
+        const $expandIcon = getImageContainer(this.option('expandButtonIcon') ?? this.option('collapseButtonIcon'));
+        const $collapseIcon = getImageContainer(this.option('collapseButtonIcon') ?? this.option('expandButtonIcon'));
+
+        $expandIcon.appendTo($node).addClass('dx-custom-expander-icon');
+        $collapseIcon.appendTo($node).addClass('dx-custom-expander-icon');
+
+        $expandIcon.addClass(CUSTOM_ICON_TOGGLE_ITEM_VISIBILITY_CLASS);
```
Krijovnick: There are too many the same actions with expandIcon and collapseIcon:
1. appendTo same container
2. add same classes
3. then same classes for disabled state
4.  then call _renderToggleItemVisibilityIconClick method for both

Can we all of this to move in extra method and call it with each icon separatly?

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: "I think it's better to separate appendTo from classes adding because it's really important moment which can be not noted now." 
What about splitting into separate lines? seems obvious to me and hard to miss this important moment:
```
$expandIcon.appendTo($node)
              .addClass('dx-custom-expander-icon');
```

testing/testcafe/tests/navigation/treeView/common.ts
AlexanderMoiseev: `await testScreenshot(t, takeScreenshot, screenshotName, { element: '#container', theme: 'material.blue.light', shouldTestInCompact: true });
`

I don't think so, if i remove the line you commented on only generic theme screenshots are taken 
image
And i cant find this logic in code, am i missing something?

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: ok, actually i managed to get rid of some of these classes, so this note is not actual anymore

testing/tests/DevExpress.ui.widgets/treeViewParts/expandedItems.js
AlexanderMoiseev: roger

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: update: test added

testing/testcafe/tests/navigation/treeView/common.ts
EugeniyKiyashko: Perhaps, you added etalons and after deleting this line the tests aren't failed. I am convinсed that we can delete this line.

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -323,6 +330,10 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
                 this._dataAdapter.setOption('recursiveSelection', args.value);
                 this.repaint();
                 break;
+            case 'expandIcon':
+            case 'collapseIcon':
+                this.repaint();
```
ksercs, line 335: Cannot we optimize it to rerender only a changed icon?
`repaint` will completely rerender the component.

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -691,7 +711,24 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
             return;
         }
 
-        this._renderToggleItemVisibilityIcon($node, node);
+        if(this._hasCustomExpanderIcons()) {
+            this._renderCustomExpanderIcons($node, node);
+        } else {
+            const $icon = $('<div>')
```
ksercs: Let's move this code to a new `_renderDefaultExpanderIcons` function to simplify this one?

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -849,27 +888,20 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         return this._updateExpandedItemsUI(node, state, e);
     },
 
-    _createLoadIndicator: function($node) {
-        const $icon = $node.children('.' + TOGGLE_ITEM_VISIBILITY_CLASS);
+    _nodeHasRenderedChildren($node) {
```
ksercs, line 878: I really like how u refactored it!

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -849,27 +888,20 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
-    _renderToggleItemVisibilityIcon: function($node, node) {
-        const $icon = $('<div>')
-            .addClass(TOGGLE_ITEM_VISIBILITY_CLASS)
-            .appendTo($node);
-
-        if(node.internalFields.expanded) {
-            $icon.addClass(TOGGLE_ITEM_VISIBILITY_OPENED_CLASS);
-            $node.parent().addClass(OPENED_NODE_CONTAINER_CLASS);
-        }
+    _renderExpanderIcon: function($icon, iconClass, $node, node) {
```
ksercs: I think `$node` and `node` parameters should be the first while it's a signature of many methods here.

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -887,6 +935,16 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         });
     },
 
+    _toggleCustomExpanderIcons: function($expandBtnIcon, $collapseBtnIcon, isOpen) {
```
ksercs: I suggest u update the naming.
We can also use a `toggle` method to get rid of the conditions.

```javascript
_toggleCustomExpanderIcons: function($expandIcon, $collapseIcon, isNodeExpanded) {
    $collapseIcon.toggle(isNodeExpanded);
    $expandIcon.toggle(!isNodeExpanded);
},
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -878,6 +910,22 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
+    _renderCustomExpanderIcons: function($node, node) {
+        const { expandIcon, collapseIcon } = this.option();
+
+        const $expandIcon = getImageContainer(expandIcon ?? collapseIcon);
+        const $collapseIcon = getImageContainer(collapseIcon ?? expandIcon);
+
+        this._renderExpanderIcon($expandIcon, CUSTOM_EXPAND_ICON_CLASS, $node, node);
+        this._renderExpanderIcon($collapseIcon, CUSTOM_COLLAPSE_ICON_CLASS, $node, node);
+
+        if(node.internalFields.expanded) {
```
ksercs: I see this property is used in 2 rows. Let's move it to a variable at the beginning of the function to make the function clearer?
```javascript
const isNodeExpanded = node.internalFields.expanded;
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -899,10 +957,17 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
             }
         }
 
-        const $icon = $node.children('.' + TOGGLE_ITEM_VISIBILITY_CLASS);
-        const $nodeContainer = $node.children(`.${NODE_CONTAINER_CLASS}`);
+        if(this._hasCustomExpanderIcons()) {
+            if(this._nodeHasRenderedChildren($node)) {
+                this._toggleCustomExpanderIcons($node.children(`.${CUSTOM_EXPAND_ICON_CLASS}`), $node.children(`.${CUSTOM_COLLAPSE_ICON_CLASS}`), state);
```
ksercs: Let's move parameters to a separate variables to make this code more short, readable and easy to debug.
```
const $childExpandIcons = $node.children(`.${CUSTOM_EXPAND_ICON_CLASS}`);
const $childCollapseIcons = $node.children(`.${CUSTOM_COLLAPSE_ICON_CLASS}`);
this._toggleCustomExpanderIcons($childExpandIcons, $childCollapseIcons, state);
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -899,10 +957,17 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
             }
         }
 
-        const $icon = $node.children('.' + TOGGLE_ITEM_VISIBILITY_CLASS);
-        const $nodeContainer = $node.children(`.${NODE_CONTAINER_CLASS}`);
+        if(this._hasCustomExpanderIcons()) {
```
ksercs: we can get rid of nested condition if we write it this way:
```
if (!this._hasCustomExpanderIcons()) {
    ...
} else if (this._nodeHasRenderedChildren($node)) {
    ...
}
```

testing/testcafe/tests/navigation/treeView/common.ts
```diff
@@ -72,3 +72,29 @@ test('TreeView: height should be calculated correctly when searchEnabled is true
...
+      items: employees,
+      width: 300,
+      showCheckBoxesMode,
+      rtlEnabled,
+      expandIcon: 'add',
+      collapseIcon: 'minus',
+      itemTemplate(item) {
+        return `<div>${item.fullName} (${item.position})</div>`;
+      },
+    }));
```
ksercs, line 101: Please, check a checkbox's position on the first 4 screenshots. It seems that a vertical alignments worsens. Is this screenshot actual?

testing/testcafe/tests/navigation/treeView/common.ts
```diff
@@ -72,3 +72,29 @@ test('TreeView: height should be calculated correctly when searchEnabled is true
...
+      await t
+        .expect(compareResults.isValid())
+        .ok(compareResults.errorMessages());
+    }).before(async () => createWidget('dxTreeView', {
+      items: employees,
+      width: 300,
+      showCheckBoxesMode,
+      rtlEnabled,
+      expandIcon: 'add',
+      collapseIcon: 'minus',
```
ksercs, line 97: Let's hover a first item to make this tests even more useful? In this case we will check how item hovering looks like when custom icon is specified

testing/tests/DevExpress.ui.widgets/treeView.markup.tests.js
```diff
@@ -122,8 +128,110 @@ QUnit.module('markup', {
...
+            collapseIcon: 'add',
+        });
+        const treeView = $treeView.dxTreeView('instance');
+
+        treeView.option('collapseIcon', 'minus');
+
+        assert.ok($(CUSTOM_COLLAPSE_ICON_CLASS_SELECTOR).hasClass('dx-icon-minus'));
+    });
+
+    QUnit.test('when both custom expander icons are set to null default icons should be used', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('default icons should be rendered if both custom expander icons are not specified', function(assert) {
```

testing/tests/DevExpress.ui.widgets/treeView.markup.tests.js
```diff
@@ -122,8 +128,110 @@ QUnit.module('markup', {
...
+            items: this.plainItems,
+            expandIcon: 'add',
+            collapseIcon: 'add',
+        });
+
+        assert.equal($(CUSTOM_EXPAND_ICON_CLASS_SELECTOR).length, 0);
+        assert.equal($(CUSTOM_COLLAPSE_ICON_CLASS_SELECTOR).length, 0);
+    });
+
+    QUnit.test('expand icon should be shown if node is collasped', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('expand icon should be shown if node is collapsed', function(assert) {
```

testing/tests/DevExpress.ui.widgets/treeView.markup.tests.js
```diff
@@ -122,8 +128,110 @@ QUnit.module('markup', {
...
+            items: data,
+            expandIcon: 'add',
+            collapseIcon: 'add',
+        });
+        const $collapseIcon = $(CUSTOM_COLLAPSE_ICON_CLASS_SELECTOR).eq(0);
+
+        assert.ok(this.isVisible($collapseIcon));
+    });
+
+    QUnit.test('expand custom item should be set with collapse icon value if omited', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('collapseIcon value should be used for an expand icon if an expandIcon is not specified', function(assert) {
```

testing/tests/DevExpress.ui.widgets/treeView.markup.tests.js
```diff
@@ -122,8 +128,110 @@ QUnit.module('markup', {
...
+        initTree({
+            items: this.treeItems,
+            collapseIcon: 'minus',
+        });
+        const $expandIcon = $(CUSTOM_EXPAND_ICON_CLASS_SELECTOR).eq(0);
+
+        assert.ok($expandIcon.hasClass('dx-icon-minus'));
+    });
+
+    QUnit.test('collapse custom item should be set with expand icon value if omited', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('expandIcon value should be used for a collapse icon if collapseIcon is not specified', function(assert) {
```

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: may be, but why bother, this approach with repaint is used with _checkBoxModeChange and some other props for tree view, and i'm almost sure this is not something that will be used a lot

ksercs: okay, sounds earnestly)

testing/testcafe/tests/navigation/treeView/common.ts
AlexanderMoiseev: good one, i think i shouldn't change the value of checkbox vertical offset if node doesn't contain custom icon. I'll fix it so the new value for checkbox vertical offset we discussed with Pavel only applied when custom icon is used

IlyaKhD (dismissed):

scss/widgets/generic/treeView/_sizes.scss
```diff
@@ -22,11 +22,15 @@ $generic-treeview-border-padding: 1px !default;
 $generic-treeview-loadincicator-offset-top: 8px !default;
 $generic-treeview-loadincicator-offset-left: -1px !default;
 
+$generic-treeview-item-side-padding: 6px !default;
+
+$generic-treeview-custom-expander-icon-offset: null !default;
```
EugeniyKiyashko, line 27: Variables with null value are above. Seems it helps us understand which variables will redeclared with new value in each of sizing mode. May be we move it there?

EugeniyKiyashko (approved):

Krijovnick (approved):

ksercs (approved):

PR #23926, opened 2023-02-24
Overlay: support capability to scroll native overflow containers

EugeniyKiyashko:
Our current implementation prevents mouse event on overlay wrapper. It cause a lot of issues with processing these events. It’s complecated task to skip events that should be prevented on elements or wrapper or not.  

As the result there are troubles with scrolling of containers with overlow style, there are frezes after zooming on mobile devices. There are a lot of use cases we supported earlier in handler, but the issues arise time to time.  

We have developed a requirements that we would to support or suggest as workaround 

- Allow customers capability to use native scrollable containers(overflow: auto) in Overlay content. 
- Get rid of and side effects after zoom and follow scroll with using Popup and Popover. ([T968188](https://supportcenter.devexpress.com/internal/ticket/details/T968188)). Mouse and touch events shouldn’t be prevented. 
- Allow customers capability to fight body scrolling. Use Overlay component as Modal window. 
- Give a solution to prevent pass scrolling from popup scrolling to parents when container scrolled to the bottom or to the top positions. (overscroll-behavior)
 
We implement the following options for better appearance.
- **preventScrollEvents**: boolean options that allows to cancel preventing of mouse events.  

Default value: true 

If `PreventScrollEvents` is true the renderScrollTerminator works and we prevent mouse events on overlay wrapper. 

If `PreventScrollEvents` is false we make content scrollable by adding `dx-popup-content-scrollable` class with styles: overflow-y: auto; overscroll-behavior: contain; 

- **enableBodyScroll**: boolean option that allows to hide body scrollbar and prevent page scrolling (on the body) when modal is open.  

Default value: true


js/ui/overlay/ui.overlay.js
```diff
@@ -396,6 +401,16 @@ const Overlay = Widget.inherit({
...
     },
 
+    _toggleBodyScroll: function(enabled) {
+        if(!hasWindow() || this.option('enableBodyScroll')) {
+            return;
+        }
+
+        const { setOverflow, restoreOverflow } = this._bodyOverflowManager ??= createBodyOverflowManager();
+
+        enabled ? restoreOverflow() : setOverflow();
```
marker-dao: Let's use if operator here.

js/ui/overlay/utils.js
```diff
@@ -24,3 +26,99 @@ export const getElementMaxHeightByWindow = ($element, startLocation) => {
...
+        setOverflow() {
+            if(isIosDevice) {
+                setBodyPositionFixed();
+            } else {
+                setBodyPaddingRight();
+                setBodyOverflow();
+            }
+        },
+        restoreOverflow() {
+            if(isIosDevice) {
```
Krijovnick: On each call of the `setOverflow` and `restoreOverflow` methods checked if this device ios or not. I think it is better on cretion check this and return what need:
```
if(isIosDevice) {
 setOverflow = setBodyPositionFixed;
 restoreOverflow = restoreBodyPositionFixed;
} else {
 setOverflow = () => { 
    setBodyPaddingRight(); 
    setBodyOverflow();
}
 restoreOverflow = () => { 
    restoreBodyPaddingRight();
    restoreBodyOverflow();
 }
}

return { setOverflow, restoreOverflow }
```

js/ui/overlay/ui.overlay.js
```diff
@@ -396,6 +401,16 @@ const Overlay = Widget.inherit({
         return this._getOptionValue('animation', this);
     },
 
+    _toggleBodyScroll: function(enabled) {
+        if(!hasWindow() || this.option('enableBodyScroll')) {
+            return;
+        }
+
+        const { setOverflow, restoreOverflow } = this._bodyOverflowManager ??= createBodyOverflowManager();
```
Krijovnick: Can we create `_bodyOverflowManager` on widget creation or on init for example? It seems unnecessary to make `??= createBodyOverflowManager()` there, because `_bodyOverflowManager` never changed the value.

js/ui/overlay/ui.overlay.js
EugeniyKiyashko: [fix notes after review](https://github.com/DevExpress/DevExtreme/pull/23926/commits/a1490d482619b19a28d18a62bb70968364aa1c68)

js/ui/overlay/utils.js
EugeniyKiyashko: Good remark. I rewrote this part and now we doesn't check `isIosDevice` every time.

js/ui/overlay/ui.overlay.js
EugeniyKiyashko: Yes, we can. I wanted to hold all logic in one method. If it raises a questions it better to do it as you ask. I don't mind. Changes are here - https://github.com/DevExpress/DevExtreme/pull/23926/commits/a1490d482619b19a28d18a62bb70968364aa1c68

marker-dao (dismissed):

Krijovnick (dismissed):

js/ui/overlay.d.ts
```diff
@@ -83,6 +83,18 @@ export interface dxOverlayOptions<TComponent> extends WidgetOptions<TComponent>
...
+     * @default true
+     * @public
+     */
+    preventScrollEvents?: boolean;
+    /**
+     * @docid
+     * @default true
+     * @public
+     */
+    enableBodyScroll?: boolean;
```
ksercs: Are u sure we should publish it on overlay level?
I think it's better to keep it private in overlays, but publish it in Popup with a default = false

ksercs: Don't we wanna deprecate it to warn the users that we don't wanna keep old behavior in the future?

js/ui/popover.d.ts
```diff
@@ -115,6 +115,12 @@ export interface dxPopoverOptions<TComponent> extends dxPopupOptions<TComponent>
      * @public
      */
     hideOnParentScroll?: boolean;
+    /**
+     * @docid
+     * @default false
+     * @public
+     */
+    preventScrollEvents?: boolean;
```
ksercs: Isn't it enough to declare it on a popup level?

js/ui/popup/ui.popup.js
```diff
@@ -608,6 +611,10 @@ const Popup = Overlay.inherit({
         }
     },
 
+    _toggleContentScrollClass() {
+        this.$content().toggleClass(POPUP_CONTENT_SCROLLABLE_CLASS, !this.option('preventScrollEvents'));
```
ksercs:
```suggestion
        const isNativeScrollingEnabled = !this.option('preventScrollEvents');
        this.$content().toggleClass(POPUP_CONTENT_SCROLLABLE_CLASS, isNativeScrollingEnabled);
```

testing/testcafe/tests/editors/overlays/scrolling.ts
```diff
@@ -0,0 +1,145 @@
...
+              $content.append(`<div>${i}</div>`);
+            }
+
+            $('#scrollable-content').append($content);
+          })();
+
+          await appendElementTo('#inner-container', 'div', 'inner-content', { width: '2000px', height: '2000px' });
+          await appendElementTo('#scrollable-container', 'div', 'overlay', {});
+
+          await createWidget('dxOverlay', {
```
ksercs: I don't think we should test a UI of private Overlay component.
Moreover, `fullScreen` property is useless for it - it's implemented on a Popup level.
Let's use a Popup for a screenshot test?

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -3843,6 +3913,60 @@ testModule('renderGeometry', {
...
+        assert.strictEqual(this.$body.get(0).style.top, '-200px');
+        assert.strictEqual(this.$body.get(0).style.left, '0px');
+
+        overlay.hide();
+
+        assert.strictEqual(this.$body.get(0).style.position, '');
+        assert.strictEqual(this.$body.get(0).style.top, '');
+        assert.strictEqual(this.$body.get(0).style.left, '');
+    });
+});
```
ksercs: don't we wanna test a runtime change?

js/ui/overlay/ui.overlay.js
```diff
@@ -155,6 +156,10 @@ const Overlay = Widget.inherit({
             // NOTE: private options
             hideTopOverlayHandler: () => { this.hide(); },
             hideOnParentScroll: false,
+
+            preventScrollEvents: true,
```
ksercs, line 160: Don't forget to disable it for a scheduler appointment form

testing/testcafe/tests/editors/overlays/scrolling.ts
```diff
@@ -0,0 +1,145 @@
...
+              .eql('auto');
+          };
+
+          await t
+            .scroll(body, 300, 0);
+
+          await checkBodyStyles({ paddingRight: '0px', overflow: 'visible' });
+          await t
+            .expect(getDocumentScrollTop())
+            .eql(258);
```
ksercs: This magic number is strange to me. We scroll on 300, but get a scrollTop 258. I'm afraid it can make this test unstable in the future.
Maybe it's better to use `dispatchEvent` or some other way to get rid of this?

testing/testcafe/model/overlay/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  hide(): Promise<{ top: number; left: number }> {
+    const { getInstance } = this;
+
+    return ClientFunction(
+      () => (getInstance() as any).hide(),
+      { dependencies: { getInstance } },
+    )();
+  }
+}
```
ksercs, line 68: Seems like this model is not useful after we change test to use popup.
Shouldn't we extend a popup model from this one?

js/ui/overlay.d.ts
EugeniyKiyashko: I made it public for Popup. Popover enherit this doc tag and will public too. In Tooltip I made it hidden. Also we don't change default behavior and this options should be `true`

js/ui/popover.d.ts
EugeniyKiyashko: We decided that we will not publish this option. It option should be private and will used for capability return old behavior if it will require. The warning will occur if customers uses it only.

js/ui/overlay.d.ts
EugeniyKiyashko: No, we don't want to mark this option as deprecated in the future. It's a feature to provide capability to prevent body scrolling if customers are going to use Popup as Modal.

js/ui/overlay/ui.overlay.js
EugeniyKiyashko: I have disabled it already.

testing/testcafe/model/overlay/index.ts
EugeniyKiyashko: I extended a popup model and also use this in contextMenu for avoid duplication.

testing/testcafe/tests/editors/overlays/scrolling.ts
EugeniyKiyashko: Yes, we may. I used another way for it

testing/tests/DevExpress.ui.widgets/overlay.tests.js
EugeniyKiyashko: Yes, I added these tests in the next commit

testing/testcafe/tests/editors/overlays/scrolling.ts
EugeniyKiyashko: Yes, great idea.

js/ui/popup/ui.popup.js
EugeniyKiyashko: Fixed

js/ui/overlay.d.ts
ksercs: Yeap, sorry, misschose a place - it was about `preventScrollEvents`. Even if we don't publish it, i think it's a good idea to raise a warning in console

andrewmakarov (dismissed):

js/ui/overlay/ui.overlay.js
```diff
@@ -217,9 +221,19 @@ const Overlay = Widget.inherit({
...
                 this._logDeprecatedOptionWarning('elementAttr', createWrapperAttrDeprecationInfo());
             }
+            if('preventScrollEvents' in options) {
+                this._logDeprecatedPreventScrollEventsInfo();
+            }
         }
     },
 
+    _logDeprecatedPreventScrollEventsInfo() {
+        this._logDeprecatedOptionWarning('preventScrollEvents', (() => ({
```
ksercs: hm, isn't it the same?
```javascript
        this._logDeprecatedOptionWarning('preventScrollEvents', {
            since: '23.1',
            message: 'If you enable this option, end-users may experience scrolling issues.'
        });
```

js/ui/popup/ui.popup.js
```diff
@@ -280,6 +271,12 @@ const Popup = Overlay.inherit({
         this.callBase();
     },
 
+    _createBodyOverflowManager: function() {
+        if(hasWindow()) {
+            this._bodyOverflowManager = createBodyOverflowManager();
```
ksercs: We can use [null object pattern](https://refactoring.guru/introduce-null-object) to get rid of conditions like this here and other places using `_bodyOverflowManager`.
You can check how it's used in [ResizeObserver implementation](https://github.com/DevExpress/DevExtreme/blob/23_1/js/core/resize_observer.js#L14).

js/ui/tooltip/tooltip.js
```diff
@@ -24,6 +24,12 @@ const Tooltip = Popover.inherit({
             */
             showCloseButton: false,
 
+            /**
+            * @name dxTooltipOptions.enableBodyScroll
+            * @hidden
+            */
+            enableBodyScroll: true,
```
ksercs, line 31: shouldn't it be `false` by default to keep an old behavior in Sheduler appointment, slider tooltip and so on?

testing/tests/DevExpress.performance/overlaysStyleRecalculations.tests.js
```diff
@@ -77,4 +77,52 @@ QUnit.performanceTest('dxOverlay should not force relayout on creation', functio
 
         assert.measureStyleRecalculation(measureFunction, shading ? 18 : 17);
     });
+
+    [true, false].forEach(enableBodyScroll => {
+        QUnit.performanceTest(`showing dxPopup with shading=${shading} & enableBodyScroll=${enableBodyScroll} should be shown with minimum count of relayouts`, function(assert) {
```
ksercs: Update in all new tests plz
```suggestion
        QUnit.performanceTest(`Popup with shading=${shading} & enableBodyScroll=${enableBodyScroll} should be shown with minimum count of relayouts`, function(assert) {
```

testing/tests/DevExpress.performance/overlaysStyleRecalculations.tests.js
```diff
@@ -77,4 +77,52 @@ QUnit.performanceTest('dxOverlay should not force relayout on creation', functio
...
+
```
ksercs:
```suggestion
```

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -2198,6 +2375,20 @@ QUnit.module('rendering', {
 
         assert.ok($('.' + POPUP_BOTTOM_CLASS).dxToolbar('instance').option('compactMode'), 'bottom toolbar has the compact option');
     });
+
+    QUnit.test('dx-popup-content-scrollable class should be attached when preventScrollEvents is used', function(assert) {
+        this.instance.show();
+
+        const $popupContent = this.instance.$content();
+
+        assert.strictEqual($popupContent.hasClass('dx-popup-content-scrollable'), true, 'scrollable class is attached');
```
ksercs: Shouldn't we move `dx-popup-content-scrollable` to a constant?

js/ui/scheduler/appointmentPopup/popup.js
```diff
@@ -25,6 +25,8 @@ const POPUP_CONFIG = {
     maxHeight: '100%',
     showCloseButton: false,
     showTitle: false,
+    preventScrollEvents: false,
```
ksercs, line 28: Don't we wanna add a QUnit test to fix the scenario, save a ticket ID and so on?

js/ui/overlay/ui.overlay.js
EugeniyKiyashko: Seems like the same, I overreacted

EugeniyKiyashko: I'll fix it

js/ui/tooltip/tooltip.js
EugeniyKiyashko: In Sheduler appointment we set it as false, because it's justified there. But in tooltip slider and others there are a lot of use cases where it may be redundant.

testing/tests/DevExpress.performance/overlaysStyleRecalculations.tests.js
EugeniyKiyashko: I will

testing/tests/DevExpress.ui.widgets/popup.tests.js
EugeniyKiyashko: Thanks, great idea, I'll do it

js/ui/scheduler/appointmentPopup/popup.js
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/23926/commits/fd2b6b773237f86f0fd2b04d5786a12da38224c0

testing/tests/DevExpress.ui.widgets/popup.tests.js
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/23926/commits/fd2b6b773237f86f0fd2b04d5786a12da38224c0

testing/tests/DevExpress.performance/overlaysStyleRecalculations.tests.js
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/23926/commits/fd2b6b773237f86f0fd2b04d5786a12da38224c0

testing/tests/DevExpress.performance/overlaysStyleRecalculations.tests.js
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/23926/commits/fd2b6b773237f86f0fd2b04d5786a12da38224c0

js/ui/popup/ui.popup.js
EugeniyKiyashko: [v](https://github.com/DevExpress/DevExtreme/pull/23926/commits/fd2b6b773237f86f0fd2b04d5786a12da38224c0)

ksercs (approved):

IlyaKhD (approved):

vladaskorohodova (approved):

PR #23945, opened 2023-02-28
Tabs: Fix scroll to disabled item when it has focus

marker-dao:


EugeniyKiyashko (approved):

Zedwag (approved):

testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -656,6 +658,30 @@ QUnit.module('Horizontal scrolling', () => {
         assert.ok(itemOffset <= rightBoundary, `item offset ${itemOffset} is lower than right boundary ${rightBoundary}`);
         assert.ok(itemOffset > contentLeft, `item offset ${itemOffset} is greater than left boundary ${contentLeft}`);
     });
+
+    QUnit.test('tabs should scroll to the disabled item when it have focus', function(assert) {
+        const items = [{ text: 'item 1' }, { text: 'item 2' }, { text: 'item 3', disabled: true }];
+        const $element = $('#scrollableTabs').dxTabs({
+            items: items,
```
ksercs:
```suggestion
            items,
```

testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -656,6 +658,30 @@ QUnit.module('Horizontal scrolling', () => {
...
+            scrollingEnabled: true,
+            visible: true,
+            width: 200,
+            showNavButtons: false,
+            focusStateEnabled: true,
```
ksercs, line 668: Testing config should be minimum config available for the case reproduction.
`visible=true` by default, `scrollingEnabled=true` by default.
Are other props meaningful here?

Let's simplify this

testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -656,6 +658,30 @@ QUnit.module('Horizontal scrolling', () => {
...
+        });
+        const $item = $element.find('.' + DISABLED_STATE_CLASS).eq(0);
```
ksercs: we usually use string literals as a more modern and short approach
```suggestion
        const $item = $element.find(`.${DISABLED_STATE_CLASS}`).eq(0);
```

testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -656,6 +658,30 @@ QUnit.module('Horizontal scrolling', () => {
...
+
+        keyboard.press('right');
+        keyboard.press('right');
+
+        const contentLeft = Math.round($element.offset().left);
+        const contentRight = Math.round(contentLeft + $element.outerWidth());
+        const itemLeft = Math.round($item.offset().left);
+        const itemRight = Math.round(itemLeft + $item.outerWidth());
+
+        assert.strictEqual(itemRight - contentRight, 0, 'focused item in disabled state in view');
```
ksercs: It's not necessary to copypaste `Math.round` for every value. We can just use a `roughEqual` instead of a `strictEqual`
```suggestion
        assert.roughEqual(itemRight, contentRight, 1, 'focused disabled item is in view');
```

js/ui/tabs.js
```diff
@@ -425,6 +425,12 @@ const Tabs = CollectionWidget.inherit({
         this.callBase();
     },
 
+    _updateScrolling($target) {
```
ksercs: This name is too common for me. The most suitable name is `_scrollToFocusedDisabledItem`.
Now we can see that this method is really specific. Do we really need?
It seems to me that not only disabled items should be visible on focus => i didn't catch y we need this condition. Maybe we can just always scroll to item in `optionChanged` handler?

testing/tests/DevExpress.ui.widgets/tabs.tests.js
marker-dao: Thank you for your attention! I fixed it here [87ee87bc1c61b02d65c6472f29438a5a1611e46f].

testing/tests/DevExpress.ui.widgets/tabs.tests.js
marker-dao: Thank you for your attention! I fixed it here [87ee87bc1c61b02d65c6472f29438a5a1611e46f].

testing/tests/DevExpress.ui.widgets/tabs.tests.js
marker-dao: Yes, of course, sorry for the copy-paste. It was fixed in this commit [87ee87bc1c61b02d65c6472f29438a5a1611e46f].

testing/tests/DevExpress.ui.widgets/tabs.tests.js
marker-dao: Thanks a lot, this method really looks better. I fixed it here [87ee87bc1c61b02d65c6472f29438a5a1611e46f].

js/ui/tabs.js
marker-dao: Yes, it's a good name. I fixed it in this commit [402689d46ecda102c8bc0f40b9fcc6495520eb83].

Yes, you are right: not only disabled items should be visible on focus. Another items become visible too, because the handler is called for them when the selectedItem option is changed in the collection widget. It is difficult logic, we don't have to fix this. I wrote about this in my post in our channel)

ksercs: Let's discuss it in voice

ksercs: I suggest u just always call `scrollToItem` in option changed

marker-dao: Yes, was fixed in [c59a0376565c8f3ab5d954d90c8f08500b96ac08]. Thanks!

EugeniyKiyashko (approved):

ksercs (approved):

PR #23958, opened 2023-03-01
Fix exception while dragging a row in grid (T1150251)

GoodDayForSurf:
Fix exception while dragging a row in grid  if angular is used in a page (T1150251)

### DIFF
--- a/js/core/utils/type.js
+++ b/js/core/utils/type.js
@@ -2,15 +2,15 @@ const types = {
     '[object Array]': 'array',
     '[object Date]': 'date',
     '[object Object]': 'object',
-    '[object String]': 'string',
-    '[object Null]': 'null' };
+    '[object String]': 'string'
+};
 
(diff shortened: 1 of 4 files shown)


js/core/utils/type.js
```diff
@@ -8,6 +8,10 @@ const types = {
 const type = function(object) {
     const typeOfObject = Object.prototype.toString.call(object);
 
+    if(object === null) {
```
MikeVitik, line 9: Remove types["Object null"]
Use type in isPlainObject instead of direct Object.prototype.toString.call

MikeVitik (changes requested):

js/core/utils/type.js
GoodDayForSurf: ok

artem-kurchenko (approved):

MikeVitik (approved):

PR #23963, opened 2023-03-02
DataGrid performs requests to the server when a user searches data in a lookup column Filter Row editor if remoteOperations|groupPaging is enabled (T1149139)

pomahtri:


js/ui/grid_core/filtering/filter_custom_operations.js
```diff
@@ -1,10 +1,10 @@
-import $ from '../../core/renderer';
```
williamvinogradov: Let's name not refactored files with prefix ```module```, e.g ```module_filter_custom_operations```
Because linter has different configs for refactored files and old code :)

js/ui/grid_core/filtering/utils.js
```diff
@@ -0,0 +1,152 @@
...
+    return group.map((item, i) => {
+        if(isString(item)) {
+            return {
+                selector: item,
+                isExpanded: i < group.length - 1,
+            };
+        }
+
+        return item;
+    });
```
williamvinogradov: Minor comment.

Let's write it shortly:
```suggestion
        return group.map((item, idx) =>
        isString(item)
            ? ({ selector: item, isExpanded: idx < group.length - 1 })
            : item);
```

js/ui/grid_core/filtering/utils.js
```diff
@@ -0,0 +1,152 @@
...
+        lookupDataSourceOptions = lookup.items;
+    } else {
+        lookupDataSourceOptions = lookup.dataSource;
+        if(isFunction(lookupDataSourceOptions) && !variableWrapper.isWrapped(lookupDataSourceOptions)) {
+            lookupDataSourceOptions = lookupDataSourceOptions({});
+        }
+    }
+
+    return normalizeDataSourceOptions(lookupDataSourceOptions);
+}
```
williamvinogradov: Let's reduce the complexity of this function:
```suggestion
export function normalizeLookupDataSource(lookup) {
    if(lookup.items) {
        return normalizeDataSourceOptions(lookup.items);
    } else if(
        isFunction(lookup.dataSource) 
        && !variableWrapper.isWrapped(lookup.dataSource)
    ) {
        return normalizeDataSourceOptions(lookup.dataSource({}));
    }

    return normalizeDataSourceOptions(lookup.dataSource);
}
```

williamvinogradov (approved):

PR #23994, opened 2023-03-10
Generate modules for types reexports

const314:
**Practical value**
To avoid name conflicts when reexporting types for components, we need to generate separate files with types that should be reexported. Also, the CI workflow controls that these generated files are up-to-date.

**Before**
No reexports in separate files

**After**
Separate component_name_types.d.ts with public exports and a workflow that checks if they are up-to-date.


package.json
```diff
@@ -233,6 +233,7 @@
     "dev:watch": "cross-env DEVEXTREME_TEST_CI=true gulp dev-watch",
     "transpile-tests": "gulp transpile-tests",
     "discover-declarations": "dx-tools discover --js-scripts ./js --exclude js/renovation/ --compiler-options \"{ \\\"typeRoots\\\": [] }\"",
+    "generate-component-reexports": "npm run internal-tool -- generate-reexports --js-scripts ./js --exclude '((dialog|export|list_light|notify|overlay|pager|palette|set_template_engine|splitter|themes|themes_callback|track_bar|utils|validation_engine|validation_message)[.d.ts])|(js/renovation/)' --compiler-options \"{ \\\"typeRoots\\\": [] }\"",
```
IlyaKhD: Let's use same naming pattern for TS generation scripts, like this:

```json
{
  "update-ts-bundle": "...",
  "update-ts-reexports": "..."
  "update-ts": "npm run update-ts-bundle && npm run update-ts-reexports"
}
```

const314: fixed

IlyaKhD (approved):

PR #24000, opened 2023-03-13
Overlay: fix closing after click on item in shadow dom (T1146455)

Krijovnick:


marker-dao (dismissed):

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -855,6 +855,30 @@ testModule('visibility', moduleConfig, () => {
         }
     });
 
+    if(!QUnit.isInShadowDomMode()) {
+        test('overlay should not close after click on item in shadow dom (T1146455)', function(assert) {
```
ksercs: Let's specify that we're talking only about element inside an overlays's content. Overlay should still close after click on element in shadow dom if this element is not inside a content.
```suggestion
        test('overlay should not close after click on content element in shadow dom (T1146455)', function(assert) {
```

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -855,6 +855,30 @@ testModule('visibility', moduleConfig, () => {
...
+            const $template = $('<div id=\'text\'></div>')
+
+            const overlay = $('#overlay').dxOverlay({
+                hideOnOutsideClick: true,
+                contentTemplate: new Template($template),
```
ksercs: `new Template()` is a kind of a private api. We try to not use a private api to simplify a future refactoring.
Other points:
- `Template` api using makes the test more difficult. We all know that `contentTemplate` accepts a function returning a node, but not everybody knows an integration api
- `Template` api makes the test more synthetic. User cannot pass `Template` instance directly. It's more sensitive to test a real scenario.

I suggest u refactor the following way, we can also simplify getters.
```javascript
    const content = $('<div>').get(0);
    const overlay = $('#overlay').dxOverlay({
        // ...
        contentTemplate: () => content
    }).dxOverlay('instance');
    
    content.attachShadow({ mode: 'open' });
    content.shadowRoot.innerHTML = '<p>Inner Text</p>';
    
    const shadowTextElement = content.shadowRoot.querySelector('p');
```

testing/tests/DevExpress.ui.widgets/overlay.tests.js
```diff
@@ -855,6 +855,30 @@ testModule('visibility', moduleConfig, () => {
         }
     });
 
+    if(!QUnit.isInShadowDomMode()) {
```
ksercs: Why don't we want run this test in shadowDom? Seems it works fine

js/ui/overlay/ui.overlay.js
```diff
@@ -303,15 +303,19 @@ const Overlay = Widget.inherit({
         this._initPositionController();
     },
 
+    _isTargetInContainer: function(container, target) {
+        return contains(container, target) || contains(container, target.getRootNode().host);
```
ksercs: Hm, this code is not related to an overlay specific => it should be placed in some common core place.
Let's try to update initial `contains` method code?

marker-dao (dismissed):

js/core/utils/dom.js
```diff
@@ -93,7 +93,11 @@ export const contains = (container, element) => {
     if(!element) {
         return false;
     }
+    const shadowHost = element.getRootNode && element.getRootNode().host;
```
ksercs: If we change this code, we should write a new test in `utils.dom.tests.js`

js/core/utils/dom.js
```diff
@@ -93,7 +93,11 @@ export const contains = (container, element) => {
...
+    return isInContainer(container, element) || shadowHost && isInContainer(container, shadowHost);
+};
 
+const isInContainer = (container, element) => {
```
ksercs: Let's move this private function to the file's beginning

js/core/utils/dom.js
```diff
@@ -93,7 +93,11 @@ export const contains = (container, element) => {
...
     if(domAdapter.isTextNode(element)) {
```
ksercs: If we think that all this conditions are necessary for the case when target is a shadow host, it's a good idea to add tests for this cases

js/core/utils/dom.js
```diff
@@ -93,22 +93,10 @@ export const contains = (container, element) => {
...
-        return container.documentElement.contains(element);
-    }
-
     if(isWindow(container)) {
         return contains(container.document, element);
     }
-
-    return container.contains
-        ? container.contains(element)
-        : !!(element.compareDocumentPosition(container) & element.DOCUMENT_POSITION_CONTAINS);
```
ksercs, line 111: Why did u remove this code?

testing/tests/DevExpress.core/utils.dom.tests.js
```diff
@@ -118,3 +118,39 @@ QUnit.test('it correctly works with svg elements', function(assert) {
...
+    const $div = $('<div>').appendTo('#qunit-fixture');
+    const divContent = $div.get(0);
+
+    divContent.attachShadow({ mode: 'open' });
+    divContent.shadowRoot.innerHTML = '<p>Inner Text</p>';
+
+    const textElement = divContent.shadowRoot.querySelector('p');
+
+    assert.ok(domUtils.contains(document, textElement));
+});
```
ksercs: empty line

testing/tests/DevExpress.core/utils.dom.tests.js
```diff
@@ -118,3 +118,39 @@ QUnit.test('it correctly works with svg elements', function(assert) {
 
     assert.ok(domUtils.contains(svgContainer, childElement));
 });
+
+QUnit.test('element in shadow dom should detects in window element', function(assert) {
```
ksercs:
```suggestion
QUnit.test('element in shadow dom should be detected if container is window', function(assert) {
```

js/core/utils/dom.js
Krijovnick: Because it was made for IE (https://github.com/DevExpress/DevExtreme/pull/13136). Now we does not support IE

ksercs: Nice!

ksercs (approved):

alexanderPolosatov (approved):

marker-dao (approved):

EugeniyKiyashko (approved):

PR #24001, opened 2023-03-13
Toolbar: _dimensionChanged should not raise any error if widget is disposed (T1147410)

marker-dao:


testing/tests/DevExpress.ui.widgets/toolbar.tests.js
```diff
@@ -1724,6 +1724,27 @@ QUnit.module('adaptivity without hiding in menu', {
     });
 });
 
+QUnit.module('Toolbar disposing', () => {
+    QUnit.test('_dimensionChanged skipped if widget is disposed (T1147410)', function(assert) {
```
ksercs: Oh, it's a really specific case, it's difficult to write a "correct" test here. Now I see the following issues:
1. Actually this test doesn't check that some method is skipped. It checks that there is no error
2. Using of `width` option change as an action seems to be confusing for me. If we refactor width option change handler to not use a `dimensionChanged`, this test will become useless.
3. I believe that we should not emulate a material theme. `dimensionChanged` issues is actual for both themes (although a promise issue is related only to material theme).

So i suggest u to write a usual "unit" test with a private api using. "'_dimensionChanged call should not raise any error if toolbar is disposed(T1147410)'". Private api using is not a good practice because it makes a future refactoring more difficult, but it's more effective and cheap to use it in this concrete scenario - we test exactly what we want and make it easy.

testing/tests/DevExpress.ui.widgets/toolbar.tests.js
```diff
@@ -1724,6 +1724,27 @@ QUnit.module('adaptivity without hiding in menu', {
...
+
+        try {
+            toolbar.option('width', 250);
+        } catch (e) {
+            assert.ok(false, e);
+        } finally {
+            assert.ok(true, 'the exception is not thrown');
+        }
+
+        themes.isMaterial = origIsMaterial;
```
ksercs: jfyi: if an error occurs this code will not be run. We should always use `try/finally` for such restoring or move it to a module configuration. However, let's refactor the test to not use a material theme at all.

testing/tests/DevExpress.ui.widgets/toolbar.tests.js
marker-dao: Thank you for review. Test was updated here: [fd3c3f1072309a7a07a8984b43e4d1efc3b23c58].

Zedwag (approved):

Krijovnick (approved):

EugeniyKiyashko (approved):

ksercs (approved):

PR #24035, opened 2023-03-16
PivotGrid: Fix broken export issue

williamvinogradov:
Change export method calls to existing methods + fix ```exportTo``` PivotGrid method

### DIFF
--- a/testing/testcafe/tests/pivotGrid/export/onExportingOption.ts
+++ b/testing/testcafe/tests/pivotGrid/export/onExportingOption.ts
@@ -1,0 +1,34 @@
+import { ClientFunction } from 'testcafe';
+import createWidget from '../../../helpers/createWidget';
+import url from '../../../helpers/getPageUrl';
+import PivotGrid from '../../../model/pivotGrid';
+
+fixture.disablePageReloads`PivotGrid_fieldChooser`
+  .page(url(__dirname, '../../container.html'));
+
+test('Should call \'onExporting\' when export button clicked', async (t) => {
+  const pivotGrid = new PivotGrid('#container');
+  const exportBtn = pivotGrid.getExportButton();
+
+  await t.click(exportBtn);
+  // eslint-disable-next-line no-underscore-dangle
+  const exportCalled = await ClientFunction(() => (window as any).__exportCalled as boolean)();
+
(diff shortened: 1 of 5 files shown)


testing/testcafe/tests/pivotGrid/export/onExportingOption.ts
```diff
@@ -0,0 +1,33 @@
...
+    store: [],
+  },
+  export: {
+    enabled: true,
+  },
+  onExporting() {
+    // eslint-disable-next-line no-underscore-dangle
+    (window as any).__exportCalled = true;
+  },
+}));
```
EugeniyKiyashko: Let's clear __exportCalled property from window store

testing/testcafe/tests/pivotGrid/export/onExportingOption.ts
```diff
@@ -0,0 +1,33 @@
...
+    fields: [{
+      caption: 'data A',
+      dataField: 'data_A',
+    }],
+    store: [],
+  },
+  export: {
+    enabled: true,
+  },
+  onExporting() {
```
EugeniyKiyashko, line 29: Let's check incomming arguments

testing/testcafe/tests/pivotGrid/export/onExportingOption.ts
williamvinogradov: Agree, done

testing/testcafe/tests/pivotGrid/export/onExportingOption.ts
williamvinogradov: Let's keep it as is (check only that ```onExporting``` was called)

Because this method doesn't have any specific arguments for export (it only passed reference of the pivot grid widget to the export tools)
[Link to the docs](https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxPivotGrid/Configuration/#onExporting)

pomahtri: > Let's check incomming arguments

we don't need to check incomming arguments in tests, now we can check it via TS :)
we'll add types in next PR with refactoring

pomahtri (approved):

PR #24047, opened 2023-03-18
DataGrid - Header Filter Customization Feature

Tucchhaa:


pomahtri (changes requested):
You also need to overload `_setDeprecatedOptions` method to show warning in dev console

js/common/grids.d.ts
```diff
@@ -570,6 +570,39 @@ export type ColumnFixingTexts = {
...
+   * @type getter|Array<getter>
+   * @default undefined
+   */
+  searchExpr?: string | Function | Array<string | Function>;
+  /**
+   * @docid GridBaseColumn.headerFilter.search.timeout
+   * @default 500
+   */
+  timeout?: number;
+};
```
pomahtri: Maybe make this interface to be extended from `HeaderFilterSearch`?

js/ui/grid_core/ui.grid_core.header_filter_core.js
```diff
@@ -360,6 +357,28 @@ export const HeaderFilterView = modules.View.inherit({
...
+            search: {
+                mode: specificHeaderFilter.searchMode,
+                enabled: specificHeaderFilter.allowSearch,
+                timeout: specificHeaderFilter.searchTimeout
+            }
+        };
+
+        return extend(true, {}, generalHeaderFilter, generalDeprecated, specificHeaderFilter, columnDeprecated);
+    },
+
```
pomahtri: Our option manager already has logic to normalize deprecated options: https://github.com/DevExpress/DevExtreme/blob/bfb0effb8e9d5ffe280bf6d9b224017e366b0167/js/core/options/index.js#L103

Please check that this logic works as we need, and in this case remove `generalDeprecated` and `columnDeprecated`

js/ui/pivot_grid.d.ts
```diff
@@ -303,16 +304,48 @@ export interface dxPivotGridOptions extends WidgetOptions<dxPivotGrid> {
...
+         * @docid
+         * @default 'contains'
+         */
+        mode?: SearchMode;
+        /**
+         * @docid
+         * @default 500
+         */
+        timeout?: number;
+      };
```
pomahtri: Maybe better to just reuse interface from grids.d.ts

testing/tests/DevExpress.ui.widgets.dataGrid/headerFilter.tests.js
```diff
@@ -27,6 +27,18 @@ import messageLocalization from 'localization/message';
...
+        this.headerFilterController.showHeaderFilterMenu(columnIndex);
+    }
+
+    const $popupContent = this.headerFilterView.getPopupContainer().$content();
+    const list = $popupContent.find('.dx-list');
+    const treeView = $popupContent.find('.dx-treeview');
+
+    return list.length ? list.dxList('instance') : treeView.dxTreeView('instance');
+}
+
```
pomahtri: Minor, maybe it's better to leave as is: it's not obvious that function with name like 'get...' not only getting some value, but performs action (showing filter menu in this case)

js/ui/grid_core/ui.grid_core.header_filter_core.js
pomahtri: discussed it, it doesn't support array options like column[].headerFilter

js/common/grids.d.ts
Tucchhaa: I think docid's will be incorrect in this way

const314: I agree that this type should be derived from HeaderFilterSearch. And, since the types are new, we can leave the docid tags empty. Will we have some problems with documentation in this case?

js/ui/gantt.d.ts
```diff
@@ -1417,6 +1431,34 @@ export interface dxGanttHeaderFilter {
     width?: number;
 }
 
+/**
+ * @docid
+ * @type object
+ * @namespace DevExpress.ui
+ */
+export interface dxGanttHeaderFilterSearch {
```
const314: This type should be derived from HeaderFilterSearch as well

js/ui/pivot_grid.d.ts
const314: Agree, use HeaderFilterSearch here

ts/aliases.d.ts
```diff
@@ -289,6 +289,8 @@ declare module DevExpress.ui.dxDataGrid {
     export type ColumnFixingTexts = DevExpress.common.grids.ColumnFixingTexts;
     /** @deprecated Use DevExpress.common.grids.ColumnHeaderFilter instead */
     export type ColumnHeaderFilter = DevExpress.common.grids.ColumnHeaderFilter;
+    /** @deprecated Use DevExpress.common.grids.ColumnHeaderFilterSearch instead */
```
const314: These types are new, it seems that there shouldn't be entries for them here.

js/ui/pivot_grid/data_source.d.ts
```diff
@@ -382,7 +383,21 @@ export interface PivotGridDataSourceField {
...
      */
-    headerFilter?: { allowSearch?: boolean; height?: number; width?: number };
+    headerFilter?: {
+        /**
+         * @deprecated
+         */
+        allowSearch?: boolean;
+        allowSelectAll?: boolean;
+        height?: number;
+        search?: {
```
const314: Use the HeaderFilterSearch type here

js/ui/pivot_grid_field_chooser.d.ts
```diff
@@ -84,16 +84,48 @@ export interface dxPivotGridFieldChooserOptions extends WidgetOptions<dxPivotGri
...
+      allowSelectAll?: boolean;
       /**
        * @docid
        * @default 325
        */
       height?: number;
+      /**
+       * @docid
+       */
+      search?: {
```
const314: Use the HeaderFilterSearch type here

js/common/grids.d.ts
```diff
@@ -597,9 +636,15 @@ export type ColumnHeaderFilter = {
    * @default undefined
    */
   height?: number;
+  /**
+   * @docid GridBaseColumn.headerFilter.search
+   * @type object
```
const314: Why do we have @type object here?

js/ui/pivot_grid/data_source.d.ts
Tucchhaa: Can I use the HeaderFilter type for all options and not just HeaderFilterSearch?

js/common/grids.d.ts
Tucchhaa: I took as example HeaderFilter.texts option, which has @type object

js/common/grids.d.ts
Tucchhaa: I just tried to follow the document's formatting rules, but It is nice that I can left @docid empty.

Tucchhaa: image

that's how it looks with empty docids in dx.all.d.ts. Is it ok?

js/ui/pivot_grid/data_source.d.ts
const314: I think yes if it is a common behavior for grid and pivot

js/common/grids.d.ts
const314: It is not necessary here, let's remove it

const314 (dismissed):
I approve this one for now, since my comments will be addressed in a separate PR

js/ui/gantt/ui.gantt.js
```diff
@@ -22,7 +22,7 @@ import { GanttToolbar, GanttContextMenuBar } from './ui.gantt.bars';
 import { GanttTreeList } from './ui.gantt.treelist';
 import { GanttView } from './ui.gantt.view';
 import { GanttDataChangesProcessingHelper } from './ui.gantt.data_changes_processing_helper';
-
+import gridCoreUtils from '../grid_core/ui.grid_core.utils';
```
MikeVitik, line 25:
```suggestion
import {logColumnsDeprecatedWarningIfNeed} from '../grid_core/ui.grid_core.utils';
```

js/ui/gantt/ui.gantt.treelist.js
```diff
@@ -21,6 +21,7 @@ export class GanttTreeList {
 
     getTreeList() {
         const { keyExpr, parentIdExpr } = this._gantt.option(GANTT_TASKS);
+        dxTreeList.prototype.disableDeprecationWarnings = true;
```
MikeVitik: Add try/finally block

js/ui/grid_core/ui.grid_core.header_filter_core.js
```diff
@@ -360,6 +357,28 @@ export const HeaderFilterView = modules.View.inherit({
         }
     },
 
+    _normalizeHeaderFilterOptions(options) {
```
MikeVitik, line 361: Suggestion: Try to make this function pure and move it to the utils part.
 Use it in place in _getSearchExpr and _initializeListContainer instead of passing the result of it as parameter.

MikeVitik (dismissed):

pomahtri (dismissed):

js/ui/grid_core/ui.grid_core.header_filter_core.js
```diff
@@ -168,10 +159,14 @@ export const HeaderFilterView = modules.View.inherit({
...
-    _getSearchExpr: function(options) {
+    _getSearchExpr: function(options, headerFilterOptions) {
         const lookup = options.lookup;
         const useDefaultSearchExpr = options.useDefaultSearchExpr;
-        const headerFilterDataSource = options.headerFilter && options.headerFilter.dataSource;
+        const headerFilterDataSource = headerFilterOptions.dataSource;
+
+        if(headerFilterOptions.search.searchExpr) {
+            return headerFilterOptions.search.searchExpr;
+        }
```
williamvinogradov: Minor comment

Let's make it shorter:
```suggestion
        const headerFilterDataSource = headerFilterOptions.dataSource;
        const filterSearchExpr = headerFilterOptions.search.searchExpr;

        if(filterSearchExpr) {
            return filterSearchExpr;
        }
```

js/ui/grid_core/ui.grid_core.header_filter_core.js
```diff
@@ -197,12 +192,12 @@ export const HeaderFilterView = modules.View.inherit({
...
 
-    _initializePopupContainer: function(options) {
+    _initializePopupContainer: function(options, headerFilterOptions) {
         const that = this;
         const $element = that.element();
-        const headerFilterOptions = that.option('headerFilter');
-        const width = options.headerFilter && options.headerFilter.width || headerFilterOptions && headerFilterOptions.width;
-        const height = options.headerFilter && options.headerFilter.height || headerFilterOptions && headerFilterOptions.height;
+        const width = headerFilterOptions.width;
+        const height = headerFilterOptions.height;
```
williamvinogradov: Minor comment.

Let's use destructuring here:
```suggestion
        const { width, height } = headerFilterOptions;
```

js/ui/grid_core/ui.grid_core.header_filter_core.js
```diff
@@ -360,6 +357,28 @@ export const HeaderFilterView = modules.View.inherit({
...
+    _normalizeHeaderFilterOptions(options) {
+        const generalHeaderFilter = this.option('headerFilter') || {};
+        const specificHeaderFilter = options.headerFilter || {};
+
+        const generalDeprecated = {
+            search: {
+                enabled: generalHeaderFilter.allowSearch,
+                timeout: generalHeaderFilter.searchTimeout,
+            }
+        };
```
williamvinogradov, line 370: Is it expected behavior, that ```search.enabled``` and ```search.timeout``` will be always overridden by ```specificDeprecated``` values?

js/ui/grid_core/ui.grid_core.header_filter_core.js
```diff
@@ -360,6 +357,28 @@ export const HeaderFilterView = modules.View.inherit({
...
+
+        const specificDeprecated = {
+            search: {
+                enabled: specificHeaderFilter.allowSearch,
+                mode: specificHeaderFilter.searchMode,
+                timeout: specificHeaderFilter.searchTimeout
+            }
+        };
+
+        return extend(true, {}, generalHeaderFilter, generalDeprecated, specificHeaderFilter, specificDeprecated);
```
williamvinogradov, line 380: Question.

Can we use the spread shell copy operator instead of the ```extend``` here?

E.g
```suggestion
        return {
          ...generalHeaderFilter,
          ...generalDeprecated,
          ...specificHeaderFilter,
          ...specificDeprecated,
        };
```

js/ui/grid_core/ui.grid_core.utils.js
```diff
@@ -701,4 +702,32 @@ export default {
 
         return lookupDataSource;
     },
+
+    logColumnsDeprecatedWarningIfNeed(componentName, columnsOptions) {
```
williamvinogradov: We already have the method for logging deprecation warnings
But it is implemented in the ```core/component.js```
[Link to the implementation](https://github.com/DevExpress/DevExtreme/blob/23_1/js/core/component.js#L133)

Let's try to use it instead of adding our custom implementation
Maybe we can prepare options for ```_logDeprecatedOptionWarning``` in utils and call it from the component

js/ui/gantt.d.ts
```diff
@@ -1388,21 +1389,34 @@ export interface dxGanttHeaderFilter {
...
+    allowSelectAll?: boolean;
     /**
      * @docid
      * @default 315 &for(Material)
      * @default 325
      */
     height?: number;
+    /**
+     * @docid
+     * @type object
```
mpreyskurantov: Remove @type here, it breaks .net wrappers

js/ui/gantt.d.ts
```diff
@@ -1388,21 +1389,34 @@ export interface dxGanttHeaderFilter {
...
+    search?: dxGanttHeaderFilterSearch;
     /**
      * @docid
      * @default 500
+     * @deprecated
      */
     searchTimeout?: number;
     /**
      * @docid
+     * @type object
```
mpreyskurantov: Remove @type here, it breaks .net wrappers

js/ui/gantt.d.ts
mpreyskurantov: Do you need to make this type visible in TypeScript? For now, it is marked as internal + deprecated in dx.all.d.ts
Consider removing "dx" prefix in its name
/cc @vladaskorohodova

mpreyskurantov (changes requested):

js/ui/gantt/ui.gantt.js
Tucchhaa: actually, it can't be like that. ui.grid_core.utils.d.ts exports default a type Utils and not each function seperate. So we need to import the gridCoreUtils

js/ui/gantt.d.ts
Tucchhaa: Do we need to do it? I made it similar to dxGanttHeaderFilterTexts. If we need, then we should change dxGanttHeaderFilterSearch too

js/ui/grid_core/ui.grid_core.header_filter_core.js
Tucchhaa: Unfortunately, no. We need deepExtend here, because we have headerFilter.search option which is object. So if in `generalHeaderFilter.search = { enabled: true }` and `specificHeaderFilter.search = { editorOptions: { placeholder: 'some' } }`, then the result headerFilter will contain just `specificHeaderFilter.search` values

js/ui/grid_core/ui.grid_core.header_filter_core.js
Tucchhaa: Yes. Example. If `datagrid.headerFilter.search.enabled = true` and in `datagrid.columns[0].headerFilter.search.enabled = false` then in first column's headerFilter search input shouldn't appear

js/__internal/grids/grid_core/module_utils.ts
```diff
@@ -696,4 +696,44 @@ export default {
 
     return lookupDataSource;
   },
+
+  logHeaderFilterDeprecatedWarningIfNeed(component) {
```
MikeVitik, line 700: Minor: Extract logHeaderFilterDeprecatedWarningIfNeed to the separated file.

MikeVitik (dismissed):

js/ui/grid_core/ui.grid_core.header_filter_core.js
williamvinogradov: Okay, our ```extend``` method has a different behavior that native ```extend```
It's not override values with ```undefinded``` value :)

williamvinogradov (dismissed):

js/ui/gantt.d.ts
mpreyskurantov: We need to add new types names according to guidelines

mpreyskurantov (changes requested):

williamvinogradov (dismissed):

williamvinogradov (dismissed):

mpreyskurantov (dismissed):

IlyaKhD (dismissed):

sergepilipchuk (dismissed):

MikeVitik (dismissed):

js/__internal/grids/grid_core/module_utils.ts
Tucchhaa: I think it is redundant

pomahtri (dismissed):

pomahtri (dismissed):

mpreyskurantov (dismissed):

IlyaKhD (dismissed):

williamvinogradov (dismissed):

const314 (approved):

williamvinogradov (approved):

PR #24057, opened 2023-03-20
PivotGrid(T1150523): Remove redundant sort http request.

williamvinogradov:
See the [T1150523](https://supportcenter.devexpress.com/ticket/details/T1150523) ticket

Remove redundant DataSource reload on the ```sort``` action

### DIFF
--- a/js/__internal/grids/pivot_grid/field_chooser/const.ts
+++ b/js/__internal/grids/pivot_grid/field_chooser/const.ts
@@ -54,3 +54,8 @@ export const SORTABLE_CONST = {
     drag: 'drag',
   },
 };
+
+export enum SortOrder {
+  descending = 'desc',
+  ascending = 'asc',
+}
--- a/js/__internal/grids/pivot_grid/field_chooser/const.ts
+++ b/js/__internal/grids/pivot_grid/field_chooser/const.ts
@@ -2,7 +2,7 @@ export const ATTRIBUTES = {
   treeViewItem: 'tree-view-item',
   allowScrolling: 'allow-scrolling',
   itemGroup: 'item-group',
-};
+} as const;
 
 export const CLASSES = {
   area: {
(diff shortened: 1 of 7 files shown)


js/__internal/grids/pivot_grid/field_chooser/const.ts
```diff
@@ -54,3 +54,8 @@ export const SORTABLE_CONST = {
     drag: 'drag',
   },
 };
+
+export enum SortOrder {
+  descending = 'desc',
+  ascending = 'asc',
+}
```
pomahtri: To be more consistent with standards in public d.ts, let's use plain union `type SortOrder = 'desc' | 'asc'`. Or maybe it's better to import it from somewhere we already have it

js/__internal/grids/pivot_grid/field_chooser/module_base.ts
```diff
@@ -126,6 +128,7 @@ const FieldChooserBase = (Widget as any)
           this._refreshDataSource();
           break;
         case 'applyChangesMode':
+        case 'remoteSort':
```
pomahtri, line 132: Is it new private option? Then we need to declare it somewhere, or else we'll forget about it. Also maybe we need to add it to d.ts, so that we could use it in angular/react/vue (but we'll need to hide it from docs)

js/__internal/grids/pivot_grid/field_chooser/const.ts
williamvinogradov: Discussed and decided to replace enum with const

js/__internal/grids/pivot_grid/field_chooser/module_base.ts
williamvinogradov: Discussed and decided add ```NOTE``` comment for now

pomahtri (approved):

PR #24073, opened 2023-03-23
DataGrid: Fix an issue where disabled editors are validated (T1152491)

markallenramirez:


js/ui/grid_core/ui.grid_core.validating.js
```diff
@@ -440,8 +440,9 @@ const ValidatingController = modules.Controller.inherit((function() {
                     if(adapter) {
                         adapter.getValue = getValue;
                         adapter.validationRequestsCallbacks = [];
+                        const oldBypass = adapter.bypass.bind(adapter);
                         adapter.bypass = () => {
-                            return parameters.row.isNewRow && !this._isValidationInProgress && !editingController.isCellModified(parameters);
+                            return oldBypass() || parameters.row.isNewRow && !this._isValidationInProgress && !editingController.isCellModified(parameters);
                         };
                     }
```
williamvinogradov: Let's change this to:
```javascript
                    if(adapter) {
                        // let's rename the variable and don't use bind
                        // it's will be better to use call or apply in the function call
                        const originBypass = adapter.bypass;
                        
                       // let's make closure with 'parameters' more obvious
                       // via moving it to a separate function
                        const defaultAdapterBypass = () =>
                            parameters.row.isNewRow && !this._isValidationInProgress && !editingController.isCellModified(parameters);

                        adapter.getValue = getValue;
                        adapter.validationRequestsCallbacks = [];
                        adapter.bypass = () => {
                            return originBypass.call(adapter) || defaultAdapterBypass();
                        };
                    }
```

williamvinogradov (approved):

PR #24093, opened 2023-03-23
TextArea: maxHeight and minHeight options should work with `em` css units when autoResizeEnabled option is true(T1155588)

EugeniyKiyashko:


Zedwag (dismissed):

js/core/utils/size.js
```diff
@@ -135,6 +135,8 @@ export const parseHeight = function(value, container) {
         value = parseInt(value);
     } else if(value.indexOf('vh') > 0) {
         value = window.innerHeight / 100 * parseInt(value.replace('vh', ''));
+    } else if(element && value.indexOf('em') > 0) {
+        value = parseFloat(value.replace('em', '')) * parseFloat(window.getComputedStyle(element).fontSize);
```
ksercs, line 139: Are u sure it's ok for performance?

ksercs (approved):

js/core/utils/size.js
EugeniyKiyashko: Getting this object doesn't require that a full recalc is performed, only its getters will force it. To get the `font-size` property is more light operation if we will compare with calling of the `width` getter.

marker-dao, line 139: What about 'rem' units? The same bug: [[codepen](https://codepen.io/marker-dao/pen/BaOvWxR)].

EugeniyKiyashko: In context of this PR we solve certain issue with `em` relative unit. I don't have information about use cases 'rem' and others where it needs. Perhaps, we will add note with limitation for using of other relative CSS units.

marker-dao (approved):

PR #24113, opened 2023-03-27
Menu: add item's url option 

Krijovnick:


js/ui/context_menu/ui.menu_base.js
```diff
@@ -177,14 +177,30 @@ class MenuBase extends HierarchicalCollectionWidget {
         this._initActions();
     }
 
+    _isValidURL(url) {
+        return url.match(/^(http|https):\/\//);
```
EugeniyKiyashko, line 181: Don't forget to describe limitation that we can't use relative paths in link

js/ui/menu.d.ts
```diff
@@ -277,6 +277,16 @@ export interface dxMenuItem extends dxMenuBaseItem {
...
+    /**
+     * @docid
+     * @public
+     */
+    url?: string;
+    /**
+     * @docid
+     * @public
+     */
+    linkAttr?: { [key: string]: any };
```
EugeniyKiyashko, line 290: I think we should add @default doc tag and default is {}, this way is simular for declare elementAttr option default value. In this case you need to extend _getDefaultOption() method with the new linkAttr option: 
```
linkAttr: {}
```
And we can change `itemData.linkAttr || {}` to itemData.linkAttr.

testing/tests/DevExpress.ui.widgets/menuBase.tests.js
```diff
@@ -333,6 +333,50 @@ QUnit.module('Menu rendering', () => {
...
+        const menuBase = createMenu({
+            items: [{ text: 'Item text', url: '/some_url' }]
+        });
+        const content = menuBase.element.find(`.${DX_MENU_ITEM_TEXT_CLASS}`);
+
+        assert.ok(!content.children()[0]);
+        assert.strictEqual(content.text(), 'Item text');
+    });
+
+    QUnit.test('Provide extra attributes for link if linkAttr option set with url', function(assert) {
```
EugeniyKiyashko: There are no tests on changing these options in runtime via api.

js/ui/context_menu/ui.menu_base.js
```diff
@@ -177,14 +177,30 @@ class MenuBase extends HierarchicalCollectionWidget {
...
+
     _getTextContainer(itemData) {
         const itemText = itemData.text;
+        const url = itemData.url;
         const $itemContainer = $('<span>').addClass(DX_MENU_ITEM_CAPTION_CLASS);
         const itemContent = isPlainObject(itemData) ? itemText : String(itemData);
-
+        if(url && this._isValidURL(url) && itemText) {
+            const $urlContainer = $('<a>');
+            $urlContainer.attr('href', url);
```
EugeniyKiyashko: We can write it shorter 
```
$urlContainer
    .attr({ 'href': url,  ...itemData.linkAttr })
    .text(itemContent)
    .appendTo($itemContainer)

File: testing/tests/DevExpress.ui.widgets/menuBase.tests.js [hunk 5]
```diff
@@ -333,6 +333,50 @@ QUnit.module('Menu rendering', () => {
...
+        const menuBase = createMenu({
+            items: [{ text: 'Item text', url: 'http://some_url', linkAttr: { target: '_blank' } }]
```
EugeniyKiyashko, line 423: Shouldn't we process the cases when linkAttr is not a object? There is not check condition for this case in code. We can do nothing in this case, but without exception.

testing/tests/DevExpress.ui.widgets/menuBase.tests.js
EugeniyKiyashko: It can be unnecessary, but time to time our customers come with this scenarios and we should sure that option change in runtime has been applied correctly.

js/ui/menu.d.ts
```diff
@@ -277,6 +277,17 @@ export interface dxMenuItem extends dxMenuBaseItem {
      * @type Array<dxMenuItem>
      */
     items?: Array<Item>;
+    /**
+     * @docid
+     * @public
+     */
+    url?: string;
```
EugeniyKiyashko, line 284: It looks strange that we add functionality(code) in `inMenuBase` component and add doc comments for `menu` component. 
- If we decide that this functionality should be used in `Menu` component only, then code should be inside menu/ui.menu.js file and belong `Menu` component that extends `MenuBase`.
- If we decide that this functionality should used `Menu` and `ContextMenu` components, but we not publish these options for `ContextMenu` then we should declare comments for `dxMenuBaseItem` interface and mark them as hidden and make it public for `Menu` component only. @IlyaKhD can help us to do it right according to best practices)

Pls, think about it

Krijovnick: This functionality we support only in menu widget. But it is not work correct if it moved to ui.menu.js file. So it live in the menuBase. But why we need to declare comments for dxMenuBaseItem? If the functionality for menu only?

testing/tests/DevExpress.ui.widgets/menuBase.tests.js
Krijovnick: I consider it overkill to add tests for options that are in another option. The items option has already been tested, the ability to update fields inside this option has been tested. Also, I can't write a red test. But if it is necessary, I will add these tests.

js/ui/menu.d.ts
EugeniyKiyashko: `This functionality we support only in menu widget. But it is not work correct if it moved to ui.menu.js file.`

Yes, it doesn't work because next menu levels is `subMenu` component that extends `contextMenu` component. So, this functionality should be on `menuBase` level for using both `menu` & `contextMenu`

testing/tests/DevExpress.ui.widgets/menuBase.tests.js
```diff
@@ -333,6 +333,62 @@ QUnit.module('Menu rendering', () => {
...
+        try {
+            createMenu({
+                items: [{ text: 'Item text', url: 'http://some_url', linkAttr: 'string' }]
+            });
+        } catch(e) {
+            assert.ok(false, `error ${e.message}`);
+        } finally {
+            assert.ok(true, 'no error');
+        }
+    });
```
ksercs, line 406: How should this api work if `item.template` is specified?
Let's add a test for this and add info in docs

js/ui/context_menu/ui.menu_base.js
```diff
@@ -177,14 +177,30 @@ class MenuBase extends HierarchicalCollectionWidget {
...
+            const $urlContainer = $('<a>');
+            $urlContainer.attr({ 'href': url, ...linkAttr })
+                .text(itemContent)
+                .appendTo($itemContainer);
+            return $itemContainer;
+        }
         return itemText && $itemContainer.text(itemContent);
     }
 
+    _getExtraFields() {
```
ksercs: This name is too common to me.
```suggestion
    _getItemLinkPropNames() {
```

js/ui/hierarchical_collection/ui.hierarchical_collection_widget.js
```diff
@@ -85,7 +89,7 @@ const HierarchicalCollectionWidget = CollectionWidget.inherit({
                     .append(this._getTextContainer(itemData))
                     .append(this._getPopoutContainer(itemData));
                 that._addContentClasses(itemData, $container.parent());
-            }.bind(this), ['text', 'html', 'items', 'icon'], this.option('integrationOptions.watchMethod'), {
+            }.bind(this), fields, this.option('integrationOptions.watchMethod'), {
```
ksercs, line 92: Seems like this code is not tested, am i right?

js/ui/context_menu/ui.menu_base.js
Krijovnick: The name should be common, because it uses in collection widget and it can return any fields (not only link related)

js/ui/hierarchical_collection/ui.hierarchical_collection_widget.js
Krijovnick: If I revert that changes then all my test will broken, because extra fields like url and linkAttr will not be in attributes

testing/tests/DevExpress.ui.widgets/menuBase.tests.js
Krijovnick: If item.template is specified, we can't control what text will be displayed. User need apply link in the template by himself.
I added this limitation in the specification for further describing in the doc. And test will be added

ksercs: Don't forget to add this to a docs card

js/ui/hierarchical_collection/ui.hierarchical_collection_widget.js
ksercs: oh, ok

js/ui/context_menu/ui.menu_base.js
ksercs: yeap, ok, then we can name it as `getItemExtraPropNames`

vladaskorohodova (dismissed):

js/ui/context_menu/ui.menu_base.js
```diff
@@ -177,14 +177,30 @@ class MenuBase extends HierarchicalCollectionWidget {
...
     _getTextContainer(itemData) {
         const itemText = itemData.text;
+        const url = itemData.url;
         const $itemContainer = $('<span>').addClass(DX_MENU_ITEM_CAPTION_CLASS);
         const itemContent = isPlainObject(itemData) ? itemText : String(itemData);
-
+        if(url && this._isValidHttpURL(url) && itemText) {
+            const linkAttr = isObject(itemData.linkAttr) ? itemData.linkAttr : {};
+            const $urlContainer = $('<a>');
+            $urlContainer.attr({ 'href': url, ...linkAttr })
```
ksercs:
```suggestion
            $urlContainer
                .attr({ 'href': url, ...linkAttr })
```

testing/tests/DevExpress.ui.widgets/menuBase.tests.js
```diff
@@ -333,6 +332,122 @@ QUnit.module('Menu rendering', () => {
 
         assert.equal($element.find('.dx-menu-item-content').first().text(), 'Test item');
     });
+
+    QUnit.test('Render item as link if url start with https', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('item should be rendered as link if item.url starts with https', function(assert) {
```

testing/tests/DevExpress.ui.widgets/menuBase.tests.js
```diff
@@ -333,6 +332,122 @@ QUnit.module('Menu rendering', () => {
...
+    QUnit.test('Render item as link if url start with http', function(assert) {
+        const menuBase = createMenu({
+            items: [{ text: 'Item text', url: 'http://some_url' }]
+        });
+        const content = menuBase.element.find(`.${DX_MENU_ITEM_TEXT_CLASS}`).children()[0];
+
+        assert.strictEqual(content.tagName, 'A');
+        assert.strictEqual(content.getAttribute('href'), 'http://some_url');
+        assert.strictEqual(content.text, 'Item text');
+    });
```
ksercs, line 348: This test and test above can be unificated

js/ui/context_menu/ui.menu_base.js
```diff
@@ -177,14 +177,31 @@ class MenuBase extends HierarchicalCollectionWidget {
         this._initActions();
     }
 
+    _isValidHttpURL(url) {
+        return url.match(/^(http|https):\/\//);
+    }
+
     _getTextContainer(itemData) {
         const itemText = itemData.text;
+        const url = itemData.url;
```
const314, line 186: Minor:
`const { text, url, linkAttr } = itemData;`

ksercs (approved):
Let's wait for .d.ts review by @IlyaKhD

const314 (approved):

IlyaKhD (approved):

PR #24121, opened 2023-03-27
PivotGrid: fix CSP setTimeout issue.

williamvinogradov:
Fix the CSP (incorrect call of the setTimeout) issue in the PivotGrid.

### DIFF
--- a/js/__internal/grids/pivot_grid/module_widget_utils.ts
+++ b/js/__internal/grids/pivot_grid/module_widget_utils.ts
@@ -45,7 +45,7 @@ function createForeachTreeFunc(isAsync) {
 
     index = index || 0;
 
-    function createForeachTreeAsyncHandler(deferred, i, isChildrenProcessing) {
+    function createForeachTreeAsyncHandler(deferred, i, isChildrenProcessing): void {
       when(foreachTreeFunc(items, callback, parentAtFirst, members, i, isChildrenProcessing))
         .done(deferred.resolve);
     }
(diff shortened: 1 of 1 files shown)


js/__internal/grids/pivot_grid/module_widget_utils.ts
```diff
@@ -59,7 +59,8 @@ function createForeachTreeFunc(isAsync) {
         foreachTreeAsyncDate = new Date();
         // @ts-expect-error
         deferred = new Deferred();
-        setTimeout(createForeachTreeAsyncHandler(deferred, i, false) as any, 0);
+        // eslint-disable-next-line @typescript-eslint/no-loop-func
+        setTimeout(() => createForeachTreeAsyncHandler(deferred, i, false) as any, 0);
```
pomahtri: It is true positive from linter, isn't it? setTimeout's callback will be called after for-loop with the same value of `i` 

```
for (var i = 0; i<5; i++) {
    setTimeout(() => console.log(i), 100)
}
```

```
5
5
5
5
5
```

pomahtri: \+ i think it's better to leave the behavior as it was: remove setTimeout, cause we can break something

williamvinogradov: Yep, I agree with you 
Sorry, I hurried and did a mistake :)

pomahtri (approved):

PR #24122, opened 2023-03-27
DataGrid - The state of command buttons is rendered incorrectly in some situations when specific virtual scrolling options are used with repaintChangesOnly and a disabled cache (T1136896)

pomahtri:

### DIFF
--- a/js/ui/grid_core/ui.grid_core.editing_row_based.js
+++ b/js/ui/grid_core/ui.grid_core.editing_row_based.js
@@ -7,6 +7,8 @@ import {
 
 } from './ui.grid_core.editing_constants';
 
+import { equalByValue } from '../../core/utils/common';
+
 const EDIT_ROW = 'dx-edit-row';
 
 export const editingRowBasedModule = {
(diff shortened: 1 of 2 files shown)


js/ui/grid_core/ui.grid_core.editing_row_based.js
```diff
@@ -31,7 +32,7 @@ export const editingRowBasedModule = {
 
                 _isDefaultButtonVisible: function(button, options) {
                     const isRowMode = this.isRowBasedEditMode();
-                    const isEditRow = options.row && options.row.rowIndex === this._getVisibleEditRowIndex();
+                    const isEditRow = options.row && options.row.key === this.option(EDITING_EDITROWKEY_OPTION_NAME);
```
MikeVitik: Minor: I suggest extracting this._getVisibleEditRowKey() from _getVisibleEditRowIndex() and using it here.
It can help to migrate from _getVisibleEditRowIndex to _getVisibleEditRowIndex in future

MikeVitik (dismissed):

williamvinogradov (dismissed):

js/ui/grid_core/ui.grid_core.editing_row_based.js
Alyar666: I think it's better to use the `equalByValue` method to compare keys

Alyar666 (changes requested):

Alyar666 (approved):

PR #24137, opened 2023-03-29
DateRangeBox: add declaration

Zedwag:


js/ui/date_range_box.d.ts
```diff
@@ -0,0 +1,129 @@
...
+     * @default 'calendar'
+     * @default 'native' &for(iOS)
+     * @default 'native' &for(Android)
+     * @public
+     */
+  pickerType?: 'calendar' | 'native';
+  /**
+     * @docid
+     * @default false
+     * @public
```
ksercs: shouldn't it be hidden?

js/ui/date_range_box.js
```diff
@@ -0,0 +1,34 @@
...
+            /**
+             * @name dxPopoverOptions.showAnalogClock
+             * @hidden
+             */
+            showAnalogClock: false,
+
+            /**
+            * @name dxDateRangeBoxOptions.type
+            * @hidden
+            */
```
ksercs: Cannot we hide all of this in d.ts, not here?

js/ui/date_range_box.d.ts
Zedwag: it's hidden but ins .js file

js/ui/date_range_box.js
Zedwag: there's literally 0 (zero) options that are hidden in d.ts file in whole project. We discussed this once. I think it's not for some some kind of joke, it's made in purpose :)

@IlyaKhD what is the reason of hiding options in .js instead of .d.ts?

js/ui/date_range_box.d.ts
Krijovnick: Why we need to declare showAnalogClock there and then hide it?

js/ui/date_range_box.d.ts
```diff
@@ -0,0 +1,129 @@
...
+     */
+  endDate?: Date | number | string;
+  /**
+     * @docid
+     * @default 'calendar'
+     * @default 'native' &for(iOS)
+     * @default 'native' &for(Android)
+     * @public
+     */
+  pickerType?: 'calendar' | 'native';
```
mpreyskurantov: Should be a type of a public enum (similar to /date_box.d.ts/DatePickerType).

js/ui/date_range_box.js
```diff
@@ -0,0 +1,34 @@
...
+
+            /**
+            * @name dxDateRangeBoxOptions.interval
+            * @hidden
+            */
+
+            /**
+             * @name dxPopoverOptions.showAnalogClock
```
mpreyskurantov: dxDateRangeBoxOptions.showAnalogClock

mpreyskurantov (changes requested):

js/ui/date_range_box.js
mpreyskurantov: Keep it as is for now

js/ui/date_range_box.d.ts
Zedwag: It's pretty common situation when we want hide option from docs but set another default value (for example tooltip: https://github.com/DevExpress/DevExtreme/blob/23_1/js/ui/tooltip/tooltip.js#L13)

In this case I thought that it's a bit strange to leave default value as true. But anyway option changes nothing, because there is no `datetime` mode in DateRangeBox, so clock will never be shown, so I think I'll get rid of this defalut value change

js/ui/date_range_box.d.ts
Zedwag: done

js/ui/date_range_box.js
Zedwag: oops, fixed, ty

js/ui/date_range_box.d.ts
```diff
@@ -0,0 +1,126 @@
...
+    & ValueChangedInfo;
+
+/** @public */
+export type DropDownButtonTemplateData = DropDownButtonTemplateDataModel;
+
+/**
+ * @deprecated use Properties instead
+ * @namespace DevExpress.ui
+ */
+export interface dxDateRangeBoxOptions extends Omit<dxDateBoxOptions<dxDateRangeBox>, 'value'> {
```
mpreyskurantov: Do not create NEW always deprecated interface: Create new type (like public alias) instead:
- DocComment (@public doc tag only)
- Take care of showAnalogClock member:

```
/**
 * @public
 */
export type Properties = Omit<dxDateBoxOptions<dxDateRangeBox>, 'value' | 'showAnalogClock'> & { }
```

js/ui/date_range_box.d.ts
```diff
@@ -0,0 +1,126 @@
...
+  /**
+     * @docid
```
mpreyskurantov: Specify docid values for all nested members after removing dxDateRangeBoxOptions with this prefix:
@docid dxDateRangeBoxOptions.endDate
etc.

js/ui/date_range_box.d.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/**
+ * @docid
+ * @isEditor
+ * @inherits dxDateBox
+ * @namespace DevExpress.ui
+ * @public
+ */
+export default class dxDateRangeBox extends dxDateBox<dxDateRangeBoxOptions> { }
```
mpreyskurantov:
```
export default class dxDateRangeBox extends dxDateBox<Properties>
```

js/ui/date_range_box.d.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/** @public */
+export type Properties = dxDateRangeBoxOptions;
+
+/** @deprecated use Properties instead */
+export type Options = dxDateRangeBoxOptions;
```
mpreyskurantov: Remove

mpreyskurantov (changes requested):

js/ui/date_range_box.d.ts
```diff
@@ -0,0 +1,119 @@
...
+     */
+  endDate?: Date | number | string;
+  /**
+     * @docid dxDateRangeBoxOptions.pickerType
+     * @default 'calendar'
+     * @default 'native' &for(iOS)
+     * @default 'native' &for(Android)
+     * @public
+     */
+  pickerType?: DateRangePickerType;
```
ksercs: Why do we omit `value` from `dxDateBoxOptions<dxDateRangeBox>`, but don't omit a `pickerType`?

Zedwag: As I understand, we have no need to omit `pickerType` because its includes types from parent. For example if parent has type `'a' | 'b'`, child can have type `'a'` without omiting. But value has completely enother type

May be ApiReviewers will correct my answer

ksercs: then y do we need to omit `value` and other props?)
@IlyaKhD consult us please

js/ui/date_range_box.js
```diff
@@ -0,0 +1,33 @@
+import registerComponent from '../core/component_registrator';
```
EugeniyKiyashko: May be we need create a folder for date_range_box component and move this code there. In this file we leave import and export only

```
import DateRangeBox from './date_range_box/ui.date_range_box';
export default DateRangeBox;
```

What do you think about it?

js/ui/date_range_box.js
```diff
@@ -0,0 +1,33 @@
...
+import { extend } from '../core/utils/extend';
+import DateBox from './date_box';
+
+const DateRangeBox = DateBox.inherit({
```
EugeniyKiyashko: Seems comment
// STYLE dateRangeBox 
is missed. I think it will be useful in the future. All public components has this comment.

EugeniyKiyashko: May be we start write component as class component?
`class DateRangeBox extends DateBox { ...`

Zedwag: I'm not sure it won't cause any issues, but let's try

Zedwag: hm, missed this, ty

js/ui/date_range_box.js
Zedwag: I planned to do this in next pr where we will need new folder because of adding strategy, but sure we can do this now

js/ui/date_range_box.d.ts
IlyaKhD: 1) If you don't omit `value`, it will have merged type:
```ts
((string | number | Date) & (string | number | Date)[]) | undefined
```

instead of
```ts
 (string | number | Date)[] | undefined
```

This happens because `dxDateBoxOptions` and anonymous type are combined with intersection. If a property is presented  in both types, its type will be a union of both possible types:

```ts
type merged = (Array<Date | number | string>) | (Date | number | string)
```

IlyaKhD: 2. AFAIU `'adaptivityEnabled' | 'interval' | 'type' | 'showAnalogClock'>` are omitted because they are just not needed in the new type (`Properties`)

IlyaKhD: 3. Technically, we don't have to include `pickerType` in the omit statement, but I would do it to explicitly point to the fact that this prop is overridden.

Why it works at the moment - because when these two types are merged, the result looks like the narrower one:

```ts
type DatePickerType = "calendar" | "list" | "native" | "rollers"
type DateRangePickerType = "calendar" | "native"

type merged = DatePickerType | DateRangePickerType // equals "calendar" | "native"
```

ksercs (dismissed):

IlyaKhD (approved):

mpreyskurantov (approved):

ksercs (approved):

sergepilipchuk (approved):

PR #24147, opened 2023-03-30
DataGrid: Fix an issue where the header filter popup is shown outside the viewport (T1156848)

markallenramirez:


pomahtri (dismissed):

pomahtri (commented):
from thread: rewrite test to testcafe

pomahtri (dismissed):

pomahtri (approved):

PR #24148, opened 2023-03-30
DataGrid - All rows are selected when deleting a row with deferred selection and refreshMode set to refresh/repaint (T1153564)

pomahtri:


js/ui/grid_core/ui.grid_core.selection.js
```diff
@@ -115,7 +115,17 @@ const SelectionController = gridCore.Controller.inherit((function() {
...
+
+            if(this.option('selection.deferred')) {
+                removedKeys = removedKeys
+                    .map(key => this._dataController.items().find(item => item.key === key))
+                    .filter(item => item.isSelected)
+                    .map(item => item.key);
+            }
+
             removedKeys.length && this.deselectRows(removedKeys);
         },
```
williamvinogradov: Let's move change this code and move it to pure function (located in this file, but outside class):

```javascript
const getRemovedKeys = (changes, dataControllerItems, isDeferredSelection) => {
    const removedKeys = changes
        .filter(change => change.type === 'remove')
        .map(change => change.key);

    if (!isDeferredSelection) {
        return removedKeys;
    }

    // Option 1: If it's ok to compare keys by hash.
    const selectedKeys = dataControllerItems
        .filter((item) => item.isSelected)
        .map((item) => item.key);
    const selectedKeysSet = new Set(selectedKeys);
    return removedKeys.filter((key) => selectedKeysSet.has(key));

    // Option 2: If we should use equalByValue.
    const selectedKeys = dataControllerItems
        .filter((item) => item.isSelected)
        .map((item) => item.key);
    return removedKeys.filter((key) => selectedKeys.find((selectedKey) => equalByValue(key, selectedKey)));
}
```

And use it here:

```javascript
        _handleDataPushed: function(changes) {
            const removedKeys = getRemovedKeys(
              changes,
              this._dataController.items(),
              this.option('selection.deferred'),
            );
            removedKeys.length && this.deselectRows(removedKeys);
        },
```

williamvinogradov: Discussed: We should choose option #2 and moving to separate function it's a minor comment

williamvinogradov (approved):

PR #24162, opened 2023-03-31
DataGrid - Customization Column Chooser feature

Tucchhaa:


js/ui/grid_core/ui.grid_core.column_chooser.js
```diff
@@ -319,23 +295,61 @@ const columnChooserMembers = {
...
+        const selectionOptions = this.option('columnChooser.selection') || {};
 
-            if(isLastColumnUnselected) {
-                e.component.selectItem(e.itemElement);
-            } else {
-                setTimeout(function() {
-                    that._columnsController.columnOption(e.itemData.id, 'visible', e.itemData.selected);
-                }, CLICK_TIMEOUT);
-            }
+        const forEachNode = (nodes, func) => {
```
MikeVitik: Sudgesstion: Try transforming forEachNode to the flatNode function returned flat array (Possibly this helper function already exists)

js/ui/grid_core/ui.grid_core.column_chooser.js
```diff
@@ -319,23 +295,61 @@ const columnChooserMembers = {
...
+        let isUpdatingSelection = false;
+
         return {
-            selectNodesRecursive: false,
-            showCheckBoxesMode: 'normal',
-            onItemSelectionChanged: selectionChangedHandler
+            selectByClick: selectionOptions.selectByClick,
+            selectNodesRecursive: selectionOptions.recursive,
+            showCheckBoxesMode: selectionOptions.allowSelectAll ? 'selectAll' : 'normal',
+            onSelectionChanged: (e) => {
```
MikeVitik: Suggestion: Try replacing the isUpdatingSelection flag to subscribe/unsubscribe logic

js/ui/grid_core/ui.grid_core.column_chooser.js
```diff
@@ -319,23 +295,61 @@ const columnChooserMembers = {
...
+                    isUpdatingSelection = true;
+                    e.component.beginUpdate();
+
+                    updateSelection(e);
+
+                    isUpdatingSelection = false;
+                    e.component.endUpdate();
+
+                    clearTimeout(_timeout);
+                    _timeout = setTimeout(() => {
```
MikeVitik: Question: Do we really need this timeout? Maybe we can solve the problem using treeView API?

MikeVitik (dismissed):

js/common/grids.d.ts
```diff
@@ -515,6 +527,50 @@ export type ColumnChooser = {
...
+  enabled?: boolean;
+  /**
+   * @docid GridBaseOptions.columnChooser.search.timeout
+   * @default 500
+   */
+  timeout?: number;
+};
+
+/**
+ * @pulibc
```
mpreyskurantov: Spell. Should be @public

js/common/grids.d.ts
```diff
@@ -515,6 +527,50 @@ export type ColumnChooser = {
...
+ * @namespace
```
mpreyskurantov: Why is it empty? What is purpose/final goal?

mpreyskurantov (dismissed):

js/common/grids.d.ts
Tucchhaa: That's mistake. I was inattentive

js/ui/grid_core/ui.grid_core.column_chooser.js
Tucchhaa: Unfortunately it is impossible here. Handlers that are assigned as options are not added into eventsStrategy._events so we can not subscribe/unsubscribe this handler

js/common/grids.d.ts
```diff
@@ -515,6 +527,50 @@ export type ColumnChooser = {
...
+/**
+ * @public
+ * @namespace DevExpress.common.grids
+ */
+export type ColumnChooserSearch = {
+  /**
+   * @docid GridBaseOptions.columnChooser.search.editorOptions
+   * @default {}
+   */
+  editorOptions?: any;
```
const314, line 538: Why don't we have a stricter type for this field?

Tucchhaa: Cause in every other editorOptions field we use type of any. There is type [TextBoxProperties](https://github.com/Tucchhaa/DevExtreme/blob/22_2/js/ui/widget/ui.search_box_mixin.d.ts#L17), but any editorOptions use it. 

If we can use this, lets do it for all editorOptions in a separate PR

js/ui/grid_core/ui.grid_core.column_chooser.js
```diff
@@ -43,12 +41,11 @@ const processItems = function(that, chooserColumns) {
...
                 expanded: true,
                 id: column.index,
-                disabled: false,
-                disableCheckBox: column.allowHiding === false,
+                disabled: column.allowHiding === false,
                 parentId: isDefined(column.ownerBand) ? column.ownerBand : null
             };
 
-            if(isSelectMode) {
+            if(isSelectMode && !(isRecursive && column.hasColumns)) {
```
williamvinogradov: Minor comment.

Let's move condition in brackets into a separate variable:

```suggestion
            const isRecursiveWithColumns = isRecursive && column.hasColumns;
            if(isSelectMode && !isRecursiveWIthColumns) {
```

sergepilipchuk (dismissed):

js/ui/grid_core/ui.grid_core.column_chooser.js
```diff
@@ -318,23 +294,67 @@ const columnChooserMembers = {
...
+            nodes.forEach(node => {
+                result.push(node);
+
+                if(node.children.length) {
+                    flatNodes(node.children, result);
+                }
+            });
+
+            return result;
+        };
```
williamvinogradov: Let's change this function:
- Incapsulate default ```result``` argument, to prevent incorrect usage of this function in the future :)
- Let's use ```reduce```
- Let's rename this function to ```getFlatNodes```

```suggestion
    const getFlatNodes = (nodes) => {
      const addNodesToArray = (nodes, flatNodesArray) => {
        return nodes.reduce((result, node) => {
            result.push(node);

            if(node.children.length) {
                addNodesToArray(node.children, result);
            }

            return result;
        }, flatNodesArray);
    };

    return addNodesToArray(nodes, []);
};
```

js/ui/grid_core/ui.grid_core.column_chooser.js
```diff
@@ -318,23 +294,67 @@ const columnChooserMembers = {
...
+
+        const updateSelection = (e, nodes) => {
+            nodes.forEach(node => {
+                if(node.itemData.allowHiding === false) {
+                    e.component.selectItem(node.key);
+                }
+            });
+        };
```
williamvinogradov: Minor comment.

Let's make this function shorter:
```suggestion
        const updateSelection = (e, nodes) => {
          nodes.filter(({ itemData: { allowHiding } }) => allowHiding === false)
                    .forEach(({ key }) => { e.component.selectItem(key); });
        };
```

js/ui/grid_core/ui.grid_core.column_chooser.js
```diff
@@ -318,23 +294,67 @@ const columnChooserMembers = {
...
+
+        const updateColumnVisibility = (nodes) => {
+            nodes.forEach(node => {
+                const columnIndex = node.itemData.id;
+                const isVisible = node.selected !== false;
+                that._columnsController.columnOption(columnIndex, 'visible', isVisible);
+            });
+        };
+
+        let _timeout;
```
williamvinogradov: Let's not use the underscore
Instead of this, we can rename this variable :)
E.g ```_timeout``` -> ```updateColumnVisibilityTimeout```

js/ui/grid_core/ui.grid_core.column_chooser.js
```diff
@@ -318,23 +294,67 @@ const columnChooserMembers = {
...
+                clearTimeout(_timeout);
+                _timeout = setTimeout(() => {
+                    that.component.beginUpdate();
+
+                    updateColumnVisibility(nodes);
+
+                    that.component.endUpdate();
                 }, CLICK_TIMEOUT);
             }
         };
```
williamvinogradov: Let's add return at the start of the function instead of wrapping all function code into one if condition:

```suggestion
        const selectionChangedHandler = (e) => {
            if(isUpdatingSelection) {
              returnp
            }
            
           const nodes = flatNodes(e.component.getNodes());

           isUpdatingSelection = true;
           
           e.component.beginUpdate();
           updateSelection(e, nodes);

           isUpdatingSelection = false;
           e.component.endUpdate();

          clearTimeout(updateColumnVisibilityTimeout);
              updateColumnVisibilityTimeout = setTimeout(() => {
                   that.component.beginUpdate();

                    updateColumnVisibility(nodes);

                    that.component.endUpdate();
                }, CLICK_TIMEOUT);
            }
        };
```

testing/testcafe/tests/dataGrid/columnChooser.ts
```diff
@@ -42,3 +42,73 @@ test('Column chooser screenshot', async (t) => {
...
 }));
+
+test('Column chooser checkboxes should be aligned correctly with plain structure', async (t) => {
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+  const dataGrid = new DataGrid('#container');
+
+  await t
+    .click(dataGrid.getHeaderPanel().getColumnChooserButton())
+    .wait(500);
```
williamvinogradov: Why do we add ```wait``` here?

testing/testcafe/tests/dataGrid/columnChooser.ts
```diff
@@ -42,3 +42,73 @@ test('Column chooser screenshot', async (t) => {
...
+}));
+
+test('Column chooser checkboxes should be aligned correctly with tree structure', async (t) => {
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+  const dataGrid = new DataGrid('#container');
+
+  await t
+    .click(dataGrid.getHeaderPanel().getColumnChooserButton())
+    .wait(500);
```
williamvinogradov: Same question about ```wait```
Let's remove it if we can or reduce it to a minimum possible value

testing/testcafe/tests/dataGrid/columnChooser.ts
Tucchhaa: I added this because of the CLICK_TIMEOUT, but you are right `wait` is redundant here, because I compare screenshot only of the column chooser and not the whole grid.

js/common/grids.d.ts
IlyaKhD: Sounds reasonable. This `editorOptions` case looks like something special - so we should check all such properties and try to figure out optimal way of typing them (and check metadata after that)

williamvinogradov (dismissed):

const314 (dismissed):

williamvinogradov (dismissed):

const314 (approved):

williamvinogradov (approved):

PR #24169, opened 2023-04-04
DataGrid: Fix blank rows with cellRenders appear during infinite scrolling (T1152498)

Alyar666:
See https://devexpress.com/issue=T1152498


MikeVitik (dismissed):

js/ui/grid_core/ui.grid_core.columns_view.js
```diff
@@ -918,17 +918,30 @@ const columnsViewMembers = {
...
+            if(this._templateDeferreds.size > 0) {
+                this.waitAsyncTemplates(forceWaiting, result);
+            } else {
+                result.resolve();
+            }
+        }).fail(result.reject);
 
-        return when.apply(this, templateDeferreds);
+        return result.promise();
     },
```
williamvinogradov: Let's don't change the ```waitAsyncTemplates``` method signature and incapsulate the recursion inside of this method

Something like:
```typescript
waitAsyncTemplates: function(forceWaiting = false) {
        // @ts-expect-error
        const result = new Deferred();
        const needWaitAsyncTemplates = forceWaiting || this.needWaitAsyncTemplates();

        if(!needWaitAsyncTemplates) {
            return result.resolve();
        }

        const waitTemplatesRecursion = () =>
            when.apply(this, Array.from(this._templateDeferreds))
                .then(() => this._templateDeferreds.size > 0
                    ? waitTemplatesRecursion()
                    : undefined);

        waitTemplatesRecursion()
            .done(result.resolve)
            .fail(result.reject);

        return result.promise();
    },
```

js/ui/grid_core/ui.grid_core.rows.js
```diff
@@ -314,72 +314,82 @@ export const rowsModule = {
...
-                                                    } else if(columnIndices) {
-                                                        this._updateCells($rowElement, $newRowElement, columnIndices);
+                        const contentChanges = this._contentChanges;
+
+                        this._contentChanges = [];
+
+                        contentChanges.forEach(({ newTableElement, change, isFixedTableRendering }) => {
+                            const tableElement = this.getTableElement(isFixedTableRendering);
+                            const contentElement = this._findContentElement(isFixedTableRendering);
+                            const changeType = change && change.changeType;
```
williamvinogradov: Can we use ```?.``` here?

```suggestion
                            const changeType = change?.changeType;
```

js/ui/grid_core/ui.grid_core.rows.js
```diff
@@ -314,72 +314,82 @@ export const rowsModule = {
...
+                            const executors = [];
+                            const highlightChanges = this.option('highlightChanges');
+                            const rowInsertedClass = this.addWidgetPrefix(ROW_INSERTED_ANIMATION_CLASS);
+
+                            switch(changeType) {
+                                case 'update':
+                                    each(change.rowIndices, (index, rowIndex) => {
+                                        const $newRowElement = this._getRowElements(newTableElement).eq(index);
+                                        const changeType = change.changeTypes && change.changeTypes[index];
+                                        const item = change.items && change.items[index];
```
williamvinogradov: Let's use ```?.``` here and rename this ```changeType``` to avoid shadowing of the variable
(we've ```changeType``` declared before this code block)

```typescript
// I'm not sure about this name, maybe we can name it better :)
const rowChangeType = change.changeTypes?.[index];
const item = change.items?.[index];
```

js/ui/grid_core/ui.grid_core.rows.js
```diff
@@ -314,72 +314,82 @@ export const rowsModule = {
...
+                                                        const columnIndices = change.columnIndices && change.columnIndices[index];
+                                                        if(isDefined(item.visible) && item.visible !== $rowElement.is(':visible')) {
+                                                            $rowElement.toggle(item.visible);
+                                                        } else if(columnIndices) {
+                                                            this._updateCells($rowElement, $newRowElement, columnIndices);
+                                                        } else {
+                                                            $rowElement.replaceWith($newRowElement);
+                                                        }
+                                                    }
+                                                    break;
```
williamvinogradov: Let's make this section a little simpler:

```typescript
case 'update':
  if (!item) {
    break;
  }

  const columnIndices = change.columnIndices?.[index];
                                                                                                
  if(isDefined(item.visible) && item.visible !== $rowElement.is(':visible')) {
    $rowElement.toggle(item.visible);
  } else if(columnIndices) {
      this._updateCells($rowElement, $newRowElement, columnIndices);
  } else {
    $rowElement.replaceWith($newRowElement);
  } 
  break;
```

js/ui/grid_core/ui.grid_core.rows.js
```diff
@@ -314,72 +314,82 @@ export const rowsModule = {
...
+                                                    if(!$rowsElement.length) {
+                                                        if(tableElement) {
+                                                            const target = $newRowElement.is('tbody') ? tableElement : tableElement.children('tbody');
+                                                            $newRowElement.prependTo(target);
+                                                        }
+                                                    } else if($rowElement.length) {
+                                                        $newRowElement.insertBefore($rowElement);
                                                     } else {
-                                                        $rowElement.replaceWith($newRowElement);
+                                                        $newRowElement.insertAfter($rowsElement.last());
```
williamvinogradov: Is that code reachable?

Because we already handle ```$rowElement.length``` and ```!$rowElement.length``` before this else block

testing/testcafe/tests/dataGrid/scrolling.ts
```diff
@@ -1287,3 +1287,66 @@ test('New virtual mode. Navigation to the last row if new row is added (T1069849
...
+            ($(options.container) as any).append(($('<div/>') as any).text(options.model.value));
+            options.deferred?.resolve();
+          }, 400);
+        },
+      });
+
+      dataGrid.repaint();
+    })();
+
+    await t.wait(500);
```
williamvinogradov: Question.

Can we remove this await?

testing/testcafe/tests/dataGrid/scrolling.ts
```diff
@@ -1287,3 +1287,66 @@ test('New virtual mode. Navigation to the last row if new row is added (T1069849
...
+      // eslint-disable-next-line no-underscore-dangle
+      dataGrid.getView('rowsView')._templatesCache = {};
+
+      // eslint-disable-next-line no-underscore-dangle
+      dataGrid._getTemplate = () => ({
+        render(options) {
+          setTimeout(() => {
+            ($(options.container) as any).append(($('<div/>') as any).text(options.model.value));
+            options.deferred?.resolve();
+          }, 400);
```
williamvinogradov: Question.

Can we reduce the time of this ```setTimeout```?

testing/testcafe/tests/dataGrid/scrolling.ts
```diff
@@ -1287,3 +1287,66 @@ test('New virtual mode. Navigation to the last row if new row is added (T1069849
...
+
+    // act
+    await dataGrid.scrollTo({ y: 200 });
+    await t.wait(200);
+    await dataGrid.scrollTo({ y: 400 });
+
+    // assert
+    await t
+      .wait(500)
+      .expect(await takeScreenshot(`grid-${scrollingMode}-scrolling-T1152498.png`, '#container'))
```
williamvinogradov: Question.

Maybe we can move last await to act part of the tests?
```suggestion
    await dataGrid.scrollTo({ y: 200 });
    await t.wait(200);
    await dataGrid.scrollTo({ y: 400 });
    await t.wait(200);

    // assert
    await t
      .expect(await takeScreenshot(`grid-${scrollingMode}-scrolling-T1152498.png`, '#container'))
```

js/ui/grid_core/ui.grid_core.rows.js
Alyar666: !$row**s**Element.length and $rowElement.length - different variables are checked here

Alyar666: I think it's better to rename the $row**s**Element variable to $rowElement**s**

js/ui/grid_core/ui.grid_core.columns_view.js
```diff
@@ -918,16 +918,33 @@ const columnsViewMembers = {
...
 
     waitAsyncTemplates: function(forceWaiting = false) {
-        const needWaitAsyncTemplates = this.needWaitAsyncTemplates();
-        const templateDeferreds = (forceWaiting || needWaitAsyncTemplates) && Array.from(this._templateDeferreds) || [];
+        // @ts-expect-error
+        const result = new Deferred();
```
pomahtri, line 922: minor
```suggestion
        const result = Deferred();
```

js/ui/grid_core/ui.grid_core.rows.js
```diff
@@ -320,72 +320,83 @@ export const rowsModule = {
...
+                            const executors = [];
+                            const highlightChanges = this.option('highlightChanges');
+                            const rowInsertedClass = this.addWidgetPrefix(ROW_INSERTED_ANIMATION_CLASS);
+
+                            switch(changeType) {
+                                case 'update':
+                                    each(change.rowIndices, (index, rowIndex) => {
+                                        const $newRowElement = this._getRowElements(newTableElement).eq(index);
+                                        const dataChangeType = change.changeTypes?.[index];
+                                        const item = change.items && change.items[index];
```
williamvinogradov, line 344: Minor comment.

Let's use ```?.``` here too:
```suggestion
                                        const item = change.items?.[index];
```

williamvinogradov (approved):

PR #24170, opened 2023-04-04
ActionSheet: set role=dialog for popover

AlexanderMoiseev:


testing/tests/DevExpress.ui.widgets/actionSheet.tests.js
```diff
@@ -51,6 +51,25 @@ QUnit.module('action sheet', {
         assert.equal(popup.option('height'), 'auto');
     });
 
+    QUnit.test('Overlay should have role="dialog" attribute when using popover', function(assert) {
+        $('#actionSheet').dxActionSheet({
+            usePopover: true,
+            target: $('#container')
+        });
+
+        const $popoverInstance = $('.dx-popover').dxPopover('instance');
```
ksercs: let's use constants

testing/tests/DevExpress.ui.widgets/actionSheet.tests.js
```diff
@@ -51,6 +51,25 @@ QUnit.module('action sheet', {
         assert.equal(popup.option('height'), 'auto');
     });
 
+    QUnit.test('Overlay should have role="dialog" attribute when using popover', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('popover should have role="dialog" attribute when usePopover=true', function(assert) {
```

testing/tests/DevExpress.ui.widgets/actionSheet.tests.js
```diff
@@ -51,6 +51,25 @@ QUnit.module('action sheet', {
...
+        $('#actionSheet').dxActionSheet({
+            usePopover: true,
+            target: $('#container')
+        });
+
+        const $popoverInstance = $('.dx-popover').dxPopover('instance');
+
+        assert.strictEqual($popoverInstance.$overlayContent().attr('role'), 'dialog');
```
ksercs: no private api plz

testing/tests/DevExpress.ui.widgets/actionSheet.tests.js
AlexanderMoiseev: i believe in this case block with ' when usePopover=true' text in not necessary, since we explicitly specified that popover is gonna be used in this test

ksercs (approved):

marker-dao (approved):

PR #24186, opened 2023-04-05
Сore: Replace "new Function" by JSON.parse

GoodDayForSurf:
Replace "new Function" by JSON.parse for CSP restrictions script-src=unsafe-eval

### DIFF
--- a/js/core/config.js
+++ b/js/core/config.js
@@ -44,13 +44,24 @@ const config = {
         if(optionsString.trim().charAt(0) !== '{') {
             optionsString = '{' + optionsString + '}';
         }
+
         try {
-            // eslint-disable-next-line no-new-func
-            return (new Function('return ' + optionsString))();
+            return JSON.parse(optionsString);
         } catch(ex) {
-            throw errors.Error('E3018', ex, optionsString);
+            try {
+                return JSON.parse(normalizeToJSONString(optionsString));
+            } catch(exNormalize) {
+                throw errors.Error('E3018', ex, optionsString);
+            }
         }
-    }
+    },
+};
+
+const normalizeToJSONString = (optionsString) => {
+    return optionsString
+        .replace(/'/g, '"') // replace all ' to "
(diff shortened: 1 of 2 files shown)


js/core/config.js
```diff
@@ -44,13 +44,25 @@ const config = {
...
+                throw errors.Error('E3018', ex, optionsString);
+            }
         }
-    }
+    },
+};
+
+const normalizeToJSONString = (optionsString) => {
+    return optionsString
+        .replace(/([{,])[\s\n]*'([^']+)'[\s\n]*:/g, '$1"$2":') // replace quotes for single quoted keys
```
timbset: Isn't it better to replace all single quotes first (like we do it on the line 64) and then do not care about keys wrapped by single quotes?

testing/tests/DevExpress.core/config.tests.js
```diff
@@ -88,3 +88,31 @@ QUnit.test('deprecated fields', function(assert) {
...
+            { 'dxTemplate': { 'name': 'title', test: { arr: ['a1', 'a2'] } } }
+        ],
+        [
+            '"dxTemplate": { "name": "title" }',
+            { 'dxTemplate': { 'name': 'title' } }
+        ],
+        [
+            '{"dxTemplate": { "name": "title" }}',
+            { 'dxTemplate': { 'name': 'title' } }
+        ]
```
timbset, line 122: We check the `\n` symbol in regular expression, but there is no test with multiline template. Maybe add it?

timbset (changes requested):

js/core/config.js
GoodDayForSurf: reasonable..

testing/tests/DevExpress.core/config.tests.js
GoodDayForSurf: fixed

timbset (dismissed):

timbset (approved):

dxvladislavvolkov (approved):

PR #24190, opened 2023-04-05
Form: fix layout update on setting "colCountByScreen" property of a group

AlexanderMoiseev:


testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4032,6 +4032,25 @@ function getColsCountFromDOM($form) {
     return result;
 }
 
+QUnit.test('Setting "colCountByScreen" property of group with dot notation should update layout', function(assert) {
+    const $form = $('#form').dxForm({
```
EugeniyKiyashko: We can get instance of the form it right away as follows
```
const form = $('#form').dxForm({}).dxForm('instance');
```
To get $element of the Form use $element() method as follows
```
assert.strictEqual(getColsCountFromDOM(form.$element()), 2);
```
JYI. It allows to write less code in tests.

testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4032,6 +4032,25 @@ function getColsCountFromDOM($form) {
     return result;
 }
 
+QUnit.test('Setting "colCountByScreen" property of group with dot notation should update layout', function(assert) {
```
EugeniyKiyashko: I didn't see tests with testing changing `colCountByScreen` option for Form & Tabbed items in runtime.
image
Are you checked these scenarios. May be these test are written and you may add permalink to demonstrate them. 
If no I think we should check these cases.

testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4032,6 +4032,25 @@ function getColsCountFromDOM($form) {
...
+                itemType: 'group',
+                colCountByScreen: { md: 3 },
+                items: [{ dataField: 'name' }, { dataField: 'email' }, { dataField: 'age' }]
+            }
+        ],
+    });
+
+    const instance = $form.dxForm('instance');
+
+    instance.option('items[0].colCountByScreen.md', 2);
```
EugeniyKiyashko: For more complicated path with nested items it works as well ? For example
```
instance.option('items[0].item[1].colCountByScreen.md', 2);
or
form.itemOption('groupName.itemName', 'colCountByScreen', { md: '2' })
// https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxForm/Methods/#itemOptionid_option_value

File: testing/tests/DevExpress.ui.widgets.form/form.tests.js (hunk 2 above)
Comment (AlexanderMoiseev) [reply testing/tests/DevExpress.ui.widgets.form/form.tests.js]: yes it uses the same code and working, ok i'll add some test

Comment (ksercs) [inline testing/tests/DevExpress.ui.widgets.form/form.tests.js]:
```suggestion
QUnit.test('group colCountByScreen property change should update layout', function(assert) {
```

testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4032,6 +4032,25 @@ function getColsCountFromDOM($form) {
...
 }
 
+QUnit.test('Setting "colCountByScreen" property of group with dot notation should update layout', function(assert) {
+    const $form = $('#form').dxForm({
+        screenByWidth: () => 'md',
+        items: [
+            {
+                itemType: 'group',
+                colCountByScreen: { md: 3 },
+                items: [{ dataField: 'name' }, { dataField: 'email' }, { dataField: 'age' }]
```
ksercs: is it necessary got the test? Config should be minified as possible

AlexanderMoiseev: 'dataField:' can be omitted

testing/tests/DevExpress.ui.widgets.form/form.tests.js
AlexanderMoiseev: i didn't check  the case with 'itemOption' - thanks

EugeniyKiyashko:
You missed the ticket number in name of your PR

testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4032,6 +4032,75 @@ function getColsCountFromDOM($form) {
...
+                itemType: 'group',
+                colCountByScreen: { md: 2 },
+                items: ['Phone', 'Email'],
+            }],
+        }],
+    }).dxForm('instance');
+
+    form.option('items[0].items[0].colCountByScreen.md', 1);
+
+    assert.strictEqual(getColsCountFromDOM($(form.$element())), 1);
```
EugeniyKiyashko: $(form.$element()) - additional wrapping is not required in this case. But you say it doesn't work for me without it. With this wrapper, you probably overcome the problem that hasClass() jquery - iterate over all elements and check against each of the array. There is a suspicion in the renderer that it does not go through all the elements, but looks only at the first one.

Let's try remove the wrapper and rewrite this piece so that it works well in both cases
```
  [1, 2, 3, 4].forEach(colCount => {
      if($lastCol.hasClass(`dx-col-${colCount - 1}`)) {
          result = colCount;
      }
  });
```
Let me know if I missed something.

EugeniyKiyashko:
Don't forget to think about the need to pass this change commit to the previous versions

testing/tests/DevExpress.ui.widgets.form/form.tests.js
AlexanderMoiseev: good catch, yes our renderer_base's 'hasClass' method seems inconsistent with jquery's 'hasClass'

ksercs:
@AlexanderMoiseev  plz, add a ticket ID to PR's name and to an appropriate test's name

ksercs (dismissed):

EugeniyKiyashko (approved):

ksercs (approved):

PR #24201, opened 2023-04-05
Feature/license checking

iBat:


js/core/utils/license.js
```diff
@@ -0,0 +1,34 @@
...
+        return;
+    }
+    const encodedLicense = /* process.env.DX_LICENSE || */ config().license || '';
+
+    try {
+        // TODO do a real license key extraction
+        // eslint-disable-next-line no-undef
+        const decodedLicense = JSON.parse(atob(encodedLicense));
+
+        if(decodedLicense?.versions.indexOf('23.1') === -1) {
```
helen-dikareva: Instead of the hardcoded we can get the current version number from the package.json file

dxbykov: AFAIR, previously we had something like DevExtreme.version
packange.json is not available in the web browser.

ts/dx.all.d.ts
```diff
@@ -1523,6 +1523,8 @@ declare module DevExpress.common {
      * [descr:GlobalConfig.useLegacyVisibleIndex]
      */
     useLegacyVisibleIndex?: boolean;
+
+    license?: string;
```
dxbykov, line 1527: licenseKey?

PR #24205, opened 2023-04-06
Menu: remove limit for url and href from linkAttr option

Krijovnick:


js/ui/context_menu/ui.menu_base.js
```diff
@@ -177,25 +177,20 @@ class MenuBase extends HierarchicalCollectionWidget {
...
         const $itemContainer = $('<span>').addClass(DX_MENU_ITEM_CAPTION_CLASS);
-        const itemContent = isPlainObject(itemData) ? itemText : String(itemData);
-        if(url && this._isValidHttpURL(url) && itemText) {
-            const linkAttr = isObject(itemData.linkAttr) ? itemData.linkAttr : {};
-            const $urlContainer = $('<a>');
-            $urlContainer
-                .attr({ 'href': url, ...linkAttr })
+        const itemContent = isPlainObject(itemData) ? text : String(itemData);
+        if(url && text) {
+            const linkAttribute = isObject(linkAttr) ? linkAttr : {};
```
ksercs:
```suggestion
            const linkAttributes = isObject(linkAttr) ? linkAttr : {};
```

ksercs (dismissed):

js/ui/context_menu/ui.menu_base.js
```diff
@@ -177,25 +177,20 @@ class MenuBase extends HierarchicalCollectionWidget {
...
-        const itemContent = isPlainObject(itemData) ? itemText : String(itemData);
-        if(url && this._isValidHttpURL(url) && itemText) {
-            const linkAttr = isObject(itemData.linkAttr) ? itemData.linkAttr : {};
-            const $urlContainer = $('<a>');
-            $urlContainer
-                .attr({ 'href': url, ...linkAttr })
+        const itemContent = isPlainObject(itemData) ? text : String(itemData);
+        if(url && text) {
+            const linkAttributes = isObject(linkAttr) ? linkAttr : {};
+            delete linkAttributes.href;
```
EugeniyKiyashko: May be will better to change to 
```
$('<a>')
                .attr({ ...linkAttributes,  'href': url })
```
In this case seems we shouldn't delete `href` property. For example:
```
{ ...{ href: 'a' }, href: 'b' } -> {href: 'b'}
```
Check it pls. Perhaps, I missing something.

Krijovnick: fixed

ksercs (approved):

PR #24227, opened 2023-04-10
HTMLEditor - Add link button should not have inactive state when cursor placed on right border of text(T1157840)

alexanderPolosatov:


testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/toolbarIntegration.tests.js
```diff
@@ -108,6 +108,27 @@ export default function() {
                 .trigger('dxclick');
         });
 
+        test('No active buttons when cursor placed on right side of formatted text(T1157840)', function(assert) {
+            const initialSelection = {
```
ksercs, line 112: I don't see y it should be "initial"
```suggestion
            const selection = {
```

testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/toolbarIntegration.tests.js
```diff
@@ -108,6 +108,27 @@ export default function() {
...
+                index: 2,
+                length: 3
+            };
+            const $editorContainer = $('#htmlEditor');
```
ksercs:
```suggestion
            const $htmlEditor = $('#htmlEditor');
```

js/ui/html_editor/modules/toolbar.js
```diff
@@ -503,20 +503,29 @@ if(Quill) {
...
-        _markActiveFormatWidget(name, widget, formats) {
+        _markActiveFormatWidget(name, widget, formats, selection) {
             if(this._isColorFormat(name)) {
                 this._updateColorWidget(name, formats[name]);
             }
 
             if('value' in widget.option()) {
                 this._setValueSilent(widget, formats[name]);
             } else {
+                if(selection?.length === 0) { // T1157840
```
ksercs: It cannot be nullish
```suggestion
                if(selection.length === 0) { // T1157840
```

js/ui/html_editor/modules/toolbar.js
```diff
@@ -503,20 +503,29 @@ if(Quill) {
...
+                    const nextPositionFormat = this.quill.getFormat({
```
ksercs: It's more concrete and native for the HtmlEditor
```suggestion
                    const nextIndexFormat = this.quill.getFormat({
```

js/ui/html_editor/modules/toolbar.js
```diff
@@ -503,20 +503,29 @@ if(Quill) {
...
+                        length: 0,
+                        index: selection.index + 1
+                    });
+                    if(!nextPositionFormat[name]) {
```
ksercs: What if there is other format on the next index? We will still show an empty link?

testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/toolbarIntegration.tests.js
```diff
@@ -108,6 +108,27 @@ export default function() {
...
+            const instance = $editorContainer.dxHtmlEditor({
+                value: '<p>text_text_text</p>',
+                toolbar: { items: ['bold'] },
+            })
+                .dxHtmlEditor('instance');
```
ksercs: Miss formatting. It should be on the previous row

js/ui/html_editor/modules/toolbar.js
alexanderPolosatov: can be. see all calls of **_markActiveFormatWidget**. about 'why we pass it only for this call' read in teams thread

js/ui/html_editor/modules/toolbar.js
alexanderPolosatov: if next position has another format, this case will(or did) processed in next(prev) iteration of [loop](https://github.com/DevExpress/DevExtreme/blob/de6ce8e404be8d629a77ef49cc1d2e6e9948d266/js/ui/html_editor/modules/toolbar.js#L498)
>We will still show an empty link?

yes

testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/toolbarIntegration.tests.js
alexanderPolosatov: let's rename to **firstSeection**? I want to show what on second changing we change selection on index + length of _initial_ selection (prev selection)

ksercs: ok, nevermind

js/ui/html_editor/modules/toolbar.js
```diff
@@ -503,20 +503,29 @@ if(Quill) {
...
-        _markActiveFormatWidget(name, widget, formats) {
+        _markActiveFormatWidget(name, widget, formats, selection) {
             if(this._isColorFormat(name)) {
                 this._updateColorWidget(name, formats[name]);
             }
 
             if('value' in widget.option()) {
                 this._setValueSilent(widget, formats[name]);
             } else {
+                if(name === 'link' && selection?.length === 0) { // T1157840
```
ksercs: This is really specific code. Let's move it to a separate function?
like
```javascript
if('value' in widget.option()) {
    this._setValueSilent(widget, formats[name]);
} else if (name === 'link') {
    this._updateLinkButton();
} else {
    widget.$element().addClass(ACTIVE_FORMAT_CLASS);
}
```

I also recommend u to write a more precise note in `_updateLinkButton` function to explain why we cannot make it active. To give a link for `descendant` method and so on.

I think it's also better to get rid of `return` and modify the condition instead of this

alexanderPolosatov: Moved to separated method
Note added
Note approved by Vlada

ksercs (approved):

Zedwag (approved):

PR #24254, opened 2023-04-12
Data grid: get rid of deprecated copy root classes to wrapper 22.2

Raushen:
Removing deprecated copyRootClassesToWrapper option in dxDataGrid


nightskylark (changes requested):
You are trying to merge 22.2 into 23.1. Don't do it. :)

PR #24264, opened 2023-04-12
DataGrid - Pager's editors don't have the "aria-label" attribute in compact mode (T1156999)

Raushen:
Adding aria-label attributes for pager's inputs (see https://supportcenter.devexpress.com/ticket/details/t1156999/datagrid-pager-s-editors-don-t-have-the-aria-label-attribute-in-compact-mode)


js/renovation/ui/pager/page_size/small.tsx
```diff
@@ -15,6 +15,8 @@ import { FullPageSize } from '../common/types';
 import { getElementMinWidth } from '../utils/get_element_width';
 import { InternalPagerProps } from '../common/pager_props';
 
+const PAGER_PAGE_SIZE_SMALL_DESCRIPTION = 'Page size';
```
pomahtri: Needs to be translated

js/renovation/ui/pager/page_size/small.tsx
```diff
@@ -28,6 +30,7 @@ export const viewFunction = ({
     value={pageSize}
     valueChange={pageSizeChange}
     width={width}
+    inputAttr={ { 'aria-label': PAGER_PAGE_SIZE_SMALL_DESCRIPTION } }
```
pomahtri: Let's move object to getter inside `PageSizeSmall` class, otherwise new object will be created each time `viewFunction` is called, it can lead to extra rerenders. Getters inside component's class are memorized

pomahtri (changes requested):

williamvinogradov (dismissed):

pomahtri (approved):

PR #24322, opened 2023-04-17
Toolbar: Fix menu button attributes (T1157065)

marker-dao:


js/ui/toolbar/internal/ui.toolbar.menu.js
```diff
@@ -218,6 +217,9 @@ export default class DropDownMenu extends Widget {
             showTitle: false,
             fullScreen: false,
             _fixWrapperPosition: true,
+            wrapperAttr: {
+                role: 'menu',
```
ksercs: Axe reports about wrapper's child without role `menuitem` and about item's parent without role `menu`. Please, check it

ksercs (changes requested):

ksercs (approved):

alexanderPolosatov (approved):

PR #24326, opened 2023-04-17
Add locale options for filtering and sorting in DataSource

GoodDayForSurf:


js/ui/grid_core/ui.grid_core.data_source_adapter.js
```diff
@@ -836,7 +837,7 @@ export default modules.Controller.inherit((function() {
                 const store = dataSource.store();
                 const dataSourceLoadOptions = dataSource.loadOptions();
                 const loadResult = {
-                    storeLoadOptions: options,
+                    storeLoadOptions: extend({}, options, { compareOptions: dataSourceLoadOptions?.compareOptions }),
```
williamvinogradov: Question to @Alyar666.

Can we use the spread operator instead of the ```extend``` here?

```typescript
storeLoadOptions: {
  ...options,
  compareOptions: dataSourceLoadOptions?.compareOption
},
```

js/ui/grid_core/ui.grid_core.data_source_adapter.js
```diff
@@ -565,7 +565,8 @@ export default modules.Controller.inherit((function() {
                 summary: !remoteOperations.summary,
                 skip: !remoteOperations.paging,
                 take: !remoteOperations.paging,
-                requireTotalCount: cachedExtra && 'totalCount' in cachedExtra || !remoteOperations.paging
+                requireTotalCount: cachedExtra && 'totalCount' in cachedExtra || !remoteOperations.paging,
+                compareOptions: !remoteOperations.filtering || !remoteOperations.sorting
```
williamvinogradov: Question to @Alyar666.

Will it be ok if we will have ```compareOptions: true``` with remote filtering but local sorting?

js/core/utils/data.js
```diff
@@ -181,7 +181,9 @@ export const compileSetter = function(expr) {
     };
 };
 
-export const toComparable = function(value, caseSensitive) {
+export const toComparable = function(value, caseSensitive, options) {
+    options = options || {};
```
williamvinogradov: Let's not override function arguments.

There are two variants:

**add a new variable e.g ```comparableOptions```**
```typescript
const comparableOptions = options || {};
```
And use this variable in the function code.
Btw, if we choose this way, let's remove ```?``` after ```options``` -> ```comparableOptions```
```typescript
// we don't need the ? symbol after the comparableOptions because it's always defined
if(comparableOptions.collateOptions?.sensitivity === 'base')
```

**add the ```?``` symbol everywhere after ```options```**
Because ```options``` may be null / undefined we can handle this just add the ```?``` after it everywhere :)

js/core/utils/data.js
```diff
@@ -191,7 +193,11 @@ export const toComparable = function(value, caseSensitive) {
     }
 
     if(!caseSensitive && typeof value === 'string') {
-        return value.toLowerCase();
+        if(options?.collateOptions?.sensitivity === 'base') {
+            value = value.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
```
williamvinogradov: As far as I understand, we remove here special unicode symbols

Let's move this RegExp to separate constant
E.g 
```typescript
const REMOVE_DIACRITICAL_MARKS_REGEXP = /[\u0300-\u036f]/g;
```

js/core/utils/data.js
```diff
@@ -191,7 +193,11 @@ export const toComparable = function(value, caseSensitive) {
     }
 
     if(!caseSensitive && typeof value === 'string') {
-        return value.toLowerCase();
+        if(options?.collateOptions?.sensitivity === 'base') {
```
williamvinogradov: Question.

Is it ok, that we apply sensitivity: 'base' only in case insensitive comparison?

js/data/abstract_store.js
```diff
@@ -92,6 +93,11 @@ const Store = Class.inherit({
     },
 
     _loadImpl: function(options) {
+        if(!isEmptyObject(this._compareOptions)) {
+            options = options || {};
+            options._compareOptions = { ...this._compareOptions, ...options._compareOptions };
+        }
```
williamvinogradov: Nothing scary will happen if we spread the empty object :)

Let's simplify these lines:
```suggestion
        options._compareOptions = { ...this._compareOptions, ...options._compareOptions };
```

js/data/array_query.js
```diff
@@ -98,9 +98,15 @@ const MapIterator = WrappedIterator.inherit({
     }
 });
 
-const defaultCompare = function(xValue, yValue) {
-    xValue = toComparable(xValue);
-    yValue = toComparable(yValue);
+const defaultCompare = function(xValue, yValue, options) {
+    if(typeof xValue === 'string' && typeof yValue === 'string' && (options?.locale || options?.collateOptions)) {
+
+        /* eslint-disable-next-line */
```
williamvinogradov: Let's directly specify eslint rule that we disable here

js/data/array_query.js
```diff
@@ -130,13 +136,16 @@ const defaultCompare = function(xValue, yValue) {
 };
 
 const SortIterator = Iterator.inherit({
-
     ctor: function(iter, getter, desc, compare) {
```
williamvinogradov, line 139: Let's not override the ctor argument and change these lines a little:

```typescript
  ctor: function(iter, getter, desc, compare) {
    const { compareOptions } = iter;

    if(iter instanceof MapIterator) {
        this.iter = iter;
    } else {
        this.iter = new MapIterator(iter, this._wrap);
        this.iter.compareOptions = compareOptions;
    }

    // Question.
    // Do we need the compareOptions field in this class?
    // I didn't see any usage of this.compareOptions
    // But maybe I missed something :)
    this.compareOptions = compareOptions;
    this.rules = [{ getter: getter, desc: desc, compare: compare, compareOptions }];
}
```

js/data/array_query.js
```diff
@@ -213,7 +222,7 @@ const SortIterator = Iterator.inherit({
             const rule = this.rules[i];
             const xValue = rule.getter(x);
             const yValue = rule.getter(y);
-            const compare = rule.compare || defaultCompare;
+            const compare = rule.compare || ((xValue, yValue) => defaultCompare(xValue, yValue, rule.compareOptions));
```
williamvinogradov: Minor comment.

Let's create ```getDefaultCompare``` function instead of an inline anonymous function here
E.g
```typescript
const getDefaultCompareFunc = (compareOptions) => 
  (xValue, yValue) => defaultCompare(xValue, yValue,compareOptions);
```

And use it here

js/data/array_query.js
```diff
@@ -362,7 +380,8 @@ const FilterIterator = WrappedIterator.inherit({
 
     ctor: function(iter, criteria) {
         this.callBase(iter);
-        this.criteria = compileCriteria(criteria);
+        this.compareOptions = iter.compareOptions;
```
williamvinogradov: Question.

Do we need the field ```compareOptions``` in this class?
I didn't see any usage of the ```this.compareOptions```
But maybe I missed something :)

js/data/data_source/data_source.js
```diff
@@ -498,6 +497,10 @@ export const DataSource = Class.inherit({
         const operationId = this._operationManager.add(deferred);
         const storeLoadOptions = this._createStoreLoadOptions();
 
+        if(this._store && !isEmptyObject(storeLoadOptions?.compareOptions)) {
+            this._store._compareOptions = { ...this._store._compareOptions, ...storeLoadOptions.compareOptions };
+        }
```
williamvinogradov: Minor comment.

Nothing scary will happen if we spread the empty object :)

Let's simplify these lines:
```suggestion
        if(this._store) {
            this._store._compareOptions = { ...this._store._compareOptions, ...storeLoadOptions?.compareOptions };
        }
```

js/data/store_helper.js
```diff
@@ -37,6 +37,10 @@ function queryByOptions(query, options, isCountQuery) {
 
     const filter = options.filter;
 
+    if(options?.compareOptions) {
+        query.setCompareOptions?.(options?.compareOptions);
```
williamvinogradov: We don't need second ```?``` here:
```suggestion
        query.setCompareOptions?.(options.compareOptions);
```

js/core/utils/data.js
pomahtri: Better to just use default value
```
function(value, caseSensitive, options = {}) {
```

js/ui/grid_core/ui.grid_core.data_source_adapter.js
williamvinogradov: No, we cannot use the spread operator here, because ```extend``` ignores keys without value
E.g { a: undefined }

js/ui/grid_core/ui.grid_core.data_source_adapter.js
williamvinogradov: Discussed, this is ok because remote operations don't handle by the local store :)

js/core/utils/data.js
GoodDayForSurf: yep, Let it be for now

js/data/data_source/data_source.js
GoodDayForSurf: i think better no reassign store._compareOptions if it's not necessary

js/data/abstract_store.js
GoodDayForSurf: if so we get two spread operations and reassign when it not necessary

js/data/array_query.js
GoodDayForSurf: we need compareOptions  as props of current iterator

js/data/array_query.js
GoodDayForSurf: ok

js/data/array_query.js
GoodDayForSurf: we need it. In other places new iterators wrap others, so we need to store compareOptions in those chains.

Zedwag (dismissed):

js/data/array_query.js
```diff
@@ -98,9 +98,15 @@ const MapIterator = WrappedIterator.inherit({
     }
 });
 
-const defaultCompare = function(xValue, yValue) {
-    xValue = toComparable(xValue);
-    yValue = toComparable(yValue);
+const defaultCompare = function(xValue, yValue, options) {
+    if(typeof xValue === 'string' && typeof yValue === 'string' && (options?.locale || options?.collateOptions)) {
```
dxvladislavvolkov: Use isString method from ../../core/utils/type

js/data/index.d.ts
```diff
@@ -42,7 +42,14 @@ export type SelectDescriptor<T> = string | Array<string> | ((source: T) => any);
...
  */
 export type FilterDescriptor = any;
-
+/**
+ * @docid
+ * @public
+ */
+export type LangParams = {
+  locale: string;
+  collateOptions?: Intl.CollatorOptions;
```
vladaskorohodova:
```suggestion
  collatorOptions?: Intl.CollatorOptions;
```
I thought we agreed with PMs that 'collator' is better than 'collate'

GoodDayForSurf: ok, i fixed it

js/data/array_query.js
GoodDayForSurf: fixed

mpreyskurantov (dismissed):

dxvladislavvolkov (approved):

pomahtri (approved):

mpreyskurantov (approved):

Zedwag (approved):

js/core/utils/data.js
```diff
@@ -191,7 +191,13 @@ export const toComparable = function(value, caseSensitive) {
     }
 
     if(!caseSensitive && typeof value === 'string') {
-        return value.toLowerCase();
+        if(options?.collatorOptions?.sensitivity === 'base') {
```
williamvinogradov, line 194: Minor comment

Let's remove the first redundant ```?```
(because options always defined and equals ```{}``` by default)

```suggestion
        if(options.collatorOptions?.sensitivity === 'base') {
```

js/core/utils/data.js
```diff
@@ -191,7 +191,13 @@ export const toComparable = function(value, caseSensitive) {
...
+            const REMOVE_DIACRITICAL_MARKS_REGEXP = /[\u0300-\u036f]/g;
+
+            value = value.normalize('NFD').replace(REMOVE_DIACRITICAL_MARKS_REGEXP, '');
+        }
+
+        return options?.locale ? value.toLocaleLowerCase(options.locale) : value.toLowerCase();
```
williamvinogradov, line 200: Minor comment

Let's remove the first redundant ```?```
(because options always defined and equals ```{}``` by default)
```suggestion
        return options.locale ? value.toLocaleLowerCase(options.locale) : value.toLowerCase();
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -560,7 +560,8 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
             recursiveExpansion: this.option('expandNodesRecursive'),
             selectionRequired: this.option('selectionRequired'),
             dataType: this.option('dataStructure'),
-            sort: this._dataSource && this._dataSource.sort()
+            sort: this._dataSource && this._dataSource.sort(),
```
williamvinogradov, line 563: Minor comment

Can we use the ```?``` here?
```suggestion
            sort: this._dataSource?.sort(),
```

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -560,7 +560,8 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
...
+            langParams: this._dataSource?.loadOptions?.()?.langParams
```
williamvinogradov, line 564: Minor comment

Let's move it to a separate variable:
e.g
```javascript
_getDataAdapterOptions: function() {
  const loadOptions  = this._dataSource?.loadOptions?.();
  const langParams = loadOptions?.langParams;
  // or const { langParams } = loadOptions || {};
  return {
    // some options.
    langParams,
  };
}
```

williamvinogradov (approved):

PR #24352, opened 2023-04-19
Fix CSP issues in ui tests

Lykoi18:


testing/helpers/includeThemesLinks.js
```diff
@@ -0,0 +1,9 @@
+const themesList = ['generic.light', 'material.blue.light'];
+
+themesList.forEach(theme => {
+    const link = document.createElement('link');
+    link.setAttribute('rel', 'dx-theme');
+    link.setAttribute('data-theme', theme);
+    link.setAttribute('href', SystemJS.normalizeSync(theme.replace(/\./g, '_') + '.css'));
+    document.head.appendChild(link);
+});
```
timbset, line 9: Can't we use `window.includeThemesLinks();` here to prevent code duplication?

testing/tests/DevExpress.ui/themes.tests.js
```diff
@@ -5,7 +5,9 @@ import themes from 'ui/themes';
 import viewPortUtils from 'core/utils/view_port';
 
 const viewPortChanged = viewPortUtils.changeCallback;
-import knownCssFiles from '/themes-test/get-css-files-list/!json';
+// import knownCssFiles from '/themes-test/get-css-files-list/!json';
+
+const knownCssFiles = [];
```
timbset: Why is this commented and used nowhere?

timbset (changes requested):

testing/helpers/includeThemesLinks.js
Lykoi18: We cannot use the includeThemesLinks function from window because the theme links must be set before initializing themes. However, since in SystemJS modules imports are executed before the rest of the code, we had to move this function to a separate file that is imported before the themes module.

Since this function is also used in another [test](https://github.com/DevExpress/DevExtreme/blob/23_1/testing/tests/DevExpress.core/devices.tests.js#L1), we cannot currently remove the implementation of this function from the window in order to avoid duplicating the code, but it will definitely be done later.

timbset: Is it a problem to make this change for that test too? I think it's better to do it right away.

Lykoi18: We can update that test in this PR.

timbset (dismissed):

mpreyskurantov (approved):

timbset (approved):

PR #24387, opened 2023-04-21
Form: fix update readOnly option for editors (T1157373)

Krijovnick:


js/ui/form/ui.form.layout_manager.js
```diff
@@ -708,14 +708,15 @@ const LayoutManager = Widget.inherit({
     },
 
     _createComponent: function($editor, type, editorOptions) {
-        const that = this;
         const readOnlyState = this.option('readOnly');
-        const instance = that.callBase($editor, type, editorOptions);
-
-        readOnlyState && instance.option('readOnly', readOnlyState);
+        const hasEditorReadOnly = isDefined(editorOptions.readOnly);
```
ksercs: I think we should we keep `editorOptions.readOnly` value even if it's specified as an `undefined` or `null` while it casts to `false` in standalone editors.
I guess it should be smth like this
```suggestion
        const hasEditorReadOnly = Object.hasOwn(editorOptions, 'readOnly');
```

testing/tests/DevExpress.ui.widgets.form/formLayoutManager.markup.tests.js
```diff
@@ -3027,3 +2991,92 @@ QUnit.module('Supported editors', () => {
...
+                editorOptions: { readOnly: true }
+            }))
+        });
+
+        checkSupportedEditors((editor, className) => {
+            assert.ok($testContainer.find(`.${FIELD_ITEM_CLASS} .${className}`).hasClass(READONLY_STATE_CLASS), `${editor}: editor is read only`);
+        });
+    });
+
+    test('editors should not be read only when readOnly option is enabled in widget, but in the editorOptions it is set to false', function(assert) {
```
ksercs: Let's concretise?
```suggestion
    test('editors should not be read only when readOnly option is enabled in form, but in the editorOptions it is set to false', function(assert) {
```

testing/tests/DevExpress.ui.widgets.form/formLayoutManager.markup.tests.js
```diff
@@ -3027,3 +2991,92 @@ QUnit.module('Supported editors', () => {
...
+                editorType: supportedEditor,
+            }))
+        });
+
+        checkSupportedEditors((editor, className) => {
+            assert.ok($testContainer.find(`.${FIELD_ITEM_CLASS} .${className}`).hasClass(READONLY_STATE_CLASS), `${editor}: editor is read only`);
+        });
+    });
+
+    test('editors should change their readonly state after change readOnly option in widget', function(assert) {
```
ksercs: am i right?
```suggestion
    test('editors should change their readonly state after change readOnly option in form if editorOptions.readOnly is not specified', function(assert) {
```

js/ui/form/ui.form.layout_manager.js
```diff
@@ -708,14 +708,15 @@ const LayoutManager = Widget.inherit({
...
+        const instance = this.callBase($editor, type, {
+            ...editorOptions,
+            readOnly: !hasEditorReadOnly ? readOnlyState : editorOptions.readOnly
+        });
 
-        that.on('optionChanged', function(args) {
-            if(args.name === 'readOnly' && !isDefined(editorOptions.readOnly)) {
+        this.on('optionChanged', function(args) {
+            if(args.name === 'readOnly' && !hasEditorReadOnly) {
```
ksercs, line 726: It does not work for `item[0].editorOptions.readonly` runtime change while we save `hasEditorReadOnly` on editor initialization. Can we try to track editors `readOnly` runtime change also? Smth like this
```javascript
        let isChangeByForm = false;
        instance.on('optionChanged', (args) => {
            if(args.name === 'readOnly' && !isChangeByForm) {
                hasEditorReadOnly = true;
            }
        });
        this.on('optionChanged', function(args) {
            if(args.name === 'readOnly' && !hasEditorReadOnly) {
                isChangeByForm = true;
                instance.option(args.name, args.value);
                isChangeByForm = false;
            }
        });
```

testing/tests/DevExpress.ui.widgets.form/formLayoutManager.markup.tests.js
```diff
@@ -3027,3 +2991,92 @@ QUnit.module('Supported editors', () => {
...
+            assert.notOk($testContainer.find(`.${FIELD_ITEM_CLASS} .${className}`).hasClass(READONLY_STATE_CLASS), `${editor}: editor is not read only`);
+        });
+
+        $testContainer.dxLayoutManager('instance').option('readOnly', true);
+
+        checkSupportedEditors((editor, className) => {
+            assert.notOk($testContainer.find(`.${FIELD_ITEM_CLASS} .${className}`).hasClass(READONLY_STATE_CLASS), `${editor}: editor is not read only`);
+        });
+    });
+});
```
ksercs, line 3125: Does it work ok for a `disabled` prop?

Krijovnick: The `disabled` state work different way. Disabled form has class on the whole widget and not pass disabled state to the editor. So you can change editor's disabled state apart of the form. In detail: if the form has disabled state, then all in the form looks like disabled. You can change disabled state of editor but visually it don't changed, inside state is changed. If turn off disabled state of the form -> editor will be with the state that was apply last time in the editorOptions.

js/ui/form/ui.form.layout_manager.js
Krijovnick: If `editorOptions.readOnly` is undefined, it is not will work as null or false with this changes. And I think we should not to make behavior with `undefined` equal to null and false values.

testing/tests/DevExpress.ui.widgets.form/formLayoutManager.markup.tests.js
```diff
@@ -3027,3 +2991,122 @@ QUnit.module('Supported editors', () => {
...
+                editorOptions: { readOnly: null }
+            }))
+        });
+
+        checkSupportedEditors((editor, className) => {
+            assert.notOk($testContainer.find(`.${FIELD_ITEM_CLASS} .${className}`).hasClass(READONLY_STATE_CLASS), `${editor}: editor is not read only`);
+        });
+    });
+
+    test('editors should has readonly state if editorOptions.readOnly is undefined', function(assert) {
```
ksercs:
```suggestion
    test('editors should has readonly state if editorOptions.readOnly is undefined and form readOnly is true', function(assert) {
```

testing/tests/DevExpress.ui.widgets.form/formLayoutManager.markup.tests.js
```diff
@@ -3027,3 +2991,122 @@ QUnit.module('Supported editors', () => {
...
+        const $testContainer = $('#container').dxLayoutManager({
+            items: supportedEditors.map(supportedEditor => ({
+                name: supportedEditor,
+                editorType: supportedEditor,
+                editorOptions: { readOnly: true }
+            }))
+        });
+
+        checkSupportedEditors((editor, className) => {
+            assert.ok($testContainer.find(`.${FIELD_ITEM_CLASS} .${className}`).hasClass(READONLY_STATE_CLASS), `${editor}: editor is read only`);
```
ksercs: I suggest u move this construction to a separate function to reduce copypaste
```javascript
            const isEditorReadOnly = ($container, editorClassName) => $container.find(`.${FIELD_ITEM_CLASS} .${editorClassName}`).hasClass(READONLY_STATE_CLASS);
```

testing/tests/DevExpress.ui.widgets.form/form.markup.tests.js
```diff
@@ -226,11 +227,27 @@ QUnit.module('Form', () => {
...
-        assert.notOk($formContainer.find('.' + FIELD_ITEM_CLASS + ' .dx-texteditor').hasClass('dx-state-readonly'), 'editor isn\'t read only');
+        assert.notOk($formContainer.find('.' + FIELD_ITEM_CLASS + ' .dx-texteditor').hasClass(READONLY_STATE_CLASS), 'editor isn\'t read only');
 
         $formContainer.dxForm('instance').option('readOnly', true);
 
-        assert.ok($formContainer.find('.' + FIELD_ITEM_CLASS + ' .dx-texteditor').hasClass('dx-state-readonly'), 'editor is read only');
+        assert.ok($formContainer.find('.' + FIELD_ITEM_CLASS + ' .dx-texteditor').hasClass(READONLY_STATE_CLASS), 'editor is read only');
+    });
+
+    test('editor should not change readonly state after update readOnly option in editor and then in form', function(assert) {
```
ksercs: It should be changed after editorOptions.readOnly change, but not after form readOnly change
```suggestion
    test('editor should not change readonly state after form readOnly option change if editorOptions.readOnly was updated before', function(assert) {
```

js/ui/form/ui.form.layout_manager.js
ksercs: okay let's leave it now

testing/tests/DevExpress.ui.widgets.form/form.markup.tests.js
```diff
@@ -226,11 +227,27 @@ QUnit.module('Form', () => {
             ]
         });
 
-        assert.notOk($formContainer.find('.' + FIELD_ITEM_CLASS + ' .dx-texteditor').hasClass('dx-state-readonly'), 'editor isn\'t read only');
+        assert.notOk($formContainer.find('.' + FIELD_ITEM_CLASS + ' .dx-texteditor').hasClass(READONLY_STATE_CLASS), 'editor isn\'t read only');
 
         $formContainer.dxForm('instance').option('readOnly', true);
 
-        assert.ok($formContainer.find('.' + FIELD_ITEM_CLASS + ' .dx-texteditor').hasClass('dx-state-readonly'), 'editor is read only');
+        assert.ok($formContainer.find('.' + FIELD_ITEM_CLASS + ' .dx-texteditor').hasClass(READONLY_STATE_CLASS), 'editor is read only');
```
marker-dao: Let's use constant for `dx-texteditor` class and another same cases.

marker-dao: Let's replace [Concatenating strings](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/First_steps/Strings#concatenating_strings) with [Template strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) in all cases.

testing/tests/DevExpress.ui.widgets.form/formLayoutManager.markup.tests.js
```diff
@@ -3027,3 +2991,123 @@ QUnit.module('Supported editors', () => {
         }
     });
 });
+
+QUnit.module('ReadOnly option', () => {
+    const getEditorClassName = editorName => `dx-${editorName.substr(2, editorName.length - 1).toLowerCase()}`;
+    const checkSupportedEditors = callBack => supportedEditors.forEach(supportedEditor => callBack(supportedEditor, getEditorClassName(supportedEditor)));
```
marker-dao: Let's create a non one line functions, because your functions is too long: [Avoid having lines of code that are longer than 100 characters](https://airbnb.io/javascript/#whitespace--max-len).

ksercs (approved):

marker-dao (approved):

PR #24388, opened 2023-04-21
DropDownButton: Remove excess aria-expanded attributes

marker-dao:


js/ui/drop_down_button.js
```diff
@@ -460,26 +461,22 @@ const DropDownButton = Widget.inherit({
...
-            this.setAria({
-                expanded: value,
-                owns: value ? this._popupContentId : undefined,
-            },
-            $(ariaElement));
-        });
-    },
-
-    _setAriaHasPopup() {
+    _getPopupButton() {
```
ksercs:
```suggestion
    _getPopupToggleButton() {
```

testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
```diff
@@ -2641,9 +2634,32 @@ QUnit.module('Accessibility', {
...
+        const instance = this.createInstance();
+
+        assert.strictEqual(this.$element.attr('aria-owns'), undefined);
+
+        instance.open();
+
+        const popupId = $(`.${POPUP_CONTENT_CLASS}`).attr('id');
+
+        assert.strictEqual(this.$element.attr('aria-owns'), popupId);
     });
```
ksercs, line 2672: let's add a specific test that toggle button is focused after `up/down` arrow is pressed if toggle button is focused

testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
```diff
@@ -2641,9 +2634,32 @@ QUnit.module('Accessibility', {
...
+    QUnit.test('check aria-owns attr for element', function(assert) {
+        const instance = this.createInstance();
+
+        assert.strictEqual(this.$element.attr('aria-owns'), undefined);
+
+        instance.open();
+
+        const popupId = $(`.${POPUP_CONTENT_CLASS}`).attr('id');
+
+        assert.strictEqual(this.$element.attr('aria-owns'), popupId);
```
ksercs, line 2671: seems like we still have action button focused after `up/down` arrow is pressed if action button is focused. Seems like screen reader does not notify a user about popup opening in this case

js/ui/drop_down_button.js
```diff
@@ -490,19 +487,19 @@ const DropDownButton = Widget.inherit({
 
         this._buttonGroup = this._createComponent($buttonGroup, ButtonGroup, this._buttonGroupOptions());
 
-        const buttonElements = this._buttonGroup.$element().find('.dx-button').toArray();
+        this._$buttonElements = this._buttonGroup.$element().find('.dx-button');
```
ksercs: if u use a cache we should clear it on a widget clean

js/ui/drop_down_button.js
```diff
@@ -490,19 +487,19 @@ const DropDownButton = Widget.inherit({
...
 
-        this._ariaExpandedElements = [ ...buttonElements, this.$element() ];
-
-        this._buttonGroup.registerKeyHandler('downArrow', this._upDownKeyHandler.bind(this));
-        this._buttonGroup.registerKeyHandler('tab', this._tabHandler.bind(this));
-        this._buttonGroup.registerKeyHandler('upArrow', this._upDownKeyHandler.bind(this));
-        this._buttonGroup.registerKeyHandler('escape', this._escHandler.bind(this));
+        this._buttonGroup._buttonsCollection.registerKeyHandler('downArrow', this._upDownKeyHandler.bind(this));
```
ksercs: `_buttonsCollection` is a private api of the buttonGroup -> it's not a good idea to use it outside of the buttonGroup.
Cannot we don't do anything with `esc` and `tab` and just forcibly focus $toggleButton in `_upDownKeyHandler`?

js/ui/drop_down_button.js
marker-dao: [31e0ee6495f20aa1900f77926ab0596130ff3325].

js/ui/drop_down_button.js
marker-dao: [31e0ee6495f20aa1900f77926ab0596130ff3325].

testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
marker-dao: [39eacc4b2b09275b18a304226d1fdaeb1cd48b41].

testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
marker-dao: Yes. I have set aria-* attrs at both buttons. This is the best solution for these cases. Fixed here: [8cb82439fa22a0e23607aba9b0eb67dcb3a3ee80].

js/ui/drop_down_button.js
marker-dao: We can don't do anything with esc and tab key handlers.

But forcibly focus $toggleButton in _upDownKeyHandler is so strange solution for me because focus will be changed twice: set to the action button in buttonsCollection handler, then set back to the toggle button in buttonGroup handler.

Moreover, I think it's right to override the parent behavior in the current component, because this avoids side effects such as setting focus to the first element.

Also I think you are right because using a private api is discouraged. But there is no public api in the current component. And the private api is used by us everywhere in our library.

if you are still against using _buttonsCollection so let's have a voice discussion.

ksercs: We discussed it in voice. Decided not to revert this changed and leave the previous behavior.

js/ui/drop_down_button.js
```diff
@@ -432,7 +432,8 @@ const DropDownButton = Widget.inherit({
     _popupHidingHandler() {
         this.option('opened', false);
 
-        this._setAriaExpanded(false);
+        this._setPopupButtonsAria(false);
+        this._setElementAria(false);
```
ksercs: Let's create a single method to update aria attributes to reduce a copypaste

testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
ksercs: decided to not change this behavior

js/ui/drop_down_button.js
marker-dao: Reverted here: [08b3ae3acf750d1245b3bd08f0c6e468113ec5c1].

js/ui/drop_down_button.js
marker-dao: Refvctored here: [08b3ae3acf750d1245b3bd08f0c6e468113ec5c1].

ksercs (dismissed):

ksercs (approved):

Krijovnick (approved):

PR #24396, opened 2023-04-23
Treeview: fix 'selectAll' item focus (T1157890)

AlexanderMoiseev:


js/ui/tree_view/ui.tree_view.search.js
```diff
@@ -58,6 +58,10 @@ const TreeViewSearch = TreeViewBase.inherit(searchBoxMixin).inherit({
         });
     },
 
+    _getRootNodeContainer: function() {
```
ksercs: Why "root"?
```suggestion
    _getNodeContainer: function() {
```

ksercs (dismissed):

Krijovnick (dismissed):

js/ui/tree_view/ui.tree_view.search.js
AlexanderMoiseev: its just that it can contain other node-containers inside, so this is like main container, but i'll rename if you dont like it)

Krijovnick (approved):

PR #24434, opened 2023-04-26
DataGrid(T1161941): Header Filter cannot switch options

williamvinogradov:
See the SC ticket: [T1161941](https://supportcenter.devexpress.com/ticket/details/T1161941)


const314 (dismissed):

js/core/utils/common.js
```diff
@@ -262,58 +262,60 @@ export const grep = function(elements, checkFunction, invert) {
...
-        if(!(propertyName in object1)) {
-            return false;
-        }
+    if(keys1.length !== keys2.length) {
+        return false;
     }
 
-    return true;
+    return !keys1.some((key) =>
+        !Object.prototype.hasOwnProperty.call(object2, key)
```
pomahtri: According to [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys), `Object.keys` already returns only own properties, so we can remove `hasOwnProperty`

js/core/utils/common.js
```diff
@@ -262,58 +262,59 @@ export const grep = function(elements, checkFunction, invert) {
...
-    for(const propertyName in object2) {
-        if(!(propertyName in object1)) {
-            return false;
-        }
+    if(keys1.length !== keys2.length) {
+        return false;
     }
 
-    return true;
+    const keys2Set = keys2.reduce((set, key) => set.add(key), new Set());
```
pomahtri, line 284:
```suggestion
    const keys2Set =new Set(keys2);
```

pomahtri:
```suggestion
    const keys2Set = new Set(keys2);
```

js/core/utils/common.js
williamvinogradov: Yup, I agree with you
Let's remove it

pomahtri (dismissed):

pomahtri (approved):

sergepilipchuk (approved):

const314 (approved):

PR #24451, opened 2023-04-27
Localization: fix getting of a day period in arabic locale (T1162346)

ksercs:
This is a spike of the fix. It should be improved by @dxvladislavvolkov


ksercs:
will be fixed in context of https://github.com/DevExpress/DevExtreme/pull/24564

PR #24484, opened 2023-05-02
HtmlEditor: fix error in console when uploadDirectory option is not specified (T1161285)

Krijovnick:


js/ui/html_editor/modules/imageUpload.js
```diff
@@ -71,11 +71,7 @@ if(Quill) {
 
         _onUploaded(data) {
             const { index: pasteIndex } = this.quill.getSelection() ?? { index: this.quill.getLength() };
-
-            const imageUrl = correctSlashesInUrl(this.options.uploadDirectory) + data.file.name;
-            urlUpload(this.quill, pasteIndex, { src: imageUrl });
-
-            this.quill.setSelection(pasteIndex + 1, 0);
+            serverUpload(this.options.uploadDirectory, data.file.name, this.quill, pasteIndex);
```
ksercs, line 74: After your change we don't call `setSelection` with third parameter = "api"`anymore. Is it expected?

js/ui/html_editor/utils/image_uploader_helper.js
```diff
@@ -533,3 +528,11 @@ export function urlUpload(quill, index, attributes) {
     quill.insertEmbed(index, 'extendedImage', attributes, USER_ACTION);
     quill.setSelection(index + 1, 0, USER_ACTION);
 }
+
+export function serverUpload(url, fileName, quill, pasteIndex) {
+    if(url) {
```
ksercs, line 528: Earlier we had a condition for `!this.useBase64` here. Does this case work ok now?

js/ui/html_editor/utils/image_uploader_helper.js
```diff
@@ -477,7 +472,7 @@ class FileStrategy extends BaseStrategy {
                 this.data = data;
                 this.onFileSelected();
             },
-            onUploaded: e => this.serverUpload(e)
+            onUploaded: e => this.onUploaded(e)
```
ksercs, line 470: We don't call `this.closeDialogPopup(data);` anymore. Is it ok?

js/ui/html_editor/modules/imageUpload.js
Krijovnick: `setSelection` called inside `serverUpload` method. But there is no `api` parameter. I don't see this parameter anywere...

js/ui/html_editor/utils/image_uploader_helper.js
Krijovnick: https://github.com/DevExpress/DevExtreme/blob/fb36a6838af35789e07094f0a4e13a2ee3d9e6c3/js/ui/html_editor/utils/image_uploader_helper.js#L413 - this flag can be only in 'base64' mode, not in server mode. And this method called only in server mode

js/ui/html_editor/utils/image_uploader_helper.js
Krijovnick: There is I wrote the test, that dialog is invisible.

js/ui/html_editor/utils/image_uploader_helper.js
ksercs: Can not it be `true` if we check `use base64` checkbox in file uploader?

js/ui/html_editor/utils/image_uploader_helper.js
ksercs: I see call of the closing only for base64. Please, test it with hands just in case :)

js/ui/html_editor/modules/imageUpload.js
ksercs: third parameter is `api` by default

js/ui/html_editor/utils/image_uploader_helper.js
Krijovnick: I tested with hands, it work as expected

js/ui/html_editor/utils/image_uploader_helper.js
Krijovnick: When we check checkbox then it uses base64 'branch' in the code and serverUpload method is not called

js/ui/html_editor/modules/imageUpload.js
Krijovnick: According to the https://quilljs.com/docs/api/#events description:

> "Changes may occur through an API but as long as they originate from the user, the provided source should still be "user""

Thus, when image is dropped by user there should be 'user' display in events, not 'api'. And new changes more correct in this case I think

ksercs (approved):

alexanderPolosatov (approved):

PR #24503, opened 2023-05-03
List: item container should be a focus and aria target for loadpanel not to be placed inside a listbox

ksercs:


js/ui/list/ui.list.base.js
```diff
@@ -656,7 +655,22 @@ export const ListBase = CollectionWidget.inherit({
         this.callBase();
         this.option('useInkRipple') && this._renderInkRipple();
 
-        this.setAria('role', this.option('_listAttributes').role);
+        this.setAria('role', 'group', this.$element());
+        this.setAria('roledescription', 'list', this.$element());
```
marker-dao:
```suggestion
        const elementAria = {
             'role': 'group',
             'roledescription': 'list',
        }
        this.setAria(elementAria, this.$element());
```

ksercs: done

marker-dao (approved):

PR #24510, opened 2023-05-04
ButtonGroup: change aria-selected to aria-pressed for buttons

AlexanderMoiseev:


js/ui/context_menu/ui.menu_base.js
```diff
@@ -534,17 +534,17 @@ class MenuBase extends HierarchicalCollectionWidget {
         const selectedIndex = this._dataAdapter.getSelectedNodesKeys();
 
         if(!selectedIndex.length || !this._selectedGetter(args.itemData) || !this._isItemSelectable(args.itemData)) {
-            this._setAriaSelected($itemElement, 'false');
+            this._setAriaSelectionAttribute($itemElement, 'false');
```
Krijovnick, line 537: How it works if `_setAriaSelectionAttribute` is empty function? Maybe this code is not need?

js/ui/radio_group/radio_group.js
```diff
@@ -107,7 +107,7 @@ class RadioCollection extends CollectionWidget {
         return this._itemContainer().children(this._itemSelector());
     }
 
-    _setAriaSelected() {}
+    _setAriaSelectionAttribute() {}
```
Krijovnick, line 110: It can be rewrite like `_setAriaSelectionAttribute: noop` everywhere

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -719,7 +719,7 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         }
     },
 
-    _setAriaSelected: function() {},
+    _setAriaSelectionAttribute: function() {},
```
Krijovnick: Only tree_view.base and ui.menu_base inherit from `HierarchicalCollectionWidget`, so `_setAriaSelectionAttribute: noop` can be moved to the `HierarchicalCollectionWidget` and can be removed from this files

js/ui/context_menu/ui.menu_base.js
AlexanderMoiseev: this is just a renaming ) also im not sure this code is not needed, if you override _setAriaSelectionAttribute it may be useful. I dont say this is the case but i didn't check, the intention was only to rename the function, not to refactor existing code.

js/ui/radio_group/radio_group.js
AlexanderMoiseev: i did not refactor anything here, just renamed the function

js/ui/tree_view/ui.tree_view.base.js
AlexanderMoiseev: i did not refactor anything here, just renamed the existing function.  anyway, if we are to do so, i dont think it should be done in this PR, this has nothing to do with refactoring

js/ui/context_menu/ui.menu_base.js
Krijovnick: It can be done in the separate pull request then. I think if we see some strange code near we should fix it or to understand how it work. Especially when it not need many time to research. And this case seems can be checked in 10 minutes and refactor in 1 minute. And the code will looks more better.

testing/tests/DevExpress.ui.widgets/buttonGroup.tests.js
```diff
@@ -233,6 +233,33 @@ QUnit.module('option changed', {
...
+
+    QUnit.test('button has aria-pressed=false attribute if unselected', function(assert) {
+        $('#widget').dxButtonGroup({ items: [{ id: '1' }, { id: '2' }] });
+        const $firstButton = $(`.${BUTTON_CLASS}`).eq(0);
+        const $secondButton = $(`.${BUTTON_CLASS}`).eq(1);
+
+        eventsEngine.trigger($firstButton, 'dxclick');
+        eventsEngine.trigger($secondButton, 'dxclick');
+
+        assert.strictEqual($firstButton.attr('aria-pressed'), 'false');
```
EugeniyKiyashko, line 260: Let's add additional assert
```
assert.strictEqual($secondButton.attr('aria-pressed'), 'true');
```

EugeniyKiyashko (dismissed):

js/ui/context_menu/ui.menu_base.js
```diff
@@ -132,7 +132,7 @@ class MenuBase extends HierarchicalCollectionWidget {
         return ITEM_CLASS;
     }
 
-    _setAriaSelected() {}
+    _setAriaSelectionAttribute() { return noop; }
```
Krijovnick: We should not return function, we need to return void. Thus, there should be:
_setAriaSelectionAttribute = noop;

js/ui/radio_group/radio_group.js
```diff
@@ -107,7 +107,7 @@ class RadioCollection extends CollectionWidget {
         return this._itemContainer().children(this._itemSelector());
     }
 
-    _setAriaSelected() {}
+    _setAriaSelectionAttribute() { return noop; }
```
Krijovnick: There also should not returns function

js/ui/tree_view/ui.tree_view.base.js
```diff
@@ -719,7 +719,7 @@ const TreeViewBase = HierarchicalCollectionWidget.inherit({
         }
     },
 
-    _setAriaSelected: function() {},
+    _setAriaSelectionAttribute() { return noop; },
```
Krijovnick: _setAriaSelectionAttribute: noop

Krijovnick (changes requested):

Krijovnick (approved):

EugeniyKiyashko (approved):

PR #24521, opened 2023-05-04
Form: provide information about parent groups on focus

Krijovnick:


js/ui/form/ui.form.js
```diff
@@ -525,21 +525,23 @@ const Form = Widget.inherit({
...
+    _itemGroupTemplate: function(item, options, $container) {
+        const id = options.editorOptions.inputAttr.id;
         const $group = $('<div>')
             .toggleClass(FORM_GROUP_WITH_CAPTION_CLASS, isDefined(item.caption) && item.caption.length)
             .addClass(FORM_GROUP_CLASS)
             .appendTo($container);
 
-        $($container).parent().addClass(FIELD_ITEM_CONTENT_HAS_GROUP_CLASS);
+        this.setAria('role', 'group', $group);
+        this.setAria('labelledby', id, $group);
```
marker-dao:
```suggestion
        const groupAria = {
            role: 'group',
            labelledby: id,
        }

        this.setAria(groupAria, $group)
```

testing/tests/DevExpress.ui.widgets.form/form.markup.tests.js
```diff
@@ -1430,42 +1430,42 @@ QUnit.module('Grouping', () => {
         const items = form._testResultItems;
 
         items[0].template.render({
-            model: {},
+            model: { editorOptions: { inputAttr: {} } },
             container: template
```
ksercs, line 1434: Why do we need all of this changed in tests?
Shouldn't we use optional chaining in the code https://github.com/DevExpress/DevExtreme/pull/24521/files#diff-aa52d76cd61878684b8de01d6631bccd4cc5ec63a16e0981048f6382a33741e6R529?

Krijovnick: editorOptions is always exist and optional chaining has not any sense in real world. This test run render method with empty model, but in the reality model is not empty

ksercs (approved):

PR #24531, opened 2023-05-04
DataGrid: Fix duplicate headers when headerCellTemplate is given (T1155453)

Alyar666:
See https://devexpress.com/issue=T1155453


testing/testcafe/model/dataGrid/index.ts
```diff
@@ -458,6 +458,19 @@ export default class DataGrid extends Widget {
     )();
   }
 
+  apiCollapseAllGroups(): Promise<void> {
+    const { getInstance } = this;
+
+    return ClientFunction(
+      () => (getInstance() as any).option('grouping.autoExpandAll', false),
```
pomahtri: minor: can we use DataGrid type from our .d.ts instead of any here?

pomahtri (dismissed):

js/ui/grid_core/ui.grid_core.column_fixing.js
```diff
@@ -243,6 +243,14 @@ const baseFixedColumns = {
         return $cell;
     },
 
+    _getContent: function(isFixedTableRendering) {
+        if(isFixedTableRendering) {
+            return this._fixedTableElement?.parent();
+        }
+
+        return this.callBase.apply(this, arguments);
+    },
```
williamvinogradov: Minor comment.

Maybe the ternary operator will be better here?
```suggestion
    _getContent: function(isFixedTableRendering) {
       return isFixedTableRendering
         ? this._fixedTableElement?.parent()
         : this.callBase.apply(this, arguments);
    },
```

js/ui/grid_core/ui.grid_core.columns_view.js
```diff
@@ -891,6 +891,16 @@ const columnsViewMembers = {
...
     },
 
+    _getContent: function() {
+        return this._tableElement?.parent();
+    },
+
+    _removeContent: function(isFixedTableRendering) {
+        const $scrollContainer = this._getContent(isFixedTableRendering);
+
+        $scrollContainer?.length && $scrollContainer.remove();
```
williamvinogradov: Minor comment.

Let's make this more obvious and add the ```if``` condition.
Also, it will help us in migration to TS because TS linter doesn't allow constructions like this :)

```suggestion
        if($scrollContainer?.length) {
          $scrollContainer.remove();
        }
```

williamvinogradov (dismissed):

williamvinogradov (approved):

PR #24543, opened 2023-05-05
Widget: aria-hidden should not be set if visible=false while "display:none" is used

ksercs:
```markdown
aria-hidden="true" should not be added when:

The HTML [hidden](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) attribute is present
The element or the element's ancestor is hidden with [display: none](https://developer.mozilla.org/en-US/docs/Web/CSS/display)
The element or the element's ancestor is hidden with [visibility: hidden](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility)
```
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden

### DIFF
--- a/testing/tests/DevExpress.ui/widget.tests.js
+++ b/testing/tests/DevExpress.ui/widget.tests.js
@@ -1541,6 +1541,13 @@ QUnit.module('aria accessibility', {}, () => {
         assert.equal($element.attr('aria-disabled'), undefined, 'attribute test on option change');
     });
 
+    QUnit.test('aria-hidden', function(assert) {
+        const $element = $('#widget').dxWidget({ visible: false });
+
+        // NOTE: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden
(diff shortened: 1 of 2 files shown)


Raushen (dismissed):

testing/tests/DevExpress.ui/widget.tests.js
```diff
@@ -1541,16 +1541,6 @@ QUnit.module('aria accessibility', {}, () => {
...
     });
 
-    QUnit.test('aria-hidden', function(assert) {
-        const $element = $('#widget').dxWidget({ visible: false });
-        const instance = $element.dxWidget('instance');
-
-        assert.equal($element.attr('aria-hidden'), 'true', 'attribute test on init');
-
-        instance.option('visible', true);
-        assert.equal($element.attr('aria-hidden'), undefined, 'attribute test on option change');
```
marker-dao, line 1551: Let's add test for non set aria-hidden.

assert.equal($element.attr('aria-hidden'), undefined, 'aria-hidden should not set');

const314 (approved):

Raushen (approved):

PR #24547, opened 2023-05-08
List: provide information about parent groups on focus

Krijovnick:


js/ui/list/ui.list.base.js
```diff
@@ -746,8 +747,17 @@ export const ListBase = CollectionWidget.inherit({
             .addClass(LIST_GROUP_CLASS)
             .appendTo(this._itemContainer());
 
+        const id = `dx-${new Guid().toString()}`;
```
marker-dao, line 750: .toString() shouldn’t be used here.

Krijovnick: It's hard to read and understand without .toString()

ksercs (approved):

alexanderPolosatov (approved):

EugeniyKiyashko (approved):

PR #24559, opened 2023-05-10
DataGrid - The "Uncaught TypeError: Cannot read properties of undefined (reading 'y')" error occurs when exporting to PDF if a pdfCell's wordWrapEnabled property is set to true (T1164070)

Raushen:
See https://supportcenter.devexpress.com/ticket/details/t1164070/datagrid-the-uncaught-typeerror-cannot-read-properties-of-undefined-reading-y-error


js/exporter/jspdf/common/rows_splitting.js
```diff
@@ -6,6 +6,8 @@ import { roundToThreeDecimals } from './draw_utils';
 import { getMultiPageRowPages, checkPageContainsOnlyHeader } from './rows_spliting_utils/get_multipage_row_pages';
 import { createOnSplitMultiPageRow } from './rows_spliting_utils/create_on_split_multipage_row';
 
+const ERROR = 0.001;
```
williamvinogradov: Let's rename this variable to ```COORDINATE_EPSILON```
or maybe ```COORDINATE_ACCURACY```

```ERROR``` isn't a very obvious name for a variable I think :)

williamvinogradov (approved):

PR #24565, opened 2023-05-10
DataGrid - update some ts types

pomahtri:

### DIFF
--- a/js/__internal/grids/grid_core/data_controller/module.ts
+++ b/js/__internal/grids/grid_core/data_controller/module.ts
@@ -1267,12 +1267,13 @@ export class DataController extends ControllerWithDataMixin {
   }
 
   _createDataSourceAdapter(dataSource) {
-    let remoteOperations = this.option('remoteOperations');
+    let remoteOperations: any = this.option('remoteOperations');
     const store = dataSource.store();
     const enabledRemoteOperations = {
       filtering: true, sorting: true, paging: true, grouping: true, summary: true,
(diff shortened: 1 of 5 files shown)


js/__internal/grids/grid_core/data_controller/module.ts
```diff
@@ -1267,12 +1267,13 @@ export class DataController extends ControllerWithDataMixin {
   }
 
   _createDataSourceAdapter(dataSource) {
-    let remoteOperations = this.option('remoteOperations');
+    let remoteOperations: any = this.option('remoteOperations');
```
williamvinogradov, line 1270: Why do we add ```any``` type here?
Now current calculated type is very ugly, but maybe the ugly type is better than ```any```?

williamvinogradov (approved):

js/__internal/grids/grid_core/data_controller/module.ts
pomahtri: Current calculated type of it is `never`, because there's no `remoteOperation` in GridBaseProperties at all, so it breaks all typing below

pomahtri: image

williamvinogradov: Ok, agree with you
Let's leave the ```any``` type for now

williamvinogradov (approved):

PR #24575, opened 2023-05-11
DataGrid - Header Filter - When allow select all is disabled checking first item has no effect

Tucchhaa:

### DIFF
--- a/js/ui/grid_core/ui.grid_core.header_filter_core.js
+++ b/js/ui/grid_core/ui.grid_core.header_filter_core.js
@@ -24,11 +24,17 @@ function resetChildrenItemSelection(items) {
     }
 }
 
-function updateSelectAllState(e, filterValues) {
+function getSelectAllCheckBox(listComponent) {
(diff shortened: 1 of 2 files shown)


js/ui/grid_core/ui.grid_core.header_filter_core.js
```diff
@@ -24,11 +24,17 @@ function resetChildrenItemSelection(items) {
     }
 }
 
-function updateSelectAllState(e, filterValues) {
+function getSelectAllCheckBox(listComponent) {
+    const selector = listComponent.NAME === 'dxTreeView' ? '.dx-treeview-select-all-item' : '.dx-list-select-all-checkbox';
```
pomahtri, line 28: generate selector using `this.addWidgetPrefix`

pomahtri (changes requested):

js/ui/grid_core/ui.grid_core.header_filter_core.js
Tucchhaa: in this case we would need to use `listComponent._addWidgetPrefix()`. And this is a private method, I don't think we need this

Alyar666 (approved):

pomahtri (approved):

PR #24584, opened 2023-05-11
DateRangeBox: support validation

ksercs:


js/ui/date_range_box/ui.date_range_box.js
```diff
@@ -385,6 +387,22 @@ class DateRangeBox extends Editor {
...
     }
 
+    _getRestErrors(allErrors, partialErrors) {
+        return allErrors.filter((error) => {
+            return !partialErrors.some((prevError) => error.message === prevError.message);
+        });
+    }
+
+    _syncValidationErrors(optionName, newPartialErrors, previousPartialErrors) {
+        newPartialErrors ||= [];
```
alexanderPolosatov, line 397: let's use default value for arguments?

ksercs: it can be `null`

EugeniyKiyashko (approved):

PR #24608, opened 2023-05-15
Chart - Remove accepted value 'shift' from valueAxis.visualRangeUpdateMode

alexanderPolosatov:


Krijovnick (dismissed):

js/viz/chart.d.ts
```diff
@@ -74,6 +74,7 @@ import {
     WordWrap,
     ZoomPanAction,
     ChartsColor,
+    ValueAxisVisualRangeUpdateMode,
```
ksercs, line 77: Seems like we should also reexport it here

alexanderPolosatov: fixed

js/common/charts.d.ts
```diff
@@ -260,6 +260,12 @@ export type VisualRange = {
     startValue?: number | Date | string;
 };
 
+/**
+ * @public
+ * @namespace DevExpress.common.charts
+ */
+export type ValueAxisVisualRangeUpdateMode = 'auto' | 'keep' | 'reset';
```
mpreyskurantov, line 267: 1) We already export enum with same name/values from polar_chart.d.ts:
https://github.com/DevExpress/DevExtreme/blob/23_1/js/viz/polar_chart.d.ts#L99
Consider exporting/using single enum/type, may be from common/charts.d.ts

2) Handle alias/mapping if necessary

mpreyskurantov (changes requested):

PR #24699, opened 2023-05-22
GridCore - move rows_view and columns_view to classes

pomahtri:


js/__internal/grids/grid_core/columns_controller/module.ts
```diff
@@ -363,11 +363,14 @@ export class ColumnsController extends modules.Controller {
 
   getCommonSettings(column?) {
     const commonColumnSettings = (!column || !column.type) && this.option('commonColumnSettings') || {};
+    // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
     const groupingOptions: any = this.option('grouping') || {};
```
williamvinogradov: Question.

Can the 'grouping' or 'groupPanel' options value be false but not null or undefined?
What is this scenario?

js/__internal/grids/grid_core/columns_controller/module.ts
```diff
@@ -363,11 +363,14 @@ export class ColumnsController extends modules.Controller {
...
+    // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
     const groupPanelOptions: any = this.option('groupPanel') || {};
 
     return extend({
       allowFixing: this.option('columnFixing.enabled'),
+      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
       allowResizing: this.option('allowColumnResizing') || undefined,
```
williamvinogradov: Question.

Same question for ```allowResizing```.
And is it ok, that ```allowResizeing: false``` will be equal ```undefined```?

js/__internal/grids/grid_core/data_controller/module.ts
```diff
@@ -542,7 +542,8 @@ export class DataController extends ControllerWithDataMixin {
     this.pushed.fire(changes);
   }
 
-  fireError() {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  fireError(...args: any[]) {
```
williamvinogradov: If possible let's replace ```arguments``` with ```args``` inside this method.

If it is not possible, let's remove the ```...ags``` from the method signature.

Anyway it's better don't mix up the ```args``` and ```arguments```

js/__internal/grids/grid_core/data_controller/module.ts
```diff
@@ -1256,7 +1258,7 @@ export class DataController extends ControllerWithDataMixin {
     return dataSourceAdapter;
   }
 
-  isLocalStore(store) {
+  isLocalStore(store?) {
     store = store || this.store();
```
williamvinogradov: By our code convention let's change it:

```suggestion
    store ||= this.store();
```

Or can we use the default parameter here?
```typescript
isLocalStore (store = this.store()) {
 // other code.
}

File: js/__internal/grids/grid_core/keyboard_navigation/module.ts [hunk 5]
```diff
@@ -546,6 +546,7 @@ export class KeyboardNavigationController extends modules.ViewController {
 
     if (
       this.option('selection')
+      // @ts-expect-error
       && this.option('selection').mode !== 'none'
```
williamvinogradov: Can we just cast ```this.option('selection)``` to ```any``` instead of adding ```@ts-expect-error``` here?

js/__internal/grids/grid_core/module_types.ts
```diff
@@ -1,11 +1,13 @@
 /* eslint-disable spellcheck/spell-checker */
 /* eslint-disable max-classes-per-file */
-import { PropertyType } from '@js/core/index';
+import { PropertyType as _PropertyType } from '@js/core/index';
```
williamvinogradov: Let's not use underscore in type names

js/__internal/grids/grid_core/module_types.ts
```diff
@@ -1,11 +1,13 @@
...
 import { Component } from '@js/core/component';
 import { dxElementWrapper } from '@js/core/renderer';
 import { GridBase, GridBaseOptions } from '@js/common/grids';
 import Widget from '@js/ui/widget/ui.widget';
 
+type PropertyType<T, TProp extends string> = _PropertyType<T, TProp> | undefined;
```
williamvinogradov: Let's change it to:

```suggestion
type Optional<T> = T | undefined;
type GridPropertyType<T, TProp extends string> = Optional<PropertyType<T, TProp>>;
```

js/__internal/grids/grid_core/rows_view/module.ts
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable @typescript-eslint/prefer-nullish-coalescing */
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
williamvinogradov: Why do we need unused vars here?
Maybe we can remove them from the methods?

js/__internal/grids/grid_core/columns_controller/module.ts
pomahtri: Can't
Disabled instead of fixing only because it was done automatically, I'll fix it

js/__internal/grids/grid_core/columns_controller/module.ts
pomahtri: Hard to tell, needs investigation. I suggest to leave it as is

js/__internal/grids/grid_core/rows_view/module.ts
pomahtri: It's for propery typing. Variable isn't used in method, but used in method override

williamvinogradov (approved):

PR #24714, opened 2023-05-23
Typescript: Move GridCore to ts (part 2.6)

Alyar666:


js/__internal/grids/pivot_grid/data_controller/module.ts
```diff
@@ -7,7 +7,7 @@ import Class from '@js/core/class';
 import { format } from '@js/core/utils/string';
 import { deferUpdate } from '@js/core/utils/common';
 import { isDefined, isString } from '@js/core/utils/type';
-import { VirtualScrollController } from '@js/ui/grid_core/ui.grid_core.virtual_scrolling_core';
+import VirtualScrollControllerModule from '@js/__internal/grids/grid_core/virtual_scrolling/module_core';
```
williamvinogradov, line 10: Let's change this import to:
```suggestion
import VirtualScrollControllerModule from '../../grid_core/virtual_scrolling/module_core';
```

Because we importing ts not js :)

williamvinogradov (approved):

js/__internal/grids/pivot_grid/data_controller/module.ts
pomahtri: I think it's better to leave absolute path with `@js`, since it will be easier to find it via regex to change it to `@ts/grid_core/virtual_scrolling/module_core`

Raushen (approved):

PR #24721, opened 2023-05-24
fix: csp setAttribute issues

ivanblinov2k17:


iBat (dismissed):

js/animation/position.js
```diff
@@ -354,7 +355,7 @@ const setScaleProperty = function(element, scale, transformProp, styleAttr, isEm
     if(stylePropIsValid) {
         element.style.transform = isEmpty ? transformProp.replace(scale, '') : transformProp;
     } else {
-        element.setAttribute('style', isEmpty ? styleAttr.replace(scale, '') : styleAttr);
+        setStyle(element, isEmpty ? styleAttr.replace(scale, '') : styleAttr);
```
ksercs: Seems like if in this function is not actual anymore. We can simplify the parameters.

js/core/utils/style.js
```diff
@@ -111,8 +111,26 @@ const setHeight = function(elements, value) {
     setDimensionProperty(elements, 'height', value);
 };
 
-const setStyle = function(element, value) {
-    element.style.cssText = value;
+const setStyle = function(element, styleString, resetStyle = true) {
+    if(!domAdapter.isNode(element.style)) {
```
ksercs: I think it's a good idea to add a ticket id here for an additional explain

ivanblinov2k17: transferred comment into style.js
but i consider the ticket id quite helpful in this comment because i was able to understand the problem better with ticket information

js/animation/position.js
```diff
@@ -345,17 +346,8 @@ const calculatePosition = function(what, options) {
...
-//       T941581
-const setScaleProperty = function(element, scale, transformProp, styleAttr, isEmpty) {
-    const stylePropIsValid = isDefined(element.style) && !domAdapter.isNode(element.style);
-    if(stylePropIsValid) {
-        element.style.transform = isEmpty ? transformProp.replace(scale, '') : transformProp;
-    } else {
-        element.setAttribute('style', isEmpty ? styleAttr.replace(scale, '') : styleAttr);
-    }
+const setScaleProperty = function(element, scale, styleAttr, isEmpty) {
+    setStyle(element, isEmpty ? styleAttr.replace(scale, '') : styleAttr, false);
```
ksercs:
```suggestion
    const newStyleValue = isEmpty ? styleAttr.replace(scale, '') : styleAttr;
    setStyle(element, newStyleValue, false);
```

js/core/utils/style.js
```diff
@@ -111,8 +111,30 @@ const setHeight = function(elements, value) {
...
+//       - a form contains an input with the name property set to "style";
+//       - a form contains a dx-validator (or other popup widget).
+//       T941581
+const setStyle = function(element, styleString, resetStyle = true) {
+    if(!domAdapter.isNode(element.style)) {
+        if(resetStyle) {
+            const styleList = [].slice.call(element.style);
+            styleList.forEach((propertyName) => {
+                element.style.removeProperty(propertyName);
+            });
```
dxvladislavvolkov: Use `element.style.cssText = ''` to remove all styles

ivanblinov2k17: https://w3c.github.io/webappsec-csp/#directive-style-src
by the csp docs csstext should also throw an error, despite it is implemented not everywhere

ksercs (approved):

dxvladislavvolkov (approved):

PR #24722, opened 2023-05-25
Lookup: fix drop down position in chrome on android (T1165271)

Krijovnick:


js/ui/lookup.js
```diff
@@ -61,6 +61,16 @@ const Lookup = DropDownList.inherit({
...
 
     _getDefaultOptions: function() {
+        const getSize = (side) => {
+            let size;
+            if(devices.real().deviceType === 'phone' && window.visualViewport) {
+                size = window.visualViewport[side];
+            } else {
+                size = side === 'width' ? getWidth(window) : getHeight(window);
+            }
+            return size * 0.8;
```
marker-dao: What is a constant (0.8)? Let's give her a name, shall we?

Krijovnick: It is ratio from the window size. I can name it something like WINDOW_RATIO? What do you think?
UPD: updated

testing/tests/DevExpress.ui.widgets.editors/lookup.tests.js
```diff
@@ -1594,6 +1596,25 @@ QUnit.module('options', {
         assert.equal(autoValue, initialValue, 'initial value equal auto value');
     });
 
+    QUnit.test('popup height should have correct size on mobile devices', function(assert) {
+        if(devices.real().deviceType !== 'phone') {
+            assert.ok(true, 'not mobile device');
+            return;
+        }
+        const initialVisualViewport = window.visualViewport;
+        window.visualViewport = { height: 510, width: 405 };
```
marker-dao: You should use try {} finally {} construction for it and for setting initial visualViewport.

js/ui/lookup.js
```diff
@@ -61,6 +62,16 @@ const Lookup = DropDownList.inherit({
...
     _getDefaultOptions: function() {
+        const getSize = (side) => {
+            let size;
+            if(devices.real().deviceType === 'phone' && window.visualViewport) {
+                size = window.visualViewport[side];
+            } else {
+                size = side === 'width' ? getWidth(window) : getHeight(window);
+            }
+            return size * WINDOW_RATIO;
+        };
```
ksercs, line 73: This function is not specific for lookup => it should not be places here. I think we should update core utils like `getWidth`/`getHeight` to integrate with a `visualViewport` api.

But it's better to make it in a separate task

ksercs (approved):

marker-dao (approved):

PR #24770, opened 2023-05-29
GridCore - Add eslint rule for setTimeouts and setIntervales

Tucchhaa:


pomahtri (changes requested):
this rule will be only applied to files with `module_` prefix (files with weak lint), however we need it for all files inside `__internal`

williamvinogradov (approved):

pomahtri (approved):

PR #24775, opened 2023-05-29
Popup: fix position after toggle fullScreen

Krijovnick:


testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -2735,35 +2731,51 @@ QUnit.module('positioning', {
             const visualPosition = this.getPosition();
             const expectedPosition = $target.position();
 
-            assert.deepEqual(visualPosition, expectedPosition, 'position is correct');
+            assert.deepEqual(visualPosition, expectedPosition);
         });
 
-        QUnit.test('popup should not restore position after fullScreen disable', function(assert) {
-            const visualPositionBeforeFullScreen = this.getPosition();
+        QUnit.test('popup should render on top-center position after change fullScreen option to false', function(assert) {
```
ksercs: Actually, not always at this position
```suggestion
        QUnit.test('popup should render on passed position after change fullScreen option to false', function(assert) {
```

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -2735,35 +2731,51 @@ QUnit.module('positioning', {
...
-            this.popup.option('position', { of: '#container' });
             this.popup.option('fullScreen', false);
 
-            this.popup.option('height', 'auto');
-            const visualPositionAfterFullScreen = this.getPosition();
+            const visualPosition = this.getPosition();
+            assert.deepEqual(visualPosition, initialPosition);
+        });
+
+        QUnit.test('popup should render on top-left position after change fullScreen option to true', function(assert) {
```
ksercs: It's not clear top left of what. It can be related to container / visualContainer / position.of and so on
```suggestion
        QUnit.test('popup should render at the left top corner of the window after change fullScreen option to true', function(assert) {
```

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -2735,35 +2731,51 @@ QUnit.module('positioning', {
...
 
-            const visualPositionAfterFullScreen = this.getPosition();
-            assert.deepEqual(visualPositionAfterFullScreen, visualPositionBeforeFullScreen, 'visual position was not changed');
+            const visualPosition = this.getPosition();
+            assert.deepEqual(visualPosition, expectedPosition);
         });
 
-        QUnit.test('popup should not restore position on rerender after fullScreen changed to false', function(assert) {
-            const visualPositionBeforeFullScreen = this.getPosition();
+        QUnit.test('popup should render on top-center position after on/off fullScreen option', function(assert) {
```
ksercs:
```suggestion
        QUnit.test('popup should render on passed position after on/off fullScreen option', function(assert) {
```

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -2710,23 +2710,19 @@ QUnit.module('positioning', {
             this.popup.dispose();
             this.init(options);
         };
-
-        this.init();
```
ksercs, line 2714: It's a base pattern in our tests. If we don't create a component manually in the test body we init it in before each. If u need to change default config in this test - u can use `reinit`. Why did u decide to change this?

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -2735,35 +2731,51 @@ QUnit.module('positioning', {
...
+        });
+
+        QUnit.test('popup should render on top-left position after change fullScreen option to true', function(assert) {
+            this.init();
+
+            this.popup.option('fullScreen', true);
+            const visualPosition = this.getPosition();
 
-            assert.deepEqual(visualPositionAfterFullScreen, visualPositionBeforeFullScreen, 'visual position was not changed');
+            assert.deepEqual(visualPosition, { top: 0, left: 0 });
```
ksercs, line 2772: Are there test which checks that popup position is not change if we change `position` option when `fullScreen=true`? If not, please, add it

js/ui/popup/popup_position_controller.js
```diff
@@ -137,11 +121,10 @@ class PopupPositionController extends OverlayPositionController {
 
     _fullScreenEnabled() {
         this.restorePositionOnNextRender(false);
-        this._lastPositionBeforeFullScreen = this._visualPosition;
     }
 
     _fullScreenDisabled() {
-        this.restorePositionOnNextRender(false);
+        this.restorePositionOnNextRender(true);
```
ksercs, line 123: This solution was implemented especially to not restore popup position to the initial after fullScreen toggle if the popup was dragged/resized before this.
We need to decide what which behavior is more expected: actual one or the behavior with restore.

marker-dao (dismissed):

js/ui/popup/popup_position_controller.js
```diff
@@ -46,7 +43,7 @@ class PopupPositionController extends OverlayPositionController {
     }
 
     get outsideDragFactor() {
-        return this._outsideDragFactor;
+        return this._getOutsideDragFactor();
```
ksercs: seems like we can get rid of internal getter and just write the content of "_getOutsideDragFactor" here

js/ui/popup/popup_position_controller.js
ksercs: Update: discussed with designer - he approved the change

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -2859,18 +2869,20 @@ QUnit.module('positioning', {
...
             this.drag();
-            const { left, top } = this.getPosition();
 
             this.popup.option('fullScreen', true);
             assert.deepEqual(visualPositionChangedStub.getCall(1).args[0].position, { top: 0, left: 0 }, 'position parameter is correct after change to true');
 
             this.popup.option('fullScreen', false);
-            assert.deepEqual(visualPositionChangedStub.getCall(2).args[0].position, { top, left }, 'position parameter is correct after change to false');
+
+            assert.deepEqual(visualPositionChangedStub.getCall(2).args[0].position, position, 'position parameter is correct after change to false');
```
ksercs: Let's concretise the case here
```suggestion
            assert.deepEqual(visualPositionChangedStub.getCall(2).args[0].position, position, 'position is equal to the initial position before drag');
```

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -2907,20 +2919,20 @@ QUnit.module('positioning', {
                 assert.strictEqual(newPosition.top, position.top - 100, 'top coordinate is correct');
             });
 
-            ['drag', 'resize'].forEach(moveMethodName => {
-                QUnit.module(moveMethodName, () => {
-                    QUnit.test('should not be changed after fullScreen is enabled and disabled', function(assert) {
```
ksercs, line 2912: Let's leave this test to pin the researched case.
"should be restored to initial position after fullScreen is enabled and disabled".
I also suggest u to add a note:
"// NOTE: we should implement "pin" special button as a separate feature if we wan't keep drag/resize position"

testing/tests/DevExpress.ui.widgets/popup.tests.js
Krijovnick: I thought it was redundant. But I can revert it

testing/tests/DevExpress.ui.widgets/popup.tests.js
Krijovnick: I added the test https://github.com/DevExpress/DevExtreme/pull/24775/files#diff-f5192573d06c413bc1642eed10810073c065b04910b3e8aced093ee663eff267R2785

testing/tests/DevExpress.ui.widgets/popup.tests.js
Krijovnick: revert test

testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -2907,11 +2925,24 @@ QUnit.module('positioning', {
                 assert.strictEqual(newPosition.top, position.top - 100, 'top coordinate is correct');
             });
 
+            QUnit.test('popup should have new position after on/off fullScreen and resize', function(assert) {
```
ksercs: Just updated to match the module name
```suggestion
            QUnit.test('on/off fullScreen and resize should not be restored', function(assert) {
```

ksercs (dismissed):

js/ui/popup/popup_position_controller.js
```diff
@@ -46,7 +43,11 @@ class PopupPositionController extends OverlayPositionController {
     }
 
     get outsideDragFactor() {
-        return this._outsideDragFactor;
+        if(this._props.dragOutsideBoundary) {
+            return 1;
```
marker-dao, line 47: What does 1 mean?

ksercs (approved):

js/ui/popup/popup_position_controller.js
AlexanderMoiseev: '+1'.  
Also just a suggestion
`
return this._props.dragOutsideBoundary ? 1 : this._outsideDragFactor;
`
But this code existed before, so it's a minor

Krijovnick: >What does 1 mean?
 
It is some factor using for drag and drop

AlexanderMoiseev (approved):

PR #24806, opened 2023-06-01
DataGrid - When columns are auto-generated and filterValue is specified Error E1047 appears - T1147719

Tucchhaa:

### DIFF
--- a/testing/tests/DevExpress.ui.widgets.dataGrid/filtering.integration.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.dataGrid/filtering.integration.tests.js
@@ -1801,4 +1801,49 @@ QUnit.module('Initialization', baseModuleConfig, () => {
         assert.deepEqual(actualFilterValues[0], expectedFilterValues[0]);
         assert.deepEqual(actualFilterValues[1], expectedFilterValues[1]);
     });
+
+    // T1147719
+    QUnit.test('When columns are auto-generated and filterValue is set, error should not be thrown', function(assert) {
+        try {
+            const grid = createDataGrid({
+                dataSource: [{ id: 0 }, { id: 1 }, { id: 2 }],
+                keyExpr: 'id',
+                filterValue: ['id', '=', '1'],
+                filterPanel: { visible: true },
+            });
+            this.clock.tick(10);
+
+            const rows = grid.getVisibleRows();
+
+            assert.strictEqual(rows.length, 1, 'Rows filtered');
+            assert.strictEqual(rows[0].data.id, 1);
+        } catch(err) {
+            assert.ok(false, 'error is thrown');
+        }
+    });
+
+    // T1147719
+    QUnit.test('When columns are auto-generated, filterValue is set and dataSource was set later, error should not be thrown', function(assert) {
+        try {
+            const grid = createDataGrid({
+                dataSource: undefined,
+                keyExpr: 'id',
+                filterValue: ['id', '=', '1'],
+                filterPanel: { visible: true },
+            });
+
+            this.clock.tick(10);
+
+            grid.option('dataSource', [{ id: 0 }, { id: 1 }, { id: 2 }]);
+
+            this.clock.tick(10);
+
+            const rows = grid.getVisibleRows();
+
+            assert.strictEqual(rows.length, 1, 'Rows filtered');
+            assert.strictEqual(rows[0].data.id, 1);
+        } catch(err) {
+            assert.ok(false, 'error is thrown');
+        }
+    });
 });
--- a/testing/tests/DevExpress.ui.widgets.dataGrid/filtering.integration.tests.js
+++ b/testing/tests/DevExpress.ui.widgets.dataGrid/filtering.integration.tests.js
@@ -396,9 +396,9 @@ QUnit.module('Initialization', baseModuleConfig, () => {
             }]
         });
 
-        assert.equal(calculateFilterExpressionCallCount, 6, 'calculateFilterExpression call count');
+        assert.equal(calculateFilterExpressionCallCount, 5, 'calculateFilterExpression call count');
         assert.ok(grid.getCombinedFilter(), 'combined filter');
-        assert.equal(calculateFilterExpressionCallCount, 7, 'calculateFilterExpression call count');
+        assert.equal(calculateFilterExpressionCallCount, 6, 'calculateFilterExpression call count');
     });
 
     function createRemoteDataSourceWithGroupPaging(arrayStore, key) {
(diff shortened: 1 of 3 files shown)


testing/tests/DevExpress.ui.widgets.dataGrid/filtering.integration.tests.js
```diff
@@ -396,9 +396,9 @@ QUnit.module('Initialization', baseModuleConfig, () => {
             }]
         });
 
-        assert.equal(calculateFilterExpressionCallCount, 6, 'calculateFilterExpression call count');
+        assert.equal(calculateFilterExpressionCallCount, 5, 'calculateFilterExpression call count');
```
Tucchhaa, line 399: because now in dataController, we call _calculateAdditionalFilter after check of isColumnsDefined, call count has decreased.

Ticket of this test (T802967), is about calculateFilterExpression being called infinite times, so this change is fine.

js/__internal/grids/grid_core/filter_panel/module.ts
```diff
@@ -41,14 +43,18 @@ const FilterPanelView = modules.View.inherit({
...
-      $leftContainer.append(that._getCheckElement())
-        .append(that._getFilterElement())
-        .append(that._getTextElement());
-      $element.append(that._getRemoveButtonElement());
-    } else {
-      $leftContainer.append(that._getFilterElement())
-        .append(that._getTextElement());
+    const isColumnsDefined = !!this._columnsController.getColumns().length;
+
+    if (isColumnsDefined) {
```
Tucchhaa: In the code below, text of filter builder is generated. So for generating it we need only column.name. Thus we don't have to check column types via columnsController.isAllDataTypesDefined() as we do in dataController

js/__internal/grids/grid_core/data_controller/module.ts
```diff
@@ -348,24 +348,27 @@ export class DataController extends ControllerWithDataMixin {
   }
 
   combinedFilter(filter, returnDataField?) {
-    const that = this;
-    const dataSource = that._dataSource;
-    const columnsController = that._columnsController;
+    const dataSource = this._dataSource;
+    const columnsController = this._columnsController;
```
williamvinogradov: Let's remove these local variables and use ```this._dataSource``` and ```this._columnsController``` directly

js/__internal/grids/grid_core/data_controller/module.ts
```diff
@@ -348,24 +348,27 @@ export class DataController extends ControllerWithDataMixin {
   }
 
   combinedFilter(filter, returnDataField?) {
```
williamvinogradov: Let's refactor this function a little and make it shorter

Something like this:
```suggestion
combinedFilter(filter, returnDataField?) {
    if (!this._dataSource) {
      return filter;
    }

    // Can we use "??" here?
    let resultFilter = filter ?? this._dataSource.filter();
    const resultReturnDataField = returnDataField ?? this._dataSource.remoteOperations().filtering;
    const isColumnsTypesDefined = this._columnsController.isDataSourceApplied()
      || this._columnsController.isAllDataTypesDefined();

    if (isColumnsTypesDefined) {
      const additionalFilter = this._calculateAdditionalFilter();
      resultFilter = additionalFilter
        ? gridCoreUtils.combineFilters([additionalFilter, resultFilter])
        : resultFilter;
    }

    return this._columnsController.updateFilter(
      resultFilter,
      resultReturnDataField,
    );
  }
```

js/__internal/grids/grid_core/filter_panel/module.ts
```diff
@@ -28,6 +28,8 @@ const FilterPanelView = modules.View.inherit({
 
   init() {
     this.getController('data').dataSourceChanged.add(() => this.render());
+
+    this._columnsController = this.component.getController('columns');
```
williamvinogradov: Let's use ```this.getController``` instead of the ```this.component.getController```:

```suggestion
    this._columnsController = this.getController('columns');
```

js/__internal/grids/grid_core/filter_panel/module.ts
```diff
@@ -41,14 +43,18 @@ const FilterPanelView = modules.View.inherit({
...
-        .append(that._getFilterElement())
-        .append(that._getTextElement());
-      $element.append(that._getRemoveButtonElement());
-    } else {
-      $leftContainer.append(that._getFilterElement())
-        .append(that._getTextElement());
+    const isColumnsDefined = !!this._columnsController.getColumns().length;
+
+    if (isColumnsDefined) {
+      if (that.option('filterValue') || that._filterValueBuffer) {
```
williamvinogradov: Let's use ```this``` instead of the ```that``` here

js/__internal/grids/grid_core/filter_panel/module.ts
```diff
@@ -41,14 +43,18 @@ const FilterPanelView = modules.View.inherit({
       .addClass(that.addWidgetPrefix(FILTER_PANEL_LEFT_CONTAINER))
```
williamvinogradov: Question.

Should we create a filter panel if ```isColumnDefined``` is false?

js/__internal/grids/grid_core/filter_panel/module.ts
williamvinogradov: Let's move this code to a separate render function

Something like this:
```typescript
  _renderCore(): void {
     // other code.

     // Is it safe to call getColumns().length here?
     const isColumnsDefined = !!this._columnsController.getColumns().length;

    if (isColumnsDefined) {
      this._renderFilterBuilderText($element, $leftContainer);
    }
  }

  _renderFilterBuilderText(
    $element: dxElementWrapper,
     $leftContainer: dxElementWrapper
  ): void {
    const $filterElement = this._getFilterElement();
    const $textElement = this._getTextElement();
    
    if(!this.option('filterValue') && !this._filterValueBuffer) {
      $leftContainer
        .append($filterElement)
        .append($textElement);
      return;
    } 
    
    const $checkElement = this._getCheckElement();
    const $removeButtonElement = this._getRemoveButtonElement();

    $leftContainer
      .append(this._getCheckElement())
      .append($filterElement)
      .append($textElement);
    $element.append(removeButtonElement);
  }
```

js/__internal/grids/grid_core/filter_panel/module.ts
Tucchhaa: Let's discuss it on the squad meeting

williamvinogradov (dismissed):

williamvinogradov (dismissed):

williamvinogradov (approved):

PR #24824, opened 2023-06-05
Accordion: fix title render in react if template is used (T1166943)

Krijovnick:


js/ui/accordion.js
```diff
@@ -371,6 +382,7 @@ const Accordion = CollectionWidget.inherit({
     },
 
     _clean: function() {
+        this._deferredTemplateItems = [];
```
alexanderPolosatov, line 390: should we reject deferreds before cleaning of array?

testing/tests/DevExpress.ui.widgets/accordion.tests.js
```diff
@@ -566,6 +566,85 @@ QUnit.module('widget options', moduleSetup, () => {
...
+        });
+
+        clock.tick(50);
+
+        const closedItems = $element.find(`.${ACCORDION_ITEM_CLOSED_CLASS}`);
+
+        assert.strictEqual(closedItems.length, 3);
+
+        for(let i = 0; i < closedItems.length; i++) {
+            assert.strictEqual(closedItems.eq(i).outerHeight(), 42.4219);
```
ksercs: Can we make this assert more common? Using of a such strict assertion value can make this test unstable / red if browser work changes a bit or we change a design

ksercs: As a minimum, we can use `roughEqual` with `1` px delta

EugeniyKiyashko (approved):

marker-dao (approved):

PR #24825, opened 2023-06-05
Draggable & Sortable - move to ts

pomahtri:


williamvinogradov (dismissed):

williamvinogradov:
Maybe we should place ```module_sortable``` / ```module_draggable``` not at the root of ```__internal```?
E.g. we can create a separate directory for modules like this.

williamvinogradov (approved):

PR #24829, opened 2023-06-06
FloatingActionButton: fix position after scrolling (T1167149)

Krijovnick:


marker-dao (dismissed):

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
```diff
@@ -323,14 +323,14 @@ QUnit.module('check action buttons position', {
...
+        const expectedPosition = 'absolute';
 
-        assert.equal($fabMainWrapper.css('position'), expectedPosition, 'position is fixed');
+        assert.equal($fabMainWrapper.css('position'), expectedPosition, 'position is absolute');
         assert.equal($fabWrapper.eq(0).css('position'), expectedPosition, 'first action has the same position with main fab');
         assert.equal($fabWrapper.eq(1).css('position'), expectedPosition, 'second action has the same position with main fab');
     });
 
-    QUnit.test('if random container', function(assert) {
+    QUnit.test('position should be absolute if visualContainer is specify', function(assert) {
```
ksercs: we shouldn't refactor this way. `visualContainer` can be specified only in our code - there is no API to patch it.
Not sure y this test was modified in this PR https://github.com/DevExpress/DevExtreme/pull/21037/files#diff-a047b2057d8df0d6520076cd9e33c0edff25bcd494d2194c5aa49b7fc470577eR339. Probably, just for test repair

```suggestion
    QUnit.test('position should be absolute if position.of is specified, function(assert) {
```

marker-dao (dismissed):

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
```diff
@@ -305,7 +305,14 @@ QUnit.module('check action buttons position', {
         fx.off = false;
     }
 }, () => {
-    QUnit.test('if container is window', function(assert) {
+    QUnit.test('position should be absolute if position.of is specify', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('position should be absolute if position.of is specified', function(assert) {
```

testing/tests/DevExpress.ui.widgets/speedDialAction.tests.js
```diff
@@ -323,21 +330,16 @@ QUnit.module('check action buttons position', {
...
+        const expectedPosition = 'absolute';
 
-        assert.equal($fabMainWrapper.css('position'), expectedPosition, 'position is fixed');
+        assert.equal($fabMainWrapper.css('position'), expectedPosition, 'position is absolute');
         assert.equal($fabWrapper.eq(0).css('position'), expectedPosition, 'first action has the same position with main fab');
         assert.equal($fabWrapper.eq(1).css('position'), expectedPosition, 'second action has the same position with main fab');
     });
 
-    QUnit.test('if random container', function(assert) {
+    QUnit.test('position should be fixed if position.of is not specify', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('position should be fixed if position.of is not specified', function(assert) {
```

js/ui/speed_dial_action/speed_dial_item.js
```diff
@@ -44,6 +41,11 @@ class SpeedDialItem extends Overlay {
         ]);
     }
 
+    _moveToContainer() {
```
ksercs, line 44: I don't like that we override overlay's private method to hack positioning algorithm: it's better to use options for this. But selection of "fixed/absolute" in overlays is not very intuitive, and it's also not clear which positioning we should use in FAB (probably, we should add an option to select), so let's postpone it for FAB rework.

ksercs (dismissed):

ksercs (approved):

PR #24838, opened 2023-06-07
RadioGroup: The radio button's dot should be centered when the browser window is zoomed out and the compact Material theme is used [T1165339]

EugeniyKiyashko:
Fixed dot alignment inside radiogroup circle when zoom is applied. 
- absolute pseudo element for positioning don't use anymore.
- styles made common between themes and more consistent
-  added some small fixes as animation of dot and borders in Material  according to guidlines. All changes were coordinate with designer.
- changed styles for radiogroup in List, that used pseudo element
- used flex for alignment radio-value and text instead of line-height static value using
- removed redundant `rtl` styles
- logical properties is using for `direction: rtl` mode


ksercs (dismissed):
Unification using mixins - great idea!

alexanderPolosatov (approved):

PR #24839, opened 2023-06-07
Toast: "message" option runtime change should change visible toast's text (T1169347)

ksercs:


alexanderPolosatov (approved):

js/ui/toast/ui.toast.js
```diff
@@ -176,12 +176,10 @@ const Toast = Overlay.inherit({
     },
 
     _renderContentImpl: function() {
-        if(this.option('message')) {
-            this._message = $('<div>')
-                .addClass(TOAST_MESSAGE_CLASS)
-                .text(this.option('message'))
-                .appendTo(this.$content());
-        }
+        this._message = $('<div>')
```
marker-dao, line 179: Why we should render message element here? Let's add method for it. And let's in optionChange method add follow code:

```
if(this._message) {
        this._message.text(args.value);
} else if (args.value) {
        this._yourRenderMessageMethod(args.value);
}
```

alexanderPolosatov: my opinion: Anton's way is more straight and easy to understand. this optimization (creating element only when message passed) do not give profit to us

marker-dao: Why?) This is an empty element. Browser spends resources on this. But it is not necessary. If we render such elements everywhere and every time, we will have a very slow codebase.

alexanderPolosatov: Did you see site with many toasts?)

ksercs: I think that there should be the single place where we render markup, and option change should only change the text

PR #24842, opened 2023-06-07
Overlays: Integrate VisualViewport API

marker-dao:


ksercs:
I close this PR as far as the solution creates new issues. We cannot just subscribe on visualViewport resize/scroll, because it breaks pinch zoom and scrolling on mobile device. 

Seems like we should go other way - try to rework actual positioning mechanism, consider fixed/absolute positioning, getting rid of "preventSafariScrolling", try to resolve it with css.

PR #24843, opened 2023-06-07
Theme builder: make integration test insensitive to deps order.

williamvinogradov:


pomahtri (commented):
Let's make immutable sorting. Mutating data outside of test's local scope is bad

OlegKipchatov (dismissed):

pomahtri (approved):

PR #24877, opened 2023-06-13
DataGrid - onFocusedCellChanged's 'cellElement' parameter is not an instance of DxElement (T1170682)

pomahtri:


sergepilipchuk (dismissed):

js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -2080,10 +2081,10 @@ export class KeyboardNavigationController extends modules.ViewController {
     const row = this._dataController.items()[localRowIndex];
     const column = this._columnsController.getVisibleColumns()[columnIndex];
     this.executeAction('onFocusedCellChanged', {
-      cellElement: $cell,
+      cellElement: getPublicElement($cell!),
```
williamvinogradov: Question.

It's okay that we write ```$cell!``` here?
Because the $cell argument has the ```dxElementWrapper | undefined``` type, it may be undefined here.
And we don't check if it's undefined or not above.

williamvinogradov (approved):

IlyaKhD (approved):

PR #24878, opened 2023-06-13
DataGrid - Searching in accent-insensitive mode (T1170192)

pomahtri:

### DIFF
--- a/js/__internal/grids/grid_core/search/m_search.ts
+++ b/js/__internal/grids/grid_core/search/m_search.ts
@@ -1,7 +1,8 @@
 /* eslint-disable @typescript-eslint/method-signature-style */
 import domAdapter from '@js/core/dom_adapter';
 import $ from '@js/core/renderer';
-import { compileGetter } from '@js/core/utils/data';
+// @ts-expect-error
+import { compileGetter, toComparable } from '@js/core/utils/data';
 import { isDefined } from '@js/core/utils/type';
 import dataQuery from '@js/data/query';
 import messageLocalization from '@js/localization/message';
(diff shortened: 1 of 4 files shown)


js/__internal/grids/grid_core/search/m_search.ts
```diff
@@ -232,8 +233,10 @@ export const searchModule = {
 
         _getStringNormalizer() {
           const isCaseSensitive = this.option('searchPanel.highlightCaseSensitive');
-          return function (str) {
-            return isCaseSensitive ? str : str.toLowerCase();
+          const langParams = this.getController('data')?.getDataSource?.()?.loadOptions?.()?.langParams;
```
williamvinogradov: This line of code looks overcomplicated.
Let's make it simple:

1) Let's move DataController to the field of rowsView (I guess):
```typescript
  init() {
    this._dataController = this.getController('data');
    // other code
  }
```
2) Let's add lines:
```typescript
// Do we need the "?" after the getDataSource method?
const dataSource = this._dataController?.getDataSource?.();
// Do we need the "?" after the loadOptions method?
const loadOptions = dataSource?.loadOptions?.();
const langParams = loadOptions?.langParams;
```

js/__internal/grids/grid_core/search/m_search.ts
```diff
@@ -232,8 +233,10 @@ export const searchModule = {
...
+
+          return function (str: string): string {
+            return toComparable(str, isCaseSensitive, langParams);
           };
```
williamvinogradov: Can we return the arrow function here?

```suggestion
          return (str: string): string => toComparable(str, isCaseSensitive, langParams);
```

williamvinogradov (dismissed):

williamvinogradov (approved):

PR #24880, opened 2023-06-13
GridCore - HeaderPanel - Don't update dimesions of toolbar if width haven't changed - T1170081

Tucchhaa:


testing/tests/DevExpress.ui.widgets.dataGrid/grouping.integration.tests.js
```diff
@@ -227,38 +227,6 @@ QUnit.module('Initialization', baseModuleConfig, () => {
         assert.equal(dataGrid.columnOption('field2', 'groupIndex'), undefined, 'field2 has no groupIndex');
     });
 
-    QUnit.test('Group panel should set correct \'max-width\' after clear grouping', function(assert) {
```
Tucchhaa, line 230: Because in _isSizeChanged_ method we check grid's visibility, we need to move this test to testcafe

PR #24914, opened 2023-06-20
dxScheduler - fix navigator does not show a proper month after navigation arrows are used to change the view(T1171295)

andrewmakarov:


js/ui/scheduler/header/calendar.js
```diff
@@ -81,9 +81,15 @@ export default class SchedulerCalendar extends Widget {
         return result;
     }
 
+    _optionChanged(args) {
+        if(args.name === 'value') {
+            this._calendar?.option('value', args.value);
+        }
+    }
```
williamvinogradov: Minor comment.

Maybe will it be better to use a switch here?
With the switch add new options to this method will be simpler:
```suggestion
    _optionChanged({ name, value }) {
        switch(name) {
          case 'value':
            this._calendar?.option('value', value);
            break;
          default:
            break;
        }
    }
```

js/ui/scheduler/header/header.js
```diff
@@ -107,6 +107,8 @@ export class SchedulerHeader extends Widget {
         this._createEventMap();
         this.$element().addClass(COMPONENT_CLASS);
 
+        this.updateCurrentDateLock = 0;
```
williamvinogradov: Let's remove this ```lock``` logic if it possible
(Related to your description of the PR)

js/ui/scheduler/header/header.js
```diff
@@ -183,24 +185,32 @@ export class SchedulerHeader extends Widget {
     }
 
     _updateCurrentDate(date) {
```
williamvinogradov: Let's rename it and leave the code inside of it as is:

```typescript
  _handleCurrentDateChange(date) {
    const onCurrentDateChange = this.option('onCurrentDateChange');
    onCurrentDateChange(date);

    this._callEvent('currentDate', date);
  }
```

And add the new method with the same name ```_updateCurrentDate```:
```typescript
_updateCurrentDate(date) {
  this._handleCurrentDateChange(date);
  this._calendar.option('value', date);
}
```

If we do it, we should use the ```__handleCurrentDateChange``` [here](https://github.com/DevExpress/DevExtreme/pull/24914/files#diff-841e3cf44beb9f50697a667bde93fa69432c1d828f2642c500e64b774dc43927R213) and leave second usage of the  ```_updateCurrentDate``` without changes :)

williamvinogradov (approved):

PR #24925, opened 2023-06-21
T1172330 - Add missing fields to GroupDescriptor

const314:


js/data/index.d.ts
```diff
@@ -11,18 +11,29 @@ export type SearchOperation = '=' | '<>' | '>' | '>=' | '<' | '<=' | 'startswith
...
  * @public
  * @type object
  * @skip
  */
-export type GroupDescriptor<T> = KeySelector<T> | BaseGroupDescriptor<T> & {
-    desc?: boolean;
+export type GroupDescriptor<T> = KeySelector<T> | OrderingDescriptor<T> & {
+    groupInterval?: number | GroupingInterval;
+    isExpanded?: boolean;
 };
```
IlyaKhD: **Minor**
Adding braces should increase readability:
```suggestion
export type GroupDescriptor<T> = KeySelector<T> | (OrderingDescriptor<T> & {
    groupInterval?: number | GroupingInterval;
    isExpanded?: boolean;
});
```

js/data/index.d.ts
```diff
@@ -31,7 +42,7 @@ export type GroupDescriptor<T> = KeySelector<T> | BaseGroupDescriptor<T> & {
  * @type object
  * @skip
  */
-export type SortDescriptor<T> = GroupDescriptor<T>;
+export type SortDescriptor<T> = OrderingDescriptor<T>;
```
IlyaKhD: Previously it can be `KeySelector` as well. Is this intended?

```suggestion
export type SortDescriptor<T> = KeySelector<T> | OrderingDescriptor<T>;
```

js/data/index.d.ts
```diff
@@ -11,18 +11,29 @@ export type SearchOperation = '=' | '<>' | '>' | '>=' | '<' | '<=' | 'startswith
...
 };
 
+type OrderingDescriptor<T> = SelectionDescriptor<T> & {
+    desc?: boolean;
+};
+
+/**
+ * @docid
+ * @public
+ */
```
IlyaKhD: Discovering unions has problems with `@docid` tag now, let's skip it for now (we probably will remove it from the `SearchOperation` as well
```suggestion
/**
 * @public
 */
```

IlyaKhD (dismissed):

js/data/index.d.ts
const314: fixed

js/data/index.d.ts
const314: fixed

js/data/index.d.ts
const314: You are right, missed it. Fixed.

IlyaKhD (approved):

mpreyskurantov (approved):

PR #24945, opened 2023-06-23
Rework Charts to TS

alexanderPolosatov:


pomahtri (changes requested):
If you want to save correct file history (to use blame), you need to merge this PR without squashing. In this case all commits will be included in main branch as is, so please change their names, and remove merge commits

alexanderPolosatov:
> If you want to save correct file history (to use blame), you need to merge this PR without squashing. In this case all commits will be included in main branch as is, so please change their names, and remove merge commits

ok. thank you

testing/tests/DevExpress.common/charts.tests.js
```diff
@@ -1,4 +1,5 @@
-import { registerPattern, registerGradient, getGraphicObjects } from 'common/charts';
+import { registerPattern, registerGradient } from 'common/charts';
+import graphicObjects from '__internal/common/m_charts';
```
Krijovnick, line 2: Could we import registerPattern and registerGradient from '__internal/common/m_charts'?

alexanderPolosatov: common/charts - it's a public path to this methods. I think we should testing this methods using public paths(or how we check, what public methods is working)

Krijovnick: ok, could we import { getGraphicObjects } there instead of graphicObjects?

alexanderPolosatov: we have
export default { getGraphicObjects };
this kind of export was added for [mock](https://github.com/DevExpress/DevExtreme/blob/23_2/testing/tests/DevExpress.viz.charts/pieChart.tests.js#L2438) this function. if we will not export { getGraphicObjects }, we can't write this mock.

Krijovnick (approved):

marker-dao (approved):

PR #24974, opened 2023-06-28
Review charts types shown in the api reference

IlyaKhD:
### Cherry-picks
- #25002


js/viz/core/base_widget.d.ts
```diff
@@ -259,6 +260,7 @@ export interface BaseWidgetExport {
 /**
  * @docid
  * @namespace DevExpress.viz
+ * @hidden
```
mpreyskurantov: Let's keep @hidden first everywhere

IlyaKhD: Done

vladaskorohodova (dismissed):

mpreyskurantov (dismissed):

js/viz/bar_gauge.d.ts
```diff
@@ -402,23 +403,24 @@ export interface dxBarGaugeLegend extends BaseLegend {
...
-}
+};
 /**
- * @docid
- * @namespace DevExpress.viz
+ * @public
+ * @docid dxBarGaugeTooltip
  */
-export interface dxBarGaugeTooltip extends BaseWidgetTooltip {
+export interface Tooltip extends BaseWidgetTooltip {
```
Krijovnick: Everywhere `interface` was changed to `type` except Tooltip. Should be there `type` too?

js/viz/chart.d.ts
```diff
@@ -1945,12 +1955,12 @@ export interface dxChartCommonAxisSettings {
      * @public
      */
     width?: number;
-}
+};
 /**
- * @docid
- * @namespace DevExpress.viz
+ * @public
+ * @docid dxChartCommonAxisSettingsConstantLineStyle
```
Krijovnick, line 1961: Should be there @type object? The same question for many places below

js/viz/bar_gauge.d.ts
IlyaKhD: Yes, it should
Thanks!

IlyaKhD: Done

js/viz/chart.d.ts
IlyaKhD: I think we can skip it here
`@type object` is not something meaningful, we have to use it in some places to correctly display the API Reference. 
In this particular place we can skip it because API Reference should be OK without it

Krijovnick (approved):

mpreyskurantov (approved):

PR #24978, opened 2023-06-28
GridCore - Show empty header message when there are command columns

Tucchhaa:


js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
```diff
@@ -632,106 +633,151 @@ export class ColumnsController extends modules.Controller {
...
-      result[i] = [];
+    const positiveIndexedColumns: any = [];
+    const negativeIndexedColumns: any = [];
+
+    for (let i = 0; i < rowCount; i += 1) {
       negativeIndexedColumns[i] = [{}];
+
+      // 0 - fixed columns on the left side
+      // 1 - not fixed columns
+      // 2 - fixed columns on the right side
```
pomahtri, line 686: If it's not hard to refactor, let's move this comments to variable names
```javascript
positiveIndexedColumns[i] = {fixedLeft: {}, notFixed: {}, fixedRight: {}}
```

pomahtri:
minor: let's add typehints for function args and result values where it's not hard

pomahtri (approved):

js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
Tucchhaa: unfortunately it'll be pretty hard. This logic needs serious refactoring, because 3-dimensional arrays are being used here. And with no reason (as far as know).

I tried to remove this hard logic, but it took too much time

PR #24979, opened 2023-06-28
NumberBox: valueChanged event should be raised on a minus sign remove by backspace  (T1170374)

AlexanderMoiseev:


testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
```diff
@@ -1820,6 +1820,105 @@ QUnit.module('format: removing', moduleConfig, () => {
...
+            onValueChanged: valueChangedStub,
+        });
+
+        this.keyboard.caret(1).press('backspace');
+
+        assert.strictEqual(this.input.val(), '1,323');
+        assert.ok(valueChangedStub.calledOnce);
+    });
+
+    QUnit.test('onValueChanged should be triggered on keyup', function(assert) {
```
EugeniyKiyashko: You may iterate values in cycle and use one test, to avoid duplication

```
['keyup', 'change'].forEach((valueChangeEvent) => {
    QUnit.test('testName', ...)
})
```

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
```diff
@@ -1820,6 +1820,105 @@ QUnit.module('format: removing', moduleConfig, () => {
         assert.equal(changeHandler.callCount, 1, 'change event has not been fired if value is not changed');
     });
 
+    QUnit.test('removing minus should trigger onValueChanged event and revert sign when using unformatted value', function(assert) {
```
EugeniyKiyashko: You don't check event arguments that comes in handler. Do you think it's not require?

js/ui/number_box/number_box.mask.js
```diff
@@ -243,6 +243,11 @@ const NumberBoxMask = NumberBoxBase.inherit({
             if(this._parsedValue < 0 || 1 / this._parsedValue === -Infinity) {
                 this._revertSign(e);
                 this._setTextByParsedValue();
+
+                if(this.option('valueChangeEvent') === 'input') {
+                    const valueChangedEventName = addNamespace(this.option(this._getValueChangeEventOptionName()), `${this.NAME}ValueChange`);
+                    eventsEngine.trigger(this._input(), valueChangedEventName);
```
EugeniyKiyashko: Why you don't trigger event as follows:
```
eventsEngine.trigger(this._input(), 'input');
```
?

js/ui/number_box/number_box.mask.js
```diff
@@ -243,6 +243,11 @@ const NumberBoxMask = NumberBoxBase.inherit({
             if(this._parsedValue < 0 || 1 / this._parsedValue === -Infinity) {
                 this._revertSign(e);
                 this._setTextByParsedValue();
+
+                if(this.option('valueChangeEvent') === 'input') {
```
EugeniyKiyashko: The 'input' event will not trigger for cases when we mix `valueChangeEvent` events.
For example this case:
`valueChangeEvent: 'input focusout'`. It should be work too.

AlexanderMoiseev: thanks, i missed that it can accept several events

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
AlexanderMoiseev: sure, but these tests (for keyup and change) differ not only by event names, but also by keyboard actions we perform to simulate these events ( this.keyboard.keyUp('backspace') and this.keyboard.press('enter') ). So i'm not sure it's worth it

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
AlexanderMoiseev: i'll add some tests for that

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
EugeniyKiyashko: I wrote code snippet for demonstration of approach. There are and other events. What about 'blur' and 'focusout' test? The scenarios look the same. You may take a decision to do or not to do yourself. It's recommendation only.

AlexanderMoiseev: yep, for blur and focusout its perfect

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
```diff
@@ -1066,6 +1066,31 @@ QUnit.module('format: text input', moduleConfig, () => {
         assert.equal(valueChangedHandler.getCall(0).args[0].event.type, 'change', 'event is correct');
     });
 
+    QUnit.test('onValueChanged should have input event as a parameter if valueChangeEvent value is input', function(assert) {
```
alexanderPolosatov: usually, we include to name of tests, related to the bug, ID of this bug

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
```diff
@@ -1820,6 +1845,102 @@ QUnit.module('format: removing', moduleConfig, () => {
...
+            });
+
+            this.keyboard.caret(1).press('backspace');
+            assert.ok(valueChangedStub.notCalled);
+
+            this.input.trigger(valueChangeEvent);
+            assert.ok(valueChangedStub.calledOnce);
+        });
+    });
+
```
ksercs, line 1945: Can we unite all this tests ti a separate module? 
```javascript
Qunit.module('valueChanged event on minus sign remove', () => {
    Qunit.test('should be triggered if valueChangeEvent="keyup"');
    Qunit.test('should be triggered if valueChangeEvent="input"');
    Qunit.test('should be triggered when format is specified');
    ...
});
```

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
```diff
@@ -1820,6 +1845,102 @@ QUnit.module('format: removing', moduleConfig, () => {
...
+            onValueChanged: valueChangedStub,
+        });
+
+        this.keyboard.caret(1).press('backspace');
+
+        assert.strictEqual(this.input.val(), '1,323');
+        assert.ok(valueChangedStub.calledOnce);
+    });
+
+    QUnit.test('onValueChanged should be triggered on keyup', function(assert) {
```
ksercs: What about `keydown`?

js/ui/number_box/number_box.mask.js
```diff
@@ -243,6 +243,10 @@ const NumberBoxMask = NumberBoxBase.inherit({
             if(this._parsedValue < 0 || 1 / this._parsedValue === -Infinity) {
                 this._revertSign(e);
                 this._setTextByParsedValue();
+
+                if(this.option('valueChangeEvent').split(' ').includes('input')) {
```
ksercs: 1. It's not necessary to split - string has [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) method
2. Let's move this to a variable `shouldTriggerInputEvent` to simplify future debug

testing/tests/DevExpress.ui.widgets.editors/numberBoxParts/mask.tests.js
AlexanderMoiseev: i'll add toDiscuss card regarding keydown

js/ui/number_box/number_box.mask.js
```diff
@@ -243,6 +243,11 @@ const NumberBoxMask = NumberBoxBase.inherit({
             if(this._parsedValue < 0 || 1 / this._parsedValue === -Infinity) {
                 this._revertSign(e);
                 this._setTextByParsedValue();
```
Krijovnick, line 245: Why is it necessary to call "_setTextByParsedValue" method? Seems to be enough to do revertSign only there and add return there.

AlexanderMoiseev: it sets text in input with parsed value ) pretty sure there are tests for that

Krijovnick: If remove it and past `return` there, `_setTextByParsedValue` method will be call with parsed value anyway and will called onValueChanged event without adding that changes

AlexanderMoiseev: plz take a look at failed tests if you do so. They are self-explanatory

Krijovnick: I have only one failed test 'should be triggered on keyup when valueChangeEvent="keyup"' - and it is worked in real example

AlexanderMoiseev: had i call with Yulia, resolved

ksercs (approved):

EugeniyKiyashko (approved):

PR #24980, opened 2023-06-28
Toast: change animation type and width for tablets and mobiles (T1169464)

marker-dao:


testing/tests/DevExpress.ui/defaultOptions.tests.js
```diff
@@ -901,78 +903,134 @@ testComponentDefaults(TagBox,
...
+            offset: '0 -20'
+        },
+        width: `calc(100vw - ${DEFAULT_MARGIN * 2}px)`,
+        displayTime: 4000,
+        hideOnOutsideClick: true,
+        animation: tabletAndMobileAnimation,
+    },
+    function() {
+        this.origIsMaterial = themes.isMaterial;
+        themes.isMaterial = function() {
```
alexanderPolosatov, line 1063: let's replace to 
themes.isMaterial = () => true

?

js/ui/toast/ui.toast.js
```diff
@@ -91,81 +90,55 @@ const Toast = Overlay.inherit({
...
+        };
+
         return this.callBase().concat([
             {
-                device: { platform: 'android' },
+                device(device) {
+                    return device.deviceType === 'phone';
+                },
                 options: {
+                    width: `calc(100vw - ${DEFAULT_MARGIN * 2}px)`,
```
Krijovnick, line 120: The `width` and `displayTime` properties have the same calculation in two places. Can we make them one time in the top of the function?

marker-dao: I don't think we should take such small calculations anywhere. In addition, we use different widths (width and maximum width).

Krijovnick: You can add calculations in the extra variable and pass it to the width and maxWidth

ksercs: I suggest using relative values. For example:
`margin: 5%, width: 90%`
It will simplify the code and improve appearance on different devices

js/ui/toast/ui.toast.js
```diff
@@ -91,81 +90,55 @@ const Toast = Overlay.inherit({
...
                 },
                 options: {
-                    width: '100vw',
-                }
+                    width: 'auto',
+                    maxWidth: `calc(100vw - ${DEFAULT_MARGIN * 2}px)`,
+                    displayTime: isMaterial() ? 4000 : 2000,
+                    hideOnOutsideClick: true,
+                    animation: tabletAndMobileAnimation,
+                },
```
ksercs, line 132: all this option seem to be the same as for phones. Can we unify it?
```javascript
device({ deviceType }) {
    return deviceType === 'phone' || deviceType === 'tablet';
},
options: {
      width: 'auto',
      maxWidth: `calc(100vw - ${DEFAULT_MARGIN * 2}px)`,
      displayTime: isMaterial() ? 4000 : 2000,
      hideOnOutsideClick: true,
      animation: tabletAndMobileAnimation,
  },
```

testing/tests/DevExpress.ui/defaultOptions.tests.js
```diff
@@ -901,78 +903,134 @@ testComponentDefaults(TagBox,
...
+    },
+    hide: {
+        type: 'fade',
+        duration: 200,
+        from: 1,
+        to: 0
+    }
+};
+
+const mobileOptions = {
```
ksercs: We should not specify such constants in this file. We test default options for all components here - it's not clear that this "mobileOptions" are related only to the toast. Let's use inline objects

js/ui/toast/ui.toast.js
marker-dao: Not all

js/ui/toast/ui.toast.js
ksercs: We decided to change maxWidth on tables to 80vw

js/ui/toast/ui.toast.js
marker-dao: I decided to move the same options into a common options object.

Krijovnick (approved):

AlexanderMoiseev (approved):

EugeniyKiyashko (approved):

PR #24999, opened 2023-07-03
DataGrid - Disable export button when data columns are not visible (T1174225)

Tucchhaa:


js/__internal/grids/data_grid/export/m_export.ts
```diff
@@ -791,13 +797,35 @@ dataGridCore.registerModule('export', {
           }
         },
 
+        _needDisableExportButton(): boolean {
+          const isDataColumnsInvisible = !this._columnsController.hasVisibleDataColumns();
```
pomahtri, line 801: minor:
```suggestion
          const areDataColumnsInvisible = !this._columnsController.hasVisibleDataColumns();
```

pomahtri (approved):

PR #25000, opened 2023-07-04
temp

AlexanderMoiseev:


js/viz/funnel/label.js
```diff
@@ -161,7 +161,9 @@ function removeEmptySpace(labels, requiredSpace, startPoint) {
         const prevLabel = labels[index + 1];
         if(requiredSpace > 0) {
             const bBox = label.getBoundingRect();
-            const point = prevLabel ? prevLabel.getBoundingRect().y + prevLabel.getBoundingRect().height : startPoint;
+
+            const prevLabelIsDrawn = prevLabel?._drawn !== false;
```
Krijovnick: The most using in charts in such cases is `prevLabel?.isVisible()`.

Krijovnick:
I think pull request can be renamed as "Funnel: crash when labels has zero values (T1172293)"

PR #25020, opened 2023-07-06
Funnel: fix crash on init when datasource has zero values (T1172293)

AlexanderMoiseev:

### DIFF
--- a/testing/tests/DevExpress.viz.funnel/funnel.label.tests.js
+++ b/testing/tests/DevExpress.viz.funnel/funnel.label.tests.js
@@ -2,7 +2,6 @@ import $ from 'jquery';
 import { createFunnel, stubAlgorithm, checkNumbersWithError } from './commonParts/common.js';
 import labelModule from 'viz/series/points/label';
 import { labelEnvironment } from './commonParts/label.js';
-import { logger } from 'core/utils/console';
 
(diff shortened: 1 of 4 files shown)


testing/tests/DevExpress.viz.funnel/funnel.label.tests.js
```diff
@@ -1300,3 +1301,57 @@ QUnit.test('change resolveLabelOverlapping option', function(assert) {
...
+                    { count: 20, level: 'Mid-Level Engineer' },
+                    { count: 0, level: 'Senior Engineer' },
+                    { count: 17, level: 'Architect' },
+                ],
+                valueField: 'count',
+                inverted: true,
+                sortData: false,
+            });
+
+            assert.strictEqual(this.loggerErrorSpy.callCount, 0);
```
Krijovnick: The logger is not useful here. Widget failed and this assert will not work. The best way to check is wrap the createFunnel in try catch like in [datebox](https://github.com/DevExpress/DevExtreme/blob/1ce4576a12c62e36f6eb714f0df8bf9a5f2e6807/testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js#L228) tests. When the test is failed, there will be a full message what is wrong

Krijovnick (approved):

EugeniyKiyashko (approved):

PR #25030, opened 2023-07-07
GridCore - Update cached selected items on data source push event (T1171774)

Tucchhaa:

### DIFF
--- a/js/__internal/grids/grid_core/selection/m_selection.ts
+++ b/js/__internal/grids/grid_core/selection/m_selection.ts
@@ -114,7 +114,11 @@ const SelectionController = modules.Controller.inherit((function () {
       this._selection = this._createSelection();
       this._updateSelectColumn();
       this.createAction('onSelectionChanged', { excludeValidators: ['disabled', 'readOnly'] });
-      this._dataController && this._dataController.pushed.add(this._handleDataPushed.bind(this));
+
+      if (!this._dataPushedHandler) {
+        this._dataPushedHandler = this._handleDataPushed.bind(this);
+        this._dataController.pushed.add(this._dataPushedHandler);
+      }
     },
 
     _handleDataPushed(changes) {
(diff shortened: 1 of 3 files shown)


testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -799,7 +799,7 @@ QUnit.module('Initialization', baseModuleConfig, () => {
         this.clock.tick(10);
 
         // assert
-        assert.equal(keyOfSpy.callCount, 55, 'keyOf call count');
+        assert.equal(keyOfSpy.callCount, 60, 'keyOf call count');
```
Tucchhaa, line 802: Needed to increase this number because keyOf is being called inside applyBatch function ([here](https://github.com/DevExpress/DevExtreme/pull/25030/files#diff-784173722ee31f6c85d8472321b20084b72811d04f0e3645848dc45ec36fc004R146))

js/__internal/grids/grid_core/selection/m_selection.ts
```diff
@@ -133,6 +139,16 @@ const SelectionController = modules.Controller.inherit((function () {
       removedKeys.length && this.deselectRows(removedKeys);
     },
 
+    _updateSelectedOnPush(changes) {
+      const updateChanges = changes.filter((change) => change.type === 'update');
+
+      applyBatch({
+        keyInfo: this._selection.options,
+        data: this.getSelectedRowsData(),
```
williamvinogradov: Question.

Is it okay, that we don't handle the [selection.deferred](https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/selection/#deferred) option here and calls the ```getSelectedRowsData``` method?
(As far as I understand it may cause additional requests to the server side)

And according to the [getSelectedRowsData documentation](https://js.devexpress.com/Documentation/ApiReference/UI_Components/dxDataGrid/Methods/#getSelectedRowsData), it may return a promise with the selection.deferred = true o_O

Or I missed something and understand it wrong? :)

Tucchhaa: You are absolutely right. I need to add condition for returning from function if selection.deferred = true

williamvinogradov (approved):

PR #25043, opened 2023-07-10
DataGrid - Excel export of selected items exports an empty array if the filtered column is bound to a collection and remote filtering is used (T1173757)

pomahtri:


js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -1365,7 +1366,7 @@ export class DataController extends ControllerWithDataMixin {
           isCustomLoading: true,
           storeLoadOptions: { isLoadingAll: true },
           loadOptions: {
-            filter: that.getCombinedFilter(),
+            filter: skipFilter ? undefined : that.getCombinedFilter(),
```
williamvinogradov: Minor comment.

Will it be better if we'll use null here?
Because undefined - it's not defined, and null is an empty reference (empty filter in our case)

js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -1432,7 +1433,8 @@ export class DataController extends ControllerWithDataMixin {
     }
   }
 
-  getRowIndexOffset() {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  getRowIndexOffset(byLoadedRows?: boolean) {
```
williamvinogradov, line 1437: Question.

Why do we need the unused variable here?

js/__internal/grids/data_grid/export/m_export.ts
```diff
@@ -30,7 +30,14 @@ export class DataProvider {
...
 
-  private _options: any;
+  private _options!: {
+    isHeadersVisible: any;
+    rtlEnabled: any;
+    summaryTexts: any;
+    groupColumns: any;
+    items: any;
+    columns: any;
+  };
```
williamvinogradov: Let's add obvious types right away, e.g:
```typescript
  private _options!: {
    isHeadersVisible: boolean;
    rtlEnabled: boolean;
    summaryTexts: any;
    groupColumns: any;
    items: any;
    columns: any;
  };
```

And question.
Can we use the ```unknown``` instead of the ```any``` here?
Will it break something?

js/__internal/grids/grid_core/data_controller/m_data_controller.ts
pomahtri: For proper typing in extenders. This method is extended in virtual scrolling with new param, but type is calculated from base class

js/__internal/grids/grid_core/data_controller/m_data_controller.ts
pomahtri: I don't see any difference. Filter is undefined -> filter is not defined -> no filter. I'll change

js/__internal/grids/data_grid/export/m_export.ts
pomahtri: boolean fields are fixed
others remain `any`, there are too many errors. For example, `items` can't even be defined as `any[]`, cause sometimes it's Deferred 🥲

williamvinogradov (approved):

PR #25086, opened 2023-07-12
Scheduler(T1175936): It's not possible to scroll with one finger on iPad

williamvinogradov:
Fixed the iPadOs detection by user agent issue.

### DIFF
--- a/packages/devextreme/js/core/devices.js
+++ b/packages/devextreme/js/core/devices.js
@@ -88,7 +88,10 @@ const UA_PARSERS = {
     android(userAgent) {
         // TODO: Check this RegExp.
         //  It looks like there may be missing android user agents.
-        if(!/android|htc_|silk/i.test(userAgent) || /windows phone/i.test(userAgent)) {
+        const isAndroid = /android|htc_|silk/i.test(userAgent);
+        const isWinPhone = /windows phone/i.test(userAgent);
+
+        if(!isAndroid || isWinPhone) {
             return null;
         }
 
--- a/packages/devextreme/js/core/devices.js
+++ b/packages/devextreme/js/core/devices.js
@@ -3,7 +3,6 @@ import $ from '../core/renderer';
 import { getWindow, getNavigator, hasWindow } from './utils/window';
 import { extend } from './utils/extend';
 import { isPlainObject } from './utils/type';
-import { each } from './utils/iterator';
 import errors from './errors';
 import Callbacks from './utils/callbacks';
 import readyCallbacks from './utils/ready_callbacks';
@@ -13,7 +12,6 @@ import { sessionStorage as SessionStorage } from './utils/storage';
 import { changeCallback, value as viewPort } from './utils/view_port';
 import Config from './config';
 
-const navigator = getNavigator();
 const window = getWindow();
 
 const KNOWN_UA_TABLE = {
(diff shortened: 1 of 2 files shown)


js/core/devices.js
```diff
@@ -295,15 +305,13 @@ class Devices {
     }
 
     _fromUA(ua) {
-        let config;
-
-        each(uaParsers, (platform, parser) => {
-            config = parser(ua);
-            return !config;
-        });
+        for(let idx = 0; idx < UA_PARSERS_ARRAY.length; idx += 1) {
```
Tucchhaa: let's rename idx to index

js/core/devices.js
```diff
@@ -84,8 +86,10 @@ const uaParsers = {
     },
 
     android(userAgent) {
-        if(!/android|htc_|silk/i.test(userAgent)) {
-            return;
+        // TODO: Check this RegExp.
+        //  It looks like there may be missing android user agents.
+        if(!/android|htc_|silk/i.test(userAgent) || /windows phone/i.test(userAgent)) {
+            return null;
```
Tucchhaa: can we rewrite it like this:
```suggestion
        const isAndroid = /android|htc_|silk/i.test(userAgent)
        const isWindowsPhone = /windows phone/i.test(userAgent)
        
        if(!isAndroid || isWindowsPhone) {
          return null;
        }
```

js/core/devices.js
```diff
@@ -13,7 +12,6 @@ import { sessionStorage as SessionStorage } from './utils/storage';
 import { changeCallback, value as viewPort } from './utils/view_port';
 import Config from './config';
 
-const navigator = getNavigator();
```
Tucchhaa: afaik, navigator is a global object like window. Why would you move it to local variables, instead of keeping it here like _window_ var below?

Raushen (dismissed):

Tucchhaa (changes requested):

js/core/devices.js
williamvinogradov: Yes, I agree with you. 
Thank you for the great suggestion!

js/core/devices.js
williamvinogradov: I change it because I need to mock the navigator in QUnit tests.
The [getNavigator](https://github.com/DevExpress/DevExtreme/blob/23_1/js/core/utils/window.js#L48) method returns the navigator from the custom window object from a closure.
And we can set a custom window object (mock) with the [setWindow](https://github.com/DevExpress/DevExtreme/blob/23_1/js/core/utils/window.js#L18) method.

So, if I leave this as is

I should mock the window before I import devices.js in QUnit's test file.
(I'm not 100% sure that it will work, but looks like it will)
From my point of view, it looks unobvious and ugly, so I decided to get navigator from the getNavigator method in functions runtime instead of declaring a variable in this scope :)

js/core/devices.js
williamvinogradov: Let's discuss this comment with the team separately.
From my point of view, it's a matter of taste, so we need the team's code style agreement here.

Tucchhaa (dismissed):

js/core/devices.js
ivanblinov2k17: Yeah, it's definitely a matter of taste, but i'm personally into index/i

ivanblinov2k17 (dismissed):

pomahtri (approved):

PR #25090, opened 2023-07-13
DataGrid: Fix auto scroll up when scrolling horizontally (T1176160)

Alyar666:
See https://devexpress.com/issue=T1176160


testing/testcafe/tests/dataGrid/virtualColumns.ts
```diff
@@ -130,3 +130,54 @@ test('The updateDimensions method should render the grid if a container was hidd
...
+    template() {
+      return ($('<div style=\'height: 300px;\'>') as any).text('details');
+    },
+  },
+  onContentReady(e) {
+    // eslint-disable-next-line no-underscore-dangle
+    if (!e.component.__initExpand) {
+    // eslint-disable-next-line no-underscore-dangle
+      e.component.__initExpand = true;
+      e.component.expandRow(9);
```
williamvinogradov: Question.

Maybe it will be better if we expand a middle row?
To see the rows before and after the master-detail.

From my point of view, it will be the nearest test case to the original user case.

williamvinogradov (dismissed):

pomahtri (approved):

PR #25109, opened 2023-07-17
DataGrid: Fix new rows rendering when scrolling horizontally then collapsing a row (T1176161)

Alyar666:
See https://devexpress.com/issue=T1176161


js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
```diff
@@ -150,10 +150,12 @@ const baseFixedColumns = {
 
     if (this._isFixedColumns) {
       const change = options?.change;
+      const $dataRows = this._getRowElements(this._fixedTableElement);
```
williamvinogradov: Minor comment.

Maybe it will be better to name this variable as ```$fixedDataRows```?
Because we get it from the fixed table :)

williamvinogradov (dismissed):

williamvinogradov (approved):

PR #25114, opened 2023-07-18
Scheduler - Remove warning about the deprecated 'preventScrollEvents' property in appointment popups

markallenramirez:


williamvinogradov (dismissed):

js/ui/scheduler/appointmentPopup/popup.js
```diff
@@ -25,7 +25,6 @@ const POPUP_CONFIG = {
     maxHeight: '100%',
     showCloseButton: false,
     showTitle: false,
-    preventScrollEvents: false,
```
williamvinogradov, line 28: Unfortunately, we cannot stop using this deprecated option.
This option is needed for the zoom & swipe events on mobile devices.
Without this option, these actions will not work correctly.

We can ask @EugeniyKiyashko for a more detailed answer as to why we should continue using this option :)

For disabling console warnings, Eugeniy suggests using the ```_ignorePreventScrollEventsDeprecation: true```
```javascript
const POPUP_CONFIG = {
  // other popup options.
  preventScrollEvents: false,
  _ignorePreventScrollEventsDeprecation: true,
  // other popup options.
};
```

testing/tests/DevExpress.ui.widgets.scheduler/appointmentPopup.tests.js
```diff
@@ -757,7 +757,6 @@ QUnit.module('Appointment Popup Content', moduleOptions, () => {
         const appointmentPopupOptions = scheduler.instance._appointmentPopup.popup.option();
 
         assert.strictEqual(appointmentPopupOptions.enableBodyScroll, false, 'enable body scroll');
-        assert.strictEqual(appointmentPopupOptions.preventScrollEvents, false, 'prevent scroll events');
```
williamvinogradov: We should leave this check in our tests.

williamvinogradov (approved):

PR #25120, opened 2023-07-18
Chart: bar should be rendered when value is on border (T1176892)

Zedwag:


packages/devextreme/js/viz/series/points/bar_point.js
```diff
@@ -46,9 +46,16 @@ export default _extend({}, symbolPoint, {
     },
 
     _calculateVisibility: function(x, y, width, height) {
+        this.inVisibleArea = this._getPointVisibility(x, y, width, height, true);
+    },
+
+    _getPointVisibility: function(x, y, width, height, allowEqual) {
```
ksercs: I suggest u use object parameter to make last boolean prop pass more explicit

packages/devextreme/js/viz/series/points/bar_point.js
```diff
@@ -46,9 +46,16 @@ export default _extend({}, symbolPoint, {
...
         const { minX, maxX, minY, maxY } = this._getVisibleArea();
 
-        this.inVisibleArea = minX < x + width && maxX > x && minY < y + height && maxY > y;
+        return (allowEqual ? maxX >= x : maxX > x)
+            && (allowEqual ? minY <= y + height : minY < y + height)
+            && (allowEqual ? minX <= x + width : minX < x + width)
+            && (allowEqual ? maxY >= y : maxY > y);
```
ksercs: let's write it as `if(allowEqual) {...` to make the code clearer

packages/devextreme/js/viz/series/points/bar_point.js
ksercs: equal what, points? `shouldDrawBorderPoints`?

packages/devextreme/js/viz/series/points/range_bar_point.js
```diff
@@ -9,6 +9,10 @@ export default _extend({}, barPoint, {
 
     _getFormatObject: rangeSymbolPointMethods._getFormatObject,
 
+    _calculateVisibility: function(x, y, width, height) {
+        this.inVisibleArea = barPoint._getPointVisibility.call(this, x, y, width, height, false);
```
ksercs: Seems like we can just override it, no?

Zedwag: override with the same function but with `>` instead of `>=`? We can, but current way seems more clear

packages/devextreme/js/viz/series/points/bar_point.js
Zedwag: equal x/y position. I don't like `shouldDrawBorderPoints` :(

Zedwag: may be `allowBorderPoints`?

packages/devextreme/js/viz/series/points/bar_point.js
Zedwag: ok

packages/devextreme/js/viz/series/points/bar_point.js
Zedwag: ok

packages/devextreme/js/viz/series/points/bar_point.js
```diff
@@ -273,16 +273,17 @@ export default _extend({}, symbolPoint, {
         that[argAxis] = arg = arg === null ? arg : arg + (that[argAxis + 'Correction'] || 0);
         that['v' + valAxis] = val;
         that['v' + argAxis] = arg + that[argIntervalName] / 2;
+        that[valIntervalName] = _abs(val - minVal);
```
Krijovnick: Current `that[valIntervalName]` and `newVal` values need for calculate visibility only and then the are recalculated. Let's move this calculations in the separate function not to clutter current method and renamed them. For example:
`
_checkVisibility: function(val, minVal, rotated) {
        const size = _abs(val - minVal);
        const startRealValue = Math.min(val, minVal);

        if(rotated) {
            that._calculateVisibility(startRealValue, arg, size, that.height);
        } else {
            that._calculateVisibility(arg, startRealValue, that.width, size);
        }
}
`

Zedwag: done

Krijovnick (approved):

EugeniyKiyashko (approved):

alexanderPolosatov (approved):

PR #25121, opened 2023-07-18
DataGrid - Cell focus in edit mode does not work correctly if a cell has a disabled editor (T1177434)

Raushen:
See https://supportcenter.devexpress.com/ticket/details/t1177434/datagrid-cell-focus-in-edit-mode-does-not-work-correctly-if-a-cell-has-a-disabled-editor

### DIFF
--- a/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
+++ b/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
@@ -1352,10 +1352,7 @@ export class KeyboardNavigationController extends modules.ViewController {
       let $cell = this._getFocusedCell();
       const isEditing = this._editingController.isEditing();
 
-      if (
-        $cell
-        && !(this._isMasterDetailCell($cell) && !this._isRowEditMode())
-      ) {
+      if (!this._isMasterDetailCell($cell) || this._isRowEditMode()) {
         if (this._hasSkipRow($cell.parent())) {
           const direction = this._focusedCellPosition && this._focusedCellPosition.rowIndex > 0
             ? 'upArrow'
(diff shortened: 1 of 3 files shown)


js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1408,6 +1402,21 @@ export class KeyboardNavigationController extends modules.ViewController {
     });
   }
 
+  _needFocusEditingCell() {
```
pomahtri: minor
```suggestion
  private needFocusEditingCell() {
```

js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1408,6 +1402,21 @@ export class KeyboardNavigationController extends modules.ViewController {
...
+    const isCellEditMode = this._editingController.getEditMode() === EDIT_MODE_CELL;
+    const isBatchEditMode = this._editingController.getEditMode() === EDIT_MODE_BATCH;
+
+    const cellEditModeHasChanges = isCellEditMode && this._editingController.hasChanges();
+    const isNewRowBatchEditMode = isBatchEditMode && this._editingController.isNewRowInEditMode();
+    const $cell = this._getFocusedCell();
+
+    return (
+      !isElementDefined($cell)
```
pomahtri: question
why we need to focus cell if it's undefined?

pomahtri (dismissed):

js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
Raushen: It is a technical condition. 1356 string says $cell can be null so the next (1415) string $cell.children() can throw NullReference exception.

pomahtri: I mean, in this case the method will return true, while at first glance it must return false
Maybe change it to `isElementDefined($cell) && ...`, or I am wrong about expected value?

Raushen: I checkedthis. _getFocusedCell() method, it returns $(smth) so it can never be null/false/undefined. I made this condition more simple in the next PR, thank you.

pomahtri (approved):

PR #25132, opened 2023-07-20
Accessibility: DataGrid Editing - Add aria-label attributes for editors

Raushen:
Make embedded editors accessible


marker-dao (dismissed):

vladaskorohodova (dismissed):

packages/devextreme/js/__internal/grids/grid_core/column_headers/m_column_headers.ts
```diff
@@ -211,7 +211,8 @@ export const columnHeadersModule = {
         _setCellAriaAttributes($cell, cellOptions) {
           this.callBase($cell, cellOptions);
           if (cellOptions.rowType === 'header') {
-            if (cellOptions.column.type !== 'groupExpand') {
+            const notHeaderTypes = ['groupExpand', 'buttons'];
```
pomahtri: After creating a new column type we'll probably forget about adding it here, I think it's better to rewrite as mapping ColumnType -> boolean, so we'll get runtime error in tests

```
const isHeaderType = {
  'groupExpand': false,
  'buttons': false,
  'detailExpand': true, // (?)
  ... // all other column types
}
``` 

Or we can spend a little time typing it, and get compile-time error instead of runtime: [playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwg9gGwK4FsB2AVcEDOUC8UA5AEZLDBxo5FQA+xA5gE5xJgCiAHmAIZoATWgyI44KCAHlgACwjN4ydFkhEA3ACgNAYyo5gUAJY4AEhF4D5KiAC4oAJQi7mAgDyLUmbDgA0UEnCI5mgAfARQAN4aUDH+5JTUdgBmvAg4ED7RsSxsnDz8Asmp6ZmxUAD05VBiEtJyCoie1nbAzEgZFVXyrMxGSVAA7tBJcMyMEAaUUBYCUGgQA1C6SmhQoJAaAL5AA)

pomahtri: irrelevant now

pomahtri (approved):

marker-dao (approved):

vladaskorohodova (approved):

PR #25149, opened 2023-07-21
Chart: bar should be rendered when value is on border, refactor (T1176892)

Zedwag:


packages/devextreme/js/viz/series/points/bar_point.js
```diff
@@ -51,14 +51,14 @@ export default _extend({}, symbolPoint, {
         this.inVisibleArea = minX <= x + width && maxX >= x && minY <= y + height && maxY >= y;
     },
 
-    _checkVisibility: function(arg, val, minVal, rotated) {
-        const size = _abs(val - minVal);
-        const valueBeforeCorrection = Math.min(val, minVal);
+    _cacheVisibility: function(x, y, minY, rotated) {
+        y = Math.min(y, minY);
```
Krijovnick, line 55: Function should be pure because it lead to side effects

Zedwag: I don't see any side effects it may cause

Krijovnick (approved):

PR #25159, opened 2023-07-25
DataGrid: Fix the error thrown when editing is enabled while the column chooser is visible (T1178698)

markallenramirez:


js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts
```diff
@@ -563,7 +563,7 @@ export const updateColumnChanges = function (that: ColumnsController, changeType
   }
   if (columnIndex === undefined || columnIndex !== columnChanges.columnIndex) {
     if (isDefined(columnIndex)) {
-      columnChanges.columnIndices ??= [columnChanges.columnIndex];
+      columnChanges.columnIndices ??= isDefined(columnChanges.columnIndex) ? [columnChanges.columnIndex] : [];
```
Tucchhaa: I agree with your solution. But I think here if `columnChanges` has columnIndices, then it means that there are at least two columns indices.  

And with this logic, it can be broken. For example, if `columnChanges.columnIndex` is `undefined` and `columnIndex` is not, then `columnChanges.columnIndices` will contain only one element. So in this case `columnChanges` should have `columnIndex` property

Tucchhaa (changes requested):

js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts
Tucchhaa: I have tried to implement such logic to make a suggestion, but it took too much time for me( It is a little bit difficult to implement it

Tucchhaa: Let's just make a separate if statement, to make it a little more readable
```suggestion
      columnChanges.columnIndices ??= [];
      
      if(isDefined(columnChanges.columnIndex)) {
        columnChanges.push(columnChanges.columnIndex)
      }
```

js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts
```diff
@@ -563,7 +563,12 @@ export const updateColumnChanges = function (that: ColumnsController, changeType
   }
   if (columnIndex === undefined || columnIndex !== columnChanges.columnIndex) {
     if (isDefined(columnIndex)) {
-      columnChanges.columnIndices ??= [columnChanges.columnIndex];
+      columnChanges.columnIndices ??= [];
+
+      if (isDefined(columnChanges.columnIndex)) {
+        columnChanges.push(columnChanges.columnIndex);
```
Tucchhaa: That's my mistake. Change to this:
```suggestion
        columnChanges.columnIndices.push(columnChanges.columnIndex);
```
I guess tests have fallen because of it

Tucchhaa (approved):

js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts
markallenramirez: Yep, I did not notice it. Thank you.

PR #25178, opened 2023-07-26
DataGrid: Throw a warning if exporting is enabled and the onExporting function is not specified

markallenramirez:


js/__internal/grids/data_grid/export/m_export.ts
```diff
@@ -608,6 +609,10 @@ export class ExportController extends dataGridCore.ViewController {
   }
 
   init() {
```
Raushen, line 611: I think we should check the same on _optionChanged too.

williamvinogradov (approved):

vladaskorohodova (approved):

PR #25204, opened 2023-07-27
TreeList - Unable to focus a node when deleting the previous node in certain scenarios (T1178893)

Raushen:
See https://supportcenter.devexpress.com/ticket/details/t1178893/treelist-unable-to-focus-a-node-when-deleting-the-previous-node-in-certain-scenarios


packages/devextreme/js/ui/gantt/ui.gantt.treelist.nodes_state.js
```diff
@@ -33,6 +33,7 @@ export class GanttTreeListNodesState {
         this._resetHash();
     }
     applyNodes(nodes, rootValue) {
+        nodes = nodes.filter(node => node);
```
Alyar666: Doubtful place for this code. Maybe discuss this with the Gantt team?

Alyar666 (changes requested):

packages/devextreme/js/ui/gantt/ui.gantt.treelist.nodes_state.js
williamvinogradov: Yup, I agree with Alyar

Also, let's don't override the function arguments
For cases like this, it's better to initialize a separate local variable :)

packages/devextreme/js/ui/gantt/ui.gantt.treelist.js
```diff
@@ -239,7 +239,13 @@ export class GanttTreeList {
         const treeList = this._treeList;
         const store = treeList?.getDataSource().store();
         const keyGetter = compileGetter(store?.key());
-        return store ? store._array.map(item => treeList.getNodeByKey(keyGetter(item))) : [];
+        let nodes = [];
```
williamvinogradov: Let's make it more obvious and get rid of the local redundant ```treeList``` variable:

```javascript
getAllNodes() {
  const store = this.treeList?.getDataSource()?.store();
  
  if (!store || !this.treeList?.getNodeByKey) {
    return [];
  }
  
  const keyGetter = compileGetter(store.key());
  return store._array
    .map((item) => this.treeList.getNodeByKey(keyGetter(item)))
    .filter((item) => !!item);
}
```

williamvinogradov (dismissed):

Alyar666 (dismissed):

Alyar666 (approved):

PR #25210, opened 2023-07-28
DataGrid - Component sends unexpected filtering request after inserting a new row if focusedRowEnabled is true (T1181477)

Raushen:
See https://supportcenter.devexpress.com/ticket/details/t1178543/datagrid-field-id-expected-a-number-but-got-string-error-occurs-on-inserting-a-new-row


packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_utils.ts
```diff
@@ -1,5 +1,9 @@
+import Guid from '@js/core/guid';
 import { isObject } from '@js/core/utils/type';
 
+const NEW_ROW_TEMP_KEY_PREFIX = '_DX_NEW_ROW_TEMP_KEY_';
```
williamvinogradov: Not a comment, just thinking out loud

Maybe it will be better to make this prefix a little shorter.
E.g:
```suggestion
const NEW_ROW_TEMP_KEY_PREFIX = '_DX_TMP_';
```

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_utils.ts
```diff
@@ -31,6 +35,16 @@ export const getEditingTexts = (options) => {
...
 
+export const generateNewRowTempKey = function (): string {
+  return NEW_ROW_TEMP_KEY_PREFIX + String(new Guid());
+};
+export const isNewRowTempKey = function (key: string): boolean {
+  return !!key
+    && typeof key === 'string'
+    && key.startsWith(NEW_ROW_TEMP_KEY_PREFIX)
+    && key.length === NEW_ROW_TEMP_KEY_PREFIX.length + GUID_LENGTH;
+};
```
williamvinogradov: Let's use arrow functions here and add the space between them.
And we can use string templates instead of string concatenation:

```suggestion
export const generateNewRowTempKey = () => 
  `${NEW_ROW_TEMP_KEY_PREFIX}${new Guid()}`;
  
export const isNewRowTempKey = (key: string) => 
    typeof key === 'string'
    && key.startsWith(NEW_ROW_TEMP_KEY_PREFIX)
    && key.length === NEW_ROW_TEMP_KEY_PREFIX.length + GUID_LENGTH;
```

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
```diff
@@ -803,6 +800,23 @@ class EditingControllerImpl extends modules.ViewController {
...
+    let keyValue;
+    if (change.data && keyExpr && !Array.isArray(keyExpr)) {
+      keyValue = change.data[keyExpr];
+    }
+    if (!isDefined(keyValue)) {
+      keyValue = generateNewRowTempKey();
+    }
+
+    change.key = keyValue;
+  }
```
williamvinogradov: Let's change this method and don't mutate the ```change.key``` property.
And let's refactor it a little :)

Something like this:
```suggestion
private getChangeKeyValue(change: any): unknown {
    if (isDefined(change.key)) {
      return change.key;
    }

    const keyExpr = this._dataController.key();
    const isKeyValueInChangeData = change.data && keyExpr && !Array.isArray(keyExpr);
    const changeDataKey = isKeyValueInChangeData
      ? change.data[keyExpr]
      : undefined;

    return changeDataKey ?? generateNewRowTempKey();
  }
```

or

```suggestion
  private getChangeKeyValue(change: any): unknown {
    if (isDefined(change.key)) {
      return change.key;
    }
    const keyExpr = this._dataController.key();
    const isKeyValueInChangeData = change.data && keyExpr && !Array.isArray(keyExpr);
    const changeDataKey = isKeyValueInChangeData
      ? change.data[keyExpr]
      : undefined;

    if (!changeDataKey) {
      return generateNewRowTempKey();
    }
    
    return changeDataKey;
  }
```

And we can rewrite the ```change.key``` in the ```_addInsertInfo``` method:

```typescript
// With mutation (old variant):
this.ensureChangeKeyValue(change);

// Without mutation (my suggestion):
change.key = this.getChangeKeyValue(change);
```

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_utils.ts
```diff
@@ -31,6 +35,16 @@ export const getEditingTexts = (options) => {
   };
 };
 
+export const generateNewRowTempKey = function (): string {
+  return NEW_ROW_TEMP_KEY_PREFIX + String(new Guid());
+};
+export const isNewRowTempKey = function (key: string): boolean {
+  return !!key
```
andrewmakarov: Isn't that a redundant check? **typeof key === 'string'** quite enough

williamvinogradov: The ```typeof key === 'string'``` pass empty strings
As far as I know, we don't want to pass empty strings here :)

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
```diff
@@ -780,12 +781,8 @@ class EditingControllerImpl extends modules.ViewController {
 
   _addInsertInfo(change, parentKey?) {
     let insertInfo;
-    let { key } = change;
-
-    if (!isDefined(key)) {
-      key = String(new Guid());
-      change.key = key;
-    }
+    this.ensureChangeKeyValue(change);
```
andrewmakarov: I would suggest that **ensureChangeKeyValue** method doesn't modify argument, but return a value

`change.key = this.createKeyValue(change);`

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_utils.ts
williamvinogradov: Oh, sorry I got your thought just right now.
I agree with @andrewmakarov here

In other conditions, we check the string length and prefix.
So, the empty string check looks redundant.
Let's remove it :)

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
```diff
@@ -780,12 +781,8 @@ class EditingControllerImpl extends modules.ViewController {
...
+    const { key } = change;
```
andrewmakarov, line 791: Such destructuring maybe overskill?

Raushen: If we avoid destructuring the following constructions appear:

```
this._addInternalData({ insertInfo, key: change.key });
return { insertInfo, key: change.key };
```

instead of current:
```
this._addInternalData({ insertInfo, key });
return { insertInfo, key };
```

williamvinogradov (approved):

PR #25236, opened 2023-08-01
Slider: remove unnecessary `aria-describedby` attribute and add `aria-label` attribute

Krijovnick:


packages/devextreme/js/ui/tooltip/tooltip.js
```diff
@@ -86,6 +80,16 @@ const Tooltip = Popover.inherit({
         if(!isWindow($target.get(0))) {
             this.setAria('describedby', label, $target);
         }
+    },
+
+    _toggleAriaAttributes: function() {
+        this._contentId = 'dx-' + new Guid();
```
alexanderPolosatov: let's use string template instead plus?

alexanderPolosatov (approved):

PR #25270, opened 2023-08-04
Gallery: add border on focus when indicator is disabled

Zedwag:


packages/devextreme/scss/widgets/generic/gallery/_index.scss
```diff
@@ -105,4 +105,8 @@
   .dx-gallery-indicator-item-selected {
     background: $gallery-indicator-focused-bg;
   }
+
+  &:not(:has(.dx-gallery-indicator)) {
```
AlexanderMoiseev: does firefox support has selector?

Zedwag: reworked

packages/devextreme/js/ui/gallery.js
```diff
@@ -263,12 +263,20 @@ const Gallery = CollectionWidget.inherit({
         return this.option('rtlEnabled') ? -1 : 1;
     },
 
+    _setBorderWidth: function() {
+        // TODO: Get rid of this code and use :has() selector when Firefox supports it
```
EugeniyKiyashko: Let's use class instead of inline styles.

packages/devextreme/scss/widgets/generic/gallery/_index.scss
```diff
@@ -102,7 +102,10 @@
 }
 
 .dx-state-focused.dx-gallery {
+  border: solid 1px $base-accent;
```
EugeniyKiyashko: Let split it to border-width, border-color, border-style

marker-dao (approved):

packages/devextreme/scss/widgets/generic/gallery/_index.scss
```diff
@@ -102,6 +102,14 @@
 }
 
 .dx-state-focused.dx-gallery {
+  border-width: 1px;
```
AlexanderMoiseev, line 105: same styles for diff themes, mb we can use some kind of mixin here?

Zedwag: as far as border color is defined for different themes separetely, I don't think we should do this

we can move only border-width and border-style to common styles, but I don't think it's a good idea to separate styles of the same border

AlexanderMoiseev (approved):

PR #25272, opened 2023-08-04
Popup: place ‘Cancel’ button at the left of the ‘Done’ button

alexanderPolosatov:


packages/devextreme/js/ui/popup/ui.popup.js
```diff
@@ -491,15 +497,17 @@ const Popup = Overlay.inherit({
...
+            if(item?.toolbar === toolbar) {
                 if(isShortcut) {
                     extend(item, { location: data.location }, this._getToolbarItemByAlias(data));
-                }
-
-                const isLTROrder = currentPlatform === 'generic';
-
-                if((data.shortcut === 'done' && isLTROrder) || (data.shortcut === 'cancel' && !isLTROrder)) {
-                    toolbarsItems.unshift(item);
+                    if(data.shortcut === 'done' || data.shortcut === 'cancel') {
```
marker-dao: You can use ACTION_BUTTONS_ORDER const here for not using a simple string:

`ACTION_BUTTONS_ORDER.includes(data.shortcut)`

Aggregate simple strings to arrays or objects looks better to me. But it is minor problem.

packages/devextreme/js/ui/popup/ui.popup.js
```diff
@@ -470,19 +479,16 @@ const Popup = Overlay.inherit({
...
-
         const toolbarItems = this.option('toolbarItems');
-
         const toolbarsItems = [];
-
         this._toolbarItemClasses = [];
-
         const currentPlatform = devices.current().platform;
         let index = 0;
+        const doneCancelButtonsInfo = [];
```
ksercs:
```suggestion
        const applyValueButtonsInfo = [];
```

packages/devextreme/js/ui/popup/ui.popup.js
```diff
@@ -491,15 +497,17 @@ const Popup = Overlay.inherit({
...
 
-            if(item && item.toolbar === toolbar) {
+            if(item?.toolbar === toolbar) {
                 if(isShortcut) {
                     extend(item, { location: data.location }, this._getToolbarItemByAlias(data));
-                }
-
-                const isLTROrder = currentPlatform === 'generic';
-
-                if((data.shortcut === 'done' && isLTROrder) || (data.shortcut === 'cancel' && !isLTROrder)) {
```
ksercs, line 501: do we ignore RTL now?

packages/devextreme/js/ui/popup/ui.popup.js
```diff
@@ -510,7 +518,9 @@ const Popup = Overlay.inherit({
             toolbarsItems.push(this._getCloseButton());
         }
 
-        return toolbarsItems;
+        const sortedActionItems = sortActionButtonsItems(doneCancelButtonsInfo).map(item => item.item);
```
ksercs: Naming is misleading for me. It seems like we name similar instances different way: "actionItems", "toolbarItems", "actionButtons", "doneCancelButtons", "applyValueButtons". Can we unify the approaches?

packages/devextreme/js/ui/popup/ui.popup.js
```diff
@@ -510,7 +518,9 @@ const Popup = Overlay.inherit({
...
+
+        return toolbarsItems.concat(...sortedActionItems);
```
ksercs: So, cancel and apply buttons are always placed in the end now? I am not sure about this because user should have an opportunity to specify the order. I thought that we should change only the default order.

packages/devextreme/js/ui/popup/ui.popup.js
alexanderPolosatov: no. There is wrong name of variable. it's not a check on RTL. it's check on theme. right now we have decided what order of buttons should not depends on themes

EugeniyKiyashko (approved):

ksercs (approved):

PR #25291, opened 2023-08-08
Calendar - implement selectWeekOnClick option

Zedwag:


packages/devextreme/js/ui/calendar/ui.calendar.base_view.js
```diff
@@ -237,6 +242,23 @@ const BaseView = Widget.inherit({
                 }
             }));
         }
+
+        if(selectionMode !== 'single') {
+            this._createWeekNumberCellClickAction();
+
+            eventsEngine.on(this._$table, CALENDAR_DXCLICK_EVENT_NAME, `.${CALENDAR_WEEK_NUMBER_CELL_CLASS}`, ((e) => {
+                const $row = $(e.currentTarget).parent();
```
EugeniyKiyashko: We should use `.closest()` in this case. Customer can use template inside cell or click on inner `span` element inside `td`

EugeniyKiyashko (approved):

PR #25300, opened 2023-08-09
DataGrid/TreeList - The 'container' parameter in a template's render function should be an HtmlElement when integrating a 3rd-party template engine (T1181910)

markallenramirez:


packages/devextreme/js/core/utils/template_manager.js
```diff
@@ -53,6 +54,10 @@ export const addOneRenderedCall = (template) => {
     const render = template.render.bind(template);
     return extend({}, template, {
         render(options) {
+            if(options && options.container) {
```
ivanblinov2k17: I don't know, if patching options inside core utils functions is considered ok in DevExtreme, fix seems pretty safe, but we must doublecheck, that it doesn't have additional sideeffects
Also, are you sure, that putting this inside addOneRenderedCall function will cover all the cases, where we need to getPublicElement? For example, if isAsyncTemplate === true, addOneRenderedCall will not be executed

markallenramirez: Nope, I missed the **isAsyncTemplate** case. **addOneRenderedCall** is indeed not called in case of **isAsyncTemplate === true**. I used a different solution.

markallenramirez: > I used a different solution

This solution is complex. It is decided to fix the issue in the template manager's side.

ivanblinov2k17 (approved):

pomahtri (approved):

PR #25313, opened 2023-08-10
DataGrid Accessibility  - Add contrast tests for all themes

Alyar666:
Added contrast tests for the following themes:

- generic.light
- generic.dark
- material.blue.light
- material.blue.dark


packages/devextreme/testing/testcafe/helpers/accessibilityUtils.ts
```diff
@@ -1,14 +1,20 @@
...
+const defaultOptions = {
+  rules: {
+    'color-contrast': { enabled: false },
+    'th-has-data-cells': { enabled: false },
+  },
+};
+
+export const a11yCheck = async (
+  t: TestController,
+  selector: any = undefined,
```
Raushen: Should we create other overloaded method instead of editing the current one? I suggest set selector argument as the third in argument order.

packages/devextreme/testing/testcafe/helpers/accessibilityUtils.ts
```diff
@@ -1,14 +1,20 @@
...
+  options: any = defaultOptions,
+):
+Promise<void> => {
```
williamvinogradov: Let's add more strict types here and make the optional ```selector``` parameter the last one.
Plus add to the ```selector``` default value.
Something like this:

```suggestion
type A11yCheckOptions: {
  rules: Record<string, { enabled: boolean }>
}

export const a11yCheck = async (
  t: TestController,
  options: A11yCheckOptions = defaultOptions,
  selector: ElementContext | undefined = undefined,
):
Promise<void> => {
```

packages/devextreme/testing/testcafe/helpers/accessibilityUtils.ts
Alyar666: What other method overload are you talking about?

Raushen: I'm sorry, it is C# way) TS requires optional selector parameter instead as williamvinogradov wrote.

Raushen (approved):

PR #25324, opened 2023-08-11
DataGrid: Fix extra page load after changing the css Display property while changing the cell value (T1089634)

Alyar666:
See https://devexpress.com/issue=T1089634


js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -599,13 +601,13 @@ const resizingControllerMembers = {
   },
   _checkSize(checkSize) {
```
williamvinogradov, line 602: Question.

Why do we need the ```checkSize``` argument here?
As far as I can see, if we pass ```checkSize: false``` the ```_checkSize``` function returns false o_O
Maybe we can don't call the ```_checkSize``` function in these cases at all?

js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -599,13 +601,13 @@ const resizingControllerMembers = {
...
     const $rootElement = this.component.$element();
+    const isWidgetVisible = $rootElement.is(':visible');
 
-    if (checkSize && (
-      this._lastWidth === getWidth($rootElement)
+    if ((checkSize
```
williamvinogradov: Let's refactor this condition a little:

```typescript
const isWidgetVisible = $rootElement.is(':visible');
const isScreenSizeNotChanged= this._lastWidth === getWidth($rootElement)
  && this._lastHeight === getHeight($rootElement)
  && this._devicePixelRatio === getWindow().devicePixelRatio;
 
 return (checkSize && isScreenSizeNotChanged) || !isWidgetVisible;
 
```

testing/testcafe/tests/dataGrid/scrolling.ts
```diff
@@ -1044,6 +1044,78 @@ safeSizeTest('The scroll position of a fixed table should be synchronized with t
...
+
+  // act
+  await dataGrid.scrollTo(t, { y: 500 });
+
+  // assert
+  await t
+    .expect(dataGrid.isReady())
+    .ok();
+
+  await takeScreenshot('T1089634-virtual-scrolling-1.png', '#container');
```
williamvinogradov: Let's replace selector with ```dataGrid.element```

```suggestion
  await takeScreenshot('T1089634-virtual-scrolling-1.png', 'dataGrid.element');
```

testing/testcafe/tests/dataGrid/scrolling.ts
```diff
@@ -1044,6 +1044,78 @@ safeSizeTest('The scroll position of a fixed table should be synchronized with t
...
+  await t
+    .expect(dataGrid.getDataCell(19, 1).isModified)
+    .ok();
+
+  // act - simulate click on button
+  await dataGrid.hide();
+  await t.click(Selector('body'));
+  await t.wait(100);
+
+  await takeScreenshot('T1089634-virtual-scrolling-2.png', '#container');
```
williamvinogradov:
```suggestion
  await takeScreenshot('T1089634-virtual-scrolling-2.png', dataGrid.element);
```

testing/testcafe/tests/dataGrid/scrolling.ts
```diff
@@ -1044,6 +1044,78 @@ safeSizeTest('The scroll position of a fixed table should be synchronized with t
...
+
+  // act
+  await dataGrid.show();
+
+  await takeScreenshot('T1089634-virtual-scrolling-3.png', '#container');
```
williamvinogradov:
```suggestion
  await takeScreenshot('T1089634-virtual-scrolling-3.png', dataGrid.element);
```

williamvinogradov:
Question.

Should we add test cafe test cases that will validate the count of requests?
Because in the initial bug report, we have an issue with extra redundant requests.
Or existing test case is enough?

williamvinogradov (approved):

PR #25394, opened 2023-08-21
DataGrid - FilterPanel - The dropdown of the editor for 'Is any of' and 'Is none of' is not shown if it is opened using the editor's field for the second time (T1182854)

pomahtri:


packages/devextreme/js/__internal/grids/grid_core/header_filter/m_header_filter.ts
```diff
@@ -345,7 +346,10 @@ const HeaderFilterController = modules.ViewController.inherit((function () {
               columnsController.columnOption(options.dataField, 'headerFilter', headerFilterByColumn, true);
             });
           },
-          onHidden: () => restoreFocus(this),
+          onHidden: () => {
+            onHidden?.();
```
williamvinogradov: Let's rename inner ```onHidden``` because two ```onHidden``` called inside each other looks very confusing :)
Maybe the ```baseOnHidden``` / ```previousOnHidden``` and etc.
Some name that you like but different from parent ```onHidden```

packages/devextreme/testing/testcafe/tests/dataGrid/filterPanel.ts
```diff
@@ -0,0 +1,41 @@
...
+
+fixture.disablePageReloads`filterPanel`
+  .page(url(__dirname, '../container.html'));
+
+const POPUP_CLASS = '.dx-popup';
+const INVISIBLE_CLASS = '.dx-state-invisible';
+
+function getVisiblePopups(): Selector {
+  return Selector(`${POPUP_CLASS}:not(${INVISIBLE_CLASS})`);
+}
```
williamvinogradov: Question.

Maybe we should add this code to the existing popup model as a static method? 
It's live here: packages/devextreme/testing/testcafe/model/popup.ts

In this case other developers can use it in their test cases :)

pomahtri: I think we shouldn't 
I wrote this code because writing good test models would require a lot of time. Ideally there should be 

```js
t.expect(filterBuilder.getValueText(0).getEditor().getPopup().visible()).ok()
```

instead of checking count of all popups. I don't want to persuade other developers to write bad code :)

packages/devextreme/testing/testcafe/tests/dataGrid/filterPanel.ts
```diff
@@ -0,0 +1,41 @@
...
+  await t.expect(getVisiblePopups().count).eql(1); // only filterPanel popup
+
+  await t.click(filterBuilder.getField().getValueText());
+  await t.expect(getVisiblePopups().count).eql(2); // filterPanel popup + editor popup
+
+  await t.click(filterBuilder.getField().getValueText());
+  await t.expect(getVisiblePopups().count).eql(1); // only filterPanel popup
+
+  await t.click(filterBuilder.getField().getValueText());
+  await t.expect(getVisiblePopups().count).eql(2); // filterPanel popup + editor popup
```
markallenramirez: Maybe use a visual test instead? Multiple popups can be visible at the same time, and a popup can overlap other popups. For instance, the editor popup can be hidden behind the filterPanel popup although both popups are visible. Though, I understand that it will not happen in this usage scenario.

pomahtri: I'm afraid visual test could be unstable, so I think for now it's better to leave this test as is

Tucchhaa: Why do think test can be unstable? If it is because of popup appear animation you could use t.wait(500).

Raushen (approved):

markallenramirez (approved):

PR #25411, opened 2023-08-22
TabPanel: Add iconPosition option

marker-dao:


EugeniyKiyashko (dismissed):

Zedwag (dismissed):

vladaskorohodova (dismissed):

packages/devextreme/js/ui/tab_panel.d.ts
```diff
@@ -18,6 +18,7 @@ import {
 
 import {
     Position,
+    TabsIconPosition,
```
IlyaKhD, line 21: Add reexport:

```ts
export {
    TabsIconPosition,
} from '../common';
```

marker-dao: I'm not sure that we should export this type separately. We export the entire dxTabPanelOptions interface.

IlyaKhD: It's the rule for all components - all public types should be reexported

marker-dao: Was updated here [e20fbd7cff57e643bde9fe631f671be9dbb154a4].

EugeniyKiyashko (approved):

IlyaKhD (approved):

PR #25453, opened 2023-08-28
Calendar - values option should have 'default' prefix in React

Zedwag:


Zedwag:
decided to get rid of values option

PR #25491, opened 2023-08-31
Fix "Invalid value used as weak map key" error (T1186521)

GoodDayForSurf:

### DIFF
--- a/packages/devextreme/js/events/core/events_engine.js
+++ b/packages/devextreme/js/events/core/events_engine.js
@@ -4,7 +4,7 @@ import domAdapter from '../../core/dom_adapter';
 import { getWindow, hasWindow } from '../../core/utils/window';
 const window = getWindow();
 import injector from '../../core/utils/dependency_injector';
-import { isWindow, isFunction, isString } from '../../core/utils/type';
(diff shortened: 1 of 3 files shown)


ivanblinov2k17 (dismissed):

packages/devextreme/js/events/core/events_engine.js
```diff
@@ -113,7 +113,11 @@ const eventsEngine = injector({
         if(!noBubble) {
             const parents = [];
             const getParents = function(element) {
-                const parent = element.parentNode ?? element.host;
+                const parent = element.parentNode
+                    ?? (typeof element.host === 'object'
```
OlegKipchatov: We can use our utils for this: https://github.com/DevExpress/DevExtreme/blob/23_2/packages/devextreme/js/core/utils/type.js#L47
Also, I think we can't use different logic for same issue: https://github.com/DevExpress/DevExtreme/blob/23_2/packages/devextreme/js/core/utils/dom.js#L17

GoodDayForSurf: isObject() - is good

OlegKipchatov (approved):

dxvladislavvolkov (approved):

PR #25509, opened 2023-09-04
Fluent: DataGrid/TreeList - Set border radius

markallenramirez:


packages/devextreme/js/__internal/grids/grid_core/m_modules.ts
```diff
@@ -256,6 +256,36 @@ const View: ModuleType<ViewType> = ModuleItem.inherit({
...
+    const BORDERED_TOP_VIEW_CLASS = 'dx-bordered-top-view';
+    const BORDERED_BOTTOM_VIEW_CLASS = 'dx-bordered-bottom-view';
+
+    const BORDERED_TOP_VIEW_NAMES = ['columnHeadersView', 'rowsView'];
+    const BORDERED_BOTTOM_VIEW_NAMES = ['rowsView', 'footerView', 'filterPanelView'];
+
+    const firstVisibleTopViewName = this.getFirstVisibleView(BORDERED_TOP_VIEW_NAMES);
+    const lastVisibleBottomViewName = this.getFirstVisibleView(BORDERED_BOTTOM_VIEW_NAMES, true);
+
+    each(BORDERED_TOP_VIEW_NAMES, (index, viewName) => {
```
andrewmakarov: Why use each function, why not use the standard array iteration?

markallenramirez: No specific reason actually. In any case, I need to refactor again. So, I'll use standard methods instead if needed.

Raushen (approved):

PR #25510, opened 2023-09-04
Add license checking functionality

iBat:


packages/devextreme/js/core/component.js
```diff
@@ -82,6 +99,32 @@ export const Component = Class.inherit({
...
+            const licenseToken = Config().license;
+            if(licenseToken) {
+                const license = parseToken(licenseToken);
+
+                if(license.kind === 'corrupted') {
+                    if(license.error === 'version') {
+                        errors.log(invalidVersionLicenseMessage);
+                    } else {
+                        errors.log(invalidFormatLicenseMessage);
+                    }
```
IlyaKhD: `version` in this context means token format version, not devextreme version
```suggestion
                    errors.log(invalidFormatLicenseMessage);
```

packages/devextreme/js/core/__tests__/license.test.ts
```diff
@@ -0,0 +1,45 @@
+import { verifyLicense } from '../component';
```
ksercs: [base approach]
It seems like u got a testing format from renovation: created a new `__tests__` directory, used ts, jest and so on.
We don't use this approach in a usual development. All our tests are placed inside the `testing` directory and we use QUnit for this. Tests for `component` are placed in `testing/tests/DevExpress.Core/component.tests.js`: [link](https://github.com/DevExpress/DevExtreme/blob/23_2/packages/devextreme/testing/tests/DevExpress.core/component.tests.js). You can create a separate testing module in this file and test this new functionality.

packages/devextreme/js/core/__tests__/license.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const token231 = 'ewogICJmb3JtYXQiOiAxLAogICJjdXN0b21lcklkIjogImIxMTQwYjQ2LWZkZTEtNDFiZC1hMjgwLTRkYjlmOGU3ZDliZCIsCiAgIm1heFZlcnNpb25BbGxvd2VkIjogMjMxCn0=.DiDceRbil4IzXl5av7pNkKieyqHHhRf+CM477zDu4N9fyrhkQsjRourYvgVfkbSm+EQplkXhlMBc3s8Vm9n+VtPaMbeWXis92cdW/6HiT+Dm54xw5vZ5POGunKRrNYUzd9zTbYcz0bYA/dc/mHFeUdXA0UlKcx1uMaXmtJrkK74=';
```
ksercs: [code style]
We usually use UPPER_SNAKE_CASE for constants
```suggestion
const TOKEN_23_1 = 'ewogICJmb3JtYXQiOiAxLAogICJjdXN0b21lcklkIjogImIxMTQwYjQ2LWZkZTEtNDFiZC1hMjgwLTRkYjlmOGU3ZDliZCIsCiAgIm1heFZlcnNpb25BbGxvd2VkIjogMjMxCn0=.DiDceRbil4IzXl5av7pNkKieyqHHhRf+CM477zDu4N9fyrhkQsjRourYvgVfkbSm+EQplkXhlMBc3s8Vm9n+VtPaMbeWXis92cdW/6HiT+Dm54xw5vZ5POGunKRrNYUzd9zTbYcz0bYA/dc/mHFeUdXA0UlKcx1uMaXmtJrkK74=';
```

packages/devextreme/js/core/__tests__/license.test.ts
```diff
@@ -0,0 +1,45 @@
...
+const tokenMissingField2 = 'ewogICJjdXN0b21lcklkIjogImIxMTQwYjQ2LWZkZTEtNDFiZC1hMjgwLTRkYjlmOGU3ZDliZCIsCiAgIm1heFZlcnNpb25BbGxvd2VkIjogMjMxCn0=.ok32DBaAgf3ijLmNQb+A0kUV2AiSivqvZJADdF607qqlAaduAVnotJtgdwm/Ib3MErfaGrDohCYoFMnKQevkRxFkA7tK3kOBnTZPUnZY0r3wyulMQmr4Qo+Sjf/fyXs4IYpGsC7/uJjgrCos8uzBegfmgfM93XSt6pKl9+c5xvc=';
+const tokenMissingField3 = 'ewogICJmb3JtYXQiOiAxLAogICJjdXN0b21lcklkIjogImIxMTQwYjQ2LWZkZTEtNDFiZC1hMjgwLTRkYjlmOGU3ZDliZCIKfQ==.resgTqmazrorRNw7mmtV31XQnmTSw0uLEArsmpzCjWMQJLocBfAjpFvKBf+SAG9q+1iOSFySj64Uv2xBVqHnyeNVBRbouOKOnAB8RpkKvN4sc5SDc8JAG5TkwPVSzK/VLBpQxpqbxlcrRfHwz9gXqQoPt4/ZVATn285iw3DW0CU=';
+const tokenUnsupportedVersion = 'ewogICJmb3JtYXQiOiAyLAogICJjdXN0b21lcklkIjogImIxMTQwYjQ2LWZkZTEtNDFiZC1hMjgwLTRkYjlmOGU3ZDliZCIsCiAgIm1heFZlcnNpb25BbGxvd2VkIjogMjMxCn0=.tTBymZMROsYyMiP6ldXFqGurbzqjhSQIu/pjyEUJA3v/57VgToomYl7FVzBj1asgHpadvysyTUiX3nFvPxbp166L3+LB3Jybw9ueMnwePu5vQOO0krqKLBqRq+TqHKn7k76uYRbkCIo5UajNfzetHhlkin3dJf3x2K/fcwbPW5A=';
+
+describe('verifyLicense', () => {
+  it('Empty license', () => {
```
ksercs: [naming]
We're trying to name our tests answering on a question "how should it work". For example:
`W0019 error should be logged if license is empty`

packages/devextreme/js/core/component.js
```diff
@@ -21,6 +23,8 @@ const isInnerOption = (optionName) => {
     return optionName.indexOf('_', 0) === 0;
 };
 
+let licenseVerified = false;
```
ksercs: [naming]
```suggestion
let isLicenseVerified = false;
```

packages/devextreme/js/core/component.js
```diff
@@ -385,3 +397,37 @@ export const Component = Class.inherit({
...
+                return 'W0020';
+            }
+        }
+    } else {
+        return 'W0019';
+    }
+}
+
+///#DEBUG
+export function resetLicenseCheckSkipCondition() {
```
ksercs: [code style]
We can get rid of this additional function and add a parameter to the setter below.

packages/devextreme/js/core/component.js
```diff
@@ -385,3 +397,37 @@ export const Component = Class.inherit({
         this.endUpdate();
     }
 });
+
+export function verifyLicense(licenseToken, ver = version) {
```
ksercs: [code splitting]
All this code about license is not related to the component's work directly => I expect all this code is isolated in a separate file in `core/utils` and we leave only call of the `verifyLicense` function here in the component's code. There also should be 2 sets of tests: for the new util (how this util works) and for the component (check that util is used, it's called only once and so on).

ksercs: [copypaste]
Seems like `version` is always passed from the call but it's also always equal to the default value. I suggest you get rid of this parameter. The duplicate name seems to hint that something went wrong :)

packages/devextreme/js/core/component.js
```diff
@@ -82,6 +86,14 @@ export const Component = Class.inherit({
         this._disposingCallbacks = _disposingCallbacks || Callbacks();
         this.postponedOperations = new PostponedOperations();
         this._createOptions(options);
+
+        if(!licenseVerified) {
+            licenseVerified = true;
+            const licenseError = verifyLicense(Config().license, version);
+            if(licenseError) {
+                errors.log(licenseError);
```
ksercs: [code splitting]
I don't see a reason why we should log error here in a components code or check that license was already verified. Let's move all this actions in util.

packages/devextreme/js/core/component.js
```diff
@@ -385,3 +397,37 @@ export const Component = Class.inherit({
...
+export function verifyLicense(licenseToken, ver = version) {
+    if(licenseToken) {
+        const license = parseToken(licenseToken);
+
+        if(license.kind === 'corrupted') {
+            return 'W0021';
+        } else {
+            const [majorStr, minorStr] = ver.split('.');
+            const major = parseInt(majorStr, 10);
+            const minor = parseInt(minorStr, 10);
```
ksercs: [code style]
We can change this 3 lines to:
```javascript
const [major, minor] = version.split('.').map(Number);
```

packages/devextreme/js/core/errors.js
```diff
@@ -249,5 +249,27 @@ export default errorUtils({
...
+    W0018: 'Setting the "position" property with a function is deprecated since v21.2',
+    /**
+    * @name ErrorsCore.W0019
+    */
+    W0019: 'DevExtreme: Valid license key not found.\n\n' +
+        'If you are using a trial version, you must uninstall all copies of DevExtreme once your 30 days trial period expires. For licensing-related information, please refer to the DevExtreme End User License Agreement: https://js.devexpress.com/EULAs/DevExtremeComplete/.\n\n' +
+        'To continue using DevExtreme in a project, you must purchase a license. For pricing/licensing options, please visit: https://js.devexpress.com/Buy/.\n\n' +
+        'If you have licensing-related questions or need help with a purchase, please email clientservices@devexpress.com. We will be happy to follow-up.',
+    /**
+     * @name ErrorsCore.W0019
```
ksercs: [error]
```suggestion
     * @name ErrorsCore.W0020
```

packages/devextreme/js/core/errors.js
```diff
@@ -249,5 +249,27 @@ export default errorUtils({
...
+     */
+    W0020: 'DevExtreme: The license key is expired.\n\n' +
+        'A mismatch exists between license key/DevExtreme version.\n\n' +
+        'To proceed, you can:\n' +
+        ' ● use a version of DevExtreme linked to your license key https://www.devexpress.com/ClientCenter/DownloadManager/\n' +
+        ' ● renew your DevExpress Subscription (once you renew your subscription, you will be entitled to product updates and support service https://www.devexpress.com/buy/renew/)\n\n' +
+        'If you have licensing-related questions or need help with a renewal, please email clientservices@devexpress.com. We will be happy to follow-up.',
+    /**
+     * @name ErrorsCore.W0019
```
ksercs: [error]
```suggestion
     * @name ErrorsCore.W0021
```

packages/devextreme/testing/tests/DevExpress.core/core.licensing.js
```diff
@@ -0,0 +1,64 @@
+import config from 'core/config';
```
ksercs: [naming]
`core` is excess in the file's name. When u move your feature's code to a separate util, you can name this similar.

packages/devextreme/testing/tests/DevExpress.core/core.licensing.js
```diff
@@ -0,0 +1,64 @@
...
+
+const validToken_23_2 = 'ewogICJmb3JtYXQiOiAxLAogICJjdXN0b21lcklkIjogIjYxMjFmMDIyLTFjMTItNDNjZC04YWE0LTkwNzJkNDU4YjYxNCIsCiAgIm1heFZlcnNpb25BbGxvd2VkIjogMjMyCn0=.RENyZ3Ga5rCB7/XNKYbk2Ffv1n9bUexYNhyOlqcAD02YVnPw6XyQcN+ZORScKDU9gOInJ4o7vPxkgh10KvMZNn+FuBK8UcUR7kchk7z0CHGuOcIn2jD5X2hG6SYJ0UCBG/JDG35AL09T7Uv/pGj4PolRsANxtuMpoqmvX2D2vkU=';
+const validToken_22_1 = 'ewogICJmb3JtYXQiOiAxLAogICJjdXN0b21lcklkIjogIjM3Yjg4ZjBmLWQ0MmMtNDJiZS05YjhkLTU1ZGMwYzUzYzAxZiIsCiAgIm1heFZlcnNpb25BbGxvd2VkIjogMjIxCn0=.NVsilC5uWlD5QGS6bocLMlsVVK0VpZXYwU2DstUiLRpEI79/onuR8dGWasCLBo4PORDHPkNA/Ej8XeCHzJ0EkXRRZ7E2LrP/xlEfHRXTruvW4IEbZt3LiwJBt6/isLz+wzXtYtjV7tpE07/Y0TFoy+mWpHoU11GVtwKh6weRxkg=';
+
+class TestComponent extends Component {
+    constructor(options) {
+        super(options);
+        this.NAME = 'TestComponent';
+    }
+}
```
ksercs: [excess]
Why do we need this `TestComponent`? Seems like we can just test a usual component

packages/devextreme/testing/tests/DevExpress.core/core.licensing.js
```diff
@@ -0,0 +1,64 @@
+import config from 'core/config';
+import errors from 'core/errors';
+import { Component, resetLicenseCheckSkipCondition } from 'core/component';
```
ksercs: [base approach]
If we test Component - test should be placed in `component.tests.js`

packages/devextreme/testing/tests/DevExpress.core/core.licensing.js
```diff
@@ -0,0 +1,64 @@
...
+}, () => {
+    test('Empty token', function(assert) {
+        config({ license: null });
+        const instance = new TestComponent();
+
+        assert.equal(log.length, 1);
+        assert.strictEqual(log[0][0], 'W0019');
+    });
+
+    test('token should be verified', function(assert) {
```
ksercs: [unclear test]
This test does not check that token is verified. It only checks that no error is logged.

packages/devextreme/testing/tests/DevExpress.exporter/exceljsParts/exceljs.dataGrid.tests.js
```diff
@@ -40,6 +41,7 @@ const moduleConfig = {
     beforeEach: function() {
         this.worksheet = new ExcelJS.Workbook().addWorksheet('Test sheet');
         this.customizeCellCallCount = 0;
+        setLicenseCheckSkipCondition();
```
ksercs: Why do we need this?

packages/devextreme/testing/tests/DevExpress.core/core.licensing.js
```diff
@@ -0,0 +1,64 @@
...
+    beforeEach: function() {
+        resetLicenseCheckSkipCondition();
+        log = [];
+        errors.log = (...args) => {
+            log.push(args);
+        };
+    },
+    afterEach: function() {
+        errors.log = originalLog;
+    }
```
ksercs: [test isolation]
shouldn't we restore `config`?

packages/devextreme/js/core/errors.js
```diff
@@ -249,5 +249,27 @@ export default errorUtils({
...
+     */
+    W0021: 'DevExtreme: License key verification failed.\n\n' +
+        'Make certain to specify a correct key in the GlobalConfig. If you continue to encounter an error, please visit https://www.devexpress.com/ClientCenter/DownloadManager/ to obtain a valid key.\n\n' +
+        'If you have a valid license and this problem persists, please submit a support ticket via the DevExpress Support Center. We will be happy to follow-up: https://supportcenter.devexpress.com/ticket/create',
```
ksercs: shouldn't be dot here in the end?

packages/devextreme/js/core/errors.js
```diff
@@ -249,5 +249,27 @@ export default errorUtils({
...
     */
-    W0018: 'Setting the "position" property with a function is deprecated since v21.2'
+    W0018: 'Setting the "position" property with a function is deprecated since v21.2',
+    /**
+    * @name ErrorsCore.W0019
+    */
+    W0019: 'DevExtreme: Valid license key not found.\n\n' +
+        'If you are using a trial version, you must uninstall all copies of DevExtreme once your 30 days trial period expires. For licensing-related information, please refer to the DevExtreme End User License Agreement: https://js.devexpress.com/EULAs/DevExtremeComplete/.\n\n' +
+        'To continue using DevExtreme in a project, you must purchase a license. For pricing/licensing options, please visit: https://js.devexpress.com/Buy/.\n\n' +
+        'If you have licensing-related questions or need help with a purchase, please email clientservices@devexpress.com. We will be happy to follow-up.',
```
ksercs: Seems like we need to add an additional `\n\n` at the end to separate a default `See: ->link_to_error<-` message with an empty line.

packages/devextreme/testing/tests/DevExpress.core/core.licensing.js
```diff
@@ -0,0 +1,64 @@
...
+let log = [];
+const originalLog = errors.log;
+
+module('License check', {
+    beforeEach: function() {
+        resetLicenseCheckSkipCondition();
+        log = [];
+        errors.log = (...args) => {
+            log.push(args);
+        };
```
ksercs: [base approach]
It's better to use `sinon.stub` instead of all this caches

packages/devextreme/testing/tests/DevExpress.exporter/exceljsParts/exceljs.dataGrid.tests.js
iBat: Because this test also checks the number of calls of the errors.log method and the new functionality interferes with this test.

packages/devextreme/js/core/component.js
```diff
@@ -82,6 +86,11 @@ export const Component = Class.inherit({
         this._disposingCallbacks = _disposingCallbacks || Callbacks();
         this.postponedOperations = new PostponedOperations();
         this._createOptions(options);
+
+        if(!isLicenseVerified) {
+            isLicenseVerified = true;
```
ksercs: [code splitting]
why don't u want incapsulate this singletone to the util?

packages/devextreme/js/core/component.js
```diff
@@ -82,6 +86,11 @@ export const Component = Class.inherit({
...
+            verifyLicense(Config().license, version);
```
ksercs: [copypaste]
I suggest u totally get rid of parameter passing here cause is does not depend from the component state

packages/devextreme/testing/tests/DevExpress.exporter/exceljsParts/exceljs.dataGrid.tests.js
ksercs: I suggest you to make the actual test more concrete instead of adding of additional actions.
```javascript
        this.stub = sinon.stub(errors, 'log').callsFake((errorId) => {
            if(errorId === 'W0003') {
                QUnit.assert.strictEqual(true, false, 'no deprecation warnings should be logged');
            }
        });
```

packages/devextreme/testing/tests/DevExpress.core/utils.license.tests.js
```diff
@@ -0,0 +1,138 @@
...
+
+        verifyLicense('in.put');
+        assert.strictEqual(errors.log.getCall(9).args[0], 'W0021');
+
+        setLicenseCheckSkipCondition(false);
+
+        verifyLicense('3.2.1', '1.2.3');
+        assert.strictEqual(errors.log.getCall(10).args[0], 'W0021');
+    });
+});
```
ksercs: [test coverage]
Now we don't check that error is not logged repeatedly if method is called several times

packages/devextreme/testing/tests/DevExpress.core/utils.license.tests.js
```diff
@@ -0,0 +1,138 @@
...
+        assert.strictEqual(errors.log.getCall(0).args[0], 'W0020');
+
+        setLicenseCheckSkipCondition(false);
+
+        verifyLicense(TOKEN_23_2, '42.4');
+        assert.equal(errors.log.callCount, 2);
+        assert.strictEqual(errors.log.getCall(1).args[0], 'W0020');
+    });
+
+    test('W0021 error should be logged if license is corrupted/invalid', function(assert) {
```
ksercs: [copypaste]

We can rewrite this using forEach to get rid of copypaste of `setLicenseCheckSkipCondition`

```javascript
[
    [TOKEN_UNVERIFIED, '1.2.3'],
    // ...
].forEach((token, version) => {
    verifyLicense(token, version);
    assert.strictEqual(errors.log.lastCall.args[0], 'W0021');
    setLicenseCheckSkipCondition(false);
});
```
P.S. we can use this approach in all tests

ksercs (dismissed):

packages/devextreme/js/core/utils/license.js
```diff
@@ -0,0 +1,43 @@
+import { parseToken } from '../../__internal/core/license/license_validation';
```
williamvinogradov: It's better to develop a new code with TS
We already have a directory in ```js/__internal``` for this feature: [link](https://github.com/DevExpress/DevExtreme/tree/23_2/packages/devextreme/js/__internal/core/license)

Also, let's use types from the ```packages/devextreme/js/__internal/core/license/license_validation.ts```
(See comments below)

packages/devextreme/js/core/utils/license.js
```diff
@@ -0,0 +1,43 @@
...
+        return;
+    }
+    isLicenseVerified = true;
+
+    let warning;
+
+    if(licenseToken) {
+        const license = parseToken(licenseToken);
+
+        if(license.kind === 'corrupted') {
```
williamvinogradov: Comment for future TS :)

Let's move the type of kind to a separate const in ```license_validation.ts```
And use this const here

packages/devextreme/js/core/utils/license.js
```diff
@@ -0,0 +1,43 @@
...
+            warning = 'W0021';
+        } else {
+            const [major, minor] = version.split('.').map(Number);
```
williamvinogradov: Comment for future TS :)

It's better to fix type of ```version``` in the ```license_validation.ts```

packages/devextreme/js/core/utils/license.js
```diff
@@ -0,0 +1,43 @@
...
+    } else {
+        warning = 'W0019';
+    }
+
+    if(warning) {
+        errors.log(warning);
+    }
+}
+
+///#DEBUG
```
williamvinogradov: Can we remove this ```//#DEBUG``` section and replace it with something another?
(E.g. mock or something)

We want to get rid of this mechanism in the future (removing some code in debug mode)

packages/devextreme/testing/tests/DevExpress.core/utils.license.tests.js
```diff
@@ -0,0 +1,96 @@
...
+        });
+    });
+
+    test('Message should be logged only once', function(assert) {
+        verifyLicense('', '1.0');
+        assert.ok(errors.log.calledOnce);
+
+        verifyLicense('', '1.0');
+        assert.ok(errors.log.calledOnce);
+    });
```
williamvinogradov: Let's modify this test case a little:

```suggestion
    test('Message should be logged only once', function(assert) {
        verifyLicense('', '1.0');
        verifyLicense('', '1.0');
        verifyLicense('', '1.0');
        
        assert.ok(errors.log.calledOnce);
    });
```

packages/devextreme/testing/tests/DevExpress.exporter/exceljsParts/exceljs.dataGrid.tests.js
```diff
@@ -40,8 +40,10 @@ const moduleConfig = {
     beforeEach: function() {
         this.worksheet = new ExcelJS.Workbook().addWorksheet('Test sheet');
         this.customizeCellCallCount = 0;
-        this.stub = sinon.stub(errors, 'log').callsFake(() => {
-            QUnit.assert.strictEqual(true, false, 'error.log should not be called');
+        this.stub = sinon.stub(errors, 'log').callsFake((errorId) => {
+            if(errorId === 'W0003') {
+                QUnit.assert.strictEqual(true, false, 'no deprecation warnings should be logged');
+            }
```
williamvinogradov, line 46: Why did we change this test case?

packages/devextreme/testing/runner/Views/Main/RunSuite.cshtml
```diff
@@ -263,6 +263,9 @@
                     'json': '@Url.Content("~/node_modules/systemjs-plugin-json/json.js")',
                     'plugin-babel': '@Url.Content("~/node_modules/systemjs-plugin-babel/plugin-babel.js")',
                     'systemjs-babel-build': '@Url.Content("~/node_modules/systemjs-plugin-babel/systemjs-babel-browser.js")',
+
+                    'sha-1': '@Url.Content("~/node_modules/sha-1/dist/sha1.umd.js")',
+
```
williamvinogradov: Let's add a comment here to make it looks consistent:

```suggestion

                     // License check
                    'sha-1': '@Url.Content("~/node_modules/sha-1/dist/sha1.umd.js")',
                    
```

packages/devextreme/testing/tests/DevExpress.exporter/exceljsParts/exceljs.dataGrid.tests.js
iBat: https://github.com/DevExpress/DevExtreme/pull/25510#discussion_r1327056916

williamvinogradov: Sorry, I cannot agree with [this comment](https://github.com/DevExpress/DevExtreme/pull/25510#discussion_r1327056916).

Before this change, the test case checked the occurrence of any error.
But after, it will check only the 'W0003'.
This test case will pass if we have another warning with another code (or another warning).

It doesn't look okay, but maybe I missed something.
@ksercs, @pomahtri could you please provide an additional explanation for this change? 
(I mean is it okay or not from your point of view and why)

pomahtri: Agree with @williamvinogradov, narrowing test case seems to be dangerous
Mocking checking license in tests looks most suitable, but code with `//#DEBUG` is fine too in case mocking in qunit is complex

williamvinogradov: Discussed it with @ksercs  in DM.
Let's solve this problem of errors/warnings for all of our products as a separate task.
So, we decided to close this comment and leave it as is for now.

packages/devextreme/js/__internal/core/license/license_validation.ts
```diff
@@ -21,12 +24,13 @@ export type Token = {
 const SPLITTER = '.';
 const FORMAT = 1;
 
-const GENERAL_ERROR: Token = { kind: 'corrupted', error: 'general' };
-const VERIFICATION_ERROR: Token = { kind: 'corrupted', error: 'verification' };
-const DECODING_ERROR: Token = { kind: 'corrupted', error: 'decoding' };
-const DESERIALIZATION_ERROR: Token = { kind: 'corrupted', error: 'deserialization' };
-const PAYLOAD_ERROR: Token = { kind: 'corrupted', error: 'payload' };
-const VERSION_ERROR: Token = { kind: 'corrupted', error: 'version' };
+const CORRUPTED = 'corrupted';
```
williamvinogradov: Let's create a ````const object``` or ```enum``` for the ```kind``` (field of ```Token``` type)


E.g.
```typescript
// enum
enum TokenKind {
  corrupted = 'corrupted',
  verified = 'verified',
}

// or static const
const TokenKind = {
  corrupted: 'corrupted',
  verified: 'verified',
} as const;
```

packages/devextreme/js/__internal/core/license/license_validation.ts
```diff
@@ -78,3 +82,53 @@ export function parseToken(encodedToken: string | undefined): Token {
...
+      }
+    }
+  } else {
+    warning = 'W0019';
+  }
+
+  if (warning) {
+    errors.log(warning);
+  }
+};
```
williamvinogradov: Minor comment.
(You can apply it or not by your opinion)

I suggest to split this function into two.
Something like this:
```suggestion
const verifyLicense = (licenseToken: string, version: string = packageVersion): void {
  if (isLicenseVerified) {
    return;
  }
  isLicenseVerified = true;

  const [success, warning] = parseLicense(licenseToken, version);
  
  if (!success) {
    errors.log(warning);
  }
}

const parseLicense = (
  licenseToken: string,
  ver: string = versionValue,
): [success: boolean, warning: string | null] {
  if (!licenseToken) {
    return [false, 'W0019'];
  }

  const license = parseToken(licenseToken);

  if (license.kind === TokenKind.currupted) {
    return [false, 'W0021'];
  }

  /* See additional comment for this part */
  const [major, minor] = ver.split('.').map(Number);

  if (major * 10 + minor > license.payload.maxVersionAllowed) {
    return [false, 'W0020'];
  }
  /* end of this part :) */

  return [true, null];
}
```

In my opinion, it makes this code clearer and more open to modification in future

packages/devextreme/js/__internal/core/license/license_validation.ts
```diff
@@ -1,3 +1,6 @@
+import errors from '@js/core/errors';
+import { version } from '@js/core/version';
```
williamvinogradov: Minor comment.
(You can apply it or not by your opinion)

Let's change naming here
```suggestion
import { version as packageVersion } from '@js/core/version';
```

To fix rename ```ver``` -> ```version``` in the ```verifyLicense ``` function.

packages/devextreme/js/__internal/core/license/license_validation.ts
```diff
@@ -78,3 +82,53 @@ export function parseToken(encodedToken: string | undefined): Token {
     },
   };
 }
+
+export type LicenseVerifyResult = 'W0019' | 'W0020' | 'W0021' | null;
+let isLicenseVerified = false;
+
+const verifyLicense = function verifyLicense(licenseToken: string, ver: string = version): void {
```
williamvinogradov: Minor comment.
(You can apply it or not by your opinion)

Let's change ```ver``` -> ```version```

packages/devextreme/js/__internal/core/license/license_validation.ts
```diff
@@ -78,3 +82,53 @@ export function parseToken(encodedToken: string | undefined): Token {
...
+    const license = parseToken(licenseToken);
+
+    if (license.kind === CORRUPTED) {
+      warning = 'W0021';
+    } else {
+      const [major, minor] = ver.split('.').map(Number);
+
+      if (major * 10 + minor > license.payload.maxVersionAllowed) {
+        warning = 'W0020';
+      }
```
williamvinogradov: If we will have ```ver = '123'``` we verify the license without any error

Because the ```123 * 10 + NaN = NaN```
And ```NaN > 10000000``` is false

Therefore, we should add an additional condition here.
Maybe we need the special error code for it :)

packages/devextreme/testing/tests/DevExpress.core/utils.license.tests.js
```diff
@@ -0,0 +1,97 @@
+import config from 'core/config';
```
williamvinogradov: Let's move this test case to ```jest```

It's not cool to have one part of tests in ```jest``` and another in ```QUnit```
(Discussed with @ksercs )

packages/devextreme/testing/tests/DevExpress.core/component.tests.js
```diff
@@ -1707,3 +1708,21 @@ QUnit.module('action API', {}, () => {
         config({ wrapActionsBeforeExecute: originFlag });
     });
 });
+
+QUnit.module('License check', {
+    beforeEach: function() {
+        sinon.spy(license, 'verifyLicense');
+        // sinon.spy(verifyLicense);
```
williamvinogradov: Do we still need this comment?

packages/devextreme/testing/tests/DevExpress.core/component.tests.js
```diff
@@ -1707,3 +1708,21 @@ QUnit.module('action API', {}, () => {
...
+        license.setLicenseCheckSkipCondition(false);
+    },
+    afterEach: function() {
+        license.verifyLicense.restore();
+        // verifyLicense.restore();
```
williamvinogradov: Do we still need this comment?

packages/devextreme/testing/tests/DevExpress.core/utils.license.tests.js
ksercs: done

packages/devextreme/js/core/utils/license.js
williamvinogradov: We decided leave it as is for now

packages/devextreme/testing/tests/DevExpress.core/component.tests.js
williamvinogradov: Fixed

packages/devextreme/testing/tests/DevExpress.core/component.tests.js
williamvinogradov: Fixed

packages/devextreme/js/__internal/core/license/license_validation.ts
iBat: I reused code W0021 because otherwise I would have to agree with Ray about 1 more error text.

williamvinogradov (dismissed):

packages/devextreme/js/core/component.js
```diff
@@ -12,6 +12,7 @@ import { PostponedOperations } from './postponed_operations';
 import { isFunction, isPlainObject, isDefined } from './utils/type';
 import { noop } from './utils/common';
 import { getPathParts } from './utils/data';
+import license from '../__internal/core/license/license_validation';
```
IlyaKhD, line 15: Can we use named import?
```suggestion
import { verifyLicense } from '../__internal/core/license/license_validation';
```

packages/devextreme/js/core/errors.js
```diff
@@ -249,5 +249,27 @@ export default errorUtils({
...
+    */
+    W0019: 'DevExtreme: Valid license key not found.\n\n' +
+        'If you are using a trial version, you must uninstall all copies of DevExtreme once your 30 days trial period expires. For licensing-related information, please refer to the DevExtreme End User License Agreement: https://js.devexpress.com/EULAs/DevExtremeComplete/.\n\n' +
+        'To continue using DevExtreme in a project, you must purchase a license. For pricing/licensing options, please visit: https://js.devexpress.com/Buy/.\n\n' +
+        'If you have licensing-related questions or need help with a purchase, please email clientservices@devexpress.com. We will be happy to follow-up.\n\n',
+    /**
+     * @name ErrorsCore.W0020
+     */
+    W0020: 'DevExtreme: The license key is expired.\n\n' +
+        'A mismatch exists between license key/DevExtreme version.\n\n' +
```
IlyaKhD:
```suggestion
        'A mismatch exists between license key and DevExtreme version.\n\n' +
```

packages/devextreme/js/core/errors.js
```diff
@@ -249,5 +249,27 @@ export default errorUtils({
...
+        'To proceed, you can:\n' +
+        ' ● use a version of DevExtreme linked to your license key https://www.devexpress.com/ClientCenter/DownloadManager/\n' +
```
IlyaKhD: I would better use .25 instead of these .50 🙂
```suggestion
        ' • use a version of DevExtreme linked to your license key: https://www.devexpress.com/ClientCenter/DownloadManager/\n' +
```

packages/devextreme/js/core/errors.js
```diff
@@ -249,5 +249,27 @@ export default errorUtils({
...
+        ' ● renew your DevExpress Subscription (once you renew your subscription, you will be entitled to product updates and support service https://www.devexpress.com/buy/renew/)\n\n' +
```
IlyaKhD:
```suggestion
        ' • renew your DevExpress Subscription: https://www.devexpress.com/buy/renew/ (once you renew your subscription, you will be entitled to product updates and support service)\n\n' +
```

packages/devextreme/js/core/component.js
```diff
@@ -82,6 +83,8 @@ export const Component = Class.inherit({
         this._disposingCallbacks = _disposingCallbacks || Callbacks();
         this.postponedOperations = new PostponedOperations();
         this._createOptions(options);
+
+        license.verifyLicense(Config().license);
```
IlyaKhD: There is no test for obtaining license token from config. We discussed this with @ksercs and I agree with his proposal: we can try moving `Config().license` inside `verifyLicense` and mocking it `Config().License`

Is this possible?

packages/devextreme/js/__internal/core/license/license_validation.ts
```diff
@@ -70,11 +81,54 @@ export function parseToken(encodedToken: string | undefined): Token {
...
+    kind: TokenKind.verified,
     payload: {
       customerId,
       maxVersionAllowed,
       ...rest,
     },
   };
 }
+
+export function verifyLicense(licenseToken: string, version: string = packageVersion): void {
```
IlyaKhD: I think better wrap the whole function body with `try-catch` to ensure all errors are handled. `W0021` should be thrown from the `catch`

packages/devextreme/js/__internal/core/license/license_validation.ts
```diff
@@ -10,23 +13,31 @@ interface Payload extends Partial<License> {
   readonly format?: number;
 }
 
+const enum TokenKind {
```
IlyaKhD, line 16: **Minor**
I wouldn't use `enum`, strings literals are great enough. As for me `enum`'s don't breathe the spirit of the modern TypeScript, so to say.

packages/devextreme/js/core/component.js
iBat: We can. But in that case we cant use sinon.spy/stub in tests.

packages/devextreme/js/__internal/core/license/license_validation.ts
iBat: const enum !== enum. It compiles into string literals instead of function wrappers as simple enums.

packages/devextreme/js/core/component.js
iBat: Unclear how to mock default exported function in tests.

packages/devextreme/js/__internal/core/license/license_validation.ts
IlyaKhD: Ah, didn't notice that. Ok then

IlyaKhD: Though I am still not sure if it's better than a literal

packages/devextreme/js/core/component.js
IlyaKhD: Maybe mocking in this case is not good

IlyaKhD (approved):

VasilyStrelyaev (dismissed):

Lykoi18 (dismissed):

vladaskorohodova (dismissed):

packages/devextreme/js/core/component.js
IlyaKhD:
```suggestion
        license.verifyLicense(Config().licenseKey);
```

IlyaKhD (changes requested):

IlyaKhD (approved):

Lykoi18 (approved):

PR #25543, opened 2023-09-07
DataGrid - Master-detail - It's not possible to focus all rows after rows are expanded and collapsed (T1187124)

pomahtri:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
@@ -1742,7 +1742,7 @@ export class KeyboardNavigationController extends modules.ViewController {
 
     const lastVisibleIndex = Math.max(
       ...dataController.items()
-        .map((item, index) => (item.visible ? index : -1)),
+        .map((item, index) => (item.visible !== false ? index : -1)),
     );
 
     return rowIndex === lastVisibleIndex;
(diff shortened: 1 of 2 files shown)


Raushen (dismissed):

Raushen (approved):

andrewmakarov (approved):

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1733,17 +1733,19 @@ export class KeyboardNavigationController extends modules.ViewController {
     return this._isCellValid($cell);
   }
 
-  _isLastRow(rowIndex) {
-    const dataController = this._dataController as any;
-    const visibleItems = dataController
-      .items()
-      .filter((item) => item.visible !== false);
+  private _isLastRow(rowIndex: number): boolean {
```
williamvinogradov, line 1736: Minor comment.

Let's get rid of the local ```dataController``` variable
Something like this:
```typescript
    if (this._isVirtualRowRender()) {
      return rowIndex >= (this._dataController as any).getMaxRowIndex();
    }
    
    const itemVisibleIndexes = this._dataController.items()
      .map((item, index) => (item.visible !== false ? index : -1));
    const lastVisibleIndex = Math.max(itemVisibleIndexes);
    
    return rowIndex === lastVisibleIndex;
```

williamvinogradov (approved):

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
pomahtri: Don't want to lose 3 approves + it's not my change, it was so before. I'll keep it in mind for next PRs

PR #25588, opened 2023-09-14
RadioGroup: validation message should be pronounced on value change

Zedwag:


ksercs (approved):
We created a separate task to research how we can extend this solution for all editors and keep the actual functionality working well.

PR #25592, opened 2023-09-14
Fluent: Remove Roboto font imports

marker-dao:


packages/devextreme/build/gulp/npm.js
```diff
@@ -141,6 +141,10 @@ gulp.task('npm-sass', gulp.series(
             .src('fonts/**/*', { base: '.' })
             .pipe(gulp.dest(`${scssDir}/widgets/material/typography`)),
 
+        () => gulp
```
EugeniyKiyashko: We are not sure that we have a need for these files for the new font-family that will be used for the fluent theme.

```
 font-family: segoe ui, -apple-system, BlinkMacSystemFont, avenir next, avenir, helvetica neue, helvetica, Cantarell, Ubuntu, roboto, noto, arial, sans-serif;
```

Now we are enclosing these files to avoid errors for use in the UI template gallery. After research, if we understand that these files are redundant or some other .ttf, .woff files are required, we will fix it.

ksercs (approved):

packages/devextreme/build/gulp/npm.js
marker-dao: In the end, we decided not to copy the font files, but to delete the import of these files, as we felt the Roboto font was a low priority.

EugeniyKiyashko (approved):

PR #25606, opened 2023-09-18
Scheduler - The `E1035 - The editor cannot be created` error appears on an attempt to edit a recurring appointment (T1188098)

pomahtri:


markallenramirez (approved):

packages/devextreme/testing/testcafe/tests/scheduler/appointmentForm/showAppointmentPopup.ts
```diff
@@ -29,6 +29,30 @@ test('Invoke showAppointmentPopup method shouldn\'t raise error if value of curr
   height: 600,
 }));
 
+test('Appointment popup shouldn\'t raise error if appoitment is recursive', async (t) => {
+  const scheduler = new Scheduler('#container');
+  await t.doubleClick(scheduler.getAppointment('Meeting of Instructors').element);
+  await t.click(Scheduler.getEditRecurrenceDialog().series);
+
+  const consoleMessages = await t.getBrowserConsoleMessages();
+  await t.expect(consoleMessages.error.length).eql(0);
```
andrewmakarov, line 38: It seems to me that the testcafe will indicate anyway if there is error in the console. Or is it impossible for the test to be without an assert?

packages/devextreme/testing/testcafe/tests/scheduler/appointmentForm/showAppointmentPopup.ts
```diff
@@ -29,6 +29,30 @@ test('Invoke showAppointmentPopup method shouldn\'t raise error if value of curr
   height: 600,
 }));
 
+test('Appointment popup shouldn\'t raise error if appoitment is recursive', async (t) => {
```
andrewmakarov, line 32: Maybe the name of test shouldn't focus on the error? Test that the form works in case that the time zone exisit and recurrence

andrewmakarov (approved):

packages/devextreme/testing/testcafe/tests/scheduler/appointmentForm/showAppointmentPopup.ts
pomahtri: Unfortunately, if I remove this `expect`, the test will pass. Seems that this error is raised not with `throw ...`, but with `console.error`, which isn't caught by testcafe

packages/devextreme/testing/testcafe/tests/scheduler/appointmentForm/showAppointmentPopup.ts
pomahtri: Test is specially to check that there are no errors, and we need to check that there's no error in the console, so I think it's okay to have this name

PR #25608, opened 2023-09-18
Charts - Publish SeriesPoint & SeriesLabel types

alexanderPolosatov:


marker-dao (dismissed):

packages/devextreme/js/viz/chart.d.ts
```diff
@@ -3788,363 +3790,7 @@ export interface dxChartSeriesTypesCommonSeriesHoverStyleHatching {
      */
     width?: number;
 }
-/** @namespace DevExpress.viz */
-export interface dxChartSeriesTypesCommonSeriesLabel {
```
IlyaKhD, line 3792: Add aliases for backward compatibility:

```ts
/**
 * @namespace DevExpress.viz
 * @deprecated use SeriesLabel from common/charts instead
 */
export type dxChartSeriesTypesCommonSeriesLabel = SeriesLabel;
```
And  make sure it remains in dx.all.d.ts in the right place. If not - use `/ts/aliases.d.ts` (next to dx.all.d.ts)

IlyaKhD (approved):

PR #25615, opened 2023-09-19
DataGrid - "W1024 - The client-side export is enabled" warning occurs when the exporting feature is used in Angular (T1188932)

pomahtri:


packages/devextreme/js/__internal/grids/data_grid/export/m_export.ts
```diff
@@ -608,11 +608,18 @@ export class ExportController extends dataGridCore.ViewController {
     return headersView && headersView.isVisible() ? headersView.getColumnWidths() : rowsView.getColumnWidths();
   }
 
-  init() {
-    if (this.option('export.enabled') && !isDefined(this.option('onExporting'))) {
+  private throwWarningIfNoOnExportingEvent(): void {
+    const hasProp = isDefined(this.option('onExporting'));
+    const hasEvent = (this.component as any)._eventsStrategy?.hasEvent('exporting');
```
Alyar666: I think it's better to use the [hasActionSubscription](https://github.com/DevExpress/DevExtreme/blob/23_2/packages/devextreme/js/core/component.js#L353) method. It is available at the component level

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/data_grid/export/m_export.ts
```diff
@@ -666,6 +673,13 @@ export class ExportController extends dataGridCore.ViewController {
     }
   }
 
+  optionChanged(args) {
+    super.optionChanged(args);
+    if (args.fullName === 'export.enabled') {
```
Alyar666: If I change the option like this:
`dataGridInstance.option('export', { enabled: true })`

Will the error be thrown?

Alyar666 (dismissed):

Alyar666 (approved):

PR #25621, opened 2023-09-19
DataGrid: Fix onContentReady was not called when the grid is rendered in DropDownBox (T1188486)

Alyar666:
See https://devexpress.com/issue=T1188486

### DIFF
--- a/js/__internal/grids/grid_core/views/m_grid_view.ts
+++ b/js/__internal/grids/grid_core/views/m_grid_view.ts
@@ -57,10 +57,21 @@ const resizingControllerMembers = {
 
     if (!this._refreshSizesHandler) {
       this._refreshSizesHandler = (e) => {
+        let resizeDeferred;
+        const changeType = e?.changeType;
+        const isDelayed = e?.isDelayed;
+
         dataController.changed.remove(this._refreshSizesHandler);
 
         if (this._checkSize()) {
-          this._refreshSizes(e);
+          resizeDeferred = this._refreshSizes(e);
+        }
+
+        if (changeType && changeType !== 'updateSelection' && changeType !== 'updateFocusedRow' && changeType !== 'pageIndex' && !isDelayed) {
+          when(resizeDeferred).done(() => {
+            this._setAriaLabel();
+            this.fireContentReadyAction();
+          });
         }
       };
       // TODO remove resubscribing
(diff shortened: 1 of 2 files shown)


js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -57,10 +57,21 @@ const resizingControllerMembers = {
 
     if (!this._refreshSizesHandler) {
       this._refreshSizesHandler = (e) => {
+        let resizeDeferred;
```
williamvinogradov: Let's modify this block of the code a little:
* add type to ```resizeDeferred```
* put condition (70 line) out of brackets in a separate variable
* pass to the ```when``` function resolved ```Deferred``` instead of the ```undefined```
```when``` function can handle the ```undefined``` value, but let's make it more obvious for the future refactoring :)

E.g. something like this:
```typescript
let resizeDeferred: DeferredObj<unknown> | null = null;
const changeType = e?.changeType;
const isDelayed = e?.isDelayed;
// I'm out of the context to come up with this variable name :)
// I guess that you can come up with a better name here
const suitableChangeType = changeType
  && changeType !== 'updateSelection'
  && changeType !== 'updateFocusedRow'
  && changeType !== 'pageIndex'
  && !isDelayed;

dataController.changed.remove(this._refreshSizesHandler);

if (this._checkSize()) {
  resizeDeferred = this._refreshSizes(e);
}

if (suitableChangeType) {
  resizeDeferred ??= new DeferredObj<null>().resolve(null);
  when(resizeDeferred).done(() => {
    this._setAriaLabel();
    this.fireContentReadyAction();
  });
}
```

js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -71,19 +82,19 @@ const resizingControllerMembers = {
   },
 
   _refreshSizes(e) {
-    let resizeDeferred;
-    const that = this;
-    const changeType = e && e.changeType;
-    const isDelayed = e && e.isDelayed;
-    const items = that._dataController.items();
+    // @ts-expect-error
+    let resizeDeferred = new Deferred().resolve();
```
williamvinogradov: Let's add types and resolve value here:

```suggestion
    let resizeDeferred = new Deferred<null>().resolve(null);
```

js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -57,10 +57,21 @@ const resizingControllerMembers = {
 
     if (!this._refreshSizesHandler) {
       this._refreshSizesHandler = (e) => {
+        let resizeDeferred;
+        const changeType = e?.changeType;
+        const isDelayed = e?.isDelayed;
+
         dataController.changed.remove(this._refreshSizesHandler);
```
williamvinogradov: Question.

Can we replace the local ```dataController``` variable with ```this.dataController``` here?

williamvinogradov (approved):

PR #25636, opened 2023-09-20
Tabs: Update scrollable after change the orientation option in runtime

marker-dao:


ksercs (dismissed):

packages/devextreme/js/ui/tabs.js
```diff
@@ -397,16 +397,12 @@ const Tabs = CollectionWidget.inherit({
...
-
-            scrollable.option('direction', scrollableDirection);
-        } else {
-            this._renderScrolling();
+    _updateScrollable() {
+        if(this.getScrollable()) {
+            this._cleanScrolling();
         }
+
+        this._renderScrolling();
```
ksercs, line 405: The solution seems to be very wide, cannot we just call `_updateNavButtonsVisibility` or localize it other way?

marker-dao: This is the cleanest solution.

The _updateNavButtonsVisibility method is responsible for setting the disabled state of the buttons. This method is not suitable for the current purpose. Moreover, I suggest renaming it.

In the following variant, I rewrote the _updateNavButtonsVisibility method, in which I used the showNavButtons option to visually remove the buttons. But I was very confused by the fact that when the showNavButtons option is changed at runtime, our code calls _invalidate. This redraws the whole component. In my opinion, this is the most unfortunate solution.

I also had an idea to add code that would expose a specific CSS class that would allow hiding buttons. But in my opinion, these options are not the best solution. So my suggestion is to leave it as it is.

ksercs: I believe we can use `showNavButton` method (possibly with some modifications). In common, it's not a good practice to rerender so wide parts. But now I see that it's already written in dimension changed handlers, so your new code definitely does not the common performance worse.

So let's postpone it for now.

ksercs (approved):

EugeniyKiyashko (approved):

PR #25640, opened 2023-09-20
DataGrid: Fix disabled state for export button when using the `toolbar|items|disabled` option (T1189621)

Alyar666:
See https://devexpress.com/issue=T1189621

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
@@ -126,6 +126,13 @@ const members = {
     }
   },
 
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
```diff
@@ -126,6 +126,13 @@ const members = {
     }
   },
 
+  _isDisabledDefinedByUser(name: string): boolean {
```
williamvinogradov, line 129: Let's add the access modifier private to this method.
It looks like it shouldn't be used outside of the m_header_panel :)

```suggestion
  private _isDisabledDefinedByUser(name: string): boolean {
```

williamvinogradov (approved):

PR #25650, opened 2023-09-23
Fluent: testing Form with new theme

EugeniyKiyashko:


EugeniyKiyashko:
We are going to add this jobs in https://github.com/DevExpress/DevExtreme/pull/25673/

PR #25675, opened 2023-09-26
Swap the buttons done and cancel in modals(grid, pivotGrid, scheduler)

andrewmakarov:


packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_form_based.ts
```diff
@@ -150,7 +151,7 @@ const editingControllerExtender = (Base: ModuleType<EditingController>) => class
     }
   }
 
-  _showEditPopup(rowIndex, repaintForm?) {
+  _showEditPopup(rowIndex, repaintForm?): void {
```
williamvinogradov, line 154: Minor comment.

Let's try to add some types for the method's arguments
At first look it may be:
```suggestion
  _showEditPopup(rowIndex: number, repaintForm?: boolean): void {
```

packages/devextreme/js/__internal/grids/grid_core/m_utils.ts
```diff
@@ -173,6 +173,15 @@ function normalizeGroupingLoadOptions(group) {
...
 }
 
+export const popupUtils = {
+  createDoneButton: (options) => ({
+    toolbar: 'bottom', location: 'after', shortcut: 'done', options,
+  }),
+  createCancelButton: (options) => ({
+    toolbar: 'bottom', location: 'after', shortcut: 'cancel', options,
+  }),
+};
```
williamvinogradov, line 183: Let's rename these methods
Because the ```create_done/end_button``` method doesn't create any button at all :)\
And let's add type to the ```options``` argument

I suggest the next namings:
```suggestion
export const popupUtils = {
  getDoneButtonConfig: (options: IToolbarButtonOptions) => ({
    toolbar: 'bottom', location: 'after', shortcut: 'done', options,
  }),
  getCancelButtonConfig: (options: IToolbarButtonOptions) => ({
    toolbar: 'bottom', location: 'after', shortcut: 'cancel', options,
  }),
};
```

packages/devextreme/js/__internal/grids/grid_core/__test__/utils.test.ts
```diff
@@ -0,0 +1,15 @@
+// eslint-disable-next-line forbidden-imports/no-restricted-imports
+import { popupUtils } from '../m_utils';
+
+describe('utils', () => {
```
williamvinogradov, line 4: Let's refactor these test cases a little and add the next scenarios:

- let's check all properties, not only the ```shortcut``` (I mean the ```toolbar``` and ```location``` too)
May be it will be good idea to match the ```result``` with the ```expectedResult``` object
- let's check that the ```option``` argument passed to these functions will exist in the ```result```

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_form_based.ts
```diff
@@ -432,8 +430,8 @@ const editingControllerExtender = (Base: ModuleType<EditingController>) => class
 
       if (!isPopupForm) {
         const $buttonsContainer = $('<div>').addClass(this.addWidgetPrefix(FORM_BUTTONS_CONTAINER_CLASS)).appendTo($container);
-        this._createComponent($('<div>').appendTo($buttonsContainer), Button, this._getSaveButtonConfig());
         this._createComponent($('<div>').appendTo($buttonsContainer), Button, this._getCancelButtonConfig());
+        this._createComponent($('<div>').appendTo($buttonsContainer), Button, this._getSaveButtonConfig());
```
williamvinogradov, line 434: Question.

Why do we change places of the cancel and save button initialization?
Does the order matter?

packages/devextreme/js/renovation/ui/scheduler/appointment_edit_form/__tests__/popup_config.test.ts
```diff
@@ -178,4 +180,16 @@ describe('API', () => {
       });
     });
   });
+
+  describe('popup dialogs', () => {
```
williamvinogradov, line 184: Let's add the next test cases for these functions:
- let's check that the passed ```location``` argument exists in the ```result```
- let's check that the passed ```options``` argument exists in the ```result```

packages/devextreme/js/__internal/grids/grid_core/m_utils.ts
andrewmakarov: I agree with renaming from button to option. But I don't agree that there is no creation. The object is being created

PR #25688, opened 2023-09-27
Text Editors: clear valid mark after resetting value to initial

AlexanderMoiseev:


packages/devextreme/js/ui/text_box/ui.text_editor.base.js
```diff
@@ -924,6 +924,10 @@ const TextEditorBase = Editor.inherit({
         } else {
             this.callBase();
         }
+
+        this.option('validationStatus', 'valid');
+        this._showValidMark = false;
+        this._renderValidationState();
```
ksercs: do we really need to rerender all the validation state? Cannot we just move the code related to validMark to a separate method and call it here?

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/validator.editors.tests.js
```diff
@@ -312,6 +312,20 @@ QUnit.module('Editors Standard Adapter', {
         });
     });
 
+    QUnit.test('Reset should clear valid mark', function(assert) {
+        const editor = this.fixture.createTextEditor();
+
+        editor.option('validationStatus', 'pending');
+        editor.option('showValidationMark', true);
```
ksercs: As far as i remember it's true by default, do we really this code row?

packages/devextreme/js/ui/text_box/ui.text_editor.base.js
AlexanderMoiseev: sure

alexanderPolosatov (approved):

ksercs (approved):

PR #25715, opened 2023-10-02
Fluent: htmlEditor redesign

EugeniyKiyashko:


packages/devextreme/scss/widgets/base/_form.scss
```diff
@@ -144,7 +136,7 @@
 
   .dx-form-group {
     &.dx-form-group-with-caption {
-      padding-left: 20px;
+      padding-inline-start: 20px;
```
marker-dao, line 139: Are you sure you shouldn't use variables here?

iBat (approved):

packages/devextreme/scss/widgets/base/_form.scss
EugeniyKiyashko: There are a lot of constant values uses in form styles. I suggest to extract these values to variables in a separate task.

Shuvaevmedia (approved):

PR #25745, opened 2023-10-04
Charts - Move Font type to common/charts

alexanderPolosatov:


packages/devextreme/js/common/charts.d.ts
```diff
@@ -341,6 +338,43 @@ export type GradientColor = {
     color: string;
 };
 
+/**
+ * @docid
+ * @namespace DevExpress.common.charts
+ * @public
+ */
+export type Font = {
```
mpreyskurantov, line 346: "devextreme-internal-tools" fix is required

mpreyskurantov (changes requested):

packages/devextreme/js/viz/core/base_widget.d.ts
```diff
@@ -915,3 +883,13 @@ export interface BaseWidgetAnnotationConfig {
      */
     y?: number;
 }
+
+// #region deprecated in 22.2
```
mpreyskurantov, line 887: 22.2 or 23.2?

packages/devextreme/js/viz/core/base_widget.d.ts
```diff
@@ -915,3 +883,13 @@ export interface BaseWidgetAnnotationConfig {
...
+
+/**
+ * @namespace DevExpress.viz
+ * @deprecated Use Font from DevExpress.common.charts.Font instead
```
IlyaKhD:
```suggestion
 * @deprecated Use Font from common/charts instead

```

packages/devextreme/js/viz/core/base_widget.d.ts
```diff
@@ -32,6 +32,7 @@ import {
     TextOverflow,
     Theme,
     WordWrap,
+    Font as CommonFont,
```
IlyaKhD, line 35: We can avoid using renaming here:
```suggestion
    Font,
```
I've tried it locally - seems to work fine:


```ts
/**
 * @namespace DevExpress.viz
 * @deprecated Use Font from DevExpress.common.charts.Font instead
 */
export type Font = Font;
```

IlyaKhD: And you'll have to add a line to the aliases.d.ts file with different deprecation message

IlyaKhD: Nope, this doesn't work. Behaves as expected, there is no magic in this world(

packages/devextreme/js/common/charts.d.ts
alexanderPolosatov: fixed

mpreyskurantov (approved):

packages/devextreme/js/viz/core/base_widget.d.ts
alexanderPolosatov: fixed

marker-dao (approved):

PR #25781, opened 2023-10-10
React Wrappers: New Template Engine

VasilyStrelyaev:


packages/devextreme-react/jest.config.js
```diff
@@ -7,4 +7,10 @@ module.exports = {
   ...base,
   name: packageName,
   displayName: packageName,
+  testPathIgnorePatterns: [
+    "component.test.tsx",
+    "nested-option.test.tsx",
+    "props-updating.test.tsx",
+    "template.test.tsx",
+  ],
```
IlyaKhD, line 15: As for me, it's better to explicitly skip tests:

```ts
    describe.skip(...
```

packages/devextreme-react/src/core/__tests__/component.test.tsx
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
IlyaKhD: Could we avoid `@ts-nocheck`?
Generally it's a bad idea to use it

packages/devextreme-react/src/core/template-wrapper.tsx
```diff
@@ -0,0 +1,130 @@
...
+          { templateFactory({ data, index, onRendered }) }
+          { hiddenNode }
+          { removalListener }
+        </RemovalLockerContext.Provider>
+      </>,
+      container,
+  );
+};
+
+export default memo(TemplateWrapper);
```
IlyaKhD: Can we avoid using default export? named are better.

packages/devextreme-react/src/core/template-wrapper.tsx
```diff
@@ -0,0 +1,130 @@
...
+  index,
+  container,
+  onRemoved,
+  onRendered,
+}) => {
+  const [removalListenerRequired, setRemovalListenerRequired] = useState(false);
+  const isRemovalLocked = useRef(false);
+  const removalLocker = useMemo(() => ({
+    lock: (): void => { isRemovalLocked.current = true; },
+    unlock: (): void => { isRemovalLocked.current = false; },
```
IlyaKhD: **Minor**
You can use simpler syntax here:

```suggestion
    lock(): void { isRemovalLocked.current = true; },
    unlock(): void { isRemovalLocked.current = false; },
```

packages/devextreme-react/src/core/template-manager.tsx
```diff
@@ -0,0 +1,180 @@
...
+    if (model && Object.prototype.hasOwnProperty.call(model, 'key')) {
+      model.dxkey = model.key;
+    }
+    return model;
+  }
+  return props;
+}
+
+export const TemplateManager: FC<TemplateManagerProps> = ({ init }) => {
+  const [renderedInstances, setRenderedInstances] = useState(new RenderedTemplateInstances());
```
IlyaKhD: They are not rendered, since they are not connected to DOM/VDOM. These are just data-transfer objects (`TemplateInstanceDefinition`).

```suggestion
  const [instantiationModels, setInstantiationModels] = useState(new RenderedTemplateInstances());
```

packages/devextreme-react/src/core/types.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export type TemplateFunc = (arg: TemplateArgs) => JSX.Element;
+
+export type DXTemplateCreator = (templateOptions: Record<string, ITemplate>) => DXTemplateCollection;
+
+export interface TemplateManagerProps {
+  init: (getDXTemplates: DXTemplateCreator, clearRenderedInstances: () => void) => void;
+}
+
+export interface TemplateInstanceDefinition {
```
IlyaKhD: **Minor**

```suggestion
export interface TemplateInstantiationModel {
```

IlyaKhD (dismissed):

packages/devextreme-react/jest.config.js
VasilyStrelyaev: This is to skip currently non-compilable tests for which I would prefer to spare their git history.
`// @ts-nocheck` to silence the linter, skipping in `jest.config.js` instead of `describe.skip()` to avoid compilation attempts.

packages/devextreme-react/src/core/__tests__/component.test.tsx
VasilyStrelyaev: This is to skip currently non-compilable tests for which I would prefer to spare their git history.
`// @ts-nocheck` to silence the linter, skipping in `jest.config.js` instead of `describe.skip()` to avoid compilation attempts.

packages/devextreme-react/src/core/component-base.ts
```diff
@@ -35,6 +35,10 @@ abstract class ComponentBase<P extends IHtmlOptions> extends React.PureComponent
       : { display: 'contents' };
   }
 
+  static contextType?: React.Context<UpdateLocker> | undefined = RemovalLockerContext;
```
ivanblinov2k17: Do we need undefined here?

packages/devextreme-react/tsconfig.json
```diff
@@ -3,7 +3,8 @@
 
   "compilerOptions": {
     "baseUrl": "./",
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "downlevelIteration": true
```
ivanblinov2k17, line 7: For what reason do we need this option?

VasilyStrelyaev: We need this option to be able to iterate maps ([here](https://github.com/DevExpress/DevExtreme/pull/25781/files#diff-8d058ac323cdd500661600d2ce8ec227cc7ede855a5246274af71525ec625dadR56)) while targeting an old ES.

ivanblinov2k17 (approved):

IlyaKhD (approved):

PR #25800, opened 2023-10-12
MultiView: Set aria-attributes

marker-dao:


alexanderPolosatov (dismissed):

Zedwag (dismissed):

packages/devextreme/js/ui/multi_view.js
```diff
@@ -213,6 +215,45 @@ const MultiView = CollectionWidget.inherit({
...
+                'dxMultiView-itemAriaLabel',
+                itemIndex + 1,
+                itemsCount,
+            ),
+        };
+
+        return aria;
+    },
+
+    _setItemsAria() {
```
ksercs: [scenarios, perf]
This method is called only on a component's initialization now. Are u sure it's enough for all cases related to a runtime change? I see the base test but not sure if it work in other cases (for example, when new items count is bigger than an old one).
Check plz.
I suggest u move this setting of an aria attribute to the `_updateItemsVisibility` method because:
1. We will be sure that all scenarios are covered
2. It's better for performance

How're u about it?

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
```diff
@@ -102,6 +101,14 @@ const animationCapturing = {
     }
 };
 
+const multiViewItemAttributeMap = (index) => {
```
ksercs: [naming]
```suggestionf
const getMultiViewItemAttributeMap = (index) => {
```

ksercs: [code splitting]
let's move it to a module?

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
```diff
@@ -1160,6 +1167,52 @@ QUnit.module('aria accessibility', {
...
+            'aria-roledescription': 'MultiView',
+            'aria-label': 'Use the arrow keys or swipe to navigate between views',
+        };
+
+        Object.keys(attributeMap).forEach(key => {
+            assert.strictEqual($element.attr(key), attributeMap[key], `${key} attribute is correct`);
+        });
+    });
+
+    function checkAttributes(assert, items) {
```
ksercs: it's more intuitive if functions are declared in the beginning of a scope

packages/devextreme/js/ui/multi_view.js
marker-dao: These methods have nothing to do with visibility.

_updateItemsVisibility is called too often, when selectedIndex or any of item option were changed. 

But setting the another aria attributes (without aria-hidden) should take place only when items were changed. And selected index shouldn't change the aria attributes. In all other cases the _initMarkup method will be called.

marker-dao: [e88d35eb9fa4d0fcb171faaf710e3d6ef8dba941]

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
marker-dao: [e88d35eb9fa4d0fcb171faaf710e3d6ef8dba941]

marker-dao: [e88d35eb9fa4d0fcb171faaf710e3d6ef8dba941]

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
marker-dao: [e88d35eb9fa4d0fcb171faaf710e3d6ef8dba941]

ksercs (approved):

alexanderPolosatov (approved):

titerman (approved):

PR #25835, opened 2023-10-19
Fix compilation targets

IlyaKhD:

### DIFF
--- a/packages/devextreme/build/gulp/transpile-config.js
+++ b/packages/devextreme/build/gulp/transpile-config.js
@@ -11,9 +11,15 @@ const common = {
     ignore: ['**/*.json'],
 };
 
+const targets = {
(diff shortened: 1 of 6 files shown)


williamvinogradov (dismissed):

pomahtri (commented):
`@babel/preset-env` should be aware of browserlist in our package.json: https://babeljs.io/docs/babel-preset-env#how-does-it-work

Why we need to set additional targets in config?

pomahtri (approved):

IlyaKhD:
@pomahtri, babel has issues with joining `esmodules: true` with browserslist
Current solution breaks "single source of truth" principle, but I'm preparing next PR to fix it

VasilyStrelyaev (approved):

PR #25876, opened 2023-10-24
TreeList and DataGrid always show the W1025 warning regardless of the height option value or CSS limiting the component height  (T1196939)

pomahtri:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
@@ -916,7 +916,7 @@ class RowsView extends ColumnsView {
     that._scrollTop = 0;
     that._scrollLeft = -1;
     that._scrollRight = 0;
-    that._hasHeight = false;
+    that._hasHeight = undefined;
     that._contentChanges = [];
     dataController.loadingChanged.add((isLoading, messageText) => {
       that.setLoading(isLoading, messageText);
--- a/packages/devextreme/testing/testcafe/tests/dataGrid/scrolling.ts
+++ b/packages/devextreme/testing/testcafe/tests/dataGrid/scrolling.ts
@@ -1647,7 +1647,6 @@ test('Warning should not be thrown if scrolling is virtual and height is specifi
 
 ['height', 'max-height'].forEach((cssOption) => {
   test(`Warning should not be thrown if scrolling is virtual and height is specified with css (${cssOption})`, async (t) => {
-    await t.debug();
     const consoleMessages = await t.getBrowserConsoleMessages();
     const warningExists = !!consoleMessages?.warn.find((message) => message.startsWith('W1025'));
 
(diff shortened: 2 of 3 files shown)


packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
```diff
@@ -916,7 +916,7 @@ class RowsView extends ColumnsView {
     that._scrollTop = 0;
     that._scrollLeft = -1;
     that._scrollRight = 0;
-    that._hasHeight = false;
+    that._hasHeight = undefined;
```
williamvinogradov, line 919: This third value of the flag variable looks like a crutch.
Because as far as I can see we should skip the first flag check.

Maybe we should move the check of this flag and warning log to another place in the ```grid_core``` / ```virtual_scrolling```?

packages/devextreme/testing/testcafe/tests/dataGrid/scrolling.ts
```diff
@@ -1624,4 +1625,47 @@ test('Warning should be thrown if scrolling is virtual and height is not specifi
...
+  scrolling: {
+    mode: 'virtual',
+  },
+  dataSource: [
+    { column: 'value' },
+  ],
+  height: 200,
 }));
+
+test('Warning should not be thrown if scrolling is virtual and height is specified with css', async (t) => {
```
williamvinogradov: Let's also add case with the ```max-height``` CSS rule just to be sure that it works :)

williamvinogradov:
I found issues during the manual testing of these changes.
If we dynamically remove the height from the DataGrid, the warning will not be thrown.
E.g set the ```height``` option to null or remove the ```height```/```msx-height``` CSS rules from the DataGrid container

packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
williamvinogradov: Discussed, decided to leave this as is for now

williamvinogradov:
Some comments about the ```this._heightWarningIsThrown``` flag:

1) Let's rename this flag from ```this._heightWarningIsThrown``` -> ```this._thrownWarnings.virtualScrollingWithoutHeight```
The main idea is to have a single object for all possible warning flags.
If we add some warnings in the future, it will not be cool to create a separate flag for each case.
It's better to store all similar things in one place.

2) Let's initialize the flag with an initial value explicitly [here (before the init method)](https://github.com/DevExpress/DevExtreme/blob/23_2/packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts#L366)
It will be more obvious (for now at the beginning our flag has ```undefined``` value) and suitable for class migration.

3) Our ```warning is thrown``` never set's back to ```false``` value
It means that if we show this warning once, we never show it again
But I think use cases where we should exist, e.g. toggling the ```height``` option of the ```DataGrid```

packages/devextreme/testing/testcafe/tests/dataGrid/scrolling.ts
```diff
@@ -1624,4 +1625,49 @@ test('Warning should be thrown if scrolling is virtual and height is not specifi
...
+  },
+  dataSource: [
+    { column: 'value' },
+  ],
+  height: 200,
 }));
+
+['height', 'max-height'].forEach((cssOption) => {
+  test(`Warning should not be thrown if scrolling is virtual and height is specified with css (${cssOption})`, async (t) => {
+    await t.debug();
```
williamvinogradov: Let's remove this redundant ```t.debug``` :)

williamvinogradov (approved):

PR #25880, opened 2023-10-25
DataGrid: Fix broken layout after resizing window when there are fixed and band columns (T1193153)

Alyar666:
See https://devexpress.com/issue=T1193153


packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
```diff
@@ -1119,15 +1119,17 @@ export class ColumnsView extends viewWithColumnStateMixin {
...
         for (let rowIndex = 0; rowIndex < $rows.length; rowIndex++) {
+          let $cell: dxElementWrapper | undefined;
           const $row = $rows.eq(rowIndex);
-
           const visibleIndex = this.getVisibleColumnIndex(columnIndex, rowIndex);
 
-          const $cell = $row.hasClass(GROUP_ROW_CLASS)
-            ? $row.find(`td[aria-colindex='${visibleIndex + 1}']:not(.${GROUP_CELL_CLASS})`)
-            : $row.find('td').eq(visibleIndex);
+          if (visibleIndex >= 0) {
```
williamvinogradov, line 1124: As far as I can see, we do nothing if the ```visibleIndex``` variable less than zero.
Let's rewrite this for-loop to make this code more clean and obvious :)

williamvinogradov (approved):

PR #25919, opened 2023-10-30
HtmlEditor: fix dialogs' design issues

marker-dao:


iBat (dismissed):

packages/devextreme/scss/widgets/fluent/htmlEditor/_index.scss
```diff
@@ -273,6 +279,10 @@
   }
 }
 
+.dx-formdialog .dx-formdialog-field-target .dx-field-item-content {
```
EugeniyKiyashko: There is rule in base/htmleditor.scss 
```
.dx-formdialog-field-target .dx-field-item-content {
  text-align: end;
}
```
So, it will be enough to use two selectors and we may don't use additional `.dx-form-dialog` selector for overriding the `text-align` property.

packages/devextreme/scss/widgets/fluent/htmlEditor/_sizes.scss
```diff
@@ -9,6 +9,8 @@ $fluent-html-editor-horizontal-padding: null !default;
 $fluent-html-editor-add-image-dialog-fileuploader-padding: null !default;
 $fluent-html-editor-toolbar-menu-separator-margin: 4px !default;
 $fluent-html-editor-fileuploader-input-wrapper-border-radius: 8px !default;
+$fluent-html-editor-fileuploader-input-wrapper-border-size: 1.5px !default;
```
EugeniyKiyashko, line 12: I don't think that using not integer values in this case is justified. It will better to change color of the border for better contrast.

packages/devextreme/scss/widgets/fluent/htmlEditor/_sizes.scss
```diff
@@ -9,6 +9,8 @@ $fluent-html-editor-horizontal-padding: null !default;
...
+$fluent-html-editor-add-image-dialog-tabs-padding: 12px !default;
```
EugeniyKiyashko: Looks like that padding should be 14px, to align the tab text relative to the content. It is also necessary to consider the compact mode and put the correct value for it

packages/devextreme/scss/widgets/fluent/htmlEditor/_sizes.scss
marker-dao: This was discussed with the designer.

packages/devextreme/scss/widgets/fluent/htmlEditor/_sizes.scss
marker-dao: [3a04d935cfc1585f5a0bedb48aba6ef6120c609d]

packages/devextreme/scss/widgets/fluent/htmlEditor/_index.scss
marker-dao: [54f050216879d6bb4a1219a7e04ed8976cb02a15]

EugeniyKiyashko (approved):

AlexanderMoiseev (approved):

alexanderPolosatov (approved):

Shuvaevmedia (approved):

PR #25952, opened 2023-11-02
TextEditor:  'outside' label should be aligned when before buttons used and input should be focused after label click

EugeniyKiyashko:


Shuvaevmedia (approved):

alexanderPolosatov (approved):

packages/devextreme/js/ui/text_box/ui.text_editor.label.js
```diff
@@ -69,6 +65,34 @@ class TextEditorLabel {
...
+        this._attachEvents();
+    }
+
+    _attachEvents() {
+        const clickEventName = addNamespace(click, this.NAME);
+        const hoverStartEventName = addNamespace(hoverStart, this.NAME);
+        const activeEventName = addNamespace(active, this.NAME);
+
+        eventsEngine.off(this._$labelSpan, clickEventName);
+        eventsEngine.off(this._$labelSpan, hoverStartEventName);
```
ksercs, line 78: why don't we unsubscribe active handler?

packages/devextreme/js/ui/text_box/ui.text_editor.label.js
```diff
@@ -69,6 +65,34 @@ class TextEditorLabel {
...
+
+        if(this._isVisible() && this._isOutsideMode()) {
+            eventsEngine.on(this._$labelSpan, clickEventName, (e) => {
+                const selectedText = getWindow().getSelection().toString();
+
+                if(selectedText === '') {
+                    this._props.onClickHandler();
+                    e.preventDefault();
+                }
```
ksercs, line 87: This code seems strange for me. As far as i understand click in selected text will not trigger input's focus. I'm not sure it's ok. What if selection is somewhere not in label? Maybe we can user pointerup here or detect selection other way?

packages/devextreme/js/ui/text_box/ui.text_editor.label.js
```diff
@@ -111,13 +139,24 @@ class TextEditorLabel {
...
+        this._updateLabelTransform();
+    }
+
+    _updateLabelTransform(offset = 0) {
+        this._$labelSpan.css('transform', '');
+
+        if(this._isVisible() && this._isOutsideMode()) {
+            const sign = this._props.rtlEnabled ? 1 : -1;
+
+            this._$labelSpan.css('transform', 'translateX(' + sign * (getWidth(this._$before) + offset) + 'px)');
```
ksercs, line 152: Let's move offset's calculation result to a separate variable to improve readability.
It's also better to use a string literal.

packages/devextreme/js/ui/text_box/ui.text_editor.label.js
```diff
@@ -111,13 +139,24 @@ class TextEditorLabel {
 
     _updateBeforeWidth() {
         this._$before.css({ width: this._props.beforeWidth });
+
+        this._updateLabelTransform();
+    }
+
+    _updateLabelTransform(offset = 0) {
```
ksercs, line 146: It's sad that we need js calculations again for labels width but I don't see a good approach to implement if fast another way. So let's leave it as it is and create a separate card for solution's rework.

EugeniyKiyashko:
I agree with all the comments. They will be fixed in the next PR. I am going create a separate card to reconsider decisions.

PR #25960, opened 2023-11-06
HtmlEditor: fix mention render in React strict mode (T1187274)

iBat:


packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
```diff
@@ -161,32 +163,53 @@ QUnit.module('Mention format', () => {
...
+            const node = MentionFormat.create({ marker: '@', value: 'val', id: 'id' });
+            spy = sinon.spy(MentionFormat.prototype, 'renderContent');
 
-        assert.strictEqual(element.innerText, '@John Smith');
+            new MentionFormat({}, node);
+
+            assert.ok(spy.calledOnce);
+            assert.strictEqual(spy.getCall(0).args[0], node.children[0]);
+        } finally {
+            spy && spy.restore();
```
alexanderPolosatov: this condition really needed?
PS: use operator ?. instead.

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
```diff
@@ -161,32 +163,53 @@ QUnit.module('Mention format', () => {
...
+        new MentionFormat({}, element);
+        assert.strictEqual(element.children[0].innerText, 'test');
 
         MentionFormat.removeTemplate({ marker: '@', editorKey: 'my_key_in_storage' });
         element = MentionFormat.create(data);
+        new MentionFormat({}, element);
+        assert.strictEqual(element.children[0].innerText, '@John Smith');
+    });
+
+    test('async template renders correctly in React strict mode (T1187274)', function(assert) {
```
alexanderPolosatov: this test know nothing about react. I suggest use kind of name like this: "mention content should be rendered to <smth> container/place" because this test checked this behavior

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
```diff
@@ -161,32 +163,53 @@ QUnit.module('Mention format', () => {
...
+        let spy;
+        try {
+            const node = MentionFormat.create({ marker: '@', value: 'val', id: 'id' });
+            spy = sinon.spy(MentionFormat.prototype, 'renderContent');
```
alexanderPolosatov: can we use heir of MentionFormat instead prototype using

alexanderPolosatov (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
```diff
@@ -142,7 +142,9 @@ QUnit.module('Mention format', () => {
         assert.strictEqual(element.dataset.marker, '@', 'correct marker');
         assert.strictEqual(element.dataset.mentionValue, 'John Smith', 'correct value');
         assert.strictEqual(element.dataset.id, 'JohnSm', 'correct id');
-        assert.strictEqual(element.innerText, '@John Smith', 'correct inner text');
+
+        new MentionFormat({}, element);
```
ksercs, line 155: Won't it work if we only call this usual constructor instead of a static `create` method above?

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
ksercs: We try to not use a private api in tests.
I believe there should be a way to test it better.
Give me time to check

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
```diff
@@ -441,8 +464,10 @@ QUnit.module('Mentions module', moduleConfig, () => {
             return;
         }
 
+        const totalItems = 60;
+        const pageSize = 50;
```
ksercs, line 461: Why does your change affect this test?

iBat: This test remains with unchanged behavior. But locally it was red, so I did a little refactoring while debugging.

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
iBat: I did some changes to avoid it.

ksercs: I suggest u to write a more "live" test using the idea from of this HtmlEditor's configuration:
```javascript
    templatesRenderAsynchronously: true,
    integrationOptions: {
        templates: {
            mention: {
                render: function({ model, container, onRendered }) {
                    setTimeout(() => {
                        const $span = $('<span>');
                        $span.attr({ 'data-id': "custom" });
                        $span.text(`${model.marker}${model.value}`);
                        container.append($span);
                    }, 1000);
                }
            }
        }
    },
```
Your actual code gives this markup:
```html
<span class="dx-mention" spellcheck="false" data-marker="@" data-mention-value="Kevin Carter" data-id="undefined">
    &#xFEFF;
    <span contenteditable="false">
        <span data-attribute="hello">@John Heart</span>
    </span>
    &#xFEFF;
</span>
```
while a code without your fix gives
```html
<span class="dx-mention" spellcheck="false" data-marker="@" data-mention-value="Kevin Carter" data-id="undefined">
    &#xFEFF;
    <span contenteditable="false"></span>
    &#xFEFF;
    <span data-attribute="hello">@Kevin Carter</span>
</span>
```
I think that better place for such a test if `mentionIntegration.tests.js` file

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
```diff
@@ -142,7 +142,9 @@ QUnit.module('Mention format', () => {
         assert.strictEqual(element.dataset.marker, '@', 'correct marker');
         assert.strictEqual(element.dataset.mentionValue, 'John Smith', 'correct value');
         assert.strictEqual(element.dataset.id, 'JohnSm', 'correct id');
-        assert.strictEqual(element.innerText, '@John Smith', 'correct inner text');
+
+        new MentionFormat({}, element);
+        assert.strictEqual(element.children[0].innerText, '@John Smith', 'correct inner text');
```
ksercs, line 156: [atomic test]
Now this test checks 2 different points:
1. ` MentionFormat.create(data);` creates a node with passed data attributes
2. ` new MentionFormat({}, element);` renders content getting the attributes from passed node

I suggest you to split this tests to make it atomic and more clear.

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
```diff
@@ -161,32 +163,55 @@ QUnit.module('Mention format', () => {
         };
 
         const element = MentionFormat.create(data);
-        assert.strictEqual(element.innerText, '#John Smith', 'correct inner text');
+
+        new MentionFormat({}, element);
+        assert.strictEqual(element.children[0].innerText, '#John Smith', 'correct inner text');
```
ksercs: [copypaste]
I believe we can just check that `element.attr('data-marker')` is correct.
We don't need to call constructor here because its work is tested in other place.

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
```diff
@@ -161,32 +163,55 @@ QUnit.module('Mention format', () => {
...
     });
 
     test('Change default content renderer', function(assert) {
-        const data = {
+        const expectedData = {
```
ksercs: [naming]
```suggestion
        const nodeData = {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionIntegration.tests.js
```diff
@@ -604,5 +604,42 @@ export default function() {
             this.$element.find('p').first().text('@');
             this.clock.tick(10);
         });
+
+        test('Mention content should be rendered to embed container (T1187274)', function(assert) {
+            const done = assert.async();
+            const that = this;
```
ksercs, line 610: I don't think we need this

ksercs (approved):

alexanderPolosatov (approved):

PR #25968, opened 2023-11-07
DataGrid: Fix the "Cannot read properties of undefined (reading 'selector')" error occurs when a column's dataField is not specified and repaintChangesOnly is enabled in popup edit mode (T1198534)

Alyar666:
See https://devexpress.com/issue=T1198534

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_form_based.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_form_based.ts
@@ -318,7 +318,7 @@ const editingControllerExtender = (Base: ModuleType<EditingController>) => class
   }
 
   getFormEditorTemplate(cellOptions, item) {
-    const column = this.component.columnOption(item.dataField);
+    const column = this.component.columnOption(item.name || item.dataField);
 
     return (options, container) => {
       const $container = $(container);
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_form_based.ts
```diff
@@ -318,7 +318,7 @@ const editingControllerExtender = (Base: ModuleType<EditingController>) => class
   }
 
   getFormEditorTemplate(cellOptions, item) {
-    const column = this.component.columnOption(item.dataField);
+    const column = this.component.columnOption(item.name || item.dataField);
```
pomahtri, line 321:
```suggestion
    const column = this.component.columnOption(item.name ?? item.dataField);
```

Alyar666: @pomahtri, I think it's better to leave '||'.
Because in this case: '' ?? 'field', the empty string will win

pomahtri: isn't it okay?

Alyar666: > isn't it okay?

It seems to me that not. Because the name is an identifier and it must contain a unique value. If it is empty ('' || null || undefined) we should target the dataField.

pomahtri (approved):

PR #25995, opened 2023-11-09
Tabs: Set focusedElement when selectedIndex, selectedItems, selectedItems was changed (T1193394)

marker-dao:


alexanderPolosatov (dismissed):

AlexanderMoiseev (dismissed):

EugeniyKiyashko:
The changes do not provide for cases of changing 'selectedItem', 'selectedItems' options in runtime. Also, there are no tests for scenario from source ticket.

EugeniyKiyashko (approved):

packages/devextreme/js/ui/tabs.js
```diff
@@ -681,6 +681,17 @@ const Tabs = CollectionWidget.inherit({
         this._toggleFocusedDisabledPrevClass(currentIndex, shouldPrevClassBeSetted);
     },
 
+    _updateFocusedElement() {
+        const { focusStateEnabled, selectedIndex } = this.option();
```
EugeniyKiyashko, line 685: Perhaps, we may think about clearing 'focusedElement' option at the right moment. It allows us to use logic from collection that determine correct element that should be focused on focusIn. 

Something like this 
```
  _focusOutHandler(e) {
        this.callBase(e);

        this.option('focusedElement', null);
    },
```

PR #26013, opened 2023-11-14
DataGrid: Fix memory leak occurs when changes are pushed and 'repaintChangesOnly' is enabled (T1196383)

Alyar666:
See https://devexpress.com/issue=T1196383

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
@@ -45,7 +45,7 @@ function isGroupRow({ rowType, column }) {
     && !column.showWhenGrouped && !column.command;
 }
 
-function watch({
+function setWatcher({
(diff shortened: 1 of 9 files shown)


Raushen (dismissed):

packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
```diff
@@ -45,6 +45,16 @@ function isGroupRow({ rowType, column }) {
     && !column.showWhenGrouped && !column.command;
 }
 
+function watch({
+  element, watch, getter, callBack,
+}) {
```
pomahtri: Let's rename function name or argument name. Or maybe inline this function
It's unsafe to have the same name for func and for variable inside of it, there was a case when we needed to rename it because of some issues in demos: https://github.com/DevExpress/DevExtreme/pull/25850/

pomahtri (approved):

PR #26025, opened 2023-11-15
Tabs: Set focusedElement to null when out of focus (T1193394)

marker-dao:


packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -506,7 +506,7 @@ QUnit.module('Tab select action', () => {
     });
 
     QUnit.test('focusedElement must be changed after changing the selectedIndex', function(assert) {
```
EugeniyKiyashko, line 508: We can use script iteration with a change in selectedIndex, selectedItem(s) to avoid duplication test scenarios.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -506,7 +506,7 @@ QUnit.module('Tab select action', () => {
...
-        assert.expect(2);
+        assert.expect(3);
```
EugeniyKiyashko: This code is not asyncroniously. All of your assertion will always execute . I suggest there is no need for this line.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
marker-dao: No, we can't. There are too many differences in configurations for selectedIndex, selectedItem and selectedItems.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
marker-dao: [25e09d20c8724be348f469868a7e2f46a2e55633]

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
EugeniyKiyashko: Okay, if that's more convenient, let's not merge then

PR #26035, opened 2023-11-15
DateRangeBox: hover style should be applied on second date pick in disableDatesOutOfRange mode

Zedwag:


ksercs (dismissed):

packages/devextreme/testing/testcafe/tests/editors/dateRangeBox/calendar.ts
```diff
@@ -696,6 +696,39 @@ test('Disabled dates on inputs focus (disableOutOfRangeSelection: true)', async
   }, '#dateRangeBox');
 });
 
+test('Hover styles appearance on second date pick (disableOutOfRangeSelection=true)', async (t) => {
```
ksercs: shouldn't we just check that class is added instead of making a screenshot?

Zedwag: great idea! will do

EugeniyKiyashko (approved):

PR #26039, opened 2023-11-16
Optimize license-check performance

IlyaKhD:
This reduces execution time of license-check and makes it insignificant

| | Before | After |
| -- | -- | -- |
| license-check execution time | 750ms | 4ms |

### DIFF
--- a/packages/devextreme/js/__internal/core/license/rsa_bigint.ts
+++ b/packages/devextreme/js/__internal/core/license/rsa_bigint.ts
@@ -24,7 +24,7 @@ export function compareSignatures(args: Args): boolean {
         e >>= one; // eslint-disable-line no-bitwise
       }
 
-      return result % modulus;
+      return result;
     };
 
     const bigIntFromBytes = (bytes: Uint8Array): bigint => bytes.reduce(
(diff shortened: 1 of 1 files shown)


packages/devextreme/js/__internal/core/license/rsa_bigint.ts
```diff
@@ -8,11 +8,27 @@ interface Args {
 export function compareSignatures(args: Args): boolean {
   try {
     const zero = BigInt(0);
+    const one = BigInt(1);
     const eight = BigInt(8);
 
+    const modExp = (base: bigint, exponent: bigint, modulus: bigint): bigint => {
```
timbset, line 14: `calculateModulusOfExponentiation`? And why is it an inner function? Isn't it better to extract it into a module root?

VasilyStrelyaev (dismissed):

packages/devextreme/js/__internal/core/license/rsa_bigint.ts
IlyaKhD: > And why is it an inner function? Isn't it better to extract it into a module root?

To keep all calls to `BigInt` within try-catch

IlyaKhD: As for me it's OK to use common shortenings in names, such as maths operations

timbset (approved):

PR #26057, opened 2023-11-20
DataGrid: Fixes sending the same request twice when using filterValue and the lookup column (T1195882)

Alyar666:
See https://devexpress.com/issue=T1195882

### DIFF
--- a/js/__internal/grids/grid_core/filter/m_filter_row.ts
+++ b/js/__internal/grids/grid_core/filter/m_filter_row.ts
@@ -490,7 +490,7 @@ const ColumnHeadersViewFilterRowExtender = (function () {
 
       if (options.lookup && this.option('syncLookupFilterValues')) {
         filterRowController.setCurrentColumnForFiltering(options);
-        const filter = this.getController('data').getCombinedFilter();
+        const filter = this.getController('data').getCombinedFilter() || null;
         filterRowController.setCurrentColumnForFiltering(null);
 
         const lookupDataSource = gridCoreUtils.getWrappedLookupDataSource(options, dataSource, filter);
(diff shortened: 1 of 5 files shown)


js/__internal/grids/grid_core/filter/m_filter_row.ts
```diff
@@ -673,10 +673,13 @@ const ColumnHeadersViewFilterRowExtender = (function () {
     },
 
     _handleDataChanged(e) {
```
williamvinogradov, line 675: Minor comment.

Let's refactor this function a little
Something like this:
```suggestion
    _handleDataChanged(e) {
      this.callBase.apply(this, arguments);
      
      const { filtering, fullReload } = e.operationTypes;
      if (!filtering && !fullReload) {
        return;
      }
      
       const dataSource = this._dataController?.dataSource?.();
       const lastLoadOptions = dataSource?.lastLoadOptions?.();
       const loadOptions = filtering ?? lastLoadOptions?.filter;
       this.updateLookupDataSource(loadOptions);
    }
```

js/__internal/grids/grid_core/filter/m_filter_row.ts
```diff
@@ -713,7 +716,7 @@ const ColumnHeadersViewFilterRowExtender = (function () {
 
           const editorDataSource = editor.option('dataSource');
           const shouldUpdateFilter = !filterChanged
-                        || !equalByValue(editorDataSource.__dataGridSourceFilter, filter);
+                        || !equalByValue(editorDataSource.__dataGridSourceFilter || null, filter);
```
williamvinogradov, line 719: Question.

Can we use the ```??``` instead of the ```||``` here?

williamvinogradov (approved):

PR #26073, opened 2023-11-23
DataGrid - Fix the E0110 error that occurs when editing a column with setCellValue in form mode (T1193894)

markallenramirez:


packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_form_based.ts
```diff
@@ -563,6 +563,18 @@ export const editingFormBasedModule = {
             this.callBase($cell, parameters);
           }
         },
+        _updateContent() {
+          const editingController = this._editingController;
```
pomahtri, line 567: minor: we usually (in new code) avoid creating variables like this and just use `this._editingController`: "editingController" isn't much shorter, but loses context (where we got this variable from) 
```suggestion
          const editingController = this._editingController;
```

pomahtri (approved):

Alyar666 (approved):

Alyar666 (approved):

PR #26080, opened 2023-11-23
DataGrid - Cannot focus an inserted row if it appears on the same position as the currently focused row (T1202646)

pomahtri:

### DIFF
--- a/packages/devextreme/testing/testcafe/tests/dataGrid/focus/focusedRow/focusedRow.ts
+++ b/packages/devextreme/testing/testcafe/tests/dataGrid/focus/focusedRow/focusedRow.ts
@@ -876,3 +876,38 @@ test('Focused row should be shown after reloading the page (T1058983)', async (t
 }).after(async () => {
   await clearLocalStorage();
 });
+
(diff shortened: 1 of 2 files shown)


markallenramirez (approved):

packages/devextreme/testing/testcafe/tests/dataGrid/focus/focusedRow/focusedRow.ts
```diff
@@ -876,3 +876,38 @@ test('Focused row should be shown after reloading the page (T1058983)', async (t
...
+
+  await t.click(dataGrid.getDataRow(0).element);
+
+  await ClientFunction(() => {
+    const grid = ($('#container') as any).dxDataGrid('instance');
+    grid.getDataSource().store().push([{
+      type: 'insert',
+      data: { value: 2 },
+    }]);
+  })();
```
williamvinogradov, line 891: Question.

Maybe we should add this ClientFunction to the ```DataGrid``` testing model?
Something like method with the name ```pushDataToTheStore``` or something :)

williamvinogradov (approved):

PR #26086, opened 2023-11-24
Chart - Remove rounding values on translation for line/area series(T1195064)

alexanderPolosatov:


Zedwag (dismissed):

packages/devextreme/testing/tests/DevExpress.viz.core/translator2D.tests.js
```diff
@@ -933,6 +933,12 @@ QUnit.test('getRangeByMinZoomValue', function(assert) {
     assert.deepEqual(translator.getRangeByMinZoomValue(100, { minVisible: 650, maxVisible: 690 }), [590, 690]);
 });
 
+QUnit.test('Do not round value when skipRound=true', function(assert) {
```
ksercs: Don't we wanna also test an end scenario in symbol point?
This tests checks the work of translator but we don't check that this API is used in SymbolPoint

packages/devextreme/js/viz/series/points/symbol_point.js
```diff
@@ -578,9 +578,9 @@ export default {
             that.minX = valTranslator.translate(that.minValue);
             that.defaultX = valTranslator.translate(CANVAS_POSITION_DEFAULT);
         } else {
-            that.vy = that.y = valTranslator.translate(that.value);
-            that.vx = that.x = argTranslator.translate(that.argument);
-            that.minY = valTranslator.translate(that.minValue);
+            that.vy = that.y = valTranslator.translate(that.value, undefined, true);
```
ksercs, line 581: I believe we should rework api of translator in the future but agree to leave it as it is now while we have only few places where it's necessary

packages/devextreme/testing/tests/DevExpress.viz.core/translator2D.tests.js
alexanderPolosatov: integration tests were added

ksercs:
```suggestion
QUnit.test('value is not rounded when skipRound=true', function(assert) {
```

packages/devextreme/js/viz/translators/translator2d.js
```diff
@@ -184,7 +184,9 @@ _Translator2d.prototype = {
         });
         that._oldMethods = Object.keys(script);
         extend(that, script);
-        that._conversionValue = options.conversionValue ? function(value) { return value; } : function(value) { return Math.round(value); };
+        that._conversionValue = options.conversionValue ? function(value) { return value; } : function(value, skipRound) {
```
ksercs: I think it's better to split to several strings

packages/devextreme/testing/tests/DevExpress.viz.charts/chartInteraction.tests.js
```diff
@@ -813,3 +815,87 @@ QUnit.module('Deprecated options', {
...
+            }, {
+                arg: 1.05,
+                val: 500.01
+            }]
+        };
+
+        return $('#chart').dxChart($.extend(true, {}, defaultOptions, options)).dxChart('instance');
+    }
+}, () => {
+
```
ksercs:
```suggestion
```

packages/devextreme/testing/tests/DevExpress.viz.charts/chartInteraction.tests.js
```diff
@@ -813,3 +815,87 @@ QUnit.module('Deprecated options', {
...
+    function getMarkerCoords(element) {
```
ksercs: `coords` seems strange for me. I think it's better to not skimp for 3 letters and use common approach. It can simplify work for other developers who does not have a habit to short such words

packages/devextreme/testing/tests/DevExpress.viz.charts/chartInteraction.tests.js
```diff
@@ -813,3 +815,87 @@ QUnit.module('Deprecated options', {
...
+        const coordsStr = transformAttribute
+            .replace('translate(', '')
+            .replace(')', '');
+
+        return coordsStr
+            .split(',')
+            .map((coord) => parseFloat(coord));
+    }
+
+    QUnit.test('Coords of points should be different when distance of values too small', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('Coordinates of points should be different when distance between values is too small', function(assert) {
```

ksercs: What about bug ID?

ksercs (approved):

EugeniyKiyashko (approved):

PR #26118, opened 2023-11-29
DataGrid - The collapse icon is missing in a group row in certain usage scenarios

Raushen:
See https://supportcenter.devexpress.com/ticket/details/t1201981/datagrid-the-collapse-icon-is-missing-in-a-group-row-in-certain-usage-scenarios


packages/devextreme/testing/testcafe/tests/dataGrid/grouping/grouping.ts
```diff
@@ -304,3 +304,42 @@ test('Group panel message should be vertically aligned (T1186613)', async (t) =>
     ],
   },
 }));
+
+test('The collapse icon should update if repaintChangesOnly option is enabled (T1201981)', async (t) => {
```
markallenramirez, line 308: Hi,

remoteOperations should be enabled and the last page should be navigated first before the previous pages (3 -> 2 -> 1) to actually reproduce the issue.

Currently, this test would yield the same attached etalons regardless of whether the solution is applied or not.

Raushen: Hi,
To reproduce the issue it is enough to navigate the next page (1 -> 2).
But you are absolutely right - I forgot to configure repaintChangesOnly in the test. Thank you. 

Attached you can find the screenshot without the solution (the problem at the "Field 1: 2" row expand button).

continued_group-collapse_icon-T1201981

markallenramirez (approved):

PR #26138, opened 2023-12-01
Chart: fix tooltip for hovered series after zooming (T1197697)

AlexanderMoiseev:


packages/devextreme/js/viz/utils.js
```diff
@@ -67,6 +67,25 @@ export {
     refreshPaths
 };
 
+export const areCanvasesDifferent = function(canvas1, canvas2) {
+    const size_changing_threshold = 1;
```
alexanderPolosatov: previously, this constant was in uppercase and lived in top of file

packages/devextreme/testing/tests/DevExpress.viz.charts/zoomAndPan.tests.js
```diff
@@ -3,6 +3,8 @@ import pointerMock from '../../helpers/pointerMock.js';
 
 import 'viz/chart';
 
+const CHART_CLASS = 'svg.dxc.dxc-chart';
```
alexanderPolosatov: I suggest use smth like "CHART_SVG_SELECTOR" for name of variable

packages/devextreme/testing/tests/DevExpress.viz.charts/zoomAndPan.tests.js
```diff
@@ -822,6 +824,70 @@ QUnit.test('Multiaxes, zoom axes only in one pane', function(assert) {
...
+        ],
+    }).dxChart('instance');
+
+    const $chartElement = $(chart.$element().find(CHART_CLASS));
+    const pointer = pointerMock($chartElement[0]).start();
+
+    pointer.start({ x: 100, y: 200 }).wheel(150);
+    $chartElement.trigger($.Event('dxpointermove', { pointerType: 'mouse', pageX: 54, pageY: 541 }));
+
+    const tooltip = $('.dxc-tooltip')[0];
```
alexanderPolosatov: move classname to constant

packages/devextreme/js/viz/utils.js
AlexanderMoiseev: i suggest renaming it to camelCase. see no point of moving it to the top of the file since there is 1 usage of this const only ( unless we have specific code style rule for this ).

packages/devextreme/js/__internal/viz/chart_components/m_advanced_chart.ts
```diff
@@ -10,6 +10,8 @@ import {
 } from '@js/viz/core/utils';
 import rangeDataCalculator from '@js/viz/series/helpers/range_data_calculator';
 import { Range } from '@js/viz/translators/range';
+// @ts-expect-error
```
iBat, line 13: Which particular error? And why?

AlexanderMoiseev: no file error. i believe this is happening because js file is imported into ts file,  i used the same approach with @ts-expect-error as we use in other places in our project, so this is not something new to our project

iBat (dismissed):

iBat (dismissed):

iBat (approved):

alexanderPolosatov (approved):

PR #26140, opened 2023-12-04
NumberBox: allow to clear the selected text by pressing the decimal separator when the format is specified [T1199553]

iBat:


packages/devextreme/testing/testcafe/tests/editors/numberBox/mask.ts
```diff
@@ -0,0 +1,29 @@
...
+  appendElementTo,
+} from '../../../helpers/domUtils';
+import { clearTestPage } from '../../../helpers/clearPage';
+import NumberBox from '../../../model/numberBox';
+
+fixture.disablePageReloads`NumberBox_Mask`
+  .page(url(__dirname, '../../container.html'))
+  .afterEach(async () => clearTestPage());
+
+test('pressing "." should clear previous value (T1199553)', async (t) => {
```
ksercs:
```suggestion
test('pressing "." should clear selected text if it contains a decimal separator (T1199553)', async (t) => {
```

packages/devextreme/testing/testcafe/tests/editors/numberBox/mask.ts
```diff
@@ -0,0 +1,29 @@
...
+  const textBox = new NumberBox('#numberBox');
```
ksercs:
```suggestion
  const numberBox = new NumberBox('#numberBox');
```

packages/devextreme/js/ui/number_box/number_box.mask.js
```diff
@@ -427,6 +427,10 @@ const NumberBoxMask = NumberBoxBase.inherit({
             return undefined;
         }
 
+        if(isTextSelected && parsedValue === null && editedText === decimalSeparator) {
+            return 0;
+        }
+
```
ksercs: I found the following behavior strange:
1. If selected texts ends before the decimal separator, press `.` will move selection to cover the separator.
2. Mask value is not cleared if minus sign or stub text is not selected.

I don't think we can rework this behavior fast - we need to make changed is core format parsing. Let's add this cases to an exidting backlog card.

About this condition - I prefer we make a condition `isWholeTextSelected` to make this condition clearer. Not this condition purely described real scenario

packages/devextreme/testing/testcafe/tests/editors/numberBox/mask.ts
```diff
@@ -0,0 +1,29 @@
...
+    .pressKey('end shift+left shift+left shift+left shift+left shift+left shift+left shift+left shift+left .')
+    .expect(input.value)
+    .eql('0.00');
+}).before(async () => {
+  await appendElementTo('#container', 'div', 'numberBox', { });
+  await createWidget('dxNumberBox', {
+    format: '#0.00',
+    value: 14500,
+  }, '#numberBox');
+});
```
ksercs: We can write a unit test instead of this:
```javascript
        this.instance.option({
            format: '0#.00',
            value: 123.45
        });

        this.keyboard
            .caret({ start: 0, end: 6 })
            .type('.');

        assert.strictEqual(this.input.val(), '0.00', 'mask value is cleared');
```

packages/devextreme/js/ui/number_box/number_box.mask.js
iBat: https://trello.com/c/Mf9HgNyC/8263-numberbox-mask-refactoring

ksercs (approved):

alexanderPolosatov (approved):

AlexanderMoiseev (approved):

EugeniyKiyashko (approved):

PR #26149, opened 2023-12-05
DataGrid - Modified cells should not require multiple Enter keys before the focus is moved to the next cell if showEditorAlways is enabled and enterKeyAction is set to 'startEdit' (T1196539)

markallenramirez:


packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1442,6 +1442,8 @@ export class KeyboardNavigationController extends modules.ViewController {
             const isFocusedElementDefined = isElementDefined(
               $focusedElementInsideCell,
             );
+            const columnIndex = this._rowsView.getCellIndex($cell);
+            const column = this.getController('columns').getVisibleColumns()[columnIndex];
```
Raushen: I found it is needed to take into account the [columnIndexOffset](https://github.com/DevExpress/DevExtreme/blob/1db274a1b82f1eafb565c0dd82bb0318b4aa1c87/packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts#L854) when you take a column. Maybe it will help if you extract a method like getColumnByCellElement.

packages/devextreme/testing/testcafe/tests/dataGrid/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -12,6 +12,50 @@ const CLASS = ClassNames;
...
+test('Next cell should be focused immediately on a single Enter key press if showEditorAlways is enabled in cell mode (T1196539)', async (t) => {
+  const dataGrid = new DataGrid('#container');
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+  await t
+    .click(dataGrid.getDataCell(0, 0).element)
+    .typeText(dataGrid.getDataCell(0, 0).element, 'test')
+    .pressKey('enter');
+
+  await takeScreenshot('cell-focus-showEditorAlways-T1196539-1.png', dataGrid.element);
```
Raushen: How about avoid screenshots and use properties like "isEditCell", "isFocused", etc?

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
markallenramirez: columnIndexOffset is used depending on whether we get all the visible columns or only the rendered ones.
Here's a different case getting only the rendered ones: [m_keyboard_navigation.ts#L1211](https://github.com/DevExpress/DevExtreme/blob/1db274a1b82f1eafb565c0dd82bb0318b4aa1c87/packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts#L1211) - here columnIndexOffset is not used

Anyway, I refactored what I can, except this part: https://github.com/DevExpress/DevExtreme/pull/26149/files#diff-cb3dbff86c1db4ac00da48f26a0f18d218f4e696d3b3b1e75785d419be1bf327R1831
since 'columnIndex' is used later in the code

packages/devextreme/testing/testcafe/tests/dataGrid/keyboardNavigation/keyboardNavigation.functional.ts
markallenramirez: Hello,

I get rid of other screenshots. I left one screenshot test because there is no property to check if the editor's text is selected or not. The isEditCell property can't be used here because showEditorAlways is enabled and the property is always true. Technically, I could check input selection, but I chose a screenshot test instead :)

Raushen (approved):

PR #26152, opened 2023-12-05
Align switch in toolbar to vertical center in material and fluent themes (T1197045)

iBat:


packages/devextreme/testing/testcafe/tests/navigation/toolbar/common.ts
```diff
@@ -162,6 +164,15 @@ const supportedWidgets = ['dxAutocomplete', 'dxCheckBox', 'dxDateBox', 'dxMenu',
           text: 'opts.stylingMode: contained',
         },
       },
+      {
```
EugeniyKiyashko, line 167: This test includes scenario with dropDownButton component. We may choose other test that test appearance all components in toolbar, for example tests with names: 'all widgets render', 'nested widget appearance'. Don't forget to test this case in multiline toolbar mode.

packages/devextreme/scss/widgets/fluent/toolbar/_index.scss
```diff
@@ -33,6 +33,10 @@
       height: $fluent-toolbar-height;
     }
   }
+
+  .dx-toolbar-item .dx-item-content {
+    display: flex;
```
EugeniyKiyashko: I believe we can't afford such a general solution. There's no justification for making all content containers flex-based. Besides, the content aligns correctly at the center. I suggest looking for a solution at the switch level in the toolbar item and/or fix issues in the switch component styles.

EugeniyKiyashko (approved):

EugeniyKiyashko:
Let's add BugID to PR caption. Also, the name should clearly reflect the problem we are fixing.

AlexanderMoiseev (approved):

PR #26154, opened 2023-12-05
TabPanel: Fix tabpanel container height (T1203780)

marker-dao:


packages/devextreme/testing/tests/DevExpress.ui.widgets/tabPanel.tests.js
```diff
@@ -158,6 +159,25 @@ QUnit.module('rendering', {
...
+
+        const $tabPanel = $('<div>').appendTo('#qunit-fixture').dxTabPanel({
+            items,
+            width: 100,
+            height: 100,
+        });
+
+        const $tabPanelContainer = $tabPanel.find(`.${TABPANEL_CONTAINER_CLASS}`);
+
+        assert.strictEqual($tabPanelContainer.get(0).clientHeight, 62);
```
iBat: Why 62 if you set 100 via options?

EugeniyKiyashko: 'tabpanel-container' is element that wrap multiview. So, it's size exclude tabs container.

marker-dao: dx-tabs height is 38px. dx-tabpanel-container is another element under dx-tabs, its height is 62px.

packages/devextreme/scss/widgets/base/_tabPanel.scss
```diff
@@ -48,4 +48,5 @@
 
 .dx-tabpanel-container {
   flex-grow: 1;
+  overflow: hidden;
```
EugeniyKiyashko: I suggest better way is using 'min-height: 0' for horizontal orientation, and 'min-width: 0' for vertical orientation. Also, the initial issue presented in our demo. I think we may improve our demo code.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabPanel.tests.js
EugeniyKiyashko: I think we should test both of tabpanel configurations - vertical & horizontal.

packages/devextreme/scss/widgets/base/_tabPanel.scss
marker-dao: [318192470d093380e5908af4d01920d6e2ffd229]

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabPanel.tests.js
marker-dao: [318192470d093380e5908af4d01920d6e2ffd229]

EugeniyKiyashko (approved):

AlexanderMoiseev (approved):

PR #26155, opened 2023-12-05
DataGrid: Fix closing editable cell when mouse pointer is dragged to copy data to other cells of the current row (T1203250)

Alyar666:
See https://devexpress.com/issue=T1203250

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
@@ -2570,7 +2570,7 @@ export const editingModule = {
             return isShowEditorAlways && allowEditing && editingController.editCell(e.rowIndex, columnIndex);
           }
 
-          if (eventName === 'click' && startEditAction === 'dblClick') {
+          if (eventName === 'click' && startEditAction === 'dblClick' && this._pointerDownTarget === $targetElement.get(0)) {
             const isError = false;
             const withoutSaveEditData = row?.isNewRow;
             editingController.closeEditCell(isError, withoutSaveEditData);
@@ -2581,6 +2581,7 @@ export const editingModule = {
           }
         },
         _rowPointerDown(e) {
+          this._pointerDownTarget = e.event.target;
           this._pointerDownTimeout = setTimeout(() => {
             this._editCellByClick(e, 'down');
           });
--- a/packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
+++ b/packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
@@ -2707,12 +2707,45 @@ QUnit.module('Editing', {
         sinon.spy(that.editingController, 'closeEditCell');
 
         // act
+        $testElement.find('td').eq(1).trigger('dxpointerdown');
         $testElement.find('td').eq(1).trigger('dxclick');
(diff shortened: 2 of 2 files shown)


packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/editing.tests.js
```diff
@@ -2715,6 +2716,38 @@ QUnit.module('Editing', {
...
+        // act
+        $(this.getCellElement(0, 0)).trigger('dxdblclick');
+        this.clock.tick(10);
+
+        // assert
+        assert.strictEqual($(this.getCellElement(0, 0)).find('input').length, 1, 'has input');
+
+        // act
+        $(this.getCellElement(0, 0)).trigger('dxpointerdown');
+        $(this.getCellElement(0, 1)).trigger('dxclick');
```
williamvinogradov: Question.

Can we use here the ```pointerEvents``` from the ```devextreme/js/events/pointer``` instead of event names in strings? :)

williamvinogradov: Minor comment.

I think it will be correct to use the ```dxpointerup``` instead of the ```dxclick``` event here.
Because in this issue we get ```pointerdown``` in the edit cell and after it ```pointerup``` in another cell on the row
(Click works similar with ```pointerup``` event only on pointer up)

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
```diff
@@ -2570,7 +2570,7 @@ export const editingModule = {
             return isShowEditorAlways && allowEditing && editingController.editCell(e.rowIndex, columnIndex);
           }
 
-          if (eventName === 'click' && startEditAction === 'dblClick') {
+          if (eventName === 'click' && startEditAction === 'dblClick' && this._pointerDownTarget === $targetElement.get(0)) {
```
williamvinogradov, line 2573: I want to suggest another solution to this issue

As far as I understand, the issue occurs because the ```click``` event triggered on the ```pointerup``` event too.
and the cell's click handler handles the click event on another cell in the same row the wrong way.

So, maybe we can just prevent this scenario [here in the column view](https://github.com/DevExpress/DevExtreme/blob/23_2/packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts#L86) where we subscribe to different mouse event?
Something like this:
```typescript
// change the click event to pointerEvents.up here
eventsEngine.on($table, [dblclickEvent, pointerEvents.down, pointerEvents.up].join(' '), '.dx-row', that.createAction((e) => {
    const { event } = e;
    const { type } = event;
    
    // Remember the event.target on pointer down
    if (type === pointerEvents.down) {
      _test = event.target;
    }

    // If it's pointer up check the target
    if (type === pointerEvents.up && _test !== event.target) {
      return;
    }
    
    // other code
}
```
And by this we can prevent another problem of "false" ```click``` event triggering on ```pointerup```, I guess :)
I checked it locally and this solution works perfectly, but I didn't run the tests on CI

williamvinogradov: DIscussed in a call.
My suggestion is bad because it can affect other existing features like row selection :)

williamvinogradov (dismissed):

williamvinogradov (approved):

PR #26178, opened 2023-12-07
LoadPanel: outside click should not be propagated when shading enabled (T1194035)

Zedwag:


Zedwag:
desided to leave this at it is and give WA

PR #26179, opened 2023-12-07
TagBox - Fix custom item rendering after deleting equal value(T1197444)

alexanderPolosatov:


packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -4770,6 +4770,39 @@ QUnit.module('the \'acceptCustomValue\' option', moduleSetup, () => {
...
+
+        const tagBoxInstance = $tagBox.dxTagBox('instance');
+
+        const $input = $tagBox.find(`.${TEXTEDITOR_INPUT_CLASS}`);
+        const keyboard = keyboardMock($input);
+
+        keyboard.type('1');
+        $input.trigger('change');
+
+        $('.dx-tag-remove-button')
```
ksercs:
```suggestion
        $(`.${TAGBOX_TAG_REMOVE_BUTTON_CLASS}`)
```

packages/devextreme/js/ui/tag_box.js
```diff
@@ -1485,6 +1485,10 @@ const TagBox = SelectBox.inherit({
         return existedItems.concat(newItems);
     },
 
+    _isCustomItemSelected() {
```
ksercs: Let's think a bit more how can we rework the actual algorithm to fix the issue.
Function with name `_isCustomItemSelected` should not return just `acceptCustomValue`. Moreover, as far as i see we already have the check for `acceptCustomValue` in the SelectBox's `valueChangeHandler`. It would be great if we research why we need this code in selectBox and prepare more intuitive solution.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
alexanderPolosatov: fixed

packages/devextreme/js/ui/tag_box.js
alexanderPolosatov: fix reworked

ksercs (approved):

iBat (approved):

PR #26189, opened 2023-12-07
TextBox: clear button should work without focus out if a mask is applied (T1193735)

iBat:


packages/devextreme/testing/testcafe/tests/editors/textBox/mask.ts
```diff
@@ -20,3 +20,20 @@ test('"!" character should not be accepted if mask restricts it (T1156419)', asy
     mask: '9',
   }, '#textBox');
 });
+
+test('clear button should work if input has not been blurred (T1193735)', async (t) => {
```
ksercs: I think that QUnit test is enough here, we can just add an appropriate check to your new QUnit test.
Let's remove this one.

ksercs:
@iBat  please update the PR's name according to our guidelines. It should answer on a questions "what I do" or "how should it work". For example: "TextBox: clear button should clear the input text even if value is not changed yet and mask is applied (T1193735)".

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/textEditorParts/mask.tests.js
```diff
@@ -1782,6 +1782,32 @@ QUnit.module('clear button', () => {
             clock.restore();
         }
     });
+
+    QUnit.test('Input should has valid value after input text & focusout', function(assert) {
```
ksercs: Let's refactor this test to check the initial scenario.

```suggestion
    QUnit.test('Input text should be cleared after click on a clear button even if value is not changed yet (T1193735)', function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/textEditorParts/mask.tests.js
```diff
@@ -1782,6 +1782,32 @@ QUnit.module('clear button', () => {
...
+        const keyboard = keyboardMock($input, true);
+
+        $input.triggerHandler('focus');
+
+        caretWorkaround($input);
+        keyboard.caret(0);
+
+        keyboard.type('9');
+
+        const $clearButton = $textEditor.find('.dx-clear-button-area');
```
ksercs: let's move this class to a constant

packages/devextreme/js/ui/text_box/ui.text_editor.mask.js
```diff
@@ -581,8 +572,24 @@ const TextEditorMask = TextEditorBase.inherit({
...
+        clearTimeout(this._caretTimeout);
+        this.callBase();
+    },
 
+    clear: function() {
+        if(this._maskRulesChain) {
+            this._maskRulesChain.clear(this._normalizeChainArguments());
+            this._updateMaskInfo();
+            const caret = this._maskRulesChain.first();
+            this._caretTimeout = setTimeout(() => {
```
ksercs: We should not use setTimeout here. Such a code always makes future maintenance much more difficult.
How now your solution works:
1.`_clear` method clears maskChain
2. In callbase we call `renderValue` which updates `text` by maskChain's text (empty now)
3. TextBox renders input value in a usual way by setting `.val` and it moves caret to the end of the string
4. Your setTimeout sets caret after the first character.

We can get rid of the setTimeout if use more direct (specific for mask) way to render a value - `_renderMaskedValue`.

I suggest you check `_clear` method of a base editor one more time. It basically
- changes value to a default one
- if value is already default, it does what we usually do on option change

It means we can just check what we do in `value` option change handler and emulate this behavior in `_clear` method.

Smth like:
```javascript
clear: function() {
      const { value: defaultValue } = this._getDefaultOptions();
      if(this.option('value') === defaultValue) {
          this._renderMaskedValue();
      }
      this.callBase();
  }
```
Please, check it out and test one more time. It's just an idea, can be not a final solution.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/textEditorParts/mask.tests.js
```diff
@@ -1782,6 +1782,32 @@ QUnit.module('clear button', () => {
...
+
+        $clearButton.trigger('dxclick');
+        $input.trigger('focusout');
+
+        assert.strictEqual(textEditor.option('value'), '');
```
ksercs, line 1832: let's add a separate test for this test case

packages/devextreme/js/ui/text_box/ui.text_editor.mask.js
```diff
@@ -581,8 +572,20 @@ const TextEditorMask = TextEditorBase.inherit({
             default:
                 this.callBase(args);
         }
-    }
+    },
 
+    _dispose: function() {
+        clearTimeout(this._caretTimeout);
```
ksercs: ?

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/textEditorParts/mask.tests.js
```diff
@@ -1782,6 +1783,57 @@ QUnit.module('clear button', () => {
...
+        caretWorkaround($input);
+        keyboard.caret(0);
+
+        keyboard.type('9');
+
+        const $clearButton = $textEditor.find(CLEAR_BUTTON_CLASS);
+
+        $clearButton.trigger('dxclick');
+
+        assert.strictEqual(textEditor.option('value'), '');
```
ksercs: We should not check value here, we should check input test.
It seems like this test is ever green now.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/textEditorParts/mask.tests.js
```diff
@@ -1782,6 +1783,57 @@ QUnit.module('clear button', () => {
...
+    });
+
+    QUnit.test('Cleared text should not be restored on focusout (T1193735)', function(assert) {
```
ksercs: It's not directly related to the ticket
```suggestion
    QUnit.test('Cleared text should not be restored on focusout', function(assert) {
```

ksercs (approved):

EugeniyKiyashko (approved):

PR #26244, opened 2023-12-12
TypeScript - CustomStore onLoaded result own param type (T1197789)

mpreyskurantov:
### Cherry-picks
- #26343

Changelog:
- Move the StoreOptions / Store types from the abstract_store.d.ts module to the store.d.ts one:
  - Change namespace for moved StoreOptions (append `.Store`), but keep namespace for Store, because it is a namespace itself (as a type) in a bundle.
  - Add re-exports for moved members in a previous module (for module users) and an alias for StoreOptions only (because of the changed namespace) in aliases.d.ts (for dx.all.d.ts bundle users);
  - Re-generate integration wrappers:
    - ~~Angular now refers to a fully-specified namespace from the `devextreme/bundles/dx.all` bundle (instead of the `devextreme/data` module);~~
    - React now refers to an actual relative (moved) devextreme module / path `devextreme/data/store` (instead of the `devextreme/data/abstract_store` one).
- Define the Options type aka AbstractStoreOptions (for AbstractStore added at #23287) that extends moved StoreOptions with the onLoaded member.
  - Remove the excessive onLoading member with the same signature from ODataStoreOptions.
- Define the CustomStoreOptions.onLoaded member with the correct signature.

A proposed Docs generation result:
https://github.com/DevExpress/devextreme-documentation/pull/5875


packages/devextreme/js/data/store.d.ts
```diff
@@ -0,0 +1,209 @@
...
+    TKey = any,
+> = StoreOptions<TItem, TKey>;
+
+/**
+ * @docid
+ * @namespace DevExpress.data
+ * @deprecated Use Options instead
+ * @hidden
+ */
+export interface StoreOptions<
```
IlyaKhD: Since we're moving StoreOptions to a new location, there is no need to keep back.compat. reexports.

```suggestion
export type Options<
```

IlyaKhD: You have to add values to docid tags then **SIGH**

packages/devextreme/js/data/abstract_store.d.ts
```diff
@@ -1,219 +1,31 @@
...
+> = AbstractStoreOptions<TItem, TKey>;
 
 /**
  * @namespace DevExpress.data
  * @deprecated Use Options instead
- * @docid
- * @hidden
  */
-export interface StoreOptions<
+export interface AbstractStoreOptions<
```
IlyaKhD: We need an alias for `StoreOptions` for back-compat.
And we have no reason to add deprecated alias for the new `AbstractStoreOptions` since no one can have existing imports of it.

```suggestion
export type Options<
```

packages/devextreme/js/data/store.d.ts
mpreyskurantov: Fixed. Removed the StoreOptions interface. Appended the `StoreOptions.` prefix to a type and all its members.

packages/devextreme/js/data/abstract_store.d.ts
mpreyskurantov: Fixed (partially). Removed most of doc tags and "export" from AbstractStoreOptions, but kept it because of failure on ts discovering on a type with an anonymous member. I also suggest not to add an alias, because moved StoreOptions is more private than AbstractStoreOptions even.

alexanderPolosatov (approved):

PR #26261, opened 2023-12-13
Accessibility: Add testAccessibility helper, add configuration import util, test Tabs and TreeView

marker-dao:


alexanderPolosatov (dismissed):

iBat (dismissed):

EugeniyKiyashko:
LGTM

EugeniyKiyashko (dismissed):

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
```diff
@@ -0,0 +1,94 @@
...
+} : {};
+
+const generateConfigurations = (
+  options: Options,
+  index = 0,
+  prevConfigurations: Options[] = [],
+) => {
+  const keys = Object.keys(options);
+
+  if (index >= keys.length) {
```
ksercs: We have `index === keys.length` as a base for recursion, `index` cannot be greater

```suggestion
  if (index === keys.length) {
```

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
```diff
@@ -0,0 +1,94 @@
...
+    return [Object.assign({}, ...prevConfigurations)];
```
ksercs: Let's use ES6 syntax
```suggestion
    return [{ ...prevConfigurations }];
```

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  const key = keys[index];
+  const values = options[key];
+
+  let configurations: object[] = [];
+
+  values.forEach((value) => {
+    const currentConfigurations = [
+      ...prevConfigurations,
```
ksercs: This destructuring wastes a linear time on each iteration. I suggest u getting rid of `prevConfigurations` and just move `push` before the recursive call and `pop` after it to improve the performance

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
```diff
@@ -0,0 +1,94 @@
...
+    selector = defaultSelector,
+    a11yCheckConfig = defaultA11yCheckConfig,
+    created = defaultCreated,
+    after = defaultAfter,
+  } = configuration;
+
+  const optionConfigurations: Options[] = getOptionConfigurations(options);
+
+  optionConfigurations.forEach((optionConfiguration, index) => {
+    test(`${component}: test with axe #${index}`, async (t) => {
```
ksercs: Cannot we use stringified `optionConfiguration` here to make the name more meaningful and simplify debug?

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
```diff
@@ -0,0 +1,94 @@
...
+      { [key]: value },
+    ];
+
+    const generatedConfigurations = generateConfigurations(
+      options,
+      index + 1,
+      currentConfigurations,
+    );
+
+    configurations = configurations.concat(generatedConfigurations);
```
ksercs: `concat` wastes a linear time. Let's use `push` here
```suggestion
    configurations = configurations.push(...generatedConfigurations);
```

packages/devextreme/testing/testcafe/tests/navigation/tabs/accessibility.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+const created = async (t: any): Promise<void> => {
+  await t.pressKey('tab');
+};
+
+const a11yCheckConfig = isMaterialBased() ? {
+  // NOTE: color-contrast issues in Material
+  runOnly: isMaterial() ? '' : 'color-contrast',
```
ksercs: Why do we need this row? Seems like exception below is enough.
Comments are also duplicated now.

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
```diff
@@ -0,0 +1,91 @@
...
+    selector = defaultSelector,
+    a11yCheckConfig = defaultA11yCheckConfig,
+    created = defaultCreated,
+  } = configuration;
+
+  const optionConfigurations: Options[] = getOptionConfigurations(options);
+
+  optionConfigurations.forEach((optionConfiguration, index) => {
+    test(`${component}: test with axe #${index}`, async (t) => {
+      await a11yCheck(t, a11yCheckConfig, selector);
```
ksercs: I see that u used already written `a11yCheck` function. Let's think how we can unify the approaches and place all stuff related to a11y nearby

packages/devextreme/testing/testcafe/tests/dataGrid/accessibility/common.ts
```diff
@@ -1,5 +1,5 @@
 import { createScreenshotsComparer } from 'devextreme-screenshot-comparer';
-import { a11yCheck } from '../../../helpers/accessibilityUtils';
+import { a11yCheck } from '../../../helpers/accessibility/accessibilityUtils';
```
pomahtri:
```suggestion
import { a11yCheck } from '../../../helpers/accessibility/utils';
```

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
```diff
@@ -0,0 +1,91 @@
...
+    options,
+    selector = defaultSelector,
+    a11yCheckConfig = defaultA11yCheckConfig,
+    created = defaultCreated,
+  } = configuration;
+
+  const optionConfigurations: Options[] = getOptionConfigurations(options);
+
+  optionConfigurations.forEach((optionConfiguration, index) => {
+    test(`${component}: test with axe #${index}`, async (t) => {
```
pomahtri: it would be hard to debug some case knowing only it's id. But I can't suggest anything, it would be bad if we include all config into test name

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
```diff
@@ -0,0 +1,91 @@
+import createWidget, { WidgetName } from '../createWidget';
+import { a11yCheck, A11yCheckOptions, ElementContext } from './accessibilityUtils';
+import { isMaterialBased } from '../themeUtils';
+
+export interface Options {
+  [key: string]: any[];
+}
```
pomahtri: let's write proper typing for this

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
pomahtri, line 1: accessibility/accessibilityUtils.ts -> accessibility/utils.ts
accessibility/testAccessibility.ts -> accessibility/test.ts

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const defaultSelector = '#container';
+const defaultOptions = {};
+const defaultCreated = async () => {};
+const defaultA11yCheckConfig = isMaterialBased() ? {
+  runOnly: 'color-contrast',
+} : {};
+
+const generateConfigurations = (
```
pomahtri: let's split matrix test generation utils and accessebility test utils. It's just different functionality, we may use matrix generation for something else + i believe in grid we'll want to use accessebility checking without matrix or with complex matrixes where not all possible variants are ok

.github/workflows/testcafe_tests.yml
```diff
@@ -62,6 +62,9 @@ jobs:
       fail-fast: false
       matrix:
         ARGS: [
+          { componentFolder: "accessibility", name: "accessibility" },
```
pomahtri, line 65: why we adding new folder 'accessebility'? It's strange for me to have ci configuration like "componentFolder: accessebility" since it's not a component ⁠

Why not just navigation/treeView/accessebility as it was before?

packages/devextreme/testing/testcafe/tests/accessibility/tabs.ts
```diff
@@ -0,0 +1,57 @@
+import url from '../../helpers/getPageUrl';
+import { clearTestPage } from '../../helpers/clearPage';
+import { testAccessibility, Options, Configuration } from '../../helpers/accessibility/testAccessibility';
+import { Item } from '../../../../js/ui/tabs.d';
```
pomahtri: minor
```suggestion
import { Item } from '../../../../js/ui/tabs';
```

packages/devextreme/testing/testcafe/tests/accessibility/tabs.ts
```diff
@@ -0,0 +1,57 @@
...
+const items = [
+  { text: 'John Heart' },
+  { text: 'Marina Thomas', disabled: true },
+  { text: 'Robert Reagan' },
+  { text: 'Greta Sims' },
+  { text: 'Olivia Peyton' },
+  { text: 'Ed Holmes' },
+  { text: 'Wally Hobbs' },
+  { text: 'Brad Jameson' },
+] as Item[];
```
pomahtri: minor
```suggestion
const items: Item[] = [
  { text: 'John Heart' },
  { text: 'Marina Thomas', disabled: true },
  { text: 'Robert Reagan' },
  { text: 'Greta Sims' },
  { text: 'Olivia Peyton' },
  { text: 'Ed Holmes' },
  { text: 'Wally Hobbs' },
  { text: 'Brad Jameson' },
];
```

packages/devextreme/testing/testcafe/tests/accessibility/tabs.ts
```diff
@@ -0,0 +1,57 @@
...
+  selectOnFocus: [true, false],
+  showNavButtons: [true, false],
+  width: [450],
+  height: [250],
+  itemHoldTimeout: [5000],
+  useInkRipple: [false],
+};
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+const created = async (t: any): Promise<void> => {
```
pomahtri: minor
```suggestion
const created = async (t: TestController): Promise<void> => {
```

packages/devextreme/testing/testcafe/helpers/accessibility/accessibilityUtils.ts
```diff
@@ -2,16 +2,18 @@
 import { axeCheck, createReport } from '@testcafe-community/axe';
 import { ElementContext, RunOptions } from 'axe-core';
 
+export { ElementContext };
```
pomahtri: let's remove it and use only import from `axe-core`?

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
marker-dao: Fixed

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
marker-dao: These expressions are not equivalent. prevConfigurations contains an array. If you use Spread syntax, you will get an object with keys 0, 1, 2 - because an array, as an object, has such keys.

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
marker-dao: I think that is not applicable in the conditions of this decision.

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
marker-dao: `Cannot we use stringified optionConfiguration`
No, we are thinking about how to improve that point. The problem is the limited output size.

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
marker-dao: Strange to me because we remove concat but add spread + push.

packages/devextreme/testing/testcafe/tests/navigation/tabs/accessibility.ts
marker-dao: `Seems like exception below is enough.`
No, the exception below is not enough. The tests are still run as specified in runOnly.

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
marker-dao: I moved both utilities to the same place.

packages/devextreme/testing/testcafe/tests/dataGrid/accessibility/common.ts
marker-dao: Fixed

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
marker-dao: Yes, we are thinking about how to improve that point. The problem is the limited output size.

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
marker-dao: We don't know how many options the developer will pass and what kind of options they will be. They may differ from component to component. But they should be in array format for convenience and in order not to complicate the algorithm of configuration generation.

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
marker-dao: Done

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
marker-dao: Let's discuss it. I separated it (moved to helpers/generateOptionMatrix.ts). But:

`i believe in grid we'll want to use accessibility checking without matrix or ...`
I get it. In this case you don't need this functionality, you need to use a11yCheck from util.ts. But we want to be able to call just one function, passing possible options to it.

.github/workflows/testcafe_tests.yml
marker-dao: Because now we want to contain accessibility tests in one place. And see the reports in one place. That's a separate kind of test.

Maybe we should add a new folder with screenshot tests next to the accessibility folder. What do you think about that?

packages/devextreme/testing/testcafe/tests/accessibility/tabs.ts
marker-dao: Fixed

packages/devextreme/testing/testcafe/tests/accessibility/tabs.ts
marker-dao: Fixed

packages/devextreme/testing/testcafe/tests/accessibility/tabs.ts
marker-dao: Done

packages/devextreme/testing/testcafe/helpers/accessibility/accessibilityUtils.ts
marker-dao: Fixed. Thanks for your review)

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
pomahtri: It's still possible to write typings using generics

packages/devextreme/testing/testcafe/tests/accessibility/tabs.ts
pomahtri: No it's not :D

pomahtri: Oh, but i got why you wrote this
eslint analyze only .js file and can't find proper type in .d.ts. I think we need to remove this rule, let's do it separately

packages/devextreme/testing/testcafe/helpers/accessibility/testAccessibility.ts
pomahtri: https://github.com/marker-dao/DevExtreme/pull/2/files

pomahtri:
my solution for generateOptions utils: (it's not review, I just couldn't stand to try do it my way, algo tasks are so rare :D )

```js
function* generateCombinations(counts) {
    const currentIndicies = counts.map(() => 0);
    const total = counts.reduce((a,b) => a*b, 1);

    for (let i=0; i<total; i++) {
        yield [...currentIndicies];
        
        currentIndicies[currentIndicies.length-1]++;
        for (let j=currentIndicies.length-1; j>0; j--) {
            if (currentIndicies[j] === counts[j]) {
                currentIndicies[j] = 0;
                currentIndicies[j-1]++;
            }
        }
    }
}
console.log(...generateCombinations([3,2]))
// [0, 0] [0, 1] [1, 0] [1, 1] [2, 0] [2, 1]

function* generateOptions(matrix) {
    const entries = Object.entries(matrix);
    const counts = entries.map(([k,v]) => v.length);

    for (let currentIndicies of generateCombinations(counts)) {
        const currentObject = {}
        entries.forEach(([k, v], i) => {
            currentObject[k] = v[currentIndicies[i]];
        })
        yield currentObject;
    }
}

console.log(...generateOptions({
    a: ['a', 'b'],
    b: ['c', 'd']
}));
// {a: 'a', b: 'c'} {a: 'a', b: 'd'} {a: 'b', b: 'c'} {a: 'b', b: 'd'}

**Comment by marker-dao** [comment]
> my solution for generateOptions utils: (it's not review, I just couldn't stand to try do it my way, algo tasks are so rare :D )
> 
> ```js
> function* generateCombinations(counts) {
> ```

Love generators :D I'll check back a little later)

File: packages/devextreme/testing/testcafe/helpers/testAccessibility.ts (hunk 5 above)
Comment (ksercs) [reply packages/devextreme/testing/testcafe/helpers/testAccessibility.ts]: spread + push iterates only through `generatedConfigurations` while `concat` recreates the whole array

File: packages/devextreme/testing/testcafe/tests/navigation/tabs/accessibility.ts (hunk 6 above)
Comment (ksercs) [reply packages/devextreme/testing/testcafe/tests/navigation/tabs/accessibility.ts]: Ok, but i still suggest removing of the duplicated comment :)

File: packages/devextreme/testing/testcafe/helpers/testAccessibility.ts (hunk 2 above)
Comment (ksercs) [reply packages/devextreme/testing/testcafe/helpers/testAccessibility.ts]: Oh, sure, I see

File: packages/devextreme/testing/testcafe/tests/navigation/tabs/accessibility.ts (hunk 6 above)
Comment (marker-dao) [reply packages/devextreme/testing/testcafe/tests/navigation/tabs/accessibility.ts]: Done

File: packages/devextreme/testing/testcafe/helpers/testAccessibility.ts (hunk 5 above)
Comment (marker-dao) [reply packages/devextreme/testing/testcafe/helpers/testAccessibility.ts]: Done

File: packages/devextreme/testing/testcafe/helpers/testAccessibility.ts (hunk 3 above)
Comment (ksercs) [reply packages/devextreme/testing/testcafe/helpers/testAccessibility.ts]: I don't see a reason why it's not applicable here. Did you try?
I mean smth like this:
```javascript
            const generateOptionMatrix = (
                options,
                index = 0,
                currentConfiguration = [],
            ) => {
                const keys = Object.keys(options);

                if (index === keys.length) {
                    return [Object.assign({}, ...currentConfiguration)];
                }

                const key = keys[index];
                const values = options[key];

                let configurationsMatrix = [];

                values.forEach((value) => {
                    currentConfiguration.push({ [key]: value });

                    const generatedConfigurations = generateOptionMatrix(
                        options,
                        index + 1,
                        currentConfiguration,
                    );
                    configurationsMatrix.push(...generatedConfigurations);

                    currentConfiguration.pop();                    
                });

                return configurationsMatrix;
            };
```

marker-dao: Okay, I get it and refactored the algorithm: [80fbb67bad49316ad89677903ec3140293ba36d0].

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
marker-dao: I have added a configuration only into failed tests: [82bb1f0fe6ff2db4976759db34ff5b7c8f3ec7ca], [b7de51ed8ff4c3e8b6c4cf04fe30e22dd06c76a2].

pomahtri (approved):

iBat (approved):

packages/devextreme/testing/testcafe/helpers/testAccessibility.ts
ksercs: I suggest u also minify items & dataSource values somehow to make the report more readable

ksercs (approved):

PR #26279, opened 2023-12-15
Calendar: it should not be possible to focus dates that are disabled using combination of disabledDates+min/max

Zedwag:


packages/devextreme/js/ui/calendar/ui.calendar.js
```diff
@@ -344,13 +344,29 @@ const Calendar = Editor.inherit({
...
-            this._waitRenderView(offset > 0 ? 1 : -1);
+            const direction = offset > 0 ? 1 : -1;
+            const isViewDisabled = direction === 1 ? this._isNextViewDisabled() : this._isPrevViewDisabled();
+
+            if(!isViewDisabled) {
+                this._waitRenderView(direction);
+            } else {
+                this._moveToClosestAvailableDate(currentDate);
+            }
+
```
EugeniyKiyashko, line 355: Let's remove excess line

EugeniyKiyashko (approved):

packages/devextreme/js/ui/calendar/ui.calendar.js
Zedwag: will do in my next pr

PR #26292, opened 2023-12-15
Refactor build scripts

IlyaKhD:


.github/workflows/packages_publishing.yml
```diff
@@ -50,17 +50,17 @@ jobs:
...
+        run: npm run ts-node build/set-timestamp-version
 
       - name: Build npm packages
         env:
           BUILD_INTERNAL_PACKAGE: true
         run: npm run all:build
 
       - name: Build artifacts package
-        run: npm run ts-script build/make-artifacts-package
+        run: npm run ts-node build/make-artifacts-package
```
pomahtri: Why not just `npx ts-node ...`? The behavior should be the same, npx calls bin packages from local node_modules if they exist

IlyaKhD: You're right, no more need in script (I used it in local experiments)

Fixed, thanks!

pomahtri: Remove it from package.json `scripts` field too please

IlyaKhD: done

alexslavr (approved):

PR #26311, opened 2023-12-19
Scheduler: Fix background color of cells for hover, focused and active states

Alyar666:


Raushen (dismissed):

packages/devextreme/scss/widgets/fluent/scheduler/_colors.scss
```diff
@@ -109,6 +112,11 @@ $header-panel-cell-date: null !default;
   $scheduler-appointment-text-color: #fff !default;
   $scheduler-panel-text-color: #999;
   $header-panel-cell-date: #adadad;
+
+  @if $color == "saas" {
+    $scheduler-workspace-focused-cell-color: darken(desaturate(adjust-hue($base-accent, -7), 65.00), 41.18);
```
EugeniyKiyashko: If we need override variable value, we should declare variable
```$scheduler-workspace-focused-cell-color: null !default; ``` 
and after that for each mode and color set required value.

File: packages/devextreme/scss/widgets/fluent/scheduler/_colors.scss [hunk 2]
```diff
@@ -109,6 +112,11 @@ $header-panel-cell-date: null !default;
   $scheduler-appointment-text-color: #fff !default;
   $scheduler-panel-text-color: #999;
```
EugeniyKiyashko, line 116: `!default` is missed

EugeniyKiyashko (approved):

PR #26407, opened 2024-01-05
Scheduler - Appointment tooltip is incorrectly displayed on the first click when adaptivity is enabled in React (T1187586)

pomahtri:


packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
```diff
@@ -139,15 +141,25 @@ export class TooltipStrategyBase {
...
-        targetedAppointmentData,
-        isButtonClicked,
-      } : appointmentData,
-      container: options.container,
-      index,
-    }));
+    return new FunctionTemplate((options) => {
+      // eslint-disable-next-line @typescript-eslint/init-declarations
+      let resolve: () => void;
+      const promise = new Promise<void>((r) => { resolve = r; });
```
williamvinogradov: Question.

Can we use the ```deferred``` here?
Or if we want to remove dependency from our custom ```deferred```, let's create new ```deferred-like``` helper object in ```__internal/core``` :)

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
```diff
@@ -14,6 +14,8 @@ const TOOLTIP_APPOINTMENT_ITEM_DELETE_BUTTON_CONTAINER = `${TOOLTIP_APPOINTMENT_
 const TOOLTIP_APPOINTMENT_ITEM_DELETE_BUTTON = `${TOOLTIP_APPOINTMENT_ITEM}-delete-button`;
 
 export class TooltipStrategyBase {
+  protected asyncTemplateDeferredList = new Set<Promise<void>>();
```
williamvinogradov: And let's rename this variable because it's not a list and it's not a deferred list :)
E.g
```suggestion
  protected asyncTemplateRenderedArray: Promise<void>[] = [];
```

Or for set:
```suggestion
  protected asyncTemplateRenderedSet = new Set<Promise<void>>();
```

williamvinogradov: Minor comment.
Why do we use ```Set``` here? :)
I didn't see any searching of the promises, can we use an array here instead?

I see the deletion of the promise from set by its ref, but we can do the following with the array easily:
```typescript
this.asyncTemplateRenderedArray = this.asyncTemplateRenderedArray.filter((item) => item !== promise);
```

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_mobile_tooltip_strategy.ts
```diff
@@ -65,13 +65,27 @@ export class MobileTooltipStrategy extends TooltipStrategyBase {
     return false;
   }
 
-  _onShowing() {
+  private setTooltipConfig(): void {
     const isTabletWidth = getWidth(getWindow()) > 700;
```
williamvinogradov, line 72: Minor comment for refactoring.

Let's move all constants related to sizes to one general place.
It may be: ```__internal/scheduler/device_const.ts``` or something like

packages/devextreme/testing/testcafe/tests/scheduler/tooltipBehaviour/tooltipBehavior.ts
```diff
@@ -117,3 +119,76 @@ safeSizeTest('The tooltip should hide after manually scrolling in the browser',
...
+    currentDate: new Date(2017, 4, 25),
+    currentView: 'month',
+    adaptivityEnabled: true,
+    templatesRenderAsynchronously: true,
+    integrationOptions: {
+      templates: {
+        appointmentTooltip: {
+          render(args) {
+            (window as any).deferred.done(() => {
+              console.log(args);
```
williamvinogradov: Let's remove this ```console.log``` :)

packages/devextreme/testing/testcafe/tests/scheduler/tooltipBehaviour/tooltipBehavior.ts
```diff
@@ -117,3 +119,76 @@ safeSizeTest('The tooltip should hide after manually scrolling in the browser',
...
+    (window as any).deferred.resolve();
+  })();
+
+  await takeScreenshot('tooltip-rendering-with-react.png');
+
+  await t.expect(compareResults.isValid()).ok(compareResults.errorMessages());
+}, [600, 1000]).before(async () => {
+  await ClientFunction(() => {
+    (window as any).deferred = $.Deferred();
+  })();
```
williamvinogradov: Let's clear the ```window.deferred``` on the ```after``` testcafe hook
Something like:
```typescript
.after(async () => {
  await ClientFunction(() => {
    delete (window as any).deferred;
  })();
});
```

packages/devextreme/testing/testcafe/tests/scheduler/tooltipBehaviour/tooltipBehavior.ts
```diff
@@ -117,3 +119,76 @@ safeSizeTest('The tooltip should hide after manually scrolling in the browser',
...
 });
+
+safeSizeTest('Tooltip on mobile devices should have enough hight if there are async templates (React)', async (t) => {
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const scheduler = new Scheduler('#container');
+
+  await t
+    .click(scheduler.collectors.find('7').element);
+
+  await ClientFunction(() => {
```
williamvinogradov: Minor comment.

In my opinion - it's better to move the client functions separately.
It makes code more clear & obvious.
But it's only my opinion :)

E.g
```typescript
  const resolveWindowDeferred = ClientFunction(() => { (window as any).deferred.resolve(); });
  
  test('my test case', async (t) => {
  // other test code
  
  await resolveWindowDeferred ();
  
  // other test code
  );
```

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
pomahtri: Because this filter operation in your example is the searching you're talking about
`set.remove(ref)` looks much better than `newArray = oldArray.filter(...)`
It's a best practice to use sets if we don't depend on items order

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
pomahtri: Why?
If you're worried about this ugly `new Promise(r => resolve = r)`, it will be changed after we'll have access to `Promise.withResolvers` API (we can't because our TS is old) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_mobile_tooltip_strategy.ts
pomahtri: Better to do separately cause I'm unsure about that (didn't deeped down to this configs)

packages/devextreme/testing/testcafe/tests/scheduler/tooltipBehaviour/tooltipBehavior.ts
pomahtri: It's bad to move it to file scope, better to keep it in test scope.
I'll extract this to separate variable

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_mobile_tooltip_strategy.ts
williamvinogradov: Yup, agree with that
I left this comment to share this idea + don't forget about it in the future

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
williamvinogradov: Yes, I worried about this ugly ```new Promise(r => resolve = r)``` and that its a custom construction to manage the promise.
In this place, you implement it this way, in other places other developers can implement it another way, and so on.
It's better to use one approach across the codebase.

I see two solutions here:

1. Use our ```deferred``` and replace it sometime in the future as a separate refactoring task
(now it's not in our plans, so in other words just use ```deferred``` for now)
Because we now use it everywhere :)
2. Create ```newDeferred``` object somewhere in the ```__internal/core``` and use it here.
I think it will be better to do it in a separate PR.
And start slowly migrate from our old ```deferred``` to ```newDeferred``` :)

packages/devextreme/testing/testcafe/tests/scheduler/tooltipBehaviour/tooltipBehavior.ts
williamvinogradov: I usually move it to file scope because these functions can be used in many test cases and one function can be used in test case, before and after hooks :)

So, I don't think that it's bad.

pomahtri: This very function is intended to be used only in this test
Plus this is a WA, we'll remove all this deferred once we can write React testcafe tests

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
pomahtri: newDeferred is no better than deferred
old Deferred is bad
This ugly construction is well-known and is used in many projects (for examle, check "Existing implementations" in https://github.com/tc39/proposal-promise-with-resolvers?tab=readme-ov-file#existing-implementations)
Moving to new API (.withResolvers) will be easier if we have Promise than if we would have Deferred (because in the future we'll change only this part of code, keeping previous typing)

williamvinogradov: Ok, got you
Maybe we should add the next utils function in the future to make this less ugly?
Something like this (and maybe we should add a "reject" too):
```typescript
export const createHandledPromise= <T>() => {
  let resolve: (value: T) => void;
  const promise = new Promise<T>((r) => resolve);
  return [resolve, promise];
}
```

packages/devextreme/js/__internal/core/utils/promise.ts
```diff
@@ -0,0 +1,22 @@
...
+  reject: (value: T) => void;
+} {
+  let resolve!: (value: T) => void;
+  let reject!: (value: T) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+
+  return { promise, resolve, reject };
```
williamvinogradov, line 21: Minor comment.

In my opinion, it's better to return the array instead of the object
It's very similar in functionality, but let's compare these two blocks of code
Array:
```typescript
const [myPromise1, myResolve1, myReject1] = createPromise<void>();
const [myPromise2, myResolve2, myReject2] = createPromise<void>();
```
Object:
```typescript
const { promise: myPromise1, resolve: myResolve1, reject: myReject1 } = createPromise<void>();
const { promise: myPromise2, resolve: myResolve2, reject: myReject2 } = createPromise<void> ();
```

But objects are more comfortable when we want to deconstruct only one property:
Array:
```typescript
const [,  , myReject1] = createPromise<void>();
const [,  , myReject2] = createPromise<void>();
```
Object:
```typescript
const { reject: myReject1 } = createPromise<void>();
const { reject: myReject2 } = createPromise<void> ();
```

But in my opinion arrays still better (but it's just my opinion)

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_mobile_tooltip_strategy.ts
```diff
@@ -65,13 +65,27 @@ export class MobileTooltipStrategy extends TooltipStrategyBase {
...
 
-    this._tooltip.option('height', MAX_HEIGHT.DEFAULT);
-    const listHeight = getOuterHeight(this._list.$element());
+    const listHeight = getOuterHeight(this._list.$element().find('.dx-scrollable-content'));
```
williamvinogradov: It's better to create the separate ```const``` variable for CSS selectors:
```suggestion
    const listHeight = getOuterHeight(this._list.$element().find(CLASS.scrollableContent));
```

packages/devextreme/testing/testcafe/tests/scheduler/tooltipBehaviour/tooltipBehavior.ts
```diff
@@ -117,3 +119,81 @@ safeSizeTest('The tooltip should hide after manually scrolling in the browser',
     adaptivityEnabled,
   }));
 });
+
+safeSizeTest('Tooltip on mobile devices should have enough hight if there are async templates (React)', async (t) => {
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const scheduler = new Scheduler('#container');
+
+  await t
+    .click(scheduler.collectors.find('7').element);
```
williamvinogradov: Let's add one more scenario with a second click on the appointment.

I mean the following test case
- Click on the appointment
- Close tooltip
- Click on the appointment again
- Take a screenshot

We can extend this test case instead of creating a new one
(hide tooltip, one more click, and one more screenshot)

**Why:**
Because of this case, we have this change:
```typescript
const listHeight = getOuterHeight(this._list.$element().find('.dx-scrollable-content'));
``` 
(And this PR failed the manual testing phase on this test case last time)

packages/devextreme/js/__internal/core/utils/promise.ts
pomahtri: I decided to use objects so that my helper had similar API as future native `Promise.withResolvers`

williamvinogradov (approved):

PR #26428, opened 2024-01-12
Set up playgrounds

markallenramirez:


playgrounds/angular/src/app/app.component.ts
```diff
@@ -30,6 +29,7 @@ import {
 import {
   DxPopoverComponent
 } from 'devextreme-angular';
+import ArrayStore from 'devextreme/artifacts/npm/devextreme/data/array_store';
```
timbset: This import looks strange. Can't we replace it with other one?

playgrounds/angular/tsconfig.spec.json
```diff
@@ -0,0 +1,18 @@
+/* To learn more about this file see: https://angular.io/config/tsconfig. */
```
timbset: Pure JSON does not support comments. It's better to remove it to be sure that everything works ok

playgrounds/react/examples/Examples.tsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
```
timbset: Isn't it the app.tsx file which is moved and renamed? If so, please use git move

playgrounds/react/examples/list-example.tsx
```diff
@@ -15,8 +15,11 @@ interface IListItemProps {
 }
 
 const items: IListItemProps[] = [
+  // @ts-expect-error
```
timbset: Why so?

playgrounds/react/examples/popup-example.tsx
```diff
@@ -21,7 +21,7 @@ export default class extends React.Component<any, { visible: boolean; text: stri
     this.handleTextUpdate = this.handleTextUpdate.bind(this);
   }
 
-  private handleTextUpdate(e: unknown) {
+  private handleTextUpdate(e: any) {
```
timbset: Don't we have a better option to avoid `any`?

playgrounds/vue/components/examples.vue
```diff
@@ -0,0 +1,69 @@
+<template>
```
timbset: Looks like it's app.vue which is moved and renamed. If so, please use git move

timbset (changes requested):

playgrounds/angular/src/app/app.component.ts
markallenramirez: I agree. Fixed.

playgrounds/angular/tsconfig.spec.json
markallenramirez: Removed.

playgrounds/react/examples/list-example.tsx
markallenramirez: Looks like the original author just reused the existing interface instead of creating another one. In any case, fixed. I also this comment from similar instances.

playgrounds/react/examples/popup-example.tsx
markallenramirez: The examples in the playground are ancient (pre-TS improvements). I proposed to remove them to simplify the playground, but the team suggested to leave it be for now just in case. Anyway, the examples need a reboot in my opinion, but I removed all 'any' types that I can for now.

playgrounds/angular/src/app/app.component.ts
```diff
@@ -100,7 +100,7 @@ import {
 export class AppComponent implements OnInit, AfterViewInit {
   @ViewChild(DxPopoverComponent) popover: DxPopoverComponent;
   
-  value: any = [
+  value: Array<Date|Number|String> = [
```
timbset: Why so? You assign only Date values. Is there any other reason why use union of types instead of `Array<Date>`?

playgrounds/react/App.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import Examples from './examples/Examples';
+
+const App = () => (
+    <Examples />
+);
+
+export default App;
```
timbset: Please do not forget a new line at the end of every file

timbset (changes requested):

playgrounds/react/examples/Examples.tsx
markallenramirez: Hi, I modified the file (moved all examples in a separate folder) and renamed it.

playgrounds/vue/components/examples.vue
markallenramirez: Hi, same goes here I modified the file (moved all examples in a separate folder) and renamed it.

playgrounds/angular/src/app/app.component.ts
markallenramirez: This variable is used by the Calender.value option: https://js.devexpress.com/Angular/Documentation/ApiReference/UI_Components/dxCalendar/Configuration/#value

The type should be
` Date | Number | String | Array<Date | Number | String>`

but I just used an array of union types just in case someone that would use the example would like to use string or number values.

timbset (dismissed):

pomahtri (dismissed):

playgrounds/README.md
```diff
@@ -0,0 +1,58 @@
+# DevExtreme Playgrounds
+
+These playgrounds are intended for testing purposes.
+
+To run a playground, clone this repository and follow the instructions below.
+
+## Prepare the devextreme package
+
+Before anything else, you need to install modules and build the devextreme package. Run the following scripts from the root directory of this repository:
```
vladaskorohodova:
```suggestion
Before you run a playground, install modules and build the devextreme package. Run the following scripts from the root directory of this repository:
```

playgrounds/README.md
```diff
@@ -0,0 +1,58 @@
...
+```
+npm run pack -w=devextreme-vue
+```
+2. Run the launch script:
+```
+cd playgrounds/vue
+npm run start
+```
+
+## Updating devextreme code
```
vladaskorohodova:
```suggestion
## Update DevExtreme code
```

playgrounds/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+To see changes applied to devextreme code in playgrounds, you need to re-build the devextreme package.
```
vladaskorohodova:
```suggestion
To observe the changes made to the DevExtreme code within playgrounds, rebuild the `devextreme` package.
```

playgrounds/README.md
```diff
@@ -0,0 +1,58 @@
...
+```
+npm run build -w=devextreme-main
+```
+If you want hot update, apply changes to the following directory:
```
vladaskorohodova:
```suggestion
If you need your changes to apply dynamically without reloading the playground, save them in the following directory:
```

williamvinogradov (approved):

timbset (approved):

PR #26432, opened 2024-01-12
DataGrid - The "aria-label" attribute value is "Show filter options for column 'undefined'" for the header filter icon in grouped fields (T1205784)

Raushen:
See https://supportcenter.devexpress.com/ticket/details/t1205784/datagrid-the-aria-label-attribute-value-is-show-filter-options-for-column-undefined-for

### DIFF
--- a/packages/devextreme/testing/testcafe/model/dataGrid/groupPanelItem.ts
+++ b/packages/devextreme/testing/testcafe/model/dataGrid/groupPanelItem.ts
@@ -1,0 +1,13 @@
+const FILTER_BUTTON_SELECTOR = '.dx-header-filter';
+
+export class GroupPanelItem {
+  element: Selector;
+
+  constructor(private readonly selector: Selector) {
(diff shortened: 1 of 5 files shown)


packages/devextreme/testing/testcafe/model/dataGrid/groupPanelItem.ts
```diff
@@ -0,0 +1,13 @@
+const FILTER_BUTTON_SELECTOR = '.dx-header-filter';
```
williamvinogradov: Let's change this const to the ```CLASS``` const
To make it uniform with other our test models: 
[link to the example](https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme/testing/testcafe/model/dataGrid/index.ts#L24)

williamvinogradov (dismissed):

williamvinogradov (approved):

PR #26442, opened 2024-01-15
Scheduler - When virtual scrolling is enabled, a template's targetedAppointmentData does not display the correct resource id (T1205120)

timbset:


marker-dao (dismissed):

packages/devextreme/testing/testcafe/tests/scheduler/layout/templates/appointmentTemplate.ts
```diff
@@ -66,3 +76,274 @@ fixture.disablePageReloads`Layout:Templates:appointmentTemplate`
     });
   });
 });
+
```
williamvinogradov: Question.

Maybe we should create a separate fixture/file in the same folder for this matrix test?
Because it has a lot of initialization and variables in the module/file scope.

packages/devextreme/testing/testcafe/tests/scheduler/layout/templates/appointmentTemplate.ts
```diff
@@ -66,3 +76,274 @@ fixture.disablePageReloads`Layout:Templates:appointmentTemplate`
...
+  scrollMode,
+  rtlEnabled,
+}) => {
+  const currentDate = new Date(2024, 0, 2);
+  const HOUR = 1000 * 60 * 60;
+
+  const startDayHour = 8;
+  const endDayHour = 20;
+
+  let resourceCount = 30;
```
williamvinogradov: Let's move the logic of setting the ```resourceCount``` value into a separate function.
It makes the code more linear and obvious.
Now, it takes time to see all cases os setting ```resourceCount``` variable :)

E.g.
```typescript
  // some variables
 const getResourceCount = (
   viewType: ViewType,
   scrollMode: ScrollMode,
   orientation: Orientation,
 ) => {
    switch (true) {
      case viewType == 'week' && scrollMode == 'virtual' && orientation == 'horizontal':
        return 10;
      // other cases
     default:
       return 30;
  }
  
  // other code
  
  testOptions.forEach({...}) => {
  // other code
  
  const resourceCount = getResourceCount(viewType, scrollMode, orientation);
  
  // other code
  }
```

packages/devextreme/testing/testcafe/tests/scheduler/layout/templates/appointmentTemplate.ts
```diff
@@ -66,3 +76,274 @@ fixture.disablePageReloads`Layout:Templates:appointmentTemplate`
...
+
+  let appointmentDates: Date[];
```
williamvinogradov: Same comment as for ```resourceCount```
Let's move the logic of setting the ```appointmentDates``` into a separate function

E.g.
```typescript
  // some variables
 const getAppointmentDates= (
   viewType: ViewType,
 ) => {
    switch (viewType) {
      case 'agenda':
        return ...;
      default:
        return ...;
  }
  
  // other code
  
  testOptions.forEach({...}) => {
  // other code
  
  const appointmentDates = getAppointmentDates(viewType);
  
  // other code
  }
```

packages/devextreme/testing/testcafe/tests/scheduler/layout/templates/appointmentTemplate.ts
```diff
@@ -66,3 +76,274 @@ fixture.disablePageReloads`Layout:Templates:appointmentTemplate`
...
+      appointmentDates
+        .map<Appointment>((date) => ({
+        text: resource.text,
+        startDate: date,
+        endDate: new Date(date.getTime() + HOUR / 2),
+        groupId: resource.id,
+      })),
+    ), []);
+
+    await createWidget('dxScheduler', {
```
williamvinogradov: Why do you initialize the widget in the test body?
We usually initialize widgets in the testcafe's ```before``` hook

If you have arguments that are better than ```before``` hook let's discuss it
If not, let's move the widget initialization to the ```before``` to make it consistent with other test cases :)

packages/devextreme/testing/testcafe/tests/scheduler/layout/templates/appointmentTemplate.ts
```diff
@@ -66,3 +76,274 @@ fixture.disablePageReloads`Layout:Templates:appointmentTemplate`
...
+    appointmentDates[appointmentDates.length - 1],
+  ];
+
+  const groupsToCheck = [
+    { groupId: 0 },
+    { groupId: Math.floor(resourceCount / 3) },
+    { groupId: resourceCount - 1 },
+  ];
+
+  test(`targetedAppointmentData should be correct with groups (viewType="${viewType}", groupOrientation="${groupOrientation}", scrollMode="${scrollMode}", rtlEnabled="${rtlEnabled}") (T1205120)`, async (t) => {
```
williamvinogradov: Optional.

Let's add a ```NOTE``` comment to this test, that it's a success if no errors in the template occurred.
It may take time to understand what going on here because this test doesn't have any ```assert/expect``` calls

packages/devextreme/testing/testcafe/tests/scheduler/layout/templates/appointmentTemplate.ts
```diff
@@ -66,3 +76,274 @@ fixture.disablePageReloads`Layout:Templates:appointmentTemplate`
...
+        date: datesToCheck,
+        group: groupsToCheck,
+      });
+
+      // eslint-disable-next-line no-restricted-syntax
+      for (const { date, group } of scrollOptions) {
+        await scrollTo(date, group);
+        await t.wait(50);
+      }
+    }
```
williamvinogradov: Let's reduce complexity of this code block and invert this ```if``` statement:

```suggestion
    if (scrollMode !== 'virtual') {
      return;
    }
    
    const scrollOptions = generateOptionMatrix({
        date: datesToCheck,
        group: groupsToCheck,
      });

      // eslint-disable-next-line no-restricted-syntax
      for (const { date, group } of scrollOptions) {
        await scrollTo(date, group);
        await t.wait(50);
      }
```

packages/devextreme/testing/testcafe/tests/scheduler/layout/templates/appointmentTemplate.ts
timbset: I put the 'createWidget' function into a test body, not into the 'before' hook, because it can throw an error on initial render, so it's a part of test logic

williamvinogradov (approved):

PR #26443, opened 2024-01-15
Scheduler(T1206036): Fix multi cell selection + offset issue.

williamvinogradov:
- Re-implemented cell selection algorithm in view_provider.
- Increased the selection e2e test matrix.

See this [T1206036](https://supportcenter.devexpress.com/ticket/details/t1206036/scheduler-cell-selection-in-the-week-view-does-not-work-as-expected-with-an-enabled-offset) ticket for more details


packages/devextreme/testing/testcafe/tests/scheduler/viewOffset/multiCellSelection.ts
```diff
@@ -8,88 +9,96 @@ fixture.disablePageReloads`Offset: Multi cell selection`
...
+      { view: 'week', dragOptions: { direction: 'decrease_2', from: [6, 5], to: [6, 3] } },
+      { view: 'week', dragOptions: { direction: 'increase_3', from: [0, 0], to: [11, 6] } },
+      { view: 'week', dragOptions: { direction: 'decrease_3', from: [11, 6], to: [0, 0] } },
+      { view: 'month', dragOptions: { direction: 'increase', from: [2, 2], to: [3, 1] } },
+      { view: 'month', dragOptions: { direction: 'decrease', from: [3, 1], to: [2, 2] } },
+      { view: 'timelineDay', dragOptions: { direction: 'increase', from: [0, 0], to: [0, 5] } },
+      { view: 'timelineDay', dragOptions: { direction: 'decrease', from: [0, 5], to: [0, 0] } },
+      { view: 'timelineMonth', dragOptions: { direction: 'increase', from: [0, 0], to: [0, 3] } },
+      { view: 'timelineMonth', dragOptions: { direction: 'decrease', from: [0, 3], to: [0, 0] } },
+    ].forEach(({ view, dragOptions }) => {
```
timbset, line 42: I would recommend to use matrix option generator instead of nested loops. They are hard to read

timbset (approved):

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
```diff
@@ -29,6 +29,7 @@ import pointerEvents from '@js/events/pointer';
 import { addNamespace, isMouseEvent } from '@js/events/utils/index';
 import messageLocalization from '@js/localization/message';
 import { getMemoizeScrollTo } from '@js/renovation/ui/common/utils/scroll/getMemoizeScrollTo';
+import { ViewType } from '@js/renovation/ui/scheduler/types';
```
andrewmakarov, line 32: Why do you use import from renovation code? If you want using typing, may be take types from _packages/devextreme/js/ui/scheduler.d.ts_?

packages/devextreme/testing/testcafe/tests/scheduler/viewOffset/multiCellSelection.ts
```diff
@@ -8,88 +9,96 @@ fixture.disablePageReloads`Offset: Multi cell selection`
 
 const SCHEDULER_SELECTOR = '#container';
 
+// TODO Vinogradov: Change it to a general dispose widget helper function.
+const disposeScheduler = ClientFunction(() => {
```
andrewmakarov, line 13: Let create hook or function **dispose**? Targeted selector have equal name in all tests and all components. Create enum with components(like dxScheduler, dxGrid) and pass to function

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
williamvinogradov: I used the type from the renovation (R1) because the code that used this type passed a variable to function to the renovated (R1) code.

packages/devextreme/testing/testcafe/tests/scheduler/viewOffset/multiCellSelection.ts
williamvinogradov: I agree with you
But let's create this ```dispose``` helper in a separate task :)

packages/devextreme/testing/testcafe/tests/scheduler/viewOffset/multiCellSelection.ts
williamvinogradov: Yup, agree with you.

But it looks like the option matrix generator will be overkill because there are only 3 arrays of options :)
I think the option generator is good when we have 5 or more options or want to exclude some cases.

williamvinogradov: Agreed that we will discuss it with the whole team and decide when we use the options matrix generator :)

PR #26480, opened 2024-01-23
Add script for updating failed screenshots

alexanderPolosatov:


packages/devextreme/testing/testcafe/update_failed_etalons.mjs
```diff
@@ -0,0 +1,91 @@
...
+      result = false;
+    }
+  });
+
+  return result;
+}
+
+function checkFilesAfterCopy(allEtalons) {
+  allEtalons.forEach((f) => {
+    console.log(`file ${f} not copyed`);
```
ksercs:
```suggestion
    console.log(`File "${f}" is not copied`);
```

packages/devextreme/testing/testcafe/update_failed_etalons.mjs
```diff
@@ -0,0 +1,91 @@
...
+const PATH_TO_CURRENT_ETALONS = argv[2];
+const PATH_TO_TESTCAFE_TESTS = './testing/testcafe/tests';
+const NEW_ETALON_FILE_NAME_PATTERN = /^(?!.*(?:_mask|_diff|_etalon)\.png$).*\.png$/i;
+const ETALONS_FOLDER_NAME = 'etalons';
+
+const getAllEtalonsNames = () => {
+  const allFileNames = readdirSync(PATH_TO_CURRENT_ETALONS);
+  const newEtalons = allFileNames.filter((name) => NEW_ETALON_FILE_NAME_PATTERN.test(name));
+
+  return newEtalons;
```
ksercs, line 22: I suggest u creating a map here instead of a plain array for an access with O(1) time.
It will speed up processing because we don't need to make linear operations like `splice` and `indexOf`

packages/devextreme/testing/testcafe/update_failed_etalons.mjs
```diff
@@ -0,0 +1,91 @@
...
+function getAllDirs(path) {
+  return readdirSync(path)
+    .map((fileName) => join(path, fileName))
+    .filter(isDirectory);
+}
+
+function processEtalonFolder(etalonFolderPath, allEtalons) {
+  const allFiles = readdirSync(etalonFolderPath);
+
+  return allFiles.some((currentFileName) => {
```
ksercs: `some` is basically used to check if we have an item passing the condition. It seems to be really hacky for me to use it for simple iteration.
I suggest u using of a usual forEach/for with an additional check inside.
Like:
```javascript
allFiles.forEach(file => {
    if (newEtalonsCount === 0) {
        return;
    }
    ...
});
```
Same for `processFolder`

packages/devextreme/testing/testcafe/update_failed_etalons.mjs
alexanderPolosatov: fixed

packages/devextreme/testing/testcafe/update_failed_etalons.mjs
alexanderPolosatov: fixed

packages/devextreme/testing/testcafe/update_failed_etalons.mjs
```diff
@@ -0,0 +1,113 @@
...
+    allEtalonsMap.forEach((value, key) => {
+      console.log(`File "${key}" is not copied`);
+    });
+  }
+}
+
+(() => {
+  if (foldersWithScreenshotsExist()) {
+    const allEtalons = getAllEtalonsNames();
+    console.log(`${allEtalons.size} screenshots to moving were found`);
```
ksercs:
```suggestion
    console.log(`${allEtalons.size} new screenshots are found`);
```

packages/devextreme/testing/testcafe/update_failed_etalons.mjs
```diff
@@ -0,0 +1,113 @@
...
+    .map((fileName) => join(path, fileName))
+    .filter(isDirectory);
+}
+
+function processEtalonFolder(etalonFolderPath, allEtalons) {
+  const allFiles = readdirSync(etalonFolderPath);
+  let hasEtalonsToMove = allEtalons.size !== 0;
+
+  allFiles.forEach((currentFileName) => {
+    if (hasEtalonsToMove) {
```
ksercs: If we use `for` here we can just return when there is no rest etalons.
```javascript
if (allEtalons.size === 0) [
    return false;
}    
````
I see following advantages of this approach:
- We will exactly stop loop keeping the time
- It'll allow us to get rid of `hasEtalonsToMove` variable
- We'll reduce nesting
- It's clearer for me to see a concrete escaping condition because u don't need to check where and why variable is updated (it's especially useful for simplifying of the `processFolder` function)

**Review by EugeniyKiyashko — DISMISSED** [review]

**Review by alexslavr — DISMISSED** [review]

File: packages/devextreme/testing/testcafe/update_failed_etalons.mjs [hunk 6]
```diff
@@ -0,0 +1,115 @@
...
+
+  for (let dir of currentDirrectoryFolders) {
+    const folderName = dir.split('/').at(-1);
+    let continueProcessFolders;
+
+    if (folderName === ETALONS_FOLDER_NAME) {
+      continueProcessFolders = processEtalonFolder(dir, allEtalons);
+    } else {
+      continueProcessFolders = processFolder(dir, allEtalons);
+    }
```
timbset:
```suggestion
    const continueProcessFolders = folderName === ETALONS_FOLDER_NAME
      ? processEtalonFolder(dir, allEtalons)
      : processFolder(dir, allEtalons);
```

packages/devextreme/testing/testcafe/update_failed_etalons.mjs
alexanderPolosatov: fixed

packages/devextreme/testing/testcafe/update_failed_etalons.mjs
alexanderPolosatov: fixed

timbset (approved):

PR #26492, opened 2024-01-25
DropDownEditors: Cover components with tests with testAccessibility

marker-dao:


packages/devextreme/testing/testcafe/tests/accessibility/autocomplete.ts
```diff
@@ -0,0 +1,64 @@
...
+import { defaultSelector, testAccessibility, Configuration } from '../../helpers/accessibility/test';
+import { Options } from '../../helpers/generateOptionMatrix';
+import { Properties } from '../../../../js/ui/autocomplete.d';
+import Autocomplete from '../../model/autocomplete';
+
+fixture.disablePageReloads`Accessibility`
+  .page(url(__dirname, '../container.html'))
+  .afterEach(async () => clearTestPage());
+
+const items = ['A', 'Aa', 'B'];
```
EugeniyKiyashko: It's better to choose more informative item's name as follows
const items = ['Item_1', 'Item_2', 'Item_3'];

marker-dao: [1d96c8aa2cf8bac45468c91f917d9e162d46d2bc]

alexanderPolosatov (dismissed):

.vscode/settings.json
```diff
@@ -3,6 +3,6 @@
          "watch": true,
     },
     "editor.codeActionsOnSave": {
-        "source.fixAll.eslint": true
+        "source.fixAll.eslint": "explicit"
```
ksercs: Is it intended? Why we do this?

.github/workflows/testcafe_tests.yml
```diff
@@ -62,9 +62,9 @@ jobs:
       fail-fast: false
       matrix:
         ARGS: [
-          { componentFolder: "accessibility", name: "accessibility" },
-          { componentFolder: "accessibility", name: "accessibility - material", theme: 'material.blue.light' },
-          { componentFolder: "accessibility", name: "accessibility - fluent", theme: 'fluent.blue.light' },
+          { componentFolder: "accessibility", name: "accessibility", concurrency: 2 },
```
ksercs: Why?

.vscode/settings.json
marker-dao: My version of VS Code changed these settings on its own. I will now revert back to the previous version.

marker-dao: [1dcfb467fcbcf0d07992d1c4023b3c01dbed1eb0]

.github/workflows/testcafe_tests.yml
marker-dao: We tried to check if the test execution time would decrease if we set the value to 2. We can remove this value now, as there is no noticeable effect.

marker-dao: [f3e0583b6c68591e7ea000377b1d886c0e1f85ee]

ksercs (approved):

EugeniyKiyashko (approved):

PR #26497, opened 2024-01-25
DropDownButton - datasource integration refactoring

iBat:


packages/devextreme/js/ui/drop_down_button.js
```diff
@@ -757,8 +752,13 @@ const DropDownButton = Widget.inherit({
             default:
                 this.callBase(args);
         }
+    },
+
+    // TODO move me?
```
ksercs: Move where? It's a public method.
```suggestion
```

packages/devextreme/js/ui/drop_down_button.js
```diff
@@ -204,12 +199,12 @@ const DropDownButton = Widget.inherit({
 
         if(this._list && this._lastSelectedItemData !== undefined) {
             const cachedResult = this.option('useSelectMode') ? this._list.option('selectedItem') : this._lastSelectedItemData;
-            return d.resolve(cachedResult);
+            return (new Deferred()).resolve(cachedResult);
```
ksercs: Why do we need this change?

packages/devextreme/js/__internal/data_controller.ts
```diff
@@ -0,0 +1,265 @@
...
+      store: new ArrayStore({
+        key,
+        data: items,
+      }),
+      pageSize: 0,
+    });
+    return new DataController(dataSource as unknown as DataSourceType);
+  }
+
+  _initDataSource(dataSourceOptions: DataSourceType): void {
```
ksercs: I think `update` is better here because we start this function with a restoring of a previous dataSource
```suggestion
  _updateDataSource(dataSourceOptions: DataSourceType): void {
```

packages/devextreme/js/__internal/data_controller.ts
```diff
@@ -0,0 +1,265 @@
...
+  store(): DataSourceType {
+    return this._dataSource.store();
+  }
+
+  items(): unknown[] {
+    return this._dataSource.items();
+  }
+
+  itemsToDataSource(value: unknown[], key: string): void {
+    if (!this._dataSource) {
```
ksercs: Do we really want somewhere for this function to make nothing if some dataSource is already created?
If not I prefer we name it as `updateDataSourceByItems` and start it with a disposing of an existing dataSource

packages/devextreme/js/ui/drop_down_button.js
```diff
@@ -117,14 +115,23 @@ const DropDownButton = Widget.inherit({
...
     },
 
+    _initDataController() {
+        const dataSource = this.option('dataSource');
+
+        if(dataSource) {
+            this._dataController = new DataController(dataSource);
+        } else {
+            this._dataController = DataController.initFromArray(this.option('items'), this._getKey());
+        }
```
ksercs: Don't we want to incapsulate this logic inside of `dataController`? It's actual not only for a DropDownButton.

packages/devextreme/js/ui/drop_down_button.js
```diff
@@ -691,14 +686,14 @@ const DropDownButton = Widget.inherit({
...
                 this._updateItemCollection(name);
                 break;
             case 'dataSource':
                 if(Array.isArray(value)) {
-                    this._updateDataSource(this.option('dataSource'));
+                    this._updateDataController(value);
                 } else {
-                    this._initDataSource();
+                    this._dataController.initDataSource(value);
                     this._updateKeyExpr();
```
ksercs: I think this `if-else` condition indicates some incorrect splitting of responsibility between functions.
For example if `itemsToDataSource` calls `ds.dispose` we can refactor it this way:
```javascript
if(Array.isArray(value)) {
    this._dataController.itemsToDataSource(items, this._getKey());
} else {
   this._dataController.initDataSource(value);
}
this._updateKeyExpr();
```
Now it's still a bit strange why we should call different updates on `dataSource` option change - possible it shoulf be also inside of a dataController

packages/devextreme/js/ui/drop_down_button.js
iBat: I can't see a smooth solution to move this logic into the DataController. Make 3 optional parameters in constructor? Its possible but in this case we will always perform unnecessary getters - this.option('items') and this._getKey().

packages/devextreme/js/__internal/data_controller.ts
```diff
@@ -0,0 +1,264 @@
...
+
+  resetDataSourcePageIndex(): void {
+    if (this.pageIndex()) {
+      this.pageIndex(0);
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
+      this.load();
+    }
+  }
+
+  initDataSource(dataSourceOptions: DataSourceType): void {
```
ksercs:
```suggestion
  updateDataSource(dataSourceOptions: DataSourceType): void {
```

packages/devextreme/js/ui/drop_down_button.js
```diff
@@ -691,16 +686,16 @@ const DropDownButton = Widget.inherit({
...
                 break;
             case 'items':
-                this._updateDataSource(this.option('items'));
+                this._updateDataController(this.option('items'));
                 this._updateItemCollection(name);
                 break;
             case 'dataSource':
                 if(Array.isArray(value)) {
-                    this._updateDataSource(this.option('dataSource'));
+                    this._dataController.updateDataSourceByItems(value, this._getKey());
```
ksercs: Cannot we move this logic to `updateDataSource` and leave the single call here?
```javascript
this._dataController.updateDataSource(value);
```

packages/devextreme/js/ui/drop_down_button.js
ksercs: I believe we're supposed to do it with a static method at all.
`dataSource` option can be a usual array too => we should handle this on dataSource creating inside a DataController.
I think it's okay to pass `dataSource || items` as a first parameter + add one more optional parameter `config`

packages/devextreme/js/ui/drop_down_button.js
```diff
@@ -635,9 +630,9 @@ const DropDownButton = Widget.inherit({
         }
     },
 
-    _updateDataSource: function(items = this._dataSource.items()) {
-        this._dataSource = undefined;
-        this._itemsToDataSource(items);
+    _updateDataController: function(items = this._dataController.items()) {
+        this._dataController.resetDataSource();
```
ksercs: shouldn't it be inside `updateDataSource`?

packages/devextreme/js/__internal/data_controller.ts
```diff
@@ -0,0 +1,257 @@
...
+
+  store(): DataSourceType {
+    return this._dataSource.store();
+  }
+
+  items(): unknown[] {
+    return this._dataSource.items();
+  }
+
+  updateDataSourceByItems(items: unknown[], key?: string): void {
```
ksercs: Don't we wanna mark it as private and move up?

packages/devextreme/js/ui/drop_down_button.js
```diff
@@ -117,14 +115,23 @@ const DropDownButton = Widget.inherit({
...
     },
 
+    _initDataController() {
+        const dataSource = this.option('dataSource');
+        let key;
+
+        if(!dataSource) {
+            key = this._getKey();
+        }
+        this._dataController = new DataController(dataSource ?? this.option('items'), key);
```
ksercs: I suggest u making the second parameter object `{ key: this._getKey() }` and always pass it.
It'll make this code more clear.
Ideologically i think we should always pass all possible-to-use props to DataController and already DataContoller should decide when to use it and when not.

packages/devextreme/testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
```diff
@@ -1565,19 +1564,16 @@ QUnit.module('common use cases', {
...
+        assert.strictEqual(this.dropDownButton.option('selectedItem'), null, 'selectedItem is correct');
 
-            const list = getList(this.dropDownButton);
-            assert.deepEqual(list.option('selectedItemKeys'), [], 'list selectedItemKey is kept');
-            assert.strictEqual(list.option('selectedItem'), undefined, 'list selectedItem is correct');
-            done();
-        });
+        const list = getList(this.dropDownButton);
+        assert.deepEqual(list.option('selectedItemKeys'), [], 'list selectedItemKey is kept');
+        // TODO research me
```
ksercs: I don't see what should happen to force us to research it later. I think we should deal with it now.
Let me check a bit later.

packages/devextreme/js/__internal/data_controller.ts
```diff
@@ -0,0 +1,257 @@
...
+    this._disposeDataSource();
```
ksercs: seems like it should be in "updateDataSource"

packages/devextreme/testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
ksercs: I added a `console.log` to List's option change and found the following difference:
before:
```text
OPTIONS CHANGE selectedItem undefined  ->  null
OPTIONS CHANGE selectedItem {id: 1, name: 'a'}  ->  null
OPTIONS CHANGE selectedItem {id: 1, file: 'vs.exe', name: 'Trial for Visual Studio', icon: 'box'}  ->  null
```
after:
```text
OPTIONS CHANGE selectedItem undefined  ->  null
OPTIONS CHANGE selectedItem {id: 1, name: 'a'}  ->  null
OPTIONS CHANGE selectedItem undefined  ->  null
```
U can see that 3rd change is made with different `previousValue`. 

**Why this difference is a reason of a red test?**

All complexity of this behavior is related to selection options sync in collection: [link](https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme/js/ui/collection/ui.collection_widget.edit.js#L299). If we change `selectedItem` option from one nullish value to another (for example, from undefined to null), no addtiional sync is made.

**Why does the last `selectedItem` option change in the last version have such a strange previous value?**

Because of this [call](https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme/js/ui/drop_down_button.js#L629).
The `_loadSelectedItem` promise is resolved with `selectedItem=null` but we anyway change `selectedItemKeys` to `[1]` and it leads to update of `selectedItem` on options sync in collection.

**Why does the last `selectedItem` option change in the updated version have an `undefined` as a previous value if we still have a call mentioned in the previous question?**

Because when we try to sync options dataSource [loads](https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme/js/ui/selection/selection.strategy.standard.js#L227) nothing in the new version, but loads an old item in the previous version. **Why? I suggest u debug this.**

Actual issue can be resolved if we change `resolve` with a `reject` [here](https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme/js/ui/drop_down_button.js#L215). It seems to be more correct from idealogical point of view.

packages/devextreme/js/ui/drop_down_button.js
```diff
@@ -384,12 +367,12 @@ const DropDownButton = Widget.inherit({
...
             grouped: this.option('grouped'),
             groupTemplate: this.option('groupTemplate'),
-            keyExpr: this._getKey(),
+            keyExpr: this._dataController.key(),
             noDataText: this.option('noDataText'),
             displayExpr: this.option('displayExpr'),
             itemTemplate: this.option('itemTemplate'),
             items: this.option('items'),
-            dataSource: this._dataSource,
+            dataSource: this._dataController.getDataSource(),
```
ksercs, line 375: don't we wanna pass a dataController instead of ds? Or u gonna make it later?

iBat: I expected to do this later.

packages/devextreme/js/__internal/data_controller.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  private _keyExpr: string;
+
+  constructor(dataSourceOptions: unknown[] | DataSourceType, { key }: DataControllerOptions) {
+    this._keyExpr = key;
+    this.updateDataSource(dataSourceOptions);
+  }
+
+  _updateDataSource(dataSourceOptions: DataSourceType): void {
+    if (dataSourceOptions) {
```
ksercs: If we write this check this way we can get rid of an additional level of indent:
```javascript
if (!dataSourceOptions) {
    return;
}
```

packages/devextreme/js/__internal/data_controller.ts
```diff
@@ -0,0 +1,278 @@
...
+      if (dataSourceOptions instanceof DataSource) {
+        this._isSharedDataSource = true;
+        this._dataSource = dataSourceOptions;
+      } else {
+        const DataSourceConstructor = DataSource;
```
ksercs: This variable is strange to me. I prefer we just use `new DataSource` below

packages/devextreme/testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
```diff
@@ -1565,19 +1564,16 @@ QUnit.module('common use cases', {
...
+        assert.strictEqual(this.dropDownButton.option('selectedItem'), null, 'selectedItem is correct');
 
-            const list = getList(this.dropDownButton);
-            assert.deepEqual(list.option('selectedItemKeys'), [], 'list selectedItemKey is kept');
-            assert.strictEqual(list.option('selectedItem'), undefined, 'list selectedItem is correct');
-            done();
-        });
+        const list = getList(this.dropDownButton);
+        assert.deepEqual(list.option('selectedItemKeys'), [], 'list selectedItemKey is kept');
+
```
ksercs, line 1578:
```suggestion
```

packages/devextreme/js/ui/drop_down_button.js
```diff
@@ -559,6 +542,10 @@ const DropDownButton = Widget.inherit({
     },
 
     _updateActionButton(selectedItem) {
+        if(selectedItem === undefined) {
```
ksercs: Why do we need it now?

packages/devextreme/testing/tests/DevExpress.ui.widgets/dropDownButton.tests.js
```diff
@@ -1565,19 +1564,16 @@ QUnit.module('common use cases', {
             useSelectMode: true,
             selectedItemKey: 1
         });
-
         const items = [{ id: 3, name: 'test' }];
         this.dropDownButton.option('items', items);
 
-        setTimeout(() => {
```
ksercs: We repaired the test => let's restore its initial state? Now we don't have a reason to change it.
If u insist on its change, let's make it for all tests added in that commit and ensure that it's not evergreen.

ksercs (dismissed):

Raushen (approved):

ksercs (approved):

packages/devextreme/js/__internal/data_controller.ts
```diff
@@ -0,0 +1,277 @@
...
+  cancelAll: () => void;
+  filter: (filter?: unknown) => unknown;
+  group: (group?: unknown) => unknown;
+  items: () => unknown[];
+  key: () => string;
+  keyOf: (item: unknown) => string;
+  paginate: (paginate?: unknown) => unknown;
+  pageIndex: <TIdx = number | undefined>(pageIndex: TIdx) => TIdx extends number
+    ? undefined
+    : number;
```
pomahtri, line 26: minor (can be skipped, just stylistic change)

```suggestion
  pageIndex: {
    (): number;
    (pageIndex: number): void;
  };
```

PR #26516, opened 2024-01-26
TreeList - Selection CheckBox in a data row isn't navigable with Tab button if this CheckBox was focused manually (T1207467)

Raushen:
See https://supportcenter.devexpress.com/ticket/details/t1207467/treelist-selection-checkbox-in-a-data-row-isn-t-navigable-with-tab-button-if-this


packages/devextreme/js/__internal/grids/grid_core/editor_factory/m_editor_factory.ts
```diff
@@ -61,7 +61,19 @@ export class EditorFactory extends ViewControllerWithMixin {
...
+      input:focus,
+      button:focus,
+      textarea:focus,
+      div[tabindex]:focus,
+      .dx-lookup-field:focus,
+      .dx-checkbox:focus,
+      .dx-switch:focus,
+      .dx-dropdownbutton
+      .dx-buttongroup:focus,
+      .dx-adaptive-item-text:focus`;
```
williamvinogradov: Let's remove spaces and new line chars from this selector string.
Something like this:
```suggestion
    const focusedElementSelector = `
      td[tabindex]:focus,
      ${rowSelector},
      input:focus,
      button:focus,
      textarea:focus,
      div[tabindex]:focus,
      .dx-lookup-field:focus,
      .dx-checkbox:focus,
      .dx-switch:focus,
      .dx-dropdownbutton
      .dx-buttongroup:focus,
      .dx-adaptive-item-text:focus`.replace(/\n| /g, ''); // after the "|" symbol there is a space!
```

Because when you create a string like this, you add additional spaces and "new line" chars to this string
```typescript
const myStr = `
  a,
  b,
  c`;
  console.log(myStr); // "\n  a,\n  b,\n  c"
  console.log(myStr.replace(/ /g, '_')); // "\n__a,\n__b,\n__c"
  console.log(myStr.replace(/\n| /g, '')); // "a,b,c"
``` 

Maybe the jQuery clears a selector string of these symbols (I'm not 100% sure), but in my opinion, it's better to replace it here :)

packages/devextreme/js/__internal/grids/grid_core/editor_factory/m_editor_factory.ts
```diff
@@ -103,7 +115,11 @@ export class EditorFactory extends ViewControllerWithMixin {
   }
 
   _needHideBorder($element) {
-    return $element.hasClass(EDITOR_INLINE_BLOCK);
+    const rowsViewElement = this.getView('rowsView').element();
+    const isRowsView = $element.closest(rowsViewElement).length > 0;
```
williamvinogradov, line 119: Question

The ```closest``` is a very expensive operation.
Can we just check the existence of the ```rowsView``` element here?
Or the ```rowsView``` may have the ```element``` when it's not attached to the DOM?

Something like this:
```typescript
const isRowsView = !!this.getView('rowsView').element?.();
```

packages/devextreme/js/__internal/grids/grid_core/editor_factory/m_editor_factory.ts
```diff
@@ -103,7 +115,11 @@ export class EditorFactory extends ViewControllerWithMixin {
...
+    const isEditing = this.getController('editing').isEditing();
+
+    return $element.hasClass(EDITOR_INLINE_BLOCK) || isRowsView && !isEditing;
```
williamvinogradov: Minor comment.

Let's add the brackets to make this bool expression more obvious:
```suggestion
    return $element.hasClass(EDITOR_INLINE_BLOCK) || (isRowsView && !isEditing);
```

packages/devextreme/js/__internal/grids/grid_core/editor_factory/m_editor_factory.ts
Raushen: We found that closest is appropriate operation because it traversing up through its ancestors in the DOM tree https://api.jquery.com/closest/

williamvinogradov: I was wrong, the ```closest``` traverse DOM tree only up and matches parent nodes.
So, using ```closest``` is ok :)

williamvinogradov (approved):

PR #26519, opened 2024-01-29
Datagrid: Fix the bottom border of the last row (T1200567)

Alyar666:
See https://devexpress.com/issue=T1200567


packages/devextreme/scss/widgets/base/_gridBase.scss
```diff
@@ -56,8 +56,11 @@
     }
 
     > .dx-#{$widget-name}-rowsview.dx-last-row-border {
-      tbody:last-child > .dx-data-row:nth-last-child(2) > td {
-        border-bottom-width: 0;
+      tbody:last-child > .dx-data-row:nth-last-child(2) {
+        &,
+        & > td {
+          border-bottom-width: 0;
+        }
```
williamvinogradov, line 63: Can we re-write this selector to get rid of the ```&```?
I'm not sure, but it looks like a reversive selector :)

williamvinogradov (approved):

packages/devextreme/scss/widgets/base/_gridBase.scss
williamvinogradov: Sorry, I was wrong here.
The ```&``` is best solution here :)

PR #26564, opened 2024-02-05
TextBox: correct handlers order after change mask option (T1214019)

iBat:


packages/devextreme/js/ui/text_box/ui.text_editor.mask.js
```diff
@@ -183,6 +183,7 @@ const TextEditorMask = TextEditorBase.inherit({
         this.$element().removeClass(TEXTEDITOR_MASKED_CLASS);
         this._maskRulesChain = null;
 
+        this._maskStrategy.clean();
```
ksercs: `clean` is a kind of opposite action for `init`. In this method I see that we added `clean` but we don't have `init` here. It means afaid that cleaned dragtimer, caretTimeout and inputHandlerTimeout is not recreated.

So I don't get why do we need this change and is it safe.

packages/devextreme/js/ui/text_box/ui.text_editor.mask.js
```diff
@@ -525,9 +526,15 @@ const TextEditorMask = TextEditorBase.inherit({
     },
 
     _updateMaskOption: function() {
+        const valueChangeEvent = addNamespace(this.option(this._getValueChangeEventOptionName()), `${this.NAME}ValueChange`);
+        const $input = this._input();
+
         this._updateHiddenElement();
         this._renderMask();
         this._validateMask();
+        // reattach events to get correct handlers order (T1214019)
```
ksercs: We usually use a `NOTE:` prefix for informative comments.
 But actually I don't see a necessity to add a comment here because this solution does not look like something specific and non-intuitive. Moreover we add a special test for this

packages/devextreme/js/ui/text_box/ui.text_editor.mask.js
```diff
@@ -525,9 +526,15 @@ const TextEditorMask = TextEditorBase.inherit({
     },
 
     _updateMaskOption: function() {
+        const valueChangeEvent = addNamespace(this.option(this._getValueChangeEventOptionName()), `${this.NAME}ValueChange`);
```
ksercs: what about similar events like keydown? Does it work even without change?

packages/devextreme/js/ui/text_box/ui.text_editor.mask.js
```diff
@@ -525,9 +526,15 @@ const TextEditorMask = TextEditorBase.inherit({
...
+        const $input = this._input();
+
         this._updateHiddenElement();
         this._renderMask();
         this._validateMask();
+        // reattach events to get correct handlers order (T1214019)
+        eventsEngine.off($input, valueChangeEvent);
+        eventsEngine.on($input, valueChangeEvent, this._valueChangeEventHandler.bind(this));
```
ksercs: Cannot we just call already written methods instead of writing an imperative code?
For example, call `_refreshValueChangeEvent`

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/textEditorParts/mask.tests.js
```diff
@@ -2580,6 +2580,34 @@ QUnit.module('Hidden input', {}, () => {
...
+            .caret(1)
+            .type('2')
+            .change();
+
+        $textEditor.blur();
+        instance.option('mask', '000');
+
+        const $hiddenInput = $textEditor.find('input[type=hidden]');
+
+        assert.equal($hiddenInput.val(), 12, 'Hidden value is saved');
```
ksercs: Let's use strict equal. It's strange to see that input text equals a number.

ksercs: Why do you check a value of the hidden input here? I think we should check usual visible input's text.
I also prefer we check how `value` prop is changed because I guess it's unexpected too

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/textEditorParts/mask.tests.js
```diff
@@ -2580,6 +2580,34 @@ QUnit.module('Hidden input', {}, () => {
         assert.equal($hiddenInput.attr('name'), 'Editor with mask', 'name of hidden input');
     });
 
+    QUnit.test('Preserve value if the mask option is updated programmatically (T1214019)', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('Value change should work as usual after mask option is updated programmatically (T1214019)', function(assert) {
```

packages/devextreme/js/ui/text_box/ui.text_editor.mask.js
iBat: Without this change, we get an error about uncleaned timeout.

ksercs (approved):

EugeniyKiyashko (approved):

marker-dao (approved):

PR #26571, opened 2024-02-06
Form: formData should not be considered as an array if it only has a "length" property (T1213983)

marker-dao:


pomahtri (changes requested):
`each` is old function which was created before existing of modern es methods like `.forEach`, `for (let el of someArray)`, `Object.keys` etc. It's better to use standard modern utilities, and not change this one

marker-dao:
> `each` is old function which was created before existing of modern es methods like `.forEach`, `for (let el of someArray)`, `Object.keys` etc. It's better to use standard modern utilities, and not change this one

Yes, I agree with you. Our team suggested that this might be a solution. So I just tried it.

pomahtri (dismissed):

pomahtri:
Please note that using `Object.keys()` differs from using `for in` loop, which was inside `each`, since `Object.keys()` ignores prototype
But I don't know whether it's good for your case or bad

Zedwag (dismissed):

AlexanderMoiseev (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -632,6 +632,22 @@ QUnit.test('form.option("onFieldDataChanged", "newHandler") -> check new handler
     assert.equal(onFieldDataChangedStub.callCount, 2, 'new handler is called after editor value is changed');
 });
 
+QUnit.test('onFieldDataChanged must be called once (T1213983)', function(assert) {
```
ksercs: Let's make scenario description more precise plz
```suggestion
QUnit.test('onFieldDataChanged must be called once if new formData contains "length" property (T1213983)', function(assert) {
```

ksercs (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
marker-dao: [9f919472bf239ce538e466cd6a61e848fbd9df1a]

ksercs (approved):

PR #26596, opened 2024-02-07
HtmlEditor - Should pass state to quill when formatting started

alexanderPolosatov:


packages/devextreme/js/ui/html_editor/utils/toolbar_helper.js
```diff
@@ -337,6 +337,9 @@ function getDefaultClickHandler(module, name) {
         const value = formats[name];
         const newValue = !(isBoolean(value) ? value : isDefined(value));
 
+        // Remove condition after adding method
+        module.quill.startFormat?.();
```
ksercs, line 341: Hm, `startFormat` is a strange name for such a specific call. We don't use it on formatting applied other way, right?
I prefer we name it explicitly by the operation we do inside - `raiseCompositionEnd`.

By the way, are u sure it's enough to call it here? Will it work if users adds a custom toolbar button with manual formatting on click? Maybe it's better to move it inside of `applyFormat`?

PR #26609, opened 2024-02-08
DataGrid / TreeList: Add private alwaysSelectByShift option (T1180554)

Alyar666:


js/ui/selection/selection.js
```diff
@@ -139,7 +140,11 @@ export default class Selection {
         const itemKey = this.options.keyOf(itemData);
 
         keys = keys || {};
-        const allowSelectByShift = keys.shift && (allowLoadByRange !== false || (!focusedItemNotInLoadedRange && !shiftFocusedItemNotInLoadedRange));
+        let allowSelectByShift = keys.shift;
+
+        if(alwaysSelectByShift === false && allowSelectByShift) {
+            allowSelectByShift &= (allowLoadByRange !== false || (!focusedItemNotInLoadedRange && !shiftFocusedItemNotInLoadedRange));
```
williamvinogradov, line 146: Do we need the bitwise operator ```&=``` here? :) 
Can we re-write this code without it?

williamvinogradov (approved):

PR #26611, opened 2024-02-08
DataGrid / TreeList: Add private alwaysSelectByShift option (T1180554)

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/selection/m_selection.ts
```diff
@@ -192,6 +192,7 @@ export class SelectionController extends modules.Controller {
       selectedKeys: this.option('selectedRowKeys'),
       mode: this._selectionMode,
       deferred,
+      alwaysSelectByShift: (selectionOptions as any).alwaysSelectByShift,
```
pomahtri: let's add this option to our private options typing in `InternalGridOptions` in `grid_core/m_types`
and then remove `as any`

williamvinogradov (approved):

PR #26655, opened 2024-02-13
Splitter: add doc declarations

EugeniyKiyashko:
Docs: https://github.com/DevExpress/devextreme-documentation/pull/6063


.github/workflows/wrapper_tests.yml
```diff
@@ -52,7 +49,6 @@ jobs:
             echo "Generated code is outdated. The following files have uncommitted changes:"
             echo "$changes";
             echo "To update generated code, use "npm run regenerate-all" and commit changes."
-            exit 1
```
mpreyskurantov: Revert CI hacks

mpreyskurantov (dismissed):

packages/devextreme/js/ui/splitter.js
```diff
@@ -3,3 +3,90 @@ import Splitter from '../__internal/ui/splitter/splitter';
...
+ */
+
+/**
+ * @name dxSplitter.focus
+ * @publicName focus()
+ * @hidden
+ */
+
+/**
+ * @name dxSplitterItem
```
IlyaKhD, line 84: Can this line be moved to .d.ts?

mpreyskurantov: Not now... Moving causes different side effects: in dx.all d.ts bundle, angular modules against dx.all or inheritance of collection item members. Let's re-try do to if after a task with angular vs dx.all.

IlyaKhD (dismissed):

mpreyskurantov (approved):

vladaskorohodova (approved):

PR #26668, opened 2024-02-14
🛠️ Refactoring | Grid native classes: PivotGrid's related migration (6)

Raushen:


packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser.ts
```diff
@@ -587,17 +590,17 @@ const FieldChooser = FieldChooserBase.inherit({
...
       that._renderFieldsTreeView($fieldsContainer);
     }
-  },
+  }
 
   _getSortableOptions() {
     // TODO
     return {
-
+      direction: '',
```
williamvinogradov, line 598: It's may be dangerous change
But if you sure that it's okay let's leave it as is :)

williamvinogradov (approved):

PR #26745, opened 2024-02-23
FileUploader: hint with "null" text should not be shown when hint option is not defined (T1218499)

Zedwag:


marker-dao (dismissed):

packages/devextreme/js/ui/file_uploader.js
```diff
@@ -301,16 +302,21 @@ class FileUploader extends Editor {
             eventsEngine.on(this._$fileInput, 'click', e => {
                 e.stopPropagation();
                 this._resetInputValue();
-                return this.option('useNativeInputClick') || this._isCustomClickEvent;
+                return useNativeInputClick || this._isCustomClickEvent;
```
ksercs: hm, I u sure it's just a refactoring without behavior change?
Earlier we got option value when event is handled, not only on subscribe. Does it work now if we change `useNativeInputClick` at runtime?
Let's also write a test.

Zedwag: thx for pointing out on this, reverted this and added test

ksercs (approved):

PR #26748, opened 2024-02-23
Demos: supporting testing across multiple themes

EugeniyKiyashko:


apps/demos/utils/visual-tests/matrix-test-helper.js
```diff
@@ -54,6 +69,33 @@ export function getChangedFiles() {
...
+export function globalReadFrom(basePath, relativePath, mapCallback) {
+  const absolute = join(basePath, relativePath);
+  if (existsSync(absolute)) {
+    const result = readFileSync(absolute, 'utf8');
+    return (mapCallback && result && mapCallback(result)) || result;
+  }
+  return null;
+}
+
+export function changeTheme(dirName, relativePath, demoPath, theme) {
```
iBat: 3 starting params looks like overkill. As I see you always use together at least 1st and 2nd. Why dont change it with, say,  (demosRoot, demoPath, theme) and join some paths in a callers.

apps/demos/utils/visual-tests/matrix-test-helper.js
```diff
@@ -54,6 +69,33 @@ export function getChangedFiles() {
...
+  if (!theme || theme === THEME.generic) {
+    return;
+  }
+
+  const indexFilePath = join(dirName, `${relativePath}${demoPath}/index.html`);
```
iBat: When you use `path.join` - no need to build strings. All strings anyway will be splitted by slashes/backslashes and rebuilded using `path.sep`. 
```suggestion
  const indexFilePath = join(dirName, relativePath, demoPath, 'index.html');
```

EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/26748/commits/5099ddceea7e349f16bf267aaa3e4127e5ef32bf

apps/demos/utils/visual-tests/matrix-test-helper.js
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/26748/commits/5099ddceea7e349f16bf267aaa3e4127e5ef32bf

.github/workflows/demos_visual_tests.yml
```diff
@@ -72,12 +72,13 @@ jobs:
       fail-fast: false
       matrix:
         STRATEGY: [screenshots, accessibility]
+        THEME: ['generic.light', 'material.blue.light', 'fluent.blue.light']
```
marker-dao, line 75: Why we are testing only light themes?

iBat (approved):

.github/workflows/demos_visual_tests.yml
EugeniyKiyashko: At the moment, yes. If there is a need to add other color schemes, we can do it later in separate pull request.

marker-dao (approved):

andrewmakarov (approved):

PR #26759, opened 2024-02-26
splitter: multiple panes resize support

AlexanderMoiseev:


packages/devextreme/js/__internal/ui/splitter/utils/layout.ts
```diff
@@ -53,12 +68,27 @@ export function getNewLayoutState(delta, handle, currentLayout, items): number[]
   const firstItemIndex: number = $(handle).prev().data().dxItemIndex;
```
EugeniyKiyashko: We should improve it and avoid interacting with the DOM

EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/26759/commits/342669b0bf3fd2cb97c049da6a67bfe515a8dac6

EugeniyKiyashko (approved):

PR #26770, opened 2024-02-27
Fix demos launcher

iBat:


build/install-wrappers.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+echo This is a workaround for working with demos until the wrappers are reworked
+echo please run 'npm run all:build' before this script
+echo this sript should be rerunned after each 'npm install'
```
IlyaKhD: and after build?
```suggestion
echo this sript should be rerunned after each 'npm install' and 'npm run all:build' 
```

build/install-wrappers.ts
```diff
@@ -0,0 +1,34 @@
...
+async function main() {
+  const version = '24.1.0';
+  const PACKAGES = ['devextreme-angular', 'devextreme-react', 'devextreme-vue'];
+
+  await Promise.all(PACKAGES.map(packageName => new Promise((resolve) => {
+    const tgzPath = `artifacts/npm/${packageName}-${version}.tgz`;
+    const targetPath = `node_modules/${packageName}`;
+
+    sh.rm('-rf', targetPath);
+    sh.mkdir(targetPath);
```
IlyaKhD: We can completely get rid off shelljs in this script, using `fs`.


Smth like this:
```suggestion
    fs.rm(targetPath, { recursive: true, force: true });
    fs.mkdirSync(targetPath);
```
```

File: build/install-wrappers.ts [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+    const tgzPath = `artifacts/npm/${packageName}-${VERSION}.tgz`;
+    const targetPath = `node_modules/${packageName}`;
+
+    fs.rmSync(targetPath, { recursive: true, force: true });
+    fs.mkdirSync(targetPath);
+
+    fs.createReadStream(tgzPath).pipe(createUnzip()).pipe(tar.extract(targetPath, { strip: 1 })).on('finish', () => {
+      console.log(`Installed ${packageName}`);
+      resolve(1);
+    }).on('error', () => {
```
IlyaKhD:
```suggestion
  await Promise.all(PACKAGES.map(packageName => new Promise<void>((resolve) => {
    const tgzPath = `artifacts/npm/${packageName}-${VERSION}.tgz`;
    const targetPath = `node_modules/${packageName}`;

    fs.rmSync(targetPath, { recursive: true, force: true });
    fs.mkdirSync(targetPath);

    fs.createReadStream(tgzPath).pipe(createUnzip()).pipe(tar.extract(targetPath, { strip: 1 })).on('finish', () => {
      console.log(`Installed ${packageName}`);
      resolve();
    }).on('error', () => {
```

IlyaKhD (approved):

PR #26783, opened 2024-02-28
feat: refactor React wrappers' base classes to functional components

VasilyStrelyaev:


packages/devextreme-react/src/core/component-base-func.tsx
```diff
@@ -0,0 +1,397 @@
...
+          renderContent(),
+          React.createElement(TemplateManager, {
+            init: setTemplateManagerHooks,
+          }),
+        ),
+        isPortalComponent && renderPortal(),
+      );
+    },
+  ),
+) as <P extends IHtmlOptions>(props: P & ComponentBaseProps, ref: React.ForwardedRef<ComponentBaseRef>) => ReactElement | null;
```
VasilyStrelyaev: Note for reviewers: This type casting is required because TypeScript is unable to infer types correctly if a functional component is a generic function and at the same time uses HOC functions (like `memo` or `forwardRef`).

Do you feel like we can go with a workaround like that, or we should come up with something else?

packages/devextreme-react/src/core/__tests__/nested-option.test.tsx
```diff
@@ -1,67 +1,132 @@
...
+const NestedComponentWithPredfeinedProps = memo(function NestedComponentWithPredfeinedProps(props: any) {
+  return (
+    <ConfigurationComponent<{ a: number } & React.PropsWithChildren>
+      {...props}
+    />
+  );
+}) as React.MemoExoticComponent<any> & {
+  OptionName: string
+  PredefinedProps: Record<string, string>
+};
```
VasilyStrelyaev, line 30: Note for reviewers: This type casting is a good illustration for my next question to you.

Currently, components like nested options have public static class properties: https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme-react/src/accordion.ts#L154

With class components, type definitions look rather nice: just classes with static props.

If we try to achieve a similar effect with functional components (by assigning additional properties to `React.FC`), type definitions will probably become intersection types like in this code. Functionally, there shouldn't be much difference for the user (besides the fact that we have FCs now instead of classes, which is already a BC). Cosmetically, we can come up with a good name for the second intersection half.

What do you think?

packages/devextreme-react/src/core/extension-component-func.tsx
```diff
@@ -0,0 +1,67 @@
...
+  useEffect,
+  useCallback,
+  ReactElement,
+  memo,
+} from 'react';
+
+import { IHtmlOptions, ComponentBaseRef, ComponentBase } from './component-base-func';
+import { ComponentProps } from './component-func';
+
+type NestedOptionElement = ReactElement<any, React.JSXElementConstructor<any> & { isExtensionComponent: boolean }>;
```
VasilyStrelyaev: Previously, we detected `ExtensionComponent` descendants just by checking their prototype (see the `isPrototypeOf` usage in `component.ts`).

Now, `Validator` component will just use `ExtensionComponent` as its render result, with no way to tell this by looking at the `ReactElement`. So, as long as we already have to assign properties to functional components (see https://github.com/DevExpress/DevExtreme/pull/26783/files#r1506083444), we may want to assign a flag property to all the extension components to make them identifiable by their React elements.
An example of how this flag is used is here: https://github.com/DevExpress/DevExtreme/pull/26783/files#diff-c2ab09c6d3ab5deec02d484b3733fdb2e341a2a66f8f0a63b0b08003df89ea85R23

packages/devextreme-react/src/core/widget-config.ts
```diff
@@ -1,11 +1,34 @@
...
+  beforeCreateWidget: () => undefined,
+  afterCreateWidget: () => undefined,
+};
+
 function isIgnoredProp(name: string) {
-  return name === 'children' || name === classNamePropName || elementPropNames.includes(name);
+  return name === 'children'
+    || name === classNamePropName
+    || elementPropNames.includes(name)
+    || Object.prototype.hasOwnProperty.call(internalProps, name);
```
VasilyStrelyaev, line 31: In this file, we parse all props and those props that we don't filter out will (or may) be sent to the Widgets in `optionChanged` events. As now base components have many props for internal usage (that previously used to be inherited class properties), we need to filter them out to avoid spamming Widgets with invalid `optionChanged` when these props change.

To ensure that future us won't forget to exclude any new "internal" props, I created a typed object of the internal props type, but with all props marked as required, and filter out all its props at runtime.

packages/devextreme-react/src/core/configuration/react/element.ts
```diff
@@ -48,7 +49,7 @@ function getElementInfo(
     };
   }
 
-  if (reactElement.type === TemplateComp) {
+  if (reactElement.type === TemplateComp || reactElement.type === TemplateFuncComp) {
```
VasilyStrelyaev: That's to make it work in production (with old class components) and testing (with new functional components) in the meantime, while we are working on generators. This will go away once we finish the feature.

I just didn't want to duplicate this file (like I did with components) just for the sake of a single line.

packages/devextreme-react/package.json
```diff
@@ -32,8 +32,8 @@
     "prop-types": "^15.8.1"
   },
   "devDependencies": {
-    "@testing-library/react": "^13.3.0",
-    "@testing-library/user-event": "^13.0.1",
+    "@testing-library/react": "^14.2.1",
```
VasilyStrelyaev, line 35: I updated testing library because with the new version, I was able to get rid of all the warnings during tests.

IlyaKhD (approved):

ivanblinov2k17 (approved):

PR #26784, opened 2024-02-28
Form: Prevent the DropDownBox from losing its value after the form is resized (T1196835)

iBat:


packages/devextreme/js/ui/form/ui.form.js
```diff
@@ -1182,10 +1183,12 @@ const Form = Widget.inherit({
     },
 
     _refresh: function() {
-        const editorSelector = `.${FOCUSED_STATE_CLASS} > :not(.dx-dropdowneditor-input-wrapper) input,`
-            + ` .${FOCUSED_STATE_CLASS} textarea`;
+        const editorSelector = '.' + FOCUSED_STATE_CLASS + ' input, .' + FOCUSED_STATE_CLASS + ' textarea';
```
marker-dao: Let's use template literals here

packages/devextreme/js/ui/form/ui.form.js
```diff
@@ -1182,10 +1183,12 @@ const Form = Widget.inherit({
...
+        const $editors = this.$element().find(editorSelector).filter((idx, el) =>
+            !$(el).parents(UNREFRESHABLE_COMPONENT_WRAPPERS.join(', ')).length
```
EugeniyKiyashko: We can call 'trigger' method for certain element here if it needs.

packages/devextreme/js/ui/form/ui.form.js
```diff
@@ -1182,10 +1183,12 @@ const Form = Widget.inherit({
...
+        );
 
-        eventsEngine.trigger(this.$element().find(editorSelector), 'change');
+        eventsEngine.trigger($editors, 'change');
```
EugeniyKiyashko: Also well know - when this 'trigger' call was added? What does it fix?

iBat: It was fix of this https://isc.devexpress.com/internal/ticket/details/T404958

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4737,37 +4737,79 @@ QUnit.module('reset', () => {
         assert.strictEqual(summaryItemsAfterValidate.length, 2, 'form has validation summary after validation');
     });
 
+    let screen;
```
ksercs, line 4740: please create a separate module and move this stuff to `beforeEach`. It's really sloppy to me to store variable somewhere between tests. We actually don't wanna to have an access to these variables from other tests

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4737,37 +4737,79 @@ QUnit.module('reset', () => {
...
-                        showClearButton: true,
-                        dataSource: [{ Value: value, Text: text }],
-                    },
-                    editorType: 'dxDropDownBox',
-                },
-            ]
-        });
+        const $form = $('#form').dxForm(formOptions);
+        const $input = $form.find(`.test-ddbox .${EDITOR_INPUT_CLASS}`);
+
```
ksercs, line 4796: nothing is focused => I don't see a reason how value can be lost.
Is input focusing missed here?

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4737,37 +4737,79 @@ QUnit.module('reset', () => {
...
+                                colSpan: 1
+                            },
+                        ]
+                    }
+                ]
+            }
+        ]
+    };
+
     QUnit.test('DropDownBox should not lose its value if form resized (T1196835)', function(assert) {
```
ksercs, line 4791:
```suggestion
    QUnit.test('DropDownBox should not lose its value if form is resized when input is focused (T1196835)', function(assert) {
```

packages/devextreme/js/ui/form/ui.form.js
```diff
@@ -1182,10 +1183,13 @@ const Form = Widget.inherit({
...
     _refresh: function() {
-        const editorSelector = `.${FOCUSED_STATE_CLASS} > :not(.dx-dropdowneditor-input-wrapper) input,`
-            + ` .${FOCUSED_STATE_CLASS} textarea`;
+        const editorSelector = `.${FOCUSED_STATE_CLASS} input, .${FOCUSED_STATE_CLASS} textarea`;
 
-        eventsEngine.trigger(this.$element().find(editorSelector), 'change');
+        this.$element().find(editorSelector).each((idx, el) => {
+            if(!$(el).parents(UNREFRESHABLE_COMPONENT_CLASS).length) {
+                eventsEngine.trigger(el, 'change');
+            }
```
ksercs, line 1191: Unfortunately it seems to me that this solution is not applicable because it will prevent triggering "change" event for cases when direct dropDownEditor is focused.
Scenario: 
- add a SelectBox with `acceptCustomValue=true` to form (i think this issue is actual for ddBox as well btw)
- start typing
- minimize window
- value (as far as i understand is lost)

Please, consider other ddEditors and `acceptCustomValue=true/false` and add tests for these cases

ksercs (changes requested):

PR #26816, opened 2024-03-01
ContextMenu: Close menu when element lose focus

marker-dao:


packages/devextreme/js/ui/context_menu/ui.context_menu.js
```diff
@@ -339,6 +340,21 @@ class ContextMenu extends MenuBase {
         this._addWidgetClass();
     }
 
+    _isTargetOutOfComponent(relatedTarget) {
+        const isInsideContextMenu = $(relatedTarget).closest(`.${DX_CONTEXT_MENU_CLASS}`).length !== 0;
+
+        return !isInsideContextMenu;
```
ksercs, line 346: I suggest u the same approach as in the lost note in Menu PR.
I believe on component recreation focusout does not have a relatedTarget and scenario in FileManager should work.

If I'm right, it will allow us to get rid of overlay private cache using.

marker-dao: [26c0e42ea5a238bd35f2d6f01d639724e1cf39d3]

ksercs (approved):

EugeniyKiyashko (approved):

iBat (approved):

PR #26823, opened 2024-03-04
Menu: Close menu when element lose focus

marker-dao:


packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -760,6 +760,33 @@ QUnit.module('Menu tests', {
         assert.notOk(submenu.option('visible'), 'submenu was closed');
     });
 
+    [true, false].forEach(hideSubmenuOnFocusOut => {
+        QUnit.test(`Close submenu when element lose focus and _hideSubmenuOnFocusOut=${hideSubmenuOnFocusOut}`, function(assert) {
```
Zedwag: let's rename the test, now it sounds like menu should be closed in both scenarios

marker-dao: Thanks, I have renamed the test: [69ee2668a0da82519efd21e28f1b049f847a0cec].

Zedwag (dismissed):

Zedwag (dismissed):

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -269,6 +269,38 @@ class Menu extends MenuBase {
...
+        const $contextMenuTarget = $(relatedTarget).parents(`.${DX_CONTEXT_MENU_CLASS}`)[0];
+
+        if(!$menuTarget && !$contextMenuTarget) {
+            return true;
+        }
+
+        const $element = this._focusTarget()[0];
+        const $overlayContent = this._visibleSubmenu?.getOverlayContent()[0];
+
+        const isTargetOutOfComponent = !($menuTarget === $element || $contextMenuTarget === $overlayContent);
```
ksercs: I don't get why we need this. Seems like it's always false because otherwise we return in 280th. Am I missing smth?

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -269,6 +269,38 @@ class Menu extends MenuBase {
     _render() {
         super._render();
         this._initAdaptivity();
+        this._attachFocusOutHandler();
```
ksercs: why do u attach additional handler, not just use existing ` _focusOutHandler`?

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -760,6 +760,30 @@ QUnit.module('Menu tests', {
...
+
+        const submenu = getSubMenuInstance($subMenu);
+
+        assert.strictEqual(submenu.option('visible'), true, 'submenu opened');
+
+        $(menu.element).trigger('focusout');
+
+        assert.strictEqual(submenu.option('visible'), false, 'submenu closed');
+    });
+
```
ksercs, line 886: We should also test that it does not close if we click inside of menu or submenu

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -760,6 +760,30 @@ QUnit.module('Menu tests', {
         assert.notOk(submenu.option('visible'), 'submenu was closed');
     });
 
+    QUnit.test('Submenu should be closed when element lose focus', function(assert) {
```
ksercs:
```suggestion
    QUnit.test('Submenu should be closed when element loses focus', function(assert) {
```

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -269,6 +269,38 @@ class Menu extends MenuBase {
...
+    }
+
+    _isTargetOutOfComponent(relatedTarget) {
+        const $menuTarget = $(relatedTarget).parents(`.${DX_MENU_CLASS}`)[0];
+        const $contextMenuTarget = $(relatedTarget).parents(`.${DX_CONTEXT_MENU_CLASS}`)[0];
```
ksercs: I think the code will be more intuitive if we write these checks in the following manner:
```javascript
    const isInsideRootMenu =  $(relatedTarget).parents(`.${DX_MENU_CLASS}`).length !== 0;
   ...
```

packages/devextreme/js/ui/menu/ui.menu.js
marker-dao: [3159bd5f6586777260033510f1278c6b4ecf6aea]

packages/devextreme/js/ui/menu/ui.menu.js
marker-dao: [3159bd5f6586777260033510f1278c6b4ecf6aea]

packages/devextreme/js/ui/menu/ui.menu.js
marker-dao: [3159bd5f6586777260033510f1278c6b4ecf6aea]

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
marker-dao: [3159bd5f6586777260033510f1278c6b4ecf6aea]

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
marker-dao: [027c84f5e4273678e72f42ec26de7d4e48fb3814]

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -760,6 +760,65 @@ QUnit.module('Menu tests', {
         assert.notOk(submenu.option('visible'), 'submenu was closed');
     });
 
+    QUnit.test('Submenu should be closed when element loses focus', function(assert) {
+        const options = {
+            focusStateEnabled: true,
```
ksercs, line 765: do we really need it?

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -760,6 +760,65 @@ QUnit.module('Menu tests', {
...
+            showFirstSubmenuMode: 'onClick',
+            items: [
+                { text: 'item 1', items: [{ text: 'item 11' }] },
+                { text: 'item 2', items: [{ text: 'item 21' }] },
+            ],
+        };
+        const menu = createMenu(options);
+        const $menuItem = $(menu.element).find(`.${DX_MENU_ITEM_CLASS}`);
+        const $subMenu = $menuItem.eq(1);
```
ksercs: so, u wanna say that in the previous row u get `$menuItem`,  but here with `.eq(1)` u get already a submenu?

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -271,6 +271,23 @@ class Menu extends MenuBase {
         this._initAdaptivity();
     }
 
+    _isTargetOutOfComponent(relatedTarget) {
+        const isInsideRootMenu = $(relatedTarget).parents(`.${DX_MENU_CLASS}`).length !== 0;
```
ksercs: this functionality is not tested

ksercs (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -760,6 +760,65 @@ QUnit.module('Menu tests', {
...
+        const submenu = getSubMenuInstance($subMenu);
+
+        assert.strictEqual(submenu.option('visible'), true, 'submenu opened');
+
+        $(menu.element).trigger('focusout');
+
+        assert.strictEqual(submenu.option('visible'), false, 'submenu closed');
+    });
+
+    QUnit.test('Submenu should not be closed when submenu item clicked', function(assert) {
```
ksercs: this test is evergreen

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
marker-dao: For testing on mobile devices - yes.

packages/devextreme/js/ui/menu/ui.menu.js
marker-dao: [d90a6b7f24312a9901977c1207f33c86766ffe5c]

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
marker-dao: [d90a6b7f24312a9901977c1207f33c86766ffe5c]

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
marker-dao: [d90a6b7f24312a9901977c1207f33c86766ffe5c]

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -271,6 +271,23 @@ class Menu extends MenuBase {
         this._initAdaptivity();
     }
 
+    _isTargetOutOfComponent(relatedTarget) {
+        const isInsideRootMenu = $(relatedTarget).closest(`.${DX_MENU_CLASS}`).length !== 0;
+        const isInsideContextMenu = $(relatedTarget).closest(`.${DX_CONTEXT_MENU_CLASS}`).length !== 0;
+
+        const isTargetOutOfComponent = !(isInsideRootMenu || isInsideContextMenu);
+
+        return isTargetOutOfComponent;
```
ksercs, line 280: We missed the scenario when disabled item is clicked. In this case `relatedTarget=null` and submenu will be badly closed.

I suggest you to add a check that related target is specified (like it's done for [overlay outside click](https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme/js/ui/overlay/ui.overlay.js#L295) for example). If relatedTarget=null we don't need to close a submenu because no element on the page is focused => a11y rule cannot be broken.

marker-dao: You're leaving out a few scenarios: when the user of a component intentionally triggers a focusout (why shouldn't we close the menu in that case?) and, for example, when the user focuses the address bar or something outside the page (also a debatable case, e.g. our DD overlays close if the focus leaves the page, why shouldn't the Menu overlay close then? Here's an example: https://codepen.io/marker-dao/pen/YzMZdXd?editors=1010. Based on this, I decided to still close Menu when relatedTarget === null).

We need to come up with an additional solution for the case with disabled items. I suggest to look for a solution in DDEditors.

I need more time to research case with new conditions.

marker-dao: We investigated all the cases in more detail and decided not to close the menu if !relatedTarget.

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -760,6 +761,138 @@ QUnit.module('Menu tests', {
         assert.notOk(submenu.option('visible'), 'submenu was closed');
     });
 
+    QUnit.test('Submenu should be closed when element loses focus', function(assert) {
+        const simpleMenu = $('#simpleMenu');
```
ksercs:
```suggestion
        const $simpleMenu = $('#simpleMenu');
```

ksercs (approved):

EugeniyKiyashko (approved):

EugeniyKiyashko (approved):

iBat (approved):

PR #26848, opened 2024-03-06
Splitter: add sizes distribution logic

AlexanderMoiseev:


packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -75,7 +76,11 @@ class Splitter extends (CollectionWidget as any) {
 
     this._toggleOrientationClass();
 
-    this._layout = getInitialLayout(this.option('items'));
+    const items = this.option('items');
+
+    const elementSize = getElementSize(this.$element(), items, this.option('orientation'), this.option('width'), this.option('height'));
```
EugeniyKiyashko: This code is not working in case: 
- Splitter is located inside TabPanel Tab's content
- Templates are rendered asynchronously
- Perhaps, after Render Splitter in invisible container

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -233,6 +238,10 @@ class Splitter extends (CollectionWidget as any) {
   }
 
   _itemOptionChanged(item: unknown, property: unknown, value: unknown): void {
+    if (property === 'size') {
```
EugeniyKiyashko: We shouldn't do full rerender of component after change size option.

packages/devextreme/js/__internal/ui/splitter/utils/layout.ts
```diff
@@ -123,14 +124,106 @@ export function updateItemsSize(items, sizeDistribution): void {
...
+  const size: number = orientation === ORIENTATION.horizontal
+    ? splitterSize.width : splitterSize.height;
+
+  const handlesSizeSum = handlesCount * DEFAULT_RESIZE_HANDLE_SIZE;
+
+  return size - handlesSizeSum;
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export function getElementSize($element, items, orientation, width, height): number {
```
EugeniyKiyashko, line 230: "If Pane.size is set to '10%', it should represent 10% of the Splitter's total width. It seems that in your case, it's 10% of the total width of the panes, excluding the width of the resize handles. Please check this and decide what way is right.

Zedwag (dismissed):

EugeniyKiyashko (approved):

Zedwag (approved):

PR #26879, opened 2024-03-11
Implement launch of testcafe tests in shadowDOM

Alyar666:


alexanderPolosatov (approved):

packages/devextreme/testing/testcafe/runner.js
```diff
@@ -16,6 +16,22 @@ const changeTheme = async(themeName) => createTestCafe.ClientFunction(() => new
...
+    await createTestCafe.ClientFunction(() => {
+        const root = document.querySelector('#parentContainer');
+        const childNodes = root.childNodes;
+
+        if(!root.shadowRoot) {
+            root.attachShadow({ mode: 'open' });
+        }
+
+        const shadowContainer = document.createElement('div');
+        shadowContainer.append.apply(shadowContainer, Array.from(childNodes));
```
Raushen, line 29: How about just shadowContainer.append(Array.from(childNodes)); ?

Alyar666: The [append](https://developer.mozilla.org/en-US/docs/Web/API/Element/append) method does not support array as argument

Raushen (approved):

PR #26885, opened 2024-03-12
TextArea: Bug Fix content exceed border when autoresize is true and max height is set (T1221869)

nikkithelegendarypokemonster:


packages/devextreme/scss/widgets/base/textArea/_index.scss
```diff
@@ -1,8 +1,10 @@
 // adduse
-
+.dx-texteditor-label{
+  margin-top: 3px;
+}
 .dx-textarea {
   height: auto;
-
+  overflow-y: hidden;
```
EugeniyKiyashko: Why we should add this property?

nikkithelegendarypokemonster: i have put that as a fix to the issue on the content exceeding the border of the parent element.

packages/devextreme/scss/widgets/base/textArea/_index.scss
```diff
@@ -1,8 +1,10 @@
 // adduse
-
+.dx-texteditor-label{
```
EugeniyKiyashko: We shouldn't add fixed values to the base styles since all themes use these styles. This is only acceptable if these styles are required in all themes. In your case, the problem occurs only with the material theme.

packages/devextreme/scss/widgets/base/textArea/_index.scss
nikkithelegendarypokemonster: upon inspecting when maxheight is defined it sets the height of the parent element disregarding the height of the child text area thats why the content in the text area exceeds the the parent element. this happens if the max height is less than the actual height of the input value inside the text area element...

packages/devextreme/js/ui/text_area.js
```diff
@@ -205,7 +204,6 @@ const TextArea = TextBox.inherit({
         if(maxHeight !== undefined) {
             const adjustedMaxHeight = maxHeight - heightDifference;
             const needScroll = inputHeight > adjustedMaxHeight;
-
```
EugeniyKiyashko: Let's remove your changes with removing empty lines. I think that it doesn't matter in your current task.

packages/devextreme/scss/widgets/base/textArea/_index.scss
```diff
@@ -1,8 +1,6 @@
 // adduse
-
```
EugeniyKiyashko: This too.

packages/devextreme/scss/widgets/material/textArea/_index.scss
```diff
@@ -10,6 +10,10 @@
 
 // adduse
 
+.dx-textarea {
```
EugeniyKiyashko: This style is better place after mixin 'dx-textarea(' call. Inside mixin you can have the same selector `.dx-textarea {}` and in this case it's will more priority that yours.

packages/devextreme/testing/testcafe/tests/editors/textArea/index.ts
```diff
@@ -1,101 +1,101 @@
...
-};
-
-testFixture()`TextArea_Height`
-  .page(url(__dirname, '../../container.html'));
+fixture.disablePageReloads`TextArea_Height`
+  .page(url(__dirname, '../../container.html'))
+  .afterEach(async () => clearTestPage());
 
 const text = 'Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry\'s standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged.';
+const longText = 'Prepare 2013 Marketing Plan: We need to double revenues in 2013 and our marketing strategy is going to be key here. R&D is improving existing products and creating new products so we can deliver great AV equipment to our customers.Robert, please make certain to create a PowerPoint presentation for the members of the executive team.';
```
EugeniyKiyashko: Why we need have 'longText'. Seems, that 'text' suits for us testing purpose.

nikkithelegendarypokemonster: i agree sir let me remove it...

packages/devextreme/testing/testcafe/tests/editors/textArea/index.ts
```diff
@@ -1,101 +1,101 @@
...
+    ...config,
+  }, '#textArea1');
+
+  await createWidget('dxTextArea', {
+    ...config,
+    value: text + text,
+  }, '#textArea2');
+});
+
+test('Bug Fix (T1221869) when there is maxHeight and autoResizeEnabled', async (t) => {
```
EugeniyKiyashko: Usually the bug ID we placed in the end of test name. It's better if we call this test as following:

`Height of TextArea input should have the correct height when the maxHeight option is set to 80px`.

The screenshot name:
`TextArea appearance when maxHeight='80px'.png'

packages/devextreme/testing/testcafe/tests/editors/textArea/index.ts
```diff
@@ -1,101 +1,101 @@
...
+  await t
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+}).before(async () => {
+  await setAttribute('#container', 'style', 'width: 300px; height: 400px;');
+  const config = {
+    value: longText,
+    width: 1000,
+    maxHeight: 80,
+    inputAttr: { 'aria-label': 'Notes' },
```
EugeniyKiyashko: In this case, we are not testing this configuration and it does not affect our scenario, so it can be omitted in the test.

packages/devextreme/testing/testcafe/tests/editors/textArea/index.ts
```diff
@@ -1,101 +1,101 @@
...
-  }).before(async () => createWidget('dxTextArea', {
-    maxHeight: '7em',
-    height: '5em',
-    autoResizeEnabled,
-    width: 200,
+    width: '100%',
     value: text,
-  }));
-
-  test(`TextArea has correct height with "autoResizeEnabled" is ${autoResizeEnabled} and maxHeight option is 5em`, async (t) => {
```
EugeniyKiyashko, line 60: We should not have deleted scenarios that were tested by these tests. We needed to preserve them, changing only the testting method to checking through screenshot testing.

packages/devextreme/scss/widgets/base/textArea/_index.scss
nikkithelegendarypokemonster: this change sir is going to cause some lint errors.

packages/devextreme/testing/testcafe/tests/editors/textArea/index.ts
```diff
@@ -1,101 +1,101 @@
...
+    autoResizeEnabled: true,
+  }, '#textArea1');
+
+  await createWidget('dxTextArea', {
+    ...config,
+    autoResizeEnabled: false,
+  }, '#textArea2');
+});
+
+test('TextArea has correct height autoResizeEnabled and different value length', async (t) => {
```
EugeniyKiyashko: Since the autoResizeEnabled option cannot be configured outside of the test and we are testing both configurations in a single screenshot, we do not need to specify 'with autoResizeEnabled' in the test name.

```
test('TextArea should have correct height ... ', async (t) => {
```

packages/devextreme/js/ui/text_area.js
nikkithelegendarypokemonster: Correction: [Commit](https://github.com/DevExpress/DevExtreme/pull/26885/commits/8c3cd0d674579d15406941118b5f92d97ddecb40)

packages/devextreme/scss/widgets/base/textArea/_index.scss
nikkithelegendarypokemonster: Correction: [Commit](https://github.com/DevExpress/DevExtreme/pull/26885/commits/8c3cd0d674579d15406941118b5f92d97ddecb40)

packages/devextreme/scss/widgets/material/textArea/_index.scss
nikkithelegendarypokemonster: Correction: [Commit](https://github.com/DevExpress/DevExtreme/pull/26885/commits/8c3cd0d674579d15406941118b5f92d97ddecb40)

EugeniyKiyashko (approved):

Zedwag (approved):

PR #26891, opened 2024-03-12
ContextMenu Share: Create TestCafe test for ContextMenu Share

jdvictoria:


jdvictoria:
[WG Tests - ContextMenu](https://trello.com/c/DMHdkU3G/8597-wg-tests-contextmenu)

apps/demos/testing/widgets/contextmenu/ContextMenuShare.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+const IMAGE_ID = 'image';
+
+runManualTest('ContextMenu', 'Templates', ['jQuery', 'React', 'Vue', 'Angular'], (test) => {
+  test('Custom Template Share Appearance', async (t) => {
+    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+    await t.rightClick(`#${IMAGE_ID}`);
+
+    await t.click($('.dx-menu-item-text').withText('Share'));
```
marker-dao: Let's create a variable for .dx-menu-item-text class.

apps/demos/testing/widgets/contextmenu/ContextMenuShare.test.js
```diff
@@ -0,0 +1,25 @@
+import { createScreenshotsComparer } from 'devextreme-screenshot-comparer';
+import { Selector as $ } from 'testcafe';
+import { runManualTest } from '../../../utils/visual-tests/matrix-test-helper';
+
+fixture('ContextMenu.Share')
+  .page('http://localhost:8080/')
+  .beforeEach(async (t) => {
```
EugeniyKiyashko: We use the following method to notify the fixture update window size for tests
```
.before(async (ctx) => {
    ctx.initialWindowSize = [900, 600];
  });
```
Your approach it will not work in CI, because each test that is running inside runManualTest has before() hook. Test before.hook() has more priority than fixture.beforeEach() hook according to testcafe documentation.

EugeniyKiyashko: You can find this approach in neighboring files.

apps/demos/testing/widgets/contextmenu/ContextMenuShare.test.js
```diff
@@ -0,0 +1,25 @@
+import { createScreenshotsComparer } from 'devextreme-screenshot-comparer';
+import { Selector as $ } from 'testcafe';
+import { runManualTest } from '../../../utils/visual-tests/matrix-test-helper';
+
+fixture('ContextMenu.Share')
```
EugeniyKiyashko: It's better to name the fixture according to the demo name that you are testing. You test ContextMenu component, the name of demo is Templates. So, the fixture name should be 'ContextMenu.Templates'

ksercs (approved):

EugeniyKiyashko (approved):

PR #26902, opened 2024-03-13
Splitter: add pane resizable option

Zedwag:


packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
```diff
@@ -172,6 +172,53 @@ QUnit.module('Resizing', moduleConfig, () => {
...
+
+        const items = this.$element.children(`.${SPLITTER_ITEM_CLASS}`);
+
+        const pointer = pointerMock(this.getResizeHandles().eq(0));
+        pointer.start().dragStart().drag(50, 50).dragEnd();
+
+        assertLayout(items, ['75', '25'], assert);
+    });
+
+    QUnit.test('resize should not work when pane resizable is disabled on runtime', function(assert) {
```
EugeniyKiyashko, line 207: Also we should the second scenario when 'items[1].resizable' is false.

EugeniyKiyashko (approved):

PR #26906, opened 2024-03-13
🛠️ Refactoring | DataGrid native classes refactoring

Raushen:


packages/devextreme/js/__internal/grids/data_grid/export/m_export.ts
```diff
@@ -593,23 +610,21 @@ export class ExportController extends dataGridCore.ViewController {
   }
 
   private _getSummaryCells(summaryTotalItems, totalAggregates) {
-    const dataController = this.getController('data');
+    const dataController = this._dataController as any;
```
williamvinogradov: Minor comment

I think it's better to remove the ```any``` and disable TS error below with ```@ts-expect-error```

williamvinogradov (dismissed):

williamvinogradov (approved):

PR #26911, opened 2024-03-13
Fix performance problem in data arrayQuery

GoodDayForSurf:


packages/devextreme/js/data/array_query.js
```diff
@@ -243,7 +244,36 @@ const compileCriteria = (function() {
...
+            }
+            return acc;
+        }, []);
+
+        return (obj) => {
+            const value = _toComparable(getter(obj));
+            let result = false;
+
+            filterValues.find((filterValue) => {
+                result = useStrictComparison(filterValue)
```
VasilyStrelyaev: This looks a little weird that a predicate mutates a variable from closure.

Does the following code do the same?

```js
return filterValues.some((filterValue) => {
    return useStrictComparison(filterValue)
        ? value === filterValue
        // eslint-disable-next-line eqeqeq
        : value == filterValue;
});
```

packages/devextreme/js/data/utils.js
```diff
@@ -232,6 +232,15 @@ const isGroupOperator = function(value) {
     return value === 'and' || value === 'or';
 };
 
+export const isUniformEqualsByOr = function(crit) {
```
VasilyStrelyaev, line 235: I feel like, ideally, we need separate unit tests for this method to make sure it rejects all the non-uniform criteria. It's the most critical point here where we can avoid (or not avoid) BCs.

Like for cases with (1) different properties (2) different operands, other than or (3) different criteria, other than = (4) more depth

packages/devextreme/js/data/array_query.js
GoodDayForSurf: I agree... it seems  .some() is better. thx

packages/devextreme/js/data/utils.js
GoodDayForSurf: ok.. i will add test for it

VasilyStrelyaev (approved):

PR #26921, opened 2024-03-14
FileUploader: Fix Custom Drop Zone Multiple Files and Incorrect File Extension Handling

jdvictoria:


apps/demos/Demos/FileUploader/CustomDropzone/jQuery/index.js
```diff
@@ -8,7 +8,16 @@ $(() => {
     uploadUrl: 'https://js.devexpress.com/Demos/NetCore/FileUploader/Upload',
     visible: false,
     onDropZoneEnter(e) {
-      if (e.dropZoneElement.id === 'dropzone-external') { toggleDropZoneActive(e.dropZoneElement, true); }
+      if (e.dropZoneElement.id === 'dropzone-external') {
+        const items = e.event.originalEvent.dataTransfer.items;
+
+        const isSingleFileDragged = items.length === 1;
+        const isValidFileExtension = `.${items[0].type.replace(/^image\//, '')}`;
```
ksercs: `is` prefix in the name says that this variable is a predicate which has true/false as a value, but use use a string as a value. Let's update the naming
```suggestion
        const draggedFileExtension = `.${items[0].type.replace(/^image\//, '')}`;
```

apps/demos/Demos/FileUploader/CustomDropzone/jQuery/index.js
```diff
@@ -8,7 +8,16 @@ $(() => {
...
+
+        if (isSingleFileDragged && this.option('allowedFileExtensions').includes(isValidFileExtension)) {
```
ksercs: We can create additional variables to make the code clearer.
It's also better to use `e.component` to get an instance because this api is available in all frameworks
```javascript
const allowedFileExtensions = this.option('allowedFileExtensions');
const isValidFileExtension = allowedFileExtensions.includes(draggedFileExtension);
```

I also suggest u splitting variables to 2 blocks:

```javascript
->additional variables<-

->meaningful variables which will be used in `if`<-
```

apps/demos/Demos/FileUploader/CustomDropzone/jQuery/index.js
```diff
@@ -8,7 +8,21 @@ $(() => {
     uploadUrl: 'https://js.devexpress.com/Demos/NetCore/FileUploader/Upload',
     visible: false,
     onDropZoneEnter(e) {
```
ksercs: it's possible to use destructuring here
```suggestion
    onDropZoneEnter({ component, dropZoneElement, event }) {
```

apps/demos/Demos/FileUploader/CustomDropzone/Angular/app/app.component.ts
```diff
@@ -24,6 +24,8 @@ export class AppComponent {
 
   progressValue = 0;
 
+  allowedFileExtensions: string[] = ['.jpg', '.jpeg', '.gif', '.png'];
```
ksercs, line 27: if we create an instance property here, let's also use it in markup `app.component.html`

jdvictoria:
[FileUploader - Update the Custom Drop Zone demo to resolve the issue with uplaoding multiple files](https://trello.com/c/BlWCKB8E/8596-fileuploader-update-the-custom-drop-zone-demo-to-resolve-the-issue-with-uplaoding-multiple-files)

apps/demos/testing/widgets/fileuploader/CustomDropzone.test.js
```diff
@@ -0,0 +1,103 @@
...
+
+    await testScreenshot(t, takeScreenshot, 'custom_dropzone_invalid_format.png');
+
+    await t
+      .expect(compareResults.isValid())
+      .ok(compareResults.errorMessages());
+  });
+});
+
+runManualTest('FileUploader', 'CustomDropzone', ['jQuery'], (test) => {
```
ksercs: I suggest u writing these 3 tests in the single `runManualTest`

apps/demos/testing/widgets/fileuploader/CustomDropzone.test.js
```diff
@@ -0,0 +1,103 @@
...
+          },
+          clientX: $dropZone.offset().left,
+          clientY: $dropZone.offset().top,
+        }),
+      }));
+    }, { dependencies: { DROPZONE_EXTERNAL_CLASS } })();
+
+    await t.expect(Selector(`#${DROPZONE_EXTERNAL_CLASS}`).hasClass('dropzone-active')).notOk();
+
+    await testScreenshot(t, takeScreenshot, 'custom_dropzone_multiple_files.png');
```
ksercs: Actually these screenshots for "invalid" cases are useless because there is no UI difference with a default UI tested in other main test. Please, leave only screenshot in the first test

apps/demos/testing/widgets/fileuploader/CustomDropzone.test.js
```diff
@@ -0,0 +1,101 @@
...
+            items: [
+              { kind: 'file', type: 'image/png' },
+              { kind: 'file', type: 'image/png' },
+            ],
+            types: ['Files'],
+          },
+          clientX: $dropZone.offset().left,
+          clientY: $dropZone.offset().top,
+        }),
+      }));
```
ksercs: Let's move event triggering to a method to reduce copypaste? We only need pass `items` as a parameter

apps/demos/testing/widgets/fileuploader/CustomDropzone.test.js
```diff
@@ -0,0 +1,62 @@
...
+const DROPZONE_EXTERNAL_CLASS = 'dropzone-external';
+
+fixture('FileUploader.CustomDropzone')
+  .page('http://localhost:8080/')
+  .before(async (ctx) => {
+    ctx.initialWindowSize = [900, 600];
+  });
+
+runManualTest('FileUploader', 'CustomDropzone', ['jQuery'], (test) => {
+  const triggerDragEnter = async (t, dropZoneSelector, fileType) => {
```
ksercs: Why do we need `t` and `dropZoneSelector` here?

apps/demos/testing/widgets/fileuploader/CustomDropzone.test.js
```diff
@@ -0,0 +1,62 @@
...
+    await t.expect(Selector(`#${DROPZONE_EXTERNAL_CLASS}`).hasClass('dropzone-active')).ok();
+  });
+
+  test('dropzone-active class is not added to the dropzone element when an invalid file format is dragged', async (t) => {
+    await triggerDragEnter(`#${DROPZONE_EXTERNAL_CLASS}`, 'image/xlsx');
+
+    await t.expect(Selector(`#${DROPZONE_EXTERNAL_CLASS}`).hasClass('dropzone-active')).notOk();
+  });
+
+  test('dropzone-active class is not added to the dropzone element when multiple items are dragged', async (t) => {
```
ksercs, line 45: this test is not working now as far as i see

ksercs (dismissed):

AlexanderMoiseev (dismissed):

apps/demos/testing/widgets/fileuploader/CustomDropzone.test.js
```diff
@@ -0,0 +1,63 @@
+import { createScreenshotsComparer } from 'devextreme-screenshot-comparer';
+import { Selector, ClientFunction } from 'testcafe';
+import { runManualTest } from '../../../utils/visual-tests/matrix-test-helper';
+import { testScreenshot } from '../../../utils/visual-tests/helpers/theme-utils';
+
+const DROPZONE_EXTERNAL_CLASS = 'dropzone-external';
```
iBat: As I see this is not a CSS class but ID of an HTML element. Can we change the name of variable depending on the kind of value it contains?

jdvictoria: Fixed

iBat (approved):

AlexanderMoiseev (approved):

PR #26941, opened 2024-03-18
Monorepo - move out testcafe tests as package

pomahtri:
review of @DevExpress/devextreme-devops:

- [x] @IlyaKhD 
- [x] @iBat 
- [x] @alexslavr


IlyaKhD:
A bunch of small questions:
1. `'e2e/testcafe-devextreme'` - shouldn't it tave project.json ? Looks like default step for creating projects, even if there is no need to configre targets at this monent
2. `'e2e/qunit/tests/Renovation'` - can it be renamed to lower case? (`'e2e/qunit/tests/renovation'`)
3. Why do `'e2e/bundlers'` and `'e2e/qunit'` extracted as projects? They have no scripts/targets
4. I suggest renaming `'testcafe-devextreme'` to `'devextreme-testcafe'`. Or renaming `'apps/react-storybook'` to `'apps/react-storybook'` - we should follow one pattern, either `PACKAGE-KIND` or `KIND-PACKAGE`

pomahtri:
@IlyaKhD 
> 'e2e/testcafe-devextreme' - shouldn't it tave project.json ? Looks like default step for creating projects, even if there is no need to configre targets at this monent

I didn't create project.json files unless there's need to configure targets. It helps avoiding duplicating (for example for 'test' target you need to write npm script, write executor that calls npm script).
Absence of this file can quickly show that all targets in this project have default settings

I don't insist tho, I will change if I didn't convince you

> 'e2e/qunit/tests/Renovation' - can it be renamed to lower case? ('e2e/qunit/tests/renovation')

This PR is only about testcafe, if you saw qunit here, please show the line so I could remove it
About renaming it when moving qunit - let's do it separately

> Why do 'e2e/bundlers' and 'e2e/qunit' extracted as projects? They have no scripts/targets

In this PR there should be no bundlers or qunit.
In later PRs:

- they do have scripts (`test` at least)
- they are e2e test, which means they test built package, which means they cannot be part of this package
- they have additional dependencies, which we will be able to remove from `devextreme` (c# in qunit, rollup/webpack/parcel/vite in bundlers)
- bundlers are already separate package (it has package.json), which means it's already separate project

> I suggest renaming 'testcafe-devextreme' to 'devextreme-testcafe'. Or renaming 'apps/react-storybook' to 'apps/react-storybook' - we should follow one pattern, either PACKAGE-KIND or KIND-PACKAGE

Let's rename storybook later

I want it to be 'testcafe-devextreme' because later there will be 'testcafe-demos'. It will be nice if they will be near in IDE folder navigation

pomahtri:
> Why do 'e2e/bundlers' and 'e2e/qunit' extracted as projects? They have no scripts/targets

upd: found & removed

IlyaKhD (dismissed):

alexslavr (approved):

williamvinogradov (approved):

PR #26944, opened 2024-03-18
Monorepo - add default workflow

pomahtri:

### DIFF
--- a/.github/workflows/default_workflow.yml
+++ b/.github/workflows/default_workflow.yml
@@ -1,0 +1,52 @@
+name: Default Workflow
+
+concurrency:
+  group: wf-${{github.event.pull_request.number || github.sha}}-${{github.workflow}}
+  cancel-in-progress: true
+
+on:
+  pull_request:
+  push:
+    branches:
+      - "[0-9][0-9]_[0-9]"
+
+env:
+  NX_SKIP_NX_CACHE: ${{ (github.event_name != 'pull_request' || contains( github.event.pull_request.labels.*.name, 'skip-cache')) && 'true' || 'false' }}
+
+jobs:
+  main:
+    name: main
+    runs-on: devextreme-shr2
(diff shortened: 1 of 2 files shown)


.github/workflows/default_workflow.yml
```diff
@@ -0,0 +1,52 @@
...
+      uses: actions/cache@v4
+      with:
+        path: '**/node_modules'
+        key: ${{ runner.os }}-node-modules-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: ${{ runner.os }}-node-modules
+
+    - name: Run npm install
+      run: npm ci --no-audit --no-fund
+    
+    - name: Run targets
```
IlyaKhD, line 36: **Minor**
```suggestion
    - name: Test
```

pomahtri: In future there will be different targets too, such as lint, compile-ts and so on

IlyaKhD: Why not extracting each target's `run-many` as a workflow step? Looks much more convenient

pomahtri: If we specify them in one `nx` command, they will run in parallel
If in separate step, each target will wait for previous to finish (because next step of workflow doesn't start until previous is finished)

pomahtri: It's nx's guideline to specify all commands in one shell command to reduce unnecessary waits (no need to wait for build to lint; no need to build qunit to run testcafe etc)
 https://nx.dev/ci/concepts/parallelization-distribution

iBat (approved):

.github/workflows/default_workflow.yml
IlyaKhD: Got it. Thank you for the explantion!

PR #26951, opened 2024-03-19
React - File Uploader component ref fix (T1183211)

ivanblinov2k17:
Re-fix: https://github.com/DevExpress/DevExtreme/pull/25420

### DIFF
--- a/packages/devextreme-react/src/core/__tests__/test-component.ts
+++ b/packages/devextreme-react/src/core/__tests__/test-component.ts
@@ -36,6 +36,16 @@ class TestComponent<P = any> extends Component<P & IHtmlOptions> {
 
   protected useDeferUpdateFlag = true;
 
+  get element(): HTMLElement {
+    return this._element;
+  }
+
+  get instance(): any {
+    return {
+      element: () => this.element,
+    };
+  }
+
   _createWidget(element?: Element): void {
     eventHandlers.optionChanged = [];
     Widget.option.mockImplementation((name: string) => name === 'integrationOptions.useDeferUpdateForTemplates');
--- a/packages/devextreme-react/src/core/widget-config.ts
+++ b/packages/devextreme-react/src/core/widget-config.ts
@@ -8,6 +8,16 @@ function isIgnoredProp(name: string) {
   return name === 'children' || name === classNamePropName || elementPropNames.includes(name);
 }
 
+function getRefElement(value: any): HTMLElement {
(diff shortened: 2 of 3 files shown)


packages/devextreme-react/src/core/widget-config.ts
```diff
@@ -8,6 +8,16 @@ function isIgnoredProp(name: string) {
   return name === 'children' || name === classNamePropName || elementPropNames.includes(name);
 }
 
+function getRefElement(value: any): HTMLElement {
```
dxvladislavvolkov, line 11: Can we use something else instead of any type?

ivanblinov2k17: I tried something like Ref< HtmlElement > | HtmlElement | Ref < MockComponent >
There are some difficulties with creating MockComponent type and type casting

ivanblinov2k17: instance getter is specified in generated react component
also we need type guards to differentiate between those types
so the code will be a little more complicated 
i'll prepare a solution and post it in channel and then we decide

packages/devextreme-react/src/core/__tests__/test-component.ts
```diff
@@ -36,6 +36,16 @@ class TestComponent<P = any> extends Component<P & IHtmlOptions> {
 
   protected useDeferUpdateFlag = true;
 
+  get element(): HTMLElement {
```
VasilyStrelyaev, line 39: Is this redundant since you now expose `element` via `instance.element()`?

packages/devextreme-react/src/core/widget-config.ts
VasilyStrelyaev: This `value` is a property value or a ref, so it may be `Ref<...> | any` which is effectively `any`.

See how the `props`  argument of `separateProps` is typed as `Record<string, any>`.

So I agree with Ivan, we may better leave types as they are...

packages/devextreme-react/src/core/__tests__/test-component.ts
ivanblinov2k17: Yeah, probably, but i use it in instance getter 
Maybe it might be useful in other tests

VasilyStrelyaev (approved):

dxvladislavvolkov (approved):

PR #26957, opened 2024-03-19
Monorepo - move out typescript e2e package

pomahtri:
Review from @DevExpress/devextreme-devops:
- [x] @alexslavr 
- [ ] @iBat 
- [x] @IlyaKhD


e2e/typescript/license_validation.ts
```diff
@@ -0,0 +1,8 @@
+import { assertType, toAssertion } from './consts';
```
IlyaKhD: I guess git gonna loose history of this file.
Could you try `git mv`? I'm not sure it will survive squaching, but we should try

e2e/typescript/package.json
```diff
@@ -1,10 +1,14 @@
...
+  "name": "devextreme-typescript-tests",
   "version": "1.1.0",
   "author": "Developer Express Inc.",
+  "scripts": {
+    "test": "tsc --noEmit"
+  },
   "dependencies": {
-    "@angular/common": "^11.2.14"
+    "@angular/common": "^11.2.14",
+    "devextreme-main": "*"
```
IlyaKhD: Can this be moved to dev-deps? This package should never be published

e2e/typescript/package.json
```diff
@@ -1,10 +1,14 @@
 {
   "private": true,
-  "name": "tests",
+  "name": "devextreme-typescript-tests",
```
IlyaKhD: Let's rename it, here are some options:

- `devextreme-compilation-cases`
- `devextreme-compilation-tests`
- `devextreme-api-compilation`
- `devextreme-api-compilation-cases`
- `devextreme-api-tests`

How do you think?

e2e/typescript/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "esModuleInterop": true,
+    "moduleResolution": "node",
+    "noEmit": true,
+    "skipLibCheck": true,
+    "typeRoots": [],
+    "rootDir": "../../",
+    "target": "es2015",
```
IlyaKhD:
```suggestion
    "target": "ES2020",
```

e2e/typescript/license_validation.ts
pomahtri: It's done with `git mv`, you can see 'move' commit. Before squash, git can blame this file properly, not sure it will do after (but probably it will be able). I don't think it's that important to save history for this file tho. Anyways, it will be possible to trace history to this PR and from it further

e2e/typescript/package.json
pomahtri: In my point of view, it will be more logical to move them all to common deps as they are all needed to run this project (and not run some side tasks like lint), but since package is never published, it doesn't matter
I'll move it to dev deps

e2e/typescript/package.json
pomahtri: Will rename to `devextreme-compilation-cases`, but what about folder name then?

e2e/typescript/tsconfig.json
pomahtri: done

e2e/typescript/package.json
IlyaKhD: It's better to rename it as well then

e2e/typescript/license_validation.ts
IlyaKhD: Agree

e2e/typescript/package.json
IlyaKhD: like 'e2e/compilation-cases'

e2e/typescript/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "noEmit": true,
+    "skipLibCheck": true,
+    "typeRoots": [],
+    "rootDir": "../../",
+    "target": "ES2020",
+    "lib": [
+        "ES2017",
+        "ES2020.BigInt",
+        "DOM"
+    ],
```
IlyaKhD: I think the `lib` field can be removed. If not, then changed like this:

```suggestion
    "lib": [
        "ES2020",
        "DOM"
    ],
```

pomahtri: It cannot be removed
Base tsconfig has `"lib": ["DOM"]` (i don't know why), so these errors occur: 
image

e2e/typescript/package.json
pomahtri: done

e2e/typescript/tsconfig.json
pomahtri: changed to suggested

IlyaKhD: thnx

IlyaKhD (approved):

e2e/typescript/tsconfig.json
IlyaKhD: @VasilyStrelyaev, JFYI, same issue we've encountered previously

PR #26994, opened 2024-03-25
TreeList: add job for testcafe tests in shadowDom

Alyar666:


packages/devextreme/testing/testcafe/helpers/accessibility/utils.ts
```diff
@@ -29,7 +29,13 @@ export const a11yCheck = async (
   configuration = {},
 ):
 Promise<void> => {
-  const { error, results } = await axeCheck(t, selector, { rules: {}, ...options });
+  let context: ElementContext | null = null;
+
+  if (process.env.shadowDom === 'true' && typeof selector === 'string') {
```
pomahtri, line 34: minor: let create file for envs config
like, env.ts:
```
export const env = {
  shadowDom: process.env.shadowDom === 'true'
};

File: packages/devextreme/testing/testcafe/helpers/shadowDom/shadowDomExtension.js [hunk 2]
```diff
@@ -7,6 +7,22 @@ function get(selector) {
...
+    const childNodes = root.childNodes;
+
+    if(root.shadowRoot) {
+        return;
+    }
+
+    root.attachShadow({ mode: 'open' });
+
+    const shadowContainer = document.createElement('div');
+    shadowContainer.append.apply(shadowContainer, Array.from(childNodes));
```
pomahtri, line 21: minor
```suggestion
    shadowContainer.append(...childNodes);
```

packages/devextreme/testing/testcafe/model/dataGrid/index.ts
```diff
@@ -160,7 +161,14 @@ export default class DataGrid extends Widget {
   }
 
   getFocusedRow(): Selector {
-    return this.dataRows.filter(`.${CLASS.focusedRow}`);
+    const { focusedRow } = CLASS;
+
+    return this.dataRows.filter(
```
pomahtri, line 166: wrote about it personally

packages/devextreme/testing/testcafe/model/internal/widget.ts
```diff
@@ -1,18 +1,24 @@
...
     () => {
       const $widgetElement = $(selector());
       const elementData = $widgetElement.data();
-      const widgetNames = elementData.dxComponents;
+      const widgetNames = elementData?.dxComponents ?? [];
 
       if (name) {
+        if (isShadowDom) {
+          return (window as any).DevExpress.ui[name].getInstance($widgetElement.get(0));
+        }
```
pomahtri, line 21: Looks like logic inside `if (isShadowDom)` should work for non-shadow mode too. If so, let's leave only this logic
Btw, why `elementData` doesn't work? It's totally not working inside shadow dom?

packages/devextreme/testing/testcafe/runner.js
```diff
@@ -18,17 +18,8 @@ const changeTheme = async(themeName) => createTestCafe.ClientFunction(() => new
...
-        if(!root.shadowRoot) {
-            root.attachShadow({ mode: 'open' });
-        }
-
-        const shadowContainer = document.createElement('div');
-        shadowContainer.append.apply(shadowContainer, Array.from(childNodes));
-
-        root.shadowRoot.appendChild(shadowContainer);
+        // eslint-disable-next-line no-undef
+        window.addShadowRootTree();
```
pomahtri, line 22: let's make `addShadowRootTree` clientFunction? Now it's js script, but if it will be clientFunction, it can be written in TS
You still can have separate file for it

pomahtri (approved):

packages/devextreme/testing/testcafe/helpers/shadowDom/shadowDomExtension.js
pomahtri: discussed, there's some error with this construction, will refactor in future for all places like this

packages/devextreme/testing/testcafe/model/internal/widget.ts
pomahtri: discussed, `else` doesn't work for all widgets (like private Scrollable), so we need to leave both

packages/devextreme/testing/testcafe/runner.js
pomahtri: discussed, we can't, because ClientFunction can be injected for all tests at once only in before hooks, but they don't run if there's manual page reloading inside test (we do have such test cases)

PR #27006, opened 2024-03-27
Grids: add aria-live a11y zones

williamvinogradov:
Add some containers with ```role="status"``` to the Grids components

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
@@ -1,5 +1,6 @@
 /* eslint-disable max-classes-per-file */
 import domAdapter from '@js/core/dom_adapter';
+import type { dxElementWrapper } from '@js/core/renderer';
 import $ from '@js/core/renderer';
 import browser from '@js/core/utils/browser';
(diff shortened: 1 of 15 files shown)


packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -921,6 +934,14 @@ export class GridView extends modules.View {
       $groupElement.appendTo($rootElement);
     }
 
+    if (!this._a11yGeneralStatusElement) {
+      this._a11yGeneralStatusElement = A11yStatusContainerComponent(
+        { a11yLiveType: 'polite' },
```
Raushen: According to the documentation the [status role](https://www.w3.org/TR/wai-aria-1.2/#status) already has implicit aria-live=polite. We can reduce this argument.
image

Raushen (changes requested):

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
williamvinogradov: Fixed

Raushen (approved):

PR #27007, opened 2024-03-27
fix: nested components render widgets in a detached DOM branch in React Strict Mode (T1191220, T1189864)

VasilyStrelyaev:


packages/devextreme-react/src/core/helpers.ts
```diff
@@ -2,10 +2,10 @@
...
 import { TemplateInstantiationModel, UpdateLocker } from './types';
 
-export const RemovalLockerContext = createContext<UpdateLocker>({
-  lock: () => undefined,
-  unlock: () => undefined,
-});
+export const RemovalLockerContext = createContext<UpdateLocker | undefined>(undefined);
+
+// eslint-disable-next-line @typescript-eslint/no-extra-parens
+export const RestoreTreeContext = createContext<(() => void) | undefined>(undefined);
```
ivanblinov2k17, line 8: won't just <undefined | () => void > work
but i see that undefined usually defined at the end, just a comment

ivanblinov2k17 (approved):

packages/devextreme-react/src/core/helpers.ts
VasilyStrelyaev: I tried both of course, and both `undefined | () => void` and `(() => void) | undefined` trigger ESLint errors for me. I just find the first one better looking.

We also don't permit using the `Function` alias.

Of course, I could use a type name for `() => void` and say `<TypeName | undefined>`, but this looks clumsy.

To be honest, part of the reason I left the `// eslint-disable` in this PR is to bring your attention to this issue and discuss maybe we need to relax ESLint rules a bit.

dxvladislavvolkov (approved):

PR #27009, opened 2024-03-27
HtmlEditor: Fix Firefox Ordered List Numbering Sequence (T1220554)

jdvictoria:


jdvictoria:
[HTML Editor - Creating a number list doesn't have the correct sequence [T1220554]](https://trello.com/c/PqkUeuRr/8600-html-editor-creating-a-number-list-doesnt-have-the-correct-sequence-t1220554)

packages/devextreme/scss/widgets/base/_htmlEditor.scss
```diff
@@ -180,11 +180,14 @@ $transparent-border: 1px solid transparent;
...
   h6,
-  ol,
   ul {
     counter-reset: add-counter-reset($max-indent);
   }
 
+  ol {
+    counter-reset: list-item 0;
+  }
+
```
jdvictoria: Removing ol only on line 183 fixed Firefox bug but this caused problem for the unit test I added on other browsers (e.g. Safari). 

Firefox by default sets counter-reset to 'list-item 0'

Safari on the other hand sets counter-reset to 'none'

To standardize it, I set it explicitly to list-item 0.

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/quillMarkup.tests.js
```diff
@@ -24,4 +24,16 @@ testModule('lists', () => {
             assert.strictEqual($listUi.css('visibility'), expectedVisibility, `UI element is ${hasUi ? '' : 'not'} visible`);
         });
     });
+
+    test('ordered list should have counter-reset property set to default (T1220554)', function(assert) {
```
AlexanderMoiseev: I believe we should test the correct behavior, not implementation details. Let's try to write a test with markup and check if the list ordering is correct and we didn't break anything. You can look at /tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/markup.tests.js for example ('render markup' test), or prolly at ui-firefox tests

EugeniyKiyashko (approved):

AlexanderMoiseev (approved):

PR #27014, opened 2024-03-27
Form: use 'captionRender' instead of 'captionTemplate' in the React GroupedFields demo

EugeniyKiyashko:


apps/demos/Demos/Form/GroupedFields/React/App.tsx
```diff
@@ -8,6 +8,14 @@ import service from './data.ts';
 
 const employee = service.getEmployee();
 
+const groupCaptionRender = (iconName) => (data) =>
```
ksercs, line 11: it can be optimized more if we create an object with precalculated functions, but for demo current approach is ok i guess

ksercs (approved):

AlexanderMoiseev (approved):

PR #27023, opened 2024-03-28
🕛 Decrease bundle size(Time Zones)

Raushen:


packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +57,66 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
 
 const getDaylightOffsetInMs = (startDate, endDate) => getDaylightOffset(startDate, endDate) * toMs('minute');
 
-const calculateTimezoneByValue = (timezone, date = new Date()) => {
-  // NOTE: This check could be removed. We don't support numerical timezones
-  if (typeof timezone === 'string') {
-    const dateUtc = createUTCDate(date);
-    return timeZoneDataUtils.getTimeZoneOffsetById(timezone, dateUtc.getTime());
+const calculateTimezoneByValue = (timeZone, date = new Date()): number | undefined => {
```
williamvinogradov: Let's set the ```timeZone``` variable type here

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +57,66 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+  const result = parseInt(hours, 10) + parseInt(minutes, 10) / 60;
+  return isMinus ? -result : result;
+};
+
+// 'GMT±XX:YY' or 'GMT' format
+const getStringOffset = (timeZone, date = new Date()): string => {
+  const dateTimeFormat = new Intl.DateTimeFormat('en-US', {
+    timeZone,
+    timeZoneName: 'longOffset',
+  } as any);
```
williamvinogradov: Why do we need this ```any``` here?
As far as I can see we can remove this type cast :)

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +57,66 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+
+  return dateTimeFormat.formatToParts(date)
+    .filter((part) => part.type === 'timeZoneName')
+    .map((p) => p.value)[0];
```
williamvinogradov: Let's refactor this part of the code and make it more safe:
```suggestion
  return dateTimeFormat.formatToParts(date)?.find((part) => part.type === 'timeZoneName')?.value
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +57,66 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
-  if (typeof timezone === 'string') {
-    const dateUtc = createUTCDate(date);
-    return timeZoneDataUtils.getTimeZoneOffsetById(timezone, dateUtc.getTime());
+const calculateTimezoneByValue = (timeZone, date = new Date()): number | undefined => {
+  if (!timeZone) {
+    return undefined;
+  }
+
+  let offset;
+  try {
```
williamvinogradov: Can we get rid of this ```try/catch``` block?

E.g:
```typescript
  const offset = getStringOffset(timeZone, date);

  if (!offset) {
    return undefined;
  }
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +57,66 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+    timeZoneName: 'longOffset',
+  } as any);
+
+  return dateTimeFormat.formatToParts(date)
+    .filter((part) => part.type === 'timeZoneName')
+    .map((p) => p.value)[0];
+};
+
+const getOffsetNamePart = (offset: string): string => {
+  if (offset === 'GMT') {
```
williamvinogradov: We use the ```GMT``` often in our new code
Let's move it to the ```const```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +57,66 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+const getTimezoneTitle = (timeZone, date = new Date()): string => {
+  const tzNamePart = timeZone.replaceAll('/', ' - ').replaceAll('_', ' ');
+  const offset = getStringOffset(timeZone, date);
+  const offsetNamePart = getOffsetNamePart(offset);
+  return `(${offsetNamePart}) ${tzNamePart}`;
 };
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
-const _getDaylightOffsetByTimezone = (startDate, endDate, timeZone) => calculateTimezoneByValue(timeZone, startDate) - calculateTimezoneByValue(timeZone, endDate);
+const _getDaylightOffsetByTimezone = (startDate, endDate, timeZone) => {
```
williamvinogradov: Let's add types to this function arguments

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +57,66 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+  if (offset === 'GMT') {
+    return 0;
   }
-  return timezone;
+
+  const isMinus = offset.substring(3, 4) === '-';
+  const hours = offset.substring(4, 6);
+  const minutes = offset.substring(7, 9);
+
+  const result = parseInt(hours, 10) + parseInt(minutes, 10) / 60;
```
williamvinogradov: If something goes wrong and we receive the not a number string in ```hours``` or ```minutes``` variables, the ```parseInt``` will return the ```NaN``` value.
So, let's make this code a little more safe:
```suggestion
  const hoursStr = offset.substring(4, 6);
  const minutesStr = offset.substring(7, 9);
  const hours = parseInt(hoursStr, 10);
  const minutes = parseInt(minutesStr, 10);

  if (Number.isNaN(hours) || Number.isNaN(minutes)) {
    return undefined;
  }

  const result = (hours + minutes) / MINUTES_IN_HOUR;
```

And let's move the "magic number" ```60``` to the ```const```, e.g 
```typescript
const MINUTES_IN_HOUR = 60;
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +57,66 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+  if (offset === 'GMT') {
+    const GMTNamePart = `${offset} +00:00`;
+    return `${GMTNamePart}`;
+  }
+  return offset.replace('GMT', 'GMT ');
+};
+
+const getTimezoneTitle = (timeZone, date = new Date()): string => {
+  const tzNamePart = timeZone.replaceAll('/', ' - ').replaceAll('_', ' ');
+  const offset = getStringOffset(timeZone, date);
```
williamvinogradov, line 157: Looks like we need a fallback here if the ```getStringOffset``` will raise an exception / returns undefined (see comment about getting rid of the ```try/catch``` block)

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Raushen: No. We can get invalid timezone as argument (for example "asia/1111") and the component must work after that.

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Raushen: I think we should fail fast in this case. I wanted to check several browsers before make a decision.

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Raushen: I think there is an error in typescript here. It requires incorrect type.

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Raushen: [formatToParts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts) returns the Array as a result, not null or undefined.

Similarly, the formatting _timeZoneName: 'longOffset'_ splits the date into parts, one of which is _'timeZoneName'_. If this is not the case, the component will not work, and it won't be possible to quickly determine why. I think there should be exceptions here as well.

Why do you want use optional chaining operator?

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Raushen: We decided to research this issue and leave the code as is

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Raushen: We agreed leave the code as is and research the issue

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Raushen: We agreed to throw an exception here

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Raushen: We agreed to throw an exception here

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Raushen: We agreed to throw an exception here

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+  return result;
+};
+
+const calculateTimezoneByValueCore = (timeZone: string, date = new Date()): number | undefined => {
+  let offset;
+  try {
+    offset = getStringOffset(timeZone, date);
+  } catch (e) {
+    errors.log('W0009', timeZone);
+    return undefined;
```
andrewmakarov, line 131: Are you sure the error needs to be handled? Maybe an error should be thrown and the code execution should be interrupted?

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -130,10 +225,19 @@ const hasDSTInLocalTimeZone = () => {
...
+    return isEqualLocalTimeZoneByDeclarationOld(timeZoneName, date);
+  }
+  return isEqualLocalTimeZoneByDeclarationCore(timeZoneName, date);
+};
+
+const isEqualLocalTimeZoneByDeclarationOld = (timeZoneName: string, date: Date): boolean => {
   const year = date.getFullYear();
-  const getOffset = (date) => -date.getTimezoneOffset() / 60;
-  const getDateAndMoveHourBack = (dateStamp) => new Date(dateStamp - 3600000);
+  const getOffset = (date) => -date.getTimezoneOffset() / MINUTES_IN_HOUR;
```
andrewmakarov: Dubbing, there is already such a function(see 279 line)

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -1,13 +1,18 @@
...
 
 import dateUtils from '../../core/utils/date';
 import DateAdapter from './m_date_adapter';
 import timeZoneDataUtils from './timezones/m_utils_timezones_data';
+import timeZoneList from './timezones/timezone_list';
 
 const toMs = dateUtils.dateToMilliseconds;
 const MINUTES_IN_HOUR = 60;
 const MS_IN_MINUTE = 60000;
+const MS_IN_HOUR = 3600000;
```
williamvinogradov: Minor

Let's delete these consts and use the ```toMs``` function instead :)

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+
+  const result = parseInt(hours, 10) + parseInt(minutes, 10) / MINUTES_IN_HOUR;
+  return isMinus ? -result : result;
+};
+
+const calculateTimezoneByValue = (timeZone: string | undefined, date = new Date()): number | undefined => {
+  if (!timeZone) {
+    return undefined;
+  }
+  const isValidTimezone = timeZoneList.value.includes(timeZone);
```
williamvinogradov:
```suggestion
  }
  
  const isValidTimezone = timeZoneList.value.includes(timeZone);
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+
+  if (!isValidDate(date)) {
+    return undefined;
+  }
+
+  let result = calculateTimezoneByValueOld(timeZone, date);
+  if (result === undefined) {
+    result = calculateTimezoneByValueCore(timeZone, date);
+  }
+  return result;
```
williamvinogradov:
```suggestion
  }
  
  return result;
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -130,10 +225,11 @@ const hasDSTInLocalTimeZone = () => {
...
 };
 
-const isEqualLocalTimeZoneByDeclaration = (timeZoneName, date) => {
+const getOffset = (date) => -date.getTimezoneOffset() / MINUTES_IN_HOUR;
+
+const isEqualLocalTimeZoneByDeclarationOld = (timeZoneName: string, date: Date): boolean => {
   const year = date.getFullYear();
-  const getOffset = (date) => -date.getTimezoneOffset() / 60;
-  const getDateAndMoveHourBack = (dateStamp) => new Date(dateStamp - 3600000);
+  const getDateAndMoveHourBack = (dateStamp) => new Date(dateStamp - MS_IN_HOUR);
```
williamvinogradov: Minor

It had been here before, but let's move out the inner scope function to outer scope
E.g.
```typescript
const getDateAndMoveHourBack = (dateStamp) => new Date(dateStamp - toMs('hour'));

const isEqualLocalTimeZoneByDeclarationOld = (...) => { ... } 
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -172,6 +268,15 @@ const isEqualLocalTimeZoneByDeclaration = (timeZoneName, date) => {
   return true;
 };
 
+const isEqualLocalTimeZoneByDeclaration = (timeZoneName: string, date: Date): boolean => {
+  const customTimezones = timeZoneDataUtils.getTimeZonesOld();
+  const targetTimezoneData = customTimezones.filter((tz) => tz.id === timeZoneName);
+  if (targetTimezoneData.length === 1) {
+    return isEqualLocalTimeZoneByDeclarationOld(timeZoneName, date);
+  }
+  return false;
```
williamvinogradov:
```suggestion
  const targetTimezoneData = customTimezones.filter((tz) => tz.id === timeZoneName);
  
  if (targetTimezoneData.length === 1) {
    return isEqualLocalTimeZoneByDeclarationOld(timeZoneName, date);
  }
  
  return false;
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+    .filter((part) => part.type === 'timeZoneName')
+    .map((p) => p.value)[0];
+
+  const isSupportedFormat = offsetFormatRegexp.test(result);
+  if (!isSupportedFormat) {
+    errors.log('W0009', timeZone);
+  }
+
+  return result;
+};
```
williamvinogradov: Let's change this function to handle all possible unhandled errors:
```suggestion
const getStringOffset = (timeZone: string, date = new Date()): string => {
  if (typeof Intl?.DateTimeFormat !== 'function') {
    return undefined;
  }

  const dateTimeFormat = new Intl.DateTimeFormat('en-US', {
    timeZone,
    timeZoneName: 'longOffset',
  } as any);
  const result = dateTimeFormat?.formatToParts?.(date)
    .find(({ type }) => type === 'timeZoneName')?.value ?? '';

  const isSupportedFormat = offsetFormatRegexp.test(result);

  return isSupportedFormat ? result : undefined;
};
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+
+const calculateTimezoneByValueCore = (timeZone: string, date = new Date()): number | undefined => {
+  let offset;
+  try {
+    offset = getStringOffset(timeZone, date);
+  } catch (e) {
+    errors.log('W0009', timeZone);
+    return undefined;
+  }
```
williamvinogradov: Take a look at the ```getStringOffset``` function comment

```suggestion
  const offset = getStringOffset(timeZone, date);

  if (offset === undefined) {
    errors.log('W0009', timeZone);
    return undefined;
  }
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+  return offset.replace(GMT, `${GMT} `);
+};
+
+const getTimezoneTitle = (timeZone: string, date = new Date()): string => {
+  if (!isValidDate(date)) {
+    return '';
+  }
+
+  const tzNamePart = timeZone.replace(/\//g, ' - ').replace(/_/g, ' ');
+  const offset = getStringOffset(timeZone, date);
```
williamvinogradov, line 157: Take a look on the ```getStringOffset``` comment
```suggestion
  const offset = getStringOffset(timeZone, date);
  
  if (offset === undefined) {
    errors.log('W0009', timeZone);
    return undefined;
  }
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+  }
+
+  return result;
+};
+
+const getOffsetNamePart = (offset: string): string => {
+  if (offset === GMT) {
+    return `${offset} +00:00`;
+  }
+  return offset.replace(GMT, `${GMT} `);
```
williamvinogradov:
```suggestion
  if (offset === GMT) {
    return `${offset} +00:00`;
  }
  
  return offset.replace(GMT, `${GMT} `);
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
-const _getDaylightOffsetByTimezone = (startDate, endDate, timeZone) => calculateTimezoneByValue(timeZone, startDate) - calculateTimezoneByValue(timeZone, endDate);
+const _getDaylightOffsetByTimezone = (startDate: Date, endDate: Date, timeZone: string): number => {
+  const startDayOffset = calculateTimezoneByValue(timeZone, startDate);
+  const endDayOffset = calculateTimezoneByValue(timeZone, endDate);
+  if (startDayOffset === undefined || endDayOffset === undefined) {
+    return 0;
+  }
+  return startDayOffset - endDayOffset;
```
williamvinogradov:
```suggestion
  if (startDayOffset === undefined || endDayOffset === undefined) {
    return 0;
  }
  
  return startDayOffset - endDayOffset;
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
andrewmakarov: Yes, I vote for it

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -1,13 +1,18 @@
...
+import timeZoneList from './timezones/timezone_list';
 
 const toMs = dateUtils.dateToMilliseconds;
 const MINUTES_IN_HOUR = 60;
 const MS_IN_MINUTE = 60000;
+const MS_IN_HOUR = 3600000;
+const GMT = 'GMT';
+const offsetFormatRegexp = /^GMT(?:[+-]\d{2}:\d{2})?$/;
 
 const createUTCDateWithLocalOffset = (date) => {
```
andrewmakarov, line 17: I found dublicate function - createUTCDateWithLocalOffset

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -34,10 +39,11 @@ const createDateFromUTCWithLocalOffset = (date) => {
...
-  const dateInUTC = createUTCDate(date);
-  return timeZoneDataUtils.getDisplayedTimeZones(dateInUTC.getTime());
-};
+const getTimeZones = (date = new Date()) => timeZoneList.value.map((tz) => ({
+  offset: calculateTimezoneByValue(tz, date),
+  title: getTimezoneTitle(tz, date),
+  id: tz,
+}));
 
 const createUTCDate = (date) => new Date(Date.UTC(
```
andrewmakarov, line 47: I found dublicate createUTCDateWithLocalOffset

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
+
+  const isMinus = offset.substring(3, 4) === '-';
+  const hours = offset.substring(4, 6);
+  const minutes = offset.substring(7, 9);
+
+  const result = parseInt(hours, 10) + parseInt(minutes, 10) / MINUTES_IN_HOUR;
+  return isMinus ? -result : result;
+};
+
+const calculateTimezoneByValue = (timeZone: string | undefined, date = new Date()): number | undefined => {
```
andrewmakarov, line 94: I see common return of undefined. May be create type Nullable?

`type Nullable<T> = T | undefined`
`const calculateTimezoneByValue = () => Nullable<number>`

@williamvinogradov @pomahtri What do uou think about it?

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -55,17 +61,106 @@ const getDaylightOffset = (startDate, endDate) => new Date(startDate).getTimezon
...
 const getDaylightOffsetInMs = (startDate, endDate) => getDaylightOffset(startDate, endDate) * toMs('minute');
 
-const calculateTimezoneByValue = (timezone, date = new Date()) => {
-  // NOTE: This check could be removed. We don't support numerical timezones
-  if (typeof timezone === 'string') {
-    const dateUtc = createUTCDate(date);
-    return timeZoneDataUtils.getTimeZoneOffsetById(timezone, dateUtc.getTime());
+const isValidDate = (date: Date) => date instanceof Date && !isNaN(date.valueOf());
+
+const calculateTimezoneByValueOld = (timezone: string, date = new Date()) => {
```
andrewmakarov: Please, add return type, otherwise it’s not clear what the function can return undefined.
And my suggestion, if function return undefined, let's nam function starting from try{name_of_function}

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
pomahtri: I'm against, looks useless (what are advantages of `Nullable<T>` vs `T | undefined`?), plus can lead to misunderstandings (why `Nullable` contains `undefined` instead of `null`)

williamvinogradov: Yup, I'm against too
For mee it looks more obvious when we just write a union ```string | undefined```

mpreyskurantov (approved):

pomahtri (approved):

williamvinogradov (approved):

PR #27080, opened 2024-04-04
LoadPanel/LoadIndicator: set default aria configuration, and avoid loading wrapper aria when the showindicator is true

nikkithelegendarypokemonster:


packages/devextreme/js/ui/load_indicator.js
```diff
@@ -103,9 +104,16 @@ const LoadIndicator = Widget.inherit({
...
 
     _init: function() {
-        this.callBase();
+        const label = messageLocalization.format('Loading');
+        const aria = {
+            role: 'alert',
+            label,
+        };
 
+        this.callBase();
```
marker-dao: Why you move this.callBase();?

packages/devextreme/js/ui/load_indicator.js
```diff
@@ -103,9 +104,16 @@ const LoadIndicator = Widget.inherit({
...
         this.$element().addClass(LOADINDICATOR_CLASS);
+
+        this.setAria(aria, this.$element());
```
marker-dao: this.$element() by default. You can remove it.

marker-dao:
```suggestion
        this.setAria(aria);
```

packages/devextreme/js/ui/load_panel.js
```diff
@@ -129,13 +132,26 @@ const LoadPanel = Overlay.inherit({
         const label = message ? defaultLabel : messageLocalization.format('Loading');
 
         const aria = {
-            role: 'alert',
+            role: 'status',
```
marker-dao: I have tested this role with NVDA: let's use role alert in LoadPanel and LoadIndicator, because role status does not allow the loading message to be spoken in most of our cases.

packages/devextreme/js/ui/load_panel.js
```diff
@@ -129,13 +132,26 @@ const LoadPanel = Overlay.inherit({
...
             label,
         };
 
         this.setAria(aria, this.$wrapper());
     },
 
+    _setLoadIndicatorAria() {
```
marker-dao: You can create just one method (_getAria, for example), which will return aria attributes. Because attributes the same for LoadPanel and LoadIndicator.

And now you can use elementAttr option to provide aria to LoadIndicator when we render LoadIndicator (_createComponent).

packages/devextreme/js/ui/load_indicator.js
nikkithelegendarypokemonster: i move callbase there so that i would like to rearrange that variables comes on top before any code its just a formatting thing.

packages/devextreme/js/ui/load_panel.js
nikkithelegendarypokemonster: [7dd1307](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7dd13076101d6fe81bbac1faed779ae225f60827)

packages/devextreme/js/ui/load_panel.js
nikkithelegendarypokemonster: [7dd1307](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7dd13076101d6fe81bbac1faed779ae225f60827)

packages/devextreme/js/ui/load_indicator.js
nikkithelegendarypokemonster: [7dd1307](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7dd13076101d6fe81bbac1faed779ae225f60827)

nikkithelegendarypokemonster: [7dd1307](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7dd13076101d6fe81bbac1faed779ae225f60827)

packages/devextreme/js/ui/load_indicator.js
nikkithelegendarypokemonster: [7dd1307](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7dd13076101d6fe81bbac1faed779ae225f60827)

packages/devextreme/js/ui/load_panel.js
```diff
@@ -136,6 +139,19 @@ const LoadPanel = Overlay.inherit({
         this.setAria(aria, this.$wrapper());
     },
 
+    _getIndicatorAria() {
```
marker-dao:
```suggestion
    _getAriaAttributes() {
```

marker-dao: Let's reduce copy past and use this method for getting aria for all of our cases, because it's the same for load panel and load indicator.

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,8 +60,38 @@ QUnit.module('init', {
         assert.equal($content.find(MESSAGE_SELECTOR).text(), 'Test Loading Message');
     });
 
+    QUnit.test('correct wrapper aria attributes load when showIndicator = false', function(assert) {
+        const label = messageLocalization.format('Loading');
+        const instance = $('#loadPanel').dxLoadPanel({ visible: true, showIndicator: false }).dxLoadPanel('instance');
+        instance.option({ message: '' });
```
marker-dao: You can set this option when you initialize the component:

```
.dxLoadPanel({
    visible: true,
    showIndicator: false,
    message: '',
})
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,8 +60,38 @@ QUnit.module('init', {
...
+        const wrapperAttributes = Array.from(instance[0].attributes).map(attr => attr.name);
+        const expectedAttributes = ['role', 'aria-label'];
+
+        assert.strictEqual(!expectedAttributes.every(item => wrapperAttributes.includes(item)), true);
+    });
+
+    QUnit.test('correct load indicator aria attributes load when showIndicator = true', function(assert) {
+        const instance = $('#loadPanel').dxLoadPanel({ visible: true, showIndicator: true }).dxLoadPanel('instance');
+        const label = messageLocalization.format('Loading');
+        const indicatorAttributes = Array.from(instance._$indicator.prop('attributes')).map(attr => attr.name);
```
marker-dao: We prefer don't use private API in tests. You can get indicator element with:

`instance.$content().find(LOADINDICATOR_CLASS);`

Before that please create a variable LOADINDICATOR_CLASS.

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
marker-dao: Explain please for what you set message: ''?

nikkithelegendarypokemonster: for this issue i noticed when i run my test when i dont change the message it doesnt set aria-label it remains as undefined making the test faulty... But once i set the message option it puts an aria-label on element.

Take note this only occurs on testcafe test not when i test in my playground

packages/devextreme/js/ui/load_panel.js
```diff
@@ -136,6 +139,19 @@ const LoadPanel = Overlay.inherit({
...
+        const { message } = this.option();
+        const defaultLabel = isMaterialBased() ? message : null;
+        const label = message ? defaultLabel : messageLocalization.format('Loading');
+
+        const aria = {
+            role: 'alert',
+            label,
```
marker-dao:
```suggestion
          'aria-label': label,
```

packages/devextreme/js/ui/load_panel.js
```diff
@@ -136,6 +139,19 @@ const LoadPanel = Overlay.inherit({
         this.setAria(aria, this.$wrapper());
```
marker-dao: Let's use _getAriaAttributes method and replace **this.setAria(aria, this.$wrapper());** with **this.$wrapper().attr(aria);**

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
marker-dao: Okay. Let's leave it like that.

packages/devextreme/js/ui/load_indicator.js
```diff
@@ -103,9 +104,16 @@ const LoadIndicator = Widget.inherit({
     },
 
     _init: function() {
-        this.callBase();
+        const label = messageLocalization.format('Loading');
+        const aria = {
+            role: 'alert',
+            label,
+        };
```
ksercs, line 115: So, why should we create this variables here, not just before using in 116th row?

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,8 +60,38 @@ QUnit.module('init', {
...
+        const label = messageLocalization.format('Loading');
+        const instance = $('#loadPanel').dxLoadPanel({ visible: true, showIndicator: false }).dxLoadPanel('instance');
+        instance.option({ message: '' });
+
+        assert.strictEqual(instance.$wrapper().attr('role'), 'alert');
+        assert.strictEqual(instance.$wrapper().attr('aria-label'), label);
+    });
+
+    QUnit.test('wrapper aria attributes must not load when showIndicator = true', function(assert) {
+        const instance = $('#loadPanel').dxLoadPanel({ visible: true, showIndicator: false });
```
ksercs: It's not an instance, it's a LoadPanel root jquery element. Let's update the naming
```suggestion
        const $loadPanel = $('#loadPanel').dxLoadPanel({ visible: true, showIndicator: false });
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,8 +60,38 @@ QUnit.module('init', {
...
+        const wrapperAttributes = Array.from(instance[0].attributes).map(attr => attr.name);
+        const expectedAttributes = ['role', 'aria-label'];
+
+        assert.strictEqual(!expectedAttributes.every(item => wrapperAttributes.includes(item)), true);
```
ksercs: This test will be much more readable it we just directly check that necessary attributes are not set.
```javascript
assert.strictEqual($loadPanel.attr('role'), undefined, 'role is not set');
assert.strictEqual($loadPanel.attr('aria-label'), undefined, 'aria-label is not set');
```

Update all similar asserts please.

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
ksercs: Hm, tests says "when showIndicator=true", but u set it to `false` here ..

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,8 +60,38 @@ QUnit.module('init', {
...
+    });
+
+    QUnit.test('correct load indicator aria attributes load when showIndicator = true', function(assert) {
+        const instance = $('#loadPanel').dxLoadPanel({ visible: true, showIndicator: true }).dxLoadPanel('instance');
+        const label = messageLocalization.format('Loading');
+        const indicatorAttributes = Array.from(instance._$indicator.prop('attributes')).map(attr => attr.name);
+        const expectedAttributes = ['role', 'aria-label'];
+        instance.option({ message: '' });
+
+        assert.strictEqual(instance._$indicator[0].role, 'alert');
```
ksercs: Please do not use private API in tests. It will make future refactoring much more difficult. For example if we change `_$indicator` private property name we'll need to refactor all such tests as well.

You can just get an indicator element through selector. Please check how it's done in similar tests.

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,8 +60,38 @@ QUnit.module('init', {
...
+        assert.strictEqual(instance._$indicator[0].ariaLabel, label);
+        assert.strictEqual(expectedAttributes.every(item => indicatorAttributes.includes(item)), true);
+    });
+
```
ksercs, line 84: Missing tests:
- showIndicator runtime change from false to true and vice versa
- `message` runtime change from empty to not empty and vice versa

I also ask you to create a separate "accessibility" module

packages/devextreme/js/ui/load_panel.js
```diff
@@ -136,6 +139,19 @@ const LoadPanel = Overlay.inherit({
...
     },
 
+    _getIndicatorAria() {
+        const { message } = this.option();
+        const defaultLabel = isMaterialBased() ? message : null;
+        const label = message ? defaultLabel : messageLocalization.format('Loading');
```
ksercs: Actually now I see that this condition for isMaterialBased is excess because label text is not theme specific. Let's remove it and always pass message as a label if it's not empty.
So we can remove defaultLabel and leave only 
```javascript
const label = message || messageLocalization.format('Loading');
```
 
You will also need to update an appropriate test (remove this condition as well)

packages/devextreme/js/ui/load_panel.js
ksercs: I agree with using `_getAriaAttributes` here to reduct copypaste but seems like `this.setAria` should still work well

marker-dao: > but seems like this.setAria should still work well

@ksercs 

We use 'aria-label' prop name in aria object when we get aria configuration. So, we should not use this.setAria(). This method will set 'aria-aria-label' attribute to element.

And we need to use 'aria-label' prop name because we set aria to loadIndicator by elementAttr option. Component will use element.attr() method to set attribute, so if we will use just 'label' prop name, we will get label attribute on element like this: label='value' instead of aria-label='value'.

ksercs: Sure, now I see. Thanks!

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,20 +103,40 @@ QUnit.module('init', {
...
+        assert.strictEqual($loadPanel.$wrapper().attr('aria-label'), undefined, 'aria-label attribute is not set');
 
-        instance.option({ message: '' });
+        $loadPanel.option({ message: '' });
 
-        assert.strictEqual(instance.$wrapper().attr('aria-label'), 'Loading...');
+        assert.strictEqual($loadPanel.$wrapper().attr('aria-label'), 'Loading...', 'aria-label attribute has value Loading...');
     });
 
     QUnit.test('aria-label on wrapper in Material based themes', function(assert) {
```
marker-dao: I think this test can be deleted, because you have removed isMaterialBased from getting aria-label function.

nikkithelegendarypokemonster: i completely agree with you..
[b0fc164](https://github.com/DevExpress/DevExtreme/pull/27080/commits/b0fc16436a7a5e17f74c1a2dfba2b2e3c17154ef)
this is the commit for this one

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,37 +103,40 @@ QUnit.module('init', {
...
 
-        instance.option({ message: '' });
+    QUnit.test('correct load indicator aria attributes load when showIndicator = true', function(assert) {
+        const $loadPanel = $('#loadPanel').dxLoadPanel({ visible: true, showIndicator: true }).dxLoadPanel('instance');
+        const label = messageLocalization.format('Loading');
+        const indicatorAttributes = Array.from($loadPanel.$content().find(LOADINDICATOR_CLASS).prop('attributes')).map(attr => attr.name);
+        const expectedAttributes = ['role', 'aria-label'];
 
-        assert.strictEqual(instance.$wrapper().attr('aria-label'), 'Loading...');
+        assert.strictEqual($loadPanel.$content().find(LOADINDICATOR_CLASS).attr('role'), 'alert', 'role attribute has value alert');
```
EugeniyKiyashko: Let's extract common parts in variables. For example: 
`$content = $loadPanel.$content().find(LOADINDICATOR_CLASS)`

Please, check it in all tests

nikkithelegendarypokemonster: [816e9ae](https://github.com/DevExpress/DevExtreme/pull/27080/commits/816e9aeb8f98a1be05fe93b205cdef06da5e0bac)

EugeniyKiyashko (dismissed):

marker-dao (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,6 +41,64 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+        const loadPanel = $('#loadPanel').dxLoadPanel({
+            visible: true,
+            showIndicator: true,
+        }).dxLoadPanel('instance');
+
+        assert.strictEqual(loadPanel.option('showIndicator'), true, 'showIndicator option is true');
+
+        loadPanel.option('showIndicator', false);
+
+        assert.strictEqual(loadPanel.option('showIndicator'), false, 'showIndicator option is false');
```
ksercs: Hm, what's the meaning of these tests? I don't see how it can be red

nikkithelegendarypokemonster: hi anton It is to replicate the scenario which showIndicator option will be set to false then i will check if it will properly set to false.

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,37 +118,55 @@ QUnit.module('init', {
...
 
-        assert.strictEqual(instance.$wrapper().attr('aria-label'), 'Loading...');
+        assert.strictEqual($indicator.attr('role'), 'alert', 'role attribute has value alert');
+        assert.strictEqual($indicator.attr('aria-label'), label, 'aria-label attribute has value Loading...');
     });
 
-    QUnit.test('aria-label on wrapper in Material based themes', function(assert) {
-        const origIsMaterial = themes.isMaterialBased;
-        themes.isMaterialBased = () => true;
+    QUnit.test('aria-label on wrapper', function(assert) {
```
ksercs: Seems like this test is not necessary anymore because you check it in "correct wrapper aria attributes load when showIndicator = false" test. How it can be red if the mentioned test is green?

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: but okay if this is not enough maybe i can check if the load indicator exist within the load panel element

ksercs: Still don't get why it's necessary. It's like test
```javascript
let a = 2;
assert.strictEqual(a, 2);
```
:)

packages/devextreme/js/ui/load_panel.js
```diff
@@ -114,26 +114,33 @@ const LoadPanel = Overlay.inherit({
...
+        const aria = this._getAriaAttributes();
 
-        const defaultLabel = isMaterialBased() ? message : null;
-        const label = message ? defaultLabel : messageLocalization.format('Loading');
+        this.$wrapper().attr(aria);
+    },
+
+    _getAriaAttributes() {
+        const { message } = this.option();
+        const label = message || messageLocalization.format('Loading');
```
ksercs, line 138: This row is not tested at the moment.
Please, add a test checking aria-label when message is set on init and when it's changed in runtime (from empty to not empty and vice versa)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,37 +118,55 @@ QUnit.module('init', {
...
-        const instance = $('#loadPanel').dxLoadPanel({ visible: true }).dxLoadPanel('instance');
+    QUnit.test('wrapper aria attributes must not load when showIndicator = true', function(assert) {
+        const $loadPanel = $('#loadPanel').dxLoadPanel({
+            visible: true,
+            showIndicator: true
+        });
+
+        assert.strictEqual($loadPanel.attr('role'), undefined, 'role is not set');
+        assert.strictEqual($loadPanel.attr('aria-label'), undefined, 'aria-label is not set');
+    });
```
ksercs: I see you test different behavior on init depending on `loadIndicator` option value => this functionality depends on `loadIndicator` option => we should also test the behavior after this option runtime change :)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -59,37 +118,55 @@ QUnit.module('init', {
...
 
-        assert.strictEqual(instance.$wrapper().attr('aria-label'), undefined);
+        const label = messageLocalization.format('Loading');
 
-        instance.option({ message: '' });
+        const $indicator = loadPanel.$content().find(LOADINDICATOR_CLASS);
 
-        assert.strictEqual(instance.$wrapper().attr('aria-label'), 'Loading...');
+        assert.strictEqual($indicator.attr('role'), 'alert', 'role attribute has value alert');
+        assert.strictEqual($indicator.attr('aria-label'), label, 'aria-label attribute has value Loading...');
```
ksercs: I suggest u moving all tests related to accessibility to a separate module

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
ksercs: It's fixed as I see

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [7c89496](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7c89496fe5ea9a3a695c7dcce3717057deeca65c)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [7c89496](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7c89496fe5ea9a3a695c7dcce3717057deeca65c)

packages/devextreme/js/ui/load_panel.js
nikkithelegendarypokemonster: [7c89496](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7c89496fe5ea9a3a695c7dcce3717057deeca65c)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [7c89496](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7c89496fe5ea9a3a695c7dcce3717057deeca65c)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [7c89496](https://github.com/DevExpress/DevExtreme/pull/27080/commits/7c89496fe5ea9a3a695c7dcce3717057deeca65c)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+        fx.off = true;
+    },
 
-            instance.option({ message: 'custom' });
+    afterEach: function() {
+        fx.off = false;
+    }
+}, () => {
+    QUnit.test('rendered markup', function(assert) {
+        const $element = $('#loadPanel').dxLoadPanel({ message: 'Test Loading Message', visible: true }); const $content = $element.dxLoadPanel('instance').$content();
```
EugeniyKiyashko: Line-break is missed before const

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+    QUnit.module('message option tests', () => {
+        ['', undefined, null].forEach(state => {
+            const stateString = (state === '') ? 'empty' : state;
+            QUnit.test(`message is set to ${stateString} on init`, function(assert) {
+                const loadPanel = $('#loadPanel').dxLoadPanel({
+                    visible: true,
+                    message: state,
+                    showIndicator: false
+                }).dxLoadPanel('instance');
+                const loadPanelMessage = loadPanel.$wrapper().find('.dx-loadpanel-message').text();
```
EugeniyKiyashko: '.dx-loadpanel-message' extract in const varible

EugeniyKiyashko:
```suggestion
                const messageText = loadPanel.$wrapper().find('.dx-loadpanel-message').text();
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+                    showIndicator: false
+                }).dxLoadPanel('instance');
+                let loadPanelMessage = loadPanel.$wrapper().find(`.${LOADPANEL_MESSAGE_CLASS}`).text();
+
+                assert.strictEqual(loadPanelMessage, 'Test', 'Load Panel message value is non-empty and correct value');
+
+                assert.timeout(1000);
+
+                loadPanel.option('message', state);
+                loadPanelMessage = loadPanel.$wrapper().find(`.${LOADPANEL_MESSAGE_CLASS}`).text();
```
EugeniyKiyashko: Can we use the link on the wrapped dom element and call text method in correspond places?

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
-        const instance = $('#loadPanel').dxLoadPanel({ visible: true }).dxLoadPanel('instance');
+    QUnit.module('wrapper / load indicator aria-label tests', () => {
+        QUnit.test('wrapper aria label is removed and insert into load indicator when showIndicator = true on runtime', function(assert) {
+            const loadPanel = $('#loadPanel').dxLoadPanel({
+                visible: true,
+                showIndicator: false
+            }).dxLoadPanel('instance');
+            const $wrapper = loadPanel.$wrapper();
+            let $indicator = loadPanel.$content().find(LOADINDICATOR_CLASS);
+            let indicatorAriaLabelExists = $indicator.is('[aria-label]');
```
EugeniyKiyashko: Can we use .attr() method insted of is(attr)?

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+            assert.strictEqual(wrapperAriaLabelExists, true, 'Wrapper aria label is added');
+            assert.strictEqual(indicatorAriaLabelExists, false, 'Indicator aria label is empty');
+
+            assert.timeout(1000);
+            loadPanel.option('showIndicator', true);
+            $indicator = loadPanel.$content().find(LOADINDICATOR_CLASS);
+            indicatorAriaLabelExists = $indicator.is('[aria-label]');
+            wrapperAriaLabelExists = $wrapper.is('[aria-label]');
+
+            assert.strictEqual(wrapperAriaLabelExists, false, 'Wrapper aria label is removed after runtime');
```
EugeniyKiyashko: We can write $indicator.attr('aria-label') directly

packages/devextreme/js/ui/load_panel.js
```diff
@@ -222,15 +229,19 @@ const LoadPanel = Overlay.inherit({
     },
 
     _optionChanged: function(args) {
+        const { showIndicator } = this.option();
```
ksercs: u don't need getting option value here, `args` parameter already contains it. It's better just using `args.value` below

packages/devextreme/js/ui/load_panel.js
```diff
@@ -222,15 +229,19 @@ const LoadPanel = Overlay.inherit({
...
             case 'delay':
                 break;
             case 'message':
             case 'showIndicator':
+                this.$wrapper().removeAttr('aria-label');
                 this._cleanPreviousContent();
                 this._renderLoadIndicator();
                 this._renderMessage();
-                this._setWrapperAria();
+                if(!showIndicator) {
```
ksercs: Now this condition is duplicated in 2 places. Let's update `setWrapperAria` to contain this condition inside.
I also suggest you renaming it to `updateWrapperAria` and moving attributes removing there too.

packages/devextreme/js/ui/load_panel.js
```diff
@@ -222,15 +229,19 @@ const LoadPanel = Overlay.inherit({
     },
 
     _optionChanged: function(args) {
+        const { showIndicator } = this.option();
         switch(args.name) {
             case 'delay':
                 break;
             case 'message':
             case 'showIndicator':
+                this.$wrapper().removeAttr('aria-label');
```
ksercs: And what about role? We set not only aria-label but role as well => we should remove it here too. 
Please read the comment below about updating `setWrapperAria` method - i think this code should be moved there too.
And don't forget to test this scenario.

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
 const MESSAGE_SELECTOR = '.' + LOADPANEL_MESSAGE_CLASS;
 const LOADPANEL_CONTENT_CLASS = 'dx-loadpanel-content';
 const LOADPANEL_PANE_HIDDEN_CLASS = 'dx-loadpanel-pane-hidden';
+const LOADINDICATOR_CLASS = '.dx-loadpanel-indicator';
```
ksercs: Not it's actually not a class.
```suggestion
const LOAD_INDICATOR_CLASS = 'dx-loadpanel-indicator';
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+
+QUnit.module('accessibility', () => {
+    QUnit.module('message option tests', () => {
```
ksercs: It's obvious that it's tests, nothing other can be here :)
```suggestion
    QUnit.module('message option', () => {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+        ['', undefined, null].forEach(state => {
+            const stateString = (state === '') ? 'empty' : state;
+            QUnit.test(`message is set to ${stateString} on init`, function(assert) {
```
ksercs: You can just write it this way and then u don't need to create additional variables
```suggestion
            QUnit.test(`message is set to "${emptyValue}" on init`, function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+                }).dxLoadPanel('instance');
+                let loadPanelMessage = loadPanel.$wrapper().find(`.${LOADPANEL_MESSAGE_CLASS}`).text();
 
-        assert.ok($element.hasClass(LOADPANEL_CLASS));
-        assert.ok($content.hasClass(LOADPANEL_CONTENT_CLASS), 'Load Indicator created');
-        assert.ok($content.find(MESSAGE_SELECTOR).length);
-        assert.equal($content.find(MESSAGE_SELECTOR).text(), 'Test Loading Message');
+                assert.strictEqual(loadPanelMessage, '', 'Load Panel message value is empty');
+
+                assert.timeout(1000);
```
ksercs: For what do you use this timeouts? For animation skipping? Than it's better just disabled the animation in before each how it's don't in a neighbor module below.

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+            let indicatorAriaLabelExists = $indicator.is('[aria-label]');
+            let wrapperAriaLabelExists = $wrapper.is('[aria-label]');
+
+            assert.strictEqual(wrapperAriaLabelExists, true, 'Wrapper aria label is added');
+            assert.strictEqual(indicatorAriaLabelExists, false, 'Indicator aria label is empty');
+
+            assert.timeout(1000);
+            loadPanel.option('showIndicator', true);
+            $indicator = loadPanel.$content().find(LOADINDICATOR_CLASS);
+            indicatorAriaLabelExists = $indicator.is('[aria-label]');
```
ksercs: Please refactor it to make explicit checks like
```javascript
assert.strictEqual($wrapper.attr('role'), 'alert', 'role attribute has value alert');
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+                    message: state,
+                    showIndicator: false
+                }).dxLoadPanel('instance');
+                const loadPanelMessage = loadPanel.$wrapper().find('.dx-loadpanel-message').text();
 
-QUnit.module('init', {
-    beforeEach: function() {
-        fx.off = true;
-    },
+                assert.strictEqual(loadPanelMessage, '', 'Load Panel message value is empty');
```
ksercs: I see that we still have different understandings of what tests are needed :)
I asked you test that we have correct aria attributes after message option runtime change. That is testing of this [code row](https://github.com/DevExpress/DevExtreme/pull/27080/files#diff-53b011e4c36d90ad74e975c0d3719f7a2a69b4893caec03a31a2086859fed5bbR135). Now if I remove message || here tests are still green because u don't check aria-attributes after message option runtime check - instead of this you check dom element text equals passed message.

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
```diff
@@ -40,56 +41,193 @@ const LOADPANEL_MESSAGE_CLASS = 'dx-loadpanel-message';
...
+                    message: state,
+                    showIndicator: false
+                }).dxLoadPanel('instance');
+                let loadPanelMessage = loadPanel.$wrapper().find(`.${LOADPANEL_MESSAGE_CLASS}`).text();
 
-        assert.ok($element.hasClass(LOADPANEL_CLASS));
-        assert.ok($content.hasClass(LOADPANEL_CONTENT_CLASS), 'Load Indicator created');
-        assert.ok($content.find(MESSAGE_SELECTOR).length);
-        assert.equal($content.find(MESSAGE_SELECTOR).text(), 'Test Loading Message');
+                assert.strictEqual(loadPanelMessage, '', 'Load Panel message value is empty');
```
ksercs: I'm afraid these tests are still not what i meant :(
I mean writing tests to check aria-attributes after message option runtime change, testing [this your code row](https://github.com/DevExpress/DevExtreme/pull/27080/files#diff-53b011e4c36d90ad74e975c0d3719f7a2a69b4893caec03a31a2086859fed5bbR135). We don't need to test everything here, only accessibility. 

For example here is how the first test can look like:
```javascript
QUnit.module('accessibility', () => {
    QUnit.module('aria-label dependent on message option', () => {
        ['', undefined, null].forEach(emptyValue => {
            QUnit.test(`equals default loading text if message equals to ${emptyValue} on init`, function(assert) {
                const loadPanel = $('#loadPanel').dxLoadPanel({
                    visible: true,
                    message: emptyValue,
                    showIndicator: false
                }).dxLoadPanel('instance');

                const $loadPanelWrapper = loadPanel.$wrapper();
                assert.strictEqual($loadPanelWrapper.attr('aria-label'), 'Loading...', 'wrapper aria-label is correct when message is empty on init');
            });
```

nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/js/ui/load_panel.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/js/ui/load_panel.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/js/ui/load_panel.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: I disregarded this suggestion because i dont need to get anymore the innerText content of load panel i should compare with aria-label values instead much correct solution.

[bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

nikkithelegendarypokemonster: [bb57796](https://github.com/DevExpress/DevExtreme/pull/27080/commits/bb57796b0f4e0953f491b0adee68e1f35762a715)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [448c03b](https://github.com/DevExpress/DevExtreme/pull/27080/commits/448c03b658281c7012397b4a2f7c91a25eb14350)

packages/devextreme/testing/tests/DevExpress.ui.widgets/loadPanel.tests.js
nikkithelegendarypokemonster: [d662d61](https://github.com/DevExpress/DevExtreme/pull/27080/commits/d662d615c8fabbb946f84f98a97b813bcbf3a628)

Alyar666 (approved):

ksercs (approved):

marker-dao (approved):

PR #27083, opened 2024-04-04
Tabs: Set correct aria and remove aria from nested buttons

marker-dao:


packages/devextreme/js/ui/tabs.js
```diff
@@ -529,20 +532,32 @@ const Tabs = CollectionWidget.inherit({
...
+                role: null,
+                'aria-label': null,
+                'aria-disabled': null,
+            },
+            onClick: () => {
+                this._updateScrollPosition(offset, 1);
+            },
+            onOptionChanged: ({ name, value, element }) => {
+                if(name === 'disabled' && value) {
+                    this.setAria({ disabled: null }, $(element));
```
ksercs: According to performance reasons it's better to do it in a place where we call option change inside of `_updateNavButtonsState` (some refactoring and extracting of an additional method can be needed). Because `optionChange` is called on almost every action (even hovering) and we actually want to minimize any handlers here.

I understand your idea, this solution keeps a kind of encapsulation and works well if we add one more "disabled" option update but I think that possible performance issues are more real than this case.

marker-dao: [2645c1e4a65e528f6ea25a986348082a6b25187f]

EugeniyKiyashko (approved):

AlexanderMoiseev (approved):

PR #27107, opened 2024-04-08
DataGrid - Toolbar is not shown when the toolbar.visible option is changed (T1225567)

pomahtri:

### DIFF
--- a/e2e/testcafe-devextreme/tests/dataGrid/headerPanel.ts
+++ b/e2e/testcafe-devextreme/tests/dataGrid/headerPanel.ts
@@ -119,7 +119,7 @@ test('Drop-down window should be positioned correctly after resizing the toolbar
 });
 
 
-test(`Toolbar should render on changing visibility if visibiliti is false initially`, async (t) => {
+test(`Toolbar should render on changing visibility if visibility is false initially`, async (t) => {
   const dataGrid = new DataGrid('#container');
 
   await t
@@ -133,7 +133,12 @@ test(`Toolbar should render on changing visibility if visibiliti is false initia
     .ok()
     .expect(dataGrid.getHeaderPanel().element.find('.dx-button').textContent)
     .eql('myTestButton');
+  
+  await dataGrid.option('toolbar.visible', true);
 
+  await t
+    .expect(dataGrid.getHeaderPanel().element.visible)
+    .notOk();
 }).before(async () => createWidget('dxDataGrid', {
   showBorders: true,
   toolbar: {
--- a/e2e/testcafe-devextreme/tests/dataGrid/headerPanel.ts
+++ b/e2e/testcafe-devextreme/tests/dataGrid/headerPanel.ts
@@ -117,3 +117,33 @@ test('Drop-down window should be positioned correctly after resizing the toolbar
     await changeTheme(Themes.genericLight);
   });
 });
+
+
+test(`Toolbar should render on changing visibility if visibiliti is false initially`, async (t) => {
+  const dataGrid = new DataGrid('#container');
+
+  await t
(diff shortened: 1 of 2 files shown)


e2e/testcafe-devextreme/tests/dataGrid/headerPanel.ts
```diff
@@ -117,3 +117,33 @@ test('Drop-down window should be positioned correctly after resizing the toolbar
...
+
+
+test(`Toolbar should render on changing visibility if visibiliti is false initially`, async (t) => {
+  const dataGrid = new DataGrid('#container');
+
+  await t
+    .expect(dataGrid.getHeaderPanel().element.visible)
+    .notOk();
+  
+  await dataGrid.option('toolbar.visible', true);
```
williamvinogradov, line 129: Let's add in this test one more check with setting the ```toolbar.visible``` false just in case

packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
```diff
@@ -205,33 +205,30 @@ export class HeaderPanel extends ColumnsView {
...
     }
     if (args.name === 'toolbar') {
-      args.handled = true;
-      if (this._toolbar) {
-        const parts = getPathParts(args.fullName);
-        const optionName = args.fullName.replace(/^toolbar\./, '');
-
-        if (parts.length === 1) {
-          // `toolbar` case
+      const parts = getPathParts(args.fullName);
```
williamvinogradov, line 208: It's not required, but let's refactor this code a little to make it more obvious/linear/easy-to-support :)

My suggestion is:
```typescript
  public optionChanged(args: any): void {
    switch (args.name) {
      case 'onToolbarPreparing':
        this._invalidate();
        args.handled = true;
        break;
      case 'toolbar':
        this.optionChangedToolbarHandler(args);
        args.handled = true;
        break;
      default:
        break;
    }

    super.optionChanged(args);
  }

  private optionChangedToolbarHandler(args: any): void {
    const optionNameParts = getPathParts(args.fullName);
    const optionName = args.fullName.replace(/^toolbar\./, '');

    if (optionNameParts[1] === 'items') {
      this.optionChangedToolbarItemsHandler(args.value, optionNameParts, optionName);
      return;
    }

    this._invalidate();
  }

  private optionChangedToolbarItemsHandler(value: unknown, optionNameParts: string[], optionName: string): void {
    // `toolbar.items` case
    if (optionNameParts.length === 2) {
      const { items } = this._getToolbarOptions();
      this._toolbar?.option('items', items);
      return;
    }

    // `toolbar.items[i]` case
    if (optionNameParts.length === 3) {
      const normalizedItem = this._normalizeToolbarItems(this._getToolbarItems(), value);
      this._toolbar?.option(optionName, normalizedItem);
      return;
    }

    // `toolbar.items[i].prop` case
    if (optionNameParts.length >= 4) {
      this._toolbar?.option(optionName, value);
      return;
    }
  }
```

pomahtri: Let's do in separate PR later, too much changes for bug fix

packages/devextreme/playground/jquery.html
```diff
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
+
```
williamvinogradov: Looks like these changes got here by accident
Let's remove them

williamvinogradov, line 3: Looks like these changes got here by accident
Let's remove them

williamvinogradov (approved):

PR #27122, opened 2024-04-09
Splitter: passing props to nested splitter 

AlexanderMoiseev:


packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -760,10 +779,15 @@ class Splitter extends (CollectionWidget as any) {
...
       case 'onItemCollapsed':
       case 'onItemExpanded':
+        this._updateNestedSplitterOption(name, value);
         this._createEventAction(name);
         break;
       case 'separatorSize':
         this._updateResizeHandlesOption(name, value);
+        this._updateNestedSplitterOption(name, value);
+        break;
+      case 'rtlEnabled':
```
EugeniyKiyashko: For rtlEnabled property we shouldn't do anything. Default logic will add dx-rtl class. We use rtlEnabled in calculation but we read this option onResize event, and there this value will have right value.

EugeniyKiyashko: In the parent components will call invalidate() and will call full rerender. It suits for all our components. The scenario of changing RTL at runtime is not targeted.

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -747,6 +765,7 @@ class Splitter extends (CollectionWidget as any) {
 
     switch (name) {
       case 'allowKeyboardNavigation':
+        this._updateNestedSplitterOption(name, value);
```
EugeniyKiyashko: Let's safe order - change for parent component and only after this - change correspond options for nested components. Please, to be sure in this order for all rest options.

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -322,6 +325,21 @@ class Splitter extends (CollectionWidget as any) {
     });
   }
 
+  _updateNestedSplitterOption(optionName: string, optionValue: unknown): void {
+    this._iterateItems((index, itemElement) => {
```
EugeniyKiyashko: I think that we should only influence splitters that are declared via items[].splitter. In other cases, templates are used, and we don't know the customers' business cases there

So, the draft version of the code could be as follows
```
this.option('items').forEach((item) => {
      if (item?.splitter) {
        getComponentInstance($(this._getElementByItem(item)).find('.dx-splitter').eq(0)).option(optionName, optionValue);
      }
    });
```
The _getElementByItem method is already used in the menu, and we can either duplicate this or move it to a common parent level

packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
```diff
@@ -2116,6 +2117,98 @@ QUnit.module('Events', moduleConfig, () => {
 });
 
 QUnit.module('Nested Splitters', moduleConfig, () => {
```
EugeniyKiyashko, line 2192: After your changes you should unskip the following test:

```
QUnit.skip(`${eventHandler} should be called when a handle in the nested splitter is dragged, ${eventHandler} has been changed at runtime`
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
```diff
@@ -2116,6 +2117,98 @@ QUnit.module('Events', moduleConfig, () => {
...
+    function getNestedSplitter(splitterElement) {
+        return splitterElement.find(`.${SPLITTER_CLASS}`);
+    }
+
+    [
+        { propertyName: 'allowKeyboardNavigation', propertyValue: false },
```
EugeniyKiyashko, line 2199: There are no tests for testing all events. onItemExpanded, onItemCollapsed for example.

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -322,6 +325,18 @@ class Splitter extends (CollectionWidget as any) {
     });
   }
 
+  _updateNestedSplitterOption(optionName: string, optionValue: unknown): void {
+    this.option('items').forEach((item) => {
```
EugeniyKiyashko: const { items } = this.option();

items.forEach(....

packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
```diff
@@ -2116,6 +2117,108 @@ QUnit.module('Events', moduleConfig, () => {
 });
 
 QUnit.module('Nested Splitters', moduleConfig, () => {
+    function getNestedSplitter(splitterElement) {
```
EugeniyKiyashko: Let's move this function to module's before hook.

packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
```diff
@@ -2116,6 +2117,108 @@ QUnit.module('Events', moduleConfig, () => {
...
+                        ]
+                    }
+                }]
+            });
+
+            this.instance.option(propertyName, propertyValue);
+
+            const $nestedItems = getNestedSplitter(this.$element);
+
+            for(let i = 0; i < $nestedItems.length; i++) {
```
EugeniyKiyashko: Let's extract this logic to utility function and place it in module before hook

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -322,6 +325,18 @@ class Splitter extends (CollectionWidget as any) {
...
+      if (item?.splitter) {
+        const splitterItem = this._findItemElementByItem(item).children(`.${SPLITTER_CLASS}`);
```
EugeniyKiyashko: Splitter elements may not always be child items in wrappers. It's better to choose a solution that does not rely on the hierarchy of DOM elements.

EugeniyKiyashko:
```suggestion
        const $nestedSplitter = this._findItemElementByItem(item).children(`.${SPLITTER_CLASS}`);
```

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -322,6 +325,18 @@ class Splitter extends (CollectionWidget as any) {
...
+
+        if (splitterItem) {
```
EugeniyKiyashko: $nestedSplitter.length?

packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
```diff
@@ -2116,6 +2117,108 @@ QUnit.module('Events', moduleConfig, () => {
...
+                                        { text: 'pane 2' }
+                                    ]
+                                }
+                            }
+                        ]
+                    }
+                }]
+            });
+
+            this.instance.option(propertyName, propertyValue);
```
EugeniyKiyashko: I think it's important that the previous options value be declared during the initialization of the component. After changing an option at runtime, events should use the new handler.

EugeniyKiyashko (approved):

PR #27123, opened 2024-04-09
Splitter: consider edge cases and sync the item.size option value with the actual pane dimensions during initialization

EugeniyKiyashko:


Zedwag (dismissed):

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -249,7 +249,7 @@ class Splitter extends (CollectionWidget as any) {
 
     const itemElement = $itemFrame.get(0);
 
-    setFlexProp(itemElement, FLEX_PROPERTY.flexGrow, 100 / getVisibleItemsCount(this.option('items')));
+    setFlexProp(itemElement, FLEX_PROPERTY.flexGrow, 100 / this.option('items').length);
```
AlexanderMoiseev, line 254: why don't we consider visible items only only anymore?

AlexanderMoiseev (dismissed):

packages/devextreme/js/__internal/ui/splitter/splitter.ts
EugeniyKiyashko: Yes, we take the pane.visible option into account during the calculation of the default layout. Now, this default value does not matter, and it will be overwritten after render all items.

AlexanderMoiseev (dismissed):

AlexanderMoiseev (approved):

PR #27127, opened 2024-04-10
React Demos - fix convert-to-js after breaking change instance -> instance()

ivanblinov2k17:


VasilyStrelyaev (changes requested):
So, what changed in 24.1: refs now expose the `instance()` method instead of the `instance` property.

Where we are dealing with refs, we should change `instance` to `instance()`.

However, in the doughnut demo (and some other demos here), we are dealing with a template argument, which functionally did not change - it used to be a widget, and it is still a widget. What is changed is how we access its type from the Ref type.
 
So, when you add parentheses to the template argument, like in the doughnut demo, it fails:

изображение

Here we should only change the way we access the type of the widget (via `ReturnType` for instance) - like in my sample in a Teams thread:

```ts
import { PieChartRef } from 'devextreme-react/pie-chart';
 
function calculateTotal(pieChart: ReturnType<PieChartRef['instance']>) {
  // ...
}
```
 
In other places, where we are actually dealing with refs, we should add parentheses to `instance()`, but I believe I should have addressed most such places when merging my PR (otherwise it would have failed). Some cases with templates somehow slipped through the checks - that's something we may want to look into...

VasilyStrelyaev (approved):

PR #27147, opened 2024-04-11
Menu: Fix disappearing of a focused root item text when menu item is opened (T1227670)

jdvictoria:


jdvictoria:
[Menu - root item text is disappeared after focusing in generic theme if menu is opened](https://trello.com/c/huXpdIsl/8728-menu-root-item-text-is-disappeared-after-focusing-in-generic-theme-if-menu-is-opened)

ksercs:
@jdvictoria  hey, u don't need to start title's each word with upper letter :)

ksercs (dismissed):

packages/devextreme/scss/widgets/generic/menu/_index.scss
```diff
@@ -46,6 +46,10 @@
   }
 
   .dx-menu-item-has-text {
+    &.dx-state-focused {
```
ksercs: Why do u do it only for `.dx-menu-item-has-text`? It does not work now if item contains an icon.
Don't forget to add a test for this case plz

packages/devextreme/scss/widgets/generic/menu/_index.scss
```diff
@@ -46,6 +46,10 @@
...
+      color: $menu-color;
```
ksercs: It's also a bit strange to me that we add a new css rule. I see that we usually use `$menu-item-expanded-color` for expanded menu. Cannot we implement a similar to how it's done in neighbor themes?

jdvictoria: The reason why I added state-focused rule is because whenever shift+tab happens, the menu focuses and disregards the css color rule from menu-item-expanded then uses the css color rule for item template. 

If we create a style test specifically for menu-item-expanded, even if we don't have the fix, it would be evergreen but the bug would be still there and still uses item template css.

Adding state-focused rule explicitly targets the focused state and enables us to test it.

ksercs: I mean that it seems that `$menu-item-expanded-color` should be applied in this case, not `$menu-color`. We definitely should add `.dx-state-focused` in some place to increase some selectors specificity and made it applied in our case. But i believe it's not a correct place (moreover, now u can see that u need duplicate this code in 2 places).

I see that this issues is resolved [this way](https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme/scss/widgets/material/menuBase/_index.scss#L49) in Material and Fluent. Cannot we use the same approach?

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -2224,6 +2224,32 @@ QUnit.module('keyboard navigation', {
...
+        assert.ok($rootMenuItem.hasClass(DX_STATE_FOCUSED_CLASS), 'root item should have focused class');
+        assert.strictEqual(checkStyleHelper.getColor($rootMenuItem[0]), 'rgb(51, 51, 51)', 'color');
+    });
+
+    [
+        [{ text: 'item1', items: [{ name: 'item_1_1' }] }],
+        [{ icon: 'imageCssClass', items: [{ name: 'item_1_1' }]}],
+        [{ text: 'item1', icon: 'imageCssClass', items: [{ name: 'item_1_1' }]}],
+    ].forEach(item => {
+        checkStyleHelper.testInChromeOnDesktopActiveWindow('root item text should not use color property from template when focused (T1227670)', function(assert) {
```
ksercs: what template?

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -2224,6 +2224,32 @@ QUnit.module('keyboard navigation', {
         assert.notOk($items.eq(0).hasClass(DX_STATE_FOCUSED_CLASS), 'first item lose focus');
     });
 
+    checkStyleHelper.testInChromeOnDesktopActiveWindow('root item text should not use color property from template when focused (T1227670)', function(assert) {
```
ksercs: seems like this test is not redundant because test below covers this case

packages/devextreme/scss/widgets/generic/menu/_index.scss
```diff
@@ -78,6 +78,11 @@
       @include dx-icon-font-centered-sizing($generic-base-icon-size);
     }
   }
+
+  .dx-menu-item-has-icon.dx-state-focused,
+  .dx-menu-item-has-text.dx-state-focused {
```
ksercs: please do it exactly how it's done in material. We wanna this styles be applied only when item is expanded

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
ksercs: @jdvictoria  i still didn't get what template you're talking about in this name :)

packages/devextreme/scss/widgets/generic/menu/_index.scss
ksercs: @jdvictoria  i found that we already have necessary styles in menubase css:
```javascript
.dx-menu-item-expanded {
  color: $menu-item-expanded-color;
  background-color: $menu-item-hover-bg;
}
```
Let's just update the selector to make it applying when item is focused or hovered

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
jdvictoria: Sorry all this time I'm talking about menu item template.
image

ksercs: How is it related to a styling?

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -2224,6 +2224,27 @@ QUnit.module('keyboard navigation', {
...
+    [
+        [{ text: 'item1', items: [{ name: 'item_1_1' }] }],
+        [{ icon: 'imageCssClass', items: [{ name: 'item_1_1' }] }],
+        [{ text: 'item1', icon: 'imageCssClass', items: [{ name: 'item_1_1' }] }],
+    ].forEach(item => {
+        checkStyleHelper.testInChromeOnDesktopActiveWindow('root item text should not use color property from item template when focused (T1227670)', function(assert) {
+            this.instance.option('items', item);
+
+            $(this.instance.itemsContainer())
+                .find('.' + DX_MENU_ITEM_CLASS)
```
ksercs:
```suggestion
                .find(`.${DX_MENU_ITEM_CLASS}`)
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -2224,6 +2224,27 @@ QUnit.module('keyboard navigation', {
...
+                .eq(0)
+                .trigger('dxhoverstart')
```
ksercs: Why do u trigger hovering?

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
jdvictoria: The bug that we're dealing right now, without our fix, the style that it uses is from the menu item template. As you can see on the first assert rgb(255, 255, 255), that's the same style that it uses the causes the disappearance.

Screenshot 2024-04-15 at 10 02 27 PM

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -2224,6 +2224,27 @@ QUnit.module('keyboard navigation', {
...
+                .trigger('dxclick');
+
+            const $rootMenuItem = $(this.instance.itemElements().eq(0));
+
+            assert.ok($rootMenuItem.hasClass(DX_MENU_ITEM_EXPANDED_CLASS), 'root item should have expanded class');
+            assert.strictEqual(checkStyleHelper.getColor($rootMenuItem[0]), 'rgb(51, 51, 51)', 'color');
```
ksercs, line 2242: this test is evergreen now :(

packages/devextreme/scss/widgets/generic/menu/_index.scss
```diff
@@ -34,7 +34,9 @@
 .dx-menu {
   color: $menu-color;
 
-  .dx-menu-item-expanded {
+  .dx-menu-item-expanded,
+  .dx-menu-item-expanded.dx-state-focused {
```
ksercs: seems like the solution works even without this selector (`.dx-menu-item-expanded` is enough)

ksercs (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
jdvictoria: That's why I separated it to state-focused because when I put the solution on menu-item-expanded, it's always evergreen :( I'll investigate this.

packages/devextreme/scss/widgets/generic/menu/_index.scss
```diff
@@ -35,7 +35,8 @@
   color: $menu-color;
 
   .dx-menu-item-expanded {
-    background-color: $menu-popup-bg;
```
ksercs, line 38: let's leave it as is

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
ksercs: In this test we check item template styles, but this styles are specified on component level, not in template or specifically for template :) 
So it's better to name it like `'root item text should be visible after focusing when it's opened (T1227670)'`

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -2224,6 +2224,26 @@ QUnit.module('keyboard navigation', {
         assert.notOk($items.eq(0).hasClass(DX_STATE_FOCUSED_CLASS), 'first item lose focus');
     });
 
+    [
+        [{ text: 'item1', items: [{ name: 'item_1_1' }] }],
+        [{ icon: 'imageCssClass', items: [{ name: 'item_1_1' }] }],
+        [{ text: 'item1', icon: 'imageCssClass', items: [{ name: 'item_1_1' }] }],
+    ].forEach(item => {
```
ksercs:
```suggestion
    ].forEach(items => {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
ksercs: That's because root menu item is not focused after popup opening.
Please, call `instance.focus()` and also add an assert that focused class is applied.

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -2224,6 +2224,26 @@ QUnit.module('keyboard navigation', {
...
+        checkStyleHelper.testInChromeOnDesktopActiveWindow('root item text should not use color property from item template when focused (T1227670)', function(assert) {
+            this.instance.option('items', item);
+
+            $(this.instance.itemsContainer())
+                .find(`.${DX_MENU_ITEM_CLASS}`)
+                .eq(0)
```
ksercs: here we can just use `$rootMenuItem`, no need to additionally get this element

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
jdvictoria: About that, I actually tried this method last week, after triggering to expand, I can't seem to focus it even with instance.focus(). Also tried $rootItem.focus() but doesn't work.

packages/devextreme/scss/widgets/generic/menu/_index.scss
ksercs: ?

ksercs: If we don't see a difference in the behavior and idealogical advantages of this addition let's get rid of it.

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
ksercs: checked - yeah, i see that there is smth strange with `focus` method in native js. You can do it this way: `this.$element.trigger('focusin');`

ksercs (approved):
Actually it's an outdated approach to test UI using QUnit tests but I see that it's usual for this piece of code, so let's not make a local revolution for this time because of such a small change :)

But for future changes i prefer we use TestCafe Screenshot tests.

AlexanderMoiseev (approved):

PR #27182, opened 2024-04-16
CheckBox: Test all configuration using TestCafe

jdvictoria:


jdvictoria:
[CheckBox: test all configuration using testcafe](https://trello.com/c/hIfp1hVu/8397-checkbox-test-all-configuration-using-testcafe)

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -1,49 +1,80 @@
-import { createScreenshotsComparer } from 'devextreme-screenshot-comparer';
+/* eslint-disable no-restricted-syntax */
```
EugeniyKiyashko, line 1: Why we need to have this comment in this file?

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -1,49 +1,79 @@
...
+const HOVER_STATE_CLASS = 'dx-state-hover';
+const FOCUSED_STATE_CLASS = 'dx-state-focused';
+const DISABLED_STATE_CLASS = 'dx-state-disabled';
 
-fixture.disablePageReloads`CheckBox`
+fixture.disablePageReloads`CheckBox render`
   .page(url(__dirname, '../../container.html'));
 
-[false, true].forEach((isColumnCountStyle) => {
-  test(`Render ${!isColumnCountStyle ? 'default' : 'with column-count style on container'}`, async (t) => {
```
EugeniyKiyashko: The initial test was written to fix an issue in a specific scenario - https://github.com/DevExpress/DevExtreme/pull/22051. After modifying the existing test, it is necessary to ensure that the new test continues to cover the scenario as well as the previous one did.

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -1,49 +1,79 @@
...
-    await t
-      .expect(compareResults.isValid())
-      .ok(compareResults.errorMessages());
-  }).before(async () => {
-    await setStyleAttribute(Selector('#container'), `padding: 5px; width: 300px; height: 200px; ${isColumnCountStyle ? 'column-count: 2' : ''}`);
+        await t
+          .expect(compareResults.isValid())
+          .ok(compareResults.errorMessages());
+      }).before(async (t) => {
+        t.ctx.ids = [];
```
EugeniyKiyashko: You don't pass ctx to body of the test. And seems that ids array is unused. In other tests it uses for changing state classes at runtime for all components in the page.

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -1,49 +1,79 @@
...
 
-fixture.disablePageReloads`CheckBox`
+fixture.disablePageReloads`CheckBox render`
   .page(url(__dirname, '../../container.html'));
 
-[false, true].forEach((isColumnCountStyle) => {
-  test(`Render ${!isColumnCountStyle ? 'default' : 'with column-count style on container'}`, async (t) => {
-    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+[false, true].forEach((rtlEnabled) => {
+  [20, 30].forEach((iconSize) => {
```
EugeniyKiyashko: Because iconSize is always defined, the size of the checkbox will be the same in both default and compact sizing modes. The default state in compact mode should be represented

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -1,49 +1,79 @@
...
-    await createWidget('dxCheckBox', { value: true, text: 'checked' }, '#checked');
+        await setStyleAttribute(Selector('#container'), `padding: 5px; width: ${limitedWidth ? croppedWidth : fullWidth}px; height: 200px; column-count: 6;`);
 
-    await appendElementTo('#container', 'div', 'unchecked');
-    await createWidget('dxCheckBox', { value: false, text: 'unchecked' }, '#unchecked');
+        await insertStylesheetRulesToPage(`.${CHECKBOX_CLASS} { display: block; }`);
 
-    await appendElementTo('#container', 'div', 'indeterminate');
-    await createWidget('dxCheckBox', { value: undefined, text: 'indeterminate' }, '#indeterminate');
+        for (const state of [
```
EugeniyKiyashko: Is there a possibility to reduce the number of screenshots and test more configurations on one screenshot?

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -1,49 +1,79 @@
...
-fixture.disablePageReloads`CheckBox`
+fixture.disablePageReloads`CheckBox render`
   .page(url(__dirname, '../../container.html'));
 
-[false, true].forEach((isColumnCountStyle) => {
-  test(`Render ${!isColumnCountStyle ? 'default' : 'with column-count style on container'}`, async (t) => {
-    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+[false, true].forEach((rtlEnabled) => {
+  [20, 30].forEach((iconSize) => {
+    [false, true].forEach((limitedWidth) => {
```
EugeniyKiyashko: The following configuration are missed:
image

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -1,49 +1,79 @@
...
+      }).before(async (t) => {
+        t.ctx.ids = [];
 
-    await insertStylesheetRulesToPage(`.${CHECKBOX_CLASS} { display: block; }`);
+        const croppedWidth = iconSize === 20 ? 400 : 450;
+        const fullWidth = iconSize === 20 ? 450 : 500;
 
-    await appendElementTo('#container', 'div', 'checked');
-    await createWidget('dxCheckBox', { value: true, text: 'checked' }, '#checked');
+        await setStyleAttribute(Selector('#container'), `padding: 5px; width: ${limitedWidth ? croppedWidth : fullWidth}px; height: 200px; column-count: 6;`);
```
EugeniyKiyashko: Why should we use CSS styles to limit the width of the checkbox? Does checkbox API not support this?

jdvictoria: I used the existing CSS styles to limit the width of checkbox and label to crop the label and not the size of the checkbox itself.

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
jdvictoria: This comment was used on neighbor tests that utilizes for loops inside .before, without it Lint TS would fail.

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,52 @@ fixture.disablePageReloads`CheckBox`
...
   });
 });
+
+['one two three', 'label'].forEach((text) => {
+  test(`Checkbox configurations label=${text}`, async (t) => {
+    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+    await testScreenshot(t, takeScreenshot, `CheckBox configurations label=${text}.png`, { element: '#container', shouldTestInCompact: true });
+
+    await testScreenshot(t, takeScreenshot, `CheckBox configurations label=${text}.png`, { element: '#container', theme: getFullThemeName().replace('light', 'dark') });
```
AlexanderMoiseev: i'd move getFullThemeName().replace('light', 'dark')  to theme utils, and call the helper function like getDarkTheneName

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,52 @@ fixture.disablePageReloads`CheckBox`
...
+    const stateClasses = [
+      READONLY_STATE_CLASS,
+      DEFAULT_STATE_CLASS,
+      ACTIVE_STATE_CLASS,
+      HOVER_STATE_CLASS,
+      FOCUSED_STATE_CLASS,
+      DISABLED_STATE_CLASS,
+    ];
+
+    for (const rtlEnabled of [false, true]) {
```
AlexanderMoiseev: i think this is too much for 1 test, it runs for too long. i believe it's better to split some checks, for example in 1 test we can check stateClasses with states, like cheked, unchecked etc, other would test the appearance with different width, the other one - different scaled container etc. and it would be easier to follow and debug if something breaks

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,52 @@ fixture.disablePageReloads`CheckBox`
...
+      for (const iconSize of [undefined, 25]) {
+        for (const width of [undefined, 35]) {
```
AlexanderMoiseev: i'd add more width, iconSize values, but prolly only if we splitt the test into several tests, it is already taking too long

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
AlexanderMoiseev: isn't setting 'width' property of checkbox is not enought?

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
AlexanderMoiseev: but, why modify existing tests, lets leave it as is? in our new tests we check checkboxes properties, i dont think we should mix them.

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
jdvictoria: Hello! Yes, Eugeniy was correct. I modified the width configuration on the CheckBox instead of the CSS style on my latest PR.

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
jdvictoria: You're right. I found also some issues with the screenshots too when I mixed them. I already left the existing test and created a new one.

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,93 @@ fixture.disablePageReloads`CheckBox`
...
+    await t
+      .expect(compareResults.isValid())
+      .ok(compareResults.errorMessages());
+  }).before(async () => {
+    await setStyleAttribute(Selector('#container'), 'padding: 5px; width: 300px; height: 400px; column-count: 2;');
+
+    await insertStylesheetRulesToPage(`.${CHECKBOX_CLASS} { display: block; }`);
+
+    for (const text of ['label', 'one two three']) {
+      for (const rtlEnabled of [false, true]) {
```
AlexanderMoiseev: you check rtl in both of your tests. If you want to test RTL alignment with and without labels in a single test, let's perform it without other configurations. I mean dedicated tests only for testing RTL with/without labels, with different label widths (normal and cropped, without different configurations , like active state etc, coz they don't affect rtl)

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,93 @@ fixture.disablePageReloads`CheckBox`
...
+
+    await testScreenshot(t, takeScreenshot, `CheckBox configurations width=${width === undefined ? 'default' : width}.png`, { element: '#container', shouldTestInCompact: true });
+
+    await testScreenshot(t, takeScreenshot, `CheckBox configurations width=${width === undefined ? 'default' : width}.png`, { element: '#container', theme: getDarkThemeName() });
+
+    await t
+      .expect(compareResults.isValid())
+      .ok(compareResults.errorMessages());
+  }).before(async () => {
+    await setStyleAttribute(Selector('#container'), 'padding: 5px; width: 300px; height: 400px; column-count: 2;');
```
AlexanderMoiseev: we have another test for column count, this is not a checkbox property, i suggest get rid of this styles. to crop label you can set checkboxes 'width' property. Anyway, if you need to test specific configurations ( styles ) for container, lets move it to separate test, eg 'test checkboxes appearance with differenc container configuration'.

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,110 @@ fixture.disablePageReloads`CheckBox`
...
+
+  await testScreenshot(t, takeScreenshot, 'CheckBox configurations different iconSizes.png', { element: '#container', shouldTestInCompact: true });
+
+  await testScreenshot(t, takeScreenshot, 'CheckBox configurations different iconSizes.png', { element: '#container', theme: getDarkThemeName() });
+
+  await t
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+}).before(async () => {
+  for (const iconSize of [20, undefined, 25]) {
```
AlexanderMoiseev: It's probably better to use a larger size, for example, 45 instead of 25. That way, the issue with 'indeterminate icon scaling in Fluent' will be more noticeable.

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,110 @@ fixture.disablePageReloads`CheckBox`
...
+  for (const width of [45, undefined, 65]) {
+    for (const text of [undefined, 'label', 'one two three']) {
+      for (const state of [
+        READONLY_STATE_CLASS,
+        DEFAULT_STATE_CLASS,
+        ACTIVE_STATE_CLASS,
+        HOVER_STATE_CLASS,
+        FOCUSED_STATE_CLASS,
+        DISABLED_STATE_CLASS,
+      ] as any[]
```
AlexanderMoiseev: why use any? won't string[] work ?

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,110 @@ fixture.disablePageReloads`CheckBox`
...
+  await testScreenshot(t, takeScreenshot, 'CheckBox configurations different iconSizes.png', { element: '#container', shouldTestInCompact: true });
+
+  await testScreenshot(t, takeScreenshot, 'CheckBox configurations different iconSizes.png', { element: '#container', theme: getDarkThemeName() });
+
+  await t
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+}).before(async () => {
+  for (const iconSize of [20, undefined, 25]) {
+    for (const text of [undefined, 'label', 'one two three']) {
```
AlexanderMoiseev: i don't think testing different text here is necessary, you test it in the next test where you check different widths and text cropping

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,110 @@ fixture.disablePageReloads`CheckBox`
...
+  await testScreenshot(t, takeScreenshot, 'CheckBox configurations different rtlOrientation.png', { element: '#container', theme: getDarkThemeName() });
+
+  await t
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+}).before(async () => {
+  for (const rtlEnabled of [false, true]) {
+    for (const width of [undefined, 45, 65]) {
+      for (const text of [undefined, 'label', 'one two three']) {
+        for (const value of valueModes) {
```
AlexanderMoiseev: dont think testing 'valueModes' in RTL tests are useful,  lets get rid of this option in this test

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,104 @@ fixture.disablePageReloads`CheckBox`
     await createWidget('dxCheckBox', { value: undefined, text: 'indeterminate', rtlEnabled: true }, '#indeterminateRTL');
   });
 });
+
+test('Checkbox configuration, different Checkbox icon sizes', async (t) => {
```
AlexanderMoiseev:
```suggestion
test('Checkbox configuration, different Checkbox icon sizes, states, value Modes', async (t) => {
```

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,104 @@ fixture.disablePageReloads`CheckBox`
...
+  await testScreenshot(t, takeScreenshot, 'CheckBox configurations different widths.png', { element: '#container', shouldTestInCompact: true });
+
+  await testScreenshot(t, takeScreenshot, 'CheckBox configurations different widths.png', { element: '#container', theme: getDarkThemeName() });
+
+  await t
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+}).before(async () => {
+  for (const width of [45, undefined, 65]) {
+    for (const text of [undefined, 'label', 'one two three']) {
```
AlexanderMoiseev, line 110: We test states and value modes in another test. For testing width and label cropping (if that is the intention of this test), we don't need to use several value modes and states. Let's remove them from this test

e2e/testcafe-devextreme/tests/editors/checkBox/common.ts
```diff
@@ -47,3 +59,104 @@ fixture.disablePageReloads`CheckBox`
...
+      for (const state of [
```
AlexanderMoiseev: We test states and value modes in another test. For testing width and label cropping (if that is the intention of this test), we don't need to use several value modes and states. Let's remove them from this test

PR #27189, opened 2024-04-17
ScrollView: Support for simulated scrolling in ShadowDOM

marker-dao:


packages/devextreme/js/events/utils/index.js
```diff
@@ -197,4 +198,6 @@ export const getChar = ({ key, which }) => key || String.fromCharCode(which);
 
 export const addNamespace = mappedAddNamespace;
 
+export const getEventTarget = mappedGetEventTarget;
```
ksercs: [code splitting]
I think it's enough to write your new function directly here, no need to create a separate file.

packages/devextreme/testing/tests/DevExpress.ui.events/events.utils.tests.js
```diff
@@ -505,6 +506,55 @@ QUnit.module('event utils', () => {
...
+            const target = getEventTarget(event);
+
+            assert.strictEqual(div, target, 'getEventTarget returned target correctly');
+        });
+
+        div.dispatchEvent(customEvent);
+    });
+
+    if(QUnit.urlParams['nojquery']) {
+        skipInShadowDomMode('getEventTarget returns the element on which the event was raised if there is ShadowDOM', function(assert) {
```
ksercs: So, we skip it because we emulate shadowDOM manually and we don't need one more level on wrapping to shadowDOM? Isn't it possible to use common wrapping and get rid of our manual wrapping?

packages/devextreme/testing/helpers/pointerMock.js
```diff
@@ -151,6 +151,13 @@
                 return this;
             },
 
+            _wheel: (config) => {
+                triggerEvent('wheel', config);
+
+                // eslint-disable-next-line spellcheck/spell-checker
+                inferno.rerender();
```
ksercs: what is this for?

packages/devextreme/testing/tests/DevExpress.ui.events/events.utils.tests.js
```diff
@@ -505,6 +506,55 @@ QUnit.module('event utils', () => {
...
+            assert.expect(1);
+
+            const shadowHost = document.getElementById('qunit-fixture');
+            const shadowRoot = shadowHost.attachShadow({ mode: 'open' });
```
ksercs: [test isolation]
oh, it's better never do smth with `qunit-fixture` because it does not change between test run, only its content is updated. Not i believe qunit-fixture is always shadowed after this tests is run

packages/devextreme/testing/tests/DevExpress.ui.widgets/scrollView.tests.js
```diff
@@ -290,6 +291,51 @@ QUnit.module('onReachBottom', () => {
     });
 });
 
+if(QUnit.urlParams['nojquery']) {
+    QUnit.module('ShadowDom', moduleConfig, () => {
+        QUnit.skipInShadowDomMode('dxscrollinit handler should be fired after triggering of wheel event', function(assert) {
```
ksercs, line 296: Again, why do we need skip common wrapping and write a specific shadowDOM test?

packages/devextreme/testing/tests/DevExpress.ui.widgets/scrollView.tests.js
```diff
@@ -290,6 +291,51 @@ QUnit.module('onReachBottom', () => {
...
+                    shadowHost,
+                ],
+            };
+
+            pointerMock(scrollViewContent)
+                .start()
+                ._wheel({
+                    target: shadowHost,
+                    originalEvent: $.Event('wheel', originalEvent),
+                });
```
ksercs, line 335: Why cannot we write more functional test with real scrolling? Like just usual trigging of a wheel events, i'm pretty sire we have similar tests for scrollview/scrollable

packages/devextreme/js/events/core/events_engine.js
```diff
@@ -274,6 +275,15 @@ function getHandlersController(element, eventName) {
                     }
                 } else {
                     e.currentTarget = e.delegateTarget || e.target;
+
+                    const isTargetInShadowDOM = Boolean(e.target?.shadowRoot);
+
+                    if(isTargetInShadowDOM) {
+                        const target = getEventTarget(e.originalEvent);
+
+                        e.target = target;
```
ksercs, line 282: I don't see this is tested on core level now, cannot we do it?

packages/devextreme/testing/helpers/pointerMock.js
ksercs: I believe it's not necessary now because actual scrollview tests should not be run in renovation

packages/devextreme/testing/helpers/pointerMock.js
```diff
@@ -151,6 +151,13 @@
                 return this;
             },
 
+            _wheel: (config) => {
```
ksercs: I think our case is too specific for adding one more method to PointerMock. You can also see that there is no other "private" methods.
Let's just directly trigger necessary event

packages/devextreme/testing/tests/DevExpress.ui.events/events.utils.tests.js
ksercs: If we write a test through common wrapping mechanism we will also be able to remove `if(QUnit.urlParams['nojquery']) {`

packages/devextreme/testing/tests/DevExpress.ui.events/events.utils.tests.js
```diff
@@ -505,6 +506,38 @@ QUnit.module('event utils', () => {
...
+
+        const $element = $('#element');
+        const $div = $('<div>');
+
+        $element.append($div);
+
+        const listener = (event) => {
+            const target = getEventTarget(event);
+
+            assert.strictEqual($div.get(0), target, 'getEventTarget returned target correctly');
```
ksercs:
```suggestion
            assert.strictEqual(target, $div.get(0), 'getEventTarget returned target correctly');
```

packages/devextreme/js/events/core/events_engine.js
ksercs: [commit](https://github.com/DevExpress/DevExtreme/pull/27189/commits/a85ed2ce8498903d30c599cb4191a7916c116bfb)

packages/devextreme/js/events/click.js
```diff
@@ -83,12 +84,9 @@ const ClickEmitter = Emitter.inherit({
         const getTarget = function(e) {
             const originalEvent = e.originalEvent;
 
-            if(originalEvent?.target?.shadowRoot) {
-                const path = originalEvent.path ?? originalEvent.composedPath?.();
-                return $(path[0]);
-            }
+            const target = originalEvent ? getEventTarget(originalEvent) : e.target;
```
ksercs: cannot we move this check for `originalEvent` into `getEventTarget` to reduce copypaste? I see that now we write similar code in all 3 places where we call it

packages/devextreme/js/events/core/events_engine.js
```diff
@@ -274,6 +275,15 @@ function getHandlersController(element, eventName) {
                     }
                 } else {
                     e.currentTarget = e.delegateTarget || e.target;
+
+                    const isTargetInShadowDOM = Boolean(e.target?.shadowRoot);
```
ksercs, line 279: This we already do inside of `getEventTarget` as well

packages/devextreme/js/events/utils/event_target.js
```diff
@@ -0,0 +1,12 @@
...
+
+    if(isShadowDOMUsed) {
+        const path = event.path ?? event.composedPath?.();
+        const target = path[0];
+
+        return target;
+    }
+
+    return event?.target;
+};
```
ksercs, line 12: So, why don't u want just to put it inside `utils/index.js`? This util seems to be small and specific enough for placing it it a separate file

packages/devextreme/js/events/core/events_engine.js
ksercs: [commit](https://github.com/DevExpress/DevExtreme/pull/27189/commits/a61f0524d90019a7baca359bd7062f2d419c1f91)

packages/devextreme/js/events/click.js
ksercs: [commit](https://github.com/DevExpress/DevExtreme/pull/27189/commits/a61f0524d90019a7baca359bd7062f2d419c1f91)

ksercs (approved):

packages/devextreme/js/events/pointer/base.js
```diff
@@ -35,11 +36,8 @@ const BaseStrategy = Class.inherit({
             timeStamp: browser.mozilla ? (new Date()).getTime() : e.timeStamp
         };
 
-        const originalEvent = e.originalEvent;
-        if(originalEvent?.target?.shadowRoot) {
-            const path = originalEvent.path ?? originalEvent.composedPath?.();
-            event.target = path[0];
-        }
+        const target = getEventTarget(e);
+        event.target = target;
```
ivanblinov2k17, line 40: why do we need a separate var for target?
couldn't we just write 
event.target = getEventTarget(e);

ivanblinov2k17 (approved):

packages/devextreme/js/events/pointer/base.js
ksercs: Yeah, we can, left it this way just for debugging simplifying

PR #27200, opened 2024-04-18
Menu scrolling: extend show and hide handlers API

iBat:


packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -1258,6 +1402,97 @@ QUnit.module('Menu tests', {
...
+        assert.strictEqual(handlerHidden.args[2][0].itemData.id, 1);
+        assert.strictEqual(getDomNode(handlerHidden.args[2][0].submenuContainer), rootSubmenu);
+
+        assert.strictEqual(getDomNode(handlerHiding.args[2][0].rootItem), $rootItem[0], '2nd level hide');
+        assert.strictEqual(handlerHiding.args[2][0].itemData.id, 12);
+        assert.strictEqual(getDomNode(handlerHiding.args[2][0].submenuContainer), nestedSubmenu2);
+        assert.strictEqual(getDomNode(handlerHidden.args[1][0].rootItem), $rootItem[0]);
+        assert.strictEqual(handlerHidden.args[1][0].itemData.id, 12);
+        assert.strictEqual(getDomNode(handlerHidden.args[1][0].submenuContainer), nestedSubmenu2);
+    });
```
ksercs, line 1563: [code style]
Please refactor it. It's totally unreadable now.
1. Do not use private API. To already have `rootSubmenu` value found in public way, but below you do it one more time using private API.
2. Reduce copypaste. Now you get same elements several times. You can also extract parameters checking to a method (on a test level). It will also help to get rid of strange `getDomMode` function
3. Why do we need to check parameters both for submenu1 and submenu2? Seems like the test cannot be red for second menu of it's greed for the first one.
4. Check the whole itemData equality, not only id. It will not only improve coverage but make the test more intuitive and clear: `handlerShowing.args[0][0].itemData === options.items[0]`
5. Add meaningful assert comments and update the test name.

Actually I also prefer splitting this test. It's okay to leave the single test if it's readable enough tho.

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -1258,6 +1402,97 @@ QUnit.module('Menu tests', {
...
+
```
ksercs, line 1564: Please add a test for scenario for which we extend this API. Like `it should be possible to restrict specific submenu height using API`

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -637,31 +637,41 @@ class Menu extends MenuBase {
         });
     }
 
-    _submenuOnShowingHandler($rootItem, submenu) {
+    _submenuOnShowingHandler($rootItem, submenu, eventArgs) {
```
ksercs: [code style]
I see`eventArgs.rootItem` and `$rootItem` but it's different things. Can we rename it to make the code clearer? Maybe we even don't need to have the whole eventArgs

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -637,31 +637,41 @@ class Menu extends MenuBase {
...
         const $border = $rootItem.children(`.${DX_CONTEXT_MENU_CONTAINER_BORDER_CLASS}`);
+        const { itemData, $submenuContainer } = this._getVisibilityChangeEventParams(eventArgs.rootItem, submenu, $rootItem);
 
         this._actions.onSubmenuShowing({
+            itemData,
             rootItem: getPublicElement($rootItem),
-            submenu: submenu
+            submenuContainer: getPublicElement($submenuContainer),
+            submenu,
```
ksercs: [copypaste]
It seems like all these params are the same for all handlers. 
Cannot we move it to `_getVisibilityChangeEventParams` to reduce copypaste?

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -674,13 +684,36 @@ class Menu extends MenuBase {
...
         });
     }
 
+    _getVisibilityChangeEventParams(submenuItem, submenu, $rootItem, isHide) {
+        let itemData;
+        let $submenuContainer;
+
+        if(submenuItem) {
+            const anchor = isHide ? $(submenuItem).closest(`.${DX_MENU_ITEM_CLASS}`)[0] : submenuItem;
```
ksercs, line 688: So, `submenuItem` has different value on showing and on hiding? Looks incorrect.
Cannot we change it? Or why is it necessary?

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -3350,6 +3642,9 @@ QUnit.module('adaptivity: behavior', {
     });
 });
 
+function getDomNode(el) {
```
ksercs: Now it's used only in 2 rows, let's get rid of it

packages/devextreme/js/ui/menu/ui.menu.js
iBat: Yes, we have different values in show and hide events. For show this is the element which we want to expand. For hide - submenu which we are closing. And there is a logic in this.

ksercs: Okay, let's leave it as is for now

packages/devextreme/js/ui/context_menu/ui.context_menu.js
```diff
@@ -555,9 +555,23 @@ class ContextMenu extends MenuBase {
...
+            const scrollableInstance = $scrollableElement.dxScrollable('instance');
+
+            scrollableInstance.scrollTo(0);
         }
     }
 
+    _getRootScrollableElement() {
+        return this._overlay.$content().find(`.${SCROLLABLE_CLASS}`).first();
+    }
+
```
ksercs: It's outdated changes, should be rebased

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -619,66 +619,84 @@ class Menu extends MenuBase {
...
-    _submenuOnShowingHandler($rootItem, submenu) {
-        const $border = $rootItem.children(`.${DX_CONTEXT_MENU_CONTAINER_BORDER_CLASS}`);
+    _submenuOnShowingHandler($menuAnchorItem, submenu, { rootItem }) {
+        const $border = $menuAnchorItem.children(`.${DX_CONTEXT_MENU_CONTAINER_BORDER_CLASS}`);
 
-        this._actions.onSubmenuShowing({
-            rootItem: getPublicElement($rootItem),
-            submenu: submenu
-        });
+        this._actions.onSubmenuShowing(this._getVisibilityChangeEventParams(rootItem, submenu, $menuAnchorItem));
```
ksercs: It's better to place params getting in a variable to simplify future debug

ksercs (approved):

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -1449,6 +1449,119 @@ QUnit.module('Menu tests', {
...
+            }],
+            onSubmenuShowing: ({ submenuContainer }) => {
+                $(submenuContainer).css('maxHeight', 42);
+                $(submenuContainer).css('boxSizing', 'border-box');
+            },
+        };
+        const menu = createMenuInWindow(options);
+        const $rootItem = $(menu.element).find(`.${DX_MENU_ITEM_CLASS}`).eq(0);
+
+        // show submenu
```
EugeniyKiyashko, line 1548: Let's remove comment

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -619,66 +619,89 @@ class Menu extends MenuBase {
...
 
-        args.rootItem = getPublicElement($rootItem);
-        args.submenu = submenu;
+        args.itemData = params.itemData;
+        args.rootItem = params.rootItem;
+        args.submenuContainer = params.submenuContainer;
+        args.submenu = params.submenu;
 
         this._actions.onSubmenuHiding(args);
         eventArgs = args;
```
EugeniyKiyashko, line 668: Let's simplify logic with args = eventArgs, eventArgs = args

packages/devextreme/js/ui/menu/ui.menu.js
```diff
@@ -619,66 +619,89 @@ class Menu extends MenuBase {
...
+        let itemData;
+        let $submenuContainer;
+
+        if(submenuItem) {
+            const anchor = isHide ? $(submenuItem).closest(`.${DX_MENU_ITEM_CLASS}`)[0] : submenuItem;
+
+            itemData = this._getItemData(anchor);
+            $submenuContainer = $(anchor).find(`.${DX_SUBMENU_CLASS}`).first();
+        } else {
+            const $overlayContent = $(submenu._overlay?.content());
```
EugeniyKiyashko, line 693: Why we should use ?. here?

PR #27209, opened 2024-04-19
Splitter: fix AXE scrollable accessibility issue

nikkithelegendarypokemonster:
**Description:**
The main problem is there is axe scrollable accessibility issue. Issue exist because item content has no focusable content inside of it thats why it make keyboard navigation not have ability to scroll inside.

**Solution:**
I applied added attribute tabindex in pane-content to make the whole content focusable allowing user the ability to scroll instead of adding a focusable item (e.g a focusable button) in the pane which it doesnt follow our original prototype design.


apps/demos/Demos/Splitter/Overview/React/PaneContent.tsx
```diff
@@ -17,7 +17,7 @@ const getFilteredDimensionOptions = (data) => Object.entries(data)
   .map(([key, value]) => ({ key, value }));
 
 const PaneContent = (data) => (
-  <div className="pane-content">
+  <div className="pane-content" tabIndex={0}>
```
ksercs, line 20: and what about demo for ReactJS?

ksercs (dismissed):

apps/demos/Demos/Splitter/Overview/React/PaneContent.tsx
nikkithelegendarypokemonster: [72a19e6](https://github.com/DevExpress/DevExtreme/pull/27209/commits/72a19e6b474cc978034926f838f491d76f49a52e)

ksercs (dismissed):

apps/demos/testing/common.test.js
```diff
@@ -138,6 +138,7 @@ const SKIPPED_TESTS = {
   React: {
     Charts: [
       { demo: 'PiesWithEqualSize', themes: [THEME.material] },
+      { demo: 'CustomAnnotations', themes: [THEME.material] },
```
ksercs, line 141: Seems like only vectorMap test should be skipped, no? 
Do not make excess skips please

ksercs: Okay I see this test is unstable as well

nikkithelegendarypokemonster: the error exists on Charts so its necessary to include it on Charts

nikkithelegendarypokemonster: The test error exists in Charts so its necessary to add the skipped test there instead of Vector Map

ksercs (approved):

EugeniyKiyashko (approved):

PR #27222, opened 2024-04-23
Fix use of cross-platform imports in react demos

tomodasheesh:


VasilyStrelyaev (dismissed):

apps/demos/Demos/Common/EditorAppearanceVariants/React/App.tsx
```diff
@@ -1,13 +1,17 @@
 import React, { useCallback, useState } from 'react';
+
+import { EditorStyle, LabelMode } from 'devextreme-react/common';
+
+import notify from 'devextreme/ui/notify';
+
 import SelectBox from 'devextreme-react/select-box';
```
mpreyskurantov: Could we use named imports for ui / components everywhere (instead of default)?

```
//import SelectBox from 'devextreme-react/select-box';

import { SelectBox } from 'devextreme-react/select-box';
```

apps/demos/Demos/Form/CustomizeItem/React/App.tsx
```diff
@@ -1,13 +1,18 @@
...
-  Item, GroupItem, Label, FormTypes,
+import { ValidationRule } from 'devextreme-react/common';
+
+import {
+  Form,
+  Item, GroupItem, Label,
+  FormTypes,
 } from 'devextreme-react/form';
-import { ValidationRule } from 'devextreme/common';
+import 'devextreme-react/text-area';
```
mpreyskurantov, line 10: strange import btw
I see that it is simply re-ordered. I am wondering, how it is used.

VasilyStrelyaev: I know, I know 😄 

There is a form in this demo with `editorType="dxSelectBox"`,  `editorType="dxDataEditor"` and `editorType="dxTextArea"`. So, this tells the form to render specific editor components.

Some editor components are already referenced by the `form` module, while others should be explicitly referenced in user's code.

https://js.devexpress.com/React/Documentation/ApiReference/UI_Components/dxForm/Item_Types/SimpleItem/#editorType

I guess a side-effectful import is enough, so it is just written here as a side effect.

mpreyskurantov (approved):

VasilyStrelyaev (approved):

PR #27303, opened 2024-04-30
Downgrade devextreme-angular to Angular 16

GoodDayForSurf:


packages/devextreme-angular/gulpfile.js
```diff
@@ -192,6 +184,30 @@ gulp.task('clean.tests', () => {
   return del([outputFolderPath]);
 });
 
+gulp.task('add.package.json', (done) => {
```
dxvladislavvolkov: I am not sure that we need to do it. We need to test native solution of ng-packagr

GoodDayForSurf: I've refactored this

dxvladislavvolkov (approved):

PR #27305, opened 2024-05-02
Splitter: items with no size should have visibility hidden to prevent focus on tab

Zedwag:


packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -810,9 +811,16 @@ class Splitter extends (CollectionWidget as any) {
     });
   }
 
-  _applyFlexGrowFromLayout(layout: number[]): void {
+  _applyStylesFromLayout(layout: number[]): void {
     this._iterateItems((index, itemElement) => {
       setFlexProp(itemElement, FLEX_PROPERTY.flexGrow, layout[index]);
+
+      const $item = $(itemElement);
+      const itemSize = this._getItemDimension(itemElement);
```
EugeniyKiyashko: Use `layout[index]` instead of `_getItemDimension`.

packages/devextreme/scss/widgets/base/splitter/_index.scss
```diff
@@ -9,6 +9,10 @@
     display: flex;
     overflow: hidden;
     flex: 1 1 auto;
+
+    &.dx-splitter-item-hidden-content {
```
EugeniyKiyashko: Let's move out of `.dx-splitter-item`

EugeniyKiyashko (approved):

PR #27309, opened 2024-05-02
Fix skipped visual demos tests

iBat:


apps/demos/testing/widgets/datagrid/EditStateManagement.test.js
```diff
@@ -14,7 +14,7 @@ runManualTest('DataGrid', 'EditStateManagement', ['jQuery', 'React', 'Vue', 'Ang
     const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
 
     // edit row
-    await t.click($('a').withText('Edit').nth(0));
+    await t.click($('.dx-link-edit').nth(0));
```
EugeniyKiyashko, line 17: Usually we use constant variables for class names to avoid duplication. But, these tests belong to the grid squad, let them decide if it's important for them or not

apps/demos/utils/visual-tests/matrix-test-helper.js
```diff
@@ -175,17 +175,17 @@ const SKIPPED_TESTS = {
       { demo: 'TooltipHTMLSupport', themes: [THEME.material] },
     ],
     DataGrid: [
-      { demo: 'BatchUpdateRequest', themes: [THEME.fluent, THEME.material] },
-      { demo: 'ColumnCustomization', themes: [THEME.fluent] },
       { demo: 'CellEditingAndEditingAPI', themes: [THEME.material] },
+      // It looks like we don't have a "close" button in the Fluent theme.
+      // Need need to update the behavior in Fluent or find another common way for all topics.
```
EugeniyKiyashko, line 180: It looks like a oversight in the Fluent theme development. It would be better to involve @Shuvaevmedia  to check the UI/UX of the DataGrid's field chooser in Fluent.

EugeniyKiyashko (approved):

apps/demos/testing/widgets/datagrid/EditStateManagement.test.js
pomahtri: We need to use our testcafe models here, we will refactor it when we'll have time (not soon)

apps/demos/utils/visual-tests/matrix-test-helper.js
ksercs: I prefer we left this on Grids responsibility. They are already aware of necessity to fix these tests.

ksercs (approved):

PR #27313, opened 2024-05-06
Fix react wrapper guard handlers (T1226902)

tomodasheesh:


ivanblinov2k17 (dismissed):

VasilyStrelyaev (dismissed):
Please see my comments in the second PR

packages/devextreme-react/src/core/__tests__/props-updating.test.tsx
```diff
@@ -717,6 +717,46 @@ describe('cfg-component option control', () => {
...
+    rerender(<TestContainer value={123} />);
+    jest.runAllTimers();
+
+    expect(Widget.option).toHaveBeenCalledWith('nestedOption.a', 123);
+  });
+
+  it('invokes option change guard handlers in strict mode', () => {
+    const TestContainer = ({ value }: { value: number }) => {
+      return (
+        <React.StrictMode>
```
VasilyStrelyaev, line 723: I guess we can leave `<React.StrictMode>`, just please note a different way to activate the Strict Mode in tests:

https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme-react/src/core/__tests__/component.test.tsx#L60

Although it's maybe a little bit more verbose because it requires the option to be reset to `false` on `afterEach`.

And this new syntax isn't available in `23_2` because I just recently updated the `@testing-library/react` dependency in `24_1`

tomodasheesh: This is noted.

ivanblinov2k17 (approved):

VasilyStrelyaev (approved):

PR #27314, opened 2024-05-06
Fix react wrapper guard handlers in 23_2  (T1226902)

tomodasheesh:


packages/devextreme-react/src/core/__tests__/props-updating.test.tsx
```diff
@@ -676,6 +676,46 @@ describe('cfg-component option control', () => {
     expect(Widget.option.mock.calls[1]).toEqual(['nestedOption.b', 'const']);
   });
 
+  it('invokes option change guard handlers', () => {
```
VasilyStrelyaev: This test is okay, just please explain what is different in this test compared to this one:

https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme-react/src/core/__tests__/props-updating.test.tsx#L607

Looks like it is almost the same, this test just calls a re-render, which, considering that we call `jest.runAllTimers`, should not make a difference.

packages/devextreme-react/src/core/__tests__/props-updating.test.tsx
```diff
@@ -676,6 +676,46 @@ describe('cfg-component option control', () => {
...
+    rerender(<TestContainer value={123} />);
+    jest.runAllTimers();
+
+    expect(Widget.option).toHaveBeenCalledWith('nestedOption.a', 123);
+  });
+
+  it('invokes option change guard handlers in strict mode', () => {
+    const TestContainer = ({ value }: { value: number }) => {
+      return (
+        <React.StrictMode>
```
VasilyStrelyaev, line 701: I guess we can leave `<React.StrictMode>`, just please note a different way to activate the Strict Mode in tests:

https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme-react/src/core/__tests__/component.test.tsx#L60

Although it's maybe a little bit more verbose because it requires the option to be reset to `false` on `afterEach`

VasilyStrelyaev (dismissed):

packages/devextreme-react/src/core/__tests__/props-updating.test.tsx
tomodasheesh: Yes, this test is similar to the test i've done. I'll remove mine out.

ivanblinov2k17 (approved):

VasilyStrelyaev (approved):

PR #27330, opened 2024-05-08
getEventTarget: Fix getting element from nullish path

marker-dao:


Zedwag (dismissed):

packages/devextreme/js/events/utils/event_target.js
```diff
@@ -6,12 +6,13 @@ export const getEventTarget = (event) => {
...
 
     const isShadowDOMUsed = Boolean(originalEvent.target?.shadowRoot);
+
     if(!isShadowDOMUsed) {
         return originalEvent.target;
     }
 
     const path = originalEvent.path ?? originalEvent.composedPath?.();
-    const target = path[0];
+    const target = path?.[0];
```
ksercs: Are you sure it's okay to return undefined in this case?
I think it's more safe to return `event.target` (do nothing like we are not in shadowDOM)

marker-dao: Yes, but it's incorrect behavior too: if target is incorrect, we get error which I fixed here: [https://github.com/DevExpress/DevExtreme/pull/27189]. Just a note. I will replace undefined with event.target.

ksercs: As far as i understand in that case path always exists, not?

ksercs (approved):

packages/devextreme/js/events/utils/event_target.js
marker-dao: [71ea9871bd24d47524de443f41651d59038bc0e1]

Zedwag (approved):

PR #27331, opened 2024-05-08
List: fix Knockout issue with console error on selectAll click

Zedwag:


packages/devextreme/js/ui/list/ui.list.edit.decorator.selection.js
```diff
@@ -167,6 +167,10 @@ registerDecorator(
         },
 
         _updateSelectAllAriaLabel() {
+            if(!this._$selectAll) {
```
ksercs, line 170: P.s we gonna deal with knockout test separately

ksercs (approved):

PR #27356, opened 2024-05-12
Splitter: improve behavior on window resize

EugeniyKiyashko:


e2e/testcafe-devextreme/tests/navigation/splitter/resize.ts
```diff
@@ -19,7 +19,7 @@ safeSizeTest('non resizable pane should not change its size during resize', asyn
 
   await t
     .expect(splitter.getItem(2).element.clientWidth)
-    .eql(300);
+    .eql(145);
```
AlexanderMoiseev, line 22: why non resizable pane changes its size? the test checks that it should not

AlexanderMoiseev (approved):

PR #27397, opened 2024-05-15
DataGrid: Fix focus stuck in grid when first column is fixed and it contains a link (T1216832)

Alyar666:
See https://devexpress.com/issue=T1216832


packages/devextreme/js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
```diff
@@ -1148,6 +1149,18 @@ const resizing = (Base: ModuleType<ResizingController>) => class ResizingColumnF
   }
 };
 
+const keyboardNavigation = (Base: ModuleType<KeyboardNavigationController>) => class KeyboardNavigationExtender extends Base {
+  protected _toggleInertAttr(value = false): void {
```
williamvinogradov: Let's remove the default argument value
Because we don't use this method without passing the value + passing the value is better than the default one defined in the method (more obvious)

```suggestion
  protected _toggleInertAttr(value: boolean): void {
```

packages/devextreme/js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
```diff
@@ -1148,6 +1149,18 @@ const resizing = (Base: ModuleType<ResizingController>) => class ResizingColumnF
...
+    const $fixedContent = this._rowsView?.element()?.children(`.${this.addWidgetPrefix(CONTENT_FIXED_CLASS)}`);
```
williamvinogradov: Let's add to the ```rowsView``` extender (in the```m_column_fixing```) the method that returns fixed content element.

Something like:
```ts
const rowsView = (Base: ModuleType<RowsView>) => class RowsViewFixedColumnsExtender extends baseFixedColumns(Base) {
  // other methods & fields
  public getFixedContentElement(): dxElementWrapper {
    const fixedContentClass = this.addWidgetPrefix(CONTENT_FIXED_CLASS);
    return this.element()?.children(`.${fixedContentClass}`);
```
And use this method here:
```suggestion
    const $fixedContent = this._rowsView?.getFixedContentElement();
```

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -160,6 +162,8 @@ export class KeyboardNavigationController extends modules.ViewController {
     this.focusedHandlerWithContext = this.focusedHandlerWithContext || this.focusedHandler.bind(this);
     this.renderCompletedWithContext = this.renderCompletedWithContext || this.renderCompleted.bind(this);
     this.rowsViewFocusHandlerContext = this.rowsViewFocusHandlerContext || this.rowsViewFocusHandler.bind(this);
+    // eslint-disable-next-line max-len
+    this.rowsViewFocusOutHandlerContext = this.rowsViewFocusOutHandlerContext ?? this.rowsViewFocusOutHandler.bind(this);
```
williamvinogradov: Let's remove the ```eslint-disable-next-line``` comment and just write the ```??``` on the next line:
```suggestion
    this.rowsViewFocusOutHandlerContext = this.rowsViewFocusOutHandlerContext 
      ?? this.rowsViewFocusOutHandler.bind(this);
```

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -688,14 +698,18 @@ export class KeyboardNavigationController extends modules.ViewController {
     }
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected _toggleInertAttr(value = false): void {}
```
williamvinogradov: Same comment for the default argument value

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -688,14 +698,18 @@ export class KeyboardNavigationController extends modules.ViewController {
...
     const editingOptions = this.option('editing');
     const direction = eventArgs.shift ? 'previous' : 'next';
     const isCellPositionDefined = isDefined(this._focusedCellPosition)
       && !isEmptyObject(this._focusedCellPosition);
+    const isLastValidCell = !eventArgs.shift && this._isLastValidCell(this._focusedCellPosition);
 
     let isOriginalHandlerRequired = !isCellPositionDefined
-      || (!eventArgs.shift && this._isLastValidCell(this._focusedCellPosition))
+      || isLastValidCell
       || (eventArgs.shift && this._isFirstValidCell(this._focusedCellPosition));
```
williamvinogradov: Minor refactoring [OPTIONAL].

Let's make the code more uniform and add the ```isFirstValidCell``` variable.
Something like:
```ts
const isFirstValidCell = eventArgs.shift && this._isFirstValidCell(this._focusedCellPosition);
const isLastValidCell = !eventArgs.shift && this._isLastValidCell(this._focusedCellPosition);

let isOriginalHandlerRequired = !isCellPositionDefined
      || isFirstValidCell
      || isLastValidCell;

File: packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/keyboardNavigation.accessibility.tests.js [hunk 6]
```diff
@@ -906,4 +906,74 @@ QUnit.module('Keyboard navigation accessibility', {
         assert.ok($headerItem.is(':focus'), 'Header cell has focus');
         assert.notOk($headersElement.hasClass('dx-state-focused'), 'Headers main element has no dx-state-focused class');
     });
+
+    // T1216832
+    testInDesktop('Navigation should not get stuck in the grid when the first cell is fixed and a template with links is specified for it', function(assert) {
```
williamvinogradov: Let's split this test case into two:
- The first one will check the ```navigationOfAllCells```
- The second one will check the ```inert``` attribute

And let's add the third test case like in the original ticket issue:
1) Focus on the last cell (in the DataGrid with fixed columns)
2) Press the ```Tab```
3) Check that DataGrid loses focus or that the element after the DataGrid gains focus

We will check the original issue case with this test case
(In this issue focus goes to the fixed column instead of "move out" from the DataGrid)

williamvinogradov: Question:
Why QUnit and not Testcafe? :)

williamvinogradov (approved):

pomahtri (approved):

PR #27408, opened 2024-05-16
SelectBox: Set appropriate "aria-invalid" attribute value when input is empty or not (T1230706)

jdvictoria:


jdvictoria:
[SelectBox - The input element doesn't contain the aria-invalid attribute after a value is changed via the keyboard if fieldTemplate is defined [T1230706]](https://trello.com/c/F6ajJ547/8815-selectbox-the-input-element-doesnt-contain-the-aria-invalid-attribute-after-a-value-is-changed-via-the-keyboard-if-fieldtemplate)

PR #27411, opened 2024-05-16
GridCore(T1194403): remove redundant refocus of the cells editor

williamvinogradov:


e2e/testcafe-devextreme/tests/dataGrid/focus/focusShowEditorAlwaysCell.ts
```diff
@@ -0,0 +1,113 @@
...
+  await takeScreenshot('focus-edit-cell_after-lookup-change.png', dataGrid.element);
+  await t.expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+}).before(async () => createDataGrid());
+
+test('Should switch focus after the textBox value change [T1194403]', async (t) => {
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const dataGrid = new DataGrid(SELECTOR);
+
+  const textEditorCellFirst = dataGrid.getDataCell(0, 1).getEditor();
```
Alyar666: In my opinion, the names of the variables do not correspond to reality. We need to swap the names: textEditorCellFirst <-> textEditorCellSecond.

williamvinogradov: Yup, agree with you that naming here creates confusion
I meant that the ```first``` - is the focused first cell (in layout this is the ```second``` cell)

So, I just renamed these variables to the ```One/Two``` instead of the ```First/Second``` :)

Alyar666 (approved):

PR #27431, opened 2024-05-21
Splitter: push api fixes

AlexanderMoiseev:


packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.push_api.tests.js
```diff
@@ -0,0 +1,361 @@
...
+        </style>
+        
+        <div id="splitter"></div>
+        <div id="splitterParentContainer">
+            <div id="splitterInContainer"></div>
+        </div>`;
+    $('#qunit-fixture').html(markup);
+});
+
+const moduleConfig = {
```
AlexanderMoiseev: probably its better to move this module to splitter.tests file

EugeniyKiyashko: Yep, it's better. It will allow us to avoid duplication.

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -291,6 +296,12 @@ class Splitter extends CollectionWidget<Properties> {
 
     this._getItemInstance($itemFrame)._renderResizeHandle();
 
+    const previousElement = $(itemElement).prev();
+    if (previousElement.length && !previousElement.hasClass(RESIZE_HANDLE_CLASS)) {
```
EugeniyKiyashko: We already have the methods that allow us to find previous or next panes.

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -276,6 +274,13 @@ class Splitter extends CollectionWidget<Properties> {
     return index === findLastIndexOfVisibleItem(items);
   }
 
+  _fireContentReadyAction(): void {
+    // @ts-expect-error badly typed base class
```
EugeniyKiyashko: We shouldn't leave the // @ts-expect-error comment. You can type this method inside edit.ts.

packages/devextreme/js/__internal/ui/splitter/splitter_item.ts
```diff
@@ -59,6 +59,13 @@ class SplitterItem extends CollectionWidgetItem {
   isLast(): boolean {
     return this._owner._isLastVisibleItem(this.getIndex());
   }
+
+  _dispose(): void {
+    // @ts-expect-error badly typed base class
```
EugeniyKiyashko, line 64: We shouldn't leave ignore comments.

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -308,10 +319,10 @@ class Splitter extends CollectionWidget<Properties> {
       const $rightItem = this._getResizeHandleRightItem($resizeHandle);
       const leftItemData = this._getItemData($leftItem);
       const rightItemData = this._getItemData($rightItem);
-      const resizable = leftItemData.resizable !== false
-        && rightItemData.resizable !== false
-        && leftItemData.collapsed !== true
-        && rightItemData.collapsed !== true;
+      const resizable = leftItemData?.resizable !== false
```
EugeniyKiyashko: Why is it necessary to check that left or right items are present? I believe that resizeHandle shouldn't exist if left or right items are missing.

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -218,9 +218,7 @@ class Splitter extends CollectionWidget<Properties> {
     this._shouldRecalculateLayout = false;
   }
 
-  _renderItems(items: Item[]): void {
-    super._renderItems(items);
-
+  updateItemsLayout(): void {
```
EugeniyKiyashko: This seems very unreliable, abandoning our usual 'renderItems' method and using 'contentReady' to finish what we didn't complete earlier. We can adjust or redraw on 'contentReady,' but we shouldn't move the main logic there—it might not be called, and then what happens? We need to find a better solution.

packages/devextreme/js/__internal/ui/splitter/splitter_item.ts
AlexanderMoiseev: we have more than 10 ts-expect-error statemenst left in our splitter component, i suggest fixing them separately. I dont think this one have a quick fix coz it looks like it should be added somewhere on a base level and has nothing to do with Push api

PR #27435, opened 2024-05-21
Form: Fix form validation error when items are cleared (T1233487)

jdvictoria:


jdvictoria:
[Form - Validation of the cleared Group causes an error [T1233487]](https://trello.com/c/nx8k1K46/8869-form-validation-of-the-cleared-group-causes-an-error-t1233487)

packages/devextreme/js/ui/validation_engine.js
```diff
@@ -27,6 +27,11 @@ const STATUS = {
     pending: 'pending'
 };
 
+const validGroupNames = [
```
ksercs: [naming]
I didn't get what do you mean when you say "validGroup". It can be not valid as well (containing failed rules).
Also we usually use UPPER_SNAKE_CASE for constants.
```suggestion
const NON_DELETABLE_GROUP_NAMES = [
```

P.s I write this not just as a feedback for you to not repeat it in the future.

packages/devextreme/js/ui/validation_engine.js
```diff
@@ -763,7 +768,7 @@ const ValidationEngine = {
 
     _shouldRemoveGroup(group, validatorsInGroup) {
         const isDefaultGroup = group === undefined;
-        const isValidationGroupInstance = group && group.NAME === 'dxValidationGroup';
+        const isValidationGroupInstance = group && validGroupNames.includes(group.NAME);
```
ksercs: [naming]
Previously `isValidationGroupInstance ` was a good name because this variable stored a boolean value which pointed if it's a validation group instance or not.
But after your change this naming is not actual because this variable now can be true even if it's not a validation group instance (in case when it's a Form instance).
```suggestion
        const isNonRemovableGroup = group && validGroupNames.includes(group.NAME);
```
P.s. I write this just as a feedback for you to not repeat it in the future

ksercs (changes requested):
[missed scenario, local solution]

This is only a local solution, it does not work if we specify Form [validationGroup](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxForm/Configuration/#validationGroup) property. (**Don't forget to add a test**)

Moreover, ideologically it's not very good that we make ValidationEngine knows about Form.

I suggest you adding `_isRemovable` prop to the groupConfig and set it to `true` when we add a group from Form and other places where it's expected to be non removable.
It will also simplify the condition in `_shouldRemoveGroup` and make it much more intuitive. 

Also I see that there is no test which checks that custom [ValidationGroup](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxValidationGroup/) is not removed after its validators are removed however this case [is handled](https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme/js/ui/validation_engine.js#L766) in the code. Please add a test for this.

packages/devextreme/js/__internal/ui/m_validation_engine.ts
```diff
@@ -778,22 +780,21 @@ const ValidationEngine = {
...
+    const groupConfig = ValidationEngine.addGroup(group, false);
     groupConfig.registerValidator.call(groupConfig, validator);
   },
 
-  _shouldRemoveGroup(group, validatorsInGroup) {
+  _shouldRemoveGroup(group, validators, isRemovable) {
     const isDefaultGroup = group === undefined;
-    const isValidationGroupInstance = group && group.NAME === 'dxValidationGroup';
-    return !isDefaultGroup && !isValidationGroupInstance && !validatorsInGroup.length;
+    return !isDefaultGroup && !validators.length && !isRemovable;
```
ksercs: [incorrect logic]
Now you return `true` here from `_shouldRemoveGroup` when it's `!isRemovable` (**not** removable). Vice versa,  we should be able to remove when it's removable! Please, reverse true/false value for this variable: it should be true when group is removable and false when it's not removable.

packages/devextreme/js/__internal/ui/m_validation_engine.ts
```diff
@@ -562,10 +563,11 @@ const ValidationEngine = {
     this.addGroup();
   },
 
-  addGroup(group) {
+  addGroup(group, isRemovable) {
     let config = this.getGroupConfig(group);
     if (!config) {
       config = new GroupConfig(group);
+      config.isRemovable = isRemovable;
```
ksercs: [broken encapsulation]
We should not set object value out of the class, it's supposed to be done in the constructor (or using class's public API but in our case it's excess while this variable cannot be changed at runtime). 
```javascript
ctor(group, isRemovable = true) {
    ...
    this._isRemovable = isRemovable;
}
```

packages/devextreme/js/__internal/ui/m_validation_engine.ts
```diff
@@ -778,22 +780,21 @@ const ValidationEngine = {
...
 
   registerValidatorInGroup(group, validator) {
-    const groupConfig = ValidationEngine.addGroup(group);
+    const groupConfig = ValidationEngine.addGroup(group, false);
     groupConfig.registerValidator.call(groupConfig, validator);
   },
 
-  _shouldRemoveGroup(group, validatorsInGroup) {
+  _shouldRemoveGroup(group, validators, isRemovable) {
     const isDefaultGroup = group === undefined;
```
ksercs: [missed refactoring]
Now we can skip this special check for default group, we can just set `isRemovable=false` when we create it

packages/devextreme/js/__internal/ui/m_validation_engine.ts
ksercs: [code style]
Please, name it starting with underscore: `_isRemovable` because it gonna be a private option. 
P.s. groupConfig can be taken by user using [this method](https://js.devexpress.com/jQuery/Documentation/ApiReference/Common/Utils/validationEngine/#getGroupConfiggroup)

packages/devextreme/js/__internal/ui/m_validation_group.ts
```diff
@@ -19,7 +19,7 @@ class ValidationGroup extends DOMComponent {
   _init() {
     // @ts-expect-error
     super._init();
-    ValidationEngine.addGroup(this);
+    ValidationEngine.addGroup(this, false);
```
ksercs, line 22: [missing test]
I changed it to true and no test is red. 
It should be tested in `validationGroup.tests`

ksercs, line 22: [broken scenario]
Custom ValidationGroup should not be removed when its validators are removed. It was supported [there](https://github.com/DevExpress/DevExtreme/pull/27435/files#diff-64529d0ddad2c54300c7f438acc04e59f59c839d7b324b6ded1f630639c8a014L787). We should not break it.

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.validationRules.tests.js
```diff
@@ -475,6 +475,62 @@ QUnit.test('Changing an validationRules options of an any item does not invalida
...
+        try {
+            form.validate();
+            assert.ok(true);
+        } catch(e) {
+            assert.ok(false, 'error is thrown');
+        }
+    });
+});
+
+QUnit.test('validation group should not be removed after its validators are removed (T1233487)', function(assert) {
```
ksercs: [Excess code]
I didn't get what u'r testing here. "validation group should not be removed after its validators are removed" - that's checked in the test above, and you even don't assert it here.
`assert.equal(validator.option('validationGroup'), 'Test', 'validation group should not be removed');` - u'r just checking that option didn't change. But it had no chance to be changed as it's just a group name passed on init ..

Explain please the idea what you tried to test, maybe i'm missing smth

ksercs (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.validationRules.tests.js
jdvictoria: Sorry, I thought it was supposed to be a separate test based on what you said above. 

"Also I see that there is no test which checks that custom [ValidationGroup](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxValidationGroup/) is not removed after its validators are removed however this case [is handled](https://github.com/DevExpress/DevExtreme/blob/24_1/packages/devextreme/js/ui/validation_engine.js#L766) in the code. Please add a test for this."

ksercs: Got it. But this scenario is not related to form, only to ValidationGroup component which you even don't use here in your test. Test i said about should check be red if we change this row: https://github.com/DevExpress/DevExtreme/pull/27435#discussion_r1640008793

packages/devextreme/js/__internal/ui/m_validation_engine.ts
jdvictoria: It seems that removing the check for default group breaks some test.
image

jdvictoria: Update: Already fixed it, seems to be flaky only.

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.validationRules.tests.js
jdvictoria: [f8e66aa](https://github.com/DevExpress/DevExtreme/pull/27435/commits/f8e66aaf1e46ace7d774494d1b930d32c43b8b69)

[c874986](https://github.com/DevExpress/DevExtreme/pull/27435/commits/c8749860b9d3db20aa0fb5abd1acd1063392b8c1)

packages/devextreme/js/__internal/ui/m_validation_group.ts
jdvictoria: [c874986](https://github.com/DevExpress/DevExtreme/pull/27435/commits/c8749860b9d3db20aa0fb5abd1acd1063392b8c1)

packages/devextreme/js/__internal/ui/m_validation_engine.ts
jdvictoria: [9cf2b5d](https://github.com/DevExpress/DevExtreme/pull/27435/commits/9cf2b5dbbeeebb0ab7f8c03aa544c69a576b74ea)

jdvictoria: [9cf2b5d](https://github.com/DevExpress/DevExtreme/pull/27435/commits/9cf2b5dbbeeebb0ab7f8c03aa544c69a576b74ea)

packages/devextreme/js/__internal/ui/m_validation_engine.ts
jdvictoria: [9cf2b5d](https://github.com/DevExpress/DevExtreme/pull/27435/commits/9cf2b5dbbeeebb0ab7f8c03aa544c69a576b74ea)

packages/devextreme/js/__internal/ui/m_validation_engine.ts
jdvictoria: [bcf7ab2](https://github.com/DevExpress/DevExtreme/pull/27435/commits/bcf7ab24f7d0f52b0117df135f481cd21f72b335)

packages/devextreme/js/__internal/ui/m_validation_group.ts
jdvictoria: Question - should we cover a custom validationGroup or just simply check validationGroup? Also there are things that I found

1. I think this [line](https://github.com/DevExpress/DevExtreme/pull/27435/files#diff-64529d0ddad2c54300c7f438acc04e59f59c839d7b324b6ded1f630639c8a014L787) seems to not work well with this `ValidationEngine.addGroup(this, false)` code we added. This makes the test evergreen regardless we make validationGroup removable or not.
2. Based from what I understand, validationGroup doesn't support custom validationGroup from [this](https://github.com/DevExpress/DevExtreme/blob/5a2883fa56ff15b64b0c24d79f84b0228c075876/packages/devextreme/js/__internal/ui/m_validation_group.ts#L22). Unlike form where it gets [this](https://github.com/DevExpress/DevExtreme/blob/5a2883fa56ff15b64b0c24d79f84b0228c075876/packages/devextreme/js/__internal/ui/form/m_form.ts#L705), should we rework how validationGroup passes group?
3. It seems that validationGroup always passed "dxValidationGroup", I think that it's sufficient that we just know if it's removable or not and proceed to omit checking if it's a validGroupInstance.

Let me know if I'm missing something out.

ksercs: 1. "I think this [line](https://github.com/DevExpress/DevExtreme/pull/27435/files#diff-64529d0ddad2c54300c7f438acc04e59f59c839d7b324b6ded1f630639c8a014L787) seems to not work well with this ValidationEngine.addGroup(this, false) code we added." Yeah, it makes. But it should be removed after we change our check to use `groupConfig._isRemovable`.  I see you already [updated the check](https://github.com/DevExpress/DevExtreme/pull/27435/files#diff-64529d0ddad2c54300c7f438acc04e59f59c839d7b324b6ded1f630639c8a014R787).
2. Didn't get ur question, sorry. The second link does not work.
3. Yeah, that was an initial idea of my note. We should just pass "isRemovable=false" here when we add a groupConfig from ValidationGroup component, and that's should be enough. But during my previous review "ValidationEngine.addGroup(this, false);" meant that this group is removable. That's why i noted

packages/devextreme/js/__internal/ui/m_validation_engine.ts
```diff
@@ -778,22 +779,21 @@ const ValidationEngine = {
   },
 
   registerValidatorInGroup(group, validator) {
-    const groupConfig = ValidationEngine.addGroup(group);
+    const groupConfig = ValidationEngine.addGroup(group, true);
```
ksercs: [minor]
Let's instead of this add a [default parameter value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) for `isRemovable` parameter in `addGroup` function
```suggestion
    const groupConfig = ValidationEngine.addGroup(group);
```

packages/devextreme/js/__internal/ui/m_validation_engine.ts
```diff
@@ -778,22 +779,21 @@ const ValidationEngine = {
...
     groupConfig.registerValidator.call(groupConfig, validator);
   },
 
-  _shouldRemoveGroup(group, validatorsInGroup) {
-    const isDefaultGroup = group === undefined;
-    const isValidationGroupInstance = group && group.NAME === 'dxValidationGroup';
-    return !isDefaultGroup && !isValidationGroupInstance && !validatorsInGroup.length;
+  _shouldRemoveGroup(validatorsInGroup, isRemovable) {
```
ksercs: [code splitting, refactoring]
Now this function is pretty small and does nothing expect logical operator with parameter + it's used in the single place. 
So, i suggest u getting rid of it and make this check directly before [this condition](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters).
```javascript
const shouldRemoveGroup = validatorsInGroup.length === 0 && isRemovable;
if (shouldRemoveGroup) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
```diff
@@ -305,6 +304,8 @@ QUnit.module('General', {
             assert.ok(isValid, 'validation is correct');
         });
         ValidationEngine.removeRegisteredValidator(group, validator2);
+
+        assert.ok(ValidationEngine.getGroupConfig(group), 'group is not removed');
```
ksercs, line 321: [atomic tests]
Please, write a separate test for this. One test = one idealogical check!
Moreover, this check is not related to T1006667 bug ticket mentioned in this old test name.

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.validationRules.tests.js
```diff
@@ -475,6 +475,33 @@ QUnit.test('Changing an validationRules options of an any item does not invalida
     assert.strictEqual(renderComponentSpy.callCount, 0, 'renderComponentSpy.callCount');
 });
 
+[false, true].forEach(hasValidationGroup => {
+    QUnit.test(`form ${hasValidationGroup ? 'with' : 'without'} validation group should validate without errors (T1233487)`, function(assert) {
```
ksercs: [test naming]
Scenario is not completely described. 
```suggestion
    QUnit.test(`form ${hasValidationGroup ? 'with' : 'without'} validationGroup property specified should validate without errors after nested validators remove (T1233487)`, function(assert) {
```

packages/devextreme/js/__internal/ui/m_validation_group.ts
ksercs: @jdvictoria  this note is still actual. If we change `false` to `true` here no test is red => your new test is evergreen

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
jdvictoria: Assert seems to fail when isRemovable = true

image

jdvictoria: image

ksercs: Well, that was problem on my side. I will remove an appropriate note. But this (about atomic test) is still actual

packages/devextreme/js/__internal/ui/m_validation_engine.ts
jdvictoria: [830b39d](https://github.com/DevExpress/DevExtreme/pull/27435/commits/830b39d99808ae11c2786a081a69c3d3d05517a4)

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.validationRules.tests.js
jdvictoria: [daf73a4](https://github.com/DevExpress/DevExtreme/pull/27435/commits/daf73a4cac0e007e5fc9a5f13c3cb1498f71d603)

packages/devextreme/js/__internal/ui/m_validation_engine.ts
jdvictoria: [913f413](https://github.com/DevExpress/DevExtreme/pull/27435/commits/913f41339e3295625f6fa15bc9fb8a0f1abcdb35)

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/validationGroup.tests.js
jdvictoria: [9b86a2e](https://github.com/DevExpress/DevExtreme/pull/27435/commits/9b86a2e8e5b2a547a739e7a9b25da52911d3f557)

packages/devextreme/js/__internal/ui/m_validation_engine.ts
jdvictoria: I don't know if I'm missing something or it really causes an error
image
image
image

packages/devextreme/js/__internal/ui/m_validation_summary.ts
```diff
@@ -37,7 +37,7 @@ const ValidationSummary = CollectionWidget.inherit({
     const $element = this.$element();
     const group = this.option('validationGroup')
             || ValidationEngine.findGroup($element, this._modelByElement($element));
-    const groupConfig = ValidationEngine.addGroup(group);
+    const groupConfig = ValidationEngine.addGroup(group, false);
```
ksercs: [mostly memory leak]
I don't see there is a call of `removeGroup` in ValidationSummary => it's more safe to leave this group removable. Otherwise we just never remove this group..

packages/devextreme/js/__internal/ui/m_validation_engine.ts
```diff
@@ -562,10 +563,10 @@ const ValidationEngine = {
     this.addGroup();
```
ksercs: [broken behavior]
This default group should be not removable. It was previously handled [here](https://github.com/DevExpress/DevExtreme/pull/27435/files#diff-64529d0ddad2c54300c7f438acc04e59f59c839d7b324b6ded1f630639c8a014L786).
Please, set isRemovable=false and write a test

ksercs (changes requested):

packages/devextreme/js/__internal/ui/m_validation_engine.ts
ksercs: That's because of this: https://github.com/DevExpress/DevExtreme/pull/27435#discussion_r1646386039.

packages/devextreme/js/__internal/ui/m_validation_engine.ts
jdvictoria: [dce6037](https://github.com/DevExpress/DevExtreme/pull/27435/commits/dce603761de8298fa7f7c8b136b4c369491a6cc4)

I believe the test is already being handled in validationEngine.test - Groups: Simple group - remove validator registration

packages/devextreme/js/__internal/ui/m_validation_summary.ts
jdvictoria: [ed5e8ca](https://github.com/DevExpress/DevExtreme/pull/27435/commits/ed5e8cae5a8f976fdac4650f50c05bc8bfe2b995)

packages/devextreme/js/__internal/ui/m_validation_engine.ts
jdvictoria: [a3d3b51](https://github.com/DevExpress/DevExtreme/pull/27435/commits/a3d3b51763ff995849707aac971fb31bb597c4d3)

ksercs (approved):

EugeniyKiyashko (approved):

PR #27439, opened 2024-05-22
Fix React TreeList Memory Leak (T1230753)

tomodasheesh:


VasilyStrelyaev (dismissed):

packages/devextreme-react/src/core/helpers.ts
```diff
@@ -12,7 +12,7 @@ export function generateID(): string {
 }
 
 export class DoubleKeyMap<TKey1, TKey2, TValue> {
-  private _map: Map<TKey1, Map<TKey2, TValue>> = new Map();
+  readonly _map: Map<TKey1, Map<TKey2, TValue>> = new Map();
```
ivanblinov2k17: i believe, this should be left as private, because with readonly, u will still be able to access _map and modify it by calling methods

tomodasheesh: I agree with you. I initially left it as private, but the linter throws an error.
image

VasilyStrelyaev: I think Ivan meant `private readonly`:

```js
private readonly _map: Map<TKey1, Map<TKey2, TValue>> = new Map();
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/selection.tests.js
```diff
@@ -3926,80 +3926,104 @@ QUnit.module('Selection with views', {
...
     // T550013
     QUnit.test('Click on selected selection checkbox with shift key', function(assert) {
-        const testElement = $('#container');
+        const clock = sinon.useFakeTimers();
 
-        this.options.selection.showCheckBoxesMode = 'onClick';
-        this.setup();
-        this.columnHeadersView.render(testElement);
-        this.rowsView.render(testElement);
+        try {
```
williamvinogradov, line 3931: Question.

Why do we add the ```try``` blocks in this test fixture?

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/selection.tests.js
```diff
@@ -3926,80 +3926,104 @@ QUnit.module('Selection with views', {
 
     // T550013
     QUnit.test('Click on selected selection checkbox with shift key', function(assert) {
-        const testElement = $('#container');
+        const clock = sinon.useFakeTimers();
```
williamvinogradov, line 3929: Question.

Can we use the ```beforeEach``` / ```afterEach``` to setup / restore fake timers?
Like here: [link](https://github.com/DevExpress/DevExtreme/pull/27439/files/a654f1dd326eb546565f7079f567ce9595e713f9..5e888a939fde8dfb2ed1ec728e0f20e8569b0ee6#diff-9aed3da6b88a7a29b47d40097ba2d19653f0cee5645bac366ccbe1b3f4557c47R4089)

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/selection.tests.js
VasilyStrelyaev: to make sure `clock.restore()` always executes, even in case of exceptions

there are lot of examples across our code base:
https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/testing/tests/DevExpress.ui/collectionWidgetParts/editingTests.js#L496

whenever we need cleanup inside test body, we wrap the cleanup in `finally` and the rest in `try`

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/selection.tests.js
VasilyStrelyaev: In the case you refer to, different tests failed in that module, so I had to migrate all of them to fake timers.

In this case, only specific tests failed, so I migrated just them.

williamvinogradov (approved):

VasilyStrelyaev (approved):

PR #27458, opened 2024-05-24
SelectBox: Set "role" when fieldTemplate is rendered (T1230635)

jdvictoria:


packages/devextreme/js/ui/drop_down_editor/ui.drop_down_list.js
```diff
@@ -487,12 +487,12 @@ const DropDownList = DropDownEditor.inherit({
...
-        this.setAria({
-            'haspopup': 'listbox',
-            'autocomplete': 'list',
-            'role': 'combobox'
-        });
+    _getAriaHasPopup() {
+        return 'true';
+    },
+
+    _getAriaAutocomplete() {
```
EugeniyKiyashko: This overload uses the same value as in the parent

packages/devextreme/js/ui/drop_down_editor/ui.drop_down_list.js
```diff
@@ -487,12 +487,12 @@ const DropDownList = DropDownEditor.inherit({
...
     },
 
-    _setDefaultAria: function() {
-        this.setAria({
-            'haspopup': 'listbox',
-            'autocomplete': 'list',
-            'role': 'combobox'
-        });
+    _getAriaHasPopup() {
+        return 'true';
```
EugeniyKiyashko: this value should be 'listbox'. Is it right?

EugeniyKiyashko (approved):

PR #27473, opened 2024-05-28
Switch, Resizable, Map: move files to TS

EugeniyKiyashko:


packages/devextreme/js/__internal/ui/m_map.ts
```diff
@@ -1,440 +1,368 @@
...
-import { titleize } from '../core/utils/inflector';
-import { each } from '../core/utils/iterator';
-import { extend } from '../core/utils/extend';
-import { wrapToArray } from '../core/utils/array';
-import { isNumeric } from '../core/utils/type';
-import { addNamespace } from '../events/utils/index';
-import pointerEvents from '../events/pointer';
-import { noop } from '../core/utils/common';
-
+import registerComponent from '@js/core/component_registrator';
```
pomahtri, line 1: let's move `@ts/ui/m_map` to all other map files `@ts/ui/map/m_map`

pomahtri (approved):

packages/devextreme/js/__internal/ui/m_map.ts
EugeniyKiyashko: I will do this move in the next PR

Zedwag (approved):

PR #27475, opened 2024-05-28
SelectBox: Set "aria-labelledby" when fieldTemplate is rendered (T1230635)

jdvictoria:


packages/devextreme/js/ui/drop_down_editor/ui.drop_down_editor.js
```diff
@@ -417,6 +417,7 @@ const DropDownEditor = TextBox.inherit({
         this._refreshEvents();
         this._refreshEmptinessEvent();
         this._setDefaultAria();
+        this._label && this._setFieldAria();
```
EugeniyKiyashko: As I know, we render the TextEditor label even when the editor label is hidden. I don't think we should add the check for `_label` here. 
I understand that we will get an error inside the `_setFieldAria` method because `_label` will not be defined on the first call `_renderValue` when fieldTemplate is rendering.
Switching the calls to renderLabel and renderValue in texteditorbase is not safe because the rendering of the label may depend on the resulting size of the input after the value is set.

I suggest to consider this scenario inside _setFieldAria function something like this:
```
const labelId = this._label?.getId();
```
Please, try it. Let me know if this does not meet any conditions.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/dropDownEditor.tests.js
```diff
@@ -2197,6 +2198,36 @@ QUnit.module('aria accessibility', () => {
         assert.strictEqual($input.attr('role'), 'combobox', 'role attribute should remain assigned to the combobox after deleting');
     });
 
+    QUnit.test('component with fieldTemplate should have proper aria-labelledby attribute after interaction (T1230635)', function(assert) {
```
EugeniyKiyashko: We should consider the following scenarios:
- fieldTemplate is used without inputAttr option, label is defined
- fieldTemplate is used with inputAttr option, label is defined
- fieldTemplate is used, label is not defined

Let's add tests for this scenarios

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/selectBox.tests.js
```diff
@@ -3356,6 +3357,33 @@ QUnit.module('search', moduleSetup, () => {
...
+        });
+        let $input = $selectBox.find(toSelector(TEXTEDITOR_INPUT_CLASS));
+        const $textEditorLabel = $selectBox.find(toSelector(TEXTEDITOR_LABEL_CLASS));
+
+        assert.strictEqual($input.attr('aria-labelledby'), $textEditorLabel.attr('id'), 'initial render should set aria-labelledby equal to the label id');
+
+        const selectBox = $selectBox.dxSelectBox('instance');
+        const keyboard = keyboardMock($input);
+
+        keyboard.type('a');
```
EugeniyKiyashko: In this case, search is not working, and you choose an item from the previously rendered items. To make it work, we should use the searchTimeout: 0 property and use a symbol that is contained in our data. Otherwise, you will get a 'no data to display' message without the opportunity to select an item in the popup.

If you have the same mistakes in your previous written tests, let's fix them there too.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/selectBox.tests.js
```diff
@@ -73,6 +73,7 @@ const STATE_FOCUSED_CLASS = 'dx-state-focused';
 const TEXTEDITOR_BUTTONS_CONTAINER_CLASS = 'dx-texteditor-buttons-container';
```
EugeniyKiyashko, line 73: Please, fix lint errors https://github.com/DevExpress/DevExtreme/actions/runs/9284097588/job/25545738944?pr=27475

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/dropDownEditor.tests.js
```diff
@@ -2197,6 +2198,45 @@ QUnit.module('aria accessibility', () => {
         assert.strictEqual($input.attr('role'), 'combobox', 'role attribute should remain assigned to the combobox after deleting');
     });
 
+    [
```
EugeniyKiyashko, line 2201: Let's try to simplify it, for example, as follows. It's difficult to understand the mapping in the test title and inside with expectedId

```
[
        { label: 'Label' },
        { label: 'Label', inputAttr: { 'aria-label': 'Label' } },
        { label: undefined },
        { label: '' },
    ].forEach((options) => {
        QUnit.test(`component with fieldTemplate should have proper aria-labelledby attribute, options: ${JSON.stringify(options)} (T1230635)`, function(assert) {
            const $dropDownEditor = $('#dropDownEditorSecond').dxDropDownEditor({
                dataSource: ['one', 'two', 'three'],
                fieldTemplate: (data) => {
                    return $('<div>').dxTextBox({ value: data });
                },
                valueChangeEvent: 'keyup',
                ...options
            });
            let $input = $dropDownEditor.find(`.${TEXT_EDITOR_INPUT_CLASS}`);
            const $label = $dropDownEditor.find(`.${TEXT_EDITOR_LABEL_CLASS}`);

            const expectedAriaLabelledByValue = !options.inputAttr && options.label ?  $label.attr('id') : undefined;

            assert.strictEqual($input.attr('aria-labelledby'), expectedAriaLabelledByValue, `aria-labelledby attribute value after initialization`);

            keyboardMock($input)
                .type('a');

            $input = $dropDownEditor.find(`.${TEXT_EDITOR_INPUT_CLASS}`);

            assert.strictEqual($input.attr('aria-labelledby'), expectedAriaLabelledByValue, `aria-labelledby attribute value after typing`);

            keyboardMock($input)
                .caret(1)
                .press('backspace');

            $input = $dropDownEditor.find(`.${TEXT_EDITOR_INPUT_CLASS}`);

            assert.strictEqual($input.attr('aria-labelledby'), expectedAriaLabelledByValue, `aria-labelledby attribute value after deleting`);
        });
    })
```

jdvictoria:
[SelectBox - The input element has incorrect role attribute value if async templates are used [T1230635]](https://trello.com/c/Zobx2cfX/8812-selectbox-the-input-element-has-incorrect-role-attribute-value-if-async-templates-are-used-t1230635)

EugeniyKiyashko (dismissed):

Zedwag (dismissed):

Zedwag (approved):

EugeniyKiyashko (approved):

PR #27478, opened 2024-05-29
chore: add comparing DevExtreme version with dependent product version

VasilyStrelyaev:


packages/devextreme/js/__internal/core/license/types.ts
```diff
@@ -28,3 +28,8 @@ export interface LicenseCheckParams {
   internal?: true;
   error: LicenseVerifyResult | undefined;
 }
+
+export interface DependentVersion {
```
IlyaKhD: It shouldn't be connected with license check

packages/devextreme/js/__internal/core/license/license_validation.test.ts
```diff
@@ -134,6 +135,142 @@ describe('license token', () => {
...
+    clearDependentVersions();
+  });
+
+  test('Perform license check if versions match', () => {
+    const token = 'ewogICJpbnRlcm5hbFVzYWdlSWQiOiAiUDdmNU5icU9WMDZYRFVpa3Q1bkRyQSIsCiAgImZvcm1hdCI6IDEKfQ==.ox52WAqudazQ0ZKdnJqvh/RmNNNX+IB9cmun97irvSeZK2JMf9sbBXC1YCrSZNIPBjQapyIV8Ctv9z2wzb3BkWy+R9CEh+ev7purq7Lk0ugpwDye6GaCzqlDg+58EHwPCNaasIuBiQC3ztvOItrGwWSu0aEFooiajk9uAWwzWeM=';
+    reportDependentVersion('DevExpress.Product.A', CORRECT_VERSION);
+    reportDependentVersion('DevExpress.Product.A', CORRECT_VERSION);
+    reportDependentVersion('DevExpress.Product.B', CORRECT_VERSION);
+    validateLicense(token, CORRECT_VERSION);
+    expect(errors.log).toHaveBeenCalledWith('W0020');
```
IlyaKhD: There should be a test for valid license as well - if license is not expired, the mismatch check should be performed as well. This check should be considered as not connnected with license check - this way it will be more clear for customets and consumers

packages/devextreme/js/__internal/utils/version.ts
```diff
@@ -25,6 +29,45 @@ export function parseVersion(version: string): Version {
   };
 }
 
+export function reportDependentVersion(dependentName: string, version: string): void {
```
IlyaKhD: It seems to be unclear from consumer package POV. Compare these statements:

(1) "I as a DevExtreme consumer I should report my version to it"
(2) "I as a DevExtreme consumer I should check it version"

The (1) leads to "why?", but the (2) is expected - dependency management always require to specify the version of a dependency.

IlyaKhD: As for me, I mostly prefer the verb "assert", as it's very common in checking prerequisites.

Maybe `assertDevExtremeVersion` instead of shorter `assertVersion`

packages/devextreme/js/__internal/utils/version.ts
```diff
@@ -25,6 +29,45 @@ export function parseVersion(version: string): Version {
...
+}
+
+export function stringifyVersionList(dependents: DependentVersion[]): string {
+  return dependents
+    .map((dependent) => `${dependent.dependentName}: ${dependent.version}`)
+    .join('\n');
+}
+
+export function versionsEqual(versionAStr: string, versionBStr: string): boolean {
+  const versionAComponents = versionAStr.split('.').map(Number);
```
IlyaKhD: Use `parseVersion` funcion from the 'version' module

Such things shouldn't be duplicated, if one day we need to change the parsing logic - we'll have to search for all these places and likely miss some.

packages/devextreme/js/__internal/core/license/license_validation.ts
```diff
@@ -162,8 +165,26 @@ export function validateLicense(licenseKey: string, versionStr: string = package
...
+  if (mismatchingDependents.length) {
+    errors.log('W0023', stringifyVersionList([
+      {
+        dependentName: 'devextreme',
+        version: versionStr,
+      },
+      ...mismatchingDependents,
+    ]));
+    return;
+  }
```
IlyaKhD: This should be a part of the version module. Version check and license check are routines of the same level and should be independent from each other

As for now, there are no other consumers for the version check output, so we can call it from within the license check, but license check should now nothing of it's implementation

packages/devextreme/js/__internal/utils/version.ts
VasilyStrelyaev: > (2) "I as a DevExtreme consumer I should check it version"

If I check something, I get some information and act by it. However, here the DevExtreme consumer does not act, it just reports its version to DevExtreme and DevExtreme decides whether to fire a warning and does it (or not)

packages/devextreme/js/core/version_check.js
```diff
@@ -0,0 +1,3 @@
+import { assertDevExtremeVersion as assertVersion } from '../__internal/utils/version';
```
IlyaKhD: **Minor**
You could use this:
```suggestion
export { assertDevExtremeVersion } from '../__internal/utils/version';
```

IlyaKhD (dismissed):

packages/devextreme/js/core/version_check.js
VasilyStrelyaev: Looks like it does

image

IlyaKhD (dismissed):

IlyaKhD (approved):

alexslavr (approved):

PR #27493, opened 2024-05-30
DataGrid: Fix row going into edit mode when popup edit mode is enabled and editRowKey is initially specified (T1218553)

Alyar666:
See https://devexpress.com/issue=T1218553


williamvinogradov:
Let's add an e2e testcafe screenshot test for this case :)

williamvinogradov (approved):

PR #27494, opened 2024-05-31
Splitter: expanded pane size should not exceed maxSize

Zedwag:


packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -804,10 +804,18 @@ class Splitter extends CollectionWidget<Properties> {
...
 
-    const itemSize = targetPaneSize >= minSize
+    let itemSize = targetPaneSize >= minSize
       ? targetPaneSize
       : minSize;
 
+    if (isDefined(item.maxSize)) {
+      const { orientation } = this.option();
+      const handlesSizeSum = this._getResizeHandlesSize();
+      const elementSize = getElementSize($(this.element()), orientation);
```
EugeniyKiyashko: I believe we can avoid getting the element sizes twice during collapsing. I will try to refactor this part in the next commit

EugeniyKiyashko (approved):

AlexanderMoiseev (approved):

PR #27495, opened 2024-05-31
Regenerate devextreme-angular/common for Angular 18 and moduleResolution=bundler (T1234715)

GoodDayForSurf:
https://devexpress.com/issue=T1234715


packages/devextreme-angular/src/common/charts/index.ts
```diff
@@ -0,0 +1,46 @@
+export {
```
mpreyskurantov: Why is it duplicated instead of moving from https://github.com/GoodDayForSurf/DevExtreme/blob/24_1_angular_regenerate_for_ng18/packages/devextreme-angular/src/common/charts.ts
?

mpreyskurantov (changes requested):
Resolve duplicate modules

mpreyskurantov (approved):

PR #27509, opened 2024-06-04
Toolbar Adaptability: Fix demos and unskip Testcafe tests (Angular, React and Vue)

jdvictoria:


jdvictoria:
[Toolbar.Adaptability demo: remove skipped TestCafe tests in Angular, React, and Vue](https://trello.com/c/RPzwD3u6/8931-toolbaradaptability-demo-remove-skipped-testcafe-tests-in-angular-react-and-vue)

apps/demos/testing/widgets/toolbar/etalons/toolbar_singleline_mode_menu_open.png
nikkithelegendarypokemonster, line 1: Good day this is my only question is it normal that the border in font size and line spacing to be removed? if it is then disregard this comment?

jdvictoria: I believe that it's also a requirement based on the card.
image

nikkithelegendarypokemonster: ok..

nikkithelegendarypokemonster (dismissed):

apps/demos/testing/widgets/toolbar/etalons/toolbar_singleline_mode_menu_open.png
ksercs: @jdvictoria 
Nikki is right, it's expected only for Material theme, not Generic.
I think you can this to get a theme and set the property dependent on it:
```javascript
import themes from 'devextreme/ui/themes';
// themes.current()
```

apps/demos/Demos/Toolbar/Adaptability/Vue/App.vue
```diff
@@ -451,4 +464,8 @@ function onFontFamilyClick() {
   height: 1px;
   border-bottom: 1px solid #ddd;
 }
+
+.dx-button-content {
```
EugeniyKiyashko, line 468: Why should we add this style in Vue only? As I can see, there is an issue with aligning the button's icon and text. This issue occurs for buttons that have locateInMenu: always.

image

The markup in Vue has particularities, and the list-item-content class is present on the Button element. To me, it looks like a bug.

Well, we can leave it as a temporary solution for the Vue demo, but let's create a Trello card (into to Discuss column) with a full description of the issue to consider this scenario in the near future.

EugeniyKiyashko (dismissed):

apps/demos/Demos/Toolbar/Adaptability/Angular/app/app.component.ts
```diff
@@ -36,6 +37,8 @@ if (!/localhost/.test(document.location.host)) {
 })
 
 export class AppComponent {
+  stylingMode = isMaterialBased() ? 'text' : undefined;
```
ksercs: It's not a public API and should not be used in demos. Please search through the written demos, find the way how this check should be implemented and use the same approach.

ksercs (changes requested):

apps/demos/Demos/Toolbar/Adaptability/Angular/app/app.component.ts
jdvictoria: [6d02fe8](https://github.com/DevExpress/DevExtreme/pull/27509/commits/6d02fe8dd0db7abe9c3a3bf2a4517a2ef078836f)

ksercs (approved):

EugeniyKiyashko (approved):

PR #27511, opened 2024-06-04
TreeView: Fix demos and unskip Testcafe tests (React and Vue) (Draft)

jdvictoria:


nikkithelegendarypokemonster:
Good day...

I started to review the changes all looks good i have just some minor suggestions in ItemSelectionAndCustomization styling
all works perfectly fine but i also researched it i think we can also do this approach i think this is much cleaner and simpler
By adding this styling only
`
#treeview{
  float: left;
  clear:none;
}
`

the main reason for the magic whitespace is how jquery react and angular do their rendering in inline-block elements

also i noticed in vue the data in DragAndDropHierarchicalDataStructure file

the last 2 items has incorrect icons not like in other approaches i suggest changing the icon of the last 2 items to file instead of activefolder

jdvictoria:
[TreeView demos: remove skipped TestCafe tests in React, and Vue](https://trello.com/c/eGfh4vxL/8932-treeview-demos-remove-skipped-testcafe-tests-in-react-and-vue)

apps/demos/Demos/TreeView/DragAndDropHierarchicalDataStructure/React/styles.css
```diff
@@ -33,3 +33,7 @@
 .drive-panel:last-of-type {
   border-right: none;
 }
+
+.dx-sortable {
```
EugeniyKiyashko: This is a workaround to achieve consistent display, but it doesn't address the root cause of the problem. We need to choose an approach here. For example, we can change the element to which we add the margin. In this case, the headers can have a bottom margin instead of the content having a top margin.
It looks like this should solve the problem.

apps/demos/Demos/TreeView/ItemSelectionAndCustomization/jQuery/styles.css
```diff
@@ -12,7 +16,7 @@
   padding: 20px;
   margin-left: 20px;
   background-color: rgba(191, 191, 191, 0.15);
-  font-size: 115%;
+  font-size: initial;
```
EugeniyKiyashko: You shouldn't change font size or imagine something else. In Vue we don't have whitespace between treeview and options elements. Why in jQuery, Angular, React we have, but not in Vue?
Pay attention to the markup in React - {{ ' ' }}.
Pay attention to the following setting in Angular demo - preserveWhitespaces.

Can we use the same additional element in the markup as used in React? 
This looks like the simplest solution.

apps/demos/Demos/TreeView/DragAndDropHierarchicalDataStructure/Angular/app/app.component.css
```diff
@@ -20,6 +15,7 @@
   min-height: auto;
   padding: 0;
   cursor: default;
+  margin-bottom: 10px;
```
EugeniyKiyashko, line 18: To be attentive. You added the style for one header only.

PR #27514, opened 2024-06-04
NX: scripts configs for cloud caches

iBat:


apps/angular/project.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "devextreme-angular-playground",
```
pomahtri, line 2: Let's delete `project.json` file for all playgrounds
I think default inputs and outputs will look more simple and more stable

apps/demos/project.json
```diff
@@ -0,0 +1,121 @@
...
+  "projectType": "application",
+  "targets": {
+    "test": {
+      "executor": "nx:run-script",
+      "options": {
+        "script": "test"
+      },
+      "dependsOn": [
+        "!^build"
+      ],
```
pomahtri: Does nx support excluding targets from `dependsOn`? Shouldn't it be like so?
```suggestion
      "dependsOn": [],
```

e2e/testcafe-devextreme/project.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "devextreme-testcafe-tests",
```
pomahtri, line 2: Same as for playgrouds, let's delete this file. Looks like defaults target settings are OK

apps/angular/project.json
iBat: Inputs - may be. But outputs - no.

e2e/testcafe-devextreme/project.json
pomahtri: resolved in DM, decided to keep settings explicitly

pomahtri (approved):

PR #27521, opened 2024-06-05
TagBox: Fix tag label display when valueExpr is set to function and hideSelectedItems is enabled (T1234032) (Draft)

jdvictoria:


jdvictoria:
[TagBox values are incorrectly displayed with keyExpr set to a function and hideSelectedItems enabled [T1234032]](https://trello.com/c/bkwJXsR6/8886-tagbox-values-are-incorrectly-displayed-with-keyexpr-set-to-a-function-and-hideselecteditems-enabled-t1234032)

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -976,7 +984,9 @@ const TagBox = (SelectBox as any).inherit({
 
     if (needFilterPlainItems) {
       const plainItems = this._getPlainItems();
-      selectedItems = this._filterSelectedItems(plainItems, values);
+      const filteredItems = this._filterSelectedItems(plainItems, values);
+
+      selectedItems = this.option('hideSelectedItems') ? this.option('selectedItems').concat(filteredItems) : filteredItems;
```
AlexanderMoiseev: There are a lot of changes, im not quite follow them.
 As far as i understand, when 'hideSelectedItems' option is true, we hide selected items from the list and they appear as tagBox values. At some point, when item in the list is selected, but not yet reflected in tagbox, to get selectedItems we actually have to combine 2 arrays. There is a code inside _getFilteredItems function to manage selected items

selectedItems = listSelectedItems || this.option('selectedItems');

In case when 'hideSelectedItems' is true we can do something like this
selectedItems = (listSelectedItems || []).concat(this.option('selectedItems') || []);

I believe this is the only change that is needed

AlexanderMoiseev (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6822,6 +6822,48 @@ QUnit.module('performance', () => {
         assert.ok($.isFunction(filter), 'filter is function');
     });
 
+    QUnit.test('tag labels should be correctly displayed with valueExpr as function and hideSelectedItems enabled (T1234032)', function(assert) {
```
AlexanderMoiseev: please also add test when hideSelecteItems option changed at runtime

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6822,6 +6822,48 @@ QUnit.module('performance', () => {
...
+        const dataSource = [
+            { id: 1, scheme: 'schema 1', name: 'item1' },
+            { id: 2, scheme: 'schema 2', name: 'item2' },
+            { id: 3, scheme: 'schema 3', name: 'item3' },
+            { id: 4, scheme: 'schema 4', name: 'item4' },
```
AlexanderMoiseev: [minor] i believe 3 items is more than enought, lest not create redundant data for tests

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6822,6 +6822,48 @@ QUnit.module('performance', () => {
...
+
+        tagBox.open();
+        $($list.find('.dx-list-item').eq(0)).trigger('dxclick');
+
+        tagBox.open();
+        $($list.find('.dx-list-item').eq(0)).trigger('dxclick');
+
+        const $tagContainer = $tagBox.find(`.${TAGBOX_TAG_CONTAINER_CLASS}`);
+
+        assert.strictEqual($.trim($tagContainer.text()), 'item1item2item3item4item5', 'label values are displayed correctly');
```
AlexanderMoiseev: just a question, why $.trim is needed here?

jdvictoria: I believe neighboring tests used this approach to check all tag labels in a single assert.

Example:
https://github.com/DevExpress/DevExtreme/blob/9227c7176ed9a46a00358285d3e197af042eaaf9/packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js#L1796

AlexanderMoiseev: ah, ok

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
jdvictoria: [875d1f5](https://github.com/DevExpress/DevExtreme/pull/27521/commits/875d1f5dad71b09a9bfe80286d68cbc533e78732)

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
jdvictoria: [dfd8a82](https://github.com/DevExpress/DevExtreme/pull/27521/commits/dfd8a8227485d7087a12fcc939aa5d36762d5f56)

packages/devextreme/js/__internal/ui/m_tag_box.ts
jdvictoria: [d66f35b](https://github.com/DevExpress/DevExtreme/pull/27521/commits/d66f35bc7c4fa23dd30f6d01d8a41748040bce4e)

AlexanderMoiseev (approved):

AlexanderMoiseev:
JFYI: there is a way to fix it on selection.strategy level i believe, but, as long this bevavior with hidden items is possible only with tagBox, i thing this WA will do. I was looking for solution with selecteion first, but decided not to tamper with base functionality

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6822,6 +6822,45 @@ QUnit.module('performance', () => {
...
+                },
+                displayExpr: 'name',
+                hideSelectedItems: !changeAtRuntime,
+                opened: true,
+            });
+
+            const tagBox = $tagBox.dxTagBox('instance');
+            const $list = tagBox._list.$element();
+
+            if(changeAtRuntime) {
```
ksercs: [excess operation]

This condition is excess: if initially property is `true`, setting it to `true` at runtime will do nothing

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6822,6 +6822,45 @@ QUnit.module('performance', () => {
...
+                tagBox.option('hideSelectedItems', true);
+            }
+
+            $($list.find('.dx-list-item').eq(0)).trigger('dxclick');
```
ksercs: [repeatable code style mistake]

Constants please!

ksercs (changes requested):
[missed scenarios, local solution]

I agree with Alex - this issue is mostly expected to be fixed not on a TagBox level. I see that proposed solution works in the target scenario but I cannot find a reason "why ideologically it should work this way?". 

I believe we cannot accept this solution because it assumes that whole hideSelectedItems prop processing was incorrect, but I see it worked well when valueExpr is not a function => the reason of the issue is related to the valueExpr option processing and mostly it's not on a TagBox level. 

Previously algorithm checked if necessary items are already loaded (are in plain items) and then just render it or additionally loaded dependent on an answer. After your change the code always believes items are already loaded and just renders them - sound not correct. **Example of a broken scenario: just set some value in initial config.**

Let's try to find a real reason this way:
- create two sample pages - with broken behavior when valueExpr is a function and with a correct behavior when valueExpr is a string
- debug both and find the difference - when exactly we get some unexpected result (mostly it's related to list selectedItems or like this)
- find why we get this incorrect result and fix (mostly somewhere in collection strategy)

Please don't forget to write separate tests additionally, at least for the scenario mentioned above.
Also wanna remind you that we will need to write a test for place where the fix is applied (for collection if the fix is on collection level).

PR #27524, opened 2024-06-06
PivotGrid - Row fields overlap data fields if dataFieldArea is set to "row" and virtual scrolling is enabled (T1210807)

Raushen:
See https://supportcenter.devexpress.com/ticket/details/t1210807/pivotgrid-row-fields-overlap-data-fields-if-datafieldarea-is-set-to-row-and-virtual


e2e/testcafe-devextreme/tests/pivotGrid/virtualScrolling_T1210807.ts
```diff
@@ -0,0 +1,194 @@
+import { createScreenshotsComparer } from 'devextreme-screenshot-comparer';
```
williamvinogradov, line 1: This e2e test's PivotGrid configuration contains a lot of redundant data.
And the data set is too big and contains a lot of unnecessary fields.

The original issue [T1210807](https://supportcenter.devexpress.com/ticket/details/t1210807/pivotgrid-row-fields-overlap-data-fields-if-datafieldarea-is-set-to-row-and-virtual) reproduction case can be simplified:
[Link to the simplified example](https://codepen.io/williamvinogradov/pen/eYaRaxN?editors=0010)

So, let's take a simplified example as a basis
To make this e2e test easiest to support/understand in the future :)

packages/testcafe-models/pivotGrid/rowsArea.ts
```diff
@@ -5,8 +5,8 @@ const CLASSES = {
 export default class RowsArea {
   private readonly element: Selector;
 
-  constructor(selector: Selector) {
-    this.element = selector.find(`.${CLASSES.root}`);
+  constructor(selector: Selector, idx?: number) {
+    this.element = selector.find(`.${CLASSES.root}`).nth(idx || 0);
```
williamvinogradov: Minor comment.

For ```undefined```/```null``` values it's better to use the ```??``` instead of the ```||```

williamvinogradov (approved):

PR #27576, opened 2024-06-11
Chart: Fix legend click when Chart is rotated and minBarSize is set (T1236326)

jdvictoria:


jdvictoria:
[Chart - minBarSize causes an error on a legend click [T1236326]](https://trello.com/c/0jg4NcGV/8971-chart-minbarsize-causes-an-error-on-a-legend-click-t1236326)

packages/devextreme/js/viz/axes/tick_generator.js
```diff
@@ -730,7 +730,7 @@ function dateGenerator(options) {
         const correctDateWithUnitBeginning = v => dateUtils.correctDateWithUnitBeginning(v, intervalObject, null, options.firstDayOfWeek);
         const floorAtStartDate = v => new Date(mathFloor((v.getTime() - v.getTimezoneOffset() * MINUTE) / divider) * divider + v.getTimezoneOffset() * MINUTE);
 
-        value = correctDateWithUnitBeginning(value);
+        value = correctDateWithUnitBeginning(new Date(value));
```
ksercs: [not working behavior]
chart_minBarSize

ksercs (changes requested):

packages/devextreme/js/viz/core/series_family.js
```diff
@@ -356,7 +356,7 @@ function updateRangeSeriesValues() {
         const minShownBusinessValue = minBarSize && valueAxisTranslator.getMinBarSize(minBarSize);
         if(minShownBusinessValue) {
             _each(singleSeries.getPoints(), function(_, point) {
-                if(!point.hasValue()) {
+                if(!point.hasValue() || that.rotated) {
```
AlexanderMoiseev, line 359: This looks strange to me, please clarify, what does the 'rotated' option have to do with this bug? and won't it affect scenarios when rotated is true, but we need to set minBarSize?

jdvictoria: I found out that the issue only happens when on rangeBar when rotated is true and minBarSize is set. Issue doesn't happen when rotated is set to false and minBarSize is set.

AlexanderMoiseev: i left a comment in teams channel

PR #27589, opened 2024-06-13
Buy Now Link Tests

tomodasheesh:


packages/devextreme/js/__internal/core/license/license_validation.test.ts
```diff
@@ -364,6 +364,18 @@ describe('license check', () => {
     expect(trialPanelSpy).not.toHaveBeenCalled();
   });
 
+  test.each([
```
VasilyStrelyaev: I have two suggestions here.

First, I think it's enough to check just one pair of values in this particular test, because the behavior we check does not depend on the value.

Second, this test is good to check that the default link is used when nothing is saved in the config. We may want to have another test that sets something in the config and checks that the specified link is used.

Because tests in the previous modules check that the correct links are set in the configs, but there's no place where we check that they are used.

VasilyStrelyaev (dismissed):

VasilyStrelyaev (approved):

PR #27636, opened 2024-06-20
HtmlEditor: retain formats when creating mentions (T1236869)

nikkithelegendarypokemonster:


packages/devextreme/js/__internal/ui/html_editor/modules/m_mentions.ts
```diff
@@ -238,11 +238,21 @@ if (Quill) {
...
       const startIndex = Math.max(0, caretPosition - markerLength);
+
+      const currentContents = this.quill.getContents();
+
+      let retainedAttributes = {};
+      currentContents.ops.forEach((op, index) => {
+        if (index <= startIndex && op.attributes) {
+          retainedAttributes = { ...retainedAttributes, ...op.attributes };
+        }
+      });
```
ksercs: [broken behavior, incorrect solution]

When u call `this.quill.getContents()` you get the whole editor content in `Delta` format including tables images and so on. Then u collect all attributes and u get smth like this 
image
because u got all attributes for all elements in the content.
And then you try to add all this attributes to the mention and inserted string: some of them are declined because it's not applicable, other - applied (text formats).

You can see unexpected behavior on this basic scenario:
HtmlEditor_mention_insert
Here only "bold" format should be applied, but we also see cross out at least.

I think you can just use `this.quill.getFormat(startIndex)` to get necessary formats.
And I also think we should apply this format only for newly inserted `" "` string because seems it's already applied for a mention

P.s test should be rewritten

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
```diff
@@ -207,6 +207,32 @@ QUnit.module('Mention format', () => {
 });
 
 QUnit.module('Mentions module', moduleConfig, () => {
+    test('retain formatting after applying mentions.(T1236869)', function(assert) {
```
ksercs: Don't need a dot :)
```suggestion
    test('retain formatting after mention insert (T1236869)', function(assert) {
```

ksercs (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/mentionModule.tests.js
nikkithelegendarypokemonster: [63d1cb1](https://github.com/DevExpress/DevExtreme/pull/27636/commits/63d1cb1d6bce06f48d449b18c49cd32cf05f8dbb)

packages/devextreme/js/__internal/ui/html_editor/modules/m_mentions.ts
nikkithelegendarypokemonster: [349e3be](https://github.com/DevExpress/DevExtreme/pull/27636/commits/349e3be9ea27f24cdc65f16749c335876438b0bd)

ksercs (approved):

AlexanderMoiseev (approved):

PR #27659, opened 2024-06-24
TagBox: Fix tag label display when valueExpr is set to function and hideSelectedItems is enabled (T1234032)

jdvictoria:


jdvictoria:
[TagBox values are incorrectly displayed with keyExpr set to a function and hideSelectedItems enabled [T1234032]](https://trello.com/c/bkwJXsR6/8886-tagbox-values-are-incorrectly-displayed-with-keyexpr-set-to-a-function-and-hideselecteditems-enabled-t1234032)

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6822,6 +6822,43 @@ QUnit.module('performance', () => {
...
+            tagBox.open();
+            $($list.find(`.${LIST_ITEM_CLASS}`).eq(0)).trigger('dxclick');
+
+            const $tagContainer = $tagBox.find(`.${TAGBOX_TAG_CONTAINER_CLASS}`);
+
+            assert.strictEqual($.trim($tagContainer.text()), 'item1item2item3', 'label values are displayed correctly');
+        });
+    });
+
     QUnit.test('loadOptions.filter should be correct when user filter is also used', function(assert) {
```
ksercs, line 6859: [still missed scenario]
On the [previous review](https://github.com/DevExpress/DevExtreme/pull/27521#pullrequestreview-2114780567) i notices that behavior is broken if we set some initial value (for example `item1 schema1`)
I see it still broken.
Please, cover it.

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/selection.tests.js
```diff
@@ -3025,6 +3025,32 @@ QUnit.module('Selection SelectAllMode', {
...
+                allowSelectAll: true
+            },
+        });
+
+        this.selectionController.selectAll();
+
+        // act
+        const selectedRows = this.selectionController.getSelectedRowsData();
+        // assert
+        assert.deepEqual(selectedRows, []);
```
ksercs: [unexpected test]

So, if we call `selectAll` why then we expect that not row is selected?
Seems like this test only shows that the proposed fix is not correct :)

Anyway, this test package is related to DataGrid, so it's not what we need.
It's better to place such tests in `DevExpress.ui.widgets/selection.test.js`.

For now - let's just remove it

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
```diff
@@ -88,8 +88,9 @@ export default class SelectionStrategy {
   }
 
   _loadFilteredData(remoteFilter, localFilter?: any, select?: any, isSelectAll?: boolean) {
+    const keyField = this.options.key();
     const filterLength = encodeURI(JSON.stringify(this._removeTemplateProperty(remoteFilter))).length;
-    const needLoadAllData = this.options.maxFilterLengthInRequest && (filterLength > this.options.maxFilterLengthInRequest);
+    const needLoadAllData = (this.options.maxFilterLengthInRequest && (filterLength > this.options.maxFilterLengthInRequest)) || isFunction(keyField);
```
ksercs: [incorrect logic]
In the previous solution you presumed that items should never be loaded if hideSelectedItems=true - you decreased the count of scenarios when we call loading, and we found broken scenarios.

Now you, vice versa, try to always **load all data** if keyExpr is a function, not filtered items. It's also incorrect because it will lead to many reduntant loadings. Just imagine that your List/DataGrid has a 1000 items and users tries to search for a specific one. Now your code will load all 1000 items.

---

I debugged a bit - I believe that the reason of the issue is here [in SelectionFilter](https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/js/core/utils/selection_filter.js#L28). If keyExpr is not a string, we consider it's a composite key (array like `['name1', 'name2']`), but it's not, it's just a regular function.

So, please
- revert the previous fix and try to update SelectionFilter code
- add a tagBox test to check that setting of an initial value works
- debug and ensure this is the reason of the issue, share a detailed explanation under the post

Regarding test for SelectionFilter - let's omit it for this time. SelectionFilter is not in our responsibility area - I discussed this case with owner and he approved we don't write an explicit test.

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/selection.tests.js
jdvictoria: [8e671fb](https://github.com/DevExpress/DevExtreme/pull/27659/commits/8e671fb170a96a989515291e4efc42a44b300c07)

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
jdvictoria: revert previous fix - [8af98af](https://github.com/DevExpress/DevExtreme/pull/27659/commits/8af98af7e18d8f135f0d0b670a94574346dbf72c)

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
jdvictoria: [4f3efe0](https://github.com/DevExpress/DevExtreme/pull/27659/commits/4f3efe0a5a2970777e7b61985586c1c8594d7af6)

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
jdvictoria: new fix

[accd01e](https://github.com/DevExpress/DevExtreme/pull/27659/commits/accd01eb772309abdd91fbc0fa37b89864aadb64)

fixed the issue but failed for this test

image

extended the condition even further to avoid failing tests

[030dd92](https://github.com/DevExpress/DevExtreme/pull/27659/commits/030dd927fbb4e35a4b613f3150b130a803969b6c)

packages/devextreme/js/core/utils/selection_filter.js
```diff
@@ -25,7 +25,7 @@ export const SelectionFilterCreator = function(selectedItemKeys, isSelectAll) {
                 filterExpr.push(isSelectAll ? 'and' : 'or');
             }
 
-            if(isString(keyExpr)) {
+            if(isString(keyExpr) || (isFunction(keyExpr) && keyExpr.length > 0)) {
```
ksercs, line 28: [excess logic]

What is this for? How length can be <= 0 for function and why do we need checking it?

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
ksercs: @jdvictoria  it's a third time when I ask you to add a test for a case when initial value is set. Please, write a separate atomic test which creates a component with function keyExpr and initial value set, and checks tags are rendered correctly

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6822,6 +6822,40 @@ QUnit.module('performance', () => {
...
+    [false, true].forEach(changeAtRuntime => {
+        QUnit.test(`tag labels should be correctly displayed when valueExpr is function and hideSelectedItems enabled ${changeAtRuntime ? 'at runtime' : ''} (T1234032)`, function(assert) {
+            const dataSource = [
+                { id: 1, scheme: 'schema 1', name: 'item1' },
+                { id: 2, scheme: 'schema 2', name: 'item2' },
+            ];
+
+            const $tagBox = $('#tagBox').dxTagBox({
+                dataSource,
+                value: [{ id: 0, scheme: 'schema 0', name: 'item0' }],
```
ksercs, line 6834: [useless prop]

This value is not correct. You have `valueExpr` property set => value should be smth like `item0 schema 0`.
But anyway you don't use this setting in your test, so let's get rid of it

packages/devextreme/js/core/utils/selection_filter.js
jdvictoria: Without checking keyExpr.length, the test `loadOptions.filter should be a function when valueExpr is function` fails.
Failing job: https://github.com/DevExpress/DevExtreme/actions/runs/9696582548/job/26759088914 

The test has this configuration

```
const $tagBox = $('#tagBox').dxTagBox({
            dataSource: {
                load
            },
            valueExpr() {
                return 'id';
            },
            displayExpr: 'text',
            opened: true,
            hideSelectedItems: true
        });
```

wherein the keyExpr has a value of 0

PR #27662, opened 2024-06-26
Support case sensitivity option in the dataQuery

tomodasheesh:


packages/devextreme/testing/tests/DevExpress.data/queryArray.tests.js
```diff
@@ -350,6 +350,20 @@ QUnit.test('filter with functional getter', function(assert) {
     });
 });
 
+QUnit.test('filter with collatorOptions.sensitivity set to "case"', function(assert) {
+    assert.expect(1);
```
williamvinogradov: This assert looks redundant - let's remove it :)

packages/devextreme/js/data/query.d.ts
```diff
@@ -166,13 +167,26 @@ export interface Query {
...
+ * @docid
+ * @type object
+ * @public
+ */
+export interface QueryOptions {
+    /**
+     * @docid
+     * @public
+     */
+    langParams?: LangParams;
```
williamvinogradov: Question.

We adding only the ```case-sensitive/case-insensitive``` comparison feature to the ```DataQuery```
But here we declare a wide ```LangParams``` type.

The ```LangParams.locale``` looks ok because this options is taken into account in comparison [here](https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/js/core/utils/data.js#L200)
But the ```LangParams.collatorOptions``` has a lot of parameters that the ```DataQuery``` doesn't take into account as far as I can see (I may be wrong)

E.g. the ```sensitivity``` has four different variants ([link to the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#sensitivity))
The ```DataQuery``` takes into account only the ```case-sensitive/case-insensitive```

So, maybe it's a good idea to make this type more strict and custom?
Something like this:
```typescript
langOptions?: {
  locale: string; // like in LangParams
  caseSensitive: boolean; 
}
```

@tomodasheesh, @GoodDayForSurf, @VasilyStrelyaev, what do you think about it? :)

packages/devextreme/testing/tests/DevExpress.data/queryArray.tests.js
```diff
@@ -350,6 +350,20 @@ QUnit.test('filter with functional getter', function(assert) {
     });
 });
 
+QUnit.test('filter with collatorOptions.sensitivity set to "case"', function(assert) {
```
williamvinogradov: Let's add the test case when the ```langParams``` isn't set

But let's clarify the questions about ```langParams``` type (the comment in the ```d.ts``` above) first
Because we would add more test cases to check all possible scenarios :)

packages/devextreme/js/data/query.d.ts
GoodDayForSurf: why we add type for query? query can be ArrayQuery or remoteQuery. QueryOptions is not used in remoteQuery..

tomodasheesh: We declare a wide `LangParams` type since this utility function requires `langParams` as the third argument
https://github.com/DevExpress/DevExtreme/blob/b9a7f67445d640c32780acfbff39453c6f4a8c1f/packages/devextreme/js/data/array_query.js#L245


I just saw that we also use `queryOptions.errorHandler`, so we should take that into account as well.
https://github.com/DevExpress/DevExtreme/blob/b9a7f67445d640c32780acfbff39453c6f4a8c1f/packages/devextreme/js/data/array_query.js#L552


Regarding `remote_query` and `array_query`, maybe I'll change the type name to be specific to `array_query` only?

image

packages/devextreme/testing/tests/DevExpress.data/queryArray.tests.js
```diff
@@ -350,6 +350,24 @@ QUnit.test('filter with functional getter', function(assert) {
     });
 });
 
+QUnit.test('filter with undefined "langParams"', function(assert) {
```
williamvinogradov, line 353: Let's add one more QUnit test with the case ```sensitivity: 'base'```
It's supported by the ```DataQuery``` [here](https://github.com/DevExpress/DevExtreme/pull/27662/files#diff-31a735c2b79c2deb183db2400f94c57fbe4c05d459afc8cafb6121f22bf5dedbR195)
[Link to the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator#base)

This test may help us in future code modifications.
With it, we don't forget that we support the ```sensitivity: 'base'``` :)

packages/devextreme/testing/tests/DevExpress.data/queryArray.tests.js
```diff
@@ -350,6 +350,36 @@ QUnit.test('filter with functional getter', function(assert) {
...
+QUnit.test('filter with collatorOptions.sensitivity set to "case"', function(assert) {
+    const input = [{ ID: 'AAA', Name: 'Name 2' }, { ID: 'aaa', Name: 'Name 3' }];
+    const filterLength = QUERY(input, {
+        langParams: {
+            collatorOptions: {
+                sensitivity: 'case'
+            }
+        }
+    }).filter(['ID', '=', 'aaa']).toArray().length;
+    assert.equal(filterLength, 1);
```
GoodDayForSurf: I think it's better also to add  assert.equal(array.ID, 'aaa')

pomahtri (approved):

vladaskorohodova (approved):

VasilyStrelyaev (approved):

williamvinogradov (approved):

PR #27678, opened 2024-06-27
List: Fix "Select All" aria-label localization (T1239880)

jdvictoria:


jdvictoria:
[List - "Select All" aria-label is not localized [T1239880]](https://trello.com/c/1y53C5nH/9035-list-select-all-aria-label-is-not-localized-t1239880)

PR #27706, opened 2024-07-03
Datagrid - Pressing the Enter key in a Detail row triggers the click event of unexpected element (T1240353)

tongsonbarbs:


e2e/testcafe-devextreme/tests/dataGrid/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -4629,3 +4629,62 @@ test('TreeList/DataGrid - Focus indicator is not visible when the Toolbar includ
...
+    .pressKey('enter');
+
+  const message = await Selector('#otherContainer').innerText;
+  const focusButtonMessage = message.includes(
+    'second button',
+  );
+
+  await t
+    .expect(focusButtonMessage)
+    .ok();
```
williamvinogradov: Let's refactor this code block a little, I think there are two options:

1) Let's rename the ```focusButtonMessage```, because this name doesn't reflect the essence
(It's a boolean flag, not the focus button message)
E.g:
```suggestion
  const message = await Selector('#otherContainer').innerText;
  const isTextFromSecondBtn = message.includes('second button');

  await t
    .expect(isTextFromSecondBtn)
    .ok();
```

Or

2) Let's use the expect comparison :)
E.g:
```suggestion
  const message = await Selector('#otherContainer').innerText;

  await t
    .expect(message)
    .eql('second button', 'The message should be from the second button');
```

e2e/testcafe-devextreme/tests/dataGrid/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -4629,3 +4629,62 @@ test('TreeList/DataGrid - Focus indicator is not visible when the Toolbar includ
...
+  ],
+  keyExpr: 'ID',
+  columns: [
+    {
+      dataField: 'Prefix',
+      caption: 'Title',
+      width: 70,
+    },
+    'FirstName',
+  ],
```
williamvinogradov: Let's simplify the ```DataGrid``` configuration a little:

```suggestion
  columns: [ 'Prefix', 'FirstName' ],
```

e2e/testcafe-devextreme/tests/dataGrid/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -4629,3 +4629,62 @@ test('TreeList/DataGrid - Focus indicator is not visible when the Toolbar includ
   keyExpr: 'field_0',
   showBorders: true,
 }));
+
+test('Enter key should not trigger other function besides the function assigned on the clicked button', async (t) => {
```
williamvinogradov, line 4634: Let's add info about the SC ticket number 
There are two options how we can do it:

1) Add the SC ticket number to the test case name
```suggestion
test('Enter key should not trigger other function besides the function assigned on the clicked button (T1240353)', async (t) => {
```
Or 
2) Add a comment with the ticket number on top:
```suggestion
// T1240353
test('Enter key should not trigger other function besides the function assigned on the clicked button', async (t) => {
```

williamvinogradov (approved):

PR #27713, opened 2024-07-03
TreeView: should have focus style for CheckBox in material and fluent (T1235403)

Zedwag:


ksercs (approved):

packages/devextreme/scss/widgets/fluent/treeView/_index.scss
```diff
@@ -13,6 +13,28 @@
   @include dx-icon-sizing($fluent-base-icon-size);
 }
 
+.dx-treeview-item-with-checkbox {
+  &.dx-state-focused > .dx-treeview-item {
+    .dx-checkbox {
+      .dx-checkbox-icon {
+        border-color: $checkbox-accent-color-focused;
```
ksercs, line 20: That's a sad story that we cannot incapsulate this logic in checkBox. I presume it's done this way because we don't wanna make checkboxes truly focusable, right?

I see a similar solution in list, so lgtm for this time

pomahtri (approved):

packages/devextreme/scss/widgets/fluent/treeView/_index.scss
Zedwag: yes, it's because whe whole item get focus, checkBox know nothing about it

Shuvaevmedia (approved):

PR #27734, opened 2024-07-05
Scheduler - The "element.get is not a function" error occurs in onAppointmentRendered (T1239368)

tongsonbarbs:


packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -2438,7 +2438,7 @@ class Scheduler extends Widget<any> {
 
         return result;
       }),
-      targetElement: getPublicElement(target),
+      targetElement: target instanceof HTMLElement ? target : getPublicElement(target),
```
pomahtri: minor:
```suggestion
      targetElement: getPublicElement($(target)),
```
it's common pattern in our code base, you can look up for `getPublicElement($(` in IDE

Also please consider moving this logic inside of `getPublicElement` to prevent similar mistakes in future

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -2419,7 +2419,7 @@ class Scheduler extends Widget<any> {
     };
   }
 
-  showAppointmentTooltipCore(target: dxElementWrapper, data, options?: any) {
+  showAppointmentTooltipCore(target: dxElementWrapper | HTMLElement, data, options?: any) {
```
pomahtri: Check please whether target can be of other type. 
Public doc for `showAppointmentTooltip` states that target can be string (idk how this is used, but we have it ¯\\\_(ツ)\_/¯ )

pomahtri:
Can you please also revert changes that I made in tests while creating this regression?
I changed some invocations of this func with `string` target. I thought it was mistaken, but occurs that it was not 
https://github.com/DevExpress/DevExtreme/pull/27250/files

pomahtri (approved):

PR #27737, opened 2024-07-07
TreeView: should have focus style for CheckBox in material and fluent (T1235403)

Zedwag:


Zedwag:
we decided to not cherry-pick this to 23_2

PR #27751, opened 2024-07-11
DataGrid - Hovering does not highlight rows properly (T1240074)

tongsonbarbs:


packages/devextreme/js/__internal/grids/grid_core/m_widget_base.ts
```diff
@@ -26,6 +24,12 @@ export default class GridCoreWidget<TProperties> extends Widget<TProperties> {
     return result;
   }
 
+  protected _init() {
+    this._activeStateUnit = '.dx-row';
```
williamvinogradov: It's better to move the "magic string" ```.dx-row``` back to the ```const``` variable :)
[Good article about "magic strings/numbers"](https://dev.to/ruben_alapont/magic-numbers-and-magic-strings-its-time-to-talk-about-it-ci2)

E.g:
```typescript
const GRID_CORE_ROW_SELECTOR = '.dx-row';

export default class GridCoreWidget<TProperties> extends Widget<TProperties> {
  // other code
  protected _init() {
    this._activeStateUnit = GRID_CORE_ROW_SELECTOR;
    // @ts-expect-error
    super._init();
  }
 // other code
}

```

e2e/testcafe-devextreme/tests/dataGrid/markup/T1240074_hoveringRows.ts
```diff
@@ -0,0 +1,27 @@
...
+  url(__dirname, '../../container.html'),
+);
+
+test('Hover over a row in DataGrid', async (t) => {
+  const dataGrid = new DataGrid('#container');
+  const firstRow = dataGrid.getDataRow(0);
+
+  await t
+    .hover(firstRow.element)
+    .expect(firstRow.element.hasClass('dx-state-hover'))
```
williamvinogradov: The ```DataGrid``` is POM (Page Object Model)
Here is [a short article about POM](https://medium.com/tech-tajawal/page-object-model-pom-design-pattern-f9588630800b)
Instead of writing direct selectors in our tests, we try to encapsulate them in our widget-related POMs :)
Our widget-related POMs are stored [in the testcafe-models nx project](https://github.com/DevExpress/DevExtreme/tree/24_2/packages/testcafe-models)
And sometimes we call this entities just the ```Test Models``` or ```Test Object Models```

The [dataGrid.getDataRow](https://github.com/DevExpress/DevExtreme/blob/24_2/packages/testcafe-models/dataGrid/index.ts#L130) method returns the [DataRow object](https://github.com/DevExpress/DevExtreme/blob/24_2/packages/testcafe-models/dataGrid/data/row.ts#L20)
The DataRow already contains functionality to check the row hover [here](https://github.com/DevExpress/DevExtreme/blob/24_2/packages/testcafe-models/dataGrid/data/row.ts#L46)

So, let's replace the direct selector usage with the ```DataRow.isHovered``` usage :)
E.g:
```typescript
  .expect(firstRow.isHovered)
  .ok();
```

e2e/testcafe-devextreme/tests/dataGrid/markup/T1240074_hoveringRows.ts
```diff
@@ -0,0 +1,27 @@
...
+    .ok();
+}).before(async () => {
+  await createWidget(
+    'dxDataGrid',
+    {
+      dataSource: getData(5, 3),
+      hoverStateEnabled: true,
+    },
+    '#container',
```
williamvinogradov: The third param of the ```createWidget``` functions is ```#container``` by default
[Link to the createWidget code](https://github.com/DevExpress/DevExtreme/blob/24_2/e2e/testcafe-devextreme/helpers/createWidget.ts#L39)

So, let's remove the third param :)
```suggestion
```

e2e/testcafe-devextreme/tests/dataGrid/markup/T1240074_hoveringRows.ts
```diff
@@ -0,0 +1,27 @@
+import DataGrid from 'devextreme-testcafe-models/dataGrid';
+import { createWidget } from '../../../helpers/createWidget';
+import url from '../../../helpers/getPageUrl';
+import { getData } from '../helpers/generateDataSourceData';
+
+fixture.disablePageReloads`HoveringRows`.page(
+  url(__dirname, '../../container.html'),
+);
+
```
williamvinogradov, line 9: Let's extend the existing test case to check more cases
Now we're checking only that the first row has the correct class on hover
But in the original issue, the last row had the correct class on hover (not the first, but I mean that checking only the first row may be an edge case)

So, let's do something like this:
```typescript
test('Hover over rows in the middle', async(t) => {
  const firstRow = dataGrid.getDataRow(10);
  const secondRow = dataGrid.getDataRow(11);
  
  await t.hover(firstRow.element)
             .expect(firstRow.isHovered)
             .ok();

  await t.hover(secondRow.element)
             .expect(firstRow.isHovered)
             .notOk()
             .expect(secondRow.isHovered)
             .ok();
}).before(async () => {
  await createWidget(
    'dxDataGrid',
    {
      dataSource: getData(20, 3),
      hoverStateEnabled: true,
    }
  )
});

test('Hover over first and last rows', async (t) => {
  const firstRow = dataGrid.getDataRow(10);
  const lastRow = dataGrid.getDataRow(20);
  
  await t.hover(firstRow.element)
             .expect(firstRow.isHovered)
             .ok();

  await t.hover(lastRow.element)
             .expect(firstRow.isHovered)
             .notOk()
             .expect(lastRow.isHovered)
             .ok();
}).before(async () => {
  await createWidget(
    'dxDataGrid',
    {
      dataSource: getData(20, 3),
      hoverStateEnabled: true,
    }
  )
});
```

packages/devextreme/js/__internal/grids/grid_core/m_widget_base.ts
williamvinogradov: Already resolved :)

williamvinogradov (approved):

PR #27771, opened 2024-07-15
Chat: Add doc declarations

marker-dao:


packages/devextreme/js/ui/chat.d.ts
```diff
@@ -0,0 +1,182 @@
...
+    ChangedOptionInfo,
+} from '../events/index';
+
+/**
+ * @docid _ui_chat_ContentReadyEvent
+ * @public
+ * @type object
+ * @inherits EventInfo
+ */
+export type ContentReadyEvent = EventInfo<dxChat>;
```
ksercs: I think we don't need contentReady in Chat

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * @docid
+     * @public
+     */
+    lastName: string;
+    /**
+     * @docid
+     * @public
+     */
+    avatarUrl?: string;
```
ksercs, line 65: Will add avatarAltText?

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * @docid
+     * @public
+     */
+    firstName: string;
+    /**
+     * @docid
+     * @public
+     */
+    lastName: string;
```
ksercs: Don't you think single `name` is enough here? Especially if it's mandatory

packages/devextreme/js/ui/chat.d.ts
marker-dao: For what? We can use user name for alt text. We don't need this option. It looks strange for me)

packages/devextreme/js/ui/chat.d.ts
artem-kurchenko: if we remove onContentReady where vanilla JS users will apply options at runtime after a component is rendered\refreshed in this component?

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * @docid
+     * @public
+     */
+    text: string;
+    /**
+     * @docid
+     * @public
+     */
+    typing?: boolean;
```
vladaskorohodova, line 97: Maybe "isTyping" would be a better name? I just didn't get what "typing" means from the first glance

packages/devextreme/js/ui/chat.d.ts
marker-dao: For now we decided to leave 2 fields

packages/devextreme/js/ui/chat.d.ts
marker-dao: "typing" is a status that tells you whether the user is currently typing or not

packages/devextreme/js/ui/chat.d.ts
marker-dao: At this time, we have decided not to publish this in the documentation

vladaskorohodova (dismissed):

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * @docid
+     * @public
+     */
+    firstName: string;
+    /**
+     * @docid
+     * @public
+     */
+    lastName: string;
```
EugeniyKiyashko: Is this field required?

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -0,0 +1,168 @@
...
+    /**
+     * @docid
+     * @public
+     */
+    id: number;
+    /**
+     * @docid
+     * @public
+     */
+    firstName: string;
```
EugeniyKiyashko: JFI: I agreed with Anton that it's better to have only one field for userName. It can be bot or autogenerated string.

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -0,0 +1,168 @@
...
+ */
+export type User = {
+    /**
+     * @docid
+     * @public
+     */
+    id: number;
+    /**
+     * @docid
+     * @public
```
EugeniyKiyashko, line 57: default value is missing

packages/devextreme/js/ui/chat.js
```diff
@@ -1,3 +1,40 @@
...
+ * @action
+ */
+
+/**
+ * @name dxChatOptions.tabIndex
+ * @hidden
+ */
+
+/**
+ * @name dxChat.focus
```
EugeniyKiyashko: Why we decide to hide focus method? It might be used for focusing of the field.

packages/devextreme/js/ui/chat.d.ts
marker-dao: [9de0c11d9b0eba05dfbb70e36e45048be20df855]

packages/devextreme/js/ui/chat.d.ts
marker-dao: [9de0c11d9b0eba05dfbb70e36e45048be20df855]

packages/devextreme/js/ui/chat.d.ts
marker-dao: [9de0c11d9b0eba05dfbb70e36e45048be20df855]

packages/devextreme/js/ui/chat.js
marker-dao: [4c429a74d08cebbf0cb75a28fa23ed6d511559ba]

EugeniyKiyashko (approved):

vladaskorohodova (approved):

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @namespace DevExpress.ui.dxChat
+ * @public
+ */
+export type Message = {
+    /**
+     * @docid
+     * @default ''
+     * @public
+     */
+    timestamp?: string;
```
EugeniyKiyashko, line 79: For history: It's requires to think about name of this option and its accepted values

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * @docid
+     * @default undefined
+     * @public
+     */
+    author?: User;
```
EugeniyKiyashko, line 85: For historical purposes: perhaps using just the userID instead of passing the full user data will suffice.

mpreyskurantov (approved):

PR #27787, opened 2024-07-17
DataGrid - If a property is removed from the state object, it becomes null rather than taking the default value passed to the component (T1233556)

tongsonbarbs:


apps/demos/Demos/DataGrid/CustomEditors/ReactJs/EmployeeDropDownBoxComponent.js
tongsonbarbs, line 1: Rollback only

PR #27800, opened 2024-07-18
Core, Data - Make array store update function copy undefined + introduce flag for deepExtendArraySafe

ivanblinov2k17:
Cherry-pick #27329

### DIFF
--- a/packages/devextreme/js/core/utils/object.js
+++ b/packages/devextreme/js/core/utils/object.js
@@ -65,9 +65,10 @@ const deepExtendArraySafe = function(target, changes, extendComplexObject, assig
             newValue = deepExtendArraySafe(goDeeper ? prevValue : {}, newValue, extendComplexObject, assignByReference, shouldCopyUndefined);
         }
 
-        if(shouldCopyUndefined && (prevValue !== newValue || prevValue === undefined)) {
-            assignValueToProperty(target, name, newValue, assignByReference);
-        } else if((newValue !== undefined) && prevValue !== newValue) {
+        if(
+            shouldCopyUndefined && (prevValue !== newValue || prevValue === undefined) ||
+            newValue !== undefined && prevValue !== newValue
+        ) {
             assignValueToProperty(target, name, newValue, assignByReference);
         }
     }
--- a/packages/devextreme/js/core/utils/object.js
+++ b/packages/devextreme/js/core/utils/object.js
@@ -48,7 +48,7 @@ const assignValueToProperty = function(target, property, value, assignByReferenc
 };
 
 // B239679, http://bugs.jquery.com/ticket/9477
-const deepExtendArraySafe = function(target, changes, extendComplexObject, assignByReference) {
+const deepExtendArraySafe = function(target, changes, extendComplexObject, assignByReference, shouldCopyUndefined) {
     let prevValue;
     let newValue;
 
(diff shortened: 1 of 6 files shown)


packages/devextreme/js/core/utils/object.js
```diff
@@ -62,10 +62,12 @@ const deepExtendArraySafe = function(target, changes, extendComplexObject, assig
...
             const goDeeper = extendComplexObject ? isObject(prevValue) : isPlainObject(prevValue);
-            newValue = deepExtendArraySafe(goDeeper ? prevValue : {}, newValue, extendComplexObject, assignByReference);
+            newValue = deepExtendArraySafe(goDeeper ? prevValue : {}, newValue, extendComplexObject, assignByReference, shouldCopyUndefined);
         }
 
-        if(newValue !== undefined && prevValue !== newValue) {
+        if(shouldCopyUndefined && (prevValue !== newValue || prevValue === undefined)) {
+            assignValueToProperty(target, name, newValue, assignByReference);
+        } else if((newValue !== undefined) && prevValue !== newValue) {
             assignValueToProperty(target, name, newValue, assignByReference);
```
williamvinogradov: Let's merge these two conditions
The ```assignValueToProperty(target, name, newValue, assignByReference);``` call similar in both logical cases

packages/devextreme/testing/tests/DevExpress.core/utils.object.tests.js
```diff
@@ -171,7 +171,21 @@ QUnit.test('deepExtendArraySafe utility does not throw an error with \'null\' de
...
 });
 
-QUnit.test('deepExtendArraySafe utility does not pollute object prototype', function(assert) {
-    objectUtils.deepExtendArraySafe({ }, JSON.parse('{ "__proto__": { "pollution": true }}'), true);
-    assert.ok(!('pollution' in { }), 'object prototype is not polluted');
+QUnit.test('deepExtendArraySafe sets undefined', function(assert) {
+    const objWithValue = { time: { duration: 50 } };
+    const objNoValue = {};
+    objectUtils.deepExtendArraySafe(objWithValue, { time: { duration: undefined } }, true);
+    objectUtils.deepExtendArraySafe(objNoValue, { time: { duration: undefined } }, true);
```
williamvinogradov: Let's set all argument values in this test
It looks a little confusing now because instead of boolean flags we pass undefined :)

packages/devextreme/testing/tests/DevExpress.core/utils.object.tests.js
```diff
@@ -171,7 +171,21 @@ QUnit.test('deepExtendArraySafe utility does not throw an error with \'null\' de
...
+
+    assert.equal(objWithValue.time.duration, 50);
+    assert.notOk(Object.prototype.hasOwnProperty.call(objNoValue.time, 'duration'));
+
+    objectUtils.deepExtendArraySafe(objWithValue, { time: { duration: undefined } }, true, false, true);
+    objectUtils.deepExtendArraySafe(objNoValue, { time: { duration: undefined } }, true, false, true);
+
+    assert.equal(objWithValue.time.duration, undefined);
```
williamvinogradov, line 181: Let's also check that the ```time``` object has own property ```duration``` here
Because if in runtime we get just an empty time object (e.g ```time: {}```),
the ```time. duration``` value will be ```undefined``` too

packages/devextreme/testing/tests/DevExpress.data/storeArray.tests.js
```diff
@@ -787,6 +787,27 @@ QUnit.test('update with key', function(assert) {
...
+
+    store.update(0, { nested: { a: undefined } }).done(function(data, key) {
+        assert.equal(key, 0);
+
+        const expectedData = {
+            nested: { a: undefined },
+            id: key
+        };
+
+        assert.deepEqual(data, expectedData);
```
williamvinogradov, line 806: Question.

Does the ```deepEqual``` check the key existence?
If it just compares values, let's check that a key with an undefined value exists

packages/devextreme/testing/tests/DevExpress.core/utils.object.tests.js
```diff
@@ -171,7 +171,21 @@ QUnit.test('deepExtendArraySafe utility does not throw an error with \'null\' de
     assert.equal(result.deepProp.toChange, 'changed value');
 });
 
-QUnit.test('deepExtendArraySafe utility does not pollute object prototype', function(assert) {
-    objectUtils.deepExtendArraySafe({ }, JSON.parse('{ "__proto__": { "pollution": true }}'), true);
-    assert.ok(!('pollution' in { }), 'object prototype is not polluted');
+QUnit.test('deepExtendArraySafe sets undefined', function(assert) {
```
williamvinogradov, line 174: Now in one test case, we test two scenarios:
1) The ```shouldCopyUndefined``` false case
2) The ```shouldCopyUndefined``` true case

It's better to have two different tests for two different test scenarios
So, let's split this test case into two

PR #27801, opened 2024-07-18
Core - deepExtendArraySafe - fix assignByReference not merging array values

ivanblinov2k17:


packages/devextreme/js/core/utils/object.js
```diff
@@ -39,16 +39,29 @@ const orderEach = function(map, func) {
     }
 };
 
-const assignValueToProperty = function(target, property, value, assignByReference) {
+const assignValueToProperty = function(target, property, value, extendComplexObject, assignByReference, shouldCopyUndefined) {
```
williamvinogradov: Question.

Maybe adding the new argument at the end of the argument list will be better?
Now in the old code the ```assignByReference: true``` became the ```extendComplexObject: true, assignByReference: undefined```

packages/devextreme/js/core/utils/object.js
```diff
@@ -39,16 +39,29 @@ const orderEach = function(map, func) {
...
+            let itemTarget = item;
+
+            if(Array.isArray(item)) {
+                itemTarget = [];
+            }
+            if(isObject(item)) {
+                itemTarget = {};
+            }
+
+            return deepExtendArraySafe(itemTarget, item, extendComplexObject, assignByReference, shouldCopyUndefined);
```
williamvinogradov: Now the ```deepExtendArraySafe``` recursive function uses the ```assignValueToProperty``` function, that uses the ```deepExtendArraySafe``` inside again

Let's make the design of these functions better and move out this new block to the ```deepExtendArraySafe```

One more point for it, the recursive call of the ```deepExtendArraySafe``` doesn't fit well with the ```assignValueToProperty``` function title

packages/devextreme/js/core/utils/object.js
```diff
@@ -39,16 +39,29 @@ const orderEach = function(map, func) {
...
+    } else if(!assignByReference && Array.isArray(value)) {
+        target[property] = value.map(item => {
+            let itemTarget = item;
+
+            if(Array.isArray(item)) {
+                itemTarget = [];
+            }
+            if(isObject(item)) {
+                itemTarget = {};
+            }
```
williamvinogradov: Let's move out this logic to a separate function

Something like:
```javascript
const getDeepCopyTarget = (item) => {
   if (isObject(item)) {
     return Array.isArray(item) ? [] : {}
   }
   
   return item;
}
```
And use it in ```map```:
```javascript
target[property] = value.map((item) => {
  const itemTarget = getDeepCopyTarget(item);
  return deepExtendArraySafe(itemTarget, item, extendComplexObject, assignByReference, shouldCopyUndefined);
});
```

packages/devextreme/js/core/utils/object.js
```diff
@@ -62,11 +75,15 @@ const deepExtendArraySafe = function(target, changes, extendComplexObject, assig
...
-            newValue = deepExtendArraySafe(goDeeper ? prevValue : {}, newValue, extendComplexObject, assignByReference);
+            newValue = deepExtendArraySafe(goDeeper ? prevValue : {}, newValue, extendComplexObject, assignByReference, shouldCopyUndefined);
         }
 
-        if(newValue !== undefined && prevValue !== newValue) {
-            assignValueToProperty(target, name, newValue, assignByReference);
+        if(
+            Array.isArray(newValue) && !assignByReference ||
+            shouldCopyUndefined && (prevValue !== newValue || prevValue === undefined) ||
+            newValue !== undefined && prevValue !== newValue
```
williamvinogradov: Let's move out this complex conditions outside the ```if``` statement into a few separate variables

packages/devextreme/js/core/utils/object.js
ivanblinov2k17: Strange diff here
As you can see, at the [branch 24_2](https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/js/core/utils/object.js#L51), this flag is already present

PR #27815, opened 2024-07-22
Chat: Basic rendering

marker-dao:


packages/devextreme/js/ui/chat.d.ts
```diff
@@ -104,6 +104,12 @@ export type Message = {
  * @docid
  */
 export interface dxChatOptions extends WidgetOptions<dxChat> {
+    /**
+     * @docid
+     * @default ''
+     * @public
+     */
+    title?: string;
```
ksercs, line 112: Hm, did we approve it with Vlada and Lena?
As far as i remember there was an idea that it'll be a `header` property and it'll be a template, not just a string.

But anyway we can omit this change on this step I think.

packages/devextreme/js/__internal/ui/chat/chat_header.ts
```diff
@@ -0,0 +1,55 @@
...
+    const { title } = this.option();
+
+    $('<p>')
+      .addClass(CHAT_HEADER_TEXT_CLASS)
+      .text((title as any))
+      .appendTo(this.element());
+  }
+
+  _updateTitle(value: any): void {
+    $(`.${CHAT_HEADER_TEXT_CLASS}`).text(value);
```
ksercs: [performance]
Don't u think it's better to cache this element in the `_initMarkup` to some variable like `_$header` to prevent repeat DOM searches?

packages/devextreme/js/__internal/ui/chat/chat_message_bubble.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    $(this.element())
+      .addClass(CHAT_MESSAGE_BUBBLE_CLASS)
+      .text((text as any));
+
+    super._initMarkup();
+  }
+
+  _updateText(value: any): void {
+    $(`.${CHAT_MESSAGE_BUBBLE_CLASS}`).text(value);
```
ksercs: [performance]
Same question here. But do we actually need text runtime update?

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -0,0 +1,139 @@
...
+import MessageBubble from './chat_message_bubble';
+
+const CHAT_MESSAGE_GROUP_CLASS = 'dx-chat-message-group';
+const CHAT_MESSAGE_GROUP_ALIGNMENT_START_CLASS = 'dx-chat-message-group-alignment-start';
+const CHAT_MESSAGE_GROUP_ALIGNMENT_END_CLASS = 'dx-chat-message-group-alignment-end';
+const CHAT_MESSAGE_GROUP_INFORMATION_CLASS = 'dx-chat-message-group-information';
+const CHAT_MESSAGE_TIME_CLASS = 'dx-chat-message-time';
+const CHAT_MESSAGE_NAME_CLASS = 'dx-chat-message-name';
+const CHAT_MESSAGE_BUBBLE_FIRST_CLASS = 'dx-chat-message-bubble-first';
+const CHAT_MESSAGE_BUBBLE_LAST_CLASS = 'dx-chat-message-bubble-last';
```
ksercs, line 17: Kindly review class names with a tech writer

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -0,0 +1,139 @@
...
+      .addClass(alignmentClass);
+
+    super._initMarkup();
+
+    if (alignment === 'start') {
+      const authorName = (messages as any)?.[0].author.name;
+      const initials = `${authorName.charAt(0).toUpperCase()}`;
+
+      this._avatar = this._createComponent($('<div>'), Avatar, {
+        text: initials,
```
ksercs: [code responsibility]
I think that logic related to text calculating by username should be incapsulated in the Avatar component

packages/devextreme/js/__internal/ui/chat/chat_header.ts
```diff
@@ -0,0 +1,55 @@
...
+import Widget from '../widget';
+
+const CHAT_HEADER_CLASS = 'dx-chat-header';
+const CHAT_HEADER_TEXT_CLASS = 'dx-chat-header-text';
+
+export interface ChatHeaderOptions extends WidgetOptions<ChatHeader> {
+  title?: string;
+}
+
+class ChatHeader extends Widget<ChatHeaderOptions> {
```
ksercs: [atchitecture]
Hm, do we really need all these components be inherited by widget? 
@EugeniyKiyashko  your thoughts are also interesting here

packages/devextreme/js/__internal/ui/chat/chat_message_box.ts
```diff
@@ -0,0 +1,47 @@
...
+    $(this.element()).addClass(CHAT_MESSAGE_BOX_CLASS);
+
+    super._initMarkup();
+
+    this._renderTextArea();
+    this._renderButton();
+  }
+
+  _renderTextArea(): void {
+    this._textArea = this._createComponent($('<div>'), TextArea, {});
```
ksercs: [memory leak]
Shouldn't we restore all this caches on disposing? Kindly check it in all places

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    return currentUserId === id;
+  }
+
+  _messageGroupAlignment(id): 'start' | 'end' {
+    return this._isCurrentUser(id) ? 'end' : 'start';
+  }
+
+  _renderMessageListContent(): void {
+    const $content = $('<ul>').addClass(CHAT_MESSAGE_LIST_CONTENT_CLASS);
```
ksercs: [markup]
It's really strange for me to render chat content as a List, at least because there will also be such element as errors and day headers. 
I expect there should be div element.
@EugeniyKiyashko  what do you think?
Maybe also designer can advise here

packages/devextreme/js/__internal/ui/chat/chat_message_box.ts
```diff
@@ -0,0 +1,47 @@
...
+  _textArea?: any;
+
+  _button?: Button;
+
+  _initMarkup(): void {
+    $(this.element()).addClass(CHAT_MESSAGE_BOX_CLASS);
+
+    super._initMarkup();
+
+    this._renderTextArea();
```
ksercs, line 23: [SSR]
Kindly ensure that your markup tests are run in server side mode. Because now I see that you render everything inside of `_initMarkup` and nothing inside of `_renderContent`. Mostly it's okay but need to be checked

packages/devextreme/js/__internal/ui/chat/chat_header.ts
EugeniyKiyashko: Or we should find inside `this.$element()`.find(), not on the whole page

marker-dao: Created _$headerText prop

packages/devextreme/js/__internal/ui/chat/chat_message_bubble.ts
marker-dao: I have deleted this method

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: Agree, I have moved this code into Avatar component

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: I have an argument for and an argument against

Behind:

The list items are related in that they are all chat elements, yes, they have a different purpose: show date, display message group, display error. What unites them is that they all go in order in a common container. And it's like if we added navigation, users would move alternately from one element to the next

Against: 

Different purpose. It's true that elements carry different information

packages/devextreme/js/ui/chat.d.ts
marker-dao: I will omit this change at this step, but will start coordinating this issue with Lena and Vlada.

packages/devextreme/js/__internal/ui/chat/chat_header.ts
EugeniyKiyashko: Now, I don't understand why the header should be a component here. It looks like a common div element. If the header should have its own states and other logic, we can consider it as a component. 
@marker-dao Please share where I can find this information.
Based on the logic inside the PR, it's unclear why we should inherit the component from Widget. If this is being considered for the future, please specify which parts of the widget need to be inherited.
@marker-dao Please share where I can find this information.

packages/devextreme/playground/jquery.html
```diff
@@ -17,45 +17,53 @@
     </script>
 
     <script type="text/javascript" src="../artifacts/js/jquery.js"></script>
-    <!-- HtmlEditor -->
```
EugeniyKiyashko: Please, don't forget to revert these changes

packages/devextreme/js/__internal/ui/chat/chat_header.ts
marker-dao: In the header we pass the text, the text can change. In the future we plan to pass html markup. I want to put all this logic in a separate place. I think it is architecturally wrong to store the header logic in some other place, for example, together with the Chat component logic. Even if we were to draw just 1 div element, it is logically another component, which is easier to manage if its code is written separately. The question is whether Header should be inherited from Widget or from any other component. Or is it better to make it a native class at all?

EugeniyKiyashko: If it's not clear what properties and states the header component should have at the moment (reasons to inherit it from the Widget are unclear).
It makes sense **to keep it as simple as possible**, and in this case, it should not be a Widget.

packages/devextreme/js/__internal/ui/chat/chat_message_box.ts
marker-dao: I checked, I see that elements are removed from the DOM when dispose is called. Only instances are stored in these properties. I believe that in this case we should not care about deleting something by hand

packages/devextreme/js/__internal/ui/chat/chat_header.ts
marker-dao: Okay, I've added that question to the aggregator. We'll deal with it separately

marker-dao: We replaced the Widget with a DOMComponent for the Header

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
```diff
@@ -0,0 +1,166 @@
...
+$chat-box-shadow: 0 0 2px 0 #0000001F, 0 4px 8px 0 #00000014;
+$chat-width: 480px;
+$chat-height: 660px;
+$chat-padding: 20px;
+$chat-border-radius: 24px;
+$chat-information-font-size: 12px;
+$chat-information-color: #707070;
+$chat-avatar-size: 32px;
+$chat-bubble-background-color-primary: #EBF3FC;
+$chat-bubble-background-color-socondary: #F5F5F5;
```
vladaskorohodova:
```suggestion
$chat-bubble-background-color-secondary: #F5F5F5;
```

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
```diff
@@ -0,0 +1,166 @@
...
+  height: $chat-avatar-size;
+  border-radius: 50%;
+  background-color: #878787;
+}
+
+.dx-chat-message-bubble {
+  padding: 8px 12px;
+  max-width: 90%;
+  border-radius: $chat-bubble-border-radius;
+  background-color: $chat-bubble-background-color-socondary;
```
vladaskorohodova:
```suggestion
  background-color: $chat-bubble-background-color-secondary;
```

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
```diff
@@ -0,0 +1,166 @@
...
+.dx-chat-message-group-alignment-end .dx-chat-message-bubble {
+  background-color: $chat-bubble-background-color-primary;
+}
+
+.dx-chat-message-box {
+  display: flex;
+  align-items: center;
+}
+
+.dx-chat .dx-chat-message-box-text-area {
```
vladaskorohodova, line 159: We need to come to agreement whether we write `textarea` or `text-area`. I would suggest the latter (you need to change other names with `textarea` though)

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
marker-dao: [87645b61de0f97694a8ad5c82a0e2b8a5937bf99]

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
marker-dao: [87645b61de0f97694a8ad5c82a0e2b8a5937bf99]

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
marker-dao: [87645b61de0f97694a8ad5c82a0e2b8a5937bf99]

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: Reviewed

packages/devextreme/js/__internal/ui/chat/chat_message_box.ts
marker-dao: I've added SSR testing. It was successful

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
EugeniyKiyashko: We shouldn't add this ignore for all file. It is necessary to eliminate the use of any. In extreme cases, it is necessary to ignore the error only for a specific line

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    super._initMarkup();
+
+    const { name } = this.option();
+
+    const text = this._getAvatarInitials((name as any));
+
+    $('<div>')
+      .addClass(CHAT_MESSAGE_AVATAR_LETTERS_CLASS)
+      .text((text as any))
```
EugeniyKiyashko: Let's avoid type casting where possible

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -0,0 +1,57 @@
...
+    return {
+      ...super._getDefaultOptions(),
+      ...{
+        name: '',
+      },
+    };
+  }
+
+  _getAvatarInitials(name: string): string {
+    const initials = `${name.charAt(0).toUpperCase()}`;
```
EugeniyKiyashko, line 25: We have utility functions for casting strings to different formats

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  _initMarkup(): void {
+    $(this.element()).addClass(CHAT_MESSAGE_AVATAR_CLASS);
+
+    super._initMarkup();
+
+    const { name } = this.option();
+
+    const text = this._getAvatarInitials((name as any));
```
EugeniyKiyashko: We need remove any here.

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -0,0 +1,57 @@
...
+      .addClass(CHAT_MESSAGE_AVATAR_LETTERS_CLASS)
+      .text((text as any))
+      .appendTo(this.element());
+  }
+
+  _optionChanged(args: Record<string, unknown>): void {
+    const { name } = args;
+
+    switch (name) {
+      case 'text':
```
EugeniyKiyashko: Why we consider 'text' when I see that only option 'name' was declared for this component.

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -3,14 +3,26 @@ import $ from '@js/core/renderer';
...
+
+  _messageBox?: MessageBox;
+
+  _messageList?: MessageList;
+
   _getDefaultOptions(): Properties {
     return {
       ...super._getDefaultOptions(),
       ...{
+        title: '',
```
EugeniyKiyashko: Does we have test for checking default value of title? (e.g. defaultOptions.tests.js)

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -21,13 +33,46 @@ class Chat extends Widget<Properties> {
...
+
+    this._chatHeader = this._createComponent($('<div>'), ChatHeader, { title });
+
+    $(this._chatHeader.element()).appendTo(this.element());
+  }
+
+  _renderMessageList(): void {
+    const { items } = this.option();
+
+    this._messageList = this._createComponent($('<div>'), MessageList, {
```
EugeniyKiyashko: It is bad practice to create a component on an element that is not attached to the DOM. First, we add the element to the page, and then we create the component on this element. This helps avoid many rendering issues.

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -21,13 +33,46 @@ class Chat extends Widget<Properties> {
...
+
+    this._renderHeader();
+    this._renderMessageList();
+    this._renderMessageBox();
+  }
+
+  _renderHeader(): void {
+    const { title } = this.option();
+
+    this._chatHeader = this._createComponent($('<div>'), ChatHeader, { title });
```
EugeniyKiyashko: The same https://github.com/DevExpress/DevExtreme/pull/27815/files#r1689329383

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -21,13 +33,46 @@ class Chat extends Widget<Properties> {
...
+    this._messageList = this._createComponent($('<div>'), MessageList, {
+      items,
+      currentUserId: MOCK_CURRENT_USER_ID,
+    });
+
+    $(this._messageList.element()).appendTo(this.element());
+  }
+
+  _renderMessageBox(): void {
+    this._messageBox = this._createComponent($('<div>'), MessageBox, {});
```
EugeniyKiyashko: The same https://github.com/DevExpress/DevExtreme/pull/27815/files#r1689329383

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -21,13 +33,46 @@ class Chat extends Widget<Properties> {
...
+    $(this._messageBox.element()).appendTo(this.element());
   }
 
   _optionChanged(args: Record<string, unknown>): void {
-    const { name } = args;
+    const { name, value } = args;
 
     switch (name) {
+      case 'title':
+        this._chatHeader?.option(name, (value as string));
```
EugeniyKiyashko: It's really that the chatHeader couldn't be created. Do you have tests for this scenario?

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
EugeniyKiyashko:
```suggestion
    const initials = name.charAt(0).toUpperCase();
```

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -0,0 +1,57 @@
...
+const CHAT_MESSAGE_AVATAR_CLASS = 'dx-chat-message-avatar';
+const CHAT_MESSAGE_AVATAR_LETTERS_CLASS = 'dx-chat-message-avatar-letters';
+
+export interface AvatarOptions extends WidgetOptions<Avatar> {
+  name?: string;
+}
+
+class Avatar extends Widget<AvatarOptions> {
+  _getDefaultOptions(): AvatarOptions {
+    return {
```
EugeniyKiyashko, line 15:
```suggestion
   return {
      ...super._getDefaultOptions(),
      name: '',
   };
```

EugeniyKiyashko: Please, check other places too

packages/devextreme/js/__internal/ui/chat/chat_header.ts
```diff
@@ -0,0 +1,65 @@
...
+    // @ts-expect-error
+    super._initMarkup();
+
+    this._renderText();
+  }
+
+  _renderText(): void {
+    const { title } = this.option();
+
+    this._$text = $('<p>')
```
EugeniyKiyashko: Why should we use `<p>` element?

packages/devextreme/js/__internal/ui/chat/chat_header.ts
```diff
@@ -0,0 +1,65 @@
...
+      .appendTo(this.element());
+  }
+
+  _optionChanged(args: Record<string, unknown>): void {
+    const { name, value } = args;
+
+    switch (name) {
+      case 'title':
+        // @ts-expect-error
+        this._$text?.text(value);
```
EugeniyKiyashko, line 53: I didn't find the test that check change of `title` option at runtime.

packages/devextreme/js/__internal/ui/chat/chat_message_box.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import TextArea from '../m_text_area';
+import Widget from '../widget';
+
+const CHAT_MESSAGE_BOX_CLASS = 'dx-chat-message-box';
+const CHAT_MESSAGE_BOX_TEXTAREA_CLASS = 'dx-chat-message-box-text-area';
+const CHAT_MESSAGE_BOX_BUTTON_CLASS = 'dx-chat-message-box-button';
+
+class MessageBox extends Widget<any> {
+  // TODO chech ts
```
EugeniyKiyashko: Why you don't use types from textArea.d.ts?

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: [5d815c8cb9c08ceb8edcac2cefb7e4c48907db00]

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: It's a linter fix. We created a non-mandatory property at the very beginning of the class

packages/devextreme/js/__internal/ui/chat/chat_header.ts
marker-dao: There is a test, it is called “Header text element should have correct text after runtime change”

packages/devextreme/ts/dx.all.d.ts
```diff
@@ -9453,6 +9453,10 @@ declare module DevExpress.ui {
    * @deprecated [depNote:dxChatOptions]
    */
   export interface dxChatOptions extends WidgetOptions<dxChat> {
+    /**
+     * [descr:dxChatOptions.title]
+     */
+    title?: string;
```
EugeniyKiyashko, line 9459: JFYI: It's better add changes for code and public types in separate PRs. This will reduce the number of changes and simplify the review process for API reviewers, tech writers, and developers.

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: I replaced it with a div

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: [90dd3acd3c6dd8b06c21608a237d774ee8b6aa43]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: [90dd3acd3c6dd8b06c21608a237d774ee8b6aa43]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: [90dd3acd3c6dd8b06c21608a237d774ee8b6aa43]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: [90dd3acd3c6dd8b06c21608a237d774ee8b6aa43]

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: [90dd3acd3c6dd8b06c21608a237d774ee8b6aa43]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: [90dd3acd3c6dd8b06c21608a237d774ee8b6aa43]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: In my code, there are four spaces there

packages/devextreme/js/__internal/ui/chat/chat_header.ts
marker-dao: [90dd3acd3c6dd8b06c21608a237d774ee8b6aa43]

packages/devextreme/js/__internal/ui/chat/chat_message_box.ts
marker-dao: [90dd3acd3c6dd8b06c21608a237d774ee8b6aa43]

packages/devextreme/js/__internal/ui/chat/chat.ts
EugeniyKiyashko: Rendering Issues: Components may not render correctly or as expected if they are not attached to the DOM. This can lead to incorrect layouts or visual artifacts. Many our components require calculations for proper rendering. For example, a text area is unlikely to be rendered correctly using the maxHeight option, or a floating label may not be positioned correctly.
You can get more details from our support team; they can provide information about non-working scenarios where components are rendered on elements that are detached from the DOM

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: [5d815c8cb9c08ceb8edcac2cefb7e4c48907db00]

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: [5d815c8cb9c08ceb8edcac2cefb7e4c48907db00]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
EugeniyKiyashko: I only wanted to point out that you are using { name: '' } in multiple places. Spaces doesn't matter here.
'...{}' excess operation, I don't understand why we should leave it as is.

packages/devextreme/js/__internal/ui/chat/chat_header.ts
```diff
@@ -0,0 +1,65 @@
+import type { Properties } from '@js/core/dom_component';
+import DOMComponent from '@js/core/dom_component';
+import type { dxElementWrapper } from '@js/core/renderer';
+import $ from '@js/core/renderer';
+
+const CHAT_HEADER_CLASS = 'dx-chat-header';
+const CHAT_HEADER_TEXT_CLASS = 'dx-chat-header-text';
+
+export interface ChatHeaderProperties extends Properties {
+  title?: string;
```
EugeniyKiyashko: Because this is not public interface or type, may be we might declare 'title' option the following way:
```
title: string;
```

marker-dao: [61fdb892ea26ec5fe331835862bfb84dc42b9632]

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -0,0 +1,137 @@
...
+  _renderName(name, $element): void {
+    $('<div>')
+      .addClass(CHAT_MESSAGE_NAME_CLASS)
+      .text(name)
+      .appendTo($element);
+  }
+
+  _renderTime(timestamp, $element): void {
+    const options: Intl.DateTimeFormatOptions = { hour: '2-digit', minute: '2-digit', hour12: false };
+    const dateTime = new Date(Number(timestamp));
```
EugeniyKiyashko, line 102: Are we going to store the date in 'tick' format?

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  _renderName(name, $element): void {
+    $('<div>')
+      .addClass(CHAT_MESSAGE_NAME_CLASS)
+      .text(name)
+      .appendTo($element);
+  }
+
+  _renderTime(timestamp, $element): void {
+    const options: Intl.DateTimeFormatOptions = { hour: '2-digit', minute: '2-digit', hour12: false };
```
EugeniyKiyashko, line 101: Seems, that this approach is incorrect and we shouldn't work with Intl directly. We also have LDML. Explore how it works in other components

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: We don't have to think about these little things in the current iteration

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: We don't have to think about these little things in the current iteration

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  _isCurrentUser(id): boolean {
+    const { currentUserId } = this.option();
+
+    return currentUserId === id;
+  }
+
+  _messageGroupAlignment(id): 'start' | 'end' {
+    return this._isCurrentUser(id) ? 'end' : 'start';
```
EugeniyKiyashko, line 41: I don't sure that in rtl we should automatically inverse alignment of messages in chat. So, seems in this case it shouldn't be 'end', 'start'. Need additional exploration.

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  _optionChanged(args: Record<string, unknown>): void {
+    const { name, value } = args;
+
+    switch (name) {
+      case 'items':
+        break;
+      case 'currentUserId':
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        this.option(name, (value as any));
```
EugeniyKiyashko: It's better to ignore this line than use type casting.

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: > I'm not sure that in rtl we should automatically inverse alignment of messages in chat
That was never in question. Where did you get that from?

Block start and block end is an appropriate, generalized name for it.

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: I understand, we will handle all TS deficiencies separately

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -0,0 +1,135 @@
...
+      messages: [],
+      alignment: 'start',
+    };
+  }
+
+  _getAlignmentClass(): string {
+    const { alignment } = this.option();
+
+    const alignmentClass = alignment === 'start'
+      ? CHAT_MESSAGE_GROUP_ALIGNMENT_START_CLASS
```
EugeniyKiyashko, line 39: I would prefer to start with a single class for elements that need to be aligned to the opposite edge. In this case, we would simply override the default styles.

marker-dao: It is always more convenient to work with two classes than with one. At least, it makes the whole code more readable. I don't see any disadvantages.

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
EugeniyKiyashko: I hope you added item in your //TODO list

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -0,0 +1,102 @@
...
```
EugeniyKiyashko: Actually I don't understand how it works. 
```
case 'currentUserId':
        this.option(name, (value as any));
```
We change the same option and set the same value that is already stored in this option

marker-dao: [f9684e39d4f173d7cab3ed1eefdc25e3c7e9a701]

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -3,31 +3,74 @@ import $ from '@js/core/renderer';
...
+    this._messageBox = this._createComponent($messageBox, MessageBox, {});
   }
 
   _optionChanged(args: Record<string, unknown>): void {
-    const { name } = args;
+    const { name, value } = args;
 
     switch (name) {
+      case 'title':
+        this._chatHeader?.option(name, (value as string));
```
EugeniyKiyashko: Let's avoid using type casting. The reason is that when you improve type definitions and the error disappears, you won't know if the cast is no longer needed. If it's a comment, it will indicate that it should be removed because the problem no longer exists.

marker-dao: [d4b674e8e7ab763f1263b0db5194a4c54e2564b9]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -0,0 +1,57 @@
...
+      ...super._getDefaultOptions(),
+      name: '',
+    };
+  }
+
+  _getAvatarInitials(name: string): string {
```
EugeniyKiyashko, line 21: Your function _getAvatarInitials extracts the first letter of a given name, why we use plural form?

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    const { name } = this.option();
+
+    if (name) {
+      const text = this._getAvatarInitials(name);
+
+      $letters.text(text);
+    }
+
+    $letters.appendTo(this.element());
```
EugeniyKiyashko: What will the avatar look like if the name is an empty string—an empty circle? Will we consider something else?

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -3,31 +3,75 @@ import $ from '@js/core/renderer';
...
+    this._renderMessageList();
+    this._renderMessageBox();
+  }
+
+  _renderHeader(): void {
+    const { title } = this.option();
+
+    const $header = $('<div>').appendTo(this.element());
+
+    this._chatHeader = this._createComponent($header, ChatHeader, { title: title ?? '' });
```
EugeniyKiyashko: Title default value of title option in ChatHeader component is ''. Why we should process undefined value here?

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
EugeniyKiyashko: Well, I gonna trust you.

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: We don't have a requirement for letters count right now. The function is so named because initials is a generalized word for any number of first letters

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
EugeniyKiyashko: Okey, add this item in //TODO/TOTHINK list

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: We don't have a requirement for that right now

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
EugeniyKiyashko: Well, what about $letters. 
```suggestion
   if (name) {
      const text = this._getAvatarInitials(name);

      $letters.text(text);
    }
```
According to name of this variable I am expecting more that one char.

marker-dao: [afbd50c9a280a38aee60feaec3ba984202894e11]

EugeniyKiyashko (dismissed):

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: [74da06c994b50425c1d5a62c53df51ab7f9f9438]

vladaskorohodova (approved):

EugeniyKiyashko (approved):

IlyaKhD (approved):

PR #27818, opened 2024-07-23
Save dx component classes when using a custom class attribute (T1243735)

dxvladislavvolkov:


ivanblinov2k17 (dismissed):

packages/devextreme-vue/src/core/component.ts
```diff
@@ -44,6 +44,7 @@ export interface IBaseComponent extends ComponentPublicInstance, IWidgetComponen
 }
 
 const includeAttrs = ['id', 'class', 'style'];
+const dxClasses: string[] = [];
```
VasilyStrelyaev: This is going to be one "static" array for all the Vue components on the page, which does not seem good

packages/devextreme-vue/src/core/component.ts
```diff
@@ -296,6 +297,14 @@ function cleanWidgetNode(node: Node) {
   return removedNodes;
 }
 
+function pickOutDxClasses(el: Element, attr: any) {
+  el.classList.forEach((item: string) => {
+    if (attr.class && attr.class.indexOf(item) === -1) {
+      dxClasses.push(item);
```
VasilyStrelyaev: Is it okay that we only accumulate class names in this array and never remove them? Could there be duplicates?

packages/devextreme-vue/src/core/component.ts
```diff
@@ -331,6 +340,8 @@ function initDxComponent() {
       thisComponent.$_instance.endUpdate();
 
       restoreNodes(this.$el, nodes);
+      pickOutDxClasses(this.$el, this.$attrs);
```
VasilyStrelyaev: I'm also worried that if we pick classes on mount, will they be relevant at the moment we restore them?

I mean, the Widget may change its dx classes dynamically, are we sure that what we pick here stays relevant at the moment of restoration?

VasilyStrelyaev (changes requested):

VasilyStrelyaev:
Overall, I think we need just tests on the mentioned scenarios:
1) two components on the page (no mixing dx classes between them)
2) component that updates a couple of times (there should be no dx class duplicates)
3) components whose dynamical class is changed after the Widget changes its dx classes (the latest version of Widget's dx classes should be restored)

Maybe its okay this way

VasilyStrelyaev (approved):

PR #27819, opened 2024-07-23
Scheduler - Resource colors are not updated (T1243679)

tongsonbarbs:


e2e/testcafe-devextreme/tests/scheduler/workSpace.ts
```diff
@@ -360,3 +360,40 @@ test('[T716993]: should has horizontal scrollbar with multiple resources and fix
 }).after(async () => {
   await removeStylesheetRulesFromPage();
 });
+
+test('Scheduler appointments should change color on update resources', async (t) => {
+  await t.wait(2000);
```
Raushen: We should avoid waiting. TestCafe has embedded functionality that can automatically wait expected conditions.

e2e/testcafe-devextreme/tests/scheduler/workSpace.ts
```diff
@@ -360,3 +360,40 @@ test('[T716993]: should has horizontal scrollbar with multiple resources and fix
...
+
+  await t.eval(() => {
```
Raushen: I think we should make the same as in the ticket sample. There is a separate button to invoke some actions. There is another div with id="otherContainer" which can be used for a such button.
Please create a separate "Change resources" button .

Raushen (approved):

PR #27822, opened 2024-07-23
🎹 Pager: Migrate dom_component_wrapper from R1

Raushen:


packages/devextreme/js/__internal/core/r1/dom_component_wrapper.tsx
```diff
@@ -0,0 +1,153 @@
...
+        } = _ref2;
+        this.instance?.option(path, value);
+      });
+      this.instance.endUpdate();
+    }
+    this.prevProps = this.getProperties();
+  }
+
+  getProperties(): any {
+    const normalizedProps = normalizeProps(this.props.componentProps);
```
pomahtri, line 128: in non-migrated `normalizeProps` is a local function, not imported from `inferno`

packages/devextreme/js/__internal/core/r1/dom_component_wrapper.tsx
```diff
@@ -0,0 +1,153 @@
...
+        ref={this.widgetRef}
+        className={this.props.componentProps.className}
+        {...this.props}
+      />
+    );
+  }
+
+  componentWillUpdate(nextProps: DomComponentWrapperProps, nextState, context): void {
+    super.componentWillUpdate(nextProps, nextState, context);
+  }
```
pomahtri, line 77:
```suggestion
```

looks like we can safely delete it

Raushen: Unfortunately not( It was added [here](https://github.com/DevExpress/DevExtreme/pull/27822/commits/1fcb92f7d73344966889989b4baf666438ed7792) because the component had the [linter issue](https://github.com/DevExpress/DevExtreme/actions/runs/10062223281/job/27814206586?pr=27822).

packages/devextreme/js/__internal/core/r1/dom_component_wrapper.tsx
Raushen: [The instruction](https://wiki.devexpress.devx/en/devextreme/workflow/grids-squad/conventions/r1-to-inferno-component-migration) suggests using the imported function.

packages/devextreme/js/__internal/core/r1/dom_component_wrapper.tsx
pomahtri: Oh, yep, type clashes between React and Inferno...
Yep, let's leave it now

packages/devextreme/js/__internal/core/r1/dom_component_wrapper.tsx
pomahtri: I think it's just a coincidence that we have same name of inferno util, and local function
All other renovation files don't have such local function
Let's also ask @williamvinogradov

pomahtri: Discussed with @williamvinogradov, there are two different 'normalizeProps': from inferno, and from local func in this file. We should distinguish them and use inferno's in compiled code, and local's where it was used locally before generation

pomahtri (approved):

PR #27825, opened 2024-07-24
Chat: Add user option

marker-dao:


packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -23,6 +21,7 @@ class Chat extends Widget<Properties> {
       ...super._getDefaultOptions(),
       title: '',
       items: [],
+      user: undefined,
```
EugeniyKiyashko: There are no tests.

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -71,8 +71,13 @@ class Chat extends Widget<Properties> {
...
         this._chatHeader?.option(name, value);
         break;
+      case 'user':
+        // @ts-expect-error
+        this._messageList?.option('currentUserId', value.id);
+        break;
       case 'items':
-        this._invalidate();
+        // @ts-expect-error
+        this._messageList?.option(name, value);
```
EugeniyKiyashko, line 83: Where are the tests?

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -47,13 +46,14 @@ class Chat extends Widget<Properties> {
   }
 
   _renderMessageList(): void {
-    const { items } = this.option();
+    const { items, user } = this.option();
 
+    const currentUserId = user?.id ?? Math.random();
```
EugeniyKiyashko: It's better to use a GUID instead of a floating-point number. Also, I don't understand why user.id might only have a numeric type according to our d.ts.

EugeniyKiyashko: Why we can't add value for `id` property inside _getDefaultOptions method?

marker-dao: [4d1584013e634fda124cfe9dc3d0bedad152ec3d]

marker-dao: [4d1584013e634fda124cfe9dc3d0bedad152ec3d]

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: Was dealing with CI at the time, hadn't gotten around to finishing it yet. It's done now

[af552e68f7ca2ad08a6f510a966f83272b925c78]

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: Was dealing with CI at the time, hadn't gotten around to finishing it yet. It's done now

[af552e68f7ca2ad08a6f510a966f83272b925c78]

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -50,7 +50,7 @@ export type User = {
      * @default undefined
```
EugeniyKiyashko: It seems incorrect that we declare a default value with `undefined`

EugeniyKiyashko (dismissed):

packages/devextreme/js/ui/chat.d.ts
marker-dao: According to the API Reviwers solution, the default type will be written as string

vladaskorohodova (dismissed):

EugeniyKiyashko (dismissed):

EugeniyKiyashko (approved):

PR #27828, opened 2024-07-25
Chat: Add renderMessage method

marker-dao:


packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -63,30 +64,34 @@ class MessageGroup extends Widget<MessageGroupOptions> {
...
-        text: message.text,
-      });
+    this._createComponent($bubble, MessageBubble, {
+      text: message.text,
+    });
+  }
+
+  _renderMessageBubbles(messages): void {
+    messages.forEach((message, index) => {
+      this._renderMessageBubble(message, index, messages.length);
```
EugeniyKiyashko: messages.length is an immutable value here, we can assign this value to a variable

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -27,7 +28,7 @@ class MessageGroup extends Widget<MessageGroupOptions> {
   _getDefaultOptions(): MessageGroupOptions {
     return {
       ...super._getDefaultOptions(),
-      messages: [],
+      items: [],
```
EugeniyKiyashko, line 31: Why are we renaming this property? If it's require I see that some renaming are missed, for example:

this._renderMessageBubbles(**items**);

_renderMessageBubbles(**messages**): void { }

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -118,11 +123,29 @@ class MessageGroup extends Widget<MessageGroupOptions> {
     $messageGroupInformation.appendTo(this.element());
   }
 
+  _removePrevBubbleLastClass(): void {
```
EugeniyKiyashko: This name is unclear. Could we change it to something more understandable?

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: I don't think it is necessary to put such data in a separate variable. This code is easy to read now

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -118,11 +123,29 @@ class MessageGroup extends Widget<MessageGroupOptions> {
...
+    const $bubbles = $(this.element()).find(`.${CHAT_MESSAGE_BUBBLE_CLASS}`);
+    const $lastBubble = $bubbles.eq($bubbles.length - 1);
+
+    $lastBubble.removeClass(CHAT_MESSAGE_BUBBLE_LAST_CLASS);
```
EugeniyKiyashko, line 130: Can't we handle this with CSS for applying borders correctly? It seems we can. What is the reason for using classes here?

packages/devextreme/testing/tests/DevExpress.ui.widgets/chat.tests.js
```diff
@@ -187,6 +189,165 @@ QUnit.module('Message group', moduleConfig, () => {
...
+
+        const newMessage = {
+            author,
+            timestamp: NOW,
+            text: 'NEW MESSAGE',
+        };
+
+        this.instance.renderMessage(newMessage, author);
+
+        const { items } = this.instance._messageList.option();
```
EugeniyKiyashko, line 224: If we need test messageList we should test this component separately.

EugeniyKiyashko (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets/chat.tests.js
```diff
@@ -187,6 +189,165 @@ QUnit.module('Message group', moduleConfig, () => {
...
+            id: MOCK_COMPANION_USER_ID,
+        };
+
+        const newMessage = {
+            author,
+            timestamp: NOW,
+            text: 'NEW MESSAGE',
+        };
+
+        const getMessageGroupElements = () => $(`.${CHAT_MESSAGE_LIST_CONTENT_CLASS}`).find(`.${CHAT_MESSAGE_GROUP_CLASS}`);
```
EugeniyKiyashko: It's bad practice to find element on the full page. The tests become slow because of this, and it leads to errors

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: I'm pretty sure the names don't have to match everywhere. I changed the name only in the _renderMessageBubbles method. In other places I didn't change it, because the word “message” is more appropriate. As for the open API: I renamed messages to items for consistency with other chat components

Fixed in: [a19e1a0b72360b36b6abd1ef8e04f34fdb291917]

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -79,8 +100,30 @@ class MessageList extends Widget<MessageListOptions> {
         this._createMessageGroupComponent(currentMessageGroupItems, currentMessageGroupUserId);
       }
     });
+  }
+
+  _renderMessage(message: Message, newItems: Message[], sender: User): void {
+    this._setOptionWithoutOptionChange('items', newItems);
```
EugeniyKiyashko, line 106: I don't understand why we should use set option without option change. Could you please clarify reasons?

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
EugeniyKiyashko: It's performance optimization. It allows us to avoid recalculating it repeatedly

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
EugeniyKiyashko: It's only a question, and may be were reasons for this change.

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: The length property is a built-in property of the array (it is a predefined property), and retrieving it is a constant access time operation. There is no repeated recalculation of values when retrieving this property. No calculations occur at this point

When optimizing performance, you should focus on the more complex operations you perform inside the loop rather than accessing the length property of the array

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: I have used more common name `_updateLastBubbleClasses` here: [b9e970178359975c6f2be01c472fdf8b32dbdf8e]

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: MessageGroup is implemented using Grid layout. First-/last-child selectors do not work properly in Grid

I don't want to use “crutches”. I don't see any point in using flex-box either, because it will complicate the layout: we will have to create new additional containers to align elements, it will complicate the layout and degrade rendering performance, we will also have to think about naming classes for these containers and maintain their existence

packages/devextreme/testing/tests/DevExpress.ui.widgets/chat.tests.js
marker-dao: I understand you. I try not to use the private options. But we are not writing separate tests for the internal components of Chat right now. However, at the same time, we need to test the functionality

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: You and I discussed this last week. Nothing has changed: we decided that if we change the entire message array, we need to call _invalidate to rerender all message groups

However, in this case, when only one new message is added, we don't want to rerender everything, we draw the message separately and update the items in silent mode

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
EugeniyKiyashko: Okay, we'll assume that you know what you're doing

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
EugeniyKiyashko: Yes, this is true for the parent component, but it's unclear why this synchronization is needed here. Alright, let's leave it as is until the next iterations.

packages/devextreme/testing/tests/DevExpress.ui.widgets/chat.tests.js
marker-dao: [8683a40ba2fa8d24ded40fdd7aa582e011adadc1]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chat.tests.js
EugeniyKiyashko: We discussed this earlier. You told, that in future development, independent tests for internal components will be created, and this test will be moved there. Okay, let's leave it as is.

EugeniyKiyashko (approved):

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
EugeniyKiyashko:
```
            const array = new Array(10000);
            console.log(array.length);
            console.time('performanceMeasurement');

            const length = array.length;
            for (let j = 0; j < length; j++) {
                const a = 100;
            }

            console.timeEnd('performanceMeasurement');
            
            console.time('performanceMeasurement');

            for (let j = 0; j < array.length; j++) {
                const a = 100;
            }

            console.timeEnd('performanceMeasurement');
```
            
On my machine the result is following:

performanceMeasurement: 0.0830078125 ms
performanceMeasurement: 0.10595703125 ms

PR #27860, opened 2024-07-31
Collection, Tabs:  add onSelectionChanging for cancelation feature

nikkithelegendarypokemonster:


packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
```diff
@@ -33,6 +33,27 @@ export default class SelectionStrategy {
     this.options[name] = value;
   }
 
+  onSelectionChanging() {
+    const { addedItemKeys } = this.options;
+    const { removedItemKeys } = this.options;
+    const { addedItems } = this.options;
+    const { removedItems } = this.options;
+    const { selectedItems } = this.options;
+    const { selectedItemKeys } = this.options;
```
ksercs: [copypaste]
I see that u implemented it similar to how `onSelectionChanged` is implemented, but at the moment it's better to use a single destructuring as a more modern syntax I think
```javascript
const {
    addedItemKeys,
    removedItemKeys,
    // ...,
    onSelectionChanging = noop
} = this.options;
```

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
```diff
@@ -33,6 +33,27 @@ export default class SelectionStrategy {
     this.options[name] = value;
   }
 
+  onSelectionChanging() {
```
ksercs: [ts]
```suggestion
  onSelectionChanging(): void {
```

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -467,12 +473,37 @@ const CollectionWidget = BaseCollectionWidget.inherit({
...
+  },
+
   _fireSelectionChangeEvent(addedItems, removedItems) {
     this._createActionByOption('onSelectionChanged', {
       excludeValidators: ['disabled', 'readOnly'],
     })({ addedItems, removedItems });
   },
 
+  _fireSelectionChangingEvent(args) {
+    this._createActionByOption('onSelectionChanging', {
```
ksercs: [performance]
Now you create a new action each time you need to fire it. It means that every time you call [_createActionByOption](https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/js/core/component.js#L286) function additionally instead of calling only a handler. It's not very good from performance point of view.

I see that you implemented it similar to the existing code in `_fireSelectionChangeEvent`, but I believe it's not the best approach. It's better to create actions only once in `_init` (and also in `_optionChanged` for updating after runtime change). Then you will be able to save execution time and also get rid of `_fireSelectionChangeEvent` and `_fireSelectionChangingEvent` method (because you will be able just to call the created action). Kindly explore how it's done in other components (Overlay, for example) and implement it similar way for both `onSelectionChanging` and `onSelectionChanged`

P.s. Don't forget to add tests for a runtime change

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -132,6 +133,9 @@ const CollectionWidget = BaseCollectionWidget.inherit({
       mode: this.option('selectionMode'),
       maxFilterLengthInRequest: this.option('maxFilterLengthInRequest'),
       equalByReference: !this._isKeySpecified(),
+      onSelectionChanging(args) {
+        that._fireSelectionChangingEvent(args);
```
ksercs: [difficult logic, excess private property]
Now u never raise `onSelectionChanging` in Selection module, only in this file's [_updateSelectedItems method](https://github.com/DevExpress/DevExtreme/pull/27860/files#diff-b8238af01aa4298df01c6c2d89d13db014bd657a316cb3bf9012c19da0dfcd86R477). So you can actually remove this and just call necessary action directly in `_updateSelectedItems`. It will also allow you to get rid of the new `this.isSelectionCancel` private property.

```typescript
_updateSelectedItems(args) {
    this._selectionChangingAction(args);
    if (args.cancel) {
    // ...
```

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -467,12 +473,37 @@ const CollectionWidget = BaseCollectionWidget.inherit({
     }
   },
 
+  _updateSelectedItems(args) {
+    this._selection.onSelectionChanging();
+    if (this.isSelectionCancel instanceof Promise) {
```
ksercs: [copypaste]
It's better to use existing `isPromise` util: [link](https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/js/core/utils/type.js#L89).

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -467,12 +473,37 @@ const CollectionWidget = BaseCollectionWidget.inherit({
...
+      this.isSelectionCancel.then((cancel) => {
+        if (!cancel) {
+          this.updatingSelectedItems(args);
+        } else {
+          this._setOptionWithoutOptionChange('selectedIndex', this.initialSelectedIndex);
+        }
```
ksercs: [missed scenario]
What if the passed promise is rejected? `catch` should also be specified.
P.s. don't forget to add a test

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+            this.clock.restore();
+        }
+    }, () => {
+        QUnit.test('cancelling selection event synchronously', function(assert) {
+            this.onSelectionChangingSpy = sinon.spy(function(e) {
+                e.cancel = true;
+            });
+
+            this.tabInstance.option('onSelectionChanging', this.onSelectionChangingSpy);
+            this.$item = this.$tabs.find('.dx-tab').eq(1);
```
ksercs: [copypaste]
Kindly use the `TABS_ITEM_CLASS` constant!
```suggestion
            this.$item = this.$tabs.find(`.${TABS_ITEM_CLASS}`).eq(1);
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+            assert.ok(this.onSelectionChangingSpy.getCall(0).args[0].cancel, 'e.cancel should be set to true');
+            assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should remain 0 when cancelled');
+        });
+
+        QUnit.test('cancelling selection event asynchronously', function(assert) {
+            this.onSelectionChangingSpy = sinon.spy(function(e) {
+                e.cancel = new Promise((resolve) => {
+                    setTimeout(() => {
+                        resolve(true);
+                    }, 5000);
```
ksercs: [refactoring, magic number]
I suggest u store the delay to a variable `CANCEL_DELAY` to get rid of magic numbers.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+            this.$item.trigger('dxclick');
+
+            this.clock.tick(6000);
+            this.onSelectionChangingSpy.getCall(0).args[0].cancel.then((cancel) => {
+                assert.strictEqual(this.onSelectionChangingSpy.callCount, 1, 'onSelectionChanging should be called');
+                assert.strictEqual(this.onSelectionChangedSpy.callCount, 0, 'onSelectionChanged should not be called');
+                assert.ok(cancel, 'e.cancel should be set to true');
+                assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should remain 0 when cancelled');
+            });
+            this.clock.tick(6000);
```
ksercs: [excess code]
You already ticked above to wait until the promise is resolved, so this tick does nothing.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+                        resolve(true);
+                    }, 5000);
+                });
+            });
+
+            this.tabInstance.option('onSelectionChanging', this.onSelectionChangingSpy);
+            this.$item = this.$tabs.find('.dx-tab').eq(1);
+            this.$item.trigger('dxclick');
+
+            this.clock.tick(6000);
```
ksercs: [refactoring]
Here you can use the same `CANCEL_DELAY` variable, no need to make delay greater.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+                onSelectionChanged: this.onSelectionChangedSpy,
+                selectedIndex: 0
+            });
+            this.tabInstance = this.$tabs.dxTabs('instance');
+        },
+        afterEach() {
+            this.clock.restore();
+        }
+    }, () => {
+        QUnit.test('cancelling selection event synchronously', function(assert) {
```
ksercs: [test naming]
Kindly rename all tests using a pattern which we usually rely on: "what should happen / how should it work in this case"
```suggestion
        QUnit.test('selection should be cancelled if cancel parameter is set to true', function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+            this.onSelectionChangingSpy.getCall(0).args[0].cancel.then((cancel) => {
+                assert.strictEqual(this.onSelectionChangingSpy.callCount, 1, 'onSelectionChanging should be called');
+                assert.strictEqual(this.onSelectionChangedSpy.callCount, 0, 'onSelectionChanged should not be called');
+                assert.ok(cancel, 'e.cancel should be set to true');
+                assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should remain 0 when cancelled');
+            });
+            this.clock.tick(6000);
+        });
+
+        QUnit.test('default to false when e.cancel is not updated', function(assert) {
```
ksercs:
```suggestion
        QUnit.test('cancel parameter is false by default', function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+
```
ksercs, line 539:
```suggestion
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+            this.tabInstance.option('onSelectionChanging', this.onSelectionChangingSpy);
+            this.$item = this.$tabs.find('.dx-tab').eq(1);
+            this.$item.trigger('dxclick');
+
+            assert.strictEqual(this.onSelectionChangingSpy.callCount, 2, 'onSelectionChanging should be called');
```
ksercs: [broken scenario]
This assert shows that current implementation is incorrect. `selectionChanging` event is supposed to be raised only once.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+        });
+
+        QUnit.test('immediate cancellation should override promise', function(assert) {
+            this.onSelectionChangingSpy = sinon.spy(function(e) {
+                e.cancel = new Promise((resolve) => {
+                    setTimeout(() => {
+                        resolve(false);
+                    }, 5000);
+                });
+                e.cancel = true;
```
ksercs: [not working test]
This test is actually evergreen because
```javascript
                e.cancel = new Promise((resolve) => {
                    setTimeout(() => {
                        resolve(false);
                    }, 5000);
                });
                e.cancel = true;
```
is processed in the code the same way as 
```javascript
e.cancel = true;
```
Moreover, you don't wait until promise is resolved.
Let's remove this test.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+            this.tabInstance.option('onSelectionChanging', this.onSelectionChangingSpy);
+            this.$item = this.$tabs.find('.dx-tab').eq(1);
+            this.$item.trigger('dxclick');
+
+            assert.strictEqual(this.onSelectionChangingSpy.callCount, 1, 'onSelectionChanging should be called');
+            assert.strictEqual(this.onSelectionChangedSpy.callCount, 0, 'onSelectionChanged should not be called');
+            assert.ok(this.onSelectionChangingSpy.getCall(0).args[0].cancel, 'e.cancel should be set to true');
+            assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should remain 0 when cancelled');
+        });
+    });
```
ksercs, line 631: [test coverage]
Missed scenarios:
1. `onSelectionChanging` with `cancel=true` passed in the initial config
2. cancel promise rejection
3. selection when `selectionMode="miltiple"` (select one more item, deselect item)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+                    setTimeout(() => {
+                        resolve(true);
+                    }, 5000);
+                });
+            });
+
+            this.tabInstance.option('onSelectionChanging', this.onSelectionChangingSpy);
+            this.$item = this.$tabs.find('.dx-tab').eq(1);
+            this.$item.trigger('dxclick');
+
```
ksercs, line 510: [test coverage]
In this place there should also be an assert than selection is not applied until cancel promise is not resolved

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+            assert.strictEqual(this.onSelectionChangingSpy.callCount, 1, 'onSelectionChanging should be called');
+            assert.strictEqual(this.onSelectionChangedSpy.callCount, 0, 'onSelectionChanged should not be called');
+            assert.ok(this.onSelectionChangingSpy.getCall(0).args[0].cancel, 'e.cancel should be set to true');
+            assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should remain 0 when cancelled');
+        });
+    });
+
```
ksercs, line 632: [test coverage]
We implemented functionality on the Collection level => we first of all should test in on the Collection's level, particularly in the `collectionWidgetParts/editingTests.js` file. 

So, kindly add these tests also for a Collection. And please take into account that on the Collection level we should test not only cancellation functionality but also the whole `onSelectionChanging` event. Particularly we should check:
- `addedItems` parameter is correct when some new items are selected
- `removedItems` parameter is correct when some items are deselected
- subscription using `on` method works well: [sample](https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/testing/tests/DevExpress.ui.widgets/dropDownButton.markup.tests.js#L644)

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -428,7 +434,7 @@ const CollectionWidget = BaseCollectionWidget.inherit({
     this._setAriaSelectionAttribute($itemElement, String(isSelected));
   },
 
-  _updateSelectedItems(args) {
+  updatingSelectedItems(args) {
```
ksercs: [naming]
Kindly leave the old name here because this function does exactly what is mentioned in the name - updates selected items:
-  we use `_` as a prefix for all methods which are "private" (are not supposed to be used from external)
- all functions should be named starting with a verb infinitive
```suggestion
  _updateSelectedItems(args) {
```

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -467,12 +473,37 @@ const CollectionWidget = BaseCollectionWidget.inherit({
     }
   },
 
+  _updateSelectedItems(args) {
```
ksercs: [naming]
```suggestion
  _processSelectionChange(args) {
```

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -467,12 +473,37 @@ const CollectionWidget = BaseCollectionWidget.inherit({
...
+    this._selection.onSelectionChanging();
+    if (this.isSelectionCancel instanceof Promise) {
+      this.isSelectionCancel.then((cancel) => {
+        if (!cancel) {
+          this.updatingSelectedItems(args);
+        } else {
+          this._setOptionWithoutOptionChange('selectedIndex', this.initialSelectedIndex);
```
ksercs: [excess code, difficult logic, performance, possible drawbacks]
So, how it works now as far as i see:
- selection is updated in the Selection module => we get notification about it through [onSelectionChanged handler](https://github.com/DevExpress/DevExtreme/pull/27860/files#diff-b8238af01aa4298df01c6c2d89d13db014bd657a316cb3bf9012c19da0dfcd86L135)
- we immediately update the Collection selectedKeys in the [onSelectionChanged](https://github.com/DevExpress/DevExtreme/pull/27860/files#diff-b8238af01aa4298df01c6c2d89d13db014bd657a316cb3bf9012c19da0dfcd86R141) handler but [store initial selectedIndex](https://github.com/DevExpress/DevExtreme/pull/27860/files#diff-b8238af01aa4298df01c6c2d89d13db014bd657a316cb3bf9012c19da0dfcd86R410) to the private variable
- then in this function we restore selection if `cancel` is true

This logic is very difficult to be understood, it definitely takes additional execution time (because we change selectedKeys and then restore it if selection was cancelled) and mostly has side effects, particularly it can not work with multiple selection.

So I suggest you reconsider the solution and try to get rid of selection restoring. Maybe for this it's enough to move  [this row](https://github.com/DevExpress/DevExtreme/pull/27860/files#diff-b8238af01aa4298df01c6c2d89d13db014bd657a316cb3bf9012c19da0dfcd86L137) to the `updatingSelectedItems ` - not sure, should be checked.

packages/devextreme/js/__internal/ui/selection/m_selection.ts
```diff
@@ -123,6 +124,10 @@ export default class Selection {
     this._selectionStrategy.setSelectedItems(keys, items);
   }
 
+  onSelectionChanging() {
```
ksercs: [hypothesis, excess code?]
I see that you never call this method in the Selection module. So possible it's excess and Selection should not have a `onSelectionChanging` at all? For the moment it seems to me like implementation on the Collection level is enough

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +422,100 @@ QUnit.module('General', () => {
...
+            });
+
+            this.tabInstance.option('onSelectionChanging', this.onSelectionChangingSpy);
+            this.$item = this.$tabs.find('.dx-tab').eq(1);
+            this.$item.trigger('dxclick');
+
+            assert.strictEqual(this.onSelectionChangingSpy.callCount, 1, 'onSelectionChanging should be called');
+            assert.strictEqual(this.onSelectionChangedSpy.callCount, 0, 'onSelectionChanged should not be called');
+            assert.ok(this.onSelectionChangingSpy.getCall(0).args[0].cancel, 'e.cancel should be set to true');
+            assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should remain 0 when cancelled');
```
ksercs: [test coverage]
There a lot of properties related to selection, not only `selectedIndex`: `selectedItem`, `selectedItems`, `selectedItemKeys`. I prefer we add asserts for all these properties to ensure they are well synchronized.

packages/devextreme/js/__internal/ui/selection/m_selection.ts
nikkithelegendarypokemonster: [b62f308](https://github.com/DevExpress/DevExtreme/pull/27860/commits/b62f30857d83924a9f3c3b9e14c53016e7d67bdb)

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
nikkithelegendarypokemonster: [b62f308](https://github.com/DevExpress/DevExtreme/pull/27860/commits/b62f30857d83924a9f3c3b9e14c53016e7d67bdb)

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
nikkithelegendarypokemonster: [b62f308](https://github.com/DevExpress/DevExtreme/pull/27860/commits/b62f30857d83924a9f3c3b9e14c53016e7d67bdb)

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
nikkithelegendarypokemonster: [b62f308](https://github.com/DevExpress/DevExtreme/pull/27860/commits/b62f30857d83924a9f3c3b9e14c53016e7d67bdb)

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
nikkithelegendarypokemonster: [b62f308](https://github.com/DevExpress/DevExtreme/pull/27860/commits/b62f30857d83924a9f3c3b9e14c53016e7d67bdb)

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
nikkithelegendarypokemonster: [b62f308](https://github.com/DevExpress/DevExtreme/pull/27860/commits/b62f30857d83924a9f3c3b9e14c53016e7d67bdb)

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
nikkithelegendarypokemonster: [b62f308](https://github.com/DevExpress/DevExtreme/pull/27860/commits/b62f30857d83924a9f3c3b9e14c53016e7d67bdb)

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
nikkithelegendarypokemonster: [b62f308](https://github.com/DevExpress/DevExtreme/pull/27860/commits/b62f30857d83924a9f3c3b9e14c53016e7d67bdb)

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
nikkithelegendarypokemonster: [b62f308](https://github.com/DevExpress/DevExtreme/pull/27860/commits/b62f30857d83924a9f3c3b9e14c53016e7d67bdb)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: I think this is correct anton... the reason its 2 because when in initialization our collection widget automatically runs selection procedure in tabs u can see it on initMarkup method when initializing our markup its also initialize selection in which it runs onSelectionChanged method that runs our onSelectionChanging first always the reason its 2 because the second call happens during clicking of next item which calls again the same steps as initialization in which it triggers our second call for onSelectionChanging  u can start to look or debug in selectItem method in collection edit. 


P.S:
based from my research when running _normalizeSelectedItems method it calls setSelection method which triggers on our selection API selectedItemKeys which it runs again onSelectionChanged method...

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [48e4ac1](https://github.com/DevExpress/DevExtreme/pull/27860/commits/48e4ac14372f22296446a8d3f5ca0147a58bb4b2)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [bebf086](https://github.com/DevExpress/DevExtreme/pull/27860/commits/bebf0867b9c1214b545d0d2fb3c7c46c2e3ff9c8)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
nikkithelegendarypokemonster: [c8cac03](https://github.com/DevExpress/DevExtreme/pull/27860/commits/c8cac032a9f69365fa631fa6fd494f442d53de32)

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
...
+            this.$item.trigger('dxclick');
+
+            assert.strictEqual(onSelectionChangingSpy.callCount, 2, 'onSelectionChanging should be called');
+            assert.strictEqual(this.onSelectionChangedSpy.callCount, 0, 'onSelectionChanged should not be called');
+            assert.ok(onSelectionChangingSpy.getCall(0).args[0].cancel, 'e.cancel should be set to true');
+            assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should remain 0 when cancelled');
+            assert.deepEqual(this.tabInstance.option('selectedItem'), { text: '0' }, 'selectedItem should remain the first item');
+            assert.deepEqual(this.tabInstance.option('selectedItems'), [{ text: '0' }], 'selectedItems should remain the first item');
+            assert.deepEqual(this.tabInstance.option('selectedItemKeys'), [{ text: '0' }], 'selectedItemKeys should remain the first key');
+        });
```
ksercs: [excess code]
I don't see a difference between this test and the previous one

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
...
+                });
+            });
+
+            this.tabInstance.option('onSelectionChanging', onSelectionChangingSpy);
+            this.tabInstance.option('onSelectionChanged', this.onSelectionChangedSpy);
+            const $item = this.$tabs.find(`.${TABS_ITEM_CLASS}`).eq(1);
+
+            $item.trigger('dxclick');
+
+            assert.strictEqual(onSelectionChangingSpy.callCount, 1, 'onSelectionChanging should be called twice after click');
```
ksercs: 'onSelectionChanging should be called twice after click' but u assert that it's called only once

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
...
+            assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'Initial selectedIndex should be 0');
+            assert.strictEqual(onSelectionChangingSpy.callCount, 0, 'onSelectionChanging should not be called initially');
+
+            $item.trigger('dxclick');
+
+            this.clock.tick(this.CANCEL_DELAY);
+
+            assert.strictEqual(onSelectionChangingSpy.callCount, 1, 'onSelectionChanging should be called once after click');
+            assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should remain 0 before promise resolves');
+            assert.ok(isPromise(onSelectionChangingSpy.getCall(0).args[0].cancel), 'e.cancel should be a Promise');
```
ksercs: [excess code]
This assert checks nothing: u set cancel to promise manually => it cannot be not a promise
```suggestion
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
...
+
+            onSelectionChangingSpy.getCall(0).args[0].cancel.then((cancel) => {
+                assert.ok(cancel, 'e.cancel should be set to true');
```
ksercs: [excess code]
This assert is excess as well. You don't check e.cancel, only resolved value here. So you assert a code written in your test, not in the code base.
```suggestion
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
...
+            $item2.trigger('dxclick');
+            $item3.trigger('dxclick');
+
+            // De-Select the first item
+            $item1.trigger('dxclick');
+
+            const selectedItems = this.tabInstance.option('selectedItems');
+            assert.deepEqual(selectedItems, expectedResult, 'Items with indexes 1 and 2 should be selected');
+            assert.deepEqual(this.tabInstance.option('selectedItem'), { text: '1' }, 'selectedItem should remain the first item');
+            assert.deepEqual(this.tabInstance.option('selectedItems'), expectedResult, 'selectedItems should remain the first item');
```
ksercs:
```suggestion
            assert.deepEqual(this.tabInstance.option('selectedItems'), expectedResult, 'selectedItems should container first and second item');
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
...
+                    { text: '3' }
+                ],
+                onSelectionChanging: this.onSelectionChangingSpy,
+                onSelectionChanged: this.onSelectionChangedSpy,
+                selectByClick: true,
+                selectedIndex: 0
+            };
+            this.init = (options) => {
+                this.$tabs = $('#tabs');
+                this.tabInstance = this.$tabs
```
ksercs: Let's rename to make the tests more compact? Some rows are really long now
```suggestion
                this.tabs = this.$tabs
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
...
+                assert.strictEqual(onSelectionChangingSpy.callCount, 1, 'onSelectionChanging should still be called once');
+                assert.strictEqual(this.onSelectionChangedSpy.callCount, 0, 'onSelectionChanged should be called after promise rejection');
+                assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should change after cancellation promise rejection');
+                assert.deepEqual(this.tabInstance.option('selectedItem'), { text: '0' }, 'selectedItem should remain the first item');
+                assert.deepEqual(this.tabInstance.option('selectedItems'), [{ text: '0' }], 'selectedItems should remain the first item');
+                assert.deepEqual(this.tabInstance.option('selectedItemKeys'), [{ text: '0' }], 'selectedItemKeys should remain the first key');
+            });
+        });
+
+        QUnit.test('Selection should be cancelled asynchronously when e.cancel is resolved', function(assert) {
```
ksercs:
```suggestion
        QUnit.test('Selection should be cancelled asynchronously when e.cancel is resolved with true', function(assert) {
```

ksercs: [test coverage]
Let's add a similar tests checking that selection is proceed correctly when promise is resolved with false

packages/devextreme/testing/tests/DevExpress.ui/collectionWidgetParts/editingTests.js
```diff
@@ -25,6 +25,60 @@ class TestComponent extends CollectionWidget {
 
 const getItemElement = (instance, itemIndex) => instance.itemElements().eq(itemIndex);
 
+module('onSelectionChanging event triggering', () => {
```
ksercs:
```suggestion
module('onSelectionChanging event', () => {
```

packages/devextreme/testing/tests/DevExpress.ui/collectionWidgetParts/editingTests.js
```diff
@@ -25,6 +25,60 @@ class TestComponent extends CollectionWidget {
...
+    test('added and removed selection should be correct', function(assert) {
```
ksercs:
```suggestion
    test('addedItems and removedItems parameters should be correct after selection change', function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui/collectionWidgetParts/editingTests.js
```diff
@@ -25,6 +25,60 @@ class TestComponent extends CollectionWidget {
...
+        const $element = $('#cmp');
+        const instance = new TestComponent($element, {
+            items: [0, 1, 2, 3],
+            selectionMode: 'multiple',
+            onSelectionChanging: function(args) {
+                assert.ok(args.addedItems.length <= 1, 'unloaded items does not present in selection');
```
ksercs: Why so strange asserts? Let's make it more precise:
```javascript
assert.strictEqual(args.addedItems, ..., 'addedItems parameter is correct');
```

packages/devextreme/testing/tests/DevExpress.ui/collectionWidgetParts/editingTests.js
```diff
@@ -25,6 +25,60 @@ class TestComponent extends CollectionWidget {
...
+        instance.on('selectionChanging', selectionChangingHandler);
+        instance.on('selectionChanged', selectionChangedHandler);
+        assert.strictEqual(instance.option('selectedIndex'), 0, 'Initially selectedIndex should be 0');
+        instance.selectItem(1);
+        assert.strictEqual(selectionChangingHandler.callCount, 1, 'selectionChanging should be raised once');
+        assert.strictEqual(selectionChangedHandler.callCount, 0, 'selectionChanged should not be raised');
+        assert.strictEqual(instance.option('selectedIndex'), 0, 'selectedIndex should be remain unchanged');
+    });
+});
+
```
ksercs, line 263: [test coverage]
Missing tests:
1. `selection is not changed if cancel parameter is set to promise which resolves with true`
2. `selection is changed if cancel parameter is set to promise which rejects`
3. `selection is changed if cancel parameter is set to promise which resolves with false`
4. `should have correct parameters` (check that all necessary parameters are passed including `component`, `event` and other)

packages/devextreme/testing/tests/DevExpress.ui/collectionWidgetParts/editingTests.js
```diff
@@ -25,6 +25,60 @@ class TestComponent extends CollectionWidget {
...
+
```
ksercs: [test coverage]
Now `onSelectionChanging` and `onSelectionChanged` properties runtime change works other way than previously. Kindly add tests for both events.

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -428,11 +446,31 @@ const CollectionWidget = BaseCollectionWidget.inherit({
     this._setAriaSelectionAttribute($itemElement, String(isSelected));
   },
 
+  onSelectionChangingHandler(args) {
```
ksercs: [excess code]
We created a new method to replace one simple code row with another simple code row. It seems to be excess

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -133,9 +151,9 @@ const CollectionWidget = BaseCollectionWidget.inherit({
       maxFilterLengthInRequest: this.option('maxFilterLengthInRequest'),
       equalByReference: !this._isKeySpecified(),
       onSelectionChanged(args) {
+        that.selectionChangingArgs = { ...args, cancel: false };
```
ksercs: [excess code]
Nikki, we create new private instance properties only in specific cases when it's really the single way to solve the problem. Usually that's very common data which can be accessed from several methods of the class.

That's not our current case, you can get rid of this new private property without changing the algorithm.

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -428,11 +446,31 @@ const CollectionWidget = BaseCollectionWidget.inherit({
...
+          this.option('selectedItems', this._getItemsByKeys(args.selectedItemKeys, args.selectedItems));
+          this._updateSelectedItems(args);
+        }
+      }).catch((e) => {
+        console.error(e);
+      });
+    } else if (!this.selectionChangingArgs.cancel) {
+      this.option('selectedItems', this._getItemsByKeys(args.selectedItemKeys, args.selectedItems));
+      this._updateSelectedItems(args);
+    }
```
ksercs, line 478: [copypaste]
Can we reduce copypaste? You run almost the same code when cancel is promise and when is not

packages/devextreme/js/__internal/ui/collection/m_collection_widget.edit.ts
```diff
@@ -428,11 +446,31 @@ const CollectionWidget = BaseCollectionWidget.inherit({
...
+  _processSelectionChanging(args) {
+    this.onSelectionChangingHandler(this.selectionChangingArgs);
+    if (isPromise(this.selectionChangingArgs.cancel)) {
+      this.selectionChangingArgs.cancel.then((cancel) => {
+        if (!cancel) {
+          this.option('selectedItems', this._getItemsByKeys(args.selectedItemKeys, args.selectedItems));
+          this._updateSelectedItems(args);
+        }
+      }).catch((e) => {
+        console.error(e);
```
ksercs: [incorrect behavior]
We should not log any error. If promise is rejected, it's the same case when it's resolved with false

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
```diff
@@ -33,14 +33,38 @@ export default class SelectionStrategy {
     this.options[name] = value;
   }
 
+  onSelectionChanging(): void {
```
ksercs: Is this code still necessary?

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
...
+            assert.strictEqual(onSelectionChangingSpy.callCount, 2, 'onSelectionChanging should be called');
+            assert.strictEqual(this.onSelectionChangedSpy.callCount, 0, 'onSelectionChanged should not be called');
+            assert.ok(onSelectionChangingSpy.getCall(0).args[0].cancel, 'e.cancel should be set to true');
+            assert.strictEqual(this.tabInstance.option('selectedIndex'), 0, 'selectedIndex should remain 0 when cancelled');
+            assert.deepEqual(this.tabInstance.option('selectedItem'), { text: '0' }, 'selectedItem should remain the first item');
+            assert.deepEqual(this.tabInstance.option('selectedItems'), [{ text: '0' }], 'selectedItems should remain the first item');
+            assert.deepEqual(this.tabInstance.option('selectedItemKeys'), [{ text: '0' }], 'selectedItemKeys should remain the first key');
+        });
+
+        QUnit.test('Selection should not change when onSelectionChanging is cancelled with a rejected promise', function(assert) {
```
ksercs: I believe that promise rejection is the same case as promise resolving with false

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
...
+            const onSelectionChangedSpy = sinon.spy();
+            this.reinit({
+                onSelectionChanging: onSelectionChangingSpy,
+                onSelectionChanged: onSelectionChangedSpy,
+            });
+
+            const $item = this.$tabs.find(`.${TABS_ITEM_CLASS}`).eq(1);
+            $item.trigger('dxclick');
+
+            assert.strictEqual(count, 2, 'onSelectionChanging should be called twice');
```
ksercs: [incorrect logic]
I saw your comment about this question and I understand that Selection's `onSelectionChanged` is fired on initialized. However, I see that Collection's `onSelectionChanged` is not fired, and I don't see any product reason to fire it on init because selection is anyway applied. So I ask you to stop firing selectionChanging on init.

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -422,6 +423,200 @@ QUnit.module('General', () => {
 });
 
 QUnit.module('Tab select action', () => {
+    QUnit.module('onSelectionChanging', {
+        beforeEach() {
+            this.onSelectionChangingSpy = sinon.stub();
```
ksercs: Let's name it as a stub while it's stub :)

packages/devextreme/testing/tests/DevExpress.ui/collectionWidgetParts/editingTests.js
```diff
@@ -25,6 +25,243 @@ class TestComponent extends CollectionWidget {
...
+                assert.strictEqual(selectionChangingHandler.callCount, 1, 'selectionChanging should be raised once');
+                assert.strictEqual(selectionChangedHandler.callCount, 1, 'selectionChanged should be raised once');
+                assert.strictEqual(instance.option('selectedIndex'), 1, 'selectedIndex should be updated');
+                done();
+            });
+        });
+    });
+});
+
+
```
ksercs:
```suggestion
```

ksercs (dismissed):

ksercs (approved):

EugeniyKiyashko (approved):

PR #27863, opened 2024-08-01
Chat: Add scroll down 

marker-dao:


packages/devextreme-scss/scss/widgets/base/chat/_index.scss
```diff
@@ -14,8 +14,8 @@ $chat-bubble-border-radius: 12px;
 $chat-text-area-height: 40px;
 
 .dx-chat {
-  display: flex;
-  flex-direction: column;
+  display: grid;
+  grid-template-rows: 24px 1fr minmax(40px, auto);
```
EugeniyKiyashko, line 18: You must be sure that all the properties you use are supported in all the browsers we claim to support. Seems, previously we didn't use grid layout in our styles.

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
```diff
@@ -117,7 +117,7 @@ class MessageGroup extends Widget<MessageGroupOptions> {
     const { timestamp, author } = message;
     const $messageGroupInformation = $('<div>').addClass(CHAT_MESSAGE_GROUP_INFORMATION_CLASS);
 
-    this._renderName(author.name, $messageGroupInformation);
+    this._renderName(author?.name, $messageGroupInformation);
```
EugeniyKiyashko: The incoming arguments or `_renderName` function is not typing. It's unclear when it's need to check exist author or not, and I'm not sure that undefined value is expected for name. 

Can we consider(process) the possible values of autor property and type incomming parameters of  renderName() method in this PR?

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -113,11 +114,21 @@ class MessageList extends Widget<MessageListOptions> {
...
 
     this._createMessageGroupComponent([message], sender.id);
+    this._scrollContentToLastMessage();
+  }
+
+  _scrollContentToLastMessage(): void {
+    const lastMessageGroup = this._messageGroups?.[this._messageGroups.length - 1];
+    const element = lastMessageGroup?.$element()[0];
+
+    element?.scrollIntoView(true);
```
EugeniyKiyashko: It is more appropriate to use our Scrollable or ScrollView components. They have the appropriate API. Their solution will need to be revised to use these components.

If ScrollIntoView is used, the entire page will scroll to the chat message when a new message arrives. This behavior will clearly be unexpected for users. 
Other issues: 
- if no scrolling is required, I will be navigated to this message.
- If count of messages inside group is a lot, this method doesn't scroll to the last message.

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -113,11 +114,21 @@ class MessageList extends Widget<MessageListOptions> {
...
         return;
       }
     }
 
     this._createMessageGroupComponent([message], sender.id);
+    this._scrollContentToLastMessage();
+  }
+
+  _scrollContentToLastMessage(): void {
+    const lastMessageGroup = this._messageGroups?.[this._messageGroups.length - 1];
```
EugeniyKiyashko: Too many uses of `?.`, in each of lines.

Can we not execute code and remove `?.` in case `this._messageGroups.length === 0`. Something like this:
```
if (this._messageGroups.length === 0) {
    return;
}
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chat.tests.js
```diff
@@ -134,6 +146,44 @@ QUnit.module('Chat initialization', moduleConfig, () => {
...
+    QUnit.test('Last message should be into view', function(assert) {
+        const messages = generateMessages(31);
+
+        this.instance.option({ items: messages });
+
+        const $messages = this.$element.find(`.${CHAT_MESSAGE_BUBBLE_CLASS}`);
+        const lastMessage = $messages[$messages.length - 1];
+        const lastMessageOffset = Math.ceil($(lastMessage).offset().top);
+
+        assert.strictEqual(lastMessageOffset, -9459);
```
EugeniyKiyashko: It is completely unclear what we are testing with this test and why we are using these magic numbers. According to the test name 'Last message should be into view,' we are checking that the element will be visible, but it cannot be visible because the element is absolutely positioned and located off-screen.

It would make more sense to check that, after initialization or after a message is entered, we have scrolled to the maximum position.

packages/devextreme/testing/tests/DevExpress.ui.widgets/chat.tests.js
```diff
@@ -134,6 +146,44 @@ QUnit.module('Chat initialization', moduleConfig, () => {
 
         assert.strictEqual(invalidateStub.callCount, 1);
     });
+
+    QUnit.test('Last message should be into view', function(assert) {
```
EugeniyKiyashko: There is no tests that check scroll position after initialization.

We can declare the reinit() utility method to simplify routine actions as follows:
https://github.com/marker-dao/DevExtreme/blob/24_2_chat_scrolling/packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/dateRangeBox.tests.js#L74

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
marker-dao: https://caniuse.com/?search=grid
https://caniuse.com/?search=grid-template-rows
https://caniuse.com/?search=minmax

marker-dao: These styles make it much easier to work with JS. That's why I replaced flex-box with grid

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: Okay, I have integrate Scrollable: [8f3fdaf2307dcd86e20f36de975133dea7e3fecf]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chat.tests.js
marker-dao: [5f471b926b7031dac4c622c538f731e7447e4126]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chat.tests.js
marker-dao: [5f471b926b7031dac4c622c538f731e7447e4126]

In the first new two tests, we can't override the scrollable scrollToElement method on initialization (just like I did in the last test). So the only way to test the performance is to get the value of scrollTop()

packages/devextreme/js/__internal/ui/chat/chat_message_group.ts
marker-dao: [40b5a534a502be1b3c2ae1cacbed67aadd0c66ba]

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: [40b5a534a502be1b3c2ae1cacbed67aadd0c66ba]

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
```diff
@@ -30,29 +30,24 @@ $chat-text-area-height: 40px;
   padding-bottom: 4px;
 }
 
-.dx-chat-header-text {
-  margin: 0;
-}
-
 .dx-chat-message-list {
   box-sizing: border-box;
   flex: 1;
```
EugeniyKiyashko: 'Flex: 1' doesn't have matter after that you use 'display: grid' 
image

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
```diff
@@ -30,29 +30,24 @@ $chat-text-area-height: 40px;
...
-  overflow: auto;
+  overflow-y: auto;
```
EugeniyKiyashko:
```suggestion
  overflow: hidden;
```

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -69,12 +76,18 @@ class MessageList extends Widget<MessageListOptions> {
     this._messageGroups?.push(messageGroup);
   }
 
+  _renderScrollable(): void {
+    this._scrollable = this._createComponent('<div>', Scrollable, {});
```
EugeniyKiyashko: Why we don't declare options: useKeyboard, showScrollbar mode? Well, let's extract it to separate task

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -39,7 +44,9 @@ class MessageList extends Widget<MessageListOptions> {
 
     super._initMarkup();
 
+    this._renderScrollable();
     this._renderMessageListContent();
```
EugeniyKiyashko, line 48: The scrollable component needs to know what content it is working with during the initialization stage. If there is no content, it will incorrectly calculate the sizes of the simulated scrollbar. 
image
With the current approach, you add content without the component knowing about it. In this case, you need to call the update() method to inform the component about the content change.

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
```diff
@@ -113,11 +126,27 @@ class MessageList extends Widget<MessageListOptions> {
...
       }
     }
 
     this._createMessageGroupComponent([message], sender.id);
+    this._scrollContentToLastMessageGroup();
+  }
+
+  _scrollContentToLastMessageGroup(): void {
+    if (
+      !this._messageGroups || this._messageGroups.length === 0 || !this._scrollable || !hasWindow()
```
EugeniyKiyashko:
```suggestion
      !(this._messageGroups?.length && this._scrollable && hasWindow())
```

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: We have a task to integrate ScrollView. Now we have problems with Scrollable when invalidate is launched, so this option is not quite suitable. I used useNative: true to avoid the problems you describe

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
marker-dao: [466559665d37a68dea3cd49bcce32b2fe48c763f]

packages/devextreme-scss/scss/widgets/base/chat/_index.scss
marker-dao: [466559665d37a68dea3cd49bcce32b2fe48c763f]

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: [466559665d37a68dea3cd49bcce32b2fe48c763f]

packages/devextreme/js/__internal/ui/chat/chat_message_list.ts
marker-dao: https://github.com/DevExpress/DevExtreme/pull/27863#discussion_r1701537214

EugeniyKiyashko (approved):

PR #27877, opened 2024-08-07
Treelist - A11y - Only provides information about the nesting level (aria-level)

tongsonbarbs:


packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -194,17 +194,23 @@ export class ResizingController extends modules.ViewController {
     return 'dxDataGrid-ariaDataGrid';
   }
 
-  private _setAriaLabel(): void {
+  protected _getExpandableWidgetAriaLabel() {
+    return 'dxDataGrid-ariaExpandableInstruction';
```
pomahtri: minor
Let's move `messageLocalization.format` into this method or rename it to clarify that it returns message id instead of actual aria-label

packages/devextreme/js/localization/messages/ar.json
```diff
@@ -182,6 +182,7 @@
         "dxDataGrid-ariaExpand": "توسيع",
         "dxDataGrid-ariaCollapsedRow": "Collapsed row",
         "dxDataGrid-ariaExpandedRow": "Expanded row",
+        "dxDataGrid-ariaExpandableInstruction": "Use Enter key to toggle expand and collapse a focused node",
```
Raushen: We do not need an instruction for DataGrid because it supports standard hotkeys.

e2e/testcafe-devextreme/tests/treeList/accessibility/aria.ts
```diff
@@ -0,0 +1,92 @@
...
+    .eql('true');
+}).before(async () => createWidget('dxTreeList', options));
+
+test('Aria expanded should be toggled false on Ctrl + ← keypress', async (t) => {
+  const treeList = new TreeList('#container');
+  const expandableRow = treeList.getDataCell(0, 0).element;
+
+  await t
+    .click(expandableRow)
+    .pressKey('ctrl+left')
```
Raushen, line 127: Row should be expanded first. And we need a check aria-checked state by default.

packages/devextreme/js/localization/messages/ru.json
```diff
@@ -195,6 +196,7 @@
         "dxDataGrid-masterDetail": "Cell with details",
 
         "dxTreeList-ariaTreeList": "Иерархическая таблица данных с {0} строками и {1} колонками",
+        "dxTreeList-ariaExpandableInstruction": "используйте комбинацию клавиш Ctrl + →, чтобы развернуть, и комбинацию клавиш Ctrl + ←, чтобы свернуть выделенный узел.",
```
Raushen: Russian translate should also be in English.

Raushen (changes requested):

packages/devextreme/js/localization/messages/ar.json
```diff
@@ -195,6 +196,7 @@
         "dxDataGrid-masterDetail": "Cell with details",
 
         "dxTreeList-ariaTreeList": "Tree list with {0} rows and {1} columns",
+        "dxTreeList-ariaExpandableInstruction": "Use Ctrl + → key combination to expand and Ctrl + ← key combination to collapse a focused node",
```
pomahtri: Question
Do all screen readers read non-ascii `←` char correctly?

tongsonbarbs: Yes, it is being pronounced "right/left arrow"

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
tongsonbarbs: Got it

packages/devextreme/js/localization/messages/ar.json
tongsonbarbs: Sorry, not sure. I only tried NVDA

packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
```diff
@@ -841,6 +841,10 @@ export class ColumnsView extends ColumnStateMixin(modules.View) {
 
     this._setCellAriaAttributes($cell, cellOptions);
 
+    if (options.row.node && options.row.node.hasChildren) {
```
Raushen: This code should be included in the method above `_setCellAriaAttributes`. If necessary, pass there an additional argument.

e2e/testcafe-devextreme/tests/treeList/accessibility/aria.ts
```diff
@@ -0,0 +1,105 @@
...
+      dataField: 'Task_Subject',
+      width: 300,
+    },
+  ],
+};
+
+test('Aria expanded should be toggled true on Ctrl + → keypress', async (t) => {
+  const treeList = new TreeList('#container');
+  const expandableRow = treeList.getDataRow(0).element;
+  const expandableCell = treeList.getDataCell(0, 0).element;
```
Raushen: You need a test that track aria-expanded of the treeList.getDataCell(0, **1**).element;
The general rule: Every mistake made while you are working on the task should be covered by a test. Otherwise it can return in the future.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -199,11 +199,16 @@ export class ResizingController extends modules.ViewController {
...
-      // @ts-expect-error Badly typed format method
+    // @ts-expect-error Badly typed format method
       .format(widgetAriaLabel, totalItemsCount, columnCount);
     // @ts-expect-error Badly typed dxElementWrapper
     const $ariaLabelElement = this.component.$element().children(`.${GRIDBASE_CONTAINER_CLASS}`);
+    // @ts-expect-error Treelist Variable
+    const expandableWidgetAriaLabel = messageLocalization.format(this._expandableWidgetAriaId);
 
+    if (expandableWidgetAriaLabel !== '') {
+      this.component.setAria('roledescription', expandableWidgetAriaLabel, $ariaLabelElement);
```
Raushen: Does aria-label attribute have the same effect here? If so, let's change aria-roledescription to aria-label.

packages/devextreme/js/__internal/grids/tree_list/rows/m_rows.ts
```diff
@@ -121,12 +120,16 @@ class TreeListRowsView extends RowsView {
   protected _createRow(row) {
     const node = row && row.node;
     const $rowElement = super._createRow.apply(this, arguments as any);
-
     if (node) {
       this.setAria('level', row.level + 1, $rowElement);
-
       if (node.hasChildren) {
-        this.setAria('expanded', row.isExpanded, $rowElement);
+        const { cells } = row;
```
Raushen: Why do you need this block of changes? I removed it and nothing changed.

Raushen (changes requested):

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
tongsonbarbs: Aria-label attribute for this element is validated on some QUnit and e2e test cases. This is the reason why I put the instructions on roledescription instead..

packages/devextreme/js/__internal/grids/tree_list/rows/m_rows.ts
tongsonbarbs: Line 126 - 131: This is the line of codes being called on expand/collapse toggle. Maybe there was a cache on your end which causes the changes to not apply.

Line 132: This adds the aria-expanded attribute to <tr> element on row creation.. same as above, this is being called on expand/collapse toggle.

packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
```diff
@@ -792,11 +792,23 @@ export class ColumnsView extends ColumnStateMixin(modules.View) {
...
+      node,
+      rowType,
+    } = options || {};
+
+    if (cellElements && node?.hasChildren) {
+      cellElements.forEach((cell) => {
+        cell.cellElement = cell.cellElement instanceof HTMLElement ? $(cell.cellElement) : cell.cellElement;
+        this.setAria('expanded', cell.isExpanded, cell.cellElement);
+      });
+    }
```
Raushen, line 806: The advantage of writing the code to the appropriate place is that it might already contain everything needed for it to work. In your case, the set of cells is already stored in the variable:
`const $cells = $rowElement.children();`
So you do not need to destructure _options_ to _cells: cellElements_. Also _rowType_ variable is not used.
Row's cells have the same _aria-expanded_ attribute as the parent row. So you do not need to use _cells[i].isExpanded_.

After the simplifications I got the folowing code. Please check:
```
protected _updateCells($rowElement, $newRowElement, columnIndices, options?) {
    const that = this;
    const $cells = $rowElement.children();
    const $newCells = $newRowElement.children();
    const highlightChanges = this.option('highlightChanges');
    const cellUpdatedClass = this.addWidgetPrefix(CELL_UPDATED_ANIMATION_CLASS);

    if (options?.node?.hasChildren) {
      $cells.each(function () {
        that.setAria('expanded', options.isExpanded, $(this));
      });
    }
...

```

Raushen (changes requested):

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -199,12 +199,13 @@ export class ResizingController extends modules.ViewController {
...
-      // @ts-expect-error Badly typed format method
+    // @ts-expect-error Badly typed format method
       .format(widgetAriaLabel, totalItemsCount, columnCount);
     // @ts-expect-error Badly typed dxElementWrapper
     const $ariaLabelElement = this.component.$element().children(`.${GRIDBASE_CONTAINER_CLASS}`);
-
-    this.component.setAria('label', widgetStatusText, $ariaLabelElement);
+    // @ts-expect-error Treelist Variable
+    const expandableWidgetAriaLabel = messageLocalization.format(this._expandableWidgetAriaId);
+    this.component.setAria('label', `${widgetStatusText}. ${expandableWidgetAriaLabel}`, $ariaLabelElement);
```
Raushen: There are different ways to build the description. To avoid changing in the existing test you could use the following:
```
const labelParts = [widgetStatusText];
if (expandableWidgetAriaLabel) {
    labelParts.push(expandableWidgetAriaLabel);
}
this.component.setAria('label', labelParts.join['. '], $ariaLabelElement);
```

packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
tongsonbarbs: This is great. Thanks for this!!

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
Raushen: Or
```
const labelParts = [
    // @ts-expect-error Badly typed format method
    messageLocalization.format(widgetAriaLabel, totalItemsCount, columnCount),
    messageLocalization.format(this._expandableWidgetAriaId),
];
const label =  labelParts.filter((x) => x).join['. '];
this.component.setAria('label', label, $ariaLabelElement);
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -173,7 +173,7 @@ QUnit.module('Initialization', baseModuleConfig, () => {
         assert.equal(filterRowWrapper.getTextEditorInput(1).attr('aria-describedby'), headersWrapper.getHeaderItem(0, 4).attr('id'));
 
         assert.equal(dataGridWrapper.getElement().find('.dx-datagrid').attr('role'), 'group', 'group role');
-        assert.equal(dataGridWrapper.getElement().find('.dx-datagrid').attr('aria-label'), 'Data grid with 2 rows and 4 columns', 'aria-label of the datagrid container');
+        assert.equal(dataGridWrapper.getElement().find('.dx-datagrid').attr('aria-label'), 'Data grid with 2 rows and 4 columns. ', 'aria-label of the datagrid container');
```
Raushen: I noticed there is no similar test for TreeList. You can use TestCafe for it, not QUnit.

tongsonbarbs: I have added it to the existing testcase I created

Raushen (dismissed):

packages/devextreme/js/localization/messages/es.json
```diff
@@ -195,6 +195,7 @@
         "dxDataGrid-masterDetail": "Cell with details",
 
         "dxTreeList-ariaTreeList": "Tree list with {0} rows and {1} columns",
+        "dxTreeList-ariaExpandableInstruction": "Use Ctrl + right arrow key combination to expand and Ctrl + left arrow key combination to collapse a focused node",
```
vladaskorohodova:
```suggestion
        "dxTreeList-ariaExpandableInstruction": "Press Ctrl + right arrow to expand the focused node and Ctrl + left arrow to collapse it",
```

tongsonbarbs: Thanks Vlada.. pushing commit now and waiting for CI check..

vladaskorohodova (approved):

Raushen (approved):

PR #27884, opened 2024-08-09
Security - scan artifacts

pomahtri:

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -4,20 +4,13 @@ on:
   workflow_dispatch:
 
 jobs:
-  analyze:
-    name: Analyze
+  analyze-javascript:
+    name: Analyze JavaScript
     runs-on: ubuntu-latest
     permissions:
       actions: read
       contents: read
       security-events: write
-
-    strategy:
-      fail-fast: false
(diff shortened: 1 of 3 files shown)


.github/workflows/codeql.yml
```diff
@@ -1,42 +1,96 @@
...
+      security-events: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
 
       - name: Get head SHA
         id: get-head-sha
-        run: echo "SHA=$(git rev-parse origin/${{ matrix.branch }})" >> "$GITHUB_OUTPUT"
+        run: echo "SHA=$(git rev-parse origin/${{ github.ref_name }})" >> "$GITHUB_OUTPUT"
```
IlyaKhD, line 87: Consider extracting these CodeQL steps as a composite action

pomahtri: I think it's better to leave it here, because looks like it will not be reused anywhere (we will have only this file)
And it seems to be too complex for just 2 times usage

IlyaKhD: 👌🏻

IlyaKhD (approved):

PR #27899, opened 2024-08-13
feat: React config component composition

VasilyStrelyaev:


packages/devextreme-react/src/accordion.ts
```diff
@@ -124,25 +124,28 @@ type IItemProps = React.PropsWithChildren<{
...
+        tmplOption: "titleTemplate",
+        render: "titleRender",
+        component: "titleComponent"
+      }],
+    },
+  });
+};
+
+const Item = Object.assign<typeof _componentItem, NestedComponentMeta>(_componentItem, {
+  componentType: "option",
```
IlyaKhD, line 147:
```suggestion
  componentKind: "option",
```

There is always mess with the word `type` in metadata, need to be careful when using it to avoid misunderstanding.

We're using [TypeScript inspired](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions
) word "kind" for such cases as this one.
A "type" is a TypeScript type and a "kind" is a specificator of a variation.

In this example `Shape` is a type (TypeScript-world type), and `'circle' | 'square' | 'triangle'` is a kind (variation specificator)
```
type Shape =
  | { kind: "circle"; radius: number }
  | { kind: "square"; x: number }
  | { kind: "triangle"; x: number; y: number };
```

packages/devextreme-react/src/core/component-base.tsx
```diff
@@ -374,35 +352,35 @@ const ComponentBase = forwardRef<ComponentBaseRef, any>(
       }
     ), [instance.current, element.current, createWidget]);
 
+    // @ts-expect-error TS2339
```
IlyaKhD: Can this be removed? Since `children` is `any`, it breaks type-checking in all places where `children` is used - not only on the line 356, but on others.

packages/devextreme-react/src/core/use-option-scanning.ts
```diff
@@ -0,0 +1,159 @@
...
+    optionElement.descriptor.initialValuesProps,
+    optionElement.descriptor.templates,
+  );
+
+  const configCollections: Record<string, IConfigNode[]> = {};
+  const configCollectionMaps: Record<string, Record<string, number>> = {};
+  const configs: Record<string, IConfigNode> = {};
+  const templates: ITemplate[] = [];
+
+  Children.map(
```
IlyaKhD: Return value is not used
```suggestion
  Children.forEach(
```

packages/devextreme-react/src/core/use-option-scanning.ts
```diff
@@ -0,0 +1,159 @@
...
+    IConfigNode,
+    NestedOptionContextContent,
+  ] {
+  const parentContext = useContext(NestedOptionContext);
+
+  const {
+    parentFullName,
+  } = parentContext;
+
+  const updateToken = Math.random();
```
IlyaKhD: Consider using a `Symbol` instead.

packages/devextreme-react/src/core/use-option-scanning.ts
```diff
@@ -0,0 +1,159 @@
...
+        if (template) {
+          templates.push(template);
+        }
+      }
+    },
+  );
+
+  const childComponentCounter = useRef(0);
+
+  const configNode: IConfigNode = {
```
IlyaKhD: Extract kinda `createConfig` function, like:

```
function useOptionScanning() {
  const config = createConfig(...);
  const context: NestedOptionContextContent = {
    ...
    onChildOptionsReady() {
      config.attachChild(...);
    }
  }
  
  useLayoutEffect(() => {
    ...
    config.updateTemplates(hasTranscludedContent)
  })
}

function createConfig(...): {
  node: {};
  attachChild();
  updateTemplates();
}
```

packages/devextreme-react/src/core/use-option-scanning.ts
```diff
@@ -0,0 +1,159 @@
...
+    configNode.templates = configNode.templates.filter((template) => !template.isAnonymous);
+    const hasTranscludedContent = templateContainer.childNodes.length > 0;
+
+    const getHasTranscludedContent = () => {
+      if (optionElement.descriptor.isCollection) {
+        return hasTranscludedContent;
+      }
+
+      return parentFullName.length > 0 ? hasTranscludedContent : false;
+    };
```
IlyaKhD: - [ ] Is it really need to be a function?

```suggestion
    const hasTranscludedContent = templateContainer.childNodes.length > 0 && (
      optionElement.descriptor.isCollection || parentFullName.length > 0
    );
```

packages/devextreme-react/src/core/configuration/utils.ts
```diff
@@ -19,6 +21,21 @@ export function parseOptionName(name: string): IOptionInfo | ICollectionOptionIn
   };
 }
 
+export function buildFullName(node: IConfigNode): string {
```
IlyaKhD: make it more specific and move to the config-node module
```suggestion
export function buildNodeFullName(node: IConfigNode): string {
```

IlyaKhD:
⚠️ `components-base.ts` should be refactored first, it's one function for 400 having serious problems with code coupling. At least added lines should be extracted in

packages/devextreme-react/src/accordion.ts
VasilyStrelyaev: Can we please leave this rename to a follow-up PR to avoid the need to re-release `generators` right now?

packages/devextreme-react/src/core/helpers.ts
```diff
@@ -1,12 +1,47 @@
...
+});
+
+export interface TemplateDiscoveryContextContent {
+  discoveryRendering: boolean;
+}
+
+export const TemplateDiscoveryContext = createContext<TemplateDiscoveryContextContent>({
+  discoveryRendering: false,
+});
+
```
IlyaKhD: Consider extracting all contexts into ./contexts.ts module or ./contexts/ folder (each context has a module then)

dxvladislavvolkov (approved):

mpreyskurantov (approved):

PR #27901, opened 2024-08-13
RadioGroup: Move aria on value container

marker-dao:


packages/devextreme/js/__internal/ui/collection/m_collection_widget.base.ts
```diff
@@ -383,11 +383,17 @@ const CollectionWidget = Widget.inherit({
     this.setAria('activedescendant', isDefined(this.option('focusedElement')) ? this.getFocusedItemId() : null, $target);
   },
 
+  _getTargetForSettingId($target) {
```
ksercs, line 386:
```suggestion
  _getIdTarget($target) {
```

packages/devextreme/js/__internal/ui/radio_group/m_radio_collection.ts
```diff
@@ -59,9 +66,29 @@ class RadioCollection extends CollectionWidget<Properties> {
...
+        // eslint-disable-next-line spellcheck/spell-checker
+        labelledby: contentId,
+      };
+
+      this.setAria(aria, $radioContainer);
     }
 
     super._postprocessRenderItem(args);
+
+    if (!html) {
```
ksercs, line 81: Looks like it's not mandatory to do it after `super._postprocessRenderItem(args);` => we can move it to the condition above

PR #27919, opened 2024-08-16
Splitter: should display properly inside a Popup component when used in React (T1241434)

EugeniyKiyashko:


marker-dao (approved):

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -214,17 +216,23 @@ class Splitter extends CollectionWidget<Properties> {
 
   _attachHoldEvent(): void {}
 
-  _resizeHandler(): void {
-    if (!this._shouldRecalculateLayout) {
-      return;
-    }
+  _isAttached(): boolean {
+    return !!contains(domAdapter.getBody(), $(this.element()).get(0));
```
ksercs, line 220: Hm, does it work in ShadowDOM? I mean if popup content is a shadow root

EugeniyKiyashko: That's why I opted for the implementation using the utility, and this approach is already in use. If it doesn't work [here (template_base.js)](https://github.com/EugeniyKiyashko/DevExtreme/blob/24_2_splitter_bug/packages/devextreme/js/core/templates/template_base.js#L37-L38), it likely won't work there either, which is improbable.

ksercs: Oh, yeah, we have shadowDOM support in the `contains` method

ksercs (approved):

nikkithelegendarypokemonster (approved):

PR #27929, opened 2024-08-19
TabPanel: add onSelectionChanging for cancelation feature

nikkithelegendarypokemonster:


packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
```diff
@@ -33,6 +33,53 @@ export default class SelectionStrategy {
...
+          }
+        })
+        .catch(() => {
+          callback();
+        });
+    } else if (!cancelResult) {
+      callback();
+    } else {
+      cancelCallback();
+    }
```
pomahtri, line 80: minor
if you have possibility, please rewrite to async/await. Awaiting non-promise will return value as is, because of that code will be easier to read

However, maybe because of that change, datagrid qunit tests will fail :(

```suggestion
    try {
      if (await cancelResult) {
        callback();
      } else {
        cancelCallback();
      }
    } catch {
      callback();
    }
```

pomahtri:
Please ask for review from grids once ready, as we use this module too

EugeniyKiyashko (approved):

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
```diff
@@ -17,6 +17,8 @@ export default class StandardStrategy extends SelectionStrategy {
 
   _lastRequestData?: any;
 
+  _isCancelingInProgress?: boolean;
```
pomahtri, line 20: minor
```suggestion
  _isSelectingInProgress?: boolean;
```

pomahtri (approved):

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
ksercs: It's possible but not so easy to rewrite - in your suggestion part of cases is not considered.
And also it would be strange to await this function call in parent function which works with deferred

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
ksercs: This property is exactly about cancelling. Selecting if more general algorithm including also items loading from store for example

PR #27934, opened 2024-08-20
DataGrid - Revert button flickers in `cell` editing mode (T1247323)

tongsonbarbs:


e2e/testcafe-devextreme/tests/dataGrid/editing.ts
```diff
@@ -2701,3 +2701,27 @@ test('Focus behavior should be correct when editing cells', async (t) => {
     mode: 'batch',
   },
 }));
+
+test.only('DataGrid - Revert button flickers in `cell` editing mode (T1247323)', async (t) => {
```
Raushen: There is the "only" forbidden method here

e2e/testcafe-devextreme/tests/dataGrid/editing.ts
```diff
@@ -2701,3 +2701,27 @@ test('Focus behavior should be correct when editing cells', async (t) => {
     mode: 'batch',
   },
 }));
+
+test('DataGrid - Revert button flickers in `cell` editing mode (T1247323)', async (t) => {
+  const dataGrid = new DataGrid('#container');
+  const revertButton = Selector('.dx-revert-button');
```
Raushen: Please use [the model method](https://github.com/DevExpress/DevExtreme/blob/3d11172b99a5750ce41b12367c39d4d87a5b2ebb/packages/testcafe-models/dataGrid/index.ts#L373). We should avoid using string selectors in the test. Please check every time is there an alternative in the model. If no, it is like you need to create one.

PR #27952, opened 2024-08-25
PivotGrid - OLAP - The E4000 error occurs if filterValues contains an item with the "unknown member" value (T1236954)

pomahtri:

### DIFF
--- a/packages/devextreme/js/__internal/grids/pivot_grid/xmla_store/m_xmla_store.ts
+++ b/packages/devextreme/js/__internal/grids/pivot_grid/xmla_store/m_xmla_store.ts
@@ -670,7 +670,9 @@ const XmlaStore = Class.inherit((function () {
 
   function preparePathValue(pathValue, dataField?) {
     if (pathValue) {
-      const shouldWrapPathValue = isString(pathValue) && pathValue.includes('&');
+      const shouldWrapPathValue = isString(pathValue) && (
+        pathValue.includes('&') || pathValue.startsWith(`${dataField}.`)
+      );
       pathValue = shouldWrapPathValue ? pathValue : `[${pathValue}]`;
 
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/grids/pivot_grid/xmla_store/m_xmla_store.ts
```diff
@@ -670,7 +670,10 @@ const XmlaStore = Class.inherit((function () {
 
   function preparePathValue(pathValue, dataField?) {
```
Raushen, line 671: Let's refactor the method: 
- 674 and 678 lines are the same, so let's extract the variable isStartWithDataField
- Avoid nesting the if statement and create a guard clause in the method
- Move number type of the path in a guard clause
- Let's avoid argument mutation and create a variable `preparedPath`
- Let's join 679 string with slice operator to the ternary operator above
- To be able to slice dataField it is necessary the path should starts with dataField, so it is needed &&-operator at 674 line
- Rename `shouldWrapPathValue` to `shouldSlicePathValue`
- Add types to the function and its arguments

I got the following:
```
  function preparePathValue(
    pathValue: string | number,
    dataField: string | undefined,
  ): string | undefined {
    if (!pathValue) {
      return undefined;
    }
    if (!isString(pathValue)) {
      return `[${pathValue}]`;
    }

    let preparedPath = pathValue;

    const shouldSlicePathValue = dataField && pathValue.includes('&') && pathValue.startsWith(`${dataField}.`);
    preparedPath = shouldSlicePathValue
      ? pathValue.slice(dataField.length + 1, pathValue.length)
      : `[${pathValue}]`;

    return preparedPath;
  }
```

Raushen (dismissed):

Alyar666 (approved):

PR #27970, opened 2024-08-29
Chat: Hide title prop and prevent Header rendering

marker-dao:


packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -46,7 +48,10 @@ class Chat extends Widget<Properties> {
 
     super._initMarkup();
 
-    this._renderHeader();
+    if (shouldHeaderBeRendered) {
```
EugeniyKiyashko: It is strange to see a magic variable used to prevent the chat header from rendering. It's better to check that the default value has not been changed; that should be sufficient. We discussed this way.

This approach also allows us to test it in the usual way without using global variables or QUnit.skip.

marker-dao: You are proposing an implementation that is too complex.

marker-dao: [1b9f995421176c5da1ede84866276a1288870383]

marker-dao: Also, now, we must support all scenarios with title change in runtime

EugeniyKiyashko (dismissed):

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: [b2a70f90cdda0c35379a3881f4e03131bf01d0a8]

pomahtri (approved):

EugeniyKiyashko (approved):

vladaskorohodova (approved):

PR #27975, opened 2024-08-29
Fix filtering issues for string in different locales (T1242756)

GoodDayForSurf:

### DIFF
--- a/packages/devextreme/js/__internal/data/m_array_query.ts
+++ b/packages/devextreme/js/__internal/data/m_array_query.ts
@@ -248,10 +248,9 @@ const SortIterator = Iterator.inherit({
 
 const compileCriteria = (function () {
   let langParams: Record<string, any> = {};
-  let useUpperCase = false;
 
   // eslint-disable-next-line @typescript-eslint/naming-convention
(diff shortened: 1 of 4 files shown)


packages/devextreme/js/__internal/data/m_array_query.ts
```diff
@@ -400,6 +391,9 @@ const compileCriteria = (function () {
 
   return function (crit, options) {
     langParams = options || {};
+    const locale = langParams?.locale?.toLowerCase();
+
+    useUpperCase = locale && !!['hy', 'el'].find((code) => locale === code || locale.startsWith(`${code}-`));
```
dxvladislavvolkov: I suggest getting rid of overriding the variable and just declare it here. And then pass it to the functions:

const useUpperCase = locale && !!['hy', 'el'].find((code) => locale === code || locale.startsWith(`${code}-`));
...
compileGroup(crit, useUpperCase);
compileUnary(crit, useUpperCase);
compileBinary(crit, useUpperCase);

VasilyStrelyaev (approved):

dxvladislavvolkov (approved):

PR #27993, opened 2024-09-03
Refactor devextreme-vue for support TS in IDE

GoodDayForSurf:


packages/devextreme-vue/src/core/index.ts
```diff
@@ -1,11 +1,7 @@
-import * as vue2Stategy from './strategy/vue2/index';
-import * as vue3Stategy from './strategy/vue3/index';
-import { isVue3 } from './version';
+import * as vue3Strategy from './strategy/vue3/index';
```
dxvladislavvolkov: Do we need strategy folder? Looks like we can move 'prepareComponentConfig', 'prepareConfigurationComponentConfig' and 'prepareExtensionComponentConfig' to this module

GoodDayForSurf: yes it can be removed

alexslavr (approved):

dxvladislavvolkov (approved):

PR #27996, opened 2024-09-04
Chat: Add avatarUrl

marker-dao:


packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -37,10 +58,25 @@ class Avatar extends Widget<Properties> {
...
+    this._$image.attr('alt', name || 'avatar');
+  }
+
+  _updateUrl(): void {
+    const { url } = this.option();
+
+    this._$image.attr('src', url ?? '');
+  }
+
+  _updateImageVisibility(): void {
```
ksercs: [logic]
I don't understand why algorithm is so difficult. For what need we render image if url is empty?

I expect: if `url` property is passed we render visible image, if not - we render initials. 
i.e.
- image should be rendered when there is no url
- initials should not be rendered when there is an image

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -37,10 +58,25 @@ class Avatar extends Widget<Properties> {
       .appendTo(this.element());
   }
 
-  _updateInitials(): void {
+  _updateName(): void {
     const { name } = this.option();
 
     this._$initials.text(this._getInitials(name));
+    // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+    this._$image.attr('alt', name || 'avatar');
```
ksercs: TODO: add this point to the card about localization
```suggestion
    this._$image.attr('alt', name || 'Avatar');
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/avatar.tests.js
```diff
@@ -63,5 +66,64 @@ QUnit.module('ChatAvatar', moduleConfig, () => {
                 assert.strictEqual(this.$element.text(), expectedInitials);
             });
         });
+
+        QUnit.test('img element should have correct alt attribute', function(assert) {
```
ksercs, line 127: Let's add a separate module like `image rendering` and move these tests there. In the future there will be also tests for `avatarAlt`.

P.s. don't forget to add a task about `avatarAlt` to aggregator

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: I thought about that, but consider the following: what about runtime option changes? If the url becomes empty, do we remove the img element and render the letters? What if it's the other way around? If there were letters but the url appears, do we remove the letters and render the img? This looks more complicated.

ksercs: Yes, it should be this way I believe.
Your scenarios are not real anyway - there is no ability to change `url` or `name` at runtime.
It's not okay to render elements which are not necessary

marker-dao: This will be a public component, users will be able to change this as they wish.

ksercs: Public component will have very different API.
I don't mind to support this scenario but believe it should be implemented using our usual way - remove one element and add other one

marker-dao: [4faa83a68b229fadabc4e402a28e04364ffa5a7b]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: [4faa83a68b229fadabc4e402a28e04364ffa5a7b]

packages/devextreme-scss/scss/widgets/base/chat/layout/chat-avatar/_index.scss
```diff
@@ -1,7 +1,18 @@
...
+  position: relative;
   grid-row: span 3;
   display: flex;
   align-items: center;
   justify-content: center;
   border-radius: 999em;
 }
+
+.dx-chat-message-avatar-image {
+  position: absolute;
```
EugeniyKiyashko: Why should we use absolute positioning for this element?

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -7,18 +7,23 @@ import Widget from '@ts/core/widget/widget';
...
-  _$initials!: dxElementWrapper;
+  _$initials?: dxElementWrapper;
+
+  _$image?: dxElementWrapper;
 
   _getDefaultOptions(): Properties {
     return {
       ...super._getDefaultOptions(),
       name: '',
+      url: '',
```
EugeniyKiyashko, line 24: Kindly use defaultOption.tests.js to test this setting

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -27,8 +32,41 @@ class Avatar extends Widget<Properties> {
...
+    this._renderImage();
+    this._renderInitials();
+  }
+
+  _renderImage(): void {
+    this._$image?.remove();
+
+    const { url } = this.option();
+
+    if (url) {
```
EugeniyKiyashko: Can we extract if conditions from '_renderImage' and '_renderInitials' into 'rencerAvatraContent' method? Seems, in this case it will more readable, something like this
```
if(shouldRenderInitials) {
     this._renderInitials();
} else {
    this._renderImage();
}
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageGroup.tests.js
```diff
@@ -115,6 +115,27 @@ QUnit.module('MessageGroup', moduleConfig, () => {
...
         });
+
+        QUnit.test('avatar component should be initialized with correct url property', function(assert) {
+            [
+                { items: [{}], passedUrlValue: undefined },
+                { items: [{ author: {} }], passedUrlValue: undefined },
+                { items: [{ author: undefined }], passedUrlValue: undefined },
+                { items: [{ author: { avatarUrl: undefined } }], passedUrlValue: undefined },
+                { items: [{ author: { avatarUrl: null } }], passedUrlValue: null },
+                { items: [{ author: { avatarUrl: '' } }], passedUrlValue: '' },
```
EugeniyKiyashko, line 126: One more condition: avatarUrl: '   '(string value with space)

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
EugeniyKiyashko: Also seems we shouldn't save _$initials & _$image and easily to remove $element().content() and render it's again when it's need

EugeniyKiyashko (changes requested):

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: I get your basic idea, but I don't think it should be exactly like this: shouldRenderInitials. I don't think we should move the name property out of _renderInitials.

packages/devextreme-scss/scss/widgets/base/chat/layout/chat-avatar/_index.scss
marker-dao: [4da065f777adfedddf3bf741d68b4ba898273434]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: [4da065f777adfedddf3bf741d68b4ba898273434]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
marker-dao: [4da065f777adfedddf3bf741d68b4ba898273434]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageGroup.tests.js
marker-dao: [4da065f777adfedddf3bf741d68b4ba898273434]

packages/devextreme/js/__internal/ui/chat/chat_avatar.ts
```diff
@@ -27,20 +30,73 @@ class Avatar extends Widget<Properties> {
 
     super._initMarkup();
 
-    this._renderInitialsElement();
-    this._updateInitials();
+    this._renderAvatarContent();
+  }
+
+  _renderAvatarContent(): void {
+    this._$content?.remove();
```
EugeniyKiyashko, line 37: I would prefer not to use a reference to an element here and, if possible, avoid using it to make the functions pure. You will have to expand this functionality further, and you might arrive at such a solution in future PRs.

EugeniyKiyashko (approved):

ksercs (approved):

PR #28021, opened 2024-09-10
fix: selectbox keyboard navigation is broken in React Strict Mode (T1230720)

VasilyStrelyaev:


dxvladislavvolkov (approved):

packages/devextreme-react/build.config.js
```diff
@@ -1,7 +1,7 @@
 
 module.exports = {
   src: './src/**/*.{ts,tsx}',
-  testSrc: './src/**/__tests__/*.*',
+  testSrc: './src/**/__tests__/**/*.*',
```
ivanblinov2k17, line 4: Why regex here is changed? Can't find tests in your pr, that satisfy your regex version, but don't satisfy previous one

VasilyStrelyaev: It's a fix for an issue that appeared when I introduced a new test subfolder - the test was included in npm package:

image

It turned out only files directly in `__tests__` directories were removed from the package. When I created a deeper-lying test file, the removal logic did not work for it.

ivanblinov2k17 (approved):

PR #28024, opened 2024-09-11
Map: add providerConfig option to .d.ts

Zedwag:


vladaskorohodova (dismissed):

packages/devextreme/js/ui/map.d.ts
```diff
@@ -205,6 +205,24 @@ export interface dxMapOptions extends WidgetOptions<dxMap> {
      * @public
      */
     focusStateEnabled?: boolean;
+    /**
+     * @docid
+     * @default { mapId: '', useAdvancedMarkers: true }
+     * @public
+     */
+    googleMapConfig?: {
```
ksercs: We don't want to make this option name too specific because it can be used in the future for other provider's config. So maybe it's better to name it like `providerConfig`?

Zedwag: renamed

ksercs (dismissed):

packages/devextreme/js/ui/map.d.ts
```diff
@@ -205,6 +205,24 @@ export interface dxMapOptions extends WidgetOptions<dxMap> {
      * @public
      */
     focusStateEnabled?: boolean;
+    /**
+     * @docid
+     * @default { mapId: '', useAdvancedMarkers: true }
```
EugeniyKiyashko: We declare defaults for mapId, and UseAdvancedMarkers inside. Should we declare default object with all properties? Moreover, mapId and useAdvancedMarkes are optional here

packages/devextreme/js/ui/map.d.ts
```diff
@@ -205,6 +205,24 @@ export interface dxMapOptions extends WidgetOptions<dxMap> {
...
+    providerConfig?: {
+      /**
+       * @docid
+       * @default ""
+       */
+      mapId?: string;
+      /**
+       * @docid
+       * @default true
+       * @deprecated
```
EugeniyKiyashko, line 222: Why you add property 'useAdvancedMarkers' and mark it as deprecated?

Zedwag: Because we want it deprecated on release according to spec

packages/devextreme/js/ui/map.d.ts
Zedwag: >Should we declare default object with all properties?

We do this in similar cases

>Moreover, mapId and useAdvancedMarkes are optional here

That doesn't mean they don't have a default values, which we want to be shown in doc

EugeniyKiyashko: I mean that each property already has its own default, but we are duplicating and listing these defaults at the object level. This could lead to maintenance issues, and if there were more than two nested properties, it would look odd. But okay, I see that this is already being done in the code, so let's leave it as is.

EugeniyKiyashko (dismissed):

ksercs (dismissed):

vladaskorohodova (dismissed):

IlyaKhD (dismissed):

EugeniyKiyashko (dismissed):

pomahtri (approved):

EugeniyKiyashko (approved):

ksercs (approved):

IlyaKhD (approved):

PR #28054, opened 2024-09-14
Fixed a bug where the unmount component template contains two childs (T1247485)

GoodDayForSurf:
T1247485 - Scheduler - "Cannot read properties of null (reading 'nextSibling')" error occurs if dateCellTemplate has more than one root element

FIX: Vue when unmounting template vnodes keeps the last element as a marker, and if we add removeListener to the end of the container, we will break the validation and an error will occur. So to fix this, we add removeListener as the first child of the container

### DIFF
--- a/packages/devextreme-vue/src/core/__tests__/button.test.ts
+++ b/packages/devextreme-vue/src/core/__tests__/button.test.ts
@@ -67,7 +67,7 @@ describe('template rendering', () => {
     expect(wrapper.vm.$el.getElementsByClassName('test')).toHaveLength(1);
   });
 
-  fit('should unmount template with two childs in root without exception', async () => {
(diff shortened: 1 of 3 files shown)


packages/devextreme-vue/src/core/__tests__/button.test.ts
```diff
@@ -66,4 +66,34 @@ describe('template rendering', () => {
 
     expect(wrapper.vm.$el.getElementsByClassName('test')).toHaveLength(1);
   });
+
+  fit('should unmount template with two childs in root without exception', async () => {
```
dxvladislavvolkov: Why do you add "fit"?

dxvladislavvolkov (approved):

VasilyStrelyaev (approved):

PR #28062, opened 2024-09-16
Chat: Finalize initials rendering

marker-dao:


packages/devextreme/js/__internal/ui/chat/avatar.ts
```diff
@@ -99,14 +99,25 @@ class Avatar extends Widget<Properties> {
...
-    if (isDefined(name)) {
-      return String(name).charAt(0).toUpperCase();
+  _getInitials(name: string | number | undefined): string {
+    if (isString(name) || isNumeric(name)) {
+      const splitValue = String(name).trim().split(' ');
+
+      const firstInitial = this._getFirstChar(splitValue[0]);
+      const secondInitial = this._getFirstChar(splitValue[1]);
+
+      const result = `${firstInitial}${secondInitial}`;
```
EugeniyKiyashko, line 105:
```suggestion
      return `${firstInitial}${secondInitial}`;
```

packages/devextreme/js/__internal/ui/chat/avatar.ts
```diff
@@ -99,14 +99,25 @@ class Avatar extends Widget<Properties> {
     return result;
   }
 
-  _getInitials(name: string | undefined): string {
-    if (isDefined(name)) {
-      return String(name).charAt(0).toUpperCase();
+  _getInitials(name: string | number | undefined): string {
+    if (isString(name) || isNumeric(name)) {
+      const splitValue = String(name).trim().split(' ');
```
EugeniyKiyashko:
```suggestion
      const splitValue = String(name).trim().split(/\s+/);
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/avatar.tests.js
```diff
@@ -40,25 +40,26 @@ QUnit.module('ChatAvatar', moduleConfig, () => {
...
             { name: 888, expectedInitials: '8' },
             { name: undefined, expectedInitials: '' },
             { name: null, expectedInitials: '' },
-            // TODO: consider scenarios
-            // { name: ' New Name', expectedInitials: 'N' },
-            // { name: NaN, expectedInitials: '' },
-            // { name: Infinity, expectedInitials: '' },
-            // { name: -Infinity, expectedInitials: '' },
-            // { name: { firstName: 'name' }, expectedInitials: '' }
+            { name: ' New Value', expectedInitials: 'NV' },
```
EugeniyKiyashko: Kindly support this scenario:
```
{ name: '   New      Value.   ', expectedInitials: 'NV' },
```

packages/devextreme/js/__internal/ui/chat/avatar.ts
```diff
@@ -99,14 +99,25 @@ class Avatar extends Widget<Properties> {
...
+
+      const firstInitial = this._getFirstChar(splitValue[0]);
+      const secondInitial = this._getFirstChar(splitValue[1]);
+
+      const result = `${firstInitial}${secondInitial}`;
+
+      return result;
     }
 
     return '';
```
EugeniyKiyashko, line 110: Did you consider variant when we return some predefined value, not empty string? For example, it can be 'UU' (Unknown User) 
image

EugeniyKiyashko (changes requested):

packages/devextreme/js/__internal/ui/chat/avatar.ts
marker-dao: I prefer to check the result before the function returns it, it's more convenient for debugging.

packages/devextreme/js/__internal/ui/chat/avatar.ts
marker-dao: The scenario you described has a broader meaning. Maybe we should set a default value for User (just like for id). Because having only a default value for initials is weird, because it might confuse users.

I asked a designer about it.

packages/devextreme/js/__internal/ui/chat/avatar.ts
marker-dao: [b5e97feb710c28caf2e6f45c42fab6585073f52d]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/avatar.tests.js
marker-dao: [b5e97feb710c28caf2e6f45c42fab6585073f52d]

packages/devextreme/js/__internal/ui/chat/avatar.ts
```diff
@@ -99,14 +99,25 @@ class Avatar extends Widget<Properties> {
     return result;
   }
 
-  _getInitials(name: string | undefined): string {
-    if (isDefined(name)) {
-      return String(name).charAt(0).toUpperCase();
+  _getInitials(name: string | number | undefined): string {
```
ksercs: Why can name be a number?

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/avatar.tests.js
```diff
@@ -40,25 +40,27 @@ QUnit.module('ChatAvatar', moduleConfig, () => {
...
+            { name: '   New      Value.   ', expectedInitials: 'NV' },
+            { name: NaN, expectedInitials: '' },
+            { name: Infinity, expectedInitials: '' },
+            { name: -Infinity, expectedInitials: '' },
+            { name: { firstName: 'name' }, expectedInitials: '' }
```
ksercs: [testing coverage]
As far as i see there is no visual test at the moment for 2-letter Avatar. Suggest u adding it

packages/devextreme/js/__internal/ui/chat/avatar.ts
marker-dao: We maintain this scenario in case the name is given by numbers.

marker-dao: During the conversation with Anton, we decided to handle all cases by casting the value to a string. If the value undefined or null, was strictly specified by the user, an empty string is returned. If the user did not specify any value, Unknown User is used by default.

Localization of all text variables is already a separate task.

packages/devextreme/js/__internal/ui/chat/avatar.ts
marker-dao: Moved to: https://github.com/DevExpress/DevExtreme/pull/28062#discussion_r1763318640

packages/devextreme/js/__internal/ui/chat/avatar.ts
marker-dao: [2c79c4f0b5cd764d261e32aae85f6b4f15127b29]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/avatar.tests.js
marker-dao: [7817343832dae488956ab0205032caab174d0ee2], [48b236474f079d7aad61cb0d15d26a3a02c04d84]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/avatar.tests.js
```diff
@@ -31,34 +31,37 @@ QUnit.module('ChatAvatar', moduleConfig, () => {
...
-            // TODO: consider scenarios
-            // { name: ' New Name', expectedInitials: 'N' },
-            // { name: NaN, expectedInitials: '' },
-            // { name: Infinity, expectedInitials: '' },
-            // { name: -Infinity, expectedInitials: '' },
-            // { name: { firstName: 'name' }, expectedInitials: '' }
+            { name: NaN, expectedInitials: 'N' },
+            { name: Infinity, expectedInitials: 'I' },
+            { name: -Infinity, expectedInitials: '-' },
+            { name: { firstName: 'name' }, expectedInitials: '[O' }
```
EugeniyKiyashko, line 64: Is it right, that you and Anton took a decision process object this way?

marker-dao: Yes, we decided that we would handle all values except null and undefined and cast them to a string, similar to how it appears in the author-name element. 

I wrote about it here: https://github.com/DevExpress/DevExtreme/pull/28062#discussion_r1763320347 and https://github.com/DevExpress/DevExtreme/pull/28062#discussion_r1763318640

Please see

EugeniyKiyashko: Sure, I saw that. I just wanted to make sure it was a conscious decision.

EugeniyKiyashko (approved):

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/avatar.tests.js
ksercs: Our logic is simple: we don't expect such values => we don't want to process it in a specific way

ksercs (approved):

PR #28073, opened 2024-09-18
Rating: create declaration and src files

Zedwag:


packages/devextreme/js/ui/rating.d.ts
```diff
@@ -0,0 +1,186 @@
...
+    ChangedOptionInfo,
+} from '../events/index';
+
+import Editor, {
+    EditorOptions,
+    ValueChangedInfo,
+} from './editor/editor';
+
+/** @public */
+export type RatingSelectionMode = 'continouus' | 'single';
```
EugeniyKiyashko: Seems, 'continouus' contains typo. Check please

packages/devextreme/js/ui/rating.d.ts
```diff
@@ -0,0 +1,186 @@
...
+    };
+    /**
+     * @docid
+     * @default 1
+     * @public
+     */
+    precision?: number;
+    /**
+     * @docid
+     * @default 'continouus'
```
EugeniyKiyashko: Typo too

packages/devextreme/js/ui/rating_types.d.ts
```diff
@@ -0,0 +1,10 @@
+export {
+  RatingSelectionMode,
+  ContentReadyEvent,
+  DisposingEvent,
+  InitializedEvent,
+  OptionChangedEvent,
+  ValueChangedEvent,
+  dxRatingOptions,
```
EugeniyKiyashko, line 8: Should we export both dxRatingOptions & Properties? I this that no. Moreover, dxRatingOptions is deprecated

Zedwag: This file is automaticaly generated. also I see that other components (chat for examplae) also includes this

packages/devextreme/js/ui/rating.d.ts
Zedwag: ty, fixed

packages/devextreme/js/ui/rating.d.ts
Zedwag: fixed

packages/devextreme/js/ui/rating_types.d.ts
EugeniyKiyashko: I am not sure if dxRatingOptions should be public and whether we should declare and deprecate it. Kindly ask the API reviewers about this, as in DateRangeBox, we only add the public type Properties.
https://github.com/EugeniyKiyashko/DevExtreme/blob/24_2_security_issue/packages/devextreme/js/ui/date_range_box.d.ts#L170-L171

packages/devextreme/js/ui/rating.d.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/** @public */
+export type Properties = dxRatingOptions;
+
+///#DEBUG
+// eslint-disable-next-line import/first
+import { CheckedEvents } from '../core';
+
+type FilterOutHidden<T> = Omit<T, 'onFocusIn' | 'onFocusOut'>;
```
EugeniyKiyashko, line 152: I don't understand why you omit the 'onFocusIn' and 'onFocusOut' properties because they are not added at the widget level or at the text editor level
image
From my perspective, widget.d.ts should contain the definitions of these properties because we are excluding them from the documentation using doc comments here:
https://github.com/Zedwag/DevExtreme/blob/ratingDeclarations__24_2/packages/devextreme/js/ui/widget/ui.widget.js#L4
Please forward this task to the API reviewers; these exceptions have been added in many places. For now, let's avoid unnecessary actions.

packages/devextreme/js/__internal/ui/m_rating.ts
```diff
@@ -0,0 +1,12 @@
+import registerComponent from '@js/core/component_registrator';
```
EugeniyKiyashko, line 1: Kindly remove prefix m_ from this file. We shouldn't ignore eslint errors for new components.

packages/devextreme/js/__internal/ui/m_rating.ts
```diff
@@ -0,0 +1,12 @@
...
+import Editor from '@js/ui/editor/editor';
+import type { Properties } from '@js/ui/rating';
+
+class Rating extends Editor<Properties> {
+
+}
+
+// @ts-expect-error
```
EugeniyKiyashko, line 9: Let's avoid using this ignore comment

packages/devextreme/js/ui/rating.js
```diff
@@ -0,0 +1,2 @@
+import Rating from '../__internal/ui/m_rating';
+export default Rating;
```
EugeniyKiyashko, line 2: Don't forget add //style comment and modify ideal dependency tree here:
https://github.com/Zedwag/DevExtreme/blob/ratingDeclarations__24_2/packages/devextreme-themebuilder/tests/data/dependencies.ts#L2

packages/devextreme/js/ui/rating.d.ts
```diff
@@ -0,0 +1,186 @@
...
+    hoverStateEnabled?: boolean;
+    /**
+     * @docid
+     * @default 5
+     * @public
+     */
+    itemCount?: number;
+    /**
+     * @docid
+     * @default {}
```
EugeniyKiyashko, line 89: Should we set default as empty object? Check it please

packages/devextreme/js/ui/rating.d.ts
```diff
@@ -0,0 +1,186 @@
...
+ */
+export type OptionChangedEvent = EventInfo<dxRating> & ChangedOptionInfo;
+
+/**
+ * @docid _ui_rating_ValueChangedEvent
+ * @public
+ * @type object
+ * @inherits NativeEventInfo,ValueChangedInfo
+ */
+export type ValueChangedEvent = NativeEventInfo<dxRating, KeyboardEvent | MouseEvent | PointerEvent | TouchEvent | Event> & ValueChangedInfo;
```
EugeniyKiyashko, line 54: I think that we shouldn't use Editor.ValueChangedInfo type because it includes value and previous values properties that have object type. In our case it should be number. Perhaps, that you copy this from Slider, seems that he has the same inaccurancy
https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxSlider/Configuration/#onValueChanged

packages/devextreme/js/ui/rating.d.ts
```diff
@@ -0,0 +1,186 @@
...
+        /**
+        * @docid
+        * @default false
+        */
+        enabled?: boolean;
+       /**
+        * @docid
+        * @default function(value, itemCount) { return `${value}/${itemCount}` }
+        */
+       format?: string | ((value: number, itemCount: number) => string);
```
EugeniyKiyashko, line 102: I think we should double-check this type, because initially, we intended to use the existing Format type. If we make this unique, our tech writers should describe it in the documentation. Please add a separate task to discuss, improve, and describe it.

packages/devextreme/js/ui/rating.d.ts
```diff
@@ -0,0 +1,186 @@
...
+     * @default 1
+     * @public
+     */
+    precision?: number;
+    /**
+     * @docid
+     * @default 'continuous'
+     * @public
+     */
+    selectionMode?: RatingSelectionMode;
```
EugeniyKiyashko, line 115: Kindly agree this type name with Vlada if you didn't do yet

packages/devextreme/js/ui/rating.d.ts
```diff
@@ -0,0 +1,186 @@
...
+    /**
+     * @docid
+     * @default 0
+     * @public
+     */
+    value?: number;
```
EugeniyKiyashko, line 121: Are you sure that option 'name' will added automatically? I see that we hide it on editor level:
https://github.com/Zedwag/DevExtreme/blob/ratingDeclarations__24_2/packages/devextreme/js/ui/editor/editor.js#L9-L10

packages/devextreme/js/ui/rating.d.ts
```diff
@@ -0,0 +1,186 @@
...
+/** @public */
+export type RatingSelectionMode = 'continuous' | 'single';
+
+/**
+ * @docid _ui_rating_ContentReadyEvent
+ * @public
+ * @type object
+ * @inherits EventInfo
+ */
+export type ContentReadyEvent = EventInfo<dxRating>;
```
ksercs, line 22: I finally think we should not add this event and appropriate property.
I don't see what it's useful for in Rating

packages/devextreme/js/ui/rating.d.ts
ksercs: Please remove this property for now because it's still questionable. U can leave only `enabled` here for the moment

packages/devextreme/js/ui/rating.d.ts
ksercs: Moreover, if we have `focus/blur` method this events are supposed to be published.

packages/devextreme/js/ui/rating.d.ts
ksercs: looks like it's missed now

PR #28078, opened 2024-09-19
🎹 Pager: Implement calculating restrictions

Raushen:


packages/devextreme/js/__internal/pager/utils/validate_utils.test.ts
```diff
@@ -0,0 +1,41 @@
+import each from 'jest-each';
+
+import { validateOptions } from './validation_utils';
+
+describe('Validate utils', () => {
+  describe('validateOption function', () => {
+    each`
+            oldPageSize | oldPageIndex | oldItemCount | expectedPageSize | expectedPageIndex | expectedItemCount | expectedPageCount
```
pomahtri, line 8: Minor (and not established as best-practice in our team yet):
Let's rewrite matrix to a lot of `it(...)` tests with only one data suite inside. In future it will help us to determine why this suite is written (because it will have description) + will allow us to quickly see how many scenarios are covered

E.g. 
```
[
  1, // not clear why we test this number
  2,
].forEach((mynumber) => {
  it(`result should be correct with ${mynumber}`, () => { // not clear what we mean by 'correct'
    ...
  }) 
})
```

->

```
it('should work with even numbers', () => { // clear description on data suite and how our func should work with it

});

it('should work with odd numbers', () => {

});
```

pomahtri (approved):

PR #28084, opened 2024-09-21
Chat: polish scrollable integration

EugeniyKiyashko:


packages/devextreme-scss/scss/widgets/base/chat/layout/chat-messagelist/_index.scss
```diff
@@ -9,6 +9,14 @@
     flex-direction: column;
     width: 100%;
   }
+
+  .dx-scrollable-native {
+    &.dx-scrollable-native-ios{
```
marker-dao, line 14:
```suggestion
    &.dx-scrollable-native-ios {
```

marker-dao (approved):

packages/devextreme-scss/scss/widgets/base/chat/layout/chat-messagelist/_index.scss
EugeniyKiyashko: If there are no other remarks, I'll fix this typo in the next PR to avoid rerunning the tests.

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -90,6 +91,8 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
         const shouldButtonBeDisabled = !this._isValuableTextEntered();
 
         this._toggleButtonDisableState(shouldButtonBeDisabled);
+
+        triggerResizeEvent(this.$element().parent());
```
ksercs, line 95: What it's necessary for?

packages/devextreme-scss/scss/widgets/base/chat/layout/chat-messagelist/_index.scss
```diff
@@ -9,6 +9,14 @@
...
+      .dx-scrollable-content {
+        min-height: 100%;
```
ksercs, line 16: Let's create a tech debt card for reconsidering this style on scrollable's level. I don't like we need override it in components where we use scrollable

packages/devextreme/js/__internal/ui/chat/messagebox.ts
EugeniyKiyashko: When expanding the message box area, we need to update the scrollbar size when using the simulated strategy. Currently, the standard mechanism with the 'dxresize' trigger to call Chat's _dimensionChanged to notify Scrollable that it should be updated.

packages/devextreme-scss/scss/widgets/base/chat/layout/chat-messagelist/_index.scss
EugeniyKiyashko: I unarchived the card in our Backlog and added this case there. It's attached to our discussion thread in Teams.

packages/devextreme/js/__internal/ui/chat/messagebox.ts
ksercs: Hm, shouldn't be there a test for this scenario?

And maybe we can make this solution more direct through direct call of a `messageList.update()`? I really don't like we trigger additional resize events for the whole chat

EugeniyKiyashko: ` I really don't like we trigger additional resize events for the whole chat `

Yes, I understand, but it looks better to add a resize observer to the message list container. 

`Hm, shouldn't be there a test for this scenario?`

Sure, it should. I would like to approve the solution and then add the test afterward. Please allow me to do this in a separate PR, if it doesn't require to add other changes in PR.

ksercs: >  but it looks better to add a resize observer to the message list container.

Sounds good to me

ksercs: Well, let's add a test and rework the solution using resizeObserver in a neighbor PR

ksercs (approved):

PR #28086, opened 2024-09-23
Chat: Add axe tests

marker-dao:


e2e/testcafe-devextreme/tests/accessibility/chat.ts
```diff
@@ -0,0 +1,69 @@
...
+      timestamp: new Date(),
+      text: 'Message text',
+      author: userSecond,
+    },
+  ];
+
+  return items;
+};
+
+const items = [[], getItems(true), getItems(false)];
```
ksercs: Now it's difficult enough for me to compile in a head what will be generated in items. Can we simplify this functional generators?
Particularly, I don't see a reason why we need generate users. Let's just create to constant object `currentUser` and `anotherUser` (with avatarUrl)? It'll be much more clearer for me

ksercs (approved):

EugeniyKiyashko (approved):

PR #28106, opened 2024-09-24
Calendar: Improve aria-label in Multiple mode

marker-dao:


vladaskorohodova (dismissed):

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
```diff
@@ -138,11 +138,46 @@ const BaseView = (Widget as any).inherit({
...
   },
 
+  _getMultipleRangesText() {
+    const { ranges } = this.option();
+
+    if (ranges.length > 2) {
+      // @ts-expect-error
+      const dateRangeCountText = messageLocalization.format('dxCalendar-selectedDateRangeCount', ranges.length);
+
+      return `${dateRangeCountText}`;
```
ksercs:
```suggestion
      return dateRangeCountText;
```

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
```diff
@@ -138,11 +138,46 @@ const BaseView = (Widget as any).inherit({
...
+    }
+
+    const selectedDatesText = messageLocalization.format('dxCalendar-selectedDates');
+    const rangesText = ranges.map((range) => this._getMultipleRangeText(range)).join(', ');
```
ksercs:
```suggestion
    const rangesText = ranges
        .map((range) => this._getMultipleRangeText(range))
        .join(', ');
```

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
```diff
@@ -138,11 +138,46 @@ const BaseView = (Widget as any).inherit({
...
+
+    const result = `${selectedDatesText}: ${rangesText}`;
+
+    return result;
+  },
+
+  _getMultipleRangeText(range) {
```
ksercs:
```suggestion
  _getRangeText(range) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/calendar.markup.tests.js
```diff
@@ -461,18 +465,41 @@ QUnit.module('Aria accessibility', {
...
+                {
+                    value: [ '2017-03-27T16:54:48' ],
+                    expectedAriaLabel: `Calendar. The selected dates: ${getFormattedDate('2017-03-27T16:54:48')}`,
+                },
+                {
+                    value: [ 1726938544957 ],
+                    expectedAriaLabel: `Calendar. The selected dates: ${getFormattedDate(1726938544957)}`,
+                },
+                {
+                    value: [ 1726938544957, 1726852144957, 1726765744957 ],
```
ksercs: [refactoring]
Let's replace numbers with usual dates? Maybe event string in a default format like "2024-01-01".
Just to make tests clearer, now it's very difficult to read

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
```diff
@@ -7,9 +8,12 @@ class CalendarMultiSelectionStrategy extends CalendarSelectionStrategy {
   }
 
   getViewOptions() {
+    const value = this.dateOption('value');
+
     return {
-      value: this.dateOption('value'),
+      value,
       range: [],
+      ranges: this._generateRanges(value),
```
ksercs: [code splitting]
As far as I see u generate `range` from `value`, and this new `ranges` property is necessary only for a label generation.
So maybe it's better to no add a new option, and generate ranges directly in the view's `_getMultipleModeAriaLabel`? 

And the code for ranges generation is definitely should be not in a component code (because it's a pure function not using component state). It's better to move it to the `utils/date` and tested separately as an util.

Then u don't need to sync about `ranges` update on value change, about its initial generation + generation code not using component's state is isolated and well tested

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
```diff
@@ -52,6 +56,60 @@ class CalendarMultiSelectionStrategy extends CalendarSelectionStrategy {
...
+        return;
+      }
+
+      const previousDate = sortedDates[index - 1];
+      const isNewRange = date - previousDate > DAY_INTERVAL;
+
+      if (isNewRange) {
+        const range = getRange(startDate, sortedDates, index);
+
+        // @ts-expect-error
```
ksercs: [ts]
This code is completely new => there should not be ts errors ignore.
Now u have this error because u didn't add a type for `ranges` when declared it, and type is generated like `never`.
It'll be ok if you declare it like `const ranges: Date[][] = [];`

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
```diff
@@ -52,6 +56,60 @@ class CalendarMultiSelectionStrategy extends CalendarSelectionStrategy {
...
+    };
+
+    const ranges = [];
+
+    let startDate = sortedDates[0];
+
+    sortedDates.forEach((date, index) => {
+      if (index === 0) {
+        return;
+      }
```
ksercs: [refactoring]
U can rewrite it like `for (let i = 1; i < sortedDates.length; ++i)`, and then u don't need this if condition.
And possibly don't need creating one more range after the loop if u do `i <= sortedDates.length`

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
```diff
@@ -52,6 +56,60 @@ class CalendarMultiSelectionStrategy extends CalendarSelectionStrategy {
 
     this.dateValue(selectedDates, event);
   }
+
+  _generateRanges(values) {
+    const datesInMilliseconds = values.map((value) => {
+      const date = new Date(new Date(value).setHours(0, 0, 0, 0)).getTime();
+
+      return date;
+    });
```
ksercs: Hm, why do we need this? Multiple mode does not support datetime as far as i remember.
Math operations works well with `Date`

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
```diff
@@ -52,6 +56,60 @@ class CalendarMultiSelectionStrategy extends CalendarSelectionStrategy {
...
+
+    let startDate = sortedDates[0];
+
+    sortedDates.forEach((date, index) => {
+      if (index === 0) {
+        return;
+      }
+
+      const previousDate = sortedDates[index - 1];
+      const isNewRange = date - previousDate > DAY_INTERVAL;
```
ksercs: [copypaste]
If I'm right and we don't support datetime then all dates here are different.
But if we really need it I believe it's better to use `sameDate` function from `utils/date`

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/calendar.markup.tests.js
marker-dao: Look at this test from the other side: I am testing the value formats that are specified in the documentation: https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxCalendar/Configuration/#value. I propose the following solution: I will simply reduce the number of tests that use milliseconds.

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
marker-dao: We have not only Date value format so I used `new Date(value)`: https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxCalendar/Configuration/#value 

Also I used `.setHours(0, 0, 0, 0)` to get the start of the day, since the might set the midday variable (in options on init) and the isNewRange constant won't work correctly:

`const isNewRange = currentDate - previousDate > msInDay;`

marker-dao: There is example: 

'2024-09-26T18:28:10' - '2024-09-25T00:28:10' = more than 1 day 

Screenshot 2024-09-26 at 18 39 57

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
marker-dao: I don't need sameDate functionality. I need to find the beginning of a new range) Moved discussion to: https://github.com/DevExpress/DevExtreme/pull/28106#discussion_r1775468176

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
marker-dao: [251e725c842b56d976f805235a429e80710a70c5]

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
marker-dao: [251e725c842b56d976f805235a429e80710a70c5]

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
marker-dao: [251e725c842b56d976f805235a429e80710a70c5]

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/calendar.markup.tests.js
marker-dao: [251e725c842b56d976f805235a429e80710a70c5]

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
marker-dao: I moved code into utils: [251e725c842b56d976f805235a429e80710a70c5]

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
marker-dao: [251e725c842b56d976f805235a429e80710a70c5]

packages/devextreme/js/core/utils/date.js
```diff
@@ -689,6 +689,45 @@ const getMachineTimezoneName = () => {
         : null;
 };
 
+const getRangesByDates = (dates) => {
+    const datesInMilliseconds = dates.map((value) => {
```
EugeniyKiyashko:
```suggestion
    const datesInMilliseconds = dates.map((value) => correctDateWithUnitBeginning(value, 'day').getTime());
```

packages/devextreme/js/core/utils/date.js
```diff
@@ -689,6 +689,45 @@ const getMachineTimezoneName = () => {
         : null;
 };
 
+const getRangesByDates = (dates) => {
```
EugeniyKiyashko, line 693: We calculate the same value for area-label three times: once for the visible view and twice for the invisible ones. Is there any way to optimize this and calculate the value only once?

EugeniyKiyashko (changes requested):

packages/devextreme/js/core/utils/date.js
```diff
@@ -689,6 +689,45 @@ const getMachineTimezoneName = () => {
...
+    });
+
+    const sortedDates = datesInMilliseconds.sort((a, b) => a - b);
+
+    const getRange = (date, dates, index) => {
+        const range = date === dates[index - 1]
+            ? [date]
+            : [date, dates[index - 1]];
+
+        return range.map((value) => new Date(value));
```
EugeniyKiyashko: It’s safer to use the [dateSerialization.deserializeDate(value)](https://github.com/marker-dao/DevExtreme/blob/24_2_chat_messagesend_timeout/packages/devextreme/js/__internal/ui/calendar/m_calendar.ts#L255-L256) function to convert many possible formats of the incoming value to a Date format

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
```diff
@@ -41,7 +41,7 @@ const CALENDAR_DXHOVERSTART_EVENT_NAME = addNamespace(hoverStartEventName, 'dxCa
 
 const CALENDAR_DATE_VALUE_KEY = 'dxDateValueKey';
 
-const DAY_INTERVAL = 86400000;
+export const DAY_INTERVAL = 86400000;
```
ksercs: [excess change]
Now we already don't need exporting this

packages/devextreme/testing/tests/DevExpress.utils/utils.date.tests.js
```diff
@@ -0,0 +1,38 @@
...
+                new Date('2024-10-27T00:00:00'),
+            ],
+            [
+                new Date('2024-11-27T00:00:00'),
+            ]
+        ];
+
+        assert.deepEqual(result, expectedResult);
+    });
+});
```
ksercs, line 76: [test coverage]
1. `[]` (empty array)
2. `['2024-11-27']` (1-length array)
3. `['2024-11-27', '2024-11-28' , '2024-11-29' ]` (single range in array)

packages/devextreme/js/core/utils/date.js
ksercs: I don't think it's important, anyway we have O(n) time complexity

packages/devextreme/js/core/utils/date.js
ksercs: I cannot explain from top of my head why it's safer but yeah, probably we should use this solution at least for a consistency

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
ksercs: Thanks, now it's clearer to me

packages/devextreme/js/core/utils/date.js
```diff
@@ -689,6 +689,45 @@ const getMachineTimezoneName = () => {
...
+
+    let startDate = sortedDates[0];
+
+    for(let i = 1; i <= sortedDates.length; ++i) {
+        const currentDate = sortedDates[i];
+        const previousDate = sortedDates[i - 1];
+        const isNewRange = currentDate - previousDate > msInDay;
+
+        if(isNewRange || i === sortedDates.length) {
+            const range = getRange(startDate, sortedDates, i);
```
ksercs: Don't u think that in this particular case code will be more readable if we get rid of the `getRange` function? Smth like
```javascript
const range = previousDate === startDate
    ? [startDate]
    : [startDate, previousDate];
    
ranges.push(range.map(/*...*/));
```
Just an additional idea: maybe it's more clear to change naming `presiousDate -> currentDate`, `currentDate -> nextDate` for making `if` looking like `if (currentDate === startDate)`

packages/devextreme/js/core/utils/date.js
marker-dao: We can, but for that I would introduce a new ranges option at the calendar level.

packages/devextreme/js/core/utils/date.js
marker-dao: [e6875c297c1d4e3c571476fa8c008d6410276286]

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
marker-dao: [e6875c297c1d4e3c571476fa8c008d6410276286]

packages/devextreme/testing/tests/DevExpress.utils/utils.date.tests.js
marker-dao: [e6875c297c1d4e3c571476fa8c008d6410276286]

packages/devextreme/js/core/utils/date.js
marker-dao: [10a380daeca82004dfd116401bdefa3c80a5813c]

packages/devextreme/testing/tests/DevExpress.utils/utils.date.tests.js
```diff
@@ -0,0 +1,76 @@
+import dateUtils from 'core/utils/date';
+const { test, module } = QUnit;
+
+module('getRangesByDates', {}, () => {
+    test('getRangesByDates should generate correctly ranges when dates is empty array', function(assert) {
```
ksercs: [minor]
In each test
```suggestion
    test('getRangesByDates should generate correct ranges when dates is empty array', function(assert) {
```

ksercs (dismissed):

packages/devextreme/testing/tests/DevExpress.utils/utils.date.tests.js
marker-dao: [10a380daeca82004dfd116401bdefa3c80a5813c]

ksercs (dismissed):

ksercs (approved):

EugeniyKiyashko (approved):

vladaskorohodova (approved):

PR #28107, opened 2024-09-25
DataGrid - Column Header filter does not properly work if the column caption contains double quotes (T1251768)

tongsonbarbs:


packages/devextreme/js/ui/shared/accessibility.js
```diff
@@ -57,7 +57,7 @@ function processKeyDown(viewName, instance, event, action, $mainElement, execute
 
 export function saveFocusedElementInfo(target, instance) {
     const $target = $(target);
-    const ariaLabel = $target.attr('aria-label');
+    const ariaLabel = $target.attr('aria-label')?.replace(/"/g, '\\"');
```
Raushen: It is needed to move the escaping to the getActiveAccessibleElements method level. Because you repaired only one case. There is [the second usage here](https://github.com/DevExpress/DevExtreme/blob/396fa5aa36e35ce0eaf7633a5fef1e6b2b6ccdd5/packages/devextreme/js/ui/shared/accessibility.js#L161).

Raushen (changes requested):

packages/devextreme/js/ui/shared/accessibility.js
```diff
@@ -158,7 +159,8 @@ export function restoreFocus(instance) {
     if(!instance.option('useLegacyKeyboardNavigation') && focusedElementInfo) {
         const viewInstance = focusedElementInfo.viewInstance;
         if(viewInstance) {
-            const $activeElements = getActiveAccessibleElements(focusedElementInfo.ariaLabel, viewInstance.element());
+            const escapedAriaLabel = focusedElementInfo.ariaLabel?.replace(/"/g, '\\"');
```
Raushen: Now you do not need this check as it is exists inside the `getActiveAccessibleElements` method

packages/devextreme/js/ui/shared/accessibility.js
tongsonbarbs: If I add the escape character in **getActiveAccessibleElements** method, it will not fix the **ariaLabel** variable inside **saveFocusedElementInfo** method, because ariaLabel variable is being used in [**focusedElementInfo**.](https://github.com/DevExpress/DevExtreme/pull/28107/files/057905cae5d6ee2b1ee86d71c5c21324c61210f7#diff-517e37cd44e0fb7d7211dc2a3f1732cc7e4585e6257898b60b9e57cb9e1a6cc8R64)

tongsonbarbs: I will just add another escape character for the variable I mentioned..

Raushen (approved):

PR #28110, opened 2024-09-25
🎹 Pager: Testcafe tests

Raushen:


e2e/testcafe-devextreme/tests/pager/index.ts
```diff
@@ -0,0 +1,60 @@
+import Pager from 'devextreme-testcafe-models/pager';
+import url from '../../helpers/getPageUrl';
+import { createWidget } from '../../helpers/createWidget';
+
+fixture`Pager Base Properties`
```
Alyar666, line 5: Do we need to enable the disablePageReloads option?

Raushen: I noticed some unstable behavior reproducing locally, so I decided not to use disablePageReloads because there are not a lot of tests

Alyar666 (approved):

PR #28111, opened 2024-09-25
Chat: use resizeObserver to update scrollbar size and position

EugeniyKiyashko:


packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -114,7 +167,9 @@ class MessageList extends Widget<Properties> {
       .appendTo(this.$element());
 
     this._scrollable = this._createComponent($scrollable, Scrollable, {
+      // useNative: false,
       useKeyboard: false,
+      // showScrollbar: 'always',
```
marker-dao: Redundant comments?

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -238,13 +300,9 @@ class MessageList extends Widget<Properties> {
...
-  update(): void {
-    // eslint-disable-next-line @typescript-eslint/no-floating-promises
-    this._scrollable.update();
-
-    this._scrollContentToLastMessage();
-  }
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// (MessageList as any).include(DataHelperMixin);
```
marker-dao: Redundant comments?

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.tests.js
```diff
@@ -379,41 +387,56 @@ QUnit.module('MessageList', moduleConfig, () => {
...
+            const done = assert.async();
+
             this.reinit({
                 width: 300,
                 height: 600,
                 items: generateMessages(52)
             });
 
-            const scrollTop = this.scrollable.scrollTop();
+            setTimeout(() => {
```
marker-dao, line 399: Is it really necessary to use setTimeout here? Doesn't component have options like onInit (or onContentReady XD), it doesn't fit?

packages/devextreme/js/__internal/ui/chat/messagelist.ts
EugeniyKiyashko: Thanks. Yep, this part from other task

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.tests.js
EugeniyKiyashko: Yes, it's really needed. Without it, we can't handle the resizeObserver call properly

packages/devextreme/js/__internal/ui/chat/messagelist.ts
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/28111/commits/4b842e1b19494eb9baf6a8a204fa91b23a683b14

packages/devextreme/js/__internal/ui/chat/messagelist.ts
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/28111/commits/4b842e1b19494eb9baf6a8a204fa91b23a683b14

Zedwag (approved):

marker-dao (approved):

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -50,7 +63,47 @@ class MessageList extends Widget<Properties> {
 
     this._renderMessageListContent();
 
-    this.update();
+    this._attachResizeObserverSubscription();
```
ksercs, line 66: [SSR logic]
I believe it should be done in `_renderContent`, not `_initMarkup`.
Then u don't need `if (hasWindow()) {` inside of this method

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -50,7 +63,47 @@ class MessageList extends Widget<Properties> {
...
+    if (hasWindow()) {
+      const element = this._getScrollContainer();
+
+      resizeObserverSingleton.unobserve(element);
+      resizeObserverSingleton.observe(element, (entry) => this._resizeHandler(entry));
+    }
+  }
+
+  _isAttached(element: Element): boolean {
+    return !!contains(domAdapter.getBody(), element);
```
ksercs, line 81: [copypaste, shadowDOM]
I think we should use `isElementInDom` from `utils/dom` instead of creating this new method.
Moreover, now the solution does now work in shadowDOM as far as I see

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -50,7 +63,47 @@ class MessageList extends Widget<Properties> {
...
+  }
+
+  _resizeHandler({ contentRect, target }: ResizeObserverEntry): void {
+    const newHeight = contentRect.height;
+
+    if (this._suppressResizeHandling
+      && this._isAttached(target)
+      && isElementVisible(target as HTMLElement)
```
ksercs, line 89: [refactoring]
Can we isolate these conditions? Let it be a first if and return if element is invisible or detached.
Now i don't like we do smth actions in this case + update cached height

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -50,7 +63,47 @@ class MessageList extends Widget<Properties> {
...
+  }
+
+  _isAttached(element: Element): boolean {
+    return !!contains(domAdapter.getBody(), element);
+  }
+
+  _resizeHandler({ contentRect, target }: ResizeObserverEntry): void {
+    const newHeight = contentRect.height;
+
+    if (this._suppressResizeHandling
```
ksercs, line 87: [naming, ?excess cache?]
u name it `_suppressResizeHandling`, but i see u don't suppress, u scroll to last message.
So, what is this for, just for indicating it's a first call, right?

Then maybe we can get rid of this private cache and just do it this way?
```javascript
const isAfterFirstRendering = this._containerClientHeight === 0; // or even undefined/null?
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.tests.js
```diff
@@ -422,12 +445,79 @@ QUnit.module('MessageList', moduleConfig, () => {
...
+                    const scrollTop = this.getScrollable().scrollTop();
+
+                    assert.notEqual(scrollTop, 0);
+                    assert.roughEqual(scrollTop, this.getScrollOffsetMax(), 1);
+                    done();
+                });
+            });
+        });
+
+        QUnit.test('should be scrolled to the last message after being rendered inside an invisible element and display correctly when shown', function(assert) {
```
ksercs, line 460:
```suggestion
        QUnit.test('should be scrolled to the last message after showing if was initially rendered inside an invisible element', function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.tests.js
```diff
@@ -379,41 +387,56 @@ QUnit.module('MessageList', moduleConfig, () => {
...
                 items: generateMessages(52)
             });
 
-            const scrollTop = this.scrollable.scrollTop();
+            setTimeout(() => {
+                const scrollTop = this.scrollable.scrollTop();
 
-            assert.strictEqual(scrollTop !== 0, true);
+                assert.notEqual(scrollTop, 0);
+                assert.roughEqual(scrollTop, this.getScrollOffsetMax(), 1);
```
ksercs, line 403: [minor]
please add assertion comments. It's not very informative now
image

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -162,23 +215,30 @@ class MessageList extends Widget<Properties> {
...
   }
 
   _scrollContentToLastMessage(): void {
-    this._scrollable.scrollTo({ top: this._$content().get(0).scrollHeight });
+    const scrollOffsetTopMax = getScrollTopMax(this._getScrollContainer());
+
+    this._scrollable.scrollTo({ top: scrollOffsetTopMax });
+  }
+
+  _getScrollContainer(): HTMLElement {
```
ksercs, line 239: [refactoring]
do we really need it? Maybe subscribe just on root?

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -50,7 +63,47 @@ class MessageList extends Widget<Properties> {
...
+
+    if (this._suppressResizeHandling
+      && this._isAttached(target)
+      && isElementVisible(target as HTMLElement)
+    ) {
+      this._scrollContentToLastMessage();
+
+      this._suppressResizeHandling = false;
+    } else {
+      const heightChange = this._containerClientHeight - newHeight;
```
ksercs, line 95: Could u please add a Unit test for this scenario? Smth like "scrollable should keep scrolling position after container resize if scrolling position is not bottom"
It was very difficult for me now to understand without unit tests what it's written for

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -50,7 +63,47 @@ class MessageList extends Widget<Properties> {
...
+
+      let { scrollTop } = target;
+
+      if (heightChange >= 1 || !isReachedBottom(target as HTMLDivElement, target.scrollTop, 0, 1)) {
```
ksercs, line 99: [excess code?]
I don't understand why we need `heightChange`.
U're trying to say "we need some scroll if we are now on bottom and height is increased" but i see no difference with and without this code

ksercs, line 99:
```suggestion
      const isReachedBottom = isReachedBottom(target as HTMLDivElement, target.scrollTop, 0, 1);
      if (!isReachedBottom) {
```

packages/devextreme/js/__internal/ui/chat/messagelist.ts
EugeniyKiyashko: I can simplified it this way:

```
if (!isElementInDom(target) || !isElementVisible(target as HTMLElement)) {
      return;
    }

    if (!isDefined(this._containerClientHeight)) {
      this._scrollContentToLastMessage();
    }
 ```

PR #28114, opened 2024-09-26
Chat: data layer integration

EugeniyKiyashko:


packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.tests.js
```diff
@@ -395,6 +407,239 @@ QUnit.module('Chat', moduleConfig, () => {
...
+            const deferred = $.Deferred();
+            const messages = [{ text: 'message_1' }, { text: 'message_2' }];
+            this.reinit({
+                dataSource: messages
+            });
+
+            this.instance.option({
+                dataSource: {
+                    load() {
+                        return deferred.promise();
```
ksercs, line 550: Is it really rejected?

EugeniyKiyashko: If we don't call 'deferred.reject()' test will fail

ksercs (dismissed):

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -183,14 +207,26 @@ class Chat extends Widget<Properties> {
...
+      const { items } = this.option();
 
-    this.option('items', newItems);
+      const newItems = [...items ?? [], message];
+      this.option('items', newItems);
+      return;
+    }
+
+    dataSource.store().insert(message).done(() => {
+      dataSource.load();
```
ksercs: Is it required? Seems like it means that we'll send addition request after single message rendering, so it would be great if we can omit it

EugeniyKiyashko: We discussed this part separately. The details can be found in the corresponding team's post.

ksercs (approved):

Zedwag (approved):

PR #28118, opened 2024-09-27
Chat: Support Accessibility

marker-dao:


packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -154,6 +150,18 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
     this._button.option('disabled', state);
   }
 
+  _updateAria(value: string | null): void {
```
ksercs: I'd like to explain what we reference to here
```suggestion
  _updateAria(emptyViewId: string | null): void {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.markup.tests.js
```diff
@@ -87,4 +87,39 @@ QUnit.module('MessageList', moduleConfig, () => {
             assert.strictEqual($emptyView.find(`.${CHAT_MESSAGELIST_EMPTY_PROMPT_CLASS}`).length, 1);
         });
     });
+
+    QUnit.module('Accessibility', () => {
+        QUnit.test('container should be have id attribute', function(assert) {
+            const id = this.$element.find(`.${CHAT_MESSAGELIST_EMPTY_VIEW_CLASS}`).attr('id');
+
+            assert.strictEqual(id !== undefined, true);
```
ksercs:
```suggestion
            assert.notStrictEqual(id, undefined);
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.markup.tests.js
```diff
@@ -87,4 +87,39 @@ QUnit.module('MessageList', moduleConfig, () => {
             assert.strictEqual($emptyView.find(`.${CHAT_MESSAGELIST_EMPTY_PROMPT_CLASS}`).length, 1);
         });
     });
+
+    QUnit.module('Accessibility', () => {
+        QUnit.test('container should be have id attribute', function(assert) {
```
ksercs:
```suggestion
        QUnit.test('emptyView should have id attribute', function(assert) {
```

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: [ef44cc43f5326a35f5bd9b69d6ee254a85950752]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.markup.tests.js
marker-dao: [ef44cc43f5326a35f5bd9b69d6ee254a85950752]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.markup.tests.js
marker-dao: [ef44cc43f5326a35f5bd9b69d6ee254a85950752]

ksercs (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.markup.tests.js
```diff
@@ -47,4 +47,12 @@ QUnit.module('MessageBox', moduleConfig, () => {
             assert.strictEqual($button.hasClass(CHAT_MESSAGEBOX_BUTTON_CLASS), true);
         });
     });
+
+    QUnit.module('Accessibility', moduleConfig, () => {
+        QUnit.test('send button should have correct aria-label', function(assert) {
+            const $button = this.$element.find(`.${BUTTON_CLASS}`);
+
+            assert.strictEqual($button.attr('aria-label'), 'Send');
```
EugeniyKiyashko, line 55: Should we provide opportunity to localize 'Send' value?

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -311,6 +306,24 @@ class MessageList extends Widget<Properties> {
     return result;
   }
 
+  _updateAria(): void {
+    const aria = {
+      role: 'log',
+      atomic: 'false',
+      label: 'Message list',
```
EugeniyKiyashko, line 305: Should we provide opportunity to localize these attributes?

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -109,6 +112,22 @@ class Chat extends Widget<Properties> {
...
+      role: 'group',
+      label: 'Chat',
+    };
+
+    this.setAria(aria, this.$element());
+  }
+
+  _updateMessageBoxAria(): void {
+    const $emptyView = this._messageList.$element().find(`.${CHAT_MESSAGELIST_EMPTY_VIEW_CLASS}`);
+    const emptyViewId = $emptyView?.attr('id') ?? null;
```
EugeniyKiyashko: Can we implement a public method inside the messageList component that will return the emptyViewId value if the empty view is rendered, and null in other cases? Seems, in this case we can avoid look for emptyview element on Chat's level. Something like this:
```
this._messageList.getEmptyViewId(): id | null {
    return this. _isEmpty() ? $(this.element()).find(`.${CHAT_MESSAGELIST_EMPTY_VIEW_CLASS}`).attr('id') ?? null;
}
```

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -154,6 +150,18 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
     this._button.option('disabled', state);
   }
 
+  _updateAria(emptyViewId: string | null): void {
```
EugeniyKiyashko: Because we use this method outside of messagebox, its name shouldn't include underscore prefics

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -154,6 +150,18 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
...
+    $(this._textArea.$element())
```
EugeniyKiyashko: I understand that you are avoiding the // @ts-expect-error comment here, but it’s better to use the _input method instead, as we use this method widely. It’s not a big issue if the comment is present; when we move the TextEditor to a native class, the comment will be removed.

https://github.com/marker-dao/DevExtreme/blob/24_2_chat_accessibility/packages/devextreme/js/__internal/ui/text_box/m_text_editor.base.ts#L153

EugeniyKiyashko: Moreover, we can use public option inputAttr here: https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxTextArea/Configuration/#inputAttr

packages/devextreme/js/__internal/ui/chat/messagebox.ts
EugeniyKiyashko: Maybe we can specify a concrete method name since we are only updating aria-labelledby here?

EugeniyKiyashko (approved):

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.markup.tests.js
```diff
@@ -87,5 +89,71 @@ QUnit.module('Chat', moduleConfig, () => {
...
+                    { text: '2' },
+                ],
+            });
+
+            const $textArea = this.$element.find(`.${TEXTEDITOR_INPUT_CLASS}`);
+
+            assert.strictEqual($textArea.attr('aria-labelledby'), undefined);
+        });
+
+        QUnit.test('textarea should get aria-labelledby attribute if items has been removed in runtime', function(assert) {
```
EugeniyKiyashko:
```suggestion
        QUnit.test('textarea should get aria-labelledby attribute if items have been removed at runtime', function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.markup.tests.js
```diff
@@ -87,5 +89,71 @@ QUnit.module('Chat', moduleConfig, () => {
...
+        });
+
+        QUnit.test('textarea should have correct aria-labelledby attribute if there are not items', function(assert) {
+            const $textArea = this.$element.find(`.${TEXTEDITOR_INPUT_CLASS}`);
+            const $emptyView = this.$element.find(`.${CHAT_MESSAGELIST_EMPTY_VIEW_CLASS}`);
+
+            assert.strictEqual($textArea.attr('aria-labelledby'), $emptyView.attr('id'));
+        });
+
+        QUnit.test('textarea should get rid of aria-labelledby attribute if items has been added in runtime', function(assert) {
```
EugeniyKiyashko:
```suggestion
        QUnit.test('textarea should get rid of aria-labelledby attribute if items have been removed at runtime', function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.markup.tests.js
```diff
@@ -87,5 +89,71 @@ QUnit.module('Chat', moduleConfig, () => {
...
+                    { text: '2' },
+                ],
+            });
+
+            const $textArea = this.$element.find(`.${TEXTEDITOR_INPUT_CLASS}`);
+
+            assert.strictEqual($textArea.attr('aria-labelledby'), undefined);
+        });
+
+        QUnit.test('textarea should have correct aria-labelledby attribute if there are not items', function(assert) {
```
EugeniyKiyashko:
```suggestion
        QUnit.test('textarea should have correct aria-labelledby attribute if there are no items', function(assert) {
```

EugeniyKiyashko (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.markup.tests.js
marker-dao: That's a separate task, I've already accomplished it in a separate PR.

packages/devextreme/js/__internal/ui/chat/messagelist.ts
marker-dao: That's a separate task, I've already accomplished it in a separate PR.

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: [ddd98c2f8675ba0e9ea72a8e3cde9824787a94a4]

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: [ddd98c2f8675ba0e9ea72a8e3cde9824787a94a4]

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: [ddd98c2f8675ba0e9ea72a8e3cde9824787a94a4]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.markup.tests.js
marker-dao: [ddd98c2f8675ba0e9ea72a8e3cde9824787a94a4]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.markup.tests.js
marker-dao: [ddd98c2f8675ba0e9ea72a8e3cde9824787a94a4]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.markup.tests.js
marker-dao: [ddd98c2f8675ba0e9ea72a8e3cde9824787a94a4]

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -325,6 +338,13 @@ class MessageList extends Widget<Properties> {
         super._optionChanged(args);
     }
   }
+
+  getEmptyViewId(): string | null {
+    const $emptyView = this.$element().find(`.${CHAT_MESSAGELIST_EMPTY_VIEW_CLASS}`);
```
EugeniyKiyashko: Why didn't we check the data here and use a search by the DOM tree? Something like this

```
return this._isEmpty() ? $emptyView.attr('id') : null;
```
_isEmpty() is reliable indicator that the empty view element is currently present in the DOM tree. 
 Another reason to use this approach is performance. If a lot of messages are rendered in the chat, we would need to spend time processing each element.
If you encounter any difficulties with this approach, feel free to let me know

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -173,6 +174,14 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
         super._optionChanged(args);
     }
   }
+
+  updateTextAreaAria(emptyViewId: string | null): void {
```
EugeniyKiyashko:
```suggestion
  updateInputAria(aria: Record<string, unknown>): void {
      this._textArea.option('inputAttr', aria);
  }
```

EugeniyKiyashko: Or we can do it more common and name it `updateInputAttr(atts)`. It could work with any attribute, not just ARIA-specific ones.

packages/devextreme/js/__internal/ui/chat/messagelist.ts
marker-dao: Okay, I will refactor it, but don't we then want to get rid of .find() in all places in our code to avoid Depth-First Search? 

This is the reason why I suggested earlier to keep references to objects such as emptyView to avoid using find().

Your approach is clear to me, I'm just trying to write more concise code. But $emptyView.attr('id') can return undefined, so we need to add some code to catch this case. Check the code, pls, I added _isEmpty().

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: Let's leave the naming associated with Aria.

marker-dao: It seems to me that it is ideologically more correct to form the aria object inside the messagebox.

packages/devextreme/js/__internal/ui/chat/messagelist.ts
marker-dao: [604f5dc44f8c0d730eca6fcecfa778976849c4d0]

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: [604f5dc44f8c0d730eca6fcecfa778976849c4d0]

EugeniyKiyashko (approved):

ksercs (approved):

PR #28127, opened 2024-10-01
Map - deprecate Bing provider

Zedwag:


packages/devextreme/js/__internal/ui/map/m_map.ts
```diff
@@ -86,6 +86,23 @@ const Map = Widget.inherit({
     ]);
   },
 
+  ctor(element, options) {
```
ksercs, line 89: [missed scenario?]
And what about switching provider at runtime?

Zedwag: updated

packages/devextreme/js/__internal/ui/map/m_map.ts
```diff
@@ -208,6 +225,9 @@ const Map = Widget.inherit({
       case 'provider':
         this._suppressAsyncAction = true;
         this._invalidate();
+        if (value === 'bing') {
+          this._logDeprecatedBingProvider();
```
ksercs, line 229: Have u checked that there is no double log now?

ksercs (approved):

marker-dao (approved):

PR #28129, opened 2024-10-01
WIP: Use true imported union / enum types instead of explicit (string) acceptable values React / Vue

mpreyskurantov:
ReOpened at #28291


packages/devextreme-vue/src/button.ts
```diff
@@ -40,25 +52,25 @@ const componentConfig = {
...
-    onInitialized: Function,
-    onOptionChanged: Function,
+    onClick: Function as PropType<(e: ClickEvent) => void>,
+    onContentReady: Function as PropType<(e: ContentReadyEvent) => void>,
+    onDisposing: Function as PropType<(e: DisposingEvent) => void>,
+    onInitialized: Function as PropType<(e: InitializedEvent) => void>,
+    onOptionChanged: Function as PropType<(e: OptionChangedEvent) => void>,
     rtlEnabled: Boolean,
-    stylingMode: String,
+    stylingMode: Object as PropType<ButtonStyle>,
```
mpreyskurantov, line 65:
```suggestion
    stylingMode: [Object, String] as PropType<ButtonStyle>,
```

```suggestion
    stylingMode: ['text', 'outlined', 'contained'] as PropType<ButtonStyle>,
```

PR #28180, opened 2024-10-16
DataGrid - The focused state of a row with the 0 key should be restored (T1252962)

markallenramirez:


siarhei-arzamasau (dismissed):

Alyar666 (dismissed):
Please, add test for this case.

siarhei-arzamasau (approved):

PR #28201, opened 2024-10-21
Scheduler a11y: fix appointment aria attrs

pomahtri:


packages/devextreme/js/__internal/scheduler/appointments/m_appointment_collection.ts
```diff
@@ -328,6 +328,10 @@ class SchedulerAppointments extends CollectionWidget {
     }
   }
 
+  _refreshActiveDescendant() {
+    // override to do nothing
```
siarhei-arzamasau, line 332: @pomahtri do we really need this comment?

siarhei-arzamasau (dismissed):

packages/devextreme/js/__internal/scheduler/appointments/m_appointment_collection.ts
pomahtri: Yes, otherwise it will be not obvious why we have empty method

pomahtri: P.S., you mentioned wrong person

Raushen (approved):

PR #28135, opened 2024-10-02
Migrate Core modules to TS

ivanblinov2k17:

PR #28209, opened 2024-10-21
Chat: Implement onTyping events

marker-dao:


packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -7,26 +8,35 @@ import type { Properties as DOMComponentProperties } from '@ts/core/widget/dom_c
...
-import type { EnterKeyEvent } from '../../../ui/text_area';
+import type { EnterKeyEvent, InputEvent } from '../../../ui/text_area';
 import type dxTextArea from '../../../ui/text_area';
 import TextArea from '../m_text_area';
 
 const CHAT_MESSAGEBOX_CLASS = 'dx-chat-messagebox';
 const CHAT_MESSAGEBOX_TEXTAREA_CLASS = 'dx-chat-messagebox-textarea';
 const CHAT_MESSAGEBOX_BUTTON_CLASS = 'dx-chat-messagebox-button';
 
+export const TYPING_START_DELAY = 1500;
```
EugeniyKiyashko: Did we approve these delay values for 'typingstart', 'typingend' with PMs?

packages/devextreme/js/core/utils/throttle.js
```diff
@@ -0,0 +1,13 @@
+export function throttle(func, delay) {
```
EugeniyKiyashko: Because you added a utility in the core area, you need to notify the Essentials team about it and ask them to review it. As far as I know, they are working on a PR for migrating to TS, and your utility could introduce additional difficulties.
https://github.com/DevExpress/DevExtreme/pull/28135 They should take new file into account after merge it

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
```diff
@@ -293,6 +297,237 @@ QUnit.module('MessageBox', moduleConfig, () => {
...
+                keyboardMock(this.$input)
+                    .focus()
+                    .type(value);
+
+                assert.strictEqual(onTypingStartStub.callCount, 0);
+            });
+        });
+
+        QUnit.test('must be called with a delay', function(assert) {
+            const clock = sinon.useFakeTimers({ now: new Date().getTime() });
```
EugeniyKiyashko: Use `test.beforeEach` hook to init `sinon.useFakeTimers()` and will restore it using 'afterEach()' hook

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -81,10 +112,12 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
       autoResizeEnabled: true,
       valueChangeEvent: 'input',
       maxHeight: '8em',
-      onInput: (): void => {
+      onInput: (e: InputEvent): void => {
         const shouldButtonBeDisabled = !this._isValuableTextEntered();
 
         this._toggleButtonDisableState(shouldButtonBeDisabled);
+
+        this._onInputTriggerTypingEventsHandler(e);
```
EugeniyKiyashko: It's unclear why we use 'onInput' & 'Events' inside this method name. Can we do this name more clear?

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -201,6 +236,19 @@ class Chat extends Widget<Properties> {
...
   }
 
+  _typingStartHandler(e: MessageBoxTypingStartEvent): void {
+    const { event } = e;
+    const { user } = this.option();
+
+    this._typingStartAction?.({ user, event });
+  }
+
+  _typingEndHandler(): void {
```
EugeniyKiyashko, line 246: Will we provide the 'event' property that initiates the firing of 'typingEnd' in the event arguments? Let's extract this question and discuss it separately, we need to get approve on it.

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -126,18 +159,52 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
     });
   }
 
+  _onInputTriggerTypingEventsHandler(e: InputEvent): void {
+    if (this._isValuableTextEntered()) {
+      this._throttledTriggerTypingStartAction?.({ event: e.event });
```
EugeniyKiyashko: What are the reasons for us to limit the number of times the typingStart event is triggered?

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -36,20 +46,41 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
...
 
     this._createMessageSendAction();
+    this._createTypingStartAction();
+    this._createTypingEndAction();
+
+    this._wrapTriggerTypingStartEvent();
+  }
+
+  _wrapTriggerTypingStartEvent(): void {
+    this._throttledTriggerTypingStartAction = throttle(
```
EugeniyKiyashko: This function is not an action.

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -126,18 +159,52 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
...
+
+    this._typingStartAction?.({ event: e.event });
+  }
+
   _sendHandler(e: ClickEvent | EnterKeyEvent): void {
     if (!this._isValuableTextEntered()) {
       return;
     }
 
+    clearTimeout(this._typingEndEventTimeout);
```
EugeniyKiyashko: Can we extract part `clearTimeout(this._typingEndEventTimeout)` to separate function `clearTypingEndTimer()' to avoid duplication?

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
```diff
@@ -293,6 +297,237 @@ QUnit.module('MessageBox', moduleConfig, () => {
...
+            this.reinit({ onTypingStart: onTypingStartStub });
+
+            keyboardMock(this.$input)
+                .focus()
+                .type('n');
+
+            assert.strictEqual(onTypingStartStub.callCount, 1);
+        });
+
+        QUnit.test('should be possible to update it at runtime', function(assert) {
```
EugeniyKiyashko, line 319: Also we should be sure that we can update at runtime when 'onTypingStart' option already has function value

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
```diff
@@ -293,6 +297,237 @@ QUnit.module('MessageBox', moduleConfig, () => {
...
+
+                clock.tick(TYPING_END_DELAY);
+
+                assert.strictEqual(onTypingEndStub.callCount, 1, 'is called once after delay');
+            } finally {
+                clock.restore();
+            }
+        });
+
+        ['', ' '].forEach(value => {
```
EugeniyKiyashko: Do you want to test all non valuable text values? In this case let's add case with line-break ' \n'

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
```diff
@@ -293,6 +297,237 @@ QUnit.module('MessageBox', moduleConfig, () => {
...
+            this.reinit({ onTypingEnd: onTypingEndStub });
+
+            try {
+                const keyboard = keyboardMock(this.$input);
+
+                keyboard
+                    .focus()
+                    .type('n');
+
+                clock.tick(TYPING_END_DELAY - 1);
```
EugeniyKiyashko: Let's increase timeout sensitivity to 10 & 20 instead of 1 & 2.

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -188,6 +209,20 @@ class Chat extends Widget<Properties> {
...
+    this._typingStartAction = this._createActionByOption(
+      'onTypingStart',
+      { excludeValidators: ['disabled', 'readOnly'] },
+    );
+  }
+
+  _createTypingEndAction(): void {
+    this._typingEndAction = this._createActionByOption(
+      'onTypingEnd',
+      { excludeValidators: ['disabled', 'readOnly'] },
```
Zedwag: I don't think we need this: there is no readOnly option in chat and there is no way to type in a disabled chat

packages/devextreme/js/__internal/ui/chat/chat.ts
ksercs: There cannot be an `event` parameter because we trigger this by timeout, not after user action

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -19,16 +21,22 @@ import ChatHeader from './header';
...
+  TypingStartEvent as MessageBoxTypingStartEvent,
 } from './messagebox';
 import MessageBox from './messagebox';
 import MessageList from './messagelist';
 
 const CHAT_CLASS = 'dx-chat';
 const TEXTEDITOR_INPUT_CLASS = 'dx-texteditor-input';
 
+type TypingStartEvent = NativeEventInfo<Chat> & { user?: User };
+type TypingEndEvent = NativeEventInfo<Chat> & { user?: User };
```
ksercs, line 33: Why types for start and end are the same although there it no `event` parameter in `end` handler?

packages/devextreme/js/__internal/ui/chat/messagebox.ts
ksercs: I think we can discuss it on testing later

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -126,18 +159,52 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
     });
   }
 
+  _onInputTriggerTypingEventsHandler(e: InputEvent): void {
+    if (this._isValuableTextEntered()) {
```
ksercs: So, if we have "       1" text and we remove last character, onTypingStart will not be called? Looks strange to me

packages/devextreme/js/__internal/ui/chat/messagebox.ts
ksercs: Moreover, it looks like now if we trigger first input event at time `0` and the next one - at time `1499` ms, second input will be ignored, so at time `2000` end event will be still triggered cause no timer update is happened, right? It looks incorrect to me - i think each input event should affect time "to end"
P.s. "each input event affects time to end" !== "each input event recreates a timeout"

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
```diff
@@ -293,6 +297,237 @@ QUnit.module('MessageBox', moduleConfig, () => {
...
+
+                keyboard.type('no');
+
+                assert.strictEqual(onTypingStartStub.callCount, 1, 'called once still');
+
+                clock.tick(TYPING_START_DELAY);
+
+                keyboard.type('not');
+
+                assert.strictEqual(onTypingStartStub.callCount, 2, 'called twice');
```
ksercs: I see no logic why `typingStart` should be repeatedly called if no `typingEnd` occured

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: Thanks a lot. [788227a8e342ba4871b28d8c1988faa534237714]

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: We decided to drop the throttle functionality [30f954d8b2d174191c45c26c9ef35cce6e9da515]

packages/devextreme/js/core/utils/throttle.js
marker-dao: We decided to drop the throttle functionality [30f954d8b2d174191c45c26c9ef35cce6e9da515]

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: [30f954d8b2d174191c45c26c9ef35cce6e9da515]

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: We decided to drop the throttle functionality [30f954d8b2d174191c45c26c9ef35cce6e9da515]

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: [30f954d8b2d174191c45c26c9ef35cce6e9da515]

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: I already improve typing in the next PR

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: [30f954d8b2d174191c45c26c9ef35cce6e9da515]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
marker-dao: We decided to drop the throttle functionality  [30f954d8b2d174191c45c26c9ef35cce6e9da515]

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: [9345dcc44d2d5822ffdceb0dca58e69e99d62bc9]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
marker-dao: Updated

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
marker-dao: Updated

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
marker-dao: Updated

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageBox.tests.js
marker-dao: Okay, I have added a test: _should be possible to update it at runtime if init value is function_

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -36,20 +44,30 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
 
   _messageSendAction?: (e: Partial<MessageSendEvent>) => void;
 
+  _typingStartAction?: (e: Partial<TypingStartEvent>) => void;
+
+  _typingEndAction?: () => void;
+
+  _typingEndTimeoutId?: ReturnType<typeof setTimeout> | undefined;
```
EugeniyKiyashko, line 51: This argument is optional. Does it require a `| undefined` declaration?

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -129,15 +150,54 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
...
+    }
+  }
+
+  _updateTypingEndTimeout(): void {
+    clearTimeout(this._typingEndTimeoutId);
+
+    this._typingEndTimeoutId = setTimeout(() => {
+      this._typingEndAction?.();
+
+      this._typingEndTimeoutId = undefined;
```
EugeniyKiyashko: Why we don't use `_clearTypingEndTimeout` method call?

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -129,15 +150,54 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
...
+
+    this._typingEndTimeoutId = undefined;
+  }
+
   _sendHandler(e: ClickEvent | EnterKeyEvent): void {
     if (!this._isValuableTextEntered()) {
       return;
     }
 
+    this._clearTypingEndTimeout();
```
EugeniyKiyashko, line 198: Should we do this when `this._typingEndTimeoutId` have value or should do it without any conditions? Kindly think about it

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -170,12 +230,27 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
       }
       case 'onMessageSend':
         this._createMessageSendAction();
+
```
EugeniyKiyashko, line 233: Excess line-break here

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: For what? We don't need to call clearTimeout here, because after the callback is called (callback from setTimeout), clearTimeout function is called by the browser automatically. And timeout is cleared by the browser. It is strange to do it manually again.

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: Yes, because in the _clearTypingEndTimeout method and in _updateTypingEndTimeout I set _typingEndTimeoutId = undefined. This type does not match the `ReturnType<typeof setTimeout>`. Even though we specified that the property is optional, TS considers it an error if we do not explicitly specify the type | undefined

packages/devextreme/js/__internal/ui/chat/chat.ts
EugeniyKiyashko, line 33: In test you don't expect that TypingEndEvent will include 'event' property.

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: There shouldn't be any conditions

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: That's right. It is not in the arguments passed.

image

packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -81,10 +99,13 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
       autoResizeEnabled: true,
       valueChangeEvent: 'input',
       maxHeight: '8em',
-      onInput: (): void => {
+      onInput: (e: InputEvent): void => {
         const shouldButtonBeDisabled = !this._isValuableTextEntered();
 
         this._toggleButtonDisableState(shouldButtonBeDisabled);
+
+        this._triggerTypingStartAction(e);
```
EugeniyKiyashko, line 107: Did you consider the scenario where the typing end event fires immediately when the input becomes empty?

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: And in the generic type we didn't pass the second parameter that would point to the event: `NativeEventInfo<dxChat>`

But we have improved the typification in the next PR. Check this: https://github.com/DevExpress/DevExtreme/pull/28228/files#diff-b631436343cddcc3bd6d8e2e22ef2ff9c8d7f46257afcfe43883cae2a6a4c932R51

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: We don't have that scenario. Scenarios that exist:

1. the user enters or deletes something in the input field
2. the user sends a message

packages/devextreme/js/__internal/ui/chat/messagebox.ts
EugeniyKiyashko: I mean that we have a method for these operations, and it's not a mistake to clear an already cleared timeout—the browser won't do anything in that case. This makes it easy to understand what I'm doing, instead of using just a small part of the implementation of the method.

ksercs (dismissed):

packages/devextreme/js/__internal/ui/chat/messagebox.ts
marker-dao: [2364c8ca9cda7da8629c766dd592cad9bb3c5c1a]

ksercs (approved):

EugeniyKiyashko (approved):

ivanblinov2k17:
Merged rebased version
#28227

PR #28216, opened 2024-10-22
Scheduler a11y: There is no info about top-left calendar button group

tongsonbarbs:


e2e/testcafe-devextreme/tests/scheduler/a11y/appointment.ts
```diff
@@ -95,6 +95,47 @@ fixture.disablePageReloads`a11y - appointment`
       ],
     });
   });
+
+  test(`Scheduler a11y: Disabled time ranges are not supported (${currentView})`, async (t) => {
+    const scheduler = new Scheduler('#container');
+    const { nextButton } = scheduler.toolbar.navigator;
+    const { prevButton } = scheduler.toolbar.navigator;
```
Raushen: const { 
    nextButton,
   prevButton,
 } = scheduler.toolbar.navigator;

packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts
```diff
@@ -40,7 +40,10 @@ export const getDateNavigator = (header, item) => {
 const getPreviousButtonOptions = (header) => ({
   key: 'previous',
   icon: 'chevronprev',
-  elementAttr: { class: PREVIOUS_BUTTON_CLASS },
+  elementAttr: {
+    class: PREVIOUS_BUTTON_CLASS,
+    ...getCurrentViewText(header) ? { 'aria-label': `Previous ${getCurrentViewText(header)}` } : {},
```
Raushen: This value should be calculated before. It is more friendly way to debug. 
https://refactoring.guru/extract-variable

packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts
```diff
@@ -143,3 +149,14 @@ const isNextButtonDisabled = (header) => {
   const nextDate = header._getNextDate(1, caption.startDate);
   return nextDate > max;
 };
+
+const getCurrentViewText = (header) => {
+  const { currentView } = header;
+  return typeof currentView === 'string'
+    ? currentView
```
Raushen: It is better to avoid ternary operator:
```
if(!header.currentView || typeof header.currentView !== 'string')
    return "";
} 
```

Raushen: What types in the header.currentView are possible?

packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts
Raushen: And the constants should be localizable, for example dxScheduler-NextMonth.
See the examples here: https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/js/localization/messages/en.json#L252

Raushen: Do not forget to ask Vlada to check them!

packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts
tongsonbarbs: I'm not sure, but this is for qUnit failing tests. There are several tests that currentView is not a string and returns undefined..

packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts
```diff
@@ -39,8 +39,8 @@ export const getDateNavigator = (header, item) => {
 };
 
 const getPreviousButtonOptions = (header) => {
-  // @ts-expect-error text
-  const ariaMessage = messageLocalization.format('dxScheduler-navigationPrevious', getCurrentViewText(header));
+  const { currentView } = header;
+  const ariaMessage = messageLocalization.format(`dxScheduler-navigationPrevious${currentView}`);
```
pomahtri: Please check that `currentView` is the right property. Users can also set custom views, so you need to make sure that when user creates custom month view, this property contains `'month'` and not custom name provided by user

packages/devextreme/js/localization/messages/ar.json
```diff
@@ -245,8 +245,25 @@
         "dxScheduler-editorLabelDescription": "الوصف",
         "dxScheduler-editorLabelRecurrence": "تكرار",
 
-        "dxScheduler-navigationPrevious": "Previous {0}",
-        "dxScheduler-navigationNext": "Next {0}",
+        "dxScheduler-navigationPreviousday": "Previous day",
```
pomahtri: Please add separator, without it it is hard to read the name
Like `dxScheduler-navigationPrevious-day`

packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts
pomahtri: Maybe we better unify the names? Not 'previous day', 'previous month' etc, but 'previous page' or 'previous range'
While 'previous day' looks ok, 'previous agenda' or 'previous timeline month' is not user-friendly

tongsonbarbs: Since user can create custom views, I think I should just stick to 'Previous Page' or 'Previous Range' instead..

pomahtri: Please don't get me wrong, users can create custom views, but they always have predefined type (like 'month', 'day' etc) 
https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/Configuration/views/#type

Custom views are only about setting some specific properties to these views. Like, user can create two month views with different grouping etc

pomahtri: But settings can still introduce some troubles. Like [intervalCount](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/Configuration/views/#intervalCount), which can make the view contain two weeks instead of one, and therefore the label should be also like "next two weeks" instead of "two week"

tongsonbarbs: In this case, I think the navigation labels can be as simple as 'Previous Page' 'Next Page'. As I see there might be a continuous conflict  if we use specific viewTypes with added logic.. What do you think? Please let me know if I should use the specific viewTypes with the required logic that can interpret intervalCount and other instances correctly..

siarhei-arzamasau (dismissed):

vladaskorohodova (dismissed):

packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts
tongsonbarbs: @Raushen @pomahtri , please review the latest commit. I haven't considered the intervalCount because if I will create it, we may not use message localization due to language differences. If I'm not to use messageLocalization, then the aria-label will only be available in english..

packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts
```diff
@@ -37,28 +40,36 @@ export const getDateNavigator = (header, item) => {
...
-  key: 'previous',
-  icon: 'chevronprev',
-  elementAttr: { class: PREVIOUS_BUTTON_CLASS },
-  clickHandler: () => header._updateDateByDirection(DIRECTION_LEFT),
-  onContentReady: (e) => {
-    const previousButton = e.component;
-    previousButton.option('disabled', isPreviousButtonDisabled(header));
+const getPreviousButtonOptions = (header) => {
+  const viewAria = getViewAria(header.currentView.type);
+  const ariaMessage = messageLocalization.format(`dxScheduler-navigationPrevious-${viewAria}`);
```
tongsonbarbs: @Raushen @pomahtri  , please review the latest commit. I haven't considered the intervalCount because if I will create it, we may not use message localization due to language differences. If I'm not to use messageLocalization, then the aria-label will only be available in english..

tongsonbarbs: I am thinking of using pattern for message localization
`Next {0} day{1}`
but as @Raushen told me, it can't be this way since there are differences in language like Greek and Russian..

Raushen (approved):

vladaskorohodova (approved):

PR #28229, opened 2024-10-22
Chat: Implement typingUsers

marker-dao:


packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -417,6 +450,10 @@ class MessageList extends Widget<Properties> {
       case 'showDayHeaders':
         this._invalidate();
         break;
+      case 'typingUsers':
+        this._typingIndicator.option(name, value);
+        this._scrollDownContent();
```
EugeniyKiyashko: I am not sure that we need scrolldown every time when typingUsers is changed. For example, I can scroll up to check Chat's history and when somebody starts typing my scroll will change. Did I get this scenario correct?

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -130,19 +132,27 @@ class Chat extends Widget<Properties> {
...
+
     this.$element().append($messageList);
 
     this._messageList = this._createComponent($messageList, MessageList, {
       items,
       currentUserId,
       showDayHeaders,
-      // @ts-expect-error
-      isLoading: this._dataController.isLoading(),
+      typingUsers,
```
EugeniyKiyashko, line 173: It will perfect if you demonstrate this configuration in Storybook

packages/devextreme-scss/scss/widgets/generic/chat/_colors.scss
```diff
@@ -76,6 +96,11 @@ $chat-errorlist-color: $base-invalid-color !default;
   $chat-messagelist-empty-icon-background-color: $base-inverted-bg !default;
   $chat-messagelist-empty-message-color: $base-text-color !default;
   $chat-messagelist-empty-prompt-color: $base-text-color !default;
+
+  $chat-typingindicator-color: $chat-information-color !default;
+  $chat-typingindicator-bubble-bg-color: $chat-bubble-background-color-secondary !default;
+  $chat-typingindicator-circle-bg-color: lighten($base-bg, 32.55) !default;
```
EugeniyKiyashko: Kindly to be sure that all colors are adjusted according to correspond color themes.
image
The main colors that are using in contrast color scheme are white, black, accent. Don't forget to get designer approve for using these colors.

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
```diff
@@ -38,6 +38,19 @@ $chat-messagelist-empty-prompt-font-size: null !default;
...
 $chat-messagelist-day-header-first-padding-top: 20px !default;
 
+$chat-typingindicator-template: $chat-messagegroup-alignment-start-template !default;
+$chat-typingindicator-padding: null !default;
+$chat-typingindicator-row-gap: $chat-messagegroup-gap !default;
+$chat-typingindicator-column-gap: $chat-messagegroup-alignment-start-gap !default;
+$chat-typingindicator-font-size: $chat-information-font-size !default;
+$chat-typingindicator-bubble-height: null !default;
+$chat-typingindicator-bubble-width: null !default;
+$chat-typingindicator-bubble-border-radius: null !default;
```
EugeniyKiyashko:
```suggestion
$chat-typingindicator-bubble-border-radius: $chat-bubble-border-radius !default;
```

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
```diff
@@ -51,12 +64,17 @@ $chat-messagelist-day-header-first-padding-top: 20px !default;
   $chat-errors-row-gap: 4px !default;
   $chat-error-padding-block: 9px !default;
 
-  $chat-messagelist-empty-icon-box-size: 64px !default;;
+  $chat-messagelist-empty-icon-box-size: 64px !default;
```
EugeniyKiyashko, line 68: Please fix this typo in others file also 
image
Thank you for your attention

packages/devextreme/js/__internal/ui/chat/messagelist.ts
ksercs: For me it also looks redundant and not expected

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.tests.js
```diff
@@ -816,7 +879,27 @@ QUnit.module('MessageList', moduleConfig, () => {
             });
         });
 
-        QUnit.test('should be scrolled to last message if items changed at runtime with an invalidate call', function(assert) {
+        QUnit.test('should be scrolled down if typingUsers changed at runtime', function(assert) {
```
ksercs: This looks unexpectedly to me

packages/devextreme/js/__internal/ui/chat/chat.ts
marker-dao: Cool! I added a task for this one

packages/devextreme-scss/scss/widgets/generic/chat/_colors.scss
marker-dao: I've updated the contrast theme colors

image


Another colors are okay

[67dfa79346db1f1db731dedfe26a5fff17e1a40e]

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
marker-dao: [678f0fc639d7a477863ea280916f5695e33e8010]

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
marker-dao: [536fec0c8bc89a7afd1a00a7114ead566de697b6]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.tests.js
marker-dao: https://github.com/DevExpress/DevExtreme/pull/28229#discussion_r1818731799

packages/devextreme/js/__internal/ui/chat/messagelist.ts
marker-dao: Fix: [47e80e6de3051c8f384466525f90a19ffcb0560a]
Tests: [a714680ada5a7e97314efe9b0168c6f6dd53ac8b]

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -301,6 +312,9 @@ class Chat extends Widget<Properties> {
       case 'showDayHeaders':
         this._messageList.option(name, value);
         break;
+      case 'typingUsers':
```
EugeniyKiyashko, line 344: If action is the same use the following notation
```
case 'showDayHeaders':
case 'typingUsers':
// action
break;
```

marker-dao: [1d5739865c023b6c7a90cfcce7a2845805f29860]

e2e/testcafe-devextreme/tests/chat/typingIndicator.ts
```diff
@@ -0,0 +1,113 @@
...
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+  await testScreenshot(t, takeScreenshot, 'Typing indicator with emptyview.png', { element: '#chat' });
+
+  await t
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+}).before(async () => {
+  await appendElementTo('#container', 'div', 'chat');
+  await insertStylesheetRulesToPage(`.${CHAT_TYPINGINDICATOR_CIRCLE_CLASS} { animation: none; }`);
```
EugeniyKiyashko: We turned off animation but on the etalons there are difference between them. Start position in material doesn't fit start position in fluent
image 
image
If I assume right the circles should have the same sizes beetween themes

e2e/testcafe-devextreme/tests/chat/typingIndicator.ts
```diff
@@ -0,0 +1,113 @@
...
+  const userFourth = createUser(4, 'Julien');
+
+  await chat.option('typingUsers', [userFirst, userSecond]);
+  await testScreenshot(t, takeScreenshot, 'Typing indicator with 2 users.png', { element: '#chat' });
+
+  await chat.option('typingUsers', [userFirst, userSecond, userThird]);
+  await testScreenshot(t, takeScreenshot, 'Typing indicator with 3 users.png', { element: '#chat' });
+
+  await chat.option('typingUsers', [userFirst, userSecond, userThird, userFourth]);
+  await testScreenshot(t, takeScreenshot, 'Typing indicator with 4 users.png', { element: '#chat' });
```
EugeniyKiyashko, line 84: Recomendation: A single configuration should be sufficient to test the appearance of the typing bubble and its layout. To validate different variations of text within the typingindicator-text, it would be more efficient to use QUnit tests.

packages/devextreme/js/__internal/ui/chat/typingindicator.ts
```diff
@@ -0,0 +1,190 @@
...
+const CHAT_TYPINGINDICATOR_CONTENT_CLASS = 'dx-chat-typingindicator-content';
+const CHAT_TYPINGINDICATOR_TEXT_CLASS = 'dx-chat-typingindicator-text';
+const CHAT_TYPINGINDICATOR_BUBBLE_CLASS = 'dx-chat-typingindicator-bubble';
+const CHAT_TYPINGINDICATOR_CIRCLE_CLASS = 'dx-chat-typingindicator-circle';
+
+export interface Properties extends DOMComponentProperties {
+  typingUsers: User[];
+}
+
+class TypingIndicator extends DOMComponent<TypingIndicator, Properties> {
```
EugeniyKiyashko, line 21: Since the typing bubble shares a similar layout with the MessageGroup component - containing an avatar, text above, and a bubble - it makes sense to reuse the MessageGroup component instead. Did you think about it? What issues did you experience?

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
```diff
@@ -51,12 +64,16 @@ $chat-messagelist-day-header-first-padding-top: 20px !default;
...
   $chat-messagelist-empty-icon-size: 40px !default;
   $chat-messagelist-empty-icon-margin-bottom: 12px !default;
   $chat-messagelist-empty-row-gap: 4px !default;
   $chat-messagelist-empty-message-font-size: 18px !default;
   $chat-messagelist-empty-prompt-font-size: 14px !default;
+
+  $chat-typingindicator-bubble-height: 2.571429em !default;
```
EugeniyKiyashko: Why we can't use the approach with paddings in the message bubbles instead of fixed sizes? It would make the values easier to understand. Additionally, the approaches between similar element would become more consistent.

EugeniyKiyashko: Using 'em' can be justified for circle sizes to support responsiveness when the font size changes

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
```diff
@@ -51,12 +64,16 @@ $chat-messagelist-day-header-first-padding-top: 20px !default;
...
+  $chat-typingindicator-bubble-width: 5.428571em !default;
```
EugeniyKiyashko: For compact theme: It should be 24px instead of 28px according to the design

packages/devextreme-scss/scss/widgets/material/chat/_sizes.scss
```diff
@@ -72,10 +89,14 @@ $chat-messagelist-day-header-first-padding-top: 20px !default;
...
-  $chat-messagelist-empty-icon-box-size: 48px !default;;
+  $chat-messagelist-empty-icon-box-size: 48px !default;
   $chat-messagelist-empty-icon-size: 32px !default;
   $chat-messagelist-empty-icon-margin-bottom: 14px !default;
   $chat-messagelist-empty-row-gap: 2px !default;
   $chat-messagelist-empty-message-font-size: 14px !default;
   $chat-messagelist-empty-prompt-font-size: 12px !default;
+
+  $chat-typingindicator-bubble-height: 2.333333em !default;
+  $chat-typingindicator-bubble-width: 5em !default;
```
EugeniyKiyashko: It should be 60x24px instead of 65x30.33px according to the design

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
marker-dao: This is not a true statement because the size of the bubble that contains the text is 28px. I believe this is a design error.

packages/devextreme-scss/scss/widgets/fluent/chat/_sizes.scss
```diff
@@ -71,10 +90,15 @@ $chat-messagelist-day-header-first-padding-top: 20px !default;
...
-  $chat-messagelist-empty-icon-box-size: 48px !default;;
+  $chat-messagelist-empty-icon-box-size: 48px !default;
   $chat-messagelist-empty-icon-size: 32px !default;
   $chat-messagelist-empty-icon-margin-bottom: 14px !default;
   $chat-messagelist-empty-row-gap: 2px !default;
   $chat-messagelist-empty-message-font-size: 14px !default;
   $chat-messagelist-empty-prompt-font-size: 12px !default;
+
+  $chat-typingindicator-bubble-height: 2.333333em !default;
+  $chat-typingindicator-bubble-width: 5em !default;
```
EugeniyKiyashko: The same mistake: https://github.com/DevExpress/DevExtreme/pull/28229/files#r1819850668

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.markup.tests.js
```diff
@@ -86,6 +87,12 @@ QUnit.module('Chat', moduleConfig, () => {
 
             assert.strictEqual($errorList.length, 1);
         });
+
+        QUnit.test('TypingIndicator should be rendered', function(assert) {
```
EugeniyKiyashko:
```suggestion
        QUnit.test('TypingIndicator element should be rendered', function(assert) {
```

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -71,7 +81,12 @@ class MessageList extends Widget<Properties> {
 
     this._renderScrollView();
     this._renderMessageListContent();
+    this._renderEmptyView();
```
EugeniyKiyashko, line 103: Did you consider scenario when Chat is empty and some users are typing simultaneously, what layout should be in this case?

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
marker-dao: I use em to resize bubble and circle proportionally. I didn't use paddings because we primarily need to keep the height and width equal to the size of the bubble that contains the text. Besides, look at the screenshots, you will see that the sizes of paddings are different, which means that you can't use the same variables and you need to introduce new ones or add an additional container, which I try to avoid.

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -417,6 +470,11 @@ class MessageList extends Widget<Properties> {
       case 'showDayHeaders':
         this._invalidate();
         break;
+      case 'typingUsers':
+        this._setIsReachedBottom();
```
EugeniyKiyashko, line 523: If I comment out lines 474 and 477, I don't see any failed QUnit tests. Perhaps I missed something. Could you please provide scenarios or test cases for what these lines are used?

packages/devextreme/js/__internal/ui/chat/typingindicator.ts
marker-dao: I'm not going to complicate the already complex logic of the MessageGroup component. These components serve different purposes, they should not be combined.

e2e/testcafe-devextreme/tests/chat/typingIndicator.ts
marker-dao: That's right. But you check that PR the moment the tests are running, and the screenshots can only be updated after I get new ones.

e2e/testcafe-devextreme/tests/chat/typingIndicator.ts
marker-dao: QUnit tests for this scenario exist

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -397,6 +428,28 @@ class MessageList extends Widget<Properties> {
     this.setAria(aria);
   }
 
+  _setIsReachedBottom(): void {
+    const scrollTop = this._scrollView.scrollTop();
+    const clientHeight = this._scrollView.clientHeight();
+    const scrollHeight = this._scrollView.scrollHeight();
+
+    const isReachedBottom = Math.floor(scrollHeight) === Math.floor(clientHeight + scrollTop);
```
EugeniyKiyashko: It might not work properly with scaling cases. I would prefer to believe more on the solution used in the List component: https://github.com/marker-dao/DevExtreme/blob/1d5739865c023b6c7a90cfcce7a2845805f29860/packages/devextreme/js/__internal/ui/list/m_list.base.ts#L563-L564

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -18,8 +18,10 @@ import Widget from '@ts/core/widget/widget';
 import { isElementVisible } from '../splitter/utils/layout';
 import type { MessageGroupAlignment } from './messagegroup';
 import MessageGroup from './messagegroup';
+import TypingIndicator from './typingindicator';
 
 const CHAT_MESSAGELIST_CLASS = 'dx-chat-messagelist';
+const CHAT_MESSAGELIST_CONTENT_CLASS = 'dx-chat-messagelist-content';
```
EugeniyKiyashko, line 28: Did I understand correctly that it needs to add typing indicator inside scrollview-content element?
Why we can't to insert typing indicator after scrollview content and not add additional level of dom hierarchy?

packages/devextreme-scss/scss/widgets/fluent/chat/_sizes.scss
marker-dao: I believe this is a design error: https://github.com/DevExpress/DevExtreme/pull/28229#discussion_r1819854218

packages/devextreme/js/__internal/ui/chat/messagelist.ts
EugeniyKiyashko: Simething like on this screenshot image

packages/devextreme-scss/scss/widgets/material/chat/_sizes.scss
marker-dao: [97c5030602394817755eb45ba60b230028eb164f]

packages/devextreme/playground/jquery.html
```diff
@@ -52,9 +52,12 @@ <h1 style="position: fixed; left: 0; top: 0; clip: rect(1px, 1px, 1px, 1px);">Te
         <div id="button"></div>
         <script>
             $(function() {
-                $("#button").dxButton({
-                    text: 'Click me!',
-                    onClick: function() { alert("clicked"); }
+                $("#button").dxChat({
+                    typingUsers: [
```
EugeniyKiyashko: Let's revert jquery.html

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/typingIndicator.markup.tests.js
```diff
@@ -0,0 +1,63 @@
...
+        QUnit.test('root element should contain bubble element', function(assert) {
+            const $bubble = this.$element.find(`.${CHAT_TYPINGINDICATOR_BUBBLE_CLASS}`);
+
+            assert.strictEqual($bubble.length, 1);
+        });
+
+        QUnit.test('bubble element should contain 3 circle elements', function(assert) {
+            const $circles = this.$element
+                .find(`.${CHAT_TYPINGINDICATOR_BUBBLE_CLASS}`)
+                .find(`.${CHAT_TYPINGINDICATOR_CIRCLE_CLASS}`);
```
EugeniyKiyashko, line 58: Is the double call of the find method a typo?

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
marker-dao: Screenshot 2024-10-28 at 23 40 51
Screenshot 2024-10-28 at 23 41 08

packages/devextreme/js/__internal/ui/chat/messagelist.ts
marker-dao: Yes reviewed, there are TestCafe tests on these scenarios and new screenshots added.

packages/devextreme/js/localization/messages/en.json
```diff
@@ -346,6 +346,10 @@
         "dxChat-errorListAriaLabel": "Error list",
         "dxChat-emptyListMessage": "There are no messages in this chat",
         "dxChat-emptyListPrompt": "Write your first message",
+        "dxChat-typingMessageSingleUser": "{0} is typing...",
+        "dxChat-typingMessageTwoUsers": "{0} and {1} are typing...",
+        "dxChat-typingMessageThreeUsers": "{0}, {1} and {2} are typing...",
+        "dxChat-typingMessageMultipleUsers": "{0} and other are typing...",
```
```suggestion
        "dxChat-typingMessageMultipleUsers": "{0} and others are typing...",
```

packages/devextreme/js/__internal/ui/chat/typingindicator.ts
EugeniyKiyashko: @ksercs Please, resolve the conversation if you agree with Roman

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
EugeniyKiyashko: We can use inner elements of your segments for animation, for example `before` element or wrap your three circle into wrapper that will have static dimention that equal maximum height value(or more) during animation

packages/devextreme/js/localization/messages/en.json
marker-dao: [a0554e66d5c13058b0b666dbfde3d442427acb15]

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/typingIndicator.markup.tests.js
marker-dao: No, it is clear from the name of the test that we are checking that some elements are inside others

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.markup.tests.js
marker-dao: [a0554e66d5c13058b0b666dbfde3d442427acb15]

packages/devextreme/playground/jquery.html
marker-dao: I haven't yet given the PR to review after the last edits.

packages/devextreme/js/__internal/ui/chat/typingindicator.ts
marker-dao: I think this is a bad idea because good architecture is based on the principle of sole responsibility. If a MessageGroup component already handles messages, including their display and layout, adding TypingIndicator logic to it can turn it into a ‘Swiss knife’ with different functionality. And as we know, the more functions a Swiss knife has, the less convenient it is to open tin cans)))

Components that perform one function are easier to test, refactor and maintain 🙂

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
marker-dao: `We can use inner elements of your segments for animation, for example before element`

We cannot use before or after elements for animation in this case due to the fact that your animation will be very complicated. Your before and after elements will be positioned absolutely to the main element, which will zoom in and out, so visually your before and after elements will move in and out.

I've tested this approach, it doesn't work.

`wrap your three circle into wrapper that will have static dimention that equal maximum height value(or more) during animation`

You're suggesting that I write even more JS code. That's an unnecessary complication. I'm not sure of its effectiveness.

packages/devextreme/js/__internal/ui/chat/messagelist.ts
marker-dao: I'm sure it's methodologically correct to add any content to an element with the dx-scrollview-content class. I don't think it's the right tactic to add other elements to the ScrollView markup. Any such additions should come from the ScrollView component, not the MessageList. 

In addition, ScrollView does not provide any public or even private API for adding anything to its markup.

packages/devextreme/js/__internal/ui/chat/messagelist.ts
marker-dao: Thanks for attention, setTimeout is excessive [5b08da15e8a2606a141b186a140255003deb5016]

vladaskorohodova (dismissed):

packages/devextreme/js/__internal/ui/chat/messagelist.ts
marker-dao: Hm, okay. Let's use it. I also created a task to include this code in scrollable, since this is the second case where we need to know this.

packages/devextreme/js/__internal/ui/chat/typingindicator.ts
ksercs: I've already set LGTM for the JS implementation - I see it's compact and clear and don't see a significant copypaste taken from MessageGroup. I definitely don't like the idea to use current MessageGroup but try configuring it for this feature purposes: at least because then we'll need to move `_getText` logic to MessageList and it'll break the isolation (or create private options in MessageGroup how Roma described above).

However, I agree that we should try to reuse MessageGroup layout. And I'm even not about code but more about styling. Now I see we have duplicated variables e.g. for bubble color. For me it's more expected that we only add several styles for "circle"/indicator itself, but do nothing for bubble/text etc. Now we'll need to make changes in several places if we decide to update bubble/text appearance. 

So, if you Roma do not have strict reasons why we should keep specific styling for the bubble/text, I'd like we try to refactor it in a separate PR later (branch TypingIndicator from MessageGroup) with DOD - remove all styles except the animation

marker-dao: We actually have reasons to add styles specifically for TypingIndicator. Let's discuss this separately in voice.

packages/devextreme/js/__internal/ui/chat/messagelist.ts
marker-dao: [91ca30ce4b31045a71987cb15c53c1eddde9c45c]

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
ksercs: [refactoring]
Hm, cannot we refactor it demonstrating how this magic number is calculated?
U'm afraid we'll not remember in the year why we've used such a strange value (and for me it's actually not clear even now xD )

packages/devextreme/playground/jquery.html
marker-dao: [1f65378087cbfdce9005809e7878398685034413]

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
EugeniyKiyashko: I understand your thoughts, Roman, but I don't understand why we use different approaches to declare the dimensions of the typing bubble (fixed sizes) and the message bubble (paddings). Are you suggesting we leave this difference?

packages/devextreme-scss/scss/widgets/fluent/chat/_colors.scss
```diff
@@ -42,10 +52,15 @@ $chat-errorlist-color: $base-invalid-color !default;
...
   $chat-bubble-background-color-primary: darken(desaturate(adjust-hue($base-accent, 356.3), 14.7), 49.4) !default;
+  
   $chat-information-color: lighten($base-bg, 43.92) !default;
   
   $chat-messagelist-day-header-color: lighten($base-bg, 43.92) !default;
   $chat-messagelist-empty-icon-color: darken($base-icon-color, 25.88) !default;
   $chat-messagelist-empty-icon-background-color: lighten($base-bg, 7.84) !default;
   $chat-messagelist-empty-prompt-color: lighten($base-label-color, 7.84) !default;
+
+  $chat-typingindicator-color: $chat-information-color !default;
```
ksercs, line 63: Cannot we set this value when declaring above?

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
ksercs: UPD: discussed and agreed to use Bubble approach with paddings

e2e/testcafe-devextreme/tests/chat/typingIndicator.ts
marker-dao: [37aabca7f1d6bc866ca3bb6e9c5eaab00e2a6ac1]

packages/devextreme-scss/scss/widgets/generic/chat/_sizes.scss
marker-dao: [18ed05071996d4112d6dd016b9d52f82587863a3]

ksercs (approved):

packages/devextreme-scss/scss/widgets/material/chat/_sizes.scss
```diff
@@ -72,10 +92,16 @@ $chat-messagelist-day-header-first-padding-top: 20px !default;
...
 
-  $chat-messagelist-empty-icon-box-size: 48px !default;;
+  $chat-messagelist-empty-icon-box-size: 48px !default;
   $chat-messagelist-empty-icon-size: 32px !default;
   $chat-messagelist-empty-icon-margin-bottom: 14px !default;
   $chat-messagelist-empty-row-gap: 2px !default;
   $chat-messagelist-empty-message-font-size: 14px !default;
   $chat-messagelist-empty-prompt-font-size: 12px !default;
+
+  $chat-typingindicator-circle-size: math.div(8, 13) * 1em !default;
```
EugeniyKiyashko, line 102: Instead of number value we should use variables, in this case it should be $material-base-font-size. Because if customer customize font-size value using Themebuilder will not get changes for circle size

EugeniyKiyashko: Let's consider it in future task. Kindly create task for this

EugeniyKiyashko (approved):

packages/devextreme-scss/scss/widgets/material/chat/_sizes.scss
marker-dao: We don't have those variables. 

If you mean the variable for font-size, it cannot be used because it contains a value in pixels:

https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme-scss/scss/widgets/fluent/_sizes.scss#L37

When interpreting scss styles, an error will occur because you are trying to mix px and em units.

Also, note that we only care about the ratio of 8 to 13 (or other), we only care about the ratio because the circle sizes will still end up being calculated relative to the font size inherited by this elements. So there is no error or oversight here, because all that matters to you at this stage is the magnification ratio.

EugeniyKiyashko: I don't see any difficulties transform it to number without units. Do you need a demonstration how we can do it?

vladaskorohodova (approved):

PR #28254, opened 2024-10-28
Scheduler a11y: Appointment collector button doesn't have info about date

tongsonbarbs:


packages/devextreme/js/__internal/scheduler/m_compact_appointments_helper.ts
```diff
@@ -181,4 +184,29 @@ export class CompactAppointmentsHelper {
...
+  _getEndDate(appointment) {
+    const date = appointment.endDate;
+    return date ? new Date(date) : null;
+  }
+
+  _getDateText(appointment) {
+    const startDateText = this._localizeDate(this._getStartDate(appointment));
+    const endDateText = this._localizeDate(this._getEndDate(appointment));
+
+    const dateText = startDateText === endDateText
```
siarhei-arzamasau, line 208: @tongsonbarbs what would happen if both of the dates would be `null`?

packages/devextreme/js/__internal/scheduler/m_compact_appointments_helper.ts
```diff
@@ -181,4 +184,29 @@ export class CompactAppointmentsHelper {
       .append($('<span>').text(text))
       .addClass(APPOINTMENT_COLLECTOR_CONTENT_CLASS);
   }
+
+  _localizeDate(date) {
+    return `${dateLocalization.format(date, 'monthAndDay')}, ${dateLocalization.format(date, 'year')}`;
+  }
+
+  _getStartDate(appointment) {
```
siarhei-arzamasau, line 192: @tongsonbarbs what do you think if you merge `_getStartDate` and `_getEndDate` into one function which receives `appointment.date`, looks like both of these functions are doing the same job.

tongsonbarbs: I wanted to set a practice where we follow Single Responsibility Pattern.. These both functions may look similar but they return different and specific data.

packages/devextreme/js/__internal/scheduler/m_compact_appointments_helper.ts
tongsonbarbs: Thanks for noticing this, I tested it and it is returning undefined as string.. I have committed the changes now to fix this..

siarhei-arzamasau (approved):

PR #28255, opened 2024-10-28
Fix T1259705 - DevExtreme - "t.getClientRects is not a function"

GoodDayForSurf:
error occurs in Angular when SSR is enabled


dxvladislavvolkov (approved):

packages/devextreme-angular/gulpfile.js
```diff
@@ -326,7 +326,7 @@ gulp.task('test.components.server.debug', (done) => {
   new karmaServer(config, done).start();
 });
 
-gulp.task('run.tests', gulp.series('test.components.client'/*, 'test.components.server'*/));
+gulp.task('run.tests', gulp.series('test.components.client', 'test.components.server'));
```
ivanblinov2k17, line 329: Do we really mean to disable these tests?

ivanblinov2k17 (approved):

PR #28256, opened 2024-10-28
Button: extend Button type with any string (T1255652)

nikkithelegendarypokemonster:


packages/devextreme/js/common.d.ts
```diff
@@ -71,7 +71,7 @@ export type ButtonStyle = 'text' | 'outlined' | 'contained';
  * @public
  * @namespace DevExpress.common
  */
-export type ButtonType = 'danger' | 'default' | 'normal' | 'success';
+export type ButtonType = 'danger' | 'default' | 'normal' | 'success' | 'string';
```
ksercs: [incorrect solution]
You've just added one more possible value `"string"`, but we need this option to accept any string value besides predefined (how it's done for text above)

ksercs (changes requested):
@nikkithelegendarypokemonster your change is related to Button, not a Button Group. Please update the name with `Button: extend ButtonType with any string`

ksercs (approved):

EugeniyKiyashko (approved):

mpreyskurantov (approved):

PR #28257, opened 2024-10-28
Chat: Add showAvatar, showMessageTimestamp, showUserName properties

Strider2342:

PR #28261, opened 2024-10-29
Chat: Add showAvatar, showMessageTimestamp, showUserName properties

Strider2342:
This PR is a recreation of https://github.com/DevExpress/DevExtreme/pull/28257
Please consider this as the relevant one.


packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -136,6 +139,9 @@ class Chat extends Widget<Properties> {
       items = [],
       user,
       showDayHeaders = true,
+      showAvatar = true,
+      showUserName = true,
+      showMessageTimestamp = true,
```
ksercs: [corner case]
These `= true` are used only in case when we intentionally pass `undefined` the the Chat's option.
And for this case it looks strange to me that we gonna handle it as `true` because undefined is usually handled as `false`. Moreover, I'm pretty sure that it'll be handled as `false` now it we do it at runtime through `.option(optionName, undefined)`.
So, I suggest you using `false` here and add tests for this.
P.s. Kindly update for `showDayHeaders` as well

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -435,6 +451,10 @@ class MessageList extends Widget<Properties> {
 
     switch (name) {
       case 'currentUserId':
+      case 'showDayHeaders':
```
ksercs, line 454: [copypaste]
We have copypaste now, it's duplicated below. Kindly union all cases when `_invalidate` is called

Strider2342: e6f6146790b208b2332e9d492319db37dd726ad1

packages/devextreme/js/__internal/ui/chat/chat.ts
Strider2342: Should I also test the case when we pass undefineed on init?

Strider2342: Altough it works as undefined is falsy, the value itself will be undefined.

Strider2342: 3db30f88525eb513cd10b51d925665c3c981bdf2

ksercs (approved):

EugeniyKiyashko (approved):

Strider2342:
[Chat: implement showAvatar, showUsername, showMessageTimestamp](https://trello.com/c/XVkZ9osa/9495-chat-implement-showavatar-showusername-showmessagetimestamp)

PR #28265, opened 2024-10-29
Chat: add integration to support the DataSource's push API

EugeniyKiyashko:


apps/react-storybook/stories/chat/Chat.stories.tsx
```diff
@@ -214,37 +213,61 @@ export const DataLoading: Story = {
...
+                    const promise = new Promise((resolve) => {
+                        setTimeout(() => {
+                            resolve(messagesRef.current);
+                        }, 500);
+                    });
+            
+                    return promise;
+                },
+                insert: (message) => {
+                    if(reloadOnChange) {
```
ksercs, line 238: Why do we need it? I believe it should be executed in both cases

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.tests.js
```diff
@@ -948,10 +952,99 @@ QUnit.module('Chat', () => {
 
             this.clock.tick(timeout);
 
+            assert.deepEqual(this.instance.option('items'), [...messages, newMessage], 'items option should contain all messages including the new one');
+            assert.strictEqual(this.getBubbles().length, 3, 'new message should be rendered in list');
+        });
+
+        QUnit.test('new message should be rendered when using store.push({ type: insert }) and reloadOnChange is true', function(assert) {
```
ksercs:
```suggestion
        QUnit.test('new message should be rendered after message is rendered if reloadOnChange is true', function(assert) {
```

EugeniyKiyashko: I'm not sure that new test name is informative, it looks like tautology

apps/react-storybook/stories/chat/Chat.stories.tsx
EugeniyKiyashko: When 'reloadOnChange' is true it requires to emulate data loading. I mean after call of built in reload() method we need to resolve() the load method with previous messages plus new one.
When 'reloadOnChange' is false the new message adds to the Chat after call store.push() method, and we render new message in list. It's not necessary to synchronize state with data here.

The main reason why I added this condition is to avoid double adding last item after switching the reloadOnChange (from false to true) mode using select panel in StoryBook.

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.tests.js
ksercs: after message is entered*

EugeniyKiyashko: Fixed

apps/react-storybook/stories/chat/Chat.stories.tsx
ksercs: I gonna trust you about this

ksercs (approved):

marker-dao (approved):

PR #28268, opened 2024-10-30
Chat: Add showDayHeaders, showAvatar, showUserName and showMessageTimestamp options to Storybook

Strider2342:


apps/react-storybook/stories/chat/Chat.stories.tsx
```diff
@@ -30,6 +30,10 @@ const commonArgs: ChatTypes.Properties = {
     activeStateEnabled: true,
     hoverStateEnabled: true,
     focusStateEnabled: true,
+    showDayHeaders: true,
```
EugeniyKiyashko: After adding options to commonArgs, we will see these options in the select panel for all stories. 
image


To make it work, we need to pass the options to the component through all stories.

We have several options:
- Demonstrate the options in all stories.
- Add a separate story to demonstrate how these options work there
- Demonstrate them only in the overview demo.

Option two is preferable to me

Strider2342: I have created a separate story (Customization) in which all the other options are hidden on the controls tab.
f6007c1b13d32d0c3698e6703fbb27445393f7e6

apps/react-storybook/stories/chat/Chat.stories.tsx
```diff
@@ -334,3 +334,84 @@ export const PopupIntegration: Story = {
...
+
+export const Customization: Story = {
+    args: {
+        showDayHeaders: true,
+        showAvatar: true,
+        showUserName: true,
+        showMessageTimestamp: true,
+        items: initialMessages,
+        user: firstAuthor,
+        errors: [],
```
EugeniyKiyashko: I would prefer to simplify it as following:
```
export const Customization: Story = {
    args: {
        width: 500,
        height: 600,
        showDayHeaders: true,
        showAvatar: true,
        showUserName: true,
        showMessageTimestamp: true,
    },
    render: ({
        width,
        height,
        showDayHeaders,
        showAvatar,
        showUserName,
        showMessageTimestamp,
    }) => {
        return (
            <div style={{ display: 'flex', justifyContent: 'center', alignItems: 'center' }}>
                <Chat
                    width={width}
                    height={height}
                    items={initialMessages}
                    user={secondAuthor}
                    showDayHeaders={showDayHeaders}
                    showAvatar={showAvatar}
                    showUserName={showUserName}
                    showMessageTimestamp={showMessageTimestamp}
                >
                </Chat>
            </div>
        );
    }
}
```

What do you think about this suggestion?

Strider2342: Yep, it was a bit overcomplicated :)
b5b3568720d89710b638977cdb03e798e0aef78e

EugeniyKiyashko (approved):

ksercs (approved):

PR #28271, opened 2024-10-30
DataGrid - The pageIndex method should return a promise if dataSource is null (T1256070)

markallenramirez:


packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -57,6 +57,10 @@ const changePaging = function (that, optionName, value) {
     return dataSource[optionName]();
   }
 
+  if (optionName === 'pageIndex' && value !== undefined) {
```
Raushen, line 60: It is better to use if-else-if construction:
```
if (dataSource) { // ok, we have dataSource
   ...
} else if(optionName === 'pageIndex' && value !== undefined) { // and dataSource is not defined
   ...
} else { 
   return 0; // if dataSource is not defined and not (optionName === 'pageIndex' && value !== undefined)
}
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/dataGrid.tests.js
```diff
@@ -2366,6 +2366,26 @@ QUnit.module('Assign options', baseModuleConfig, () => {
         assert.equal(contentReadyCallCount, 2);
     });
 
+    QUnit.test('pageIndex should return correct values when dataSource is null', function(assert) {
```
Raushen: For functional tests we prefer TestCafe. In your case it will be a trivial test case:

```
test('T1256070 - DataGrid - "pageIndex(...).done is not a function" error occurs if dataSource is not defined and pageIndex is chained with the done function', async (t) => { // Test name
   // Just check the page loaded without errors
    await t
      .expect(dataGrid.isReady())
      .ok();
  }).before(() => createWidget('dxDataGrid', {
   // And your config
    keyExpr: 'ID',
    columns: ['CompanyName', 'City', 'State', 'Phone', 'Fax'],
    showBorders: true,
    onContentReady: function(e) {
      e.component.pageIndex(1).done(() => {});
    }
  }));
```

siarhei-arzamasau (dismissed):

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
markallenramirez: Hi, thank you for the recommendation. However, I cannot apply it because of `no-else-return` (allowElseIf=false).

Raushen (approved):

PR #28275, opened 2024-10-30
R1: move utils

pomahtri:


packages/devextreme/js/__internal/ui/scroll_view/utils/get_translate_values.test.tsx
```diff
@@ -1,8 +1,9 @@
-import { getTranslateValues } from '../get_translate_values';
-import { getElementTransform } from '../get_element_style';
+/* eslint-disable @typescript-eslint/no-unsafe-return */
```
marker-dao, line 1: Do we really need that comment here?

marker-dao (approved):

packages/devextreme/js/__internal/ui/scroll_view/utils/get_translate_values.test.tsx
pomahtri: Unfortunately, we do
`jest.requireActual` in code below returns any and this lint rule fails

image
image

EugeniyKiyashko (approved):

PR #28285, opened 2024-10-31
Chat - add messageTemplate in .d.ts

Zedwag:


packages/devextreme/js/ui/chat.d.ts
```diff
@@ -151,6 +160,20 @@ export type Message = {
     typing?: boolean;
 };
 
+/** @public */
+export type MessageTemplateData = {
+    readonly component: dxChat;
+    readonly message?: Message;
+    readonly text?: string;
```
ksercs: We don't need text property, it's included in `message`

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -151,6 +160,20 @@ export type Message = {
...
+};
+
+/**
+ * @docid
+ * @namespace DevExpress.ui.dxChat
+ * @public
+ */
+export type MessageTemplate = ((data: MessageTemplateData, messageBubbleElement: DxElement) => string | UserDefinedElement);
```
ksercs: why do you write `messageBubbleElement`?
I thought it's just `container` in all out templates

Zedwag: no, in .d.ts there are different names, but not container, like `itemElement`, `groupElement`,`fieldElement`, `contentElement` for different templates. I decided it's ok to name it `messageBubbleElement` in this case

p.s. just to make it clear, it's not related to a name we use in templates code, it's just a second arg of function and user can name it whatever he wants

packages/devextreme/js/ui/chat.d.ts
Zedwag: oh, right, fixed ty

packages/devextreme/js/ui/chat.d.ts
mpreyskurantov: Do you really need a new standalone (docid - documentable) type for a template? AFAIK it is a single case for a such extraction, except for https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/js/ui/filter_builder.d.ts#L168
Looks like the `@public` is enough, or even better - to inline for now
The rest is fine to me

Zedwag: Changed to inline

mpreyskurantov (approved):

EugeniyKiyashko (approved):

PR #28286, opened 2024-11-01
DataGrid - Sticky Columns - fixes in master-detail, summary & grouping

pomahtri:


packages/devextreme/js/__internal/grids/grid_core/sticky_columns/m_sticky_columns.ts
```diff
@@ -131,6 +131,8 @@ const baseStickyColumns = <T extends ModuleType<ColumnsView>>(Base: T) => class
     const $cell = super._createCell(options);
     const hasStickyColumns = this.hasStickyColumns();
     const rowIndex = rowType === 'header' ? options.rowIndex : null;
+    const isSummary = rowType === 'groupFooter' || rowType === 'totalFooter' || rowType === 'group';
```
siarhei-arzamasau, line 134: What do you think to move such expression in utility function?

siarhei-arzamasau (dismissed):

packages/devextreme/js/__internal/grids/grid_core/sticky_columns/m_sticky_columns.ts
pomahtri: As I think, it's better not to, unfortunately
Utility like `isSummaryRow` can be reused, but this `isSummary` definition is local.
The key difference is that summary can be in group row (`rowType === 'group'` case), but it doesn't mean that it is summary row. This nuance (whether to consider group row as summary) is local and depends, and in other cases it's better to think twice when reusing this code

siarhei-arzamasau (dismissed):

Alyar666 (approved):

PR #28293, opened 2024-11-01
List: Add group header to focus order if collapsibleGroups is true

marker-dao:


packages/devextreme/js/__internal/ui/list/m_list.base.ts
```diff
@@ -345,6 +371,15 @@ export const ListBase = CollectionWidget.inherit({
   },
 
   _init() {
+    const { collapsibleGroups } = this.option();
+
+    if (collapsibleGroups) {
+      this._activeStateUnit = [
+        ...DEFAULT_ACTIVE_STATE_UNIT_ARRAY,
+        `.${LIST_GROUP_HEADER_CLASS}`,
+      ].join(',');
```
ksercs: [refactoring]
This can be done though defaultOptionRules mechnism

packages/devextreme/js/__internal/ui/list/m_list.base.ts
```diff
@@ -605,31 +640,62 @@ export const ListBase = CollectionWidget.inherit({
...
+    };
+
+    const action = this._createAction(actionCallback, actionParams);
+
+    action({ event: e });
+  },
+
+  _enterKeyHandler(e): void {
+    const { collapsibleGroups, focusedElement } = this.option();
+    const isGroupHeader = $(focusedElement).hasClass(LIST_GROUP_HEADER_CLASS);
```
ksercs, line 691: It looks more logical to me checking if e.target is a group header, not a focused element. Theoretically enter can be triggered for not focused item

packages/devextreme/js/__internal/ui/list/m_list.base.ts
ksercs: [missed scenario?]
What if `collapsibleGroups` is changed at runtime?
I also see no tests for this case

marker-dao: By the way, ‘working defaultOptionRules mechanism’ sounds like a good topic for the documentation section.

marker-dao: I'm pretty sure defaultOptionsRules can't help in this scenario, as _activeStateUnit is not a component option, but a class property:

https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/js/__internal/ui/list/m_list.base.ts#L62 

I'm also pretty sure that defaultOptionsRules ideologically can only be used exclusively for options.

Also, I'm sure that rewriting _activeStateUnit to an option would be difficult within this PR because this property is inherent in many classes (see screenshot below).

Let's make a separate task for this? 

image

packages/devextreme/js/__internal/ui/list/m_list.base.ts
marker-dao: The mechanism by which focusing works in collection_widget is a bit different. Let me explain. document.activeElement is scrollview in this case, so when you press Enter in e.target it will be the scrollview and not the header. We store the focused element in the focusedElement option.

See how it works in collection_widget: 

https://github.com/DevExpress/DevExtreme/blob/24_2/packages/devextreme/js/__internal/ui/collection/m_collection_widget.base.ts#L89

packages/devextreme/js/__internal/ui/list/m_list.base.ts
marker-dao: [10038c0b8e7df4df3c692d3c1f64327e9da1d4b0], [88dd969107513bc1a62715a7692d711a986024c7]

ksercs: Yeah, I haven't noticed it's not an option.
I agree to create a separate card for refactoring - this code looks very strange for me, even in terms of naming

packages/devextreme/js/__internal/ui/list/m_list.base.ts
ksercs: Oh, okay, I see

ksercs (approved):
Focus move on header click/enter press and aria attributes will be updated in separate PRs

EugeniyKiyashko (approved):

PR #28295, opened 2024-11-02
Fix executing angular tests on CI

GoodDayForSurf:


.github/workflows/wrapper_tests.yml
```diff
@@ -80,9 +83,8 @@ jobs:
       - name: Angular - Build
         run: pnpx nx build devextreme-angular
 
-# NOTE: temporary skipped during migrating to the PNPM
```
dxvladislavvolkov, line 83: Do we need the note?

packages/devextreme-angular/karma.test.shim.js
```diff
@@ -8,6 +8,6 @@ testing.TestBed.initTestEnvironment(
   browser.platformBrowserDynamicTesting(),
 );
 
-const context = require.context('./tests/dist', true, /^.\/(?!.*\/ssr-.*\.spec.js$).*\.spec\.js$/);
+const context = require.context('./tests/dist', true, /^.\/(?!.*\/ssr-components\.spec.js$).*\.spec\.js$/);
```
dxvladislavvolkov, line 11: Why only components? We have tests for ajax and platform server

dxvladislavvolkov (dismissed):

alexslavr (approved):

PR #28299, opened 2024-11-04
MultiView: fix on infinite looping issue during normalizing selectIndex

nikkithelegendarypokemonster:


packages/devextreme/js/__internal/ui/m_multi_view.ts
```diff
@@ -100,9 +100,19 @@ const MultiView = CollectionWidget.inherit({
...
+  _isAllItemsHidden() {
+    const { items } = this.option();
+
+    return items.every((_, index) => !this._isItemVisible(index));
+  },
+
   _normalizeIndex(index, direction, loop = true) {
     const count = this._itemsCount();
 
+    if (count === 0 || this._isAllItemsHidden()) {
```
ksercs: [excess code?]
Why do we need `count === 0` here? If `items=[]` then `this._isAllItemsHidden()` will return `true`, so it seems redundant

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
```diff
@@ -504,6 +504,26 @@ QUnit.module('interaction via swipe', {
         });
     });
 
+    QUnit.test('when all items are hidden selectedIndex should be initialized to 0 after swipe', function(assert) {
```
ksercs: You are fixing not this, you're fixing infinite loop error. Moreover, it seems like setting selectedIndex to 0 when all items are hidden are absolutely not related to the code you re changing, it's made separately [there](https://github.com/DevExpress/DevExtreme/pull/28299/files#diff-b95a070f9f0a64ed59233db53b817dce631e45fc0a7e5c12ff624c54ffd8b93aR164)
```suggestion
    QUnit.test('there should not be infinite loop if try swiping when all items are hidden ', function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
```diff
@@ -504,6 +504,26 @@ QUnit.module('interaction via swipe', {
...
+            swipeEnabled: true
+        });
+        const instance = $multiView.dxMultiView('instance');
+        const $itemContainer = $multiView.find(`.${MULTIVIEW_ITEM_CONTAINER_CLASS}`);
+        const pointer = pointerMock($multiView);
+
+        pointer.start().swipeStart().swipe(-0.5).swipeEnd(-1);
+        assert.strictEqual(position($itemContainer), 0, 'container did not move');
+        assert.strictEqual(instance.option('selectedIndex'), 0, 'selectedIndex is not changed');
+    });
```
ksercs, line 536: [broken behavior]
We cannot write such a test because it breaks the page if it's red. If it becomes red one day, it'll be impossible to see the test report / debug / find out the reason because page will be just unresponsive => this test is not only not helpful but also harmful.
For testing your scenario you can mock `_isVisible` method to check how many times it's called. Actually that's also not very good approach because it tests implementation and if we refactor it some day the test can become red or worse evergreen. But that's the easiest approach.

```javascript
        const isItemVisibleStub = sinon
            .stub(instance, '_isItemVisible')
            .callsFake(() => {
                if(itemVisibleStub.callCount === 20) {
                    assert.ok(false, 'infinite loop');
                    return true;
                }
                return false;
            });

        pointer.start().swipeStart().swipe(-0.5).swipeEnd(-1);

        assert.ok(true, 'no infinite loop');
```

ksercs (changes requested):

packages/devextreme/js/__internal/ui/m_multi_view.ts
nikkithelegendarypokemonster: i agree also on this i didnt forsee this

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
nikkithelegendarypokemonster: [563cea3](https://github.com/DevExpress/DevExtreme/pull/28299/commits/563cea32b40366930aa50fa71d7ae64a401c7b3a)

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
nikkithelegendarypokemonster: [563cea3](https://github.com/DevExpress/DevExtreme/pull/28299/commits/563cea32b40366930aa50fa71d7ae64a401c7b3a)

packages/devextreme/js/__internal/ui/m_multi_view.ts
nikkithelegendarypokemonster: [563cea3](https://github.com/DevExpress/DevExtreme/pull/28299/commits/563cea32b40366930aa50fa71d7ae64a401c7b3a)

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
```diff
@@ -504,6 +504,36 @@ QUnit.module('interaction via swipe', {
...
+        const isItemVisibleStub = sinon.stub(instance, '_isItemVisible').callsFake(() => {
+            if(isItemVisibleStub.callCount === 20) {
+                assert.ok(false, 'infinite loop detected');
+                return true;
+            }
+            return false;
+        });
+
+        pointer.start().swipeStart().swipe(-0.5).swipeEnd(-1);
+
```
EugeniyKiyashko: Let's leave single line break

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
```diff
@@ -504,6 +504,36 @@ QUnit.module('interaction via swipe', {
...
+
+        assert.ok(true, 'no infinite loop');
+
+        isItemVisibleStub.restore();
```
EugeniyKiyashko: I would prefer to be safe and, if something went wrong, revert to the state before the test execution
```
const isItemVisibleStub = sinon.stub(...);

try {
    // problematic code, operation
} finally {
    isItemVisibleStub.restore();
}
```

nikkithelegendarypokemonster: [4938424](https://github.com/DevExpress/DevExtreme/pull/28299/commits/493842406d6cb591fd111d55fee5de9c62515e68)

packages/devextreme/testing/tests/DevExpress.ui.widgets/multiView.tests.js
nikkithelegendarypokemonster: [4938424](https://github.com/DevExpress/DevExtreme/pull/28299/commits/493842406d6cb591fd111d55fee5de9c62515e68)

EugeniyKiyashko (approved):

ksercs (approved):

PR #28304, opened 2024-11-05
Chat: support for dataSource operations 'update' and 'remove'

EugeniyKiyashko:


Zedwag (dismissed):

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -100,15 +99,10 @@ class Chat extends Widget<Properties> {
...
-      applyBatch({
-        // // @ts-expect-error
-        keyInfo: dataSource,
-        data: this.option('items'),
-        changes: e.changes,
-      });
+      this._setOptionWithoutOptionChange('items', newItems.slice());
+      this._messageList._setOptionWithoutOptionChange('items', newItems.slice());
+
+      this._messageList._renderEmptyView();
```
ksercs: [naming]
This looks strange that we call `_renderEmptyView` although we don't wanna render in here in the most of cases. I'd expect we call it like `_updateEmptyViewVisibility` / `_toggleEmptyView` / smth else

packages/devextreme/js/__internal/ui/chat/messagegroup.ts
```diff
@@ -111,7 +113,9 @@ class MessageGroup extends Widget<Properties> {
   }
 
   _renderMessageBubble(message: Message): void {
-    const $bubble = $('<div>');
+    const $bubble = $('<div>')
+      .data(MESSAGE_DATA_KEY, message);
```
ksercs, line 117: Why have u decided to store all message data, not only id?

packages/devextreme/js/__internal/ui/chat/messagebubble.ts
```diff
@@ -44,6 +44,9 @@ class MessageBubble extends Widget<Properties> {
 
       const messageTemplate = this._getTemplateByOption('template');
 
+      // @ts-expect-error
+      templateData.message.text = text;
```
ksercs, line 48: [will be refactored in a [separate PR](https://github.com/DevExpress/DevExtreme/pull/28320/files)]

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -400,10 +393,24 @@ class MessageList extends Widget<Properties> {
     $lastAlignmentEndGroup.addClass(CHAT_LAST_MESSAGEGROUP_ALIGNMENT_END_CLASS);
   }
 
+  _getLastMessageGroup(): MessageGroup | undefined {
+    const $lastMessageGroup = this._$content.find(`.${CHAT_MESSAGEGROUP_CLASS}`).last();
+
+    if ($lastMessageGroup.length) {
+      return this._getMessageGroupInstanceByElement($lastMessageGroup);
```
ksercs: I think that creation of a separate method for this operation only complicated the code (at least, it should be not a class method because it does not use `this` and ideologically not related to the MessageList).
I think it's better to so `MessageGroup.getInstance($element) as MessageGroup` right here

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -431,6 +438,80 @@ class MessageList extends Widget<Properties> {
...
+
+      if (messageData.id === key) {
+        result = $(item);
+        return false;
+      }
+
+      return true;
+    });
+
+    return result;
```
ksercs, line 458: [refactoring]
This looks very imperative to me. And probably it's not very good from performance point of view.
I'm not quire familiar with jQuery `data`, but if we use a usual attribute like `data-message-id`, we'll be able to find necessary bubble though a single operation 
```javascript
$('[data-message-id="`${key}`")
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.tests.js
```diff
@@ -25,6 +25,10 @@ const CHAT_MESSAGEBOX_BUTTON_CLASS = 'dx-chat-messagebox-button';
 const CHAT_MESSAGEBOX_TEXTAREA_CLASS = 'dx-chat-messagebox-textarea';
 const CHAT_MESSAGELIST_EMPTY_VIEW_CLASS = 'dx-chat-messagelist-empty-view';
 const SCROLLVIEW_REACHBOTTOM_INDICATOR = 'dx-scrollview-scrollbottom';
+const CHAT_MESSAGELIST_DAY_HEADER_CLASS = 'dx-chat-messagelist-day-header';
+
+const CHAT_LAST_MESSAGEGROUP_ALIGNMENT_START_CLASS = 'dx-chat-last-messagegroup-alignment-start';
+const CHAT_LAST_MESSAGEGROUP_ALIGNMENT_END_CLASS = 'dx-chat-last-messagegroup-alignment-end';
```
ksercs, line 31: [refactoring idea]
Don't u want import all this classes from the implementation directly?

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.tests.js
```diff
@@ -1091,10 +1089,397 @@ QUnit.module('Chat', () => {
...
+
+            assert.strictEqual(this.getEmptyView().length, 0, 'empty view is not rendered');
+
+            store.push([{ type: 'remove', key: 1 }]);
+            this.clock.tick(timeout * 2);
+
+            assert.strictEqual(this.getEmptyView().length, 1, 'empty view is rendered');
+        });
+
+        QUnit.test('emptyview should be rendered after each removal and removed after each addition from the store', function(assert) {
```
ksercs: I didn't get what u're testing here. The test name says we should add/remove emptyView after reach add/remove operation.. But you assert once after all operations, so it looks the same as the ["emptyview should be rendered after the last message is deleted from the store" test](https://github.com/DevExpress/DevExtreme/pull/28304/files#diff-c28d8d0fcc791fc81e519f30bf992c6eb3607f134382d592b9518688d76a640eR1390).

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.tests.js
```diff
@@ -677,6 +679,56 @@ QUnit.module('MessageList', () => {
             assert.strictEqual($secondMessageGroupBubbles.length, 1, 'correct bubble count');
         });
 
+        QUnit.test('new messages should be rendered after the last group', function(assert) {
```
ksercs, line 682: Why is scenario new? How it can be rendered not after last group?

EugeniyKiyashko: This test was added because when I broke this functionality, and no existing tests revealed the issue.

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.tests.js
EugeniyKiyashko: Sure, I want to use this approach in all of Chat's tests and other places. It also needs to be implemented as part of the code styling. I’ve extracted your suggestion into a separate card.

packages/devextreme/js/__internal/ui/chat/messagegroup.ts
EugeniyKiyashko: We use this approach in other components, e.g., in all collections. I would prefer not to come up with a new solution, although I understand that in our case, the ID is sufficient

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.tests.js
EugeniyKiyashko: Thanks. Scenario was simplified: https://github.com/DevExpress/DevExtreme/pull/28304/commits/f7b68176e3e980d976fa7a3b8f77a8fa36812a9b

packages/devextreme/js/__internal/ui/chat/messagelist.ts
EugeniyKiyashko: Ged rid of its using: https://github.com/DevExpress/DevExtreme/commit/f7b68176e3e980d976fa7a3b8f77a8fa36812a9b

packages/devextreme/js/__internal/ui/chat/chat.ts
EugeniyKiyashko: Fixed: https://github.com/DevExpress/DevExtreme/commit/f7b68176e3e980d976fa7a3b8f77a8fa36812a9b

packages/devextreme/js/__internal/ui/chat/messagelist.ts
EugeniyKiyashko: The same here: https://github.com/DevExpress/DevExtreme/pull/28304#discussion_r1834013999. Now, I am not see reasons to refactor it.

marker-dao (approved):

packages/devextreme/js/__internal/ui/chat/messagelist.ts
ksercs: I'm worried about performance, but gonna trust you

ksercs (approved):

PR #28305, opened 2024-11-05
SelectBox: proposal for combobox option for popup (a11y)

Strider2342:
We can now set whether popup gets aria role `dialog` or `combobox`. Let's discuss this solution!
(I've commited with --no-verify as eslint considered `combobox` a misspell)


Strider2342:
Unnecessary change, got wrong conclusions during first research.

PR #28315, opened 2024-11-07
FilterBuilder - First field uses the dataField property while subsequent fields use the name property in the filter value

tongsonbarbs:


packages/devextreme/js/__internal/filter_builder/m_filter_builder.ts
```diff
@@ -516,6 +516,15 @@ class FilterBuilder extends Widget<any> {
     }
   }
 
+  _getUniqueFieldIdentifier(item, currentCondition) {
+    const isDataFieldUnique = !this._model.some(
+      (condition) => condition !== currentCondition
+        && Array.isArray(condition)
+        && condition[0] === item.dataField,
+    );
+    return isDataFieldUnique ? item.dataField : item.name ? item.name : item.dataField;
```
siarhei-arzamasau: @tongsonbarbs please avoid nested ternary operators.

a bit more readable:
```
if (isDataFieldUnique) {
  return item.dataField;
}

return item.name ? item.name : item.dataField;
```

tongsonbarbs: Got it, thanks @siarhei-arzamasau

tongsonbarbs: https://github.com/DevExpress/DevExtreme/pull/28315/commits/1b85c6fdf74f75fce4ccdb1340307bbc9b483523

Raushen (approved):

PR #28343, opened 2024-11-12
Scheduler a11y: remove tables from a11y tree

pomahtri:


e2e/testcafe-devextreme/tests/scheduler/a11y/scheduler.ts
```diff
@@ -27,3 +27,21 @@ test('Scheduler should have right aria attributes', async (t) => {
...
   });
 });
+
+test('Scheduler table elements have right aria attributes', async (t) => {
+  const scheduler = new Scheduler('#container');
+
+  const tables = scheduler.element.find('table');
+  await t.expect(tables.count).eql(4);
+
+  for (let i = 0; i < await tables.count; i += 1) {
```
siarhei-arzamasau, line 37: Could you explain why do you use await before`tables.count`? Does it return Promise?
Also, what do you think to replace `for` loop on `for of` loop for better readability?

siarhei-arzamasau (dismissed):

e2e/testcafe-devextreme/tests/scheduler/a11y/scheduler.ts
pomahtri: Yes, it's a promise
`for of` loop is not supported by Selector, tsc gives `Type 'Selector' must have a '[Symbol.iterator]()' method that returns an iterator`

siarhei-arzamasau: got it. Yeah, for of works only with entities which implement iterable protocol

siarhei-arzamasau (approved):

Raushen (approved):

PR #28370, opened 2024-11-15
Scheduler - add label for edit appointment

pomahtri:

### DIFF
--- a/e2e/testcafe-devextreme/tests/scheduler/a11y/scheduler.ts
+++ b/e2e/testcafe-devextreme/tests/scheduler/a11y/scheduler.ts
@@ -76,9 +76,9 @@ test('Scheduler table elements have right aria attributes', async (t) => {
         currentView,
         currentDate: new Date(2021, 3, 29),
         startDayHour: 9,
-      })
+      });
     }).after(async () => {
       await changeTheme('generic.light');
-    });;
+    });
   });
 });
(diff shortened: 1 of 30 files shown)


e2e/testcafe-devextreme/tests/scheduler/a11y/scheduler.ts
```diff
@@ -45,3 +52,25 @@ test('Scheduler table elements have right aria attributes', async (t) => {
     currentView: 'month',
   });
 });
+
```
Raushen, line 56: Could you please add all themes?

packages/devextreme/js/localization/messages/ja.json
```diff
@@ -250,6 +250,7 @@
         "dxPivotGrid-filterFieldArea": "フィルター フィールドをこちらへドラッグ＆ドロップ",
 
         "dxScheduler-ariaLabel": "Scheduler. {0} view",
+
```
Raushen, line 253: I'd suggest move new string here instead of blank row

pomahtri: This new line appears only in generated file. In normal file (en.json), there's no blank rows

e2e/testcafe-devextreme/tests/scheduler/a11y/scheduler.ts
pomahtri: Contrast is disabled. Is there any need to check other themes?

Raushen: Yes. We have different predefined settings for themes so some of them can be course of issues.

packages/devextreme/js/localization/messages/ja.json
Raushen: Ok

Raushen (approved):

vladaskorohodova (approved):

PR #28388, opened 2024-11-18
List: Focus group header after interaction with header if collapsibleGroups: true, skip invisible items

marker-dao:


EugeniyKiyashko (dismissed):

ksercs (commented):
1. Items in collapsed groups are still focusable through kbn. It breaks "focus not obscured" a11y criteria
2. After group is collapsed, group header's id is updated (for some reason, looks incorrectly), but body's labelled-by is old. You can see an Axe report

marker-dao:
> 1. Items in collapsed groups are still focusable through kbn. It breaks "focus not obscured" a11y criteria

We put this bug in a separate one. Therefore, this issue has not been fixed in this PM. Now we have agreed that I will fix it here. Check pls: [1c73455fe3984156e398d0e239db2399eb64d663, 1fd50b01d94e520ba4be5cd8b91664a48e71b418].

> 2. After group is collapsed, group header's id is updated (for some reason, looks incorrectly), but body's labelled-by is old. You can see an Axe report

I fixed that problem. The explanation will be in our channel. [a6b2816cb2286be602a7a6d5e70464c54f1e0eac]

EugeniyKiyashko (dismissed):

Zedwag (dismissed):

EugeniyKiyashko (approved):

EugeniyKiyashko (approved):

ksercs (approved):

PR #28394, opened 2024-11-19
Menu: fix unable to unselect an menuitem (T1253750)

nikkithelegendarypokemonster:


Zedwag (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -1179,6 +1179,20 @@ QUnit.module('Menu - selection', {
         assert.equal($items.find('.' + DX_MENU_ITEM_TEXT_CLASS).text(), 'item3');
     });
 
+    QUnit.test('should be able to unselect current selected item (T1253750)', function(assert) {
```
EugeniyKiyashko:
```suggestion
    QUnit.test('should be able to unselect currently selected item (T1253750)', function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
```diff
@@ -1179,6 +1179,20 @@ QUnit.module('Menu - selection', {
...
+        const menu = createMenu({
+            items: [
+                { text: 'menu item 1', selectable: true },
+            ],
+        });
+        const item = $(menu.element).find('.' + DX_MENU_ITEM_CLASS).eq(0);
+
+        menu.instance.selectItem(item[0]);
```
EugeniyKiyashko, line 1193: I assume we should simplify this test and avoid checking how the selectItem method works.

1) If there are no tests verifying that the DX_MENU_ITEM_SELECTED_CLASS is added to a selected item when the selectItem method is called, let's create separate tests for it. Perhaps two separate tests should be added. (similar the test that I will add to the end of this message)
2) To configure the initial state of the item in the test, it's better to use the Item.selected property.
3) Instead of assert.ok(notOk), I would prefer using assert.strictEqual because there are more than two assertions in the test, and I want to understand exactly what each assertion verifies.
4) Don't forget to replace string concatenations with modern string literals. If we copy old tests, we should refactor them.

So, the test could look as follows:

```
QUnit.test('should be able to unselect the currently selected item (T1253750)', function(assert) {
    const menu = createMenu({
        items: [{
            text: 'Item_1',
            selectable: true,
            selected: true,
        }],
    });
    const item = $(menu.element).find(`.${DX_MENU_ITEM_CLASS}`).eq(0);

    assert.strictEqual(item.hasClass(DX_MENU_ITEM_SELECTED_CLASS), true, 'item has the selected class after initialization'); 

    menu.instance.unselectItem(item[0]);

    assert.strictEqual(item.hasClass(DX_MENU_ITEM_SELECTED_CLASS), false, 'item does not have the selected class after unselecting');
});
```

nikkithelegendarypokemonster: [59b9512](https://github.com/DevExpress/DevExtreme/pull/28394/commits/59b951296079f492dd4d8a786255364d461940fb)

nikkithelegendarypokemonster: added test for .selectItem since there is not test that its testing that method: [de5eda2](https://github.com/DevExpress/DevExtreme/pull/28394/commits/de5eda2ffa6e44404b033a3151847cfc09b6c0ad)

packages/devextreme/testing/tests/DevExpress.ui.widgets/menu.tests.js
nikkithelegendarypokemonster: [59b9512](https://github.com/DevExpress/DevExtreme/pull/28394/commits/59b951296079f492dd4d8a786255364d461940fb)

EugeniyKiyashko (approved):

Zedwag (approved):

PR #28415, opened 2024-11-22
DataGrid - Selected (Blanks) in the header filter should be applied if the Filter Row and Filter Panel is visible (T1257261)

markallenramirez:


siarhei-arzamasau (dismissed):

tongsonbarbs (dismissed):

e2e/testcafe-devextreme/tests/dataGrid/filterRow/filterRow.ts
```diff
@@ -9,6 +9,49 @@ import { getNumberData } from '../helpers/generateDataSourceData';
...
+  headerFilter: { visible: true },
+  filterRow: { visible: true },
+  filterPanel: { visible: true },
+  filterValue: ['Text', '=', 'i'],
+  columns: ['ID', {
+    dataField: 'Text',
+    selectedFilterOperation: '=',
+  }],
+  onEditorPreparing(e: any) {
+    e.updateValueTimeout = 100;
```
Alyar666, line 51: Why do you need this timeout?

e2e/testcafe-devextreme/tests/dataGrid/filterRow/filterRow.ts
```diff
@@ -9,6 +9,49 @@ import { getNumberData } from '../helpers/generateDataSourceData';
...
+    .eql('[Text] Equals \'i\'')
+    // act
+    .click(filterEditor.input)
+    .pressKey('backspace')
+    .wait(100) // updateValueTimeout
+    // assert
+    .expect(filterPanelText.element.textContent)
+    .eql('Create Filter')
+    // act
+    .click(dataGrid.element)
```
Alyar666, line 29: Why do you need this click?

e2e/testcafe-devextreme/tests/dataGrid/headerFilter/headerFilter.ts
```diff
@@ -10,6 +10,46 @@ fixture.disablePageReloads`Header Filter`
...
+  const list = headerFilter.getList();
+
+  await t.click(filterIconElement)
+    .click(list.getItem(1).element)
+    .click(buttons.nth(0));
+
+  result[0] = await dataCell.element().innerText;
+
+  await t.click(filterIconElement)
+    .click(filterIconElement)
```
Alyar666, line 31: Why do you need double click by filterIconElement?

e2e/testcafe-devextreme/tests/dataGrid/headerFilter/headerFilter.ts
```diff
@@ -10,6 +10,46 @@ fixture.disablePageReloads`Header Filter`
...
+  const dataGrid = new DataGrid(GRID_CONTAINER);
+  const headerCell = dataGrid.getHeaders().getHeaderRow(0).getHeaderCell(1);
+  const dataCell = dataGrid.getDataRow(0).getDataCell(0);
+  const filterIconElement = headerCell.getFilterIcon();
+  const headerFilter = new HeaderFilter();
+  const buttons = headerFilter.getButtons();
+  const list = headerFilter.getList();
+
+  await t.click(filterIconElement)
+    .click(list.getItem(1).element)
```
Alyar666: I suggest adding comments about what items you select/unselect. For example:
`.click(list.getItem(1).element) // Select second item with value 'Item 1'`

Alyar666 (changes requested):

e2e/testcafe-devextreme/tests/dataGrid/filterRow/filterRow.ts
markallenramirez: It's related to an issue I encountered with my initial solution. The use case is as follows:
1) User input in the filter row.
2) User clears the input via the backspace key.

There's a delay/timeout at this point before the new input is registered. After the timeout, the grid's filter is cleared.

3) User clicked outside the filter row editor (here's the reason why there's the last click). The focus is removed from the editor (blur event).

Since the editor's default `valueChangeEvent` includes 'change', the blur event triggers the editor's value change event. Unfortunately, at this point, the editor re-applies its old value.

This test case ensures that the same filter applied via input persists even after the 'change' event is triggered.

e2e/testcafe-devextreme/tests/dataGrid/filterRow/filterRow.ts
markallenramirez: Please see [this comment](https://github.com/DevExpress/DevExtreme/pull/28415/files#r1889765733).

e2e/testcafe-devextreme/tests/dataGrid/headerFilter/headerFilter.ts
markallenramirez: Good suggestion. Added.

e2e/testcafe-devextreme/tests/dataGrid/headerFilter/headerFilter.ts
markallenramirez: Not needed, nice catch! Fixed.

Alyar666 (approved):

PR #28424, opened 2024-11-25
Add external deps bundles for demos

GoodDayForSurf:


apps/demos/Demos/HtmlEditor/MarkdownSupport/Vue/index.html
```diff
@@ -10,26 +10,8 @@
     <script src="../../../../node_modules/typescript/lib/typescript.js"></script>
     <script type="module">
       import * as vueCompilerSFC from "../../../../node_modules/@vue/compiler-sfc/dist/compiler-sfc.esm-browser.js";
-
```
mpreyskurantov, line 13: please discard this diff, we have this copy pasted in all vue demos, it will be easier to remove it using identical occurrence / regex

apps/demos/configs/React/config.js
```diff
@@ -125,6 +116,15 @@ window.config = {
 
     'devextreme': 'npm:devextreme/cjs',
     'devextreme-react': 'npm:devextreme-react/cjs',
+
+    'unified': 'bundles:unified/unified.bundle.js',
```
mpreyskurantov: wrap with identical begin/end comment (like in other files) and local this section at same place of this file everywhere

mpreyskurantov (approved):

PR #28425, opened 2024-11-25
Chat: override box-sizing value on scrollable-content on ios to 'border-box'

nikkithelegendarypokemonster:


ksercs (approved):

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/messageList.tests.js
```diff
@@ -67,6 +69,17 @@ QUnit.module('MessageList', () => {
             assert.ok(this.instance instanceof MessageList);
         });
 
+        QUnit.test('scrollable content box-sizing value should be border-box', function(assert) {
+            if(devices.real().platform !== 'ios') {
```
EugeniyKiyashko, line 73: I assume that on all platforms, this style should have the 'border-box' value, so we don't need to add this condition. However, we emphasize that the issue was specific to iOS platforms only. I don't have any abjections to keep this condition.

EugeniyKiyashko (approved):

PR #28431, opened 2024-11-25
Header close button spike

Raushen:


packages/devextreme/js/__internal/grids/new/card_view/headers/header_item.tsx
```diff
@@ -1,19 +1,28 @@
...
 
 export function HeaderItem(props: HeaderItemProps): JSX.Element {
   return (
     <div className={CLASSES.headerItem}>
       {props.column.caption}
+      <Button
+          icon='close'
+          stylingMode='text'
+          elementAttr={{ class: CLASSES.button }}
+          onClick={(): void => { props.onRemoveButtonClicked?.(); }}
```
siarhei-arzamasau: should we explicitly define `void` type? looks like you can avoid it

packages/devextreme/js/__internal/grids/new/card_view/headers/headers.tsx
```diff
@@ -34,6 +36,9 @@ export class Headers extends Component<HeadersProps> {
           {this.props.columns.map((column) => (
             <HeaderItem
               column={column}
+              onRemoveButtonClicked={
+                (): void => this.props.onHeaderRemoveButtonClicked?.(column.name)
```
siarhei-arzamasau: similar question regarding `void`

packages/devextreme/js/__internal/grids/new/card_view/headers/view.tsx
```diff
@@ -25,6 +26,16 @@ export class HeadersView extends View {
     super();
   }
 
+  public onHeaderCloseButtonClick(name: string): void {
+    const index = this.getColumnIndexByName(name);
+
+    this.columnsController.columns.updateFunc((columns) => {
+      const newColumns = columns.slice();
```
siarhei-arzamasau: As I understood, you want to make a copy of array, if so than let's use more readable version with spread operator.
`columns.slice()` -> `[ ...columns ]`

packages/devextreme/js/__internal/grids/new/card_view/headers/header_item.tsx
pomahtri: It's linter rule
image

packages/devextreme/js/__internal/grids/new/card_view/headers/view.tsx
pomahtri: It's quite understandable too and used widely in grids code
image

pomahtri: If we want to use unified approach, let's discuss separately and apply strict rules to it (like linter)

siarhei-arzamasau: It's a question of the habbit or personal preferences but historically slice method was used as a quick workaround/hack to make a clone of array because there were no other eloquent ways in JS standard. For me it's like a "Hello from 2015".

Today, there are many ways how to make a copy of array:
- for loop
- slice
- Array.from
- spread operator

I suggest to use modern approachs like Array.from or spread operator and avoid hacks old approaches. Agree that we need to add a rule to the linter.

PR #28455, opened 2024-11-29
Process migration to jQuery 4

tongsonbarbs:


packages/devextreme/js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
```diff
@@ -641,7 +641,8 @@ const columnHeadersView = (Base: ModuleType<ColumnHeadersView>) => class ColumnH
     const $transparentColumnElement = that.getTransparentColumnElement();
     if (columnElements && $transparentColumnElement && $transparentColumnElement.length) {
       const transparentColumnIndex = getTransparentColumnIndex(that.getFixedColumns());
-      columnElements.splice(transparentColumnIndex, $transparentColumnElement.get(0).colSpan, $transparentColumnElement.get(0));
+      // @ts-expect-error
```
Raushen: I think we should not use @ts-expect-error here. It is better to use required types in the method. Also it is better to format the method. For example:
```
[].splice.apply(
        columnElements,
        [
          transparentColumnIndex,
          $transparentColumnElement.get(0)?.colSpan,
          $transparentColumnElement.get(0) as never,
        ],
      );
```

Raushen (approved):

PR #28458, opened 2024-11-29
Chat: add testcafe test for customization demo

nikkithelegendarypokemonster:


apps/demos/testing/widgets/chat/Customization.test.js
```diff
@@ -0,0 +1,130 @@
...
+    await t
+      .expect(compareResults.isValid())
+      .ok(compareResults.errorMessages());
+  });
+
+  test('Test dayHeaderFormat', async (t) => {
+    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+    const option = ['dd_MM_yyyy', 'dd.MM.yyyy', 'MMMM dd, yyyy', 'EEEE, MMMM dd'];
+
+    for (let index = 0; index < 4; index++) {
```
EugeniyKiyashko: I think that changing the value once is enough to check that the value change of the selectbox works

EugeniyKiyashko: The same for timestamp selectbox

apps/demos/testing/widgets/chat/Customization.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+    await t
+      .expect(compareResults.isValid())
+      .ok(compareResults.errorMessages());
+  });
+
+  test('Test showUsername', async (t) => {
+    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+    await testScreenshot(t, takeScreenshot, 'Chat when showUsername = true.png');
```
EugeniyKiyashko: I suggest leaving only the screenshot after changing the option, as the initial values of the fields are tested during demo initialization

apps/demos/testing/widgets/chat/Customization.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+fixture('Chat.Customization')
+  .page('http://localhost:8080/apps/demos/Demos/Chat/Customization/jQuery/')
+  .before(async (ctx) => {
+    ctx.initialWindowSize = [900, 600];
+  });
+
+// runManualTest('Chat', 'Customization', ['jQuery'/* , 'React', 'Vue', 'Angular' */], (test) => {
+  test('Sending Message', async (t) => {
+    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
```
EugeniyKiyashko, line 14: Can we have one test with the following chain of scenarios? 

hide avatar -> screenshot -> hide username -> screenshot -> change day header format -> screenshot -> hide day header format -> screenshot -> change timestamp format -> screenshot -> hide timestamp -> screenshot -> disable chat -> screenshot

apps/demos/testing/widgets/chat/Customization.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+  test('Test showDayHeaders', async (t) => {
+    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+    await testScreenshot(t, takeScreenshot, 'Chat when showDayHeaders = true.png');
+
+    await t
+        .click('#show-day-headers');
+
+    await testScreenshot(t, takeScreenshot, 'Chat when showDayHeaders = false.png');
```
EugeniyKiyashko: Let's stick to the format (notation) that we use for almost all scenario screenshot tests: widget_demoname_what_happened.png. I'm not sure why, but let's keep this format

apps/demos/testing/widgets/chat/Customization.test.js
```diff
@@ -0,0 +1,130 @@
...
+import { runManualTest } from '../../../utils/visual-tests/matrix-test-helper';
+import { testScreenshot } from '../../../utils/visual-tests/helpers/theme-utils';
+
+fixture('Chat.Customization')
+  .page('http://localhost:8080/apps/demos/Demos/Chat/Customization/jQuery/')
+  .before(async (ctx) => {
+    ctx.initialWindowSize = [900, 600];
+  });
+
+// runManualTest('Chat', 'Customization', ['jQuery'/* , 'React', 'Vue', 'Angular' */], (test) => {
```
EugeniyKiyashko: I think the variant with only the jQuery item in the array should work. When we add other demos (react, vue, angular), we will uncomment them.

After uncomment don't forget to patch pageUrl -> .page('http://localhost:8080/')

apps/demos/testing/widgets/chat/Customization.test.js
EugeniyKiyashko: done

EugeniyKiyashko (approved):

PR #28501, opened 2024-12-05
DataGrid - Canceled rows are hidden when multiple rows are added in batch mode (T1250405)

tongsonbarbs:


packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -80,6 +80,7 @@ interface Item {
   values?: unknown[];
   visible?: boolean;
   isExpanded?: boolean;
+  isNewRow: boolean;
```
Alyar666: I think, we need to mark this prop as being optional

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
```diff
@@ -1709,6 +1713,25 @@ class EditingControllerImpl extends modules.ViewController {
     return hasSavedData;
   }
 
+  private _updateUnsavedRowsInsertKey(tempKey, changes: any[]) {
+    const topSavedRowKey = this._getTopSavedRowKey();
+    if (!topSavedRowKey) return;
```
Alyar666: The problem will remain if there is no data in the grid

e2e/testcafe-devextreme/tests/dataGrid/editing/editingEvents.ts
```diff
@@ -172,3 +174,52 @@ testCases.forEach(({ caseName, expected, onRowRemoving }) => {
...
   });
 });
+
+// T1250405
+test('DataGrid - Canceled rows are hidden when multiple rows are added in batch mode', async (t) => {
+  const dataGrid = new DataGrid('#container');
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const addBtn = dataGrid.getToolbar().getItem();
+  const saveBtn = dataGrid.getToolbar().getItem(1);
+  await t
```
Alyar666, line 185: Please add a check that the grid is ready before starting all actions - `await t.expect(dataGrid.isReady()).ok()`.
This will increase the stability of the test.

e2e/testcafe-devextreme/tests/dataGrid/editing/editingEvents.ts
```diff
@@ -172,3 +174,52 @@ testCases.forEach(({ caseName, expected, onRowRemoving }) => {
...
+
+  const overlay1 = new Overlay(Selector('.dx-overlay-wrapper'), 0);
+  const overlay2 = new Overlay(Selector('.dx-overlay-wrapper'), 1);
+  const cancelBtn = overlay2.getToolbar(1).getItem(1);
+  const saveBtnPopup = overlay1.getToolbar(1).getItem(0);
+
+  await t
+    .click(cancelBtn)
+    .click(saveBtnPopup);
+  await t
```
Alyar666, line 199: Also, I recommend adding this check here too (before taking a screenshot). Since the data is refreshed when saving. And this is an asynchronous operation.

Alyar666 (changes requested):

Alyar666 (approved):

PR #28512, opened 2024-12-06
fix: framework demos fail when system.js faces some auto re-exports

VasilyStrelyaev:


apps/demos/utils/bundle/index.js
```diff
@@ -223,6 +224,9 @@ const prepareConfigs = (framework)=> {
                   fetch(load).then(r => resolve(r));
                   console.log('Unexpected transipling error (babel 7): ' + err);
                 } else {
+                  if(result.code.includes(GRID_COMMON_STAR_IMPORT)) {
+                    result.code = result.code.replace(GRID_COMMON_STAR_IMPORT, '');
+                  }
```
IlyaKhD: Extract this as a named function with a name reflecting its WA nature

IlyaKhD (dismissed):

IlyaKhD (approved):

PR #28513, opened 2024-12-06
Chat: added screenshot tests for chat overview demo

Strider2342:


Zedwag (dismissed):

apps/demos/testing/widgets/chat/Overview.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+runManualTest('Chat', 'Overview', ['jQuery', 'React', 'Vue', 'Angular'], (test) => {
+  test('Overview', async (t) => {
+    const styleElement = document.createElement('style');
+    styleElement.innerHTML = `.dx-chat-typingindicator-circle { animation: none !important; }`;  
+    document.head.appendChild(styleElement);
+
+    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+    await t.typeText('#user-chat .dx-texteditor-input', 'testing');
```
EugeniyKiyashko, line 24: It's better to extract selector in separate variables to avoid duplication

apps/demos/testing/widgets/chat/Overview.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+    await testScreenshot(t, takeScreenshot, 'chat_typing_indicator_appears_in_user_chat.png');
+
+    await t.typeText('#support-chat .dx-texteditor-input', 'testing');
```
EugeniyKiyashko: This approach will not work because IDs are not added t Chats in the all frameworks. We should use a common approach to determine a specific Chat element on the page, something like this
`
Selector('.dx-chat').nth(1)
`

apps/demos/testing/widgets/chat/Overview.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+    await testScreenshot(t, takeScreenshot, 'chat_typing_indicator_appears_in_support_chat.png');
+
+    await t
```
EugeniyKiyashko, line 28: Do you think that checking whether the message was entered and displayed in both Chats is excessive?

ksercs: It's possible that binding is broken for the second chat only, so why not? And it was stated in the initial card)
UPD: it's much less stable, typeText with Selectors does not trigger input event for some reason. So I've left single screenshot

apps/demos/testing/widgets/chat/Overview.test.js
ksercs: https://github.com/DevExpress/DevExtreme/pull/28513/commits/3bde3a32c992d44752bf0839478bdfc6e4cb44b3

apps/demos/testing/widgets/chat/Overview.test.js
ksercs: https://github.com/DevExpress/DevExtreme/pull/28513/commits/3bde3a32c992d44752bf0839478bdfc6e4cb44b3

Zedwag (dismissed):

PR #28515, opened 2024-12-06
fix: single entry point - add JS re-exports to old locations

VasilyStrelyaev:


packages/devextreme/js/events/core/events_engine.d.ts
```diff
@@ -0,0 +1,3 @@
+import eventsEngine from '../../common/core/events/core/events_engine';
+
+export default eventsEngine;
```
IlyaKhD, line 3:
```suggestion
export { eventsEngine as default } from '../../common/core/events/core/events_engine';
```

IlyaKhD (approved):

packages/devextreme/js/events/core/events_engine.d.ts
IlyaKhD: Discussed in person, eslint doesn't like it

PR #28522, opened 2024-12-09
DataGrid - Contrast between icons in the Filter Row menu and their background doesn't comply with WCAG accessibility standards (T1257970)

tongsonbarbs:


packages/testcafe-models/dataGrid/index.ts
```diff
@@ -217,13 +219,22 @@ export default class DataGrid extends Widget {
     return this.element.find(`.${this.addWidgetPrefix(CLASS.filterRow)}`);
   }
 
+  getFilterRowSearchIcon(columnIndex?): Selector {
```
Raushen, line 222: There is method getSearchIcon() already, you do not need to create one more

packages/testcafe-models/dataGrid/index.ts
```diff
@@ -217,13 +219,22 @@ export default class DataGrid extends Widget {
...
+
   getFilterCell(columnIndex: number): Selector {
     return this.getFilterRow().find(`[aria-colindex='${columnIndex + 1}']`);
   }
 
   getFilterRangeOverlay(): Selector {
     return this.body.find(`.${this.addWidgetPrefix(CLASS.filterRangeOverlay)}`);
   }
+
+  getFilterMenuOverlay(): Selector {
```
Raushen: Please use [the following approach](https://github.com/DevExpress/DevExtreme/blob/d37b38b6bfbdf8fa28f9d31278beaeea9d6090a1/e2e/testcafe-devextreme/tests/dataGrid/filterRow/filterRow.ts#L40) to work with filter menu

Raushen (changes requested):

packages/testcafe-models/dataGrid/index.ts
tongsonbarbs: Hi @Raushen , how do you mean by this? I see the link points to FilterEditor. Also to add, Filter Menu is found in an overlay wrapper which is outside the datagrid element..

packages/testcafe-models/dataGrid/filter/cell.ts
```diff
@@ -2,6 +2,7 @@ import FocusableElement from '../../internal/focusable';
 
 const CLASS = {
   filterMenu: 'dx-filter-menu',
+  filterMenuItem: 'dx-menu-item',
```
Raushen: You do not need changes in the testcafe model

packages/devextreme-scss/scss/widgets/generic/gridBase/_colors.scss
```diff
@@ -86,12 +86,6 @@ $datagrid-row-focused-bg: null !default;
 $datagrid-hover-bg: null !default;
 $datagrid-row-hovered-color: null !default;
 
-/**
-* $name 30. Context menu icon color
```
Raushen, line 90: We should not delete Theme Builder variables as them already became a part of public API. Project where these variables are used will be broken.

Shuvaevmedia (dismissed):

Raushen (approved):

PR #28535, opened 2024-12-11
Form: Hide opened DropDownEditor popup on label click in form (T1257945)

anna-shakhova:


packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
```diff
@@ -33,7 +59,9 @@ export function convertToRenderFieldItemOptions({
   labelMode,
   onLabelTemplateRendered,
 }) {
-  const isRequired = isDefined(item.isRequired) ? item.isRequired : !!_hasRequiredRuleInSet(item.validationRules);
+  const isRequired = isDefined(item.isRequired)
```
anna-shakhova, line 65: eslint fixed some styles here and after

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
```diff
@@ -153,10 +196,46 @@ function _convertToEditorOptions({
   const stylingMode = externalEditorOptions?.stylingMode || editorStylingMode;
   const useSpecificLabelOptions = EDITORS_WITH_SPECIFIC_LABELS.includes(editorType);
 
+  let editorOptionsWithDropDown: Partial<dxDropDownEditorOptions<dxTextBox>> = {};
```
anna-shakhova: used component the editor is based on, but still options types are unknown - what component is best to use here?

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
```diff
@@ -153,10 +196,46 @@ function _convertToEditorOptions({
...
+  const useDropDownOptions = DROP_DOWN_EDITORS.includes(editorType);
+
+  if (useDropDownOptions) {
+    editorOptionsWithDropDown = {
+      onContentReady: (e) => {
+        const { component } = e;
+        const openOnFieldClick = component.option('openOnFieldClick');
+        const initialHideOnOutsideClick = component.option('dropDownOptions.hideOnOutsideClick');
```
ksercs: I think it's okay to overwrite default internal hideOnOutsideClick if user specifies a custom one (not only through onContentReady, but always). I see dropDownEditor itself works this way now.

Cannot we just add a default implementation for dropDownOptions.hideOnOutsideClick, not using content ready?

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4540,6 +4541,98 @@ QUnit.test('form should be dirty when some editors are dirty', function(assert)
...
+                    colCount: 2,
+                    items: [{
+                        dataField: 'CustomerID',
+                        editorType: 'dxSelectBox',
+                        editorOptions: {
+                            items: ['VINET', 'VALUE', 'VINS'],
+                            value: '',
+                            openOnFieldClick,
+                            dropDownOptions: {
+                                hideOnOutsideClick: () => hideOnOutsideClickValue,
```
ksercs: [refactoring]
I suggest u renaming the parameter and pass a pure boolean here (and functions if u think it's important to be tested)
```suggestion
                                hideOnOutsideClick
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4540,6 +4541,98 @@ QUnit.test('form should be dirty when some editors are dirty', function(assert)
...
+                                onContentReady: ({ component }) => {
+                                    initialHideOnOutsideClick = component.option('hideOnOutsideClick');
+                                }
+                            }
+                        },
+                    }],
+                }],
+            });
+
+            const $dropDownButton = $form.find('.dx-dropdowneditor-button');
```
ksercs: [code style]
Please do not use classes explicitly, add a constant

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4540,6 +4541,98 @@ QUnit.test('form should be dirty when some editors are dirty', function(assert)
...
+            const hideOnOutsideClickSpy = sinon.spy(initialHideOnOutsideClick);
+            const formInstance = $form.dxForm('instance');
+            const editorInstance = formInstance.getEditor('CustomerID');
+
+            formInstance.option('items[0].items[0].editorOptions.dropDownOptions.hideOnOutsideClick', hideOnOutsideClickSpy);
+
+            assert.true(editorInstance.option('opened'), 'drop down list is visible');
+
+            const $label = $form.find('.dx-field-item-label-text');
+            pointerMock($label).click();
```
ksercs: Hm, does finally this click using `pointerMock` emulates real `label` behavior? I mean, doesn't it close the popup?
If it does, it's much better to omit hideOutsideClick spying and assert popup visibility state explicitly

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4540,6 +4541,98 @@ QUnit.test('form should be dirty when some editors are dirty', function(assert)
...
+
+            const expected = openOnFieldClick ? false : hideOnOutsideClickValue;
+            assert.ok(hideOnOutsideClickSpy.returned(expected), `hideOnOutsideClick returned ${expected}`);
+
+            sinon.restore();
+        });
+    });
+
+    QUnit.test(`Opened DropDownList must hide on input label click, openOnFieldClick: ${openOnFieldClick}, hideOnOutsideClick not defined (T1257945)`, function(assert) {
```
ksercs: [excess test?]
Hm, what's the point of this test? If `hideOnOutsideClick` is not specified explicitly, it takes a default value (true/false). So seems like your test above covers this case

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
```diff
@@ -153,10 +196,46 @@ function _convertToEditorOptions({
...
+
+        if (openOnFieldClick) {
+          component.option('dropDownOptions', {
+            hideOnOutsideClick: (e) => {
+              if (isBoolean(initialHideOnOutsideClick)) {
```
ksercs: [missed scenario?]
If initial value is true/false, why don't we consider label clicking as it's done for a case if initial value is a function?
p.s. if it's a missed scenario, kindly add a test

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
ksercs: UPD: got it, that's because of `openOnFieldClick`: in the hideOnOutsideClick handler you don't have information about editor's state

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
anna-shakhova: If hideOnOutsideClick is set to boolean value - then customer expects it always works the same way
Like in dxLookup it set to true and overwrites the default behavior for ddEditor. If we here add label check, then we'll get the same blinking on label click.
I moved applying functional hideOnOutsideClick after label check - because we discussed that it would be great to enrich customer's custom callback with our fix for the case

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
anna-shakhova: Unfortunately, it is not
But in each case we know by whom the popup should be closed - by label or input. So we can check the opened state.
it should be `false `when closed by label
and `true `when closed by triggered input

rewrote test, as it is much simpler without spy

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
anna-shakhova: The default value for hideOnOutsideClick is function, that checks whether ffEditor input was clicked. So this cases check default behavior

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
anna-shakhova: in that case initialHideOnOutsideClick we spy on is not a function and test fails on spy

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
anna-shakhova: option method returns unknown,

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
anna-shakhova: thanks, forgot before push

anna-shakhova:
[Form + DropDownBox - Opened DropDownBox opens unexpectedly again when the label area is clicked [T1257945]](https://trello.com/c/pB4kwQka/9470-form-dropdownbox-opened-dropdownbox-opens-unexpectedly-again-when-the-label-area-is-clicked-t1257945)

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4540,6 +4542,47 @@ QUnit.test('form should be dirty when some editors are dirty', function(assert)
...
+
+            assert.true(editorInstance.option('opened'), 'drop down list is visible');
+
+            const $label = $form.find(`.${FIELD_ITEM_LABEL_TEXT_CLASS}`);
+
+            pointerMock($label).click();
+
+            const willHideByInput = openOnFieldClick || (hideOnOutsideClick === false);
+
+            assert.equal(editorInstance.option('opened'), willHideByInput, `drop down list is hidden by ${willHideByInput ? 'triggered input click' : 'label click'}`);
```
ksercs: [minor, more jfyi]
We try to always use `strictEqual` instead of `equal`. It's like `==` vs `===` - just for safety :)

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4540,6 +4542,47 @@ QUnit.test('form should be dirty when some editors are dirty', function(assert)
...
+            const editorInstance = $form.dxForm('instance').getEditor('CustomerID');
+
+            assert.true(editorInstance.option('opened'), 'drop down list is visible');
+
+            const $label = $form.find(`.${FIELD_ITEM_LABEL_TEXT_CLASS}`);
+
+            pointerMock($label).click();
+
+            const willHideByInput = openOnFieldClick || (hideOnOutsideClick === false);
+
```
ksercs, line 4580: NOTE: In the real environment, clicking the label triggers a click on the editor, toggling the popup visibility if openOnFieldClick=true. This assertion only takes hideOnOutsideClick into account.

ksercs: Let's add a comment like this just for future us
```
NOTE: In the real environment, clicking the label triggers a click on the editor, toggling the popup visibility if openOnFieldClick=true. This assertion only takes hideOnOutsideClick into account.
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
ksercs:
```suggestion
            assert.strictEqual(editorInstance.option('opened'), willHideByInput, `drop down list is hidden by ${willHideByInput ? 'triggered input click' : 'outside click'}`);
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4540,6 +4542,47 @@ QUnit.test('form should be dirty when some editors are dirty', function(assert)
...
+            const editorInstance = $form.dxForm('instance').getEditor('CustomerID');
+
+            assert.true(editorInstance.option('opened'), 'drop down list is visible');
+
+            const $label = $form.find(`.${FIELD_ITEM_LABEL_TEXT_CLASS}`);
+
+            pointerMock($label).click();
+
+            const willHideByInput = openOnFieldClick || (hideOnOutsideClick === false);
```
ksercs: I didn't get the second part of this expression.
Only `openOnFieldClick` says will be popup closed because of "label + editor" click or not.
Seems like you write this code just for the correct assertion below.
Can we make this part more clear?

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
```diff
@@ -153,10 +197,48 @@ function _convertToEditorOptions({
   const stylingMode = externalEditorOptions?.stylingMode || editorStylingMode;
   const useSpecificLabelOptions = EDITORS_WITH_SPECIFIC_LABELS.includes(editorType);
 
+  let editorOptionsWithDropDown: Partial<dxDropDownEditorOptions<dxTextBox>> = {};
+  const useDropDownOptions = DROP_DOWN_EDITORS.includes(editorType);
+
+  if (useDropDownOptions) {
```
ksercs: [refactoring / code splitting]
I suggest you moving this new code to a separate method with an `editorType` parameter.
It'll simplify reading of the main function and will allow us to get of this condition

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
anna-shakhova: fixed

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
anna-shakhova: fixed

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
anna-shakhova: fixed

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
anna-shakhova: moved

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
```diff
@@ -122,8 +167,55 @@ export function convertToLabelMarkOptions({
   };
 }
 
+// eslint-disable-next-line @typescript-eslint/naming-convention
+function _getDropDownEditorOptions(
```
ksercs, line 171: Actually no need to add an underscore here while it's not a method but a function. It's clear that this function will not be used from external while it's not exported
```suggestion
function getDropDownEditorOptions(
```

ksercs (approved):

packages/devextreme/js/__internal/ui/form/m_form.layout_manager.utils.ts
anna-shakhova: I've added underscore as many other functions (like `_convertToEditorOptions`) in this file are named that way. I thought it is like a standard

Zedwag (approved):

PR #28536, opened 2024-12-11
Chat AI Demo: Add model to params, fix import in Vue demo

Zedwag:


apps/demos/Demos/Chat/AIAndChatbotIntegration/React/App.tsx
```diff
@@ -35,7 +35,7 @@ async function getAIResponse(messages) {
     temperature: 0.7,
   };
 
-  const response = await chatService.chat.completions.create(params);
+  const response = await chatService.chat.completions.create(params as any);
```
anna-shakhova: I think we should avoid 'any' and assertions as untipatterns in our demos
We can define params using required type, add examples with some artificial type - you can change it for the required one 

const params: ChatServiceCreateParams = {
    messages,
    max_tokens: 1000,
    temperature: 0.7,
  };

Zedwag: i did it to avoid required `model` passing
image
but ok i guess it's ok to use the one we use in service config

apps/demos/Demos/Chat/AIAndChatbotIntegration/React/App.tsx
```diff
@@ -31,6 +31,7 @@ const chatService = new AzureOpenAI(AzureOpenAIConfig);
 async function getAIResponse(messages) {
   const params = {
     messages,
+    model: AzureOpenAIConfig.deployment,
```
marker-dao, line 34: Why did you add that?

ksercs, line 34: don't we want to apply this in all frameworks to get rid of ts warnings everywhere?

apps/demos/Demos/Chat/AIAndChatbotIntegration/Vue/App.vue
```diff
@@ -76,6 +76,8 @@ const copyButtonIcon = ref('copy');
 
 onBeforeMount(() => {
   loadMessages(dictionary);
+
+  chatService = new AzureOpenAI(AzureOpenAIConfig);
```
ksercs: I've checked this change on az-server and seems like it does not resolve the issue.
However, replacing current import with `import openai from 'openai';` helps.
So I suggest you to revert this change but update the import

apps/demos/Demos/Chat/AIAndChatbotIntegration/React/App.tsx
marker-dao: okay, got it (cuz ts)

marker-dao: [b83ec79eabe4b3dfb5b357bd0ea9b4bfe39a0859]

apps/demos/Demos/Chat/AIAndChatbotIntegration/Vue/App.vue
marker-dao: [b83ec79eabe4b3dfb5b357bd0ea9b4bfe39a0859]

apps/demos/Demos/Chat/AIAndChatbotIntegration/Vue/App.vue
```diff
@@ -53,10 +53,9 @@ import { ref, onBeforeMount } from 'vue';
 import DxChat from 'devextreme-vue/chat';
 import DxButton from 'devextreme-vue/button';
 import { loadMessages } from 'devextreme/localization';
-import { AzureOpenAI } from 'openai';
+import openai from 'openai';
 import {
   dictionary,
-  store,
```
ksercs, line 59: Is this change intentional?

apps/demos/Demos/Chat/AIAndChatbotIntegration/Vue/data.ts
```diff
@@ -26,8 +26,7 @@ export const CHAT_DISABLED_CLASS = 'dx-chat-disabled';
 export const ALERT_TIMEOUT = 1000 * 60;
 
 export const user = {
-  id: 'c94c0e76-fb49-4b9b-8f07-9f93ed93b4f3',
-  name: 'John Doe',
+  id: 'user',
```
ksercs, line 29: For what?

apps/demos/Demos/Chat/AIAndChatbotIntegration/Vue/App.vue
marker-dao: Yes, store is not used anywhere, its import was an oversight.

apps/demos/Demos/Chat/AIAndChatbotIntegration/Vue/data.ts
marker-dao: Look at the other demos (jQuery for example): there's different data there, it just wasn't copied correctly.

ksercs (approved):

PR #28595, opened 2024-12-19
Core: renderer, `removeAttr` remove an attribute from a set of elements (T1261932)

chaosmirage:
To reproduce the case in versions `>=24.2` of the devextreme package the `legacyMode` should be enabled in the DataGrid config.

```
  columnFixing: {
      legacyMode: true,
  },
```

### DIFF
--- a/packages/devextreme/js/core/renderer.d.ts
+++ b/packages/devextreme/js/core/renderer.d.ts
@@ -36,6 +36,14 @@ export interface dxElementWrapper {
 
   detach(): this;
 
+  /**
+   * Iterates over each element in the set of matched elements, executing the provided function for each element.
+   *
+   * @param {(this: Element, index: number, element: Element) => boolean} func - A function to execute for each matched element.
+   *        The function receives the index position of the element in the set and the element itself as arguments.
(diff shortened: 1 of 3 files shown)


packages/devextreme/js/core/renderer.d.ts
```diff
@@ -36,6 +36,14 @@ export interface dxElementWrapper {
...
   detach(): this;
 
+  /**
+   * Iterates over each element in the set of matched elements, executing the provided function for each element.
+   *
+   * @param {(this: Element, index: number, element: Element) => boolean} func - A function to execute for each matched element.
+   *        The function receives the index position of the element in the set and the element itself as arguments.
+   *        If the function returns `false`, the iteration will stop.
+   * @returns {this} The current instance for chaining.
+   */
```
pomahtri: Let's not add doc-comments in public .d.ts files. They might interfere with metadata generations, and they will require approve from api reviewers & tech writers

chaosmirage: removed

VasilyStrelyaev (approved):

dxvladislavvolkov (approved):

PR #28614, opened 2024-12-23
List, TagBox, Splitter: localize aria-roledescription attribute (T1267579)

anna-shakhova:


packages/devextreme/js/__internal/ui/splitter/resize_handle.ts
```diff
@@ -305,7 +305,7 @@ class ResizeHandle extends Widget<ResizeHandleOptions> {
     this.setAria({
       role: 'application',
       // eslint-disable-next-line spellcheck/spell-checker
-      roledescription: 'separator',
+      roledescription: messageLocalization.format('dxSplitter-ariaRoleDescription'),
```
ksercs: While we're localizing `separator` not a "splitter" itself I think we should show it explicitly in the variable's name. Maybe smth like this?
```suggestion
      roledescription: messageLocalization.format('dxSplitter-separatorAriaRoleDescription'),
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/list.markup.tests.js
```diff
@@ -278,6 +278,9 @@ if(devices.real().deviceType === 'desktop') {
                         items: [{ text: 'Item_1' }, { text: 'Item_2' }, { text: 'Item_3' }],
                     }, options))
             });
+            const localizedRoleDescription = 'list';
```
ksercs: If you trying to check that custom value can be loaded to the dictionary, let's use move explicit custom value
```suggestion
            const localizedRoleDescription = 'My Custom List';
```

packages/devextreme/js/__internal/ui/splitter/resize_handle.ts
anna-shakhova: renamed

packages/devextreme/testing/tests/DevExpress.ui.widgets/list.markup.tests.js
anna-shakhova: fixed

packages/devextreme/js/localization/messages/en.json
```diff
@@ -372,6 +373,7 @@
         "dxTagBox-allSelected": "All selected ({0})",
         "dxTagBox-moreSelected": "{0} more",
         "dxTagBox-tagRoleDescription": "Tag. Press the delete button to remove this tag",
+        "dxTagBox-ariaRoleDescription": "Tagbox",
```
```suggestion
        "dxTagBox-ariaRoleDescription": "Tag box",
```

anna-shakhova: fixed

vladaskorohodova (approved):

ksercs (approved):

PR #28626, opened 2024-12-24
fix: optimize React template discover renders

VasilyStrelyaev:


packages/devextreme-react/src/core/use-option-scanning.ts
```diff
@@ -75,12 +78,11 @@ export function useOptionScanning(
   };
 
   useLayoutEffect(() => {
-    if (!templateContainer) {
-      return;
-    }
-
-    const hasTemplateRendered = templateContainer.childNodes.length > 0;
+    const container = templateContainer && 'current' in templateContainer
```
ivanblinov2k17, line 81: Minor - maybe use optional chaining here, smth like
const container = templateContainer?.current || templateContainer

ivanblinov2k17 (dismissed):

packages/devextreme-react/src/core/use-option-scanning.ts
VasilyStrelyaev: `templateContainer?.current` throws a TypeScript error, so `'current' in templateContainer` looks like _the_ valid way to narrow the type here

chaosmirage (dismissed):

packages/devextreme-react/src/core/component-base.tsx
```diff
@@ -424,18 +429,14 @@ const ComponentBase = forwardRef<ComponentBaseRef, any>(
...
-          { templateContainer
-            && createPortal(
-              <TemplateDiscoveryContext.Provider value={{ discoveryRendering: true }}>
-                {_renderChildren()}
-              </TemplateDiscoveryContext.Provider>,
-              templateContainer,
-            )
-          }
-          <div {...getElementProps()}>
+        <TemplateRenderingContext.Provider value={renderContextValue}>
```
chaosmirage, line 438: Here's an abstract example in isolation:
 
https://codesandbox.io/p/sandbox/7c597q
 
If you click the "Trigger Unnecessary Render" button, there will be redraws. 
 
If you put `{ isTemplateRendering: false }` into a module-level constant, there will be no redraws. 
 
I suggest to put { isTemplateRendering: false } into constants or memoize it to avoid such side effects (unless it is done on purpose).

chaosmirage (approved):

PR #28642, opened 2024-12-26
Fix export in chart demo (T1266498)

GoodDayForSurf:


apps/demos/Demos/Charts/ExportCustomMarkup/Angular/app/app.component.html
```diff
@@ -1,6 +1,14 @@
 <div class="chart_environment">
   <div id="custom_markup_container">
     <svg version="1.1" width="820px" height="420px">
+      <path
```
mpreyskurantov, line 4: Why do you change `path` tag location within svg (all frameworks)?

mpreyskurantov: Btw, the `opacity` style is removed

GoodDayForSurf: actually this <path  is background for chart so before replacement exported image had black background

apps/demos/package.json
```diff
@@ -28,6 +28,7 @@
         "@rollup/plugin-commonjs": "19.0.2",
         "@vue/compiler-sfc": "3.4.27",
         "@vue/shared": "^3.2.47",
+        "canvg": "4.0.2",
```
mpreyskurantov, line 31: What canvg version used before this?

apps/demos/configs/Vue/config.js
```diff
@@ -95,9 +95,7 @@ window.config = {
     /**/
 
     /** canvg */
-    'stackblur-canvas': 'npm:stackblur-canvas/dist/stackblur.min.js',
-    'rgbcolor': 'npm:rgbcolor/index.js',
-    'canvg': 'npm:canvg/dist/browser/canvg.min.js',
+    'canvg': 'externals:canvg.bundle.js',
```
mpreyskurantov, line 98: How the other deps were used before?

apps/demos/package.json
GoodDayForSurf: It seems last working version was < 3

apps/demos/configs/Vue/config.js
GoodDayForSurf: I don't know full story but some deps was used like rgbcolor, now I think it's better to keep them in bundle

mpreyskurantov (dismissed):

mpreyskurantov (approved):

PR #28660, opened 2025-01-02
DataGrid - The "Cannot read properties of undefined error" occurs when using Tab while saving a promise (T1190566)

tongsonbarbs:


packages/devextreme/js/__internal/grids/grid_core/validating/m_validating.ts
```diff
@@ -844,7 +844,7 @@ export const validatingEditingExtender = (Base: ModuleType<EditingController>) =
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   protected _beforeSaveEditData(change, editIndex?) {
     let result: any = super._beforeSaveEditData.apply(this, arguments as any);
-    const validationData = this._validatingController._getValidationData(change?.key);
+    const validationData = this._validatingController._getValidationData(change?.key, true);
 
     if (change) {
       const isValid = change.type === 'remove' || validationData.isValid;
```
Tucchhaa, line 850: I've noticed that when we tab to different cell after editing previous, `_beforeSaveEditData` is called with `change` parameter as undefined. So no validation data is found.

But there's already an if-statement that checks if `change` is not undefined. Maybe then we can move `_getValidationData` into it? And we don't need to pass `create` parameter. I tested it and it seems to work.

```suggestion
    if (change) {
      const validationData = this._validatingController._getValidationData(change.key);
      const isValid = change.type === 'remove' || validationData.isValid;
```

What do you think?

Alyar666 (approved):

PR #28674, opened 2025-01-07
NX - dte spike

pomahtri:


.github/workflows/default_workflow.yml
```diff
@@ -53,6 +63,36 @@ jobs:
...
+    #       if ! wait $pid; then
+    #         exit 1  # exit with an error status if any process fails
+    #       fi
+    #     done
+
+    #     exit 0 # exits with success status if a all processes complete successfully
+
     - name: Run targets
       run: >
         pnpx nx run-many
```
IlyaKhD:
```suggestion
        pnpm nx run-many
```
nx is already installed at this point

PR #28687, opened 2025-01-08
T1261532: DataGrid - FocusedRowChanged event isn't raised when the push API is used to remove the last row

Raushen:


e2e/testcafe-devextreme/tests/dataGrid/focus/focus.ts
```diff
@@ -241,6 +241,41 @@ test('DataGrid - FilterRow cell loses focus when focusedRowEnabled is true and e
   });
 });
 
+test('DataGrid - FocusedRowChanged event isnt raised when the push API is used to remove the last row (T1261532)', async (t) => {
+  await t
+    .expect(Selector('#otherContainer').innerText)
+    .eql('Success');
```
Tucchhaa: I feel strange about the checking the text of #otherContainer. Is it possible to use this instead of setting text in onFocusedRowChanged? 
```suggestion
const key = await grid.option('focusedRowKey');
const index = await grid.option('focusedRowIndex');

await t.expect(key === null && index === -1).ok();
```

Raushen: TestCafe is designed to work with markup, and in our tests, we use the [Page Object Model for the DataGrid](https://github.com/DevExpress/DevExtreme/blob/c3ae57e7ae8d9738ada90b58b9bd62e245d676d4/packages/testcafe-models/dataGrid/index.ts#L129), which operates with a markup (CCS classes and DOM elements) and does not have a method for retrieving options from the business object. It seems that transforming the state into markup and verifying it using TestCafe is the most appropriate and natural approach.

Tucchhaa: https://github.com/Tucchhaa/DevExtreme/blob/c3ae57e7ae8d9738ada90b58b9bd62e245d676d4/e2e/testcafe-devextreme/tests/dataGrid/scrolling.ts#L996-L999

https://github.com/Tucchhaa/DevExtreme/blob/c3ae57e7ae8d9738ada90b58b9bd62e245d676d4/packages/testcafe-models/internal/widget.ts#L73

Seems that we can retrieve options from the datagrid.

Tucchhaa: the current test implementation does its job, but it just feels unnaturally to me to do it like that. If using grid.option() will also do the job, then I suggest to use it 🙂

Raushen: Yep, finally I see, thank you! The correct check looks slightly different)
The whole expression should be wrapped to the "except" method so the expression could calculate continuously, not only once while a declaration.

Tucchhaa (approved):

PR #28691, opened 2025-01-09
TagBox: re-render tags when dataSource is updated (T1253312)

nikkithelegendarypokemonster:


packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1592,6 +1609,10 @@ const TagBox = (SelectBox as any).inherit({
         this.$element().toggleClass(TAGBOX_SINGLE_LINE_CLASS, !value);
         this._renderSingleLineScroll();
         break;
+      case 'dataSource':
```
anna-shakhova, line 1612: here we lost default behavior from parent class which set new dataSource to editor and list, it also forced tags rerender
So, there is no need to rewrite the whole function, just add some logic to `_setListDataSource` to support `_selectedItems`, and, perhaps, `value` update (later it will be set to list options in `_refreshSelected`)

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6693,6 +6693,58 @@ QUnit.module('dataSource integration', moduleSetup, () => {
...
+            this.$tagBox = $('#tagBox').dxTagBox({
+                dataSource: this.data,
+                value: [1, 2],
+                displayExpr: 'Name',
+                valueExpr: 'ID',
+            });
+            this.tagBox = this.$tagBox.dxTagBox('instance');
+        }
+    }, () => {
+        QUnit.test('Tagbox should only render existing tags (T1253312)', function(assert) {
```
anna-shakhova: this test almost duplicates the next one, in current case there hardly ever will present issue in markup, if options are valid, so. I think. we can omit this test

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6693,6 +6693,58 @@ QUnit.module('dataSource integration', moduleSetup, () => {
...
+            this.tagBox.option('dataSource', this.data.slice(0, 1));
+
+            assert.strictEqual(this.$tagBox.find(`.${TAGBOX_TAG_CLASS}`).length, 1, 'Only one tag is rendered after the dataSource is reduced.');
+            assert.strictEqual(this.$tagBox.find(`.${TAGBOX_TAG_CLASS}`).eq(0).text(), 'HD Video Player', 'Rendered tags match the updated dataSource.');
+        });
+
+        QUnit.test('Tagbox updates value and selectedItems (T1253312)', function(assert) {
+            this.tagBox.option('dataSource', this.data.slice(0, 1));
+
+            assert.deepEqual(this.tagBox.option('selectedItems'), this.data.slice(0, 1), 'selectedItems match the updated dataSource.');
```
anna-shakhova: Here and after, remove dots in assertions, please

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6693,6 +6693,58 @@ QUnit.module('dataSource integration', moduleSetup, () => {
...
+            assert.deepEqual(this.tagBox.option('value'), [1], 'value is updated based on the new dataSource.');
+        });
+
+        QUnit.test('Tagbox shouldnt render any tags if dataSource is updated to either null or empty (T1253312)', function(assert) {
```
anna-shakhova: Again, I think. we can omit this case, as it is covered with the next one

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6693,6 +6693,58 @@ QUnit.module('dataSource integration', moduleSetup, () => {
...
+            assert.strictEqual(this.$tagBox.find(`.${TAGBOX_TAG_CLASS}`).length, 0, 'No tags are rendered after null dataSource');
+        });
+
+        QUnit.test('Tagbox should update its selectedItems and value to empty if dataSource is updated to empty or null (T1253312)', function(assert) {
+            this.tagBox.option('dataSource', []);
+
+            assert.deepEqual(this.tagBox.option('selectedItems'), [], 'selectedItems is updated to empty array when dataSource is empty');
+            assert.deepEqual(this.tagBox.option('value'), [], 'value is updated to empty array when dataSource is empty');
+
+            this.tagBox.option('dataSource', null);
```
anna-shakhova, line 6735: let's split the test to two - in first we set [] dataSouce, in second - null, for now there is no change in component after second change

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -6693,6 +6693,58 @@ QUnit.module('dataSource integration', moduleSetup, () => {
...
+    }, () => {
+        QUnit.test('Tagbox should only render existing tags (T1253312)', function(assert) {
+            this.tagBox.option('dataSource', this.data.slice(0, 1));
+
+            assert.strictEqual(this.$tagBox.find(`.${TAGBOX_TAG_CLASS}`).length, 1, 'Only one tag is rendered after the dataSource is reduced.');
+            assert.strictEqual(this.$tagBox.find(`.${TAGBOX_TAG_CLASS}`).eq(0).text(), 'HD Video Player', 'Rendered tags match the updated dataSource.');
+        });
+
+        QUnit.test('Tagbox updates value and selectedItems (T1253312)', function(assert) {
+            this.tagBox.option('dataSource', this.data.slice(0, 1));
```
anna-shakhova, line 6714: let's add assertion that before dataSource is changed, we had 2 selectedItems
same for the later tests

anna-shakhova (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
nikkithelegendarypokemonster: [709c69b](https://github.com/DevExpress/DevExtreme/pull/28691/commits/709c69b71c6126e545de23a9579bd33ab53c2e79)

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
nikkithelegendarypokemonster: [709c69b](https://github.com/DevExpress/DevExtreme/pull/28691/commits/709c69b71c6126e545de23a9579bd33ab53c2e79)

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
nikkithelegendarypokemonster: [709c69b](https://github.com/DevExpress/DevExtreme/pull/28691/commits/709c69b71c6126e545de23a9579bd33ab53c2e79)

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
nikkithelegendarypokemonster: [709c69b](https://github.com/DevExpress/DevExtreme/pull/28691/commits/709c69b71c6126e545de23a9579bd33ab53c2e79)

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
nikkithelegendarypokemonster: [709c69b](https://github.com/DevExpress/DevExtreme/pull/28691/commits/709c69b71c6126e545de23a9579bd33ab53c2e79)

packages/devextreme/js/__internal/ui/m_tag_box.ts
nikkithelegendarypokemonster: hello Anna i replicated your issue yes dataSource is not updating in _list and editor can you try my solution here using callBase(args) instead since in my opinion i don't want to use setListDataSource since we are changing selectedItems and value options not on _list but on tag box level instead so i like to use a separate method for it.... can you check my solution as well... thank you

nikkithelegendarypokemonster: [709c69b](https://github.com/DevExpress/DevExtreme/pull/28691/commits/709c69b71c6126e545de23a9579bd33ab53c2e79)

anna-shakhova: we should keep editor and list options synchronized, in `_refreshSelected,` `se;ectedItems` list option is set to be equal to `this,_selectedItems`, so we should update the last one as well
i suggested to combine syncLogic with `_setListDataSource`, to avoid setting value twice

PR #28700, opened 2025-01-10
DataGrid - Gray boxes appear when the push method is used to remove rows in infinite scrolling mode (T1240079)

tongsonbarbs:


packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
```diff
@@ -792,8 +792,8 @@ export class RowsView extends ColumnsView {
     }, options));
 
     that._checkRowKeys(options.change);
-
     that._renderFreeSpaceRow($table, options.change);
+
     if (!that._hasHeight) {
```
Tucchhaa: It's better to not change files, that have nothing to do with the fix. Let's remove these changes

packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts
```diff
@@ -1613,6 +1613,10 @@ export const rowsView = (Base: ModuleType<RowsView>) => class VirtualScrollingRo
       .viewportItemSize(rowHeight);
 
     if (isVirtualMode(this) || gridCoreUtils.isVirtualRowRendering(this)) {
+      if (this._dataController.isEmpty()) {
+        return;
+      }
```
Tucchhaa: Minor suggestion: move the expression inside if statement into a variable
```suggestion
      const skipRenderEmptyRows = this._dataController.isEmpty();
      
      if (skipRenderEmptyRows) {
        return;
      }
```

e2e/testcafe-devextreme/tests/dataGrid/scrolling.ts
```diff
@@ -1720,3 +1720,47 @@ test('The row alternation should display correctly when grouping and virtual scr
   grouping: { autoExpandAll: true },
   scrolling: { mode: 'virtual', useNative: false },
 })));
+
+test('DataGrid - Gray boxes appear when the push method is used to remove rows in infinite scrolling mode (T1240079)', async (t) => {
+  const dataGrid = new DataGrid('#container');
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  await t
+    .click(dataGrid.getDataCell(0, 0).element)
```
Tucchhaa: Let's use .apiPush() method instead of clicking for deleting items

```suggestion
  await dataGrid.apiPush([
    { type: 'remove', key: 1 },
    { type: 'remove', key: 2 },
  ]);
```

Tucchhaa (changes requested):

Tucchhaa (approved):

PR #28709, opened 2025-01-14
Form: DropDownBox loses its value after the Form is resized (T1196835)

EugeniyKiyashko:


packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.tests.js
```diff
@@ -4916,39 +4954,124 @@ QUnit.module('reset', () => {
...
+                            editorType,
+                        }],
+                    }],
+                }],
+            }).dxForm('instance');
+
+            const dropDownEditor = form.getEditor('name');
+
+            screen = 'sm';
+            dropDownEditor.focus();
```
ksercs, line 5032: Is it expected that ddEditor is focused in this test? I thought that we want to focus tabPanel here to text scenario different from the previous test

ksercs (approved):

Zedwag (approved):

PR #28758, opened 2025-01-17
DataGrid demos - hide context menu in table in dropdown

Ambrozy:

### DIFF
--- a/apps/demos/Demos/DataGrid/CustomEditors/Angular/app/app.component.ts
+++ b/apps/demos/Demos/DataGrid/CustomEditors/Angular/app/app.component.ts
@@ -58,7 +58,7 @@ export class AppComponent {
       },
     });
   }
-  
+
   getSelectedRowKeys<T>(value: T): T[] {
     return value !== null && value !== undefined ? [value] : [];
   }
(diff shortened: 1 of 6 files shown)


apps/demos/Demos/DataGrid/CustomEditors/Angular/app/app.component.ts
```diff
@@ -58,7 +58,7 @@ export class AppComponent {
       },
     });
   }
-
```
tongsonbarbs, line 61: Let's remove this change since it has nothing to do with the fix. There is a possibility that it may cause confusion for Git History that this PR is the last touch. Other developers might assume that there could be a change in this file in accordance to this PR's purpose.

apps/demos/Demos/DataGrid/CustomEditors/React/EmployeeDropDownBoxComponent.tsx
```diff
@@ -12,7 +12,7 @@ const dropDownOptions = { width: 500 };
 const ownerLabel = { 'aria-label': 'Owner' };
 
 const EmployeeDropDownBoxComponent = (props) => {
-  const {data: { value: dataValue }} = props;
+  const { data: { value: dataValue } } = props;
```
tongsonbarbs, line 15: You may revert this line too. Not a big deal but for the same reason in the first comment.

apps/demos/Demos/DataGrid/CustomEditors/Angular/app/app.component.ts
pomahtri: I agree that it's isn't really connected to PR, but we don't really have rule to omit this type of changes, and it might be painful to revert it and trigger CI again

apps/demos/Demos/DataGrid/CustomEditors/React/EmployeeDropDownBoxComponent.tsx
pomahtri: see 1st comment

pomahtri (approved):

PR #28803, opened 2025-01-21
Lookup: Fix load indicator when minSearchLength specified

anna-shakhova:


packages/devextreme/js/__internal/ui/m_lookup.ts
```diff
@@ -811,6 +811,21 @@ const Lookup = DropDownList.inherit({
     }
   },
 
+  _presetListDataSource() {
+    if (this._list && this._list?.option('dataSource') && !this._needPassDataSourceToList()) {
```
EugeniyKiyashko: It's unclear why is it necessary to use ?. on this line but not on line 819.

ksercs: Ofc optional chaining is never useful here - we should use it either `this._list &&` but not together.
I personally prefer optional chaining as a more modern and short version

ksercs (approved):

EugeniyKiyashko (approved):

PR #28842, opened 2025-01-27
CardView - implement headerPanel

pomahtri:


packages/devextreme/js/__internal/grids/new/card_view/header_panel/column_sortable.tsx
```diff
@@ -0,0 +1,139 @@
...
+      x: e.event.clientX,
+      y: e.event.clientY,
+    };
+
+    const yDistance = Math.min(
+      Math.abs(dragCoords.y - containerCoords.y),
+      Math.abs(dragCoords.y - containerCoords.y + containerCoords.height),
+    );
+
+    if (yDistance <= 64) {
```
Raushen: Magic constant

pomahtri: I added `TODO` comment for next PR (with implementing styles)
I think it should be css variable (because it depends on the theme)

packages/devextreme/js/__internal/grids/new/card_view/header_panel/options.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+      expect(rootElement).toMatchSnapshot();
+    });
+  });
+  describe('headerItemCssClass', () => {
+    it('should override content of headerPanel item', () => {
+      const { rootElement } = setup({
+        columns: [{
+          dataField: 'column1',
+          headerItemCssClass: 'my-css-class',
```
Raushen, line 143: To discuss

packages/devextreme/js/__internal/grids/new/card_view/header_panel/options.test.ts
```diff
@@ -0,0 +1,150 @@
...
+          columns: ['column1'],
+          headerPanel: {
+            // @ts-expect-error
+            itemTemplate: ({ column }) => $('<div>')
+              .addClass('my-class')
+              .text(column.caption),
+          },
+        });
+
+        expect(rootElement).toMatchSnapshot();
```
Raushen, line 102: To discuss

packages/devextreme/js/__internal/grids/new/card_view/header_panel/options.test.ts
```diff
@@ -0,0 +1,150 @@
...
+            },
+          },
+        });
+
+        // @ts-expect-error
+        expect(renderSpy.mock.calls[0][0]).toMatchObject({
+          dropFeedbackMode: 'push',
+          scrollSpeed: 555,
+          scrollSensitivity: 111,
+        });
```
Raushen, line 58: To discuss

packages/devextreme/js/__internal/grids/new/card_view/header_panel/__snapshots__/options.test.ts.snap
```diff
@@ -0,0 +1,156 @@
...
+        >
+          <div
+            class="dx-scrollable-content"
+            style="left: 0px; top: 0px; transform: none;"
+          >
+            <div
+              class="dx-cardview-headers-content"
+            >
+              <div
+                class="dx-cardview-header-item undefined undefined"
```
Raushen: to discuss

packages/devextreme/js/__internal/grids/new/card_view/header_panel/options.test.ts
pomahtri: It's not possible to check it in jest
We override callbacks as we need to use them too, so we can't just check that callbacks match props. We need to trigger reordering and then check that callbacks are called, we can do it only in testcafe

packages/devextreme/js/__internal/grids/new/card_view/header_panel/column_sortable.tsx
```diff
@@ -0,0 +1,141 @@
...
+export class ColumnSortable extends Component<Props, State> {
+  status: Status = 'moving';
+
+  dragItemProps?: {
+    container: HTMLElement;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    props: any;
+  };
+
+  private readonly onDragStart = (e): void => {
```
wdevfx: Here and in some other event handlers, the type of ```e``` is missing
Let's add types to them
(As far as I can see, there are all ok with types in the ```<Sortable />``` props)

packages/devextreme/js/__internal/grids/new/card_view/header_panel/column_sortable.tsx
```diff
@@ -0,0 +1,141 @@
...
+      Math.abs(dragCoords.y - containerCoords.y),
+      Math.abs(dragCoords.y - containerCoords.y + containerCoords.height),
+    );
+
+    // TODO: move to scss variable
+    if (yDistance <= 64) {
+      this.status = 'moving';
+    } else {
+      this.status = 'forbid';
+    }
```
wdevfx: Minor comment.

We can use a ternary operator instead of a single ```if / else``` 
It makes code a little shorter and simpler (but it's the subjective opinion)

E.g:
```ts
  // TODO: move the 64 magic const to scss variable
  this.status = yDistance <= 64 ? 'moving' : 'forbid'; 
```

packages/devextreme/js/__internal/grids/new/card_view/header_panel/column_sortable.tsx
```diff
@@ -0,0 +1,141 @@
...
+    this.props.onMove(
+      e.itemData.column,
+      e.toIndex,
+      e.itemData.source,
+    );
+  };
+
+  // TODO: move all none-native approaches to sortable wrapper
+  private readonly renderDragTemplate = (): void => {
+    if (this.dragItemProps) {
```
wdevfx: Minor comment.

Let's reduce the nesting (and complexity level) of this method
Also, we can get rid of the ```no-non-null-assertion``` by checking the ```this.props.dragTemplate``` here

Something like this:
```ts
if (!this.dragItemProps || !this.props?.dragTemplate) {
  return;
}

// other code
```

packages/devextreme/js/__internal/grids/new/card_view/header_panel/header_panel.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  visible: boolean;
+
+  draggingOptions?: DraggingOptions;
+}
+
+export class HeaderPanel extends Component<HeaderPanelProps> {
+  public render(): JSX.Element {
+    if (!this.props.visible) {
+      return <></>;
```
wdevfx, line 40: Question.

Can we return ```null``` here instead of fragment (```<></>```)?

packages/devextreme/js/__internal/grids/new/card_view/header_panel/item.tsx
```diff
@@ -0,0 +1,86 @@
...
+  status?: Status;
+  showSortIndexes?: boolean;
+  onSortClick?: () => void;
+  template?: ComponentType<{ column: Column }>;
+  cssClass?: string;
+}
+
+export function Item(props: ItemProps): JSX.Element {
+  const Template = props.column.headerItemTemplate ?? props.template;
+  const cssClass = `${CLASSES.item} ${props.column.headerItemCssClass ?? ''} ${props.cssClass ?? ''}`;
```
wdevfx, line 66: Now if the ```props.column.headerItemCssClass``` and ```props.cssClass``` is undefined or null we will get CSS string with space in the end of it
E.g: ```"dx-cardview-header-item  "```

To prevent this, let's move ```combineClasses``` utils from R1 and use it
https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/core/r1/utils/render_utils.ts#L1

packages/devextreme/js/__internal/grids/new/card_view/header_panel/header_panel.tsx
pomahtri: We cannot, typing issue :(

packages/devextreme/js/__internal/grids/new/card_view/header_panel/column_sortable.tsx
pomahtri: There were problems with that public types, and I decided to make type any, because otherwise there were too much '@ts-expect-error`s
But now I think better to explicitly ignore all errors

pomahtri: I will rewrite it

packages/devextreme/js/__internal/grids/new/card_view/header_panel/item.tsx
pomahtri: I tried to use it, but unfortunately since `props.column.headerItemTemplate` can be undefined, it's impossible to use it in combineClasses: typing issue, and plus maybe it will be wrongly stringified:
```
combineClasses({
  [undefined]: false,
  [undefined]: true.   // rewrite. Is it okay?
})
```

pomahtri: So I think we'll need to enhance the utility in future. Maybe smth like
```
combineClasses({
  [CLASSES.conditionClass]: myCondition,
}, [CLASSES.alwaysClass])
```

wdevfx: Yup, agree to postpone it but let's leave a ```TODO``` record

PR #28843, opened 2025-01-27
CardView: DataController: implement Remote Operations

Raushen:


packages/devextreme/js/__internal/grids/new/grid_core/data_controller/utils.ts
```diff
@@ -31,6 +33,110 @@ export function normalizeDataSource(
...
+}
+
+export function normalizeRemoteOptions(
+  remoteOperations: boolean | object | string,
+  localStore: boolean,
+  customStore: boolean,
+): OperationOptions {
+  if (remoteOperations instanceof String && remoteOperations !== 'auto') {
+    throw new Error('Remote operations do not support any string values except \'auto\'');
+  }
```
pomahtri: Let's move this check to TS and specify more precise type of `remoteOperations` instead of `object | string`

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/types.ts
```diff
@@ -1,3 +1,9 @@
 export type DataObject = Record<string, unknown>;
 export type Key = unknown;
 export type KeyExpr = unknown;
+export interface OperationOptions {
+  filtering?: boolean;
+  sorting?: boolean;
+  paging?: boolean;
+  summary?: boolean;
```
pomahtri: Let's remove summary as we will not support it in this iteration. It is okay to remove it only from public type so that customers will not be confused

pomahtri (approved):

PR #28855, opened 2025-01-28
fix: Next.js hydration throws a class mismatch error

VasilyStrelyaev:


ivanblinov2k17 (dismissed):

packages/devextreme/js/__internal/core/utils/m_common.ts
```diff
@@ -7,6 +7,9 @@ import {
   isDefined, isFunction, isObject, isString, type,
 } from '@js/core/utils/type';
 
+// @ts-expect-error new deferred
+export const waitUntilFirstWidgetCreating = new Deferred();
+
```
GoodDayForSurf, line 12: at this point it is assumed that will be only one instance of module but in module federation a module can have more than one instance...  won't this interfere?

packages/devextreme/js/__internal/core/m_devices.ts
```diff
@@ -4,6 +4,7 @@ import { EventsStrategy } from '@js/core/events_strategy';
 import $ from '@js/core/renderer';
 import type { Callback } from '@js/core/utils/callbacks';
 import Callbacks from '@js/core/utils/callbacks';
+import { when } from '@js/core/utils/deferred';
```
IlyaKhD, line 7: I guess it was already migrated (other imports should be eventually fixed as well:
```suggestion
import { when } from '@ts/core/utils/m_deferred';
```

packages/devextreme/js/__internal/core/utils/m_common.ts
```diff
@@ -7,6 +7,9 @@ import {
   isDefined, isFunction, isObject, isString, type,
 } from '@js/core/utils/type';
 
+// @ts-expect-error new deferred
+export const waitUntilFirstWidgetCreating = new Deferred();
```
IlyaKhD: I guess it's `Created`, not `Creating`. If I'm not mistaken Promise callbacks will be called after component's ctor syncronously finished it's work, i.e. component is created.

IlyaKhD: I suggest renaming it:

```suggestion
export const firstComponentCreated = new Deferred();
```
or better smth generalized, maybe to highlight the logical segregation of "core/components" layers. Like this:
```suggestion
export const domLayerInitialized = new Deferred();
```
or
```suggestion
export const uiLayerInitialized = new Deferred();
```
or 
```suggestion
export const componentsCreated = new Deferred();
```
or
```suggestion
export const componentsLayerInitialized = new Deferred();
```

etc

packages/devextreme/js/__internal/core/widget/dom_component.ts
```diff
@@ -86,6 +87,8 @@ class DOMComponent<
   }
 
   ctor(element: Element, options: TProperties): void {
+    waitUntilFirstWidgetCreating.resolve();
```
IlyaKhD: As I've written above - since it's a promise, not direct callback call, it looks like `Created`, not `Creating`

So, shouldn't it be moved to the end of ctor body to avoid misinterpretation in the future? Maybe, it should be called in the leaf child, so that it would be truly marks a componetn initialization is over?

packages/devextreme/js/__internal/core/m_devices.ts
VasilyStrelyaev: `when` from `m_deferred` expects 0 arguments

an argument appears only in `js/core/utils/deferred.d.ts`

packages/devextreme/js/__internal/core/utils/m_common.ts
VasilyStrelyaev: The NX docs says that it's a common threat when multiple versions of shared libraries are used. See https://nx.dev/concepts/module-federation/module-federation-and-nx#managing-versions

That's why it recommends SVP for libraries with internal state: https://nx.dev/concepts/module-federation/manage-library-versions-with-module-federation#our-approach

My takeaway from this is that having internal state in a library is not a showstopper for using it with Module Federation. Users just have to take it into account when building the app.

We already have license checking data, and the _entire config_ as the library internal state. Will we get something if we don't add this Deferred here?

packages/devextreme/js/__internal/core/m_devices.ts
IlyaKhD: Looks like the declaration in the m_deferred.ts module should be updated, as nothing is changed in the export:

https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/core/utils/deferred.js

```ts
export { fromPromise, setStrategy, Deferred, when } from '../../__internal/core/utils/m_deferred';
```

The idea is that internal code should reference other internal code directly, without going through the public facade.

IlyaKhD (approved):

GoodDayForSurf (approved):

PR #28856, opened 2025-01-28
TagBox: tags should be actualized after data source change (T1253312)

Zedwag:


packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -798,6 +806,7 @@ const TagBox = (SelectBox as any).inherit({
       })
       .done((data, extra) => {
         this._isDataSourceChanged = false;
+        this._newDataSource = false;
```
EugeniyKiyashko: The `_newDataSource` doesn't say that it stores value with boolean type.

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1005,6 +1028,12 @@ const TagBox = (SelectBox as any).inherit({
     return selectedItems;
   },
 
+  _processDataSourceChanging() {
+    this._newDataSource = true;
```
EugeniyKiyashko:
```suggestion
    this._isDataSourceChanging = true;
```

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -831,10 +841,23 @@ const TagBox = (SelectBox as any).inherit({
...
-        }) as never);
+        if (!this._dataSource) {
+          return;
+        }
+
+        loadItemPromises.push(
+          this._loadItem(value, cache)
+            .done((item) => {
+              const newItem = this._createTagData(item, value);
+              items.splice(index, 0, newItem as never);
```
EugeniyKiyashko: This cast for the newItem type looks like a hack. Please add the correct types. 
If that's difficult for you, at least use a // @ts-expect-error comment instead in all places.

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1058,13 +1087,15 @@ const TagBox = (SelectBox as any).inherit({
       this._tagElements().remove();
     } else {
       const $tags = this._tagElements();
-      const values = this._getValue();
+
+      const selectedItems = this._selectedItems ?? [];
```
EugeniyKiyashko:
```suggestion
      const selectedItems = this._fieldRenderData() ?? [];
```

packages/devextreme/js/__internal/ui/m_tag_box.ts
ksercs:
```suggestion
        this._isDataSourceOptionChanged = false;
```

packages/devextreme/js/__internal/ui/m_tag_box.ts
ksercs: actually it's already not changing at this moment

packages/devextreme/js/__internal/ui/m_tag_box.ts
ksercs: tbh Alex's option is more clear to me)

Zedwag: i will use `selectedItems` from option(), should be ok in this place, since it's already updated

packages/devextreme/js/__internal/ui/m_tag_box.ts
Zedwag: renamed to `_isDataSourceOptionChanged`

packages/devextreme/js/__internal/ui/m_tag_box.ts
Zedwag: add comments here and in other places, i don't want to think on types updates in scope of this task which is arguable enough itself

packages/devextreme/js/__internal/ui/m_tag_box.ts
Zedwag: renamed to _isDataSourceOptionChanged

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -831,13 +841,30 @@ const TagBox = (SelectBox as any).inherit({
...
+        if (!this._dataSource) {
+          return;
+        }
+
+        loadItemPromises.push(
+          // @ts-expect-error
+          this._loadItem(value, cache)
+            .done((item) => {
+              const newItem = this._createTagData(item, value);
+              // @ts-expect-error// @ts-expect-error
```
EugeniyKiyashko:
```suggestion
              // @ts-expect-error
```

EugeniyKiyashko (approved):

ksercs (approved):

PR #28869, opened 2025-01-30
Pagination bug in compact theme

tongsonbarbs:


packages/devextreme/js/__internal/pagination/content.tsx
```diff
@@ -216,6 +216,7 @@ export class PaginationContent extends InfernoComponent<PaginationContentProps>
       hoverStateEnabled,
     } = this.props;
 
+    const isLargeDisplayMode = this.getIsLargeDisplayMode();
```
tongsonbarbs: Hi @Raushen , these changes are unnecessary but I just added these refactoring to avoid possible change in value during runtime.. Also it's a good practice..

wdevfx (approved):

PR #28876, opened 2025-01-30
Splitter re-initializes templates for items that are at the same level as an item being shown/hidden (T1267030)

EugeniyKiyashko:


packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -659,14 +668,107 @@ class Splitter extends CollectionWidgetLiveUpdate<Properties> {
...
+
+        this._itemVisibleOptionChanged(item, value as boolean, prevValue as boolean);
         break;
+      }
       default:
         super._itemOptionChanged(item, property, value, prevValue);
     }
   }
 
+  _itemVisibleOptionChanged(item: Item, value: boolean, prevValue: boolean): void {
```
anna-shakhova: Check condition is confusing, as long we defined `value` and `prevValue` as booleans, `Item` interface allows them to be undefined.
Can we simplify it? 
If value is undefined, will the later logic work properly?

```suggestion
  _itemVisibleOptionChanged(item: Item, value = true, prevValue = true): void {
      if (prevValue === value) {
```

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -659,14 +668,107 @@ class Splitter extends CollectionWidgetLiveUpdate<Properties> {
       case 'collapsible':
         this._updateResizeHandlesCollapsibleState();
         break;
-      case 'visible':
-        this._invalidate();
+      case 'visible': {
+        super._itemOptionChanged(item, property, value, prevValue);
+
+        this._itemVisibleOptionChanged(item, value as boolean, prevValue as boolean);
```
anna-shakhova:
```suggestion
        this._itemVisibleOptionChanged(item, value as Item['visible'], prevValue as Item['visible']);
```

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -659,14 +668,107 @@ class Splitter extends CollectionWidgetLiveUpdate<Properties> {
...
+      !value,
+      this._panesCacheSizeVisible,
+      direction,
+    );
+
+    this._itemRestrictions.map((pane) => {
+      pane.maxSize = undefined;
+      pane.resizable = undefined;
+
+      return item;
```
anna-shakhova: item has different type than itemRestrictions, perhaps, pane was meant

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -659,14 +668,107 @@ class Splitter extends CollectionWidgetLiveUpdate<Properties> {
...
+  _getDistributionData(item: Item, value: boolean | undefined): {
+    direction: CollapseExpandDirection;
+    paneIndex: number;
+  } {
+    const itemIndex = this._getIndexByItem(item);
+    const $item = $(this._itemElements()[itemIndex]);
+    const { items = [] } = this.option();
+
+    const isLastVisibleItem = itemIndex >= findLastIndexOfVisibleItem(items);
+    const isLastVisible = this._isLastVisibleItem(itemIndex) || isLastVisibleItem;
```
anna-shakhova: Redundant, `this._isLastVisibleItem(index)` returns `true` if `index === findLastIndexOfVisibleItem(items)` and we've already checked it before

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -659,14 +668,107 @@ class Splitter extends CollectionWidgetLiveUpdate<Properties> {
...
+          direction: CollapseExpandDirection.Next,
+          paneIndex: this._getPaneIndexByElement(this._getResizeHandleLeftItem($item)),
+        };
+      }
+
+      return {
+        direction: CollapseExpandDirection.Previous,
+        paneIndex: this._getPaneIndexByElement($item),
+      };
+    } if (isLastVisible) {
```
anna-shakhova: Move to next line, please
It is hard to read 4 returns, perhaps, it would be easier to calculate direction and paneIndex first
Suggestion requires to set default value for `value` to true to work proper with undefined value
```suggestion
    const direction = value === isLastVisible ? CollapseExpandDirection.Previous : CollapseExpandDirection.Next;
    const paneIndex = isLastVisible ? this._getPaneIndexByElement(this._getResizeHandleLeftItem($item)) : this._getPaneIndexByElement($item);
    
    return {
      direction,
      paneIndex,
    };
```

packages/devextreme/js/__internal/ui/splitter/splitter_item.ts
```diff
@@ -27,8 +27,15 @@ class SplitterItem extends CollectionWidgetItem<Item> {
     this._owner = options.owner;
   }
 
+  _renderVisible(
+    value: boolean | undefined,
+    oldValue?: boolean | undefined,
```
anna-shakhova: redundant

packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
```diff
@@ -1328,6 +1328,285 @@ QUnit.module('Pane sizing', moduleConfig, () => {
...
+            });
+        });
+
+        QUnit.test('Changing pane visibility does not trigger full rerender (T1267030)', function(assert) {
+            this.reinit({
+                orientation,
+                dataSource: [{}, {}, {}]
+            });
+
+            const $pane = this.getPanes().eq(0).attr('id');
```
anna-shakhova:
```suggestion
            const initialId = this.getPanes().eq(0).attr('id');
```

packages/devextreme/js/__internal/ui/splitter/splitter_item.ts
EugeniyKiyashko: Yep, I rewrote the implementation and we can remove all method. Thanks for your attention

packages/devextreme/js/__internal/ui/splitter/splitter.ts
EugeniyKiyashko: The base logic is not working correctly in cases where the value resets to the default state, which is why we need to add an additional condition. In theory, itemOptionChanged should not be called in some cases—when changing from undefined to true and vice versa -because it's the same meaning in our scenario.

I would like to safely cast to a boolean if something different from the user was passed, as it looks safer. Unfortunately, in jQuery, it can pass values like 0, 1, '', or 'some string'.

packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/28876/commits/58a0820fcd17a0422a9d90b12eaee01c5e286602

packages/devextreme/js/__internal/ui/splitter/splitter.ts
EugeniyKiyashko: I made all conditions separate because it is easier to understand and debug. Both variants are okay for me if the shorter one suits you.

https://github.com/DevExpress/DevExtreme/pull/28876/commits/58a0820fcd17a0422a9d90b12eaee01c5e286602

packages/devextreme/js/__internal/ui/splitter/splitter.ts
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/28876/commits/58a0820fcd17a0422a9d90b12eaee01c5e286602

packages/devextreme/js/__internal/ui/splitter/splitter.ts
EugeniyKiyashko: Actually, it doesn't matter in this case. I would prefer to use Array.forEach() to iterate through the array of restrictions.

https://github.com/DevExpress/DevExtreme/pull/28876/commits/58a0820fcd17a0422a9d90b12eaee01c5e286602

packages/devextreme/js/__internal/ui/splitter/splitter.ts
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/28876/commits/58a0820fcd17a0422a9d90b12eaee01c5e286602

packages/devextreme/js/__internal/ui/splitter/splitter.ts
EugeniyKiyashko: https://github.com/DevExpress/DevExtreme/pull/28876/commits/58a0820fcd17a0422a9d90b12eaee01c5e286602

Zedwag (dismissed):

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -650,24 +668,111 @@ class Splitter extends CollectionWidgetLiveUpdate<Properties> {
...
         break;
+      }
       default:
         super._itemOptionChanged(item, property, value, prevValue);
     }
   }
 
-  _itemCollapsedOptionChanged(item: Item, value: boolean, prevValue: boolean): void {
+  _itemVisibleOptionChanged(item: Item, value = true, prevValue = true): void {
+    if (Boolean(prevValue) === Boolean(value)) {
```
anna-shakhova, line 697: no need in type conversion - we already overwrote undefined with true

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -659,28 +668,33 @@ class Splitter extends CollectionWidgetLiveUpdate<Properties> {
...
         break;
+      }
       case 'resizable':
         this._updateResizeHandlesResizableState();
         break;
       case 'collapsible':
         this._updateResizeHandlesCollapsibleState();
         break;
       case 'visible': {
+        type PropertyType = Item[typeof property];
```
anna-shakhova, line 684: Nice!

anna-shakhova (approved):

Zedwag (approved):

packages/devextreme/js/__internal/ui/splitter/splitter.ts
EugeniyKiyashko: Discussed in chat. Anna agreed to leave current option

PR #28947, opened 2025-02-07
FileUploader: Attach Enter keyup handler to select file dialog button (T1256752)

marker-dao:


packages/devextreme/js/__internal/ui/m_file_uploader.ts
```diff
@@ -861,19 +864,30 @@ class FileUploader extends Editor<FileUploaderProperties> {
     this._isCustomClickEvent = false;
   }
 
-  _attachSelectFileDialogHandler(target) {
+  _selectFileDialogKeyUpHandler(e): void {
```
anna-shakhova: Please, define type for `e`

packages/devextreme/js/__internal/ui/m_file_uploader.ts
```diff
@@ -861,19 +864,30 @@ class FileUploader extends Editor<FileUploaderProperties> {
...
+    if (normalizeKeyName(e) === ENTER_KEY) {
+      this._selectFileDialogClickHandler();
+    }
+  }
+
+  _attachSelectFileDialogHandlers(target) {
```
anna-shakhova: Please, define type for `target`

packages/devextreme/js/__internal/ui/m_file_uploader.ts
```diff
@@ -861,19 +864,30 @@ class FileUploader extends Editor<FileUploaderProperties> {
...
-    eventsEngine.on($(target), 'click', this._selectFileDialogHandler);
+
+    this._detachSelectFileDialogHandlers(target);
+
+    eventsEngine.on($(target), 'click', this._selectFileDialogClickHandler);
+    eventsEngine.on($(target), 'keyup', this._selectFileDialogKeyUpHandler.bind(this));
   }
 
-  _detachSelectFileDialogHandler(target) {
+  _detachSelectFileDialogHandlers(target) {
```
anna-shakhova: Please, define type for `target`

anna-shakhova (changes requested):

packages/devextreme/js/__internal/ui/m_file_uploader.ts
```diff
@@ -861,19 +864,30 @@ class FileUploader extends Editor<FileUploaderProperties> {
...
     if (!isDefined(target)) {
       return;
     }
-    eventsEngine.off($(target), 'click', this._selectFileDialogHandler);
+
+    eventsEngine.off($(target), 'click', this._selectFileDialogClickHandler);
+    eventsEngine.off($(target), 'keyup', this._selectFileDialogKeyUpHandler.bind(this));
```
ksercs: [mistake]
this is not working while `.bind` creates a new function
Please check how it's done in other components. Probably you can create a new event namespace

packages/devextreme/js/__internal/ui/m_file_uploader.ts
```diff
@@ -861,19 +864,30 @@ class FileUploader extends Editor<FileUploaderProperties> {
...
     if (!isDefined(target)) {
       return;
     }
-    this._detachSelectFileDialogHandler(target);
-    eventsEngine.on($(target), 'click', this._selectFileDialogHandler);
+
+    this._detachSelectFileDialogHandlers(target);
+
+    eventsEngine.on($(target), 'click', this._selectFileDialogClickHandler);
+    eventsEngine.on($(target), 'keyup', this._selectFileDialogKeyUpHandler.bind(this));
```
ksercs: [refactoring]
Can we also get rid of copypaste for target wrapping?
Ans I personally prefer arrow functions to `bind` just because it's a more modern syntax, but it's up to you

packages/devextreme/js/__internal/ui/m_file_uploader.ts
ksercs: Btw maybe we don't need a separate method for such a simple task - just run this `if` in arrow function when subscribe?

packages/devextreme/js/__internal/ui/m_file_uploader.ts
```diff
@@ -834,15 +836,16 @@ class FileUploader extends Editor<FileUploaderProperties> {
       disabled: this.option('readOnly'),
       hoverStateEnabled: this.option('hoverStateEnabled'),
     });
-    this._selectFileDialogHandler = this._selectButtonClickHandler.bind(this);
+
+    this._selectFileDialogClickHandler = this._selectButtonClickHandler.bind(this);
```
ksercs: I believe we can get rid of this caches if we use namespaces

packages/devextreme/js/__internal/ui/m_file_uploader.ts
ksercs: btw now i see that for avoiding this issue `_selectFileDialogHandler` private cache was used. I suggest you to not use this approach but refactor both old and new subscription

marker-dao: I'm assuming the cache was used for a reason. It is needed to ensure that a given handler is removed from the subscription. You suggest using arrow functions, but you can't remove such a subscription because your function is not saved anywhere. You need to pass the same handler (to clear the handlersController) to correctly unsubscribe.

I also see this hack being used in our code: https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/m_sortable.ts#L66

If I understood correctly, it is not used only where unsubscribe is not important.

This has nothing to do with adding namespace. Nevertheless, I use it to make the code more correct.

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: https://github.com/DevExpress/DevExtreme/pull/28947#discussion_r1954548220

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: https://github.com/DevExpress/DevExtreme/pull/28947#discussion_r1954548220

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: https://github.com/DevExpress/DevExtreme/pull/28947#discussion_r1954548220

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: [6f89b7f67a4112e1f45570a4cfb3a3c4ea8f1247]

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: [ffcb81850d2ebcbc378673ccd0b789e9c41cb351]

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: [ffcb81850d2ebcbc378673ccd0b789e9c41cb351]

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: [ffcb81850d2ebcbc378673ccd0b789e9c41cb351]

packages/devextreme/js/__internal/ui/m_file_uploader.ts
ksercs: U can unsubscribe namespace

anna-shakhova (dismissed):

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: [de296c41c1a81df6f6ecf35bd5f5d86bf831cc8a, 11b18755b6a59bf22ead285faa81a797cda4a09a]

packages/devextreme/js/__internal/ui/m_file_uploader.ts
```diff
@@ -60,6 +60,16 @@ const FILEUPLOADER_AFTER_LOAD_DELAY = 400;
 const FILEUPLOADER_CHUNK_META_DATA_NAME = 'chunkMetadata';
 const DRAG_EVENT_DELTA = 1;
 
+const DIALOG_TRIGGER_EVENT_NAMESPACE = 'dxFileUploaderDialogTrigger';
+
+const keyUpEventName = 'keyup';
+
+const dialogTriggerClickEventName = addNamespace(clickEventName, DIALOG_TRIGGER_EVENT_NAMESPACE);
+const dialogTriggerKeyUpEventName = addNamespace(keyUpEventName, DIALOG_TRIGGER_EVENT_NAMESPACE);
```
ksercs: Why do you create it in a global scope if it's used in the single place? Please move it to the appropriate method

packages/devextreme/js/__internal/ui/m_file_uploader.ts
```diff
@@ -834,19 +844,21 @@ class FileUploader extends Editor<FileUploaderProperties> {
       disabled: this.option('readOnly'),
       hoverStateEnabled: this.option('hoverStateEnabled'),
     });
-    this._selectFileDialogHandler = this._selectButtonClickHandler.bind(this);
 
     // NOTE: click triggering on input 'file' works correctly only in native click handler when device is used
     if (devices.real().deviceType === 'desktop') {
-      this._selectButton.option('onClick', this._selectFileDialogHandler);
+      this._selectButton.option('onClick', this._selectFileDialogClickHandler);
```
ksercs: [mistake?]
How tests can be green if context is lost here? You should use arrow function or bind here to keep it working with `this.` if i miss nothing

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: c1058a4c403d37dcd8e1337200ce9985c3463039

packages/devextreme/js/__internal/ui/m_file_uploader.ts
marker-dao: c1058a4c403d37dcd8e1337200ce9985c3463039

Zedwag (approved):

ksercs (approved):

PR #28961, opened 2025-02-10
fix(VueWrapper): correctly sync wrapper classes with classes added by a JS component

chaosmirage:
**Problem**
Once a dx-class is added dynamically through the Vue's attrs API, it is added but isn’t removed from a DOM element after it should according to dynamic addition/removing logic.

**Solution**
Save previous classes received from the attrs API (because if a dynamic class is added and then removed the attrs API doesn't contain it and we don't know which dx-classes were added by a JS component internal logic or via attrs API) and cleanup them before an update.

### DIFF
--- a/packages/devextreme-vue/src/core/__tests__/component.test.ts
+++ b/packages/devextreme-vue/src/core/__tests__/component.test.ts
@@ -3,7 +3,7 @@ import { mount } from '@vue/test-utils';
 import * as events from 'devextreme/events';
 import config from 'devextreme/core/config';
 import {
-  App, createVNode, defineComponent, h, nextTick, renderSlot,
+  App, createVNode, defineComponent, h, nextTick, renderSlot, ref,
 } from 'vue';
 import { createRouter, createWebHistory } from 'vue-router';
 
@@ -18,8 +18,8 @@ import { getNodeOptions } from '../vue-helper';
 import {
   prepareComponentConfig,
   prepareExtensionComponentConfig,
-  prepareConfigurationComponentConfig
-} from "../index";
+  prepareConfigurationComponentConfig,
+} from '../index';
 
 interface CustomApp extends App {
   test: string;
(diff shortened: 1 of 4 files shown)


ivanblinov2k17 (dismissed):

packages/devextreme-vue/src/core/__tests__/textbox.test.ts
```diff
@@ -103,7 +103,7 @@ describe('two-way binding', () => {
     const component = wrapper.getComponent('#component');
     await wrapper.setProps({ customClass: false });
     await nextTick(() => {
-      expect(component.element.classList.toString()).toBe(' dx-show-invalid-badge dx-textbox dx-texteditor dx-editor-outlined dx-texteditor-empty dx-widget');
+      expect(component.element.classList.toString()).toBe('dx-show-invalid-badge dx-textbox dx-texteditor dx-editor-outlined dx-texteditor-empty dx-widget');
```
ivanblinov2k17, line 106: interesting typo, didn't know that to be operator compares strings without lead/trail spaces

packages/devextreme-vue/src/core/component.ts
```diff
@@ -50,12 +50,32 @@ config({
...
+
+function prepareAttrs(attrs, dxClassesFromElement: string) {
   const attributes = {};
   includeAttrs.forEach((attr) => {
     const attrValue = attrs[attr];
     if (attrValue !== undefined && attrValue !== null) {
-      attributes[attr] = attr === 'class' && dxClasses.length ? `${attrValue} ${dxClasses.join(' ')}` : attrValue;
+      // a JS component adds classes by an internal logic,
+      // here these classes are forwarded as attrs
+      const shouldEnrichClassesViaDxClassesFromElement = attr === 'class' && dxClassesFromElement.length;
```
dxvladislavvolkov: I think the variable name is too detailed. It seems to me that this variable is not necessary. Just add this expression to the condition

packages/devextreme-vue/src/core/component.ts
```diff
@@ -50,12 +50,32 @@ config({
...
+  }
+}
+
+function prepareAttrs(attrs, dxClassesFromElement: string) {
   const attributes = {};
   includeAttrs.forEach((attr) => {
     const attrValue = attrs[attr];
     if (attrValue !== undefined && attrValue !== null) {
-      attributes[attr] = attr === 'class' && dxClasses.length ? `${attrValue} ${dxClasses.join(' ')}` : attrValue;
+      // a JS component adds classes by an internal logic,
```
dxvladislavvolkov: We don't usually add comments to the code

packages/devextreme-vue/src/core/__tests__/textbox.test.ts
chaosmirage: It compares with lead/trail spaces. This change is a consequence of the fix for the case when 

```
attributes[attr] = attr === 'class' && dxClasses.length ? `${attrValue} ${dxClasses.join(' ')}` : attrValue;
```

in this example the `attrValue` is empty, result will be ` some-class` instead of `some-class`.

packages/devextreme-vue/src/core/component.ts
chaosmirage: changed

packages/devextreme-vue/src/core/component.ts
chaosmirage: removed

ivanblinov2k17 (dismissed):

packages/devextreme-vue/src/core/component.ts
```diff
@@ -44,18 +44,31 @@ export interface IBaseComponent extends ComponentPublicInstance, IWidgetComponen
...
+const dxClassesPrefix = 'dx-';
 
 config({
   buyNowLink: 'https://go.devexpress.com/Licensing_Installer_Watermark_DevExtremeVue.aspx',
   licensingDocLink: 'https://go.devexpress.com/Licensing_Documentation_DevExtremeVue.aspx',
 });
 
-function getAttrs(attrs, dxClasses: string[]) {
+function parseClassList(classList: string): string[] {
+  return classList.split(' ').filter((currentClass) => currentClass);
```
GoodDayForSurf: how about .trim().split(/\s+/)?

chaosmirage: replaced

ivanblinov2k17 (dismissed):

packages/devextreme-vue/src/core/__tests__/component.test.ts
```diff
@@ -156,6 +156,119 @@ describe('component rendering', () => {
...
+
+      expect(componentContainer.element.className).toBe('dx-chat dx-hover dx-chat-disabled');
+
+      await wrapper.find('button').trigger('click');
+
+      expect(componentContainer.element.className).toBe('dx-chat dx-hover');
+
+      componentContainer.element.classList.remove('dx-hover');
+
+      expect(componentContainer.element.className).toBe('dx-chat');
```
GoodDayForSurf: does component rerender here? if not this expect is always true

chaosmirage: removed

GoodDayForSurf (approved):

PR #28968, opened 2025-02-11
DataGrid - Filter Row's Reset button does not work after a custom filter is set in Filter Builder (T1267481)

tongsonbarbs:


e2e/testcafe-devextreme/tests/dataGrid/filterRow/filterRow.ts
```diff
@@ -106,3 +106,52 @@ test('FilterRow range overlay screenshot', async (t) => {
...
   },
 }));
+
+// T1267481
+test('Filter Row\'s Reset button does not work after a custom filter is set in Filter Builder', async (t) => {
+  const dataGrid = new DataGrid('#container');
+  const filterEditor = dataGrid.getFilterEditor(0, FilterTextBox);
+
+  await t
+    .expect(await dataGrid.dataRows.count)
```
Raushen: You must not use await inside expect method. Now the expression calculated only once and the exact value (for example integer, "10") passes to the expect method. The best practice to pass Promise to the method, Promise will be calculated multiple times until its result is equal to an expected one.

packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_row.ts
```diff
@@ -604,6 +604,7 @@ const columnHeadersView = (Base: ModuleType<ColumnHeadersView>) => class ColumnH
         let notFocusEditor = false;
         const isOnClickMode = isOnClickApplyFilterMode(that);
         const options = {};
+        const isNotFireEvent = selectedFilterOperation ? undefined : false;
```
Raushen: const isResetFilterOperation = !properties.itemData.name;

Raushen: const isNotFireEvent = isResetFilterOperation ? false : undefined;

Raushen (dismissed):

Raushen (approved):

PR #28983, opened 2025-02-12
NumberBox: it should be possible to use percent format with more than 7 fractional digits (T1277123)

Zedwag:


packages/devextreme/js/__internal/ui/number_box/m_number_box.mask.ts
```diff
@@ -452,7 +452,19 @@ const NumberBoxMask = NumberBoxBase.inherit({
...
+      const interval = this._getIntervalFromPrecision(maxPrecision);
+
+      return adjustPercentValue(parsedValue, interval);
+    }
+
+    return parsedValue;
+  },
+
+  _getIntervalFromPrecision(precision = 1) {
+    const result = `0.${'0'.repeat(precision)}1`;
```
marker-dao:
```suggestion
    const result = 1 / Math.pow(10, precision + 1);
```

marker-dao (dismissed):

marker-dao (approved):

packages/devextreme/js/__internal/ui/number_box/m_number_box.mask.ts
```diff
@@ -452,7 +452,21 @@ const NumberBoxMask = NumberBoxBase.inherit({
...
+
+    return parsedValue;
+  },
+
+  _getIntervalFromPrecision(precision: number) {
+    if (precision < 1) {
+      return 1;
+    }
+
+    return 10 ** -precision;
```
ksercs, line 469: [minor]
if your method does not use `this` it should be an util (or global function), not a class method

ksercs (approved):

PR #29014, opened 2025-02-14
Core - improve `combineClasses` utility

pomahtri:


packages/devextreme/js/__internal/core/utils/combine_classes.ts
```diff
@@ -1,5 +1,40 @@
-export function combineClasses(classesMap: { [key: string]: boolean }): string {
-  return Object.keys(classesMap)
-    .filter((p) => classesMap[p])
+/* eslint-disable @typescript-eslint/init-declarations */
+/* eslint-disable @typescript-eslint/unified-signatures */
+export function combineClasses(
+  classList: (undefined | string)[],
+  classesMap: { [key: string]: boolean },
+): string;
```
wdevfx, line 6: Idea.

We can use ```combineClasses``` in the array instead and simplify types and implementation
E.g:

```ts
const classes = combineClasses([
  'A', 
  'B',
  combineClasses({ 
    C: true, 
    D: false 
  }),
]);
```

packages/devextreme/js/__internal/core/utils/combine_classes.ts
```diff
@@ -1,5 +1,40 @@
...
+      classesList = classListOrMap;
+      classesMap = {};
+    } else {
+      classesList = [];
+      classesMap = classListOrMap;
+    }
+  } else {
+    classesList = classListOrMap as (undefined | string)[];
+    classesMap = classesMapOrUndefined as { [key: string]: boolean };
+  }
```
wdevfx, line 31: If we decide to leave ```combineClasses``` interface without modifications,
let's simplify this code and move normalization to a separate fn

E.g:

```ts
export function normalizeArgs(
  classListOrMap: (undefined | string)[] | Record<string, boolean>,
  classesMapOrUndefined?: Record<string, boolean>,
): [list: (undefined | string)[], map: Record<string, boolean>] {
  if (Array.isArray(classListOrMap)) {
    return [classListOrMap, classesMapOrUndefined ?? {}];
  }

  return [[], classListOrMap];
};

export function combineClasses(
  classListOrMap: (undefined | string)[] | { [key: string]: boolean },
  classesMapOrUndefined?: { [key: string]: boolean },
): string {
  const [classesList, classesMap] = normalizeArgs(classListOrMap, classesMapOrUndefined);
  const mapClasses = Object.keys(classesMap).filter((p) => classesMap[p]);
  
  return classesList
    .filter((className): className is string => !!className)
    .concat(mapClasses)
    .join(' ');
}
```

wdevfx: It's not mandatory, but my suggestion is to change the interface of ```combineClasses```

Option A
```ts
function combineClasses(classes: (string | undefined | Record<string, boolean>)[]): string {};
```
Inspired by [Vue classes binding](https://vuejs.org/guide/essentials/class-and-style#binding-to-arrays)

Option B
```ts
function combineClasses(list: (string | undefined)[]): string {};
function combineClasses(map: Record<string, boolean>): string {};
```

PR #29020, opened 2025-02-14
fix(localization): correctly format number with escaped semicolon (T1275922)

chaosmirage:
<h3><strong>Problem 1</strong></h3>
<p><code>'\\';\\'0’</code> formats the <code>8</code> as <code>‘</code> instead of <code>;8</code></p>
<p><strong>Reason</strong></p>
<p>In the <code>getSignParts</code> used the <code>split</code> without counting that format may contain escaped <code>;</code> .</p>
<blockquote>
<p><strong>‘</strong></p>
<p>Used to quote special characters in a prefix or suffix, for example, <code>&quot;'#'#&quot;</code> formats 123 to <code>&quot;#123&quot;</code>. To create a single quote itself, use two in a row: <code>&quot;# o''clock&quot;</code>.</p>
<p><strong>;</strong></p>
<p>subpattern is formed from the positive pattern with a prefixed - (ASCII U+002D HYPHEN-MINUS).</p>
</blockquote>

</blockquote>
<p><a href="https://unicode.org/reports/tr35/tr35-numbers.html">https://unicode.org/reports/tr35/tr35-numbers.html</a>

<strong>Solution</strong>
Ignore <code>;</code> if it was escaped</p>
<h3><strong>Problem 2</strong></h3>
<p>After adding a more complex case ( <code>'\\';\\'plus \\';\\' 0;minus \\';\\' 0\\';\\''</code>) as a test format I found a new problem.</p>

Expected: | "minus ; 8;"
-- | --
Result: | "minus ; ;"

<p><strong>Reason</strong></p>
<p>This is because the regex <code>/'.+'/g</code>  used in <code>removeStubs</code> finds this:</p>
<p>
image
</p>
<p><strong>Solution</strong></p>
<p>Use the <code>'[^']*'/g</code> regex to get the following result:</p>
<p>
image
</p>

### DIFF
--- a/packages/devextreme/js/common/core/localization/intl/number.js
+++ b/packages/devextreme/js/common/core/localization/intl/number.js
@@ -90,7 +90,9 @@ export default {
             return getFormatter(format)(value);
         }
 
-        return this.callBase.apply(this, arguments);
+        const result = this.callBase.apply(this, arguments);
+
+        return result;
     },
     _getCurrencySymbolInfo: function(currency) {
         const formatter = getCurrencyFormatter(currency);
--- a/packages/devextreme/js/common/core/localization/ldml/number.js
+++ b/packages/devextreme/js/common/core/localization/ldml/number.js
@@ -19,8 +19,36 @@ function getGroupSizes(formatString) {
     });
 }
 
+function splitSignParts(format, separatorChar = ';', escapingChar = ESCAPING_CHAR) {
+    const parts = [];
+    let currentPart = '';
+    let state = 'searchingSeparator';
+
+    for(let i = 0; i < format.length; i++) {
+        const char = format[i];
+        if(state === 'searchingSeparator' && char === escapingChar) {
+            state = 'skippingSeparationInsideEscaping';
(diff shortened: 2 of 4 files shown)


packages/devextreme/js/common/core/localization/intl/number.js
```diff
@@ -90,7 +90,9 @@ export default {
             return getFormatter(format)(value);
         }
 
-        return this.callBase.apply(this, arguments);
+        const result = this.callBase.apply(this, arguments);
```
ivanblinov2k17, line 93: Is this diff meaningful?

VasilyStrelyaev (approved):

packages/devextreme/js/common/core/localization/intl/number.js
chaosmirage: Yes, it was useful during debugging and will be useful for future work (the same as in other places).
If there are no strict rules for reducing the line count I suggest the current change.

ivanblinov2k17 (approved):

PR #29022, opened 2025-02-14
Drawer: hide animation should work on hide using opened option (T1277636) + Promise should be resolved after call show() for already opened drawer (T1263952)

Zedwag:


packages/devextreme/js/__internal/ui/drawer/m_drawer.ts
```diff
@@ -508,6 +508,9 @@ const Drawer = (Widget as any).inherit({
         this._dimensionChanged();
         break;
       case 'opened':
+        if (!this._whenAnimationCompleted) {
```
ksercs: What if it's not a first animation, and this cache already exists but the deferred is resolved?

Zedwag: the solution was reworked

ksercs (dismissed):

marker-dao (approved):

ksercs (approved):

PR #29023, opened 2025-02-14
ColorBox: fix invalid value setting error (T1274477)

r-farkhutdinov:

### DIFF
--- a/packages/devextreme/js/__internal/ui/color_box/m_color_view.ts
+++ b/packages/devextreme/js/__internal/ui/color_box/m_color_view.ts
@@ -884,7 +884,7 @@ const ColorView = (Editor as any).inherit({
         }
 
         this._updateByDrag = false;
-        this.callBase(args);
+        this.callBase({ ...args, value: this.option('value') });
         break;
       case 'matchValue':
         this._setBaseColor(value);
--- a/packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/colorBox.tests.js
+++ b/packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/colorBox.tests.js
@@ -500,6 +500,21 @@ QUnit.module('Color Box', {
         assert.equal(colorBox.option('value'), 'rgba(100, 0, 0, 1)');
     });
 
+    QUnit.test('In "instantly" mode value should not throw an error when invalid value is set (T1274981)', function(assert) {
+        try {
+            const colorBox = showColorBox.call(this, {
+                applyValueMode: 'instantly',
+            }).dxColorBox('instance');
+
+            colorBox.option('value', 'invalid');
+            assert.strictEqual(colorBox.option('value'), '#000000');
+
+            assert.ok(true, 'no errors');
+        } catch(e) {
+            assert.ok(false, `The '${e.message}' is raised`);
+        }
+    });
+
 
     QUnit.test('Option changes', function(assert) {
         const colorBox = showColorBox.call(this).dxColorBox('instance');
--- a/packages/devextreme/js/__internal/ui/color_box/m_color_box.ts
+++ b/packages/devextreme/js/__internal/ui/color_box/m_color_box.ts
@@ -320,8 +320,7 @@ const ColorBox = (DropDownEditor as any).inherit({
   },
 
   _optionChanged(args) {
-    const { value } = args;
-    const { name } = args;
+    const { name, value } = args;
 
     switch (name) {
       case 'value':


packages/devextreme/js/__internal/ui/color_box/m_color_view.ts
```diff
@@ -907,6 +907,10 @@ const ColorView = (Editor as any).inherit({
         this.callBase(args);
     }
   },
+
+  _getInstanceValue() {
```
marker-dao: Please specify the return type

packages/devextreme/js/__internal/ui/editor/editor.ts
```diff
@@ -402,7 +402,9 @@ class Editor<
   }
 
   _optionChanged(args: OptionChanged<TProperties> | Record<string, unknown>): void {
-    const { name, value, previousValue } = args;
+    const { name, previousValue } = args;
+
+    const value = this._getInstanceValue(args);
```
marker-dao: Have you tried just putting the value-option takeup inside the switch case 'value'?

marker-dao: I would prefer to call this method from case 'value' due to the fact that this is where we get the error. For example:

```
case 'value':
    const actualValue = this._getInstanceValue(args);
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/colorBox.tests.js
```diff
@@ -500,6 +500,21 @@ QUnit.module('Color Box', {
...
     });
 
+    QUnit.test('In \'instantly\' mode value should not throw an error when invalid value is set (T1274981)', function(assert) {
+        try {
+            const colorBox = showColorBox.call(this, {
+                applyValueMode: 'instantly',
+            }).dxColorBox('instance');
+
+            colorBox.option('value', 'invalid');
+            assert.equal(colorBox.option('value'), '#000000');
```
marker-dao:
```suggestion
            assert.strictEqual(colorBox.option('value'), '#000000');
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/colorBox.tests.js
```diff
@@ -500,6 +500,21 @@ QUnit.module('Color Box', {
         assert.equal(colorBox.option('value'), 'rgba(100, 0, 0, 1)');
     });
 
+    QUnit.test('In \'instantly\' mode value should not throw an error when invalid value is set (T1274981)', function(assert) {
```
marker-dao:
```suggestion
    QUnit.test('In "instantly" mode value should not throw an error when invalid value is set (T1274981)', function(assert) {
```

packages/devextreme/js/__internal/ui/color_box/m_color_view.ts
r-farkhutdinov: Fixed

packages/devextreme/js/__internal/ui/editor/editor.ts
r-farkhutdinov: Fixed

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/colorBox.tests.js
r-farkhutdinov: Fixed

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/colorBox.tests.js
r-farkhutdinov: Fixed

packages/devextreme/js/__internal/ui/editor/editor.ts
```diff
@@ -467,6 +469,12 @@ class Editor<
     this.option('value', this._initialValue);
   }
 
+  _getInstanceValue({
```
ksercs: Do I understand correctly that u decided to implement a separate method only to override it in ColorView for minimizing the count of code branches affected by your change? 
If so, I suggest you implementing the change directly in text_editor and try to not add this new method - I guess it's safe enough

This method looks a bit strange to me because:
- It's used in the single place only. It has a general name but is aimed for the specific place of using and never gonna be used somewhere else;
- It's basically a getter but has a parameter and returns it

r-farkhutdinov: Correct, this was the main reason. Agree with the naming concerns. 

_I suggest you implementing the change directly in text_editor and try to not add this new method - I guess it's safe enough_
Do you mean directly pass `.option('value')` here? If so, it didn't work as there are some inherited components that stop working properly.

r-farkhutdinov: Fixed in https://github.com/DevExpress/DevExtreme/pull/29023/commits/c49a94c0fb7ee324965ba8a8ce9b3324aba337b1

marker-dao (approved):

ksercs (approved):

PR #29031, opened 2025-02-15
List: empty message should be hidden when new item is inserted to empty store (T1277567)

Zedwag:


packages/devextreme/js/__internal/ui/collection/m_collection_widget.live_update.ts
```diff
@@ -249,6 +249,7 @@ class CollectionWidgetLiveUpdate<
     changes.forEach((change) => this[`_${change.type}ByChange`](keyInfo, items, change, isPartialRefresh));
     this._renderedItemsCount = items.length;
     this._refreshItemsCache();
+    this._renderEmptyMessage();
```
ksercs: I think now it's called to frequently. This is needed only after `insert` operation but not others.
Shouldn't we move it to the `_afterItemElementInserted`?

Zedwag: good point, that should me much better

ksercs (approved):

marker-dao (approved):

PR #29053, opened 2025-02-17
T1235433: fix Drag-n-Drop inside the group with virtual scrolling

Ambrozy:

### DIFF
--- a/e2e/testcafe-devextreme/tests/scheduler/dragAndDrop/T1235433.ts
+++ b/e2e/testcafe-devextreme/tests/scheduler/dragAndDrop/T1235433.ts
@@ -1,0 +1,105 @@
+import Scheduler from 'devextreme-testcafe-models/scheduler';
+import { createWidget } from '../../../helpers/createWidget';
+import url from '../../../helpers/getPageUrl';
+import { scrollTo } from '../virtualScrolling/utils';
+
+fixture.disablePageReloads`Scheduler Drag-and-Drop inside Group`
+  .page(url(__dirname, '../../container.html'));
+
+test('T1235433: Scheduler - Drag-n-Drop works inside the group with virtual scrolling', async (t) => {
+  const scheduler = new Scheduler('#container');
+
+  await t.expect(scheduler.element.exists).ok();
+
+  const appointment1 = scheduler.getAppointment('Book 1');
+  const appointment2 = scheduler.getAppointment('Book 2');
+  const appointment3 = scheduler.getAppointment('Book 3');
+  const appointment4 = scheduler.getAppointment('Book 4');
+
+  await t.drag(appointment1.element, -200, 0)
+    .expect(appointment1.date.time)
+    .eql('9:00 AM - 10:00 AM');
+  await t.drag(appointment2.element, -200, 0)
+    .expect(appointment2.date.time)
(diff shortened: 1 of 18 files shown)


e2e/testcafe-devextreme/tests/scheduler/utils.ts
```diff
@@ -0,0 +1,14 @@
+import { ClientFunction } from 'testcafe';
+
+export const scrollTo = ClientFunction((x, y) => {
```
wdevfx, line 3: Like for moving similar code to the separate util func 👍

e2e/testcafe-devextreme/tests/scheduler/dragAndDrop/T1235433.ts
```diff
@@ -0,0 +1,105 @@
+import Scheduler from 'devextreme-testcafe-models/scheduler';
+import { createWidget } from '../../../helpers/createWidget';
+import url from '../../../helpers/getPageUrl';
+import { scrollTo } from '../utils';
+
+fixture.disablePageReloads`Scheduler Drag-and-Drop inside Group`
+  .page(url(__dirname, '../../container.html'));
+
+test('T1235433: Scheduler - Drag-n-Drop works inside the group with virtual scrolling', async (t) => {
```
wdevfx, line 49: Let's add more test cases

- Include 'timelineMonth' because the initial issue was in this view
- Simplify test case/cases

About simplification, I prepared a simple reproducing example (from the customer one):
https://codepen.io/wdevfx/pen/MYWexJp?editors=0010
(And let's add simple reproduction example to the SC ticket notes for the history)

packages/devextreme/js/__internal/scheduler/m_subscribes.ts
```diff
@@ -72,42 +72,45 @@ const subscribes = {
...
+    const appointmentGridStartDate = createAppointmentAdapter(
+      rawAppointment,
+      this._dataAccessors,
+      this.timeZoneCalculator,
+    ).clone({ pathTimeZone: 'toGrid' }).startDate;
+    const startDateTime = appointmentGridStartDate.getTime();
+    const cellStartDateTime = new Date(targetCell.startDate).getTime();
+    const cellEndDateTime = new Date(targetCell.endDate).getTime();
+    // check that appointment start date is in cell date interval: [start, end)
+    const isDropToTheSameCell = cellStartDateTime <= startDateTime && cellEndDateTime > startDateTime;
```
wdevfx: Discussed in a call
Unfortunately, the comparison of dates breaks drag-and-drop between groups ('rows')

packages/devextreme/js/__internal/core/utils/date.ts
```diff
@@ -7,7 +7,10 @@ const addOffsets = (date: Date, offsets: number[]): Date => {
 
   return new Date(newDateMs);
 };
+// eslint-disable-next-line max-len
+const isValidDate = (date: Date | null | undefined): boolean => Boolean(date && !isNaN(new Date(date).valueOf()));
```
wdevfx: Question.

Maybe it will be better to set the type of ```date``` argument to ```any```?
Cause the runtime check for date is too strict and it looks like that ```date``` may be any type in runtime

wdevfx (approved):

PR #29056, opened 2025-02-18
DataGrid - Filter Row's Reset button does not work after a custom filter is set in Filter Builder (T1267481)

tongsonbarbs:


packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_row.ts
```diff
@@ -623,7 +623,10 @@ const columnHeadersView = (Base: ModuleType<ColumnHeadersView>) => class ColumnH
           options[isOnClickMode ? 'bufferedSelectedFilterOperation' : 'selectedFilterOperation'] = column.defaultSelectedFilterOperation || null;
         }
 
-        that._columnsController.columnOption(column.index, options);
+        // @ts-expect-error
```
Raushen, line 626: What is an issue here?

Raushen (approved):

PR #29074, opened 2025-02-19
Map: Add location field to ClickEvent type and pass event to onClick

r-farkhutdinov:
Docs: https://github.com/DevExpress/devextreme-documentation/pull/7086

### DIFF
--- a/packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
+++ b/packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
@@ -195,7 +195,10 @@ const AzureProvider = DynamicProvider.inherit({
 
   _clickActionHandler(e) {
     if (e.type === 'click') {
-      this._fireClickAction({ location: this._normalizeLocation(e.position) });
+      this._fireClickAction({
+        location: this._normalizeLocation(e.position),
+        event: e.originalEvent,
+      });
     }
   },
 
--- a/packages/devextreme/js/__internal/ui/map/m_provider.dynamic.google.ts
+++ b/packages/devextreme/js/__internal/ui/map/m_provider.dynamic.google.ts
@@ -228,7 +228,7 @@ const GoogleProvider = DynamicProvider.inherit({
   },
 
   _clickActionHandler(e) {
-    this._fireClickAction({ location: this._normalizeLocation(e.latLng) });
+    this._fireClickAction({ location: this._normalizeLocation(e.latLng), event: e.domEvent });
   },
 
   updateDimensions() {
--- a/packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/azureTests.js
+++ b/packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/azureTests.js
@@ -466,6 +466,7 @@ QUnit.module('basic options', moduleConfig, () => {
     QUnit.test('Should add onClick handler with correct args', function(assert) {
         const done = assert.async();
         let clickFired = 0;
+        const originalEvent = new PointerEvent({ type: 'click' });
 
         const map = $('#map').dxMap({
             provider: 'azure',
@@ -473,11 +474,16 @@ QUnit.module('basic options', moduleConfig, () => {
                 assert.strictEqual(e.component, map, 'click event includes component instance');
                 assert.strictEqual($(e.element).is($('#map')), true, 'click event includes root element');
                 assert.deepEqual(e.location, { lat: 88, lng: 88 }, 'click event includes correct location');
+                assert.deepEqual(e.event, originalEvent, 'click event is equal to passed originalEvent');
 
                 clickFired++;
             },
             onReady: () => {
-                atlas.clickActionCallback({ type: 'click', position: [88, 88] });
+                atlas.clickActionCallback({
+                    type: 'click',
+                    position: [88, 88],
+                    originalEvent,
+                });
                 assert.strictEqual(clickFired, 1, 'click action fired');
 
                 done();
--- a/packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/googleStaticTests.js
+++ b/packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/googleStaticTests.js
@@ -471,12 +471,17 @@ QUnit.test('click', function(assert) {
     let clicked = 0;
     let eventFired = 0;
 
+    const clickEvent = $.Event('click');
+
     return new Promise(function(resolve) {
-        new Map($('#map'), {
+        const map = new Map($('#map'), {
             provider: 'googleStatic',
             width: 400,
             height: 500,
-            onClick: function() {
+            onClick: function(e) {
+                assert.strictEqual(e.component, map, 'component is passed correctly');
+                assert.strictEqual($(e.element).is($('#map')), true, 'click event includes root element');
+                assert.strictEqual(e.event.originalEvent, clickEvent, 'event is passed correctly');
                 clicked++;
             },
             onReady: function(e) {
@@ -484,7 +489,7 @@ QUnit.test('click', function(assert) {
                 $element.dxMap('instance').on('click', function() {
                     eventFired++;
                 });
-                $element.children().trigger('dxclick');
+                $element.children().trigger(clickEvent);
 
                 resolve();
             }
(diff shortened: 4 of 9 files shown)


marker-dao:
Pls don't forget to add a tests

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.google.ts
```diff
@@ -228,7 +228,7 @@ const GoogleProvider = DynamicProvider.inherit({
   },
 
   _clickActionHandler(e) {
-    this._fireClickAction({ location: this._normalizeLocation(e.latLng) });
+    this._fireClickAction({ location: this._normalizeLocation(e.latLng), event: e });
```
ksercs: please add tests for all changes in js

r-farkhutdinov: added in https://github.com/DevExpress/DevExtreme/pull/29074/commits/6eba642be3e72e7de638242bfc65454cfeaa5ad2

Zedwag (dismissed):

marker-dao (dismissed):

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
```diff
@@ -195,7 +195,10 @@ const AzureProvider = DynamicProvider.inherit({
 
   _clickActionHandler(e) {
     if (e.type === 'click') {
-      this._fireClickAction({ location: this._normalizeLocation(e.position) });
+      this._fireClickAction({
+        location: this._normalizeLocation(e.position),
+        event: e.originalEvent,
```
ksercs, line 200: [scenarios]
I'm worried about consistency across the providers:
- I see in Azure and dynamic google you pass an original event. This at least means that there will never be a Jquery event. Does it correspond with the type and docs?
- Afair googleStatic already had `event` parameter passed. But it's not an original event, it can be a jQuery event. Moreover, googleStatic does not have `location` parameter. So its type is still not actual?
- Why don't you make changes for Bing? I still see no `event` there

packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/azureTests.js
```diff
@@ -473,11 +473,12 @@ QUnit.module('basic options', moduleConfig, () => {
                 assert.strictEqual(e.component, map, 'click event includes component instance');
                 assert.strictEqual($(e.element).is($('#map')), true, 'click event includes root element');
                 assert.deepEqual(e.location, { lat: 88, lng: 88 }, 'click event includes correct location');
+                assert.strictEqual(e.event, true, 'click event includes event instance');
 
                 clickFired++;
             },
             onReady: () => {
-                atlas.clickActionCallback({ type: 'click', position: [88, 88] });
+                atlas.clickActionCallback({ type: 'click', position: [88, 88], originalEvent: true });
```
ksercs: [synthetic test]
Why don't u just create an event here and pass it? It also allow you to be sure that exactly same event is passed and it's not a coincidence.

packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/googleStaticTests.js
```diff
@@ -476,7 +476,8 @@ QUnit.test('click', function(assert) {
             provider: 'googleStatic',
             width: 400,
             height: 500,
-            onClick: function() {
+            onClick: function(e) {
+                assert.ok(e.event, 'click event includes event instance');
```
ksercs: [test coverage]
That's not enough :) I don't like it'll be green if we pass something incorrect to the event parameter. Now you only check that smth is passed, but not that passed stuff is exactly what we want.

Moreover, tests are not only about insurance, it also describes the contract for behavior/api. Now your change does not help us with this purpose.

Also if you see that another parameters are not tests it's a good idea to test it as well

ksercs (changes requested):

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
r-farkhutdinov: 1. Yes, we have discussed this in the thread
2. Will re-check
3. This is the issue of the bing provider, it doesn't have an event field. We decided not to pass any event at all as it is deprecated.

packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/azureTests.js
r-farkhutdinov: We can only check that the handler correctly resolves the `originalEvent ` field

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
ksercs: 1. So we think it's okay to leave the old description which contains jQuery even because it's wider that we need? 
image

3. Well. So we agree to keep type not actual for Bing while it's deprecated. Maybe we should add a note to the docs?

packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/azureTests.js
ksercs: I meant to create an event, pass it here to `originalEvent` (emulating dispatching from original map), then check in onClick that exactly this event is passed

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
r-farkhutdinov: 1. Should we wrap the native event to JQuery or update the doc?

packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/googleTests.js
```diff
@@ -1534,6 +1536,8 @@ QUnit.test('click', function(assert) {
                 lat: 2,
                 lng: 10
             }, 'correct location passed');
+            assert.ok(e.event instanceof MouseEvent, 'click event includes event instance');
```
ksercs: this assert looks excess - I don't see how it can be ref is the second one is green

packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/azureTests.js
```diff
@@ -466,18 +466,25 @@ QUnit.module('basic options', moduleConfig, () => {
...
+        const originalEvent = new PointerEvent({ type: 'click' });
 
         const map = $('#map').dxMap({
             provider: 'azure',
             onClick: (e) => {
                 assert.strictEqual(e.component, map, 'click event includes component instance');
                 assert.strictEqual($(e.element).is($('#map')), true, 'click event includes root element');
                 assert.deepEqual(e.location, { lat: 88, lng: 88 }, 'click event includes correct location');
+                assert.ok(e.event instanceof PointerEvent, 'click event includes event instance');
+                assert.deepEqual(e.event, originalEvent, 'click event is equal to passed originalEvent');
```
ksercs, line 477: Our goal is to check that it's the same even which was displatched by original map, not just an event with the same data
```suggestion
                assert.strictEqual(e.event, originalEvent, 'click event is equal to passed originalEvent');
```

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
ksercs: definitely not manual wrapping. Let's go with docs

packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/googleTests.js
r-farkhutdinov: Fixed

packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/azureTests.js
r-farkhutdinov: Fixed

marker-dao (dismissed):

Zedwag (approved):

marker-dao (approved):

ksercs (approved):

PR #29079, opened 2025-02-20
Remove redundant code, semaphores and todos

Ambrozy:


packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -63,15 +62,12 @@ export class AppointmentForm {
 
   form: any;
 
-  // TODO: Why we need the "semaphore" in the sync code?
-  //  We should research it and delete it if redundant
-  semaphore: Semaphore;
+  isFormUpdating: boolean;
```
wdevfx: Let's make this field private
```ts
 // NOTE: flag to prevent double value set during form updating
 private isFormUpdating = false;
```

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -63,15 +62,12 @@ export class AppointmentForm {
...
 
   constructor(scheduler) {
     this.scheduler = scheduler;
     this.form = null;
-
-    this.semaphore = new Semaphore();
+    this.isFormUpdating = false;
```
wdevfx: Minor comment.
(It's up to u, we can leave it as is)

Let's move field initialization to the inline shorthand (see previous comment).

wdevfx (approved):

PR #29121, opened 2025-02-27
SelectBox: Skip custom item creating if customItemCreateEvent is '' and component loses focus (T1269852)

marker-dao:


packages/devextreme/js/__internal/ui/m_select_box.ts
```diff
@@ -534,16 +534,20 @@ const SelectBox = (DropDownList as any).inherit({
...
         acceptCustomValue,
         text,
         selectedItem: initialSelectedItem,
+        customItemCreateEvent,
       } = this.option();
 
       if (acceptCustomValue) {
         if (!saveEditingValue && !this._isValueChanging) {
-          this._updateField(initialSelectedItem ?? this._createCustomItem(text));
+          const initialItem = initialSelectedItem
```
r-farkhutdinov: We don't have a convention for this, but for me this expression looks too complex, we can split it into two (1 with nullish coalescing and 1 with ternary operator)

ksercs: Agree. And I'm also not sure that this is the best place to add this `if`. I see why it fixes the scenario but I cannot read the code and see it's logical. It'll be more expected to me that maybe this method (or upper "if") is not called at all.
But it's up to you, I didn't debug

marker-dao: > We don't have a convention for this, but for me this expression looks too complex, we can split it into two (1 with nullish coalescing and 1 with ternary operator)

This would be a bug, because this._createCustomItem(text) could be called even if initialSelectedItem exists. This would cause an unnecessary call to all subsequent code and lead to unexpected behavior (the user came to us with this error: the code is called unnecessarily).

marker-dao: Refactored in [48bf3c701f36fad4c3795695f3b97e3217ecca48]

marker-dao: >  is not called at all

It's supposed to be called)

ksercs (dismissed):

ksercs (approved):

PR #29160, opened 2025-03-03
TextEditorBase, TextEditorMask, TextBox: convert into ES6 class

EugeniyKiyashko:


packages/devextreme/js/__internal/ui/text_box/m_text_editor.mask.ts
```diff
@@ -48,23 +54,41 @@ const buildInMaskRules = {
...
+class TextEditorMask<
+  TProperties extends TextEditorBaseProperties= TextEditorBaseProperties,
+> extends TextEditorBase<TProperties> {
+  _changedValue?: any;
+
+  _maskStrategy!: MaskStrategy;
+
+  _$hiddenElement!: dxElementWrapper;
+
+  _typingDirection?: 'forward' | 'backward';
```
marker-dao, line 79: It is possible to create a separate type

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/actionButtons.test.js
```diff
@@ -1,1065 +0,0 @@
-import $ from 'jquery';
```
marker-dao: Why did you delete these tests?

marker-dao (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/actionButtons.test.js
EugeniyKiyashko: I was going to rename it to 'tests.js' to match the naming of the other files, but something went wrong. Well, it doesn’t matter here, so I reverted the change.

packages/devextreme/js/__internal/ui/text_box/m_text_editor.mask.ts
EugeniyKiyashko: These changes do not aim to make TypeScript typing great for each places. All improvements within your area will be handled by the squad in step 2. I sure that the squad understands the code in its own area better.

marker-dao (approved):

PR #29161, opened 2025-03-04
Get rid of the R1 runtime deps

tongsonbarbs:


packages/devextreme/webpack.config.js
```diff
@@ -34,12 +35,12 @@ module.exports = {
...
-            '@devextreme/runtime/declarations': require.resolve('@devextreme/runtime/cjs/declarations/index.js'),
-            '@devextreme/runtime/angular': require.resolve('@devextreme/runtime/cjs/angular/index.js'),
-            '@devextreme/runtime/vue': require.resolve('@devextreme/runtime/cjs/vue/index.js'),
-            '@devextreme/runtime/react': require.resolve('@devextreme/runtime/cjs/react/index.js')
+            // '@devextreme/runtime/common': require.resolve('./js/__internal/core/r1/runtime/common/index.ts'),
+            '@devextreme/runtime/inferno': path.join(__dirname, '/js/__internal/core/r1/runtime/inferno/index.ts'),
+            // '@devextreme/runtime/declarations': require.resolve('./js/__internal/core/r1/runtime/declarations/index.ts'),
+            // '@devextreme/runtime/angular': require.resolve('./js/__internal/core/r1/runtime/angular/index.ts'),
+            // '@devextreme/runtime/vue': require.resolve('./js/__internal/core/r1/runtime/vue/index.ts'),
+            // '@devextreme/runtime/react': require.resolve('./js/__internal/core/r1/runtime/react/index.tsx')
```
pomahtri: let's either revert changes in this file or delete `resolve: {...}` config completely here

packages/devextreme/tsconfig.json
```diff
@@ -48,6 +48,7 @@
     "js/**/*.d.ts",
     "js/renovation",
     "../devextreme-themebuilder",
-    "testing/testcafe"
+    "testing/testcafe",
+    "js/__internal/core/r1/runtime/**/*.ts"
```
pomahtri: why is it needed?

packages/devextreme/webpack.config.js
tongsonbarbs: I noticed this is being used in the build although we are not using this in the codebase. But I will try to remove this..

packages/devextreme/tsconfig.json
tongsonbarbs: The runtime is not being compiled even __internal has its own tsconfig. So I added the runtime in the includes of the root's tsconfig to compile the runtime. But I will try to remove this too and see if it will work this time..

pomahtri (dismissed):

ivanblinov2k17:
We also need to remove/change dependency for demos https://github.com/DevExpress/DevExtreme/blob/25_1/apps/demos/package.json#L28
And check, if they work correctly both locally and in WG 
I propose putting force all test label on this PR to run through all the demos to make sure if nothing is missing

packages/devextreme/js/__internal/core/r1/config_context.ts
```diff
@@ -1,4 +1,4 @@
-import { createContext } from '@devextreme/runtime/inferno';
+import { createContext } from '@ts/core/r1/runtime/inferno/index';
```
ivanblinov2k17, line 1: Personally i dont like the reexport without package (import from index)
I propose to add typescript path alias (or at least use file specific imports)

ivanblinov2k17:
It's also a good idea to check all of instances of usage @devextreme/runtime package
I was able to find it in 
.github/renovate.json
systemjs-builder in testing 
RunSuite in testing
if we plan to remove the package completely, those places should be checked also

ivanblinov2k17 (dismissed):

packages/devextreme/js/__internal/core/r1/config_context.ts
pomahtri: Reexports are good, they allow us to make some files inside a folder private
E.g. you can create a lot of exports inside myfolder/utils.ts, but reexport only few of them in myfolder/index.ts

pomahtri:
> It's also a good idea to check all of instances of usage @devextreme/runtime package I was able to find it in .github/renovate.json systemjs-builder in testing RunSuite in testing if we plan to remove the package completely, those places should be checked also

that's necessary, but not in scope of current task

Ambrozy (dismissed):

pomahtri (dismissed):

pomahtri (approved):

Ambrozy (approved):

PR #29165, opened 2025-03-04
DropDownEditor, DropDownList, SelectBox, TagBox, DropDownBox: convert into ES6 class

EugeniyKiyashko:


packages/devextreme/js/__internal/ui/drop_down_editor/m_drop_down_editor.ts
```diff
@@ -925,30 +1031,30 @@ const DropDownEditor = TextBox.inherit({
...
 
-  content() {
+  content(): Element | null {
     return this._popup ? this._popup.content() : null;
-  },
-});
+  }
+}
 
 registerComponent('dxDropDownEditor', DropDownEditor);
```
marker-dao, line 1060: Why do we still use export at the end of the file when this can be shortened by writing export before the class definition?

export default DropDownEditor; ––> export default class DropDownEditor { ... }

I think we can do this where we don't use the include method for mixins. Am I correct in understanding that we eventually want to get rid of the include method?

marker-dao (dismissed):

packages/devextreme/js/__internal/ui/drop_down_editor/m_drop_down_editor.ts
EugeniyKiyashko: Now, there is no mechanism to exclude Mixins without reworking the code logic, so the mixin will continue to be used for some time. The card for changing its usage was created more than two years ago, and it is not part of the task related to typing directly.

```suggestion
export default DropDownEditor; ––> export default class DropDownEditor { ... }
```
I have no objections if the code owners decide to use this notation and change it in all places at once.

marker-dao (approved):

packages/devextreme/js/__internal/ui/drop_down_editor/m_drop_down_editor.ts
```diff
@@ -565,66 +638,71 @@ const DropDownEditor = TextBox.inherit({
...
 
-  _attachPopupKeyHandler() {
+  _attachPopupKeyHandler(): void {
+    // @ts-expect-error ts-error
     eventsEngine.on(this._popup.$overlayContent(), addNamespace('keydown', this.NAME), (e) => this._popupKeyHandler(e));
-  },
+  }
 
-  _popupKeyHandler(e) {
+  _popupKeyHandler(e): void {
```
r-farkhutdinov, line 652: Why don't we assign a type for the event parameter?

r-farkhutdinov (approved):

packages/devextreme/js/__internal/ui/drop_down_editor/m_drop_down_editor.ts
EugeniyKiyashko: I mention in the initial post details. 
Shortly:
the current task involves changes to make the code work with the new syntax. In the process, I can improve some types (to avoid a huge number of ts-expect-error), which can be resolved quickly while ensuring everything works as before without changing execution logic (refactoring) - were done a lot here.
Other improvements within your area will be handled by the squad in step 2. I sure that the squad understands the code in its own area better.

PR #29198, opened 2025-03-06
DataGrid - Scrolling position is reset to far right on an attempt to scroll left if the most right cell is focused using the keyboard (T1277214)

tongsonbarbs:


packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -315,9 +315,10 @@ export class KeyboardNavigationController extends modules.ViewController {
 
     if (!$focusedElement.length && e?.virtualColumnsScrolling) {
       const focusedColumnIndex = this._focusedCellPosition?.columnIndex ?? -1;
-      const focusedColumnIndexWithoutOffset = focusedColumnIndex - this._getFocusedColumnIndexOffset(focusedColumnIndex);
+      const visibleColumns = this._columnsController.getVisibleColumns();
+      const isColumnVisible = visibleColumns.some((column) => column.index === focusedColumnIndex);
```
Alyar666: `column.index` - This is the index that is equal to the column's index in the columns array. This does not match the visible index of the cell in the markup. You can read more about indexes [here](https://js.devexpress.com/jQuery/Documentation/Guide/UI_Components/DataGrid/Columns/Column_and_Row_Indexes/).

I suggest using the `_isColumnRendered` method to check if a cell is in the markup.

Alyar666 (dismissed):

Alyar666 (approved):

PR #29214, opened 2025-03-07
Stepper: enable active, hover, selected, focused states for steps

EugeniyKiyashko:


packages/devextreme/js/__internal/ui/stepper/stepper.ts
```diff
@@ -45,11 +63,20 @@ class Stepper extends CollectionWidgetAsync<StepperProperties> {
...
     $(this.element()).addClass(STEPPER_CLASS);
 
+    this._appendStepContainer();
+
     this._toggleOrientationClass();
 
     super._initMarkup();
   }
 
+  _appendStepContainer(): void {
```
anna-shakhova: let's use plural form, as in variable: _appendStepsContainer

anna-shakhova (dismissed):

anna-shakhova (approved):

PR #29216, opened 2025-03-07
CardView - Column Chooser

Tucchhaa:


packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/controller.ts
```diff
@@ -0,0 +1,98 @@
...
+        const newItems = sortedColumns.map((c, index) => ({
+          id: index,
+          columnName: c.name,
+          selected: c.visible,
+          text: c.caption,
+          disabled: !c.allowHiding,
+          visible: c.showInColumnChooser,
+        }) as TreeViewItemProperties);
+
+        const items = this.items?.unreactive_get();
```
Raushen, line 39: It is better to retrieve items by reactive way, and add `this.items` in dependencies

Raushen, line 39: Maybe it is better to create something like...
` private readonly itemsInternal: SubsGets<TreeViewItemProperties[]>;`
to separate states?

packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.ts
```diff
@@ -160,4 +160,25 @@ export class ColumnsController {
       return newColumns;
     });
   }
+
+  public updateColumns(func: (columns: PreNormalizedColumn[]) => PreNormalizedColumn[]): void {
```
Raushen, line 117: It seems there is a code duplication. Please [compare with my implementation](https://github.com/DevExpress/DevExtreme/pull/29202/files#diff-6d7ed67cee3b58c72be0825512be4cf5e4cdd55a131998e0c6b9841df03af6ccR129).

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+    this.mode = computed(
+      (v) => (v ?? 'select') as ColumnChooserMode,
+      [this.options.oneWay('columnChooser.mode')],
+    );
+
+    this.toolbarController.addDefaultItem(
+      computed(
+        (enabled) => ({
+          name: 'columnChooserButton',
```
Raushen, line 57: These options seem to be the options by default. Please check [my example](https://github.com/DevExpress/DevExtreme/pull/29202/files#diff-94016181098e5e4d1aa18f957394b8bcf6a0584520b0ca05815865afaf7afa74R16) how to define them separately of the view.

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,218 @@
...
+          },
+          [this.options.oneWay('columnChooser.title')],
+        ),
+
+        onShown: () => { this.setPopupAttributes(); },
+        onHidden: () => { this.popupVisible.update(false); },
+      } as MapMaybeSubscribable<PopupProperties>),
+
+      treeViewConfig: combined({
+        dataStructure: 'plain',
```
Raushen, line 147: Here there are default options also

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+    return combined({
+      popupRef: this.popupRef,
+      treeViewRef: this.columnChooserController.treeViewRef,
+
+      visible: this.popupVisible,
+      mode: this.mode,
+
+      popupConfig: combined({
+        shading: false,
```
Raushen, line 100: Can these default options be extracted?

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  private setPopupAttributes(): void {
+    const isSelectMode = this.mode.unreactive_get() === 'select';
+    const isBandColumnsUsed = false; // TODO: band columns aren't yet implemented in cardview
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const popup = this.popupRef.current as any;
+
+    popup.setAria({
+      role: 'dialog',
```
Raushen: Can it be included to default options too?

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,218 @@
...
+    );
+  }
+
+  public show(): void {
+    this.popupVisible.update(true);
+  }
+
+  public hide(): void {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    this.popupRef.current?.hide();
```
Raushen, line 83: How about:
`this.popupVisible.update(false);`
Does current approach update all options correctly?

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,218 @@
...
+
+  constructor(
+    private readonly toolbarController: ToolbarController,
+    private readonly columnChooserController: ColumnChooserController,
+    private readonly options: OptionsController,
+  ) {
+    super();
+
+    this.mode = computed(
+      (v) => (v ?? 'select') as ColumnChooserMode,
```
Raushen: This can be extracted to default options

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,218 @@
...
+      [this.options.oneWay('columnChooser.mode')],
+    );
+
+    this.toolbarController.addDefaultItem(
+      computed(
+        (enabled) => ({
```
Raushen: How about make the behavior consistent with DataGrid and pass `null` (for example) to the addDefaultItem method if column chooser is disabled?

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
Tucchhaa: If just use `popupVisible.update(false)`, then there will be no popup closing animation.
I think while this way isn't perfect it should be acceptable. 

jfyi popupVisible is set to false in [Popup.onHidden](https://github.com/DevExpress/DevExtreme/blob/514eb4bad377ffb435d5911e40fd178c3c972567/packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx#L143) event

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
Tucchhaa: Then `cardView.option('columnChooser.enabled', true)` will not affect anything. Another solution could be to subscribe on 'columnChooser.enabled' and add or remove defaultItem, but I implemented it this way because it's simpler.

packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.ts
Tucchhaa: I see, we implemented the same func :), but your code isn't merged yet, so I guess the one whose PR will merge later should resolve the merge conflict.

And I want to mention that in the [.updateColumn()](https://github.com/DevExpress/DevExtreme/blob/514eb4bad377ffb435d5911e40fd178c3c972567/packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.ts#L118-L130) func, there's visible indexes normalization performed after column update. Which I think we have to perform in .updateColumns too.

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
Tucchhaa: Thank you for pointing this out. Actually there's no need of this `computed`, I will remove it

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
Tucchhaa: Do you mean to create a variable that will contain default options?

But I'd keep all the options in the one place, rather than splitting it.

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/controller.ts
```diff
@@ -0,0 +1,115 @@
...
+      }
+      if (!column.visible && isSelected) {
+        treeView.unselectItem(index);
+      }
+    });
+
+    treeView.endUpdate();
+  }
+
+  private getChooserColumns(): Column[] {
```
Raushen: It seems it is possible to create smth like:
`public readonly orderedColumns: SubsGets<Column[]>;
orderedColumns = computed(
   calculationFunction,
   [this.options.oneWay('columnChooser.sortOrder'), this.columnsController.columns]
)
`

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
Raushen: Ok, the last chance, what if this.toolbarController.addDefaultItem could return a button instance so it could be updated here related on the enabled property?

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
Raushen: I'm not sure it is enough only to extract variables. There is a huge logic with calculating popup's props:
isMaterialOrGeneric(), getPopupWrapperClass(), setPopupAttributes() methods. I'd move all of this to a separate file.

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/controller.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        if (items?.length === columns.length) {
+          for (let i = 0; i < columns.length; i += 1) {
+            items[i].selected = columns[i].visible;
+          }
+
+          // if only column visibility changed, then we don't
+          // create new items, but just update selection
+          if (equalByValue(items, newItems)) {
+            this.updateSelection();
```
Raushen: I think it is important to use side effects only in the effect methods. I will allow us to build clear reactive constructions.

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/controller.ts
```diff
@@ -0,0 +1,115 @@
...
+        const columns = this.getChooserColumns();
+        const newItems = columns.map((c, index) => ({
+          id: index,
+          columnName: c.name,
+          selected: c.visible,
+          text: c.caption,
+          disabled: !c.allowHiding,
+        }) as TreeViewItemProperties);
+
+        const items = this.items?.unreactive_get();
```
Raushen: Maybe it is better to create a separate field for the previous state. Also it can be useful do it out of the box, please make a TODO note to include the previous state in our computed method.

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
Tucchhaa: There's no need in setting role='dialog', it is implemented out-of-the-box from popup. But we still have to set aria-label.

And we can't set attributes to the $content, through popup config. So we can't move this method to default options completely

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
Tucchhaa: I refactored toolbar addDefaultItem, now the second parameter needRender is passed to determine whether to render toolbar item or not

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/column_chooser.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  if (!visible) {
+    return null;
+  }
+
+  return (
+    <Popup
+      visible={visible}
+      componentRef={popupRef}
+      { ...popupConfig }
```
wdevfx, line 38: Let's not spread props and write it direct way
Because it's a little hard to understand wich props we passed when we spread them :)

E.g:
```tsx
  <Popup
    componentRef={popupRef}
    visible={visible}
    shading={false}
    showCloseButton={popupConfig.showCloseButton}
    // other props
/>
```

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/column_chooser.tsx
```diff
@@ -0,0 +1,46 @@
...
+    >
+      <TreeView
+        componentRef={treeViewRef}
+        { ...treeViewConfig }
```
wdevfx, line 42: Same comment for spreading props :)

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/controller.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  public readonly treeViewRef = createRef<dxTreeView>();
+
+  constructor(
+    private readonly columnsController: ColumnsController,
+    private readonly options: OptionsController,
+  ) {
+    this.chooserColumns = computed(
+      (columns, sortOrder) => {
+        let chooserColumns = columns.filter((c) => c.showInColumnChooser);
```
wdevfx: Minor comment.

Let's write full name of variable, ```c``` -> ```column``` to make it more obvious

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/controller.ts
```diff
@@ -0,0 +1,117 @@
...
+        this.columnsController.columns,
+        this.options.oneWay('columnChooser.sortOrder'),
+      ],
+    );
+
+    this.chooserColumns.subscribe((columns) => {
+      let onlyVisibleChanged = false;
+
+      this.items.updateFunc((oldItems) => {
+        const newItems = columns.map((c, index) => ({
```
wdevfx: Minor comment.

Let's write full name of variable, ```c``` -> ```column``` to make it more obvious

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/controller.ts
```diff
@@ -0,0 +1,117 @@
...
+        if (oldItems.length === columns.length) {
+          for (let i = 0; i < columns.length; i += 1) {
+            oldItems[i].selected = columns[i].visible;
+          }
+
+          if (equalByValue(oldItems, newItems)) {
+            onlyVisibleChanged = true;
+            return oldItems;
+          }
+        }
```
wdevfx: Let's make this ```items``` option update mutation in the ```TreeViewWrapper```
The main point here is that we don't want mutations in our new code, 
but it is sometimes required  for the correct work of old components
So, let's move this logic with mutation to the component wrapper

After this change I think it will be possible to convert ```items``` back to the ```computed```

An example of how it is implemented the in toolbar wrapper:
https://github.com/DevExpress/DevExtreme/blob/grids/cardview/draft/packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/toolbar.ts#L15

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/controller.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  const columnChooserController = new ColumnChooserController(columnsController, optionsController);
+
+  return {
+    columnChooserController,
+    columnsController,
+    optionsController,
+  };
+};
+
+describe('select mode: ', () => {
```
wdevfx: Let's change ```describe``` a little
Something like:

```ts
describe('ColumnChooser', () => {
  describe('Controller', () => {
    // test code
  });
});
```

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/controller.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  it('updates treeView items on applying column options', () => {
```
wdevfx: Let's cover by test ```onSelectionChanged``` method too

wdevfx: Looks like now not all cases from [this calculation](https://github.com/DevExpress/DevExtreme/pull/29216/files#diff-f80c12759b9cf3a1d5d0bd80393a1e5f7284ec6e592660ecf73aa1f405fc5ed6R44) covered

So, let's cover all cases (please take into account comment of [this line](https://github.com/DevExpress/DevExtreme/pull/29216/files#diff-f80c12759b9cf3a1d5d0bd80393a1e5f7284ec6e592660ecf73aa1f405fc5ed6R44) before)

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/public_methods.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable spellcheck/spell-checker */
```
wdevfx, line 1: Let's cover public methods by tests

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,221 @@
...
+import type { ColumnChooserProps } from './column_chooser';
+import { ColumnChooser } from './column_chooser';
+import { ColumnChooserController } from './controller';
+
+const COLUMN_CHOOSER_CLASS = 'column-chooser';
+const COLUMN_CHOOSER_BUTTON_CLASS = 'column-chooser-button';
+const COLUMN_CHOOSER_LIST_CLASS = 'column-chooser-list';
+const COLUMN_CHOOSER_PLAIN_CLASS = 'column-chooser-plain';
+const COLUMN_CHOOSER_DRAG_CLASS = 'column-chooser-mode-drag';
+const COLUMN_CHOOSER_SELECT_CLASS = 'column-chooser-mode-select';
```
wdevfx: Not a mandatory comment.

Let's merge all CSS-related const into one:
```ts
const CLASS = {
  root: 'column-chooser',
  // other
};
```

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,221 @@
...
+        showCheckBoxesMode: computed(
+          (v) => (v ? 'selectAll' : 'normal'),
+          [this.options.oneWay('columnChooser.selection.allowSelectAll')],
+        ),
+        selectByClick: this.options.oneWay('columnChooser.selection.selectByClick'),
+        onSelectionChanged: controller.onSelectionChanged.bind(controller),
+      };
+    }
+
+    return {};
```
wdevfx: Minor comment.

Let's invert logic and reduce complexity a little
```ts
if (!this.isSelectMode()) {
  return {}
}

// other code
```

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+        ...this.getTreeViewConfig(),
+      } as MapMaybeSubscribable<TreeViewProperties>),
+    });
+  }
+
+  protected getTreeViewConfig(): MapMaybeSubscribable<TreeViewProperties> {
+    const mode = this.mode.unreactive_get();
+
+    if (mode === 'select') {
```
wdevfx: Let's use here ```isSelectMode``` private method

```ts
if (!this.isSelectMode())
```

packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.ts
```diff
@@ -125,21 +125,34 @@ export class ColumnsController {
...
       return newColumns;
     });
   }
+
+  private normalizeColumnsVisibleIndexes(columns: PreNormalizedColumn[]): void {
+    const visibleIndexes = normalizeVisibleIndexes(
+      columns.map((c) => c.visibleIndex),
+    );
+
+    visibleIndexes.forEach((visibleIndex, i) => {
```
wdevfx, line 134: Let's not mutate ```columns``` arguments array and return new one from this function instead

wdevfx:
1) Let's write some e2e testcafe tests too
At least a few visuals (take screenshots) and some functional ones to ensure that all integrations work well :)

2) Let's create a storybook story, but we can postpone it till PR in the main branch

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/view.tsx
Tucchhaa: There will be dragAndDrop mode config soon, so I think there's no point of changing it

packages/devextreme/js/__internal/grids/new/grid_core/toolbar/controller.ts
```diff
@@ -38,25 +40,27 @@ export class ToolbarController {
...
-        ...oldDefaultItems,
-        [item.name]: item,
-      }));
-    });
-  }
+    this.itemSubscriptions[name] = effect(
+      (item, needRender) => {
+        this.defaultItems.updateFunc((oldDefaultItems) => {
+          if (needRender) {
+            oldDefaultItems[item.name] = item;
```
Raushen: Let's avoid the argument mutation here

packages/devextreme/js/__internal/ui/overlay/m_overlay.ts
```diff
@@ -406,6 +406,7 @@ class Overlay<
       const tabbableElements = overlayStack[i]._findTabbableBounds();
 
       if (tabbableElements.first || tabbableElements.last) {
+        // @ts-ignore
```
Raushen: Roman asked to apply this change in a separate request

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/column_chooser.tsx
Tucchhaa: I think it's a matter of taste. I feel current way is more convenient, because if there's no spreading operator, then there will be a duplication of code which can be annoying to add/remove props in both places all the time.

Moreover, I think it's easy to navigate to the getProps() and see all the props passed.

wdevfx: Discussed with @pomahtri and decided to left it as is for a while

e2e/testcafe-devextreme/tests/cardView/columnChooser/api.functional.ts
```diff
@@ -0,0 +1,26 @@
+import CardView from 'devextreme-testcafe-models/cardView';
+import url from '../../../helpers/getPageUrl';
+import { createWidget } from '../../../helpers/createWidget';
+
+fixture.disablePageReloads`CardView - ColumnChooser.Functional`
+  .page(url(__dirname, '../../container.html'));
+
+test('public methods showColumnChooser and hideColumnChooser', async (t) => {
```
wdevfx: Not critical (we can postpone it).

Let's split this one test case into two separate
It's better to have smaller, accurate test cases than one that checks a few things

e2e/testcafe-devextreme/tests/dataGrid/common/columnChooser.ts
```diff
@@ -54,7 +54,7 @@ test('Column chooser checkboxes should be aligned correctly with plain structure
   const columnChooser = dataGrid.getColumnChooser();
 
   await t
-    .expect(await takeScreenshot('column-chooser-checkbox-alignment-plain-structure.png', columnChooser.content))
+    .expect(await takeScreenshot('column-chooser-checkbox-alignment-plain-structure.png', columnChooser.popup.element))
```
wdevfx: It will be better to move the existing ```ColumnChooser``` POM into a separate PR into ```25_1``` branch
Because it may affect the existing ```DataGrid``` functionality and lead to some conflicts with main branches in the future

So, let's do the following:
- Create a separate PR with POM change and ```DataGrid``` e2e test related changes
- Merge it into ```25_1```
- After it I or @pomahtri will merge ```25_1``` in ```main``` branch
- I or @pomahtri will merge ```main``` into ```draft``` branch
- U will merge ```draft``` into ur feature branch

@pomahtri cc

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/options.ts
```diff
@@ -0,0 +1,12 @@
+import type { ColumnChooser, ColumnChooserSelectionConfig } from '@js/common/grids';
```
wdevfx, line 1: Discussed and decided that we should add integration tests for public options

packages/devextreme/js/__internal/grids/new/grid_core/column_chooser/utils.test.ts
```diff
@@ -0,0 +1,126 @@
+/* eslint-disable spellcheck/spell-checker */
```
wdevfx: Not critical (we can postpone it)

Let's rename this file to ```test-utils.test.ts```  / ```test-helpers.test.ts``` or sth like that to make it super obvious

packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/tree_view.ts
```diff
@@ -7,4 +8,65 @@ export class TreeView extends InfernoWrapper<TreeViewProperties, dxTreeView> {
...
+
+    this.component?.getScrollable()?.scrollTo({ top: scrollTop });
+  }
+
+  private isItemsOnlySelectionChanged(
+    prevProps: TreeViewProperties,
+    props: TreeViewProperties,
+  ): boolean {
+    const oldItems = prevProps.items ?? [];
+    const newItems = props.items ?? [];
```
wdevfx: Let's not mutate props and get rid of ```selection``` field mutation by the following change:

```ts
const oldItems = (prevProps.items ?? []).map(({...restItemFields, selected}) => ({...restItemFields});
const newItems = (prevProps.items ?? []).map(({...restItemFields, selected}) => ({...restItemFields});

// and we don't need the selected mutation because we exclude these fields from items
```

Tucchhaa: Wow, that's a really nice way to do that. Thank you :)

e2e/testcafe-devextreme/tests/dataGrid/common/columnChooser.ts
Tucchhaa: Because ColumnChooser POM is related to gridCore POM, we decided to revert POM changes and apply them after cardView is merged to 25.1.

wdevfx (approved):

PR #29221, opened 2025-03-07
Spike: HTMLEditor AI Features UI

r-farkhutdinov:


packages/devextreme/playground/jquery.html
```diff
@@ -42,22 +49,171 @@
...
+
+                const editor = $('#editor').dxHtmlEditor({
+                    height: 400,
+                    value: princ,
+                    toolbar: {
+                        items: [
+                            {
+                                name: 'ai',
+                                options,
+                                widget: 'dxDropDownButton'
```
marker-dao, line 207: When using the name 'ai', the component is supposed to figure out which widget to use, similar to the 'undo', 'redo' buttons. I.e. the user does not have to specify the widget field here.

packages/devextreme/js/__internal/ui/html_editor/modules/m_toolbar.ts
```diff
@@ -311,6 +312,9 @@ if (Quill) {
 
         return this._getToolbarItem(buttonItemConfig);
       }
+      if (item.name && this._isAcceptableItem(item.widget, 'dxDropDownButton')) {
```
marker-dao, line 315: If you leave the code as it is, any element containing a widget field with the value 'dxDropDownButton' will be configured with the above instructions. It is best to encapsulate ai separately.

PR #29252, opened 2025-03-12
🎴 CardView - filterPanel - implement

Raushen:


packages/devextreme/js/__internal/grids/grid_core/filter/filter_panel_clear_filter_button.tsx
```diff
@@ -0,0 +1,56 @@
...
+    this.clearFilter();
+  };
+
+  public render(): JSX.Element {
+    return (
+        <>
+            <div ref={this.ref}
+              className={this.props.className}
+              tabIndex={this.props.tabIndex}
+              onClick={(): void => this.props.clearFilter()}
```
wdevfx: Let's use ```clearFilter``` here too

```suggestion
              onClick={this.clearFilter}
```

packages/devextreme/js/__internal/grids/grid_core/filter/filter_panel_clear_filter_button.tsx
```diff
@@ -0,0 +1,56 @@
...
+import * as accessibility from '@js/ui/shared/accessibility';
+import { Component, createRef } from 'inferno';
+
+export interface Props {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  component: any;
+
+  className: string;
+
+  tabIndex: number;
```
wdevfx: Question.

Is ```tabIndex``` may be not equal zero?
If it cannot, let's remove this prop :)

packages/devextreme/js/__internal/grids/grid_core/filter/filter_panel_clear_filter_button.tsx
```diff
@@ -0,0 +1,56 @@
...
+              role='button'
+            >
+              {this.props.text}
+            </div>
+        </>
+    );
+  }
+
+  public componentDidMount(): void {
+    accessibility.registerKeyboardAction(
```
wdevfx: Let's unsubscribe from this helper's events
It returns a function for this:
https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/ui/shared/accessibility.js#L151

packages/devextreme/js/__internal/grids/new/grid_core/core/view.tsx
```diff
@@ -45,6 +45,16 @@ export abstract class View<T extends {}> {
...
 
+  public isCompatibilityMode(): boolean {
+    return false;
+  }
+
+  // eslint-disable-next-line @stylistic/max-len
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/no-unused-vars, @typescript-eslint/explicit-module-boundary-types
+  public optionChanged(_args: any): void {
+
+  }
```
wdevfx, line 60: I think this empty ```optionChanged``` needed for some reason
But it's very hard to guess why we need an empty method

So, let's add a comment here
```ts
// NOTE: We need this empty method ...
```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_controller.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  public static dependencies = [OptionsController] as const;
+
+  public readonly displayFilter = computed(
+    (filter, filterEnabled) => {
+      if (!filterEnabled) {
+        return null;
+      }
+      return filter;
+    },
```
wdevfx: Minor comment.

We can write it shorter:
```ts
(filter, filterEnabled) => !filterEnabled ? null : filter
```

packages/devextreme/js/__internal/grids/grid_core/filter/filter_panel_clear_filter_button.tsx
Raushen: Yes, the KBN can be turned off, so tabIndex could be equal -1

wdevfx (approved):

PR #29254, opened 2025-03-12
Map: Support custom route modes

Zedwag:


packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
```diff
@@ -48,7 +48,12 @@ class AzureProvider extends DynamicProvider {
       driving: 'car',
       walking: 'pedestrian',
     };
-    return movementTypes[type] || movementTypes.driving;
+
+    if (!isDefined(type)) {
+      return movementTypes.driving;
+    }
+
+    return movementTypes[type] || type;
```
r-farkhutdinov: Minor suggestion, this will be semantically more correct (and in the files below too)
```suggestion
    return movementTypes[type] ?? type;
```

Zedwag: updated

marker-dao (dismissed):

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.bing.ts
```diff
@@ -53,7 +53,12 @@ class BingProvider extends DynamicProvider {
       driving: Microsoft.Maps.Directions.RouteMode.driving,
       walking: Microsoft.Maps.Directions.RouteMode.walking,
     };
-    return movementTypes[type] || movementTypes.driving;
+
+    if (!type) {
+      return movementTypes.driving;
+    }
+
+    return movementTypes[type] ?? Microsoft.Maps.Directions.RouteMode[type];
```
ksercs: [excess code]
Looks like we don't need `movementTypes` at all here, no?

r-farkhutdinov (dismissed):

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.bing.ts
Zedwag: i wanted to keep the same content for all 3 providers, but updated if u think it's better

ksercs (approved):

PR #29257, opened 2025-03-12
Update and fix TS in Demos.

GoodDayForSurf:
Add CI TS check for Vue


apps/demos/utils/demo-vue-loader.js
```diff
@@ -0,0 +1,11 @@
+const vueLoader = System.import("vue-loader");
+
+module.exports.translate = function (data) {
```
mpreyskurantov: Implement without non es5 / browser keywords.
Why only for vue? Will it work with other (react - tsx, angular -ts, etc.) inputs with `import { type ABC, } from path`?
Blocked until 24.2.6 / WG approval

mpreyskurantov (changes requested):

apps/demos/utils/demo-vue-loader.js
GoodDayForSurf: only for vue. In react all demos tests ok

mpreyskurantov: Nice, let's trim `type` from `import` inside https://github.com/DevExpress/dx-systemjs-vue-browser then

mpreyskurantov (dismissed):

mpreyskurantov (approved):

pomahtri (approved):

wdevfx (approved):

PR #29262, opened 2025-03-12
Stepper: add navigation modes

anna-shakhova:


packages/devextreme/js/__internal/ui/collection/collection_widget.base.ts
```diff
@@ -112,6 +112,8 @@ export interface CollectionWidgetBaseProperties<
   _itemAttributes?: Record<string, unknown>;
 
   selectOnFocus?: boolean;
+
+  loopItemFocus?: boolean;
```
ksercs, line 116: What is this for? I don't see any place of using

packages/devextreme/testing/tests/DevExpress.ui.widgets/stepper.tests.js
```diff
@@ -62,6 +62,87 @@ QUnit.module('Initialization', moduleConfig, () => {
...
+        this.getItems().eq(3).trigger('dxclick');
+
+        assert.equal(this.instance.option('selectedIndex'), 1, 'selectedIndex not changed');
+
+        this.getItems().eq(2).trigger('dxclick');
+
+        assert.equal(this.instance.option('selectedIndex'), 2, 'selectedIndex changed');
+    });
+
+    [true, false].forEach((linear) => QUnit.test(`selectionChanged callback should not be triggered when is already selected, linear=${linear}`, function(assert) {
```
ksercs: [code style]
Could you please update formatting to stop this function return result of ` QUnit.test`, how we do it in another places? It's more readable for me
```suggestion
    [true, false].forEach((linear) => {
        QUnit.test(`selectionChanged callback should not be triggered when is already selected, linear=${linear}`, function(assert) {
```

packages/devextreme/js/__internal/ui/stepper/stepper.ts
```diff
@@ -85,11 +99,13 @@ class Stepper extends CollectionWidgetAsync<StepperProperties> {
     return {
       ...super._getDefaultOptions(),
       orientation: 'horizontal',
+      linear: true,
```
EugeniyKiyashko, line 108: Option change processing is missing. Please add it to optionChanged() and add tests to cover scenarios

packages/devextreme/js/__internal/ui/stepper/stepper.ts
```diff
@@ -142,6 +158,37 @@ class Stepper extends CollectionWidgetAsync<StepperProperties> {
...
+  }
+
+  _itemPointerDownHandler(e: DxEvent): void {
+    if (!this._shouldPreventItemEvent(e.currentTarget)) {
+      super._itemPointerDownHandler(e);
+    }
+  }
+
+  _itemSelectHandler(e: DxEvent): void {
+    if (this.isItemSelected(e.currentTarget) || this._shouldPreventItemEvent(e.currentTarget)) {
```
EugeniyKiyashko: Let's try 'selectionRequired' option instead. It should do the same

packages/devextreme/testing/tests/DevExpress.ui/defaultOptions.tests.js
```diff
@@ -1547,11 +1547,13 @@ testComponentDefaults(Stepper,
     {},
     {
         orientation: 'horizontal',
+        linear: true,
         selectionMode: 'single',
         selectOnFocus: true,
         activeStateEnabled: true,
         hoverStateEnabled: true,
         focusStateEnabled: true,
+        loopItemFocus: false,
```
EugeniyKiyashko, line 1556:
```suggestion
        loopItemFocus: false,
        selectionRequired: true,
```

packages/devextreme/js/__internal/ui/collection/collection_widget.base.ts
anna-shakhova: it is just typings, we redefine default value for this collection property, but it was absent in type and caused ts error

packages/devextreme/testing/tests/DevExpress.ui.widgets/stepper.tests.js
anna-shakhova: fixed

packages/devextreme/js/__internal/ui/stepper/stepper.ts
anna-shakhova: changed

packages/devextreme/testing/tests/DevExpress.ui/defaultOptions.tests.js
anna-shakhova: added

packages/devextreme/js/__internal/ui/stepper/stepper.ts
anna-shakhova: added

EugeniyKiyashko (approved):

PR #29268, opened 2025-03-13
fix(react-wrapper): fixed T1279646

sergey-kras:


apps/react/examples/Examples.tsx
```diff
@@ -51,7 +52,7 @@ const Examples = () => {
 
             {/* <FormExample /> */}
 
-            <ListExample />
+            {/*<ListExample />*/}
```
VasilyStrelyaev: Can we return this example? Or is there a reason it is still commented out?

apps/react/runner.js
```diff
@@ -1,17 +1,25 @@
...
 createTestCafe('127.0.0.1')
     .then(tsc => {
         testCafe = tsc;
-
         const runner = tsc.createRunner();
+
         return runner
-            .src('test.js')
-            .browsers('chrome:headless --disable-gpu --window-size=1200,800')
+            .src(path.join(__dirname, './__test__/*.ts'))
```
VasilyStrelyaev: Do we need these changes to the runner since we are not adding new tests in this PR?

apps/react/examples/Examples.tsx
sergey-kras: Yes, reverted

apps/react/runner.js
sergey-kras: reverted

VasilyStrelyaev (approved):

chaosmirage (approved):

PR #29310, opened 2025-03-17
CardView - create .d.ts (1st part)

pomahtri:


pomahtri:
Moved to https://github.com/DevExpress/DevExtreme/pull/29455

PR #29318, opened 2025-03-17
Inferno - update to v8

pomahtri:
Commit-by-commit:

- updated inferno to v8 (NOTE: it is not last version anymore, the last is v9 already)
- moved inferno version to pnpm's catalog to prevent duplication and desyncing of version in different packages
- added `jsx.ts` file to provide typing for `JSX.Element`. I don't know why they deleted it in inferno, but without it we don't receive enough strictness in TS (any component becomes `any` and doesn't check its properties)
- updated updated `r1/types`, resolved pending "todo in v8 inferno"
- applied v8 inferno's BC: renamed `findDOMFromVNode` util
- updated `/r1/runtime/inferno/mocked/hydrate.ts` file following the https://github.com/infernojs/inferno/blob/master/packages/inferno-hydrate/src/index.ts as reference
- removed usage of removed global types `CSSProperties` and `HTMLAttribues`
- removed usage of `className` on non-html components (as we checked with @wdevfx, it didn't have any affect even before)


packages/devextreme/js/__internal/scheduler/r1/components/base/row.tsx
```diff
@@ -61,9 +61,9 @@ export class Row extends BaseInfernoComponent<RowProps> {
           hasRightVirtualCell
           && rightVirtualCellCount != null
           && splitNumber(rightVirtualCellCount, MAX_COL_SPAN).map(
-            // @ts-ignore
-            (colSpan, index) => <VirtualCell
-              className={`right-virtual-cell-${index}`}
+            (colSpan) => <VirtualCell
+              // TODO: do we need this?
+              // className={`right-virtual-cell-${index}`}
```
Ambrozy, line 66: I think better to just remove **className**, because we don't use it and it affects on nothing.

Ambrozy (approved):

packages/devextreme/js/__internal/scheduler/r1/components/base/row.tsx
pomahtri: It was @wdevfx's suggestion, please discuss it in squad if there's a need

VasilyStrelyaev (approved):

wdevfx (approved):

packages/devextreme/js/__internal/scheduler/r1/components/base/row.tsx
wdevfx: Hi @Ambrozy, @pomahtri 

The ```VirtualCell``` component as far as I can see never have the ```className``` prop
We checked with @pomahtri that Inferno cannot add a class to the component root automatically
So, it looks like this functionality never worked

But, this code goes from the R1 -> Inferno migration (very old) and we may miss something
So, I asked to comment this line instead of deleting just in case for future cases)

PR #29321, opened 2025-03-17
Overlay: tab should focus next focusable element if target is deleted (T1248343)

r-farkhutdinov:

### DIFF
--- a/packages/devextreme/js/__internal/ui/overlay/m_overlay.ts
+++ b/packages/devextreme/js/__internal/ui/overlay/m_overlay.ts
@@ -783,47 +783,47 @@ class Overlay<
   _findTabbableBounds(): { first: dxElementWrapper | null; last: dxElementWrapper | null } {
     const $elements = this._$wrapper.find('*');
     const elementsCount = $elements.length - 1;
-    const result = { first: null, last: null };
 
-    for (let i = 0; i <= elementsCount; i++) {
+    let first = null;
+    let last = null;
+
+    for (let i = 0; i <= elementsCount; i += 1) {
       // @ts-expect-error ts-error
-      if (!result.first && $elements.eq(i).is(tabbable)) {
+      if (!first && $elements.eq(i).is(tabbable)) {
         // @ts-expect-error ts-error
-        result.first = $elements.eq(i);
+        first = $elements.eq(i);
       }
       // @ts-expect-error ts-error
-      if (!result.last && $elements.eq(elementsCount - i).is(tabbable)) {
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/ui/overlay/m_overlay.ts
```diff
@@ -814,16 +816,20 @@ class Overlay<
...
-    const isOutsideTarget = !domUtils.contains(this._$wrapper.get(0), e.target);
 
-    if (isTabOnLast || isShiftTabOnFirst
-            || isEmptyTabList || isOutsideTarget) {
+    const activeElement = domUtils.getDeepActiveElement();
+
+    const isTabOnLast = !e.shiftKey && activeElement === $lastTabbable?.get(0);
+    const isShiftTabOnFirst = e.shiftKey && activeElement === $firstTabbable?.get(0);
+    const isEmptyTabList = Object.values(tabbableElements).filter(Boolean).length === 0;
+    const isOutsideTarget = !domUtils.contains(this._$wrapper.get(0), activeElement);
```
ksercs: I see that `isOutsideTarget` worked incorrectly but I actually don't see why it's necessary. It looks like information "tab on last" / "shift+tab on first" should be enough to loop when needed, no? Isn't it possible that we can just remove this condition, scenario is fixed but nothing is changed?

packages/devextreme/js/__internal/core/utils/m_dom.ts
```diff
@@ -166,6 +166,16 @@ export const isElementInDom = ($element) => {
   return !!$(shadowHost || element).closest(getWindow().document).length;
 };
 
+export function getDeepActiveElement(doc: Document = document): Element | null {
```
ksercs: I have doubts about this new utility:
 - If existing `getActiveElement` does not work well in shadowDOM we should fix it, not add a new function. 
-  But I see a code related to shadowDOM like `getRootNode` in `getActiveElement`. Cannot it be an issue in the test? 
- Also it's strange to see while - is really active element could be deeper then first level? I don't remember we thought about such cases when was supporting shadowDOM
- As far as i remember `utils/dom` is exactly for helper functions, but such staff as getters/setters/checkers are in `dom_adapter`
- If you add a new util, it should be tested on `core` level + review from owners

I believe there should not be a new utility. Kindly check again cannot we do it another way

ksercs (dismissed):

ksercs (approved):

Zedwag (approved):

PR #29325, opened 2025-03-18
CardView: Implement Selection

Alyar666:


packages/devextreme/js/__internal/grids/new/card_view/content_view/content/card/card.tsx
```diff
@@ -1,12 +1,14 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-/* eslint-disable @typescript-eslint/prefer-nullish-coalescing */
-
+import { isCommandKeyPressed } from '@js/common/core/events/utils/index';
+import { off, on } from '@js/events';
```
pomahtri: please add '/index', otherwise qunit will fail

packages/devextreme/js/__internal/grids/new/card_view/content_view/content/card/card.tsx
```diff
@@ -54,15 +59,23 @@ export interface CardProps {
 
   toolbar?: CardHeaderItem[];
 
+  width?: number;
```
pomahtri: looks like mistake after merge conflict solving. Do we need it?

packages/devextreme/js/__internal/grids/new/grid_core/selection/const.ts
```diff
@@ -0,0 +1,17 @@
+export enum SelectionMode {
```
pomahtri, line 1: better not to use enums, only type unions
`export type SelectionMode = 'single' | 'none' ` etc
And better to import existing type from .d.ts

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  private updateSelectionToolbarButtons(
+    selectedCardKeys: SelectedCardKeys,
+    selectionOption: SelectionOptions,
+  ) {
+    if (selectionOption.mode === SelectionMode.Multiple && selectionOption.allowSelectAll) {
+      const isSelectAll = this.isSelectAll();
+      const isOnePageSelectAll = this.isOnePageSelectAll();
+
+      this.toolbarController.addDefaultItem({
```
pomahtri: let's use approach from @Tucchhaa's PR to hide items

pomahtri (approved):

PR #29331, opened 2025-03-18
Update demos to use platform-specific imports

DuckTieCorpMember:
Change Angular/Vue/React demos to use platform-specific common imports. For example DataSource, CustomStore etc.


apps/demos/Demos/Charts/BarSparklines/Angular/app/app.component.ts
```diff
@@ -5,8 +5,8 @@ import { HttpClient, HttpClientModule } from '@angular/common/http';
 import { lastValueFrom } from 'rxjs';
 
 import { DxSparklineModule, DxSelectBoxModule } from 'devextreme-angular';
-import DataSource from 'devextreme/data/data_source';
-import CustomStore from 'devextreme/data/custom_store';
+import { DataSource } from 'devextreme-angular/common/data';
+import { CustomStore } from 'devextreme-angular/common/data';
```
GoodDayForSurf: it's better to replace by import { DataSource, CustomStore  } from 'devextreme-angular/common/data';
also in other demos

apps/demos/Demos/Lookup/Basics/Angular/app/app.component.ts
```diff
@@ -2,8 +2,8 @@ import { NgModule, Component, enableProdMode } from '@angular/core';
 import { BrowserModule } from '@angular/platform-browser';
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { DxLookupModule } from 'devextreme-angular';
-import DataSource from 'devextreme/data/data_source';
-import ArrayStore from 'devextreme/data/array_store';
+import { DataSource } from 'devextreme-angular/common/data';
+import { ArrayStore } from 'devextreme-angular/common/data';
```
VasilyStrelyaev: Great job!

I have just one note I did not pay attention before - it's cosmetic, but since these are demos, it's nevertheless important: there should not be two lines that import from the same module. In this PR, there is a number of files that violate this rule. Lint should have caught it, but for some reason it didn't.

```
// Any occurrence of

import { a } from 'foo/bar';
import { b } from 'foo/bar';

// should be refactored to

import { a, b } from 'foo/bar';
```

Looks like all such occurrences import from /common/data, so maybe if you search for `/common/data';` and notice which files have more than one result, you will quickly find where to fix.

VasilyStrelyaev (approved):

GoodDayForSurf (approved):

Raushen (approved):

Ambrozy (approved):

wdevfx (approved):

PR #29337, opened 2025-03-18
TreeList - The W1025 warning occurs when dragging a row even if the component's height is specified. (T1280519)

tongsonbarbs:


wdevfx:
Let's cover these changes by tests

packages/devextreme/js/__internal/grids/grid_core/row_dragging/m_row_dragging.ts
```diff
@@ -213,6 +217,7 @@ const rowsView = (Base: ModuleType<RowsView>) => class RowsViewRowDraggingExtend
         fixed: column.fixed,
         fixedPosition: column.fixedPosition,
       })),
+      isDragging,
```
Tucchhaa: Is it possible to just write: `isDragging: true`? I suggest to do that because this options is passed only to draggable grid, so it should be always true. And there will be no need to track dragging state in `onDragStart` and 'onDragEnd'

packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts
```diff
@@ -1789,10 +1789,15 @@ export const rowsView = (Base: ModuleType<RowsView>) => class VirtualScrollingRo
     super.setLoading.call(this, isLoading, messageText);
   }
 
+  private isRowDragging() {
+    return this.component.option('isDragging');
+  }
```
Tucchhaa: Can we rename this method to 'isDraggableGrid' and move it to another place? Doesn't seem correct to me to place this kind of method in VirtualScrolling module

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/rowDragging.tests.js
```diff
@@ -197,6 +197,7 @@ QUnit.module('Drag and Drop rows', moduleConfig, () => {
             pager: {
                 visible: false
             },
+            rootElementHeight: undefined,
```
Tucchhaa: seems that we don't need this anymore

Tucchhaa (approved):

wdevfx (approved):

PR #29344, opened 2025-03-18
Stepper: split variables values to support compact mode

EugeniyKiyashko:


packages/devextreme-scss/scss/widgets/base/stepper/layout/step/_index.scss
```diff
@@ -1,6 +1,19 @@
 .dx-step {
   position: relative;
   cursor: pointer;
+
+  &.dx-step-completed,
```
anna-shakhova, line 5: completed steps have font-weight 400

EugeniyKiyashko: According to designer's remarks now it should be 600

anna-shakhova (approved):

PR #29345, opened 2025-03-18
AI: Init, add .d.ts, add separate build, add tests

marker-dao:


packages/devextreme/js/__internal/core/ai/templates/index.ts
```diff
@@ -0,0 +1,6 @@
+export const templates = {
```
marker-dao, line 1: Add type

packages/devextreme/js/__internal/core/ai/__tests__/core/ai.test.ts
pomahtri, line 1: we don't use `__tests__` subfolders anymore, we make `a.ts` and `a.test.ts` files in same dir

marker-dao: @pomahtri Okay, are you going to remove this from scheduler --> workspaces --> view_model --> __tests__?

pomahtri: We are

apps/demos/Demos/Chat/AIAndChatbotIntegration/Angular/app/app.component.ts
```diff
@@ -55,6 +58,21 @@ export class AppComponent {
...
+          promise: Promise.resolve('AI response'),
+        }
+      },
+    };
+    const ai: AIClass = new AI(provider);
+
+    ai.translate(
+      { text: 'text', lang: 'lang' },
+      { onComplete: (result) => { console.log(result) } },
+    );
```
ksercs: What is this for? Looks like a spike you forgot to remove

packages/devextreme/js/__internal/core/ai/commands/translate.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      expect(promptManager.buildPrompt).toHaveReturnedWith({
+        system: 'You are a translation assistant, who speaks French at a native level.',
+        user: 'Translate "text to translate" to French language.',
+      });
+
+      expect(typeof abort).toBe('function');
+      expect(sendRequestSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('pulls onChunk callback when getting a chunk', () => {
```
ksercs: Aren't these last 3 tests a pure duplicate of tests for base command? I see no specific code in execute for this command'd code nor in the test

packages/devextreme/js/__internal/core/ai/core/request_manager.ts
```diff
@@ -0,0 +1,32 @@
+import type { AIProvider, Prompt, RequestCallbacks } from '@js/ai';
+
+export const ERROR_MESSAGE = {
```
ksercs:
```suggestion
export const ERROR_MESSAGES = {
```

packages/devextreme/js/__internal/core/ai/core/request_manager.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      const onChunkSpy = jest.fn();
+
+      requestManager.sendRequest({ user: 'test' }, { onChunk: onChunkSpy });
+
+      expect(onChunkSpy).toHaveBeenCalledTimes(2);
+      expect(onChunkSpy).toHaveBeenNthCalledWith(1, 'AI');
+      expect(onChunkSpy).toHaveBeenNthCalledWith(2, ' response');
+    });
+
+    it('after completion of the promis calls onComplete with accumulated data', async () => {
```
ksercs:
```suggestion
    it('after completion of the promise calls onComplete with accumulated data', async () => {
```

apps/demos/Demos/Chat/AIAndChatbotIntegration/Angular/app/app.component.ts
marker-dao: [6fc0a500f05c7afd090d61a12403b7d8ea7ea0b4]

packages/devextreme/js/__internal/core/ai/commands/translate.test.ts
marker-dao: [bd71b24e153307161edc9fa3c795a2b595726f4e]

packages/devextreme/js/__internal/core/ai/core/request_manager.ts
marker-dao: [bd71b24e153307161edc9fa3c795a2b595726f4e]

packages/devextreme/js/__internal/core/ai/core/request_manager.test.ts
marker-dao: [bd71b24e153307161edc9fa3c795a2b595726f4e]

packages/devextreme-angular/src/common/index.ts
```diff
@@ -74,6 +74,7 @@ export {
     VerticalEdge,
 } from 'devextreme/common';
 
+export * as AiIntegration from './ai-integration';
```
marker-dao, line 77: No need to pay attention to this, it's internal auto-generated code. This type is eventually named AIIntegration, etc.

packages/devextreme/js/common/ai-integration.d.ts
```diff
@@ -0,0 +1,140 @@
...
+   */
+  onChunk: (chunk: string) => void;
+}
+
+/**
+ * @docid
+ * @namespace DevExpress.aiIntegration
+ * @public
+ */
+export interface ResponseParams {
```
IlyaKhD: How about just `Response` ? At 1st glance I wouldn't expect from a response obj to have params. Parameters are like input, but reponse is output.
```suggestion
export interface ResponseParams {
```

packages/devextreme/js/common/ai-integration.d.ts
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * @docid
+   * @public
+   */
+  prompt: Prompt;
+  /**
+   * @docid
+   * @public
+   */
+  onChunk: (chunk: string) => void;
```
IlyaKhD, line 34: If I got it right, both `onChunk` and `Response.promise` are used for getting the result of the request.

It feels a bit unintuitive that these two things (".onChunk" and `.promise`) are separated and belong to different counterparts.


In my last experience with paginated responses,  you'd get a promise or a special object as a result. After that, if pagination is used you'd process this special object assuming that you get the data page by page until everything is received.

packages/devextreme/js/common/ai-integration.d.ts
```diff
@@ -0,0 +1,140 @@
...
+ * @docid
+ * @namespace DevExpress.aiIntegration
+ * @public
+ */
+export interface ResponseParams {
+  /**
+   * @docid
+   * @public
+   */
+  promise: Promise<string>;
```
IlyaKhD, line 47: **Minor**
In general it's not a good idea to name entites as the types, though it's not a big deal in this case

```suggestion
  result: Promise<string>;
```

packages/devextreme/js/common/ai-integration.d.ts
```diff
@@ -0,0 +1,140 @@
...
+   * @docid
+   * @param provider
+   */
+  constructor(provider: AIProvider);
+  /**
+   * @docid
+   * @publicName translate(params, callbacks)
+   * @public
+   */
+  translate(params: TranslateCommandParams, callbacks: RequestCallbacks): () => void;
```
IlyaKhD, line 109: **Minor** **Question**

What is the returned ` () => void` used for?

packages/devextreme/js/common/ai-integration.d.ts
```diff
@@ -0,0 +1,140 @@
...
+ * @public
+ */
+export type TranslateCommandResult = string;
+
+/**
+ * @docid
+ * @namespace DevExpress.aiIntegration
+ * @public
+ */
+export type BaseCommandResult = TranslateCommandResult;
```
IlyaKhD: As both aliases point to `string` - maybe we could keep them private for now?
Are their specific cases where they are expected to be used in the end project?

```suggestion
/**
 * @namespace DevExpress.aiIntegration
 */
export type TranslateCommandResult = string;

/**
 * @namespace DevExpress.aiIntegration
 */
export type BaseCommandResult = TranslateCommandResult;
```

packages/devextreme/js/common/ai-integration.d.ts
```diff
@@ -0,0 +1,140 @@
...
+   */
+  abort: () => void;
+}
+
+/**
+ * @docid
+ * @namespace DevExpress.aiIntegration
+ * @public
+ */
+export interface TranslateCommandParams {
```
IlyaKhD: If it's not a problem with generation or usage scenaros, we still should follow this rule:

> All public types are `type`s (type aliases, not `interface`s) to disallow [interface merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces) on end project.
> It's also because type aliases are better suited for creating a public facade. In the future if we need to change a public type from `interface` to `type` (e.g. to make it a discriminated union) we will add potential BC. On the other hand, converting from `type` to `interface` is expected to be BC-less

```suggestion
export type TranslateCommandParams = {
```

IlyaKhD: Same for other public `interface`'s: Prompt, RequestParams, ResponseParams, RequestCallbacks, interface, AIProvider

packages/devextreme/js/common/ai-integration.d.ts
marker-dao: I have no objection. [ff70e3116f89e204c3de3bfcdd26b9ab5ba78816]

packages/devextreme/js/common/ai-integration.d.ts
marker-dao: The data we want to retrieve is just a string. We don't want complex objects and we don't want to write logic to handle complex objects, also I got rid of generators (*yield) to make the code very easy to understand. Everything is set up this way because a query can be a streaming query, or it can be a single-shot query. The two methods (regular and streaming) were combined into one sendRequest method because they were virtually indistinguishable.

The streaming method returns a part of the string each time (simple use of promise is not enough here, and I deliberately did not use a generator). Callbacks solve these problems.

If I misunderstood you, please clarify the question.

packages/devextreme/js/common/ai-integration.d.ts
marker-dao: I have no objection. [9e280fc242ebd899caec42cb6de6285186d75abe]

packages/devextreme/js/common/ai-integration.d.ts
marker-dao: This function returns an abort function that cancels the current request. abort can be called from the UI as follows: onStopButtonClick: () => { abort() }.

packages/devextreme/js/common/ai-integration.d.ts
marker-dao: I added this type here purely because it is used to describe RequestCallbacks, and RequestCallbacks are used to describe AIIntegration.translate. In the final application, the user should only use AIIntegration, AIProvider (described with RequestParams (described with Prompt), ResponseParams. Please, if you think some types should be private - let's make them private.

packages/devextreme/js/common/ai-integration.d.ts
marker-dao: I have no objection. [d94d903a19ca36a6308df8cc83f72396558917dd]

r-farkhutdinov (approved):

r-farkhutdinov (approved):

IlyaKhD (approved):

PR #29348, opened 2025-03-19
List: selected item should receive focus when the list is grouped (T1278005)

EugeniyKiyashko:


anna-shakhova (dismissed):

anna-shakhova (approved):

packages/devextreme/js/__internal/ui/list/m_list.edit.ts
```diff
@@ -317,6 +318,19 @@ class ListEdit extends ListBase {
...
 
+  _getFlatIndex(): number | undefined {
+    const { selectedIndex } = this.option();
+
+    if (isNumeric(selectedIndex) || !selectedIndex) {
+      return selectedIndex;
+    }
+
+    const $item = this._editStrategy.getItemElement(selectedIndex);
+    const index = this.getFlatIndexByItemElement($item);
```
ksercs, line 329: It looks more logically to make this specific processing under `if` condition (we're trying to change only this, right?)  and return usual selectedIndex in all another scenarios

EugeniyKiyashko: It involves different editing strategies, and I would prefer not to add an additional if condition at this level. I hope that getItemElement has everything needed to process it the right way according to the strategy.

PR #29350, opened 2025-03-19
TabPanel: borders of nested TabPanels should be shown correctly in generic themes (T1280729)

EugeniyKiyashko:


anna-shakhova (dismissed):

packages/devextreme-scss/scss/widgets/generic/tabPanel/layout/tab/_position-top.scss
```diff
@@ -13,8 +13,8 @@
 // adduse
 
 .dx-tabpanel-tabs-position-top {
-  .dx-tabpanel-tab {
```
ksercs, line 16: Hm, so it was earlier broken even for not nested tabPanel?)

ksercs (approved):

packages/devextreme-scss/scss/widgets/generic/tabPanel/layout/tab/_position-top.scss
EugeniyKiyashko: For non-nested cases, it was almost okay

PR #29355, opened 2025-03-19
CardView: search text highlighting feature

wdevfx:


packages/devextreme/js/__internal/grids/new/grid_core/search/search_highlight_text_processor.ts
```diff
@@ -0,0 +1,33 @@
...
+    OptionsController,
+  ] as const;
+
+  public readonly highlightTextOptions$: SubsGets<HighlightTextOptions> = computed((
+    searchOptions: DefinedSearchOptions,
+  ) => ({
+    enabled: searchOptions.highlightSearchText,
+    caseSensitive: searchOptions.highlightCaseSensitive,
+    searchStr: searchOptions.text,
+  }), [this.options.oneWay('searchPanel') as SubsGets<DefinedSearchOptions>]);
```
pomahtri: Let's use `this.options.oneWay('searchPanel.enabled')`, `this.options.oneWay('searchPanel.caseSensitive')` separately
And let's get searchText from SearchController. It is bad to claim same option in different controllers. i think it will be even good idea to add runtime check for that

apps/react-storybook/stories/card_view/search/Search.stories.tsx
```diff
@@ -0,0 +1,65 @@
+import type {Meta, StoryObj} from "@storybook/react";
+import {wrapDxWithReact} from "../../utils";
+import dxCardView from "devextreme/ui/card_view";
+
+const CardView = wrapDxWithReact(dxCardView);
+
+const DATA = [{
```
pomahtri: we usually create all stories in same component to be sure that default config/data used is the same

packages/devextreme/js/__internal/grids/new/grid_core/search/search_highlight_text_processor.ts
wdevfx: Discussed and fixed
We decided to split logic into controller + view and delete processor

apps/react-storybook/stories/card_view/search/Search.stories.tsx
wdevfx: Fixed in terms of a separate PR:
https://github.com/DevExpress/DevExtreme/pull/29364

(Because it needed for headerFilter feature too)

pomahtri (approved):

PR #29356, opened 2025-03-20
DataGrid - `Cannot read properties of undefined (reading 'done')` error occurs when column fixing and state storing are used (T1283168)

tongsonbarbs:


e2e/testcafe-devextreme/tests/dataGrid/common/stateStoring/stateStoring.ts
```diff
@@ -110,3 +110,30 @@ test('The focused state of a row with the 0 key should be restored (T1252962)',
     },
   },
 }));
+
+test('DataGrid - Cannot read properties of undefined (reading \'done\') error occurs when column fixing and state storing are used (T1283168)', async (t) => {
+  const dataGrid = new DataGrid('#container');
```
Raushen: Unfortunately I was unable to make this test red without your changes on clear 25_1 branch. When the test runs on CI there is no any state in the store, so Grid won't restore anything.
Please check how to write such tests here: https://github.com/DevExpress/DevExtreme/blob/0ac0de0483c84eb9b02482c530504f5366d8d854/e2e/testcafe-devextreme/tests/treeList/focusedRow.ts#L68

Also there is no need to check errors in the console because if there are any, the test will fail itself. You can just leave a comment why there are not additional actions.

Raushen (changes requested):

packages/devextreme/js/__internal/grids/grid_core/sticky_columns/m_sticky_columns.ts
```diff
@@ -136,7 +136,6 @@ const baseStickyColumns = <T extends ModuleType<ColumnsView>>(Base: T) => class
       hasStickyColumns,
       this.addWidgetPrefix.bind(this),
     );
-
```
Raushen: Let's avoid unnecessary changes.

Raushen (approved):

PR #29362, opened 2025-03-20
HtmlEditor: click on variable toolbar format button before popup is hidden after variable select should not add a undefined variable (T1278577)

Zedwag:


r-farkhutdinov (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/toolbarIntegration.tests.js
```diff
@@ -1000,5 +1000,36 @@ export default function() {
...
+                    dataSource: ['test'],
+                    escapeChar: '%'
+                },
+            }).dxHtmlEditor('instance');
+
+            $('#htmlEditor')
+                .find(`.${TOOLBAR_FORMAT_WIDGET_CLASS}`)
+                .trigger('dxclick');
+
+            $('.dx-suggestion-list .dx-list-item').trigger('dxclick');
```
ksercs: [refactoring]
Let's use string literals? I see we already have `LIST_ITEM_CLASS` specified, would be cool if you add a constant for `dx-suggestion-list` and get rid of direct using of classes

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/toolbarIntegration.tests.js
```diff
@@ -1000,5 +1000,36 @@ export default function() {
...
+            $('#htmlEditor')
+                .find(`.${TOOLBAR_FORMAT_WIDGET_CLASS}`)
+                .trigger('dxclick');
+
+            this.clock.tick(10);
+
+            const value = editor.option('value');
+
+            assert.strictEqual(prepareEmbedValue(value), expectedValue);
+            done();
```
ksercs: [excess code?]
So, what's the point to use `async-done` if you just call it in the end of sync code? Kindly remove it or move to `onValueChanged` as it's done in neighbor tests

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/toolbarIntegration.tests.js
Zedwag: done

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/toolbarIntegration.tests.js
Zedwag: `onValueChanged` is not suitable here since we expect undefined only on second value changed. but fair point about async, removed it

ksercs (approved):

PR #29401, opened 2025-03-24
fix nested option changes in scheduler toolbar

Ambrozy:


e2e/testcafe-devextreme/tests/scheduler/common/header/toolbar_option_change.ts
```diff
@@ -0,0 +1,63 @@
+import { createScreenshotsComparer } from 'devextreme-screenshot-comparer';
```
wdevfx: Let's cover the following cases by tests too:
- Root option change
- Item[idx] whole option change

Now these cases aren't working:

```js
// Root option change
$('#container').dxScheduler('instance').option('toolbar', { items: [] });

$('#container').dxScheduler('instance').option('toolbar', { disabled: true});
```

About the second case, if I don't pass optionName in the ```toolbar``` option object, should this configuration be applied ```as is``` or it should be merged with default?

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -115,6 +116,31 @@ export class SchedulerHeader extends Widget<dxSchedulerOptions> {
     }
   }
 
+  onToolbarOptionChanged(args: ChangedOptionInfo) {
+    if (args.name === 'toolbar') {
```
wdevfx: Looks like the whole ```toolbar``` option update is missed
(See comment in e2e tests)

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -115,6 +116,31 @@ export class SchedulerHeader extends Widget<dxSchedulerOptions> {
     }
   }
 
+  onToolbarOptionChanged(args: ChangedOptionInfo) {
```
wdevfx: Middle comment (let's discuss if u disagree)

Let's move knowledge about options and ```ChangedOptionInfo``` to the root scheduler's ```optionChanged``` method (in ```m_scheduler.ts```)
As far as I can see we need here only ```args.fullName``` and ```args.value```
So, we can change method signature to

```ts
 // Maybe we can provide good type for value arg
 onToolbarOptionChanged(fullOptionName: string, value: ???): void {
   // method's code
 }
```

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -115,6 +116,31 @@ export class SchedulerHeader extends Widget<dxSchedulerOptions> {
...
+    if (args.name === 'toolbar') {
+      const parts = getPathParts(args.fullName);
+      const optionName = args.fullName.replace(/^toolbar\./, '');
+
+      switch (true) {
+        case parts[1] === 'items' && parts.length === 2:
```
wdevfx: Minor comment.

Let's replace this condition with:
```ts
  case optionName === items:
```

packages/devextreme/js/__internal/scheduler/header/m_header.ts
wdevfx: Minor comment.

Let's move the ```this._toolbar``` existance check on top of this method to remove the ```?```
Because if toopbar doesn't exist -> option change handling make no sense :)

E.g:
```ts
// Maybe we can provide good type for value arg
 onToolbarOptionChanged(fullOptionName: string, value: ???): void {
   if (!this._toolbar) {
     return;
   }
   // and next in the code use this._toolbar without ?
   this._toolbar.option('blablabla');
 }
```

e2e/testcafe-devextreme/tests/scheduler/common/header/toolbar_option_change.ts
```diff
@@ -0,0 +1,63 @@
...
+import Scheduler from 'devextreme-testcafe-models/scheduler';
+import { ClientFunction } from 'testcafe';
+import { createWidget } from '../../../../helpers/createWidget';
+import url from '../../../../helpers/getPageUrl';
+
+fixture.disablePageReloads`Scheduler header nested options change`
```
wdevfx: Minor comment.

Let's change the fixture name a little to make it more obvious:
```Scheduler: Toolbar options change```
(And more consistent with file name)

File: e2e/testcafe-devextreme/tests/scheduler/common/header/toolbar_option_change.ts [hunk 6]
```diff
@@ -0,0 +1,63 @@
...
+  ($('#container') as any).dxScheduler('instance').option('toolbar.items', buttons);
+});
+const changeButtonText = ClientFunction(() => {
+  ($('#container') as any).dxScheduler('instance').option('toolbar.items[0].options.text', 'Changed text');
+});
+const changeFirstItemLocation = ClientFunction(() => {
+  ($('#container') as any).dxScheduler('instance').option('toolbar.items[0].location', 'after');
+});
+
+test('Scheduler should change nested toolbar options', async (t) => {
```
wdevfx: Let's split this one big test case into a few small ones

E.g:
- Should change item location in runtime
- Should setup toolbar options in runtime
- Should turn on multiline in runtime

 and so on
 
 A few small tests are better than one big one because if sth is broken,
 we can see exactly which case works wrong

e2e/testcafe-devextreme/tests/scheduler/common/header/toolbar_option_change.ts
```diff
@@ -0,0 +1,63 @@
...
+  await changeButtonText();
+  await t
+    .expect(await takeScreenshot('scheduler-toolbar-item-option-changed.png', scheduler.toolbar.element))
+    .ok();
+
+  await t
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+}).before(async () => createWidget('dxScheduler', {
+  views: ['day', 'week', 'workWeek', 'month'],
```
wdevfx: Minor comment.

Because we don't see the ```viewSwitcher``` in etalons and don't use other views, let's reduce views count to make the test configuration more clean:
```ts
  views: ['week']
```

e2e/testcafe-devextreme/tests/scheduler/common/header/viewSwitcher.ts
```diff
@@ -31,3 +32,43 @@ test('it should correctly switch a differently typed views (T1080992)', async (t
...
+    .expect(selectBox.value)
+    .eql(defaultSelectBoxValue);
+
+  await t.click(scheduler.toolbar.viewSwitcher.getButton('Month').element)
+    .expect(scheduler.checkViewType('month'))
+    .ok();
+
+  await t.expect(selectBox.value).eql(defaultSelectBoxValue);
+}).before(async () => createWidget('dxScheduler', {
+  views: ['day', 'week', 'workWeek', 'month'],
```
wdevfx: Minor comment.

Because we don't see the ```viewSwitcher``` in etalons and don't use other views, let's reduce views count to make the test configuration more clean:
```ts
  views: ['week']
```

e2e/testcafe-devextreme/tests/scheduler/common/header/viewSwitcher.ts
```diff
@@ -31,3 +32,43 @@ test('it should correctly switch a differently typed views (T1080992)', async (t
...
+  currentView: 'week',
+  currentDate: new Date(2021, 3, 27),
+  height: 200,
+  useDropDownViewSwitcher: false,
```
wdevfx: Minor comment.

Let's remove this two options, because we don't need them in this test case

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -175,7 +201,7 @@ export class SchedulerHeader extends Widget<dxSchedulerOptions> {
       }
     }
 
-    return item;
+    return extend(true, {}, item);
```
wdevfx, line 204: Question.

Sorry, missed this point in a previous review:
Is it ok that we log the error if the item name isn't a default element type?

Because if a user will specify the name in the custom toolbar element -> we will log this error
I mean sth like this:
```ts
$('#scheduler').dxScheduler('instance').option('toolbar.items', [{
    name: 'mySuperButton', // <----
    location: 'before',
    locateInMenu: 'auto',
    widget: 'dxButton',
    options: { text: `Button` },
}]);
```

But maybe it's ok because the ```toolbar.items``` doesn't have the ```name``` option :)
[Link to the docs](https://js.devexpress.com/jQuery/Documentation/Guide/UI_Components/Toolbar/Getting_Started_with_Toolbar/)

e2e/testcafe-devextreme/tests/scheduler/common/header/toolbar_option_change.ts
Ambrozy: it will be merged with default

e2e/testcafe-devextreme/tests/scheduler/common/header/toolbar_option_change.ts
Ambrozy: Comparison of the screenshots executes on the end of the test, so if some of the screenshot fails test will continue

packages/devextreme/js/__internal/scheduler/header/m_header.ts
Ambrozy: Yes, it's okay. The same behavior in the DataGrid https://github.com/DevExpress/DevExtreme/blob/1945d65b25337687dd55fafe4b9f47d28716b0ee/packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts#L90

wdevfx (approved):

PR #29402, opened 2025-03-24
CardView: Search Panel - implement core

Raushen:


packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
```diff
@@ -188,13 +202,14 @@ export class DataController {
...
-        if (dataSource.requireTotalCount() !== true) {
+
+        if (!dataSource.requireTotalCount()) {
           dataSource.requireTotalCount(true);
           someParamChanged ||= true;
         }
         if (dataSource.filter() !== displayFilter) {
-          dataSource.filter(displayFilter);
+          const currentFilter = displayFilter ?? null;
+          dataSource.filter(currentFilter);
```
wdevfx: Minor comment.

```ts
dataSource.filter(displayFilter ?? null);
```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_controller.ts
```diff
@@ -1,21 +1,34 @@
...
 
   public readonly displayFilter = computed(
-    (filter, filterEnabled) => (filterEnabled ? filter : null),
-    [this.filter, this.filterEnabled],
+    (filterEnabled, filter, searchFilter) => {
+      if (!filterEnabled) {
+        return searchFilter;
+      }
+      return gridCoreUtils.combineFilters([filter, searchFilter]);
+    },
```
wdevfx, line 21: Minor comment.

```ts
(filterEnabled, filter, searchFilter) => !filterEnabled
  ? searchFilter
  : gridCoreUtils.combineFilters([filter, searchFilter])
```

packages/devextreme/js/__internal/grids/new/grid_core/search/controller.ts
```diff
@@ -24,9 +32,27 @@ export class SearchController {
 
   private readonly _searchTextOption = this.options.twoWay('searchPanel.text');
 
+  private readonly _searchVisibleColumnsOnly = this.options.oneWay('searchPanel.searchVisibleColumnsOnly');
+
   public readonly searchTextOption: SubsGets<string> = this._searchTextOption;
 
-  constructor(private readonly options: OptionsController) {}
+  public readonly searchFilter = computed(
```
wdevfx, line 33: Let's cover this ```computed``` (in fact the private method ```calculateSearchFilter```) by jest tests in ```controller.test.ts```

packages/devextreme/js/__internal/grids/new/grid_core/search/utils.ts
```diff
@@ -1,4 +1,20 @@
-import type { HighlightedTextItem, HighlightTextOptions } from './types';
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
wdevfx, line 1: Let's cover these utils func by jest tests

packages/devextreme/js/__internal/grids/new/grid_core/search/utils.ts
```diff
@@ -44,3 +60,92 @@ export const splitHighlightedText = (
...
 };
+
+export const allowSearch = (column: Column, searchVisibleColumnsOnly: boolean): boolean => {
+  const allowSearchByDataField = !!column.dataField && isString(column.dataField);
+  const allowSearchByVisibility = !searchVisibleColumnsOnly || column.visible;
+  const allowSearchByConfig = column.allowSearch ?? column.allowFiltering;
+  return allowSearchByDataField && allowSearchByVisibility && allowSearchByConfig;
+};
+
+export const parseValue = (column: Column, text: string): unknown => {
```
wdevfx: Let's reduce the complexity of this function:

Something like this
```ts
export const parseNumberValue = (
  column: Column,
  text: string
): unknown => {
  switch (true) {
    case isString(text) && column.format:
      return strictParseNumber(text.trim(), column.format);
    case isDefined(text) && isNumeric(text):
      return Number(text);
    default:
      return null;
  }
}

// other functions like parseBooleanValue / parseDateValue and so on

export const parseValue = (
  column: Column,
  text: string
): unknown => {
  switch (true) {
    case column.dataType === 'number':
      return parseNumberValue(column, text);
    case column.dataType === 'boolean':
      return parseBooleanValue(column, text);
    case gridCoreUtils.isDateType(column.dataType):
      return parseDateValue(column, text);
    default:
      return text;
  }
}

```

packages/devextreme/js/__internal/grids/new/grid_core/search/utils.ts
```diff
@@ -44,3 +60,92 @@ export const splitHighlightedText = (
...
+): unknown => {
+  // eslint-disable-next-line @typescript-eslint/init-declarations
+  let result;
+  if (column.calculateFilterExpression) {
+    result = column.calculateFilterExpression(filterValue, selectedFilterOperation, target);
+  }
+  if (isFunction(result)) {
+    result = [result, '=', true];
+  }
+  return result;
```
wdevfx: Now we can return ```undefined``` here if all conditions are false
Is it ok? :)

packages/devextreme/js/__internal/grids/new/grid_core/search/utils.ts
```diff
@@ -44,3 +60,92 @@ export const splitHighlightedText = (
...
+};
+
+const FILTERING_TIMEOUT = 700;
+
+const CLASS = {
+  searchPanel: 'search-panel',
+};
```
wdevfx: Let's move these const on top of the file

packages/devextreme/js/__internal/grids/new/grid_core/search/utils.ts
```diff
@@ -44,3 +60,92 @@ export const splitHighlightedText = (
...
+
+// eslint-disable-next-line @typescript-eslint/init-declarations
+let timer;
```
wdevfx, line 170: We should clear all timers on the component ```dispose```
(We have strict checks in QUnit about it for each component)

So, let's also export a function that will clear the scheduled timer
As an example, this is how new macro tasks are executed in the Scheduler:
https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/scheduler/utils/macro_task_array/dispatcher.ts#L2

Maybe it's a good idea to add some util macro task dispatcher for CardView too or sth like that

packages/devextreme/js/__internal/grids/new/grid_core/search/utils.ts
```diff
@@ -44,3 +60,92 @@ export const splitHighlightedText = (
...
+
+  widget: 'dxTextBox',
+  options: {
+    onInput: (e: NativeEventInfo<TextBoxInstance, UIEvent>): void => {
+      clearTimeout(timer);
+      const component = e.component as any;
+      const newValue = component._input().val();
+      timer = setTimeout(() => {
+        props.onValueChanged?.(newValue);
+      }, FILTERING_TIMEOUT);
```
wdevfx, line 186: Should we have configurable filtering timeout?
And why do we use ```700```ms here?

packages/devextreme/js/__internal/grids/new/grid_core/search/controller.ts
```diff
@@ -24,9 +32,27 @@ export class SearchController {
 
   private readonly _searchTextOption = this.options.twoWay('searchPanel.text');
 
+  private readonly _searchVisibleColumnsOnly = this.options.oneWay('searchPanel.searchVisibleColumnsOnly');
```
wdevfx: Let's remove these ```_``` (underscore) from the private field names

As far as I can see, we don't have any special convention for naming class properties 

But there are some arguments against underscore:
1) In most classes we don't use underscore in private fields naming
2) In this class we have two private fields without it (```options``` and columnsController```)

So, let's make it consistent :)

packages/devextreme/js/__internal/grids/new/grid_core/search/options.ts
```diff
@@ -6,15 +8,27 @@ export interface Options {
     highlightCaseSensitive?: boolean;
     // Specifies whether found substrings should be highlighted.
     highlightSearchText?: boolean;
+    // Specifies a placeholder for the search panel.
+    placeholder?: string;
```
wdevfx, line 12: Should we add some tests to the ```options.integration.test.ts``` or ```testcafe``` e2e tests will be enough?

@pomahtri cc

packages/devextreme/js/__internal/grids/new/grid_core/test_utils.ts
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable spellcheck/spell-checker */
```
wdevfx: Can we separate these jest test changes to a separate PR?

Because now, in one PR, we have two unrelated changes:
- our technical test setup refactoring that touch a lot of tests
- implementation of the search feature

I'm not insist of split this changes, but it's a good practise to do a few small PR instead of big one if changes unrelated between each other :)
Also, it may help us to provide better review in future and produce less conflicts

packages/devextreme/js/ui/shared/filtering.js
```diff
@@ -31,11 +31,18 @@ const getGroupInterval = function(column) {
     }
 };
 
+const getNormalizedCalculateDisplayValue = function(column) {
+    if(column.calculateDisplayValue?.context) {
+        return column.calculateDisplayValue;
+    }
+    return null;
```
wdevfx: Minor comment.

```ts
  return column.calculateDisplayValue?.context ? column.calculateDisplayValue : null;
```

packages/devextreme/js/ui/shared/filtering.js
```diff
@@ -31,11 +31,18 @@ const getGroupInterval = function(column) {
...
+};
+
 export default (function() {
     const getFilterSelector = function(column, target) {
         let selector = column.dataField || column.selector;
         if(target === 'search') {
-            selector = column.displayField || column.calculateDisplayValue || selector;
+            selector = column.displayField || getNormalizedCalculateDisplayValue(column) || selector;
```
wdevfx, line 42: Are we 100% sure that these changes don't affect other widgets and don't fail any tests?

packages/testcafe-models/dataGrid/index.ts
```diff
@@ -16,7 +16,6 @@ import { Overlay } from './overlay';
 import MasterRow from './masterRow';
 import AdaptiveDetailRow from './adaptiveDetailRow';
 import ColumnChooser from './columnChooser';
-import TextBox from '../textBox';
```
wdevfx: I suggest to leave the existing ```DataGrid``` ```POM``` untouched
And after the ```CardView``` will be merged, we delete the redundant overrides

Why:
- Less count of conflicts
- Now we removed the ```getColumnsSeparator``` and looks like the related tests will be red

Raushen: There are no usings of getColumnsSeparator

packages/devextreme/js/__internal/grids/new/grid_core/test_utils.ts
Raushen: I would agree with you in other cases... But not for this one) Without these change there are A LOT of conflicts in jest tests. So if I'll extract these changes in a separate PR I will have to support the old architecture of tests with a lot of conflicts. And I'll have to do the same after each merge to draft branch. 
So I just want to save our time as an emergency measure.

packages/testcafe-models/dataGrid/index.ts
Raushen: However the conflicts is the big problem I've reverted changes and have created a separate task to update DataGrid model

packages/devextreme/js/ui/shared/filtering.js
Raushen: This change should not affect existing implementation. But let's see on the Grid tests state to ensure.

packages/devextreme/js/__internal/grids/new/grid_core/search/utils.ts
Raushen: There is no timeout option or TextBox options in [Search panel settings](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/searchPanel/). The implementation is [the same as in DataGrid](https://github.com/DevExpress/DevExtreme/blob/016df420a828326c9c80d9ac7b073fd8fd169918/packages/devextreme/js/__internal/grids/grid_core/search/m_search.ts#L20).

packages/devextreme/js/__internal/grids/new/grid_core/search/options.ts
Raushen: I think it is enough to have e2e tests for public options. But I decided to add integration tests in the context of render testing.

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_controller.ts
Raushen: It seems I thought about the next step when I wrote this code) Please allow me to make a refactoring (if needed) when another parameter of headerFilter will be added.

packages/devextreme/js/__internal/grids/new/grid_core/search/controller.ts
Raushen: Will research in the context of testing utils

vladaskorohodova (approved):

packages/devextreme/js/__internal/grids/new/grid_core/search/utils.ts
wdevfx: Still actual :)

wdevfx: Sorry, missed this new [onDisposing](https://github.com/DevExpress/DevExtreme/pull/29402/files#diff-c9fcf48982188e5af75a5cd648ed2ea3d1757fda11e086373fca7ece0008e1a1R199) function

Raushen:
Reviewed, all seems good.

PR #29415, opened 2025-03-25
HtmlEditor: Add AI toolbar option types

r-farkhutdinov:


packages/devextreme/js/ui/html_editor.d.ts
```diff
@@ -46,7 +46,109 @@ export type HtmlEditorImageUploadTab = 'url' | 'file';
...
+    | 'french'
+    | 'german'
+    | 'japanese'
+    | 'spanish';
+
+/**
+ * @namespace DevExpress.ui.dxHtmlEditor
+ */
+export interface HtmlEditorAICommandBase {
+    name: HtmlEditorAICommandName | 'custom';
```
marker-dao: I didn't understand the reason for separating “custom” from HtmlEditorAICommandName.

packages/devextreme/js/ui/html_editor.d.ts
```diff
@@ -46,7 +46,109 @@ export type HtmlEditorImageUploadTab = 'url' | 'file';
...
+export interface HtmlEditorAISimpleCommand extends HtmlEditorAICommandBase {
+    name: HtmlEditorAISimpleCommandName;
+}
+
+/**
+ * @namespace DevExpress.ui.dxHtmlEditor
+ */
+export interface HtmlEditorAIChangeStyleCommand extends HtmlEditorAICommandBase {
+    name: 'changeStyle';
+    options?: HtmlEditorAIChangeStyleOption[];
```
marker-dao: Could we remove code duplication, for example by using the Basic interface using generics? Something like this? 
```suggestion
export interface HtmlEditorAICommandBase<
    T extends HtmlEditorAICommandName,
    Options = undefined
> {
    name: T;
    text?: string;
    options?: Options;
}
```

packages/devextreme/js/ui/html_editor.d.ts
```diff
@@ -46,7 +46,109 @@ export type HtmlEditorImageUploadTab = 'url' | 'file';
...
+ */
+export interface HtmlEditorAICommandBase {
+    name: HtmlEditorAICommandName | 'custom';
+    text?: string;
+}
+
+/**
+ * @namespace DevExpress.ui.dxHtmlEditor
+ */
+export interface HtmlEditorAISimpleCommand extends HtmlEditorAICommandBase {
```
marker-dao: I would rename HtmlEditorAISimpleCommand to HtmlEditorAIBasicCommand, but Vlada can help us with the naming issue later.

packages/devextreme/js/ui/html_editor.d.ts
r-farkhutdinov: If we add `custom` to `HtmlEditorAICommandName` this will mean that the user can specify custom command just as string, but it requires parameters. 
To make it look better, we can separate a new type where we will have all possible names + all possible names except "custom"

packages/devextreme/js/ui/html_editor.d.ts
r-farkhutdinov: Updated in: https://github.com/DevExpress/DevExtreme/pull/29415/commits/054b856a8651904a8dce3fa6781c49cdb63625cd

packages/devextreme/js/ui/html_editor.d.ts
```diff
@@ -46,7 +46,138 @@ export type HtmlEditorImageUploadTab = 'url' | 'file';
...
+export type HtmlEditorPredefinedToolbarItem = 'background' | 'bold' | 'color' | 'font' | 'italic' | 'link' | 'image' | 'size' | 'strike' | 'subscript' | 'superscript' | 'underline' | 'blockquote' | 'header' | 'increaseIndent' | 'decreaseIndent' | 'orderedList' | 'bulletList' | 'alignLeft' | 'alignCenter' | 'alignRight' | 'alignJustify' | 'codeBlock' | 'variable' | 'separator' | 'undo' | 'redo' | 'clear' | 'cellProperties' | 'tableProperties' | 'insertTable' | 'insertHeaderRow' | 'insertRowAbove' | 'insertRowBelow' | 'insertColumnLeft' | 'insertColumnRight' | 'deleteColumn' | 'deleteRow' | 'deleteTable' | 'ai';
+/** @public */
```
marker-dao, line 51: Please add indents between types

packages/devextreme/js/ui/html_editor.d.ts
```diff
@@ -46,7 +46,138 @@ export type HtmlEditorImageUploadTab = 'url' | 'file';
...
+}
+
+/**
+ * @docid
+ * @public
+ * @type object
+ * @inherits HtmlEditorAICommandBase
+ * @namespace DevExpress.ui.dxHtmlEditor
+ */
+export type HtmlEditorAIBasicCommand = HtmlEditorAICommandBase<HtmlEditorAIBasicCommandName>;
```
marker-dao: I still don't understand why we need HtmlEditorAIBasicCommandName) Can you explain where it will be needed in the code? If it will not be used in the code, I would prefer to use the 

> export type HtmlEditorAIBasicCommand =
>   | HtmlEditorAICommandBase<'summarize', never>
>   | HtmlEditorAICommandBase<'proofread', never>
>   | HtmlEditorAICommandBase<'expand', never>
>   | HtmlEditorAICommandBase<'shorten', never>
>   | HtmlEditorAICommandBase<'askAI', never>;

“never” to prevent the user from passing options to these commands.

marker-dao: Then you can get rid of HtmlEditorAIBasicCommandName.

packages/devextreme/js/ui/html_editor.d.ts
marker-dao: Why? You have defined **HtmlEditorAICustomCommand** and **HtmlEditorAICommand**. If the user wants to pass 'custom', **prompt** will be required and **options** will not. 

One more question, have we agreed with PMs what the default value of the **text** property for **custom**-command will contain? What if the user doesn't specify the **text** prop? 

Also, I think that in prompt: **(param: string) => string;** the **param** argument should be optional, because if the user didn't pass us the **options**-list, we have nothing to return to them.

r-farkhutdinov: 1. Options are not required for the custom command, why do you think so? What if there's just a simple prompt and no options?
2. Great question, made the text property required for the custom command
3. Agree

packages/devextreme/js/ui/html_editor.d.ts
r-farkhutdinov: I don't think it will be used in our code, agree, we can simplify

packages/devextreme/js/ui/html_editor.d.ts
r-farkhutdinov: Fixed in https://github.com/DevExpress/DevExtreme/pull/29415/commits/79605fc489c05a95e7d0266b1349fe292e9e972a

packages/devextreme/js/ui/html_editor.d.ts
r-farkhutdinov: Fixed in https://github.com/DevExpress/DevExtreme/pull/29415/commits/79605fc489c05a95e7d0266b1349fe292e9e972a

packages/devextreme/js/ui/html_editor.d.ts
r-farkhutdinov: Fixed in https://github.com/DevExpress/DevExtreme/pull/29415/commits/79605fc489c05a95e7d0266b1349fe292e9e972a

marker-dao: > Options are not required for the custom command, why do you think so?

I didn't write that, I wrote that they are not mandatory, but prompt is mandatory.

> Great question, made the text property required for the custom command

I don't think it should be made mandatory. This is a question that should be asked of PMs. The answer could be anything.

marker-dao: Okay, now I get your idea of “custom”.

packages/devextreme/js/ui/html_editor.d.ts
```diff
@@ -46,7 +46,150 @@ export type HtmlEditorImageUploadTab = 'url' | 'file';
...
+    | HtmlEditorAITranslateCommand
+    | HtmlEditorAICustomCommand;
+
+/**
+ * @docid
+ * @public
+ * @type object
+ * @namespace DevExpress.ui.dxHtmlEditor
+ */
+export interface HtmlEditorAIToolbarItem {
```
ksercs: Kindly explore which options from `dxToolbarItem` are useful for this item. I believe there should be extend + omit

packages/devextreme/js/ui/html_editor.d.ts
```diff
@@ -46,7 +46,150 @@ export type HtmlEditorImageUploadTab = 'url' | 'file';
...
+    | HtmlEditorAICommandBase<'askAI', never>;
+
+/**
+ * @docid
+ * @public
+ * @type object
+ * @inherits HtmlEditorAICommandBase
+ * @namespace DevExpress.ui.dxHtmlEditor
+ */
+export type HtmlEditorAIChangeStyleCommand = HtmlEditorAICommandBase<'changeStyle', HtmlEditorAIChangeStyleOption[]>;
```
ksercs: Should be `| string` here?
I don't see where an ability for specifying custom param is supported

packages/devextreme/js/ui/html_editor.d.ts
r-farkhutdinov: Proposed a set of properties to Omit: https://github.com/DevExpress/DevExtreme/pull/29415/commits/0b4dc94fe347776e204ee553181ecf72ba137cb2

packages/devextreme/js/ui/html_editor.d.ts
r-farkhutdinov: Added to all command options: https://github.com/DevExpress/DevExtreme/pull/29415/commits/0b4dc94fe347776e204ee553181ecf72ba137cb2

ksercs: I'm not sure it's optimal but will see what API reviewers say

ksercs (dismissed):

packages/devextreme/js/ui/html_editor.d.ts
```diff
@@ -46,7 +46,162 @@ export type HtmlEditorImageUploadTab = 'url' | 'file';
...
+    | 'german'
+    | 'japanese'
+    | 'spanish'
+    | string;
+
+/**
+ * @docid
+ * @public
+ */
+export type HtmlEditorAICommandNameAll = HtmlEditorAICommandName | 'custom';
```
vladaskorohodova: I don't really like the "all" suffix. Maybe we can change it to HtmlEditorAICommandNameExtended or smth like this?

packages/devextreme/js/ui/html_editor.d.ts
```diff
@@ -46,7 +46,162 @@ export type HtmlEditorImageUploadTab = 'url' | 'file';
...
+    | 'business'
+    | 'creative'
+    | 'journalistic'
+    | 'academic'
+    | 'persuasive'
+    | 'narrative'
+    | 'expository'
+    | 'descriptive'
+    | 'conversational'
+    | string;
```
IlyaKhD:
```suggestion
    | 'conversational';
```

I think if we remove `string` out of the these types it will make it more convenient to reusethem  in end projects.

Assuming that string will go here:

```ts
export type HtmlEditorAIChangeStyleCommand = HtmlEditorAICommandBase<'changeStyle', (HtmlEditorAIChangeStyleOption | string)[]>;
```

IlyaKhD: Maybe it's also better for reading when surfing through d.ts (I can't see literal values until I go directly to them):


image

packages/devextreme/js/ui/html_editor.d.ts
r-farkhutdinov: Renamed to HtmlEditorAICommandNameExtended

IlyaKhD (approved):

marker-dao (approved):

PR #29425, opened 2025-03-25
TagBox: previously rendered tags should not rerender on value change (T1246066)

Zedwag:


r-farkhutdinov (dismissed):

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1249,12 +1248,6 @@ class TagBox<
     const itemModel = this._getItemModel(item, displayValue);
 
     if ($tag) {
-      if (isDefined(displayValue)) {
-        $tag.empty();
-        this._applyTagTemplate(itemModel, $tag);
-      }
-
-      $tag.removeClass(TAGBOX_CUSTOM_TAG_CLASS);
       this._updateElementAria($tag.attr('id'));
```
ksercs, line 1258: Is it still necessary? If tag is already rendered -> labelledby already contains its ID. No?

Zedwag: i see red tests (related to kbn) if i remove this. i don't see a point to research it deeper, looks like it still necessary

ksercs (approved):

PR #29426, opened 2025-03-26
CardView - implement rootElementUpdater

pomahtri:

### DIFF
--- a/packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
@@ -88,7 +88,7 @@ export class RootElementUpdater extends Component<Properties> {
     const element = this.props.rootElementRef.current;
 
     if (!element) {
-      throw new Error('root element is not providen');
+      throw new Error('root element is not provided');
     }
 
     this.updateClasses(element);
--- a/packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
@@ -1,0 +1,101 @@
+/* eslint-disable spellcheck/spell-checker */
+import type { RefObject } from 'inferno';
+import { Component } from 'inferno';
+
+export function normalizeEventName(name: string): string {
+  return name.substring(2).toLowerCase();
+}
+export type Properties = JSX.IntrinsicElements['div'] & {
+  rootElementRef?: RefObject<HTMLDivElement>;
+};
+
+export class RootElementUpdater extends Component<Properties> {
+  private previousClasses: string[] = [];
+
+  private previousAttributes: Record<string, unknown> = {};
+
+  public render(): JSX.Element {
+    if (!this.props.rootElementRef?.current) {
+      return <div {...this.props}>
+        {this.props.children}
+      </div>;
+    }
+
+    return <>
+      {this.props.children}
+    </>;
+  }
+
+  private updateClasses(element: HTMLDivElement): void {
+    const currentClassName = this.props.className;
+
+    const currentClasses = currentClassName?.split(' ') ?? [];
+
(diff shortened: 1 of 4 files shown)


packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
```diff
@@ -0,0 +1,101 @@
+/* eslint-disable spellcheck/spell-checker */
+import type { RefObject } from 'inferno';
+import { Component } from 'inferno';
+
+export function normalizeEventName(name: string): string {
+  return name.substring(2).toLowerCase();
+}
+export type Properties = JSX.IntrinsicElements['div'] & {
+  rootElementRef?: RefObject<HTMLDivElement>;
```
wdevfx: We don't use this component without this ref, so let's make it required in props

packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+export class RootElementUpdater extends Component<Properties> {
+  private previousClasses: string[] = [];
+
+  private previousAttributes: Record<string, unknown> = {};
+
+  public render(): JSX.Element {
```
wdevfx, line 17: Question.

As far as I understand, we render the fragment until the ```ref``` to the root element will be initialized
Is it ok, and why do we need this behavior?

Maybe we should render just ```<div>``` without passing props if the root element isn't initialized yet.

packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
```diff
@@ -0,0 +1,101 @@
...
+      ref,
+      className,
+      children,
+      ...currentAttributes
+    } = this.props;
+
+    const currentAttributeKeys = Object.keys(currentAttributes);
+    const previousAttributeKeys = Object.keys(this.previousAttributes);
+
+    currentAttributeKeys.forEach((attrName) => {
```
wdevfx, line 54: Looks like we missed here logic for ```addedAttributes``` like ```addedClasses``` in ```updateClasses```

Because if we pass twice ```onMyEventAttr``` we will create two event listeners

packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
pomahtri: No, ref in cardview's case is always initialized
ref is optional for future purposes if we'll want to use inferno component deattached from whole cardView

packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
pomahtri: See comment below

packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
pomahtri: fixed
`addedAttributes` are intendedly missed. In classes we don't need to change class that was present previously and still remains, in attributes we need to update it anyways because value might be changed, so here we go through all present attributes & delete removed ones

packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/root_element_updater.tsx
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    this.previousAttributes = currentAttributes;
+  }
+
+  private updateClassesAndAttributes(): void {
+    const element = this.props.rootElementRef.current;
+
+    if (!element) {
+      throw new Error('root element is not providen');
```
Raushen: A typo, 'provided'

Raushen (approved):

wdevfx (approved):

PR #29430, opened 2025-03-26
CardView - rewrite to signals

pomahtri:
- added `@preact/signals-core` package
- patch for `jest.config.json`. Jest for some reason mistakenly tries to load `.module.js` file instead of `.js`, but it doesn't support es modules
- patched qunit systemjs config
- removed `@ts/core/reactive`
- rewrote cardview


packages/devextreme/js/__internal/grids/new/grid_core/options_controller/options_controller_base.ts
```diff
@@ -20,25 +18,43 @@ import type {
 } from './types';
 import { getTreeNodeByPath, mergeOptionTrees } from './utils/index';
 
+function getOr<T>(cache: Record<string, T>, key: string, orElse: () => T): T {
```
wdevfx, line 21: [Let's do it as a separate task]

Let's move this cache logic to a separate entity with cache state
And use more obvious naming, clear cache on destroy and etc

Sth like (not necessarily class, closure will be fine too):
```ts
export class OptionsCache {
  private readonly cache = ...;
  
  public getOptionValue<T>(key: string, orElse: actionFn: () => T): T {
  // some code
  }
  
  public clearCache(): void {
  // some code
  }
}
```

packages/devextreme/js/__internal/grids/new/grid_core/options_controller/options_controller_base.ts
```diff
@@ -57,47 +73,72 @@ export class OptionsController<
...
-          callback?.(value);
-        }
-      },
-      // @ts-expect-error
-      unreactive_get: obs.unreactive_get.bind(obs),
-    };
+  ): Signal<PropertyWithDefaults<TProps, TDefaultProps, TProp>> {
+    return getOr(this.cache.twoWay, name, () => {
+      // eslint-disable-next-line @typescript-eslint/no-this-alias
+      const that = this;
```
wdevfx, line 104: [Let's do it as a separate task]

Let's get rid of ```that```
We can use the arrow function in ```set value(value: any)``` instead

packages/devextreme/js/__internal/grids/new/grid_core/options_controller/options_controller_base.ts
```diff
@@ -57,47 +73,72 @@ export class OptionsController<
...
+      const obs = signal(this.component.option(name));
```
wdevfx, line 105: [Let's do it as a separate task]

Also, let's get rid of this signal
It looks like a middle state, which is always updated from the ```oneWay``` binding

packages/devextreme/js/__internal/grids/new/grid_core/options_controller/options_controller_base.ts
```diff
@@ -57,47 +73,72 @@ export class OptionsController<
...
+            that.component.option(name, value);
+            callback?.(value);
+          }
+        },
+        peek(): any {
+          return obs.peek();
+        },
+        subscribe(...params: any): any {
+        // @ts-expect-error
+          return obs.subscribe(...params);
```
wdevfx, line 129: Question.

Why TS error here?

wdevfx (approved):

packages/devextreme/js/__internal/grids/new/grid_core/options_controller/options_controller_base.ts
pomahtri: Yep, let's do it when we'll be checking all subscriptions

packages/devextreme/js/__internal/grids/new/grid_core/options_controller/options_controller_base.ts
pomahtri: We cannot, setters cannot be arrow function

packages/devextreme/js/__internal/grids/new/grid_core/options_controller/options_controller_base.ts
pomahtri: Actually, it's a needed one
Let's have a call to discuss

packages/devextreme/js/__internal/grids/new/grid_core/options_controller/options_controller_base.ts
pomahtri: Cause `params` are not of tuple type, and only tuples can be spreaded into function which doesn't have rest params
`any[]` is not tuple too, only finite arrays are tuple

packages/devextreme/js/__internal/grids/new/grid_core/options_controller/options_controller_base.ts
pomahtri: NVM, i already forgot the point
Will check in separate PR and either fix or tell exact reason

wdevfx: Yup, it's ok
I don't think that with middle state break something, looks like we just don't need it
So, it can be easily postponed for the future changes

PR #29454, opened 2025-03-31
CardView - fix ssr

pomahtri:


packages/devextreme/js/__internal/grids/new/grid_core/core/view.tsx
```diff
@@ -3,24 +3,34 @@
...
 import { toSubscribable } from '@ts/core/reactive/index';
-import { Component, type ComponentType, render } from 'inferno';
+import { Component, type ComponentType } from 'inferno';
 
 export abstract class View<T extends {}> {
   private inferno: undefined | ComponentType;
 
+  private props?: T;
+
+  private readonly firstRender = true;
```
wdevfx: Where do we change the ```firstRender```?
Now it looks like that it always first render :)

packages/devextreme/js/__internal/grids/new/grid_core/core/view.tsx
```diff
@@ -3,24 +3,34 @@
...
 
-  protected abstract getProps(): Subscribable<T>;
+  protected abstract getProps(): SubsGets<T>;
 
   public render(root: HTMLDivElement): Subscription {
     this.root = root;
     const ViewComponent = this.component;
     return toSubscribable(this.getProps()).subscribe((props: T) => {
-      render(<ViewComponent {...props}/>, root);
+      this.props = props;
```
wdevfx, line 28: Why do we need to reassign props to the class instance?

pomahtri: For debug purposes only

packages/devextreme/js/__internal/grids/new/grid_core/core/view.tsx
pomahtri: Bad cherry-pick I guess

VasilyStrelyaev (approved):

PR #29455, opened 2025-03-31
CardView - create .d.ts

pomahtri:


packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -0,0 +1,1880 @@
...
+/** @public */
+export type CardHeaderPredefinedToolbarItem = 'selectionCheckBox' | 'updateButton' | 'deleteButton';
+
+/**
+ * @docid
+ * @inherits dxToolbarItem
+ * @public
+ * @namespace DevExpress.ui.dxCardView
+ */
+export type CardHeaderToolbarItem = dxToolbarItem & {
```
IlyaKhD:
```suggestion
export type CardHeaderItem = dxToolbarItem & {
```
Since it is only used in CardHeader, the `Toolbar` part is unnecessary

packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -0,0 +1,1880 @@
...
+    aspectRatio?: string;
+    /**
+     * @docid
+     * @public
+     */
+    template?: template | ((data: CardTemplateData, container: DxElement) => string | UserDefinedElement);
+};
+
+/** @public */
+export type CardHeaderPredefinedToolbarItem = 'selectionCheckBox' | 'updateButton' | 'deleteButton';
```
IlyaKhD:
```suggestion
export type CardHeaderPredefinedItem = 'selectionCheckBox' | 'updateButton' | 'deleteButton';
```
Since it is only used in CardHeader, the `Toolbar` part is unnecessary

packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -0,0 +1,1880 @@
...
+    /** @docid */
+    readonly field: FieldInfoType;
+    /** @docid */
+    readonly fieldCaptionElement: DxElement;
+};
+
+/**
+ * @docid
+ */
+type WithFieldValueInfo = {
```
IlyaKhD, line 556: **Question**
Just curious, why `WithFieldValueInfo`, not `FieldValueInfo`?

packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -0,0 +1,1880 @@
...
+// #endregion
+
+// #region ColumnsController
+
+/**
+ * @docid
+ * @public
+ * @namespace DevExpress.ui.dxCardView
+ */
+export type FieldInfoType = { // TODO: rename to FieldInfo
```
IlyaKhD:
```suggestion
export type FieldInfo = {
```

I see the TODO, just to have a conversation entry for ir

packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -0,0 +1,1880 @@
...
+    /**
+     * @public
+     * @docid
+     * @type object
+     */
+    value: unknown;
```
IlyaKhD: Not 100% sure about that, but it seems that we shouldn't use `unknown` for values that can be passed to consuming code. It's generallly a good practise, but all the other components currently use `any`.

```suggestion
    value: any;
```

packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -0,0 +1,1880 @@
...
+ * @public
+ * @namespace DevExpress.ui.dxCardView
+ */
+export type HeaderPanel<TCardData = unknown, TKey = unknown> = {
+    /**
+     * @docid
+     * @public
+     * @type object
+     */
+    dragging?: Pick<dxSortableOptions, 'dropFeedbackMode' | 'scrollSpeed' | 'scrollSensitivity' | 'onDragChange' | 'onDragEnd' | 'onDragMove' | 'onDragStart' | 'onRemove' | 'onReorder'>;
```
IlyaKhD: **Suggestion**
Maybe it would be useful to extract a public type for dragging:`HeaderPanelDragging`

packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -0,0 +1,1880 @@
...
+     * @docid
+     * @public
+     * @type object
+     */
+    filterBuilderPopup?: PopupProperties;
+    /**
+     * @docid
+     * @public
+     */
+    filterBuilder?: dxFilterBuilderOptions;
```
IlyaKhD, line 1469: You can create a local public alias, so that it get into `card_view_types.d.ts` and aggregated exports in framework packages:

```ts
/**
 * @public
 */
type FilterBuilder = dxFilterBuilderOptions;
export {
  FilterBuilder
}
```

```suggestion
    filterBuilder?: FilterBuilder;
```

packages/devextreme/js/ui/card_view.d.ts
wdevfx: Fixed

But one thought worried me a bit
If we would like to have some items in the ```Card``` in future -> the ```CardHeaderItem``` name will be already reserved

packages/devextreme/js/ui/card_view.d.ts
wdevfx: Fixed

Same thoughts like about the ```CardHeaderToolbarItem``` fix
But the chance here is extremely low,I guess

packages/devextreme/js/ui/card_view.d.ts
wdevfx: Because this type is used to add sth to existing types, I guess :)

E.g: 
```ts
export type FieldValuePreparedEvent = EventInfo<dxCardView> & WithFieldValueInfo;
```
An event with field value info

packages/devextreme/js/ui/card_view.d.ts
wdevfx: Fixed

But @pomahtri warned that the ```FieldInfo``` causes some error (I don't have more details)
So, if we will face complex issues with this name -> let's rollback my fix to the ```FieldInfoType``` with ```TODO```

packages/devextreme/js/ui/card_view.d.ts
wdevfx: Agree with u here
Let's not risk and use ```any``` for now

Btw, @IlyaKhD, what do u think about ```unknown``` in generics?
I guess it should be alright, but at night it started worried me a bit because ```unknown``` !== ```any```

packages/devextreme/js/ui/card_view.d.ts
wdevfx: Agree -> fixed

@IlyaKhD, could u please re-check that I did it right (I may have missed some jsDoc comment or sth)?
But as far as I can see, everything should be alright :)

packages/devextreme/js/ui/card_view.d.ts
wdevfx: This change leads to the following error:

```ts
System.Exception: "dxCardView.filterBuilder" has unknown type. Try adding @docid tag to the type.
         at DevExtreme.NgSmdGenerator.StrongType.Builder.Build(String optionName, String optionFullName, DataType[] types, Boolean hasAcceptableValues, Boolean hasNestedOptions)
         at DevExtreme.NgSmdGenerator.Program.BuildStrongType(Builder builder, HierarchicalMember member, String optionFullName)
         at DevExtreme.NgSmdGenerator.Program.<>c__DisplayClass2_0.<ToSerializable>b__5(HierarchicalMember m)
         at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
         at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector)
         at DevExtreme.NgSmdGenerator.Program.ToSerializable(HierarchicalMember member, Builder typesBuilder, String name)
         at DevExtreme.NgSmdGenerator.Program.<>c__DisplayClass1_0.<MainCore>b__3(HierarchicalMember m)
         at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](TSource[] source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
         at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
         at DevExtreme.NgSmdGenerator.Program.MainCore(ProgramConfig programConfig)
         at DevExtreme.NgSmdGenerator.Program.Main(String[] args)
```

As logs adviced -> I added the ```docid```:
```ts
/**
 * @docid
 * @public
 */
export type FilterBuilder = dxFilterBuilderOptions;
```

But I'm not sure that it's ok (we already have ```docid``` on the ```dxFilterBuilderOptions```)
And get the following error on the last Angular regenerate step:
```ts
 [04:00:49] TypeError: Cannot set properties of undefined (setting 'baseClass')
          at ComponentMetadataGenerator.generateComplexOptionByType (/Users/wdevfx/work/devextreme_prod/node_modules/.pnpm/devextreme-internal-tools@18.0.0-beta.1/node_modules/devextreme-internal-tools/ts/angular-components-generator/metadata-generator.js:444:43)
          at ComponentMetadataGenerator.generate (/Users/wdevfx/work/devextreme_prod/node_modules/.pnpm/devextreme-internal-tools@18.0.0-beta.1/node_modules/devextreme-internal-tools/ts/angular-components-generator/metadata-generator.js:283:47)
          at /Users/wdevfx/work/devextreme_prod/packages/devextreme-angular/gulpfile.js:38:13
          at bound (node:domain:433:15)
          at runBound (node:domain:444:12)
          at asyncRunner (/Users/wdevfx/work/devextreme_prod/node_modules/.pnpm/async-done@1.3.2/node_modules/async-done/index.js:55:18)
          at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
```

@IlyaKhD, may we temporarily leave it as is with direct ```dxFilterBuilderOptions``` type?

packages/devextreme/js/ui/card_view.d.ts
alexslavr: Hi, @IlyaKhD, @wdevfx. `FieldInfo` causes a conflict with `FieldInfo` from `FilterBuilder`, which is also used in CardView and imported in generated components directly from 'filter_builder' module. This change will lead to re-generation in CardView and **other grids** with `import { FieldInfo } from 'filter_builder' ` replaced to `import { FieldInfo } from 'card_view'` which is totally wrong. So we can revert renaming or use WA with prefixed internal type.

mpreyskurantov (approved):

packages/devextreme/js/ui/card_view.d.ts
IlyaKhD: It sounds like a name of a HOC

packages/devextreme/js/ui/card_view.d.ts
IlyaKhD: Well, I don't have a general idea for all generics, better investigate some samples. We have reasons to avoid giving unknown outside, and can leverage from marking as unknown for data going inside. Is there a gurantee that generics never given outside - hm... not sure about that, it will be hard to lint, I think. Not sure about possible covariance / contravariance issues, 2-way bindings in Angualr, etc

PR #29479, opened 2025-04-01
Map - add test on map ready callback and route adding fail

Zedwag:


packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/azureTests.js
```diff
@@ -35,6 +36,7 @@ const moduleConfig = {
         let azureMockCreated = false;
 
         AzureProvider.remapConstant(fakeURL);
+        window.postponeMapReadyPromise = false;
```
ksercs: Don't you wanna do it vice verse - default `false` and manual explicit true in a separate module?
Just because when we add new tests we don't want to have this specificity enabled

Zedwag: it is false by default, i've added it to restore it to false after new tests are finished. but let's move them to a separate module

ksercs (approved):

PR #29486, opened 2025-04-01
Fix/options manager resetter

sergey-kras:


packages/devextreme-react/src/core/__tests__/nested-option.test.tsx
```diff
@@ -654,7 +654,69 @@ describe('conditional rendering', () => {
       <ComponentWithConditionalOption enableOption={false} />,
     );
 
-    expect(Widget.resetOption.mock.calls.length).toBe(1);
-    expect(Widget.resetOption.mock.calls[0]).toEqual(['option']);
+    expect(Widget.resetOption.mock.calls.length).toBeGreaterThan(0);
```
VasilyStrelyaev: By looking at the code changes in the wrappers, I could not figure out why the number of calls here can increase from 1 to more than one.

Did it really happen? If not, let's maybe leave the condition as strict as it was before?

sergey-kras: It was mistake. I fixed

VasilyStrelyaev (dismissed):

VasilyStrelyaev (approved):

PR #29494, opened 2025-04-01
DateRangeBox, DateBox, Calendar: extend type for date props with null (T1282901)

ksercs:


packages/devextreme/js/ui/calendar.d.ts
```diff
@@ -141,10 +143,11 @@ export interface dxCalendarOptions extends EditorOptions<dxCalendar> {
     hoverStateEnabled?: boolean;
     /**
      * @docid
+     * @type Date|number|string|null
```
ksercs, line 146: NOTE: Angular metadata cannot be generated without this `@type`

packages/devextreme/js/ui/calendar.d.ts
```diff
@@ -226,7 +231,7 @@ export default class dxCalendar extends Editor<dxCalendarOptions> {
      * @publicName reset(value)
      * @public
      */
-    reset(value?: Date | number | string | Array<Date | number | string> | null): void;
+    reset(value?: DateLike | Array< Date | number | string | null >): void;
```
ksercs: NOTE: `DateLike[]` or `Array<DateLike>` breaks `declarations-validation` with no explanations

packages/devextreme/js/ui/date_range_box.d.ts
```diff
@@ -288,10 +298,11 @@ export type Properties = Omit<DateBoxBaseOptions<dxDateRangeBox>, 'inputAttr' |
     startDateText?: string;
     /**
      * @docid dxDateRangeBoxOptions.value
+     * @type Array<Date|number|string|null>
      * @default [null,null]
      * @public
      */
-    value?: Array<Date | number | string>;
+    value?: DateLike[];
```
ksercs, line 305: NOTE: It would be better to specify type like `[DateLike, DateLike]` but it breaks generation

Zedwag (dismissed):

r-farkhutdinov (dismissed):

packages/devextreme/js/ui/date_box.d.ts
```diff
@@ -318,16 +327,18 @@ export interface DateBoxBaseOptions<TComponent> extends dxDropDownEditorOptions<
     displayFormat?: Format;
     /**
      * @docid
+     * @type Date|number|string|null
```
alexslavr: I don't like that we lost `undefined` in angular generation, it's because `type` doctag overrides entire type union. Try adding `|undefined` here and for `min` as well. It should not affect other wrappers, only angular

ksercs: Thanks, fixed it. You're right, it affects angular only
7a54a527009096c19a0db4ff5606f25876d37812

IlyaKhD (approved):

PR #29545, opened 2025-04-07
CardView: calculateCellValue should return dates for date-like type columns

tongsonbarbs:


packages/devextreme/js/__internal/grids/new/grid_core/items_controller/items_controller.ts
```diff
@@ -52,6 +52,37 @@ export class ItemsController {
     this.selectedCardKeys.update(keys);
   }
 
+  private updateColumnDataType(column: Column, value: unknown): unknown {
```
pomahtri: please move this logic into ColumnsController
Also, avoid mutating of the objects, they should be modified immutably

pomahtri (changes requested):

pomahtri:
Also, please switch the PR to grids/cardview/main branch

packages/devextreme/js/__internal/grids/new/grid_core/items_controller/items_controller.ts
Raushen: Hi guys! I'd suggest to create two utility methods such as getActualColumnType and parseValue and split the logic to them and then cover by unit tests.
If you need parsing methods you can [use mine ones](https://github.com/DevExpress/DevExtreme/pull/29402/files#diff-c9fcf48982188e5af75a5cd648ed2ea3d1757fda11e086373fca7ece0008e1a1R78) with moving them to the root level (not Search). They are already [covered by tests](https://github.com/DevExpress/DevExtreme/pull/29402/files#diff-02543bae99bf61173eaa229616c8eb401ce023e923f9740992cebe0e856377bdR325), feel free to add your own cases.

tongsonbarbs: Please see new PR: https://github.com/DevExpress/DevExtreme/pull/29578

PR #29553, opened 2025-04-07
Stepper: add accessibility attributes and tests

anna-shakhova:


packages/devextreme/testing/tests/DevExpress.ui.widgets/stepper.markup.tests.js
```diff
@@ -575,3 +576,92 @@ QUnit.module('Step.hint', moduleConfig, () => {
...
+        assert.strictEqual(this.getStepByIndex(0).attr('role'), 'tab');
+    });
+
+    QUnit.test('Selected step element should have aria-selected=true', function(assert) {
+        this.reinit({
+            items: [{}, {}, {}],
+            selectedIndex: 1,
+        });
+
+        assert.strictEqual(this.getStepByIndex(0).attr('aria-selected'), 'false');
```
EugeniyKiyashko: There are more than one assert in test scenario, comments are required

packages/devextreme/js/__internal/ui/stepper/stepper.ts
```diff
@@ -38,6 +38,8 @@ export const STEP_OPTIONAL_MARK_CLASS = 'dx-step-optional-mark';
 
 export const STEPPER_ITEM_DATA_KEY = 'dxStepperItemData';
 
+export const STEPPER_ARIA_LABEL = 'stepper';
```
EugeniyKiyashko:
```suggestion
export const STEPPER_ARIA_LABEL = 'Stepper';
```

packages/devextreme/js/__internal/ui/stepper/stepper.ts
```diff
@@ -240,6 +245,8 @@ class Stepper extends CollectionWidgetAsync<StepperProperties> {
   _init(): void {
     super._init();
 
+    this.setAria('role', 'tablist');
+    this.setAria('label', STEPPER_ARIA_LABEL);
```
ksercs: Why do we need this basic label?
This attribute is usually used to describe a purpose what element is userd for, like "Order Steps" or "Authentication steps". So maybe it should not be specified by default but we should provide a way to set it on a user side (for example through `elementAttr`)

packages/devextreme/js/__internal/ui/stepper/stepper.ts
```diff
@@ -293,6 +301,13 @@ class Stepper extends CollectionWidgetAsync<StepperProperties> {
     $(this.element()).append(this._$stepsContainer);
   }
 
+  _setAriaOrientation(): void {
+    const isHorizontal = this._isHorizontalOrientation();
+    const orientation = isHorizontal ? ORIENTATION.horizontal : ORIENTATION.vertical;
+
+    this.setAria('orientation', orientation, $(this.element()));
```
ksercs: is third parameter necessary here? 
If it's a `focusTarget`, then not

packages/devextreme/testing/tests/DevExpress.ui.widgets/stepper.markup.tests.js
anna-shakhova: added

packages/devextreme/js/__internal/ui/stepper/stepper.ts
anna-shakhova: removed aria-label

packages/devextreme/js/__internal/ui/stepper/stepper.ts
anna-shakhova: removed

packages/devextreme/js/__internal/ui/stepper/stepper.ts
anna-shakhova: fixed

ksercs (approved):

EugeniyKiyashko (approved):

PR #29556, opened 2025-04-08
CardView: create initial columns from dataSource first object

tongsonbarbs:


packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.ts
```diff
@@ -32,6 +32,8 @@ export class ColumnsController {
 
   public static dependencies = [OptionsController] as const;
 
+  public columnsInitialized = false;
```
Raushen: How about apply a reactive way and store the state of the first item?
`public readonly firstItem = state<Record<string, unknown>>(null);`

Then, if dataSource initialized or changed, you will update the state in data controller:
`this.columnsController.firstItem.update(newItem);`

And finally you should not mutate columnsSettings. Instead of this it is possible to add `this.firstItem` to the dependencies list of `this.columns` and calculate the final state accordingly:

```
this.columns = computed(
      (
        columnsSettings,
        headerFilterRootOptions,
        firstItem,
      ) =>...,
      [
        this.columnsSettings,
        this.headerFilterConfiguration,
        this.firstItem,
      ],
    );
```

packages/devextreme/js/__internal/grids/new/card_view/header_panel/options.test.ts
```diff
@@ -22,7 +22,12 @@ const setup = (options: Options) => {
   const filterController = new FilterController(optionsController);
```
Raushen, line 22: Please ask Roman about a possibility to apply changes to the main branch instead of the draft.

Raushen: Please ensure a test under a comment
`// TODO: resolve the situation when colums are not set in config`
works without column config.

pomahtri: > Please ask Roman about a possibility to apply changes to the main branch instead of the draft.

Yes, it should be rebased to main

tongsonbarbs: I will rebase it to main once the changes is approved so the files changed remains the same. It is easier to check if only changes made are present..

packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.ts
```diff
@@ -72,6 +100,9 @@ export class ColumnsController {
       [this.columns],
     );
 
+    // @ts-expect-error
+    this.columnsInitialized = !!this.columns.value.length;
```
Raushen: There is no using

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
```diff
@@ -216,6 +222,9 @@ export class DataController {
       items = updateItemsImmutable(items, e.changes, dataSource.store());
     }
 
+    // @ts-expect-error
```
Raushen: What is the ts error here?

packages/devextreme/js/__internal/grids/new/card_view/header_panel/options.test.ts
Raushen: It is better to rebase now to avoid a lot of changes in test files.

packages/devextreme/js/__internal/grids/new/card_view/header_panel/options.test.ts
```diff
@@ -22,7 +22,12 @@ const setup = (options: Options) => {
...
   const columnsController = new ColumnsController(optionsController);
```
Raushen, line 23: I'll start a new conversation just not to miss.

`It is better to rebase now to avoid a lot of changes in test files.`

packages/devextreme/js/__internal/grids/new/card_view/header_panel/options.test.ts
```diff
@@ -22,7 +22,12 @@ const setup = (options: Options) => {
...
   const sortingController = new SortingController(optionsController, columnsController);
-  const dataController = new DataController(optionsController, sortingController, filterController);
+  const dataController = new DataController(
```
Raushen, line 25:
```
Please ensure a test under a comment
// TODO: resolve the situation when colums are not set in config
works without column config.
```

Also the test should be changed (and remove column configuration in it)

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
tongsonbarbs: a type error.. to avoid the error I will need to add 'items[0] as Record<string, unknown>'.. do you suggest it?

PR #29557, opened 2025-04-08
Stepper: implement `Step Template` demo

EugeniyKiyashko:


apps/demos/Demos/Stepper/StepTemplate/React/data.ts
```diff
@@ -0,0 +1,22 @@
+export const steps = [{
+  text: 'A',
```
anna-shakhova, line 2: Text is excessive item data for this demo, let's remove it. For other platforms too

ksercs (approved):

anna-shakhova (approved):

apps/demos/Demos/Stepper/StepTemplate/Angular/app/app.component.html
```diff
@@ -0,0 +1,41 @@
+<div class="stepper-label">Custom Step Shape</div>
+<dx-stepper
```
ksercs, line 2: please don't forget to tune a11y in this or another PR. I believe there should be `labelledby` linked to title

apps/demos/Demos/Stepper/StepTemplate/React/data.ts
EugeniyKiyashko: I decided to use the same configuration as in the Overview demo, which demonstrates that the template's result doesn't depend on the configuration

PR #29571, opened 2025-04-09
The "info" icon should be exclamation mark instead of "i" in the Fluent Theme

EugeniyKiyashko:


packages/devextreme-scss/tests/icon-font.test.ts
```diff
@@ -39,7 +39,7 @@ describe('Equals svg to font', () => {
     const countElementFluentFont = getCountElementInFont(`${BASE_PATH}/icons/dxiconsfluent.ttf`);
     const countElementFluentSvg = getCountElementInSvg(`${BASE_PATH}/images/icons/fluent`);
 
-    expect(countElementFluentFont).toBe(countElementFluentSvg);
+    expect(countElementFluentFont).toBe(countElementFluentSvg + 15);
```
pharret31, line 42: What about using named consts instead of plain numbers?

EugeniyKiyashko: Good suggestion, but in this case it might be excessive and seems minor. It looks clear enough and doesn’t seem like a magic number - it reflects the difference between the expected count in the font icon set and the actual SVG icons count

pharret31 (approved):

Shuvaevmedia (approved):

PR #29579, opened 2025-04-10
List should not display "[object Object]" in the aria-label attribute in multiple selection mode, especially when displayExpr and default text field are missed (T1285078)

EugeniyKiyashko:


packages/devextreme/js/localization/messages/en.json
```diff
@@ -55,6 +55,7 @@
         "dxList-selectAll-checked": "Checked",
         "dxList-selectAll-notChecked": "Not checked",
         "dxList-ariaRoleDescription": "List",
+        "dxList-listAriaLabel-itemContent": "List Item Content",
```
```suggestion
        "dxList-listAriaLabel-itemContent": "List item content",
```

Please also replace in other locales

EugeniyKiyashko: Fixed

anna-shakhova (approved):

vladaskorohodova (approved):

PR #29578, opened 2025-04-10
CardView: calculateCellValue should return dates for date-like type columns

tongsonbarbs:

PR #29593, opened 2025-04-11
CardView: create initial columns from dataSource first object

tongsonbarbs:


packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.ts
```diff
@@ -44,17 +46,28 @@ export class ColumnsController {
...
-      (
-        columnsSettings,
-        headerFilterRootOptions,
-      ) => normalizeColumns(
-        columnsSettings ?? [],
-        this.options.normalizeTemplate.bind(this.options),
-      ).map((column) => headerFilterUtils
-        .mergeColumnHeaderFilterOptions(column, headerFilterRootOptions)),
+      (columnsSettings, headerFilterRootOptions, firstItem) => {
+        let finalColumns = columnsSettings;
```
Raushen: I think it is better to extract getActualColumnsSettings method into utils and cover it by tests.

packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.ts
```diff
@@ -44,17 +46,28 @@ export class ColumnsController {
...
+
+        if ((!finalColumns || finalColumns.length === 0) && firstItem) {
```
pomahtri: In case columns options is set to [], it should not be calculated

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
```diff
@@ -218,6 +225,8 @@ export class DataController {
       items = updateItemsImmutable(items, e.changes, dataSource.store());
     }
 
+    this.columnsController.firstItem.update(items[0] as Record<string, unknown> ?? null);
```
pomahtri, line 236: Please add logic to firstly check whether we already have first item, if we have it, it shouldn't be re-assigned

pomahtri (changes requested):

packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.ts
```diff
@@ -30,31 +34,38 @@ export class ColumnsController {
...
     this.columnsConfiguration = this.options.oneWay('columns');
     this.headerFilterConfiguration = this.options.oneWay('headerFilter');
-
-    this.columnsSettings = interruptableComputed(
-      (columnsConfiguration) => preNormalizeColumns(columnsConfiguration ?? []),
-      [
-        this.columnsConfiguration,
-      ],
-    );
+    this.columnsSettings = getActualColumnSettings(this.columnsConfiguration);
```
Raushen: I meant getActualColumnSettings should get `columnsSettings` and `firstItem` as args and could be able to return actualSettings (the equal of `finalColumns` in your code).

Then you can write test and research possible cases. So if Roman said if columnOptions set to [], it should not be changes:

columnsSettings == [],               firstItem == { test: 1 }   // the result is []
columnsSettings == undefined, firstItem = { test: 2}      // the result is ['test']
columnSettings == ['test1'],       firstItem = { test2: 3}    // the result is ['test1']

etc...

It is better to check every case with DataGrid settings.

packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/columns_controller.test.ts
```diff
@@ -157,4 +157,55 @@ describe('ColumnsController', () => {
...
+      };
+
+      const { columnsController } = setup({
+        dataSource: [dataItem],
+      });
+
+      const columns = columnsController.columns.unreactive_get();
+
+      expect(columns).toHaveLength(3);
+      expect(columns.map((col) => col.dataField)).toEqual(['id', 'name', 'age']);
```
pomahtri: Please replace `toEqual` with snapshot of all columns to check dataType too and maybe other column options
Or let's add tests for `dataType` only with similar approach as for dataField right now

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
tongsonbarbs: https://github.com/DevExpress/DevExtreme/pull/29578/files#diff-0ab150c5ebc9f5b8eec72bc0fa0733a8358f3feb9ac7385bdccfb2e2438e2ef5R236

PR #29631, opened 2025-04-16
HtmlEditor: Remove formDialogOptions option && Improve types

marker-dao:


packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
```diff
@@ -25,29 +25,55 @@ import type { DeferredObj } from '@js/core/utils/deferred';
...
 import { prepareScrollData } from '@ts/ui/text_box/m_utils.scroll';
 
 import type { AIDialogResult, AIDialogShowPayload } from './ui/aiDialog';
 import AIDialog from './ui/aiDialog';
 
-const HTML_EDITOR_CLASS = 'dx-htmleditor';
 const QUILL_CONTAINER_CLASS = 'dx-quill-container';
 const QUILL_CLIPBOARD_CLASS = 'ql-clipboard';
+const HTML_EDITOR_CLASS = 'dx-htmleditor';
+const HTML_EDITOR_STYLING_MODE_PREFIX = 'dx-htmleditor-';
```
ksercs: I don't see any point to create a new variable for this. You can just use `${HTML_EDITOR_CLASS}-` for this

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
```diff
@@ -25,29 +25,55 @@ import type { DeferredObj } from '@js/core/utils/deferred';
...
 const HTML_EDITOR_CONTENT_CLASS = 'dx-htmleditor-content';
 
 const ANONYMOUS_TEMPLATE_NAME = 'htmlContent';
 
-const isIos = devices.current().platform === 'ios';
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type QuillInstance = any;
+
+type ModulesConfig = Record<string, unknown>;
+type TextFormat = Record<HtmlEditorFormat | string, string>;
```
ksercs: [mistake]
I see it's used not only for text formats but for any formats, e.g. it can be a table format or another.
Also I see you state that this object's values are strings but actually it's much wider. For example, here's a link format's possible value: ` {href: "https://www.google.com/",  text: "Google",   title: "Go to Google" }`

I suggest to not type it for now, it's not so fast

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
```diff
@@ -25,29 +25,55 @@ import type { DeferredObj } from '@js/core/utils/deferred';
...
+}
+
+interface Bounds {
+  bottom: number;
+  height: number;
+  left: number;
+  right: number;
+  top: number;
+  width: number;
+}
```
ksercs, line 72: [excess code?]
I see Quill's original type does now have a half of these properties: https://quilljs.com/docs/api#getbounds

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
```diff
@@ -57,17 +83,17 @@ class HtmlEditor extends Editor<Properties> {
...
 
-  _quillInstance?: any;
+  _quillInstance?: QuillInstance;
 
   _cleanCallback!: Callback;
 
   _contentInitializedCallback!: Callback;
 
-  _htmlConverter!: Converter;
+  _htmlConverter?: Converter;
```
ksercs, line 90: [excess change?]
For what? I see this property is always set but a bit later in constructor. I believe it was written this way to not duplicate option chaining operators you introduced now

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
```diff
@@ -412,94 +392,131 @@ class HtmlEditor extends Editor<Properties> {
...
+        variables: this._getModuleConfigByOption('variables'),
+      },
+      this._getCustomModules(),
+    );
+
+    return modulesConfig;
   }
 
-  _getModulesConfig() {
+  _getUploaderConfig(): {
```
ksercs: [naming]
```suggestion
  _getUploaderModuleConfig(): {
```
I suggest you to update it for all similar modules to point explicitly that it's a config for module. "uploaderConfig" sound like a config for subcomponent, "keyboardConfig" sounds just strange)

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
```diff
@@ -614,24 +636,26 @@ class HtmlEditor extends Editor<Properties> {
...
+        this._processHtmlContentUpdating(currentValue);
         break;
       }
       case 'value': {
-        this._processHtmlContentUpdating(args.value);
+        this._processHtmlContentUpdating(value);
+
+        const { value: currentValue } = this.option();
 
-        // NOTE: value can be optimized by Quill
```
ksercs, line 630: Why did u remove this note? Do you think it's obvious?

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
marker-dao: I compiled this type based on the values that were obtained by the function in realtime. I propose to make them optional.

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
marker-dao: > this property is always set but a bit later in constructor

in constructor: 

```
_prepareHtmlConverter(): void {
    const { converter } = this.option();
    if (converter) {
      this._htmlConverter = converter;
    }
}
```

so this prop is not always set, cause in the options it looks like this:
```
_getDefaultOptions() {
    converter: null,
}
```

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
marker-dao: [081e5880faafd04de7b57955e52c34b00540bbbb]

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
marker-dao: [081e5880faafd04de7b57955e52c34b00540bbbb]

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
marker-dao: [081e5880faafd04de7b57955e52c34b00540bbbb]

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
marker-dao: [081e5880faafd04de7b57955e52c34b00540bbbb]

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
marker-dao: [081e5880faafd04de7b57955e52c34b00540bbbb]

ksercs (approved):

Zedwag (approved):

PR #29640, opened 2025-04-17
cache timezone calculation

Ambrozy:


packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -23,7 +23,6 @@ export interface TimezoneData extends TimezoneLabel {
 const toMs = dateUtils.dateToMilliseconds;
```
wdevfx, line 23: Let's cover this ```getTimeZones``` function's new parameter by tests
We already have some QUnits that check this function, so the better way will be to just add some new test cases:
https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/common.tests.js#L905

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -333,32 +332,32 @@ const addOffsetsWithoutDST = (date: Date, ...offsets: number[]): Date => {
...
   date = new Date(),
-): Promise<TimezoneLabel[]> => macroTaskArray.map(
+  timeZones = timeZoneList.value,
+): TimezoneData[] => timeZones.map((timezoneId) => ({
+  id: timezoneId,
+  title: getTimezoneTitle(timezoneId, date),
+  offset: calculateTimezoneByValue(timezoneId, date),
+}));
+
+// START getTimeZonesCache.ts
```
wdevfx: Let's get rid of these section comments

If we would like to separate this code section -> let's just move it to another TS module

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -333,32 +332,32 @@ const addOffsetsWithoutDST = (date: Date, ...offsets: number[]): Date => {
...
+let timeZoneDataCache: TimezoneLabel[] = [];
```
wdevfx, line 345: Question.

Should we have some method to clear this cache?

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -333,32 +332,32 @@ const addOffsetsWithoutDST = (date: Date, ...offsets: number[]): Date => {
...
+const cacheTimeZones = (
+  date = new Date(),
+): Promise<void> => macroTaskArray.map(
```
wdevfx: Question.

Should we check that the cache already exists and do nothing in this case?
E.g now for two Scheduler components we will perform calculations twice

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -333,32 +332,32 @@ const addOffsetsWithoutDST = (date: Date, ...offsets: number[]): Date => {
...
-  GET_TIMEZONES_BATCH_SIZE,
-);
-
-const getTimeZoneLabel = (
-  timezoneId: string,
-  date = new Date(),
-): TimezoneLabel => ({
-  id: timezoneId,
-  title: getTimezoneTitle(timezoneId, date),
+  10,
```
wdevfx: Let's move this ```magic number``` to some const with name

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
Ambrozy: Global cache

pomahtri (dismissed):

pomahtri (approved):

wdevfx (approved):

PR #29649, opened 2025-04-18
Improve appointment accessibility

Ambrozy:

### DIFF
--- a/e2e/testcafe-devextreme/tests/scheduler/common/a11y/appointment.ts
+++ b/e2e/testcafe-devextreme/tests/scheduler/common/a11y/appointment.ts
@@ -43,7 +43,7 @@ const currentDate = Date.UTC(2021, 1, 1);
       .expect(appointment.getAriaLabel())
       .eql('App 1: February 1, 2021, 12:00 PM - 1:00 PM')
       .expect(await appointment.getAriaDescription())
-      .eql('Group: resource1; Resource 1: resource1');
+      .eql('Group: resource1; Group 1: resource1');
 
     await a11yCheck(t, checkOptions, '#container');
   }).before(async () => {
@@ -63,7 +63,7 @@ const currentDate = Date.UTC(2021, 1, 1);
             text: 'resource1',
             id: 1,
           }],
-          label: 'Resource 1',
+          label: 'Group 1',
         },
       ],
     });
@@ -77,7 +77,7 @@ const currentDate = Date.UTC(2021, 1, 1);
       .expect(appointment.getAriaLabel())
       .eql('App 1: February 1, 2021, 12:00 PM - 1:00 PM')
       .expect(await appointment.getAriaDescription())
-      .eql('Group: resource11, resource21; Resource 1: resource11; Resource 2: resource21, resource22');
+      .eql('Group: resource11, resource21; Group 1: resource11; Group 2: resource21, resource22');
 
     await a11yCheck(t, checkOptions, '#container');
   }).before(async () => {
@@ -98,7 +98,7 @@ const currentDate = Date.UTC(2021, 1, 1);
             text: 'resource11',
             id: 1,
           }],
-          label: 'Resource 1',
+          label: 'Group 1',
         },
         {
           fieldExpr: 'groupId2',
@@ -109,7 +109,7 @@ const currentDate = Date.UTC(2021, 1, 1);
             text: 'resource22',
             id: 2,
           }],
-          label: 'Resource 2',
+          label: 'Group 2',
         },
       ],
     });
--- a/e2e/testcafe-devextreme/tests/scheduler/common/a11y/appointment.ts
+++ b/e2e/testcafe-devextreme/tests/scheduler/common/a11y/appointment.ts
@@ -7,49 +7,54 @@ import { checkOptions } from './axe_options';
 fixture.disablePageReloads`a11y - appointment`
   .page(url(__dirname, '../../../container.html'));
 
+const appointmentItem = {
+  text: 'App 1',
+  startDate: Date.UTC(2021, 1, 1, 12),
+  endDate: Date.UTC(2021, 1, 1, 13),
+};
+const currentDate = Date.UTC(2021, 1, 1);
+
 ['month', 'week', 'day'].forEach((currentView) => {
-  test(`appointment should have correct aria-label without grouping (${currentView})`, async (t) => {
+  test(`appointment should have correct aria-label without description (${currentView})`, async (t) => {
     const scheduler = new Scheduler('#container');
+    const appointment = scheduler.getAppointment('App 1');
 
     await t
-      .expect(
-        scheduler.getAppointment('App 1').element.attributes['aria-label'],
-      )
-      .eql(undefined);
+      .expect(appointment.getAriaLabel())
+      .eql('App 1: February 1, 2021, 12:00 PM - 1:00 PM')
+      .expect(await appointment.hasAriaDescription())
+      .notOk();
 
     await a11yCheck(t, checkOptions, '#container');
   }).before(async () => {
     await createWidget('dxScheduler', {
-      dataSource: [{
-        text: 'App 1',
-        startDate: new Date(2021, 1, 1, 12),
-        endDate: new Date(2021, 1, 1, 13),
-      }],
+      timeZone: 'UTC',
+      dataSource: [appointmentItem],
       currentView,
-      currentDate: new Date(2021, 1, 1),
+      currentDate,
     });
   });
 
   test(`appointment should have correct aria-label with one group (${currentView})`, async (t) => {
     const scheduler = new Scheduler('#container');
-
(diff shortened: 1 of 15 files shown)


e2e/testcafe-devextreme/tests/scheduler/common/a11y/appointment.ts
```diff
@@ -7,49 +7,54 @@ import { checkOptions } from './axe_options';
 fixture.disablePageReloads`a11y - appointment`
   .page(url(__dirname, '../../../container.html'));
 
+const appointmentItem = {
+  text: 'App 1',
+  startDate: Date.UTC(2021, 1, 1, 12),
```
wdevfx, line 12: Not a comment at all

Also, we may set dates in the Scheduler as UTC string
It's very human-readable :)
E.g: ```Date.UTC(2021, 1, 1, 12)``` -> ```'2021-02-01T12:00:00Z'```

Btw, now we set dates as numbers because the ```Date.UTC``` method returns number

e2e/testcafe-devextreme/tests/scheduler/common/a11y/appointment.ts
```diff
@@ -58,32 +63,33 @@ fixture.disablePageReloads`a11y - appointment`
             text: 'resource1',
             id: 1,
           }],
+          label: 'Resource 1',
```
wdevfx: Minor comment.

Let's rename this label to ```Group 1``` to make test data more obvious
Now we have ```resource1``` as resource value and ```Resource 1``` as group value, it's a little bit confusing

e2e/testcafe-devextreme/tests/scheduler/common/dragAndDrop/T1235433.ts
```diff
@@ -7,91 +7,7 @@ import { scrollTo } from '../../helpers/utils';
...
-
-  await t.click(scheduler.toolbar.viewSwitcher.getButton('Timeline Work Week').element)
-    .expect(scheduler.checkViewType('timelineWorkWeek'))
-    .ok();
-  await scrollTo(2400, 0);
-  await dragAppointmentByCircle(t, scheduler.getAppointment('Book 1'), appointmentDescriptions, appointment1Times);
-  await scrollTo(3400, 0);
-  await dragAppointmentByCircle(t, scheduler.getAppointment('Book 2'), appointmentDescriptions, appointment2Times);
-
-  await t.click(scheduler.toolbar.viewSwitcher.getButton('Timeline Month').element)
```
wdevfx, line 68: Question.

In the previous impl we also check view switching inside the test case
Should we add more test cases that cover it?

packages/devextreme/js/__internal/scheduler/appointments/appointment/agenda_appointment.ts
```diff
@@ -12,17 +11,13 @@ export class AgendaAppointment extends Appointment {
     return this.$element().find(`.${APPOINTMENT_CONTENT_CLASSES.AGENDA_MARKER}`);
   }
 
-  get resourceProcessor(): ResourceProcessor {
-    const getResourceProcessor = this.option('getResourceProcessor') as () => ResourceProcessor;
-    return getResourceProcessor();
-  }
-
   _renderResourceList(): void {
+    const parent = this.$element().find(`.${APPOINTMENT_CONTENT_CLASSES.APPOINTMENT_CONTENT_DETAILS}`);
```
wdevfx: Because ```getAppointmentResourcesValues``` is async, let's move back parent element selection inside ```then```
Because for the resource uploading time we may have unexpected (or expected) re-render
And this parent DOM node may became detached (unactual)

packages/devextreme/js/__internal/scheduler/appointments/appointment/m_appointment.ts
```diff
@@ -2,20 +2,18 @@ import { move } from '@js/common/core/animation/translator';
...
 import $ from '@js/core/renderer';
 import { extend } from '@js/core/utils/extend';
-import { isDefined } from '@js/core/utils/type';
 import Resizable from '@js/ui/resizable';
+import type { ResourceProcessor } from '@ts/scheduler/resources/resource_processor';
 import { hide, show } from '@ts/ui/tooltip/m_tooltip';
 
 import {
-  ALL_DAY_APPOINTMENT_CLASS,
+  ALL_DAY_APPOINTMENT_CLASS, APPOINTMENT_CONTENT_CLASSES,
```

Formatting, let's move added const to a new line
In the future, it will be perfect to add ```prettier``` to our repo :)

packages/devextreme/js/__internal/scheduler/appointments/appointment/m_appointment.ts
```diff
@@ -178,46 +185,22 @@ export class Appointment extends DOMComponent {
...
-    $element.attr('aria-roledescription', `${ariaLabel}, `);
-
-    const id = `dx-${new Guid()}`;
-
-    $element.attr('aria-describedby', id);
-    $element.find('.dx-item-content').attr('id', id);
+    $element.attr('aria-label', getAriaLabel(this.option()));
+
+    // eslint-disable-next-line no-void
+    void getAriaDescription(this.option())
```
wdevfx, line 194: [Critical]

Now it looks like for each appointment (in render) we download all resources
Functions call chain is:
- text_utils.getAriaDescription
- text_utils.getResourceText
- resource_processor.getAppointmentResourcesValues
- resource_processor.loadResource

And the ```resource_processor.loadResource``` under the hood calls DataSource ```load``` method here:
[Link to the code](https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/scheduler/resources/resource_processor.ts#L29)

Therefore, it looks like that we call ```DataSource.load()``` for each appointment render
The resources DataSource may be remote and it may produce additional requests

Let's double-check it

packages/devextreme/js/__internal/scheduler/appointments/appointment/text_utils.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const date = new Date(result);
+  const gridDate = options.timeZoneCalculator?.createDate(
+    date,
+    { path: PathTimeZoneConversion.fromSourceToGrid },
+  );
+
+  return gridDate ?? date;
+};
+const getDateText = (options: AppointmentProperties): string => {
```
wdevfx: Minor comment.

Let's add an empty line between functions
(One more poin for ```prettier```)

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/integration.resources.tests.js
```diff
@@ -457,7 +457,8 @@ QUnit.module('Integration: Resources', moduleConfig, () => {
             }]
         });
 
-        assert.equal(loadStub.callCount, 1, 'Resources are loaded only once');
+        // TODO(6): fix it when after resource refactoring
+        assert.equal(loadStub.callCount, 2, 'Resources are loaded only once');
```
wdevfx, line 466: This test assertion change may be very dangerous (see critical comment)
So, let's not change it and fix the logic in terms of this PR

Also, let's add more test that checks resources loading, to make right behavior more obvious

packages/devextreme/js/__internal/scheduler/appointments/appointment/m_appointment.ts
Ambrozy: It's not true. We load resources only once. There is a test for it. Duplication exists and it existed before because we have 2 mechanisms to download resources. But for this I created the different task

wdevfx: Yup, I see
We call ```load``` but prevent real loading here:
[Link to the code](https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/scheduler/resources/resource_processor.ts#L79)

But loading two times isn't good too (but not so critical)
Let's create a separate task to fix it until major release

e2e/testcafe-devextreme/tests/scheduler/common/dragAndDrop/T1235433.ts
Ambrozy: I think no, because initial problem, that was covered under tests didn't dependent on view changes

e2e/testcafe-devextreme/tests/scheduler/common/a11y/appointment.ts
Ambrozy: Yes, I left it like this to diverse our test data. Generally it's better to use obvious way. For this test I would like to stay it different

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/integration.resources.tests.js
Ambrozy: I fixed one test to make it more obvious

wdevfx (approved):

PR #29652, opened 2025-04-18
refactor resource processor with promises

Ambrozy:


Ambrozy:
Pipeline didn't pass, so it was removed

PR #29655, opened 2025-04-18
🐞 T1284200. PivotGrid. Select all of field chooser doesn't work with remote dataSource

wdevfx:


e2e/testcafe-devextreme/tests/pivotGrid/headerFilter.ts
```diff
@@ -52,3 +55,40 @@ test.meta({ unstable: true })('Header filter popup', async (t) => {
...
+    ],
+    store: new Array(100).fill(null).map((_, idx) => ({
+      id: idx,
+    })),
+  },
+  allowSorting: true,
+  allowFiltering: true,
+  fieldPanel: {
+    visible: true,
+  },
```
Ambrozy, line 93: This configuration works even without fix and bug appears only after cancel click
1. Open filter
2. Select all -> all selected
3. Select all -> all deselected
4. Click cancel
5. Open filter
6. Select all -> bug

wdevfx: Discussed, in the PivotGrid it looks like that dxList has biggest page size
And the id = 25 on the first page

I changed excluded value to the ```70``` to be 100% sure that this value isn't on the first page :)

Ambrozy (approved):

PR #29680, opened 2025-04-23
Fixed focused link bg color

surohak:


packages/testcafe-models/dataGrid/index.ts
```diff
@@ -69,6 +69,7 @@ export const CLASS = {
   summaryTotal: 'dx-datagrid-summary-item',
   scrollableContainer: 'dx-scrollable-container',
   columnsSeparator: 'dx-datagrid-columns-separator',
+  dxLink: 'dx-link',
```
wdevfx: Let's rollback changes in DataGrid ```POM``` because we don't need them for now

surohak: done ✅

wdevfx (approved):

surohak:
https://github.com/DevExpress/DevExtreme/pull/29680

PR #26052, opened 2023-11-17
Prepare internal packages

Lykoi18:

PR #29697, opened 2025-04-24
feat(StateManager): create the first version of the state manager

chaosmirage:


packages/devextreme/build/gulp/js-bundles.js
```diff
@@ -29,15 +29,44 @@ const DEBUG_BUNDLES = BUNDLES.concat([ '/bundles/dx.custom.js' ]);
...
+                            return;
+                        }
+
+                        const newRequest = resource.request.replace('reactive/index', 'reactive/development');
+
+                        resource.request = newRequest;
+                    }
+                )] : [])
+        ]
+    }, webpackConfig) ;
```
IlyaKhD: It's hard to read. AFAICS, it can be refactored:

```js
    const plugins = [];
    const internalBuild = env.BUILD_INTERNAL_PACKAGE || env.BUILD_TEST_INTERNAL_PACKAGE;

    if (internalBuild) {
        plugins.push(new webpack.NormalModuleReplacementPlugin(/(.*)\/license_validation/, resource => {
            resource.request = resource.request.replace('license_validation', 'license_validation_internal');
        }));
    }

    if (mode === 'production' && internalBuild) {
        plugins.push(new webpack.NormalModuleReplacementPlugin(/(.*)\/state_manager/, (resource) => {/*...*/}));
    }

    if (mode === 'development' && !internalBuild) {
        plugins.push(new webpack.NormalModuleReplacementPlugin(/(.*)\/reactive/,(resource) => {/*...*/}));
    }

    return Object.assign({ plugins }, webpackConfig); // I would change it to `assign(webpackConfig, { plugins })`, looks like a typo from the past.
```

IlyaKhD: Could you explain these conditions combination, why `state_manager/production` is removed only from internal package, and reguilar one removes `reactive/development`?
Can it be determined by `mode` only?

IlyaKhD: Maybe it'd be good to collect all `state_manager` paths in here:
packages/devextreme/build/gulp/state_manager/constants.js, to explicitly link all the `state_manager`-related parts together.

packages/devextreme/build/gulp/state_manager/transpile_state_manager_prod.js
```diff
@@ -0,0 +1,59 @@
...
+const babel = require('@babel/core');
+const transpileConfig = require('../transpile-config');
+const { STATE_MANAGER_FOLDER_PATH } = require('./constants');
+
+const ERROR_PREFIX = 'Error during transpiling the state manager for prod mode:';
+const PRODUCTION_FILE_NAME = 'production.js';
+
+function transpileStateManagerProd(isEsm) {
+    return through2.obj(function(file, enc, callback) {
+        if (file.path.includes(path.join(STATE_MANAGER_FOLDER_PATH, 'setup_state_manager.js'))) {
```
IlyaKhD:
```suggestion
        if (file.path.includes(SETUP_STATE_MANAGER)) {
```
I'd also move this as constant to packages/devextreme/build/gulp/state_manager/constants.js

packages/devextreme/build/gulp/state_manager/transpile_state_manager_prod.js
```diff
@@ -0,0 +1,59 @@
...
+
+const through2 = require('through2');
+const path = require('path');
+const fs = require('fs');
+const babel = require('@babel/core');
+const transpileConfig = require('../transpile-config');
+const { STATE_MANAGER_FOLDER_PATH } = require('./constants');
+
+const ERROR_PREFIX = 'Error during transpiling the state manager for prod mode:';
+const PRODUCTION_FILE_NAME = 'production.js';
```
IlyaKhD: Maybe it's worth moving this to constants.js

packages/devextreme/build/gulp/state_manager/transpile_state_manager_prod.js
```diff
@@ -0,0 +1,59 @@
...
+                    let productionContent = fs.readFileSync(productionFilePath, 'utf8');
+
+                    if (!isEsm) {
+                        const babelConfig = {
+                            ...transpileConfig.tsCjs,
+                            filename: productionFilePath
+                        };
+
+                        try {
+                            const transformResult = babel.transformSync(productionContent, babelConfig);
```
IlyaKhD: Is it possible to avoid extra babel call? I'm not sure, but something similar is made with licensing moudles: https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/build/gulp/npm.js#L82-L91,

(added in #26052)

IlyaKhD (commented):
A few comments on changes in the build scripts

packages/devextreme/js/__internal/core/state_manager/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    if (a === null || b === null || a === undefined || b === undefined) {
+      return a === b;
+    }
+
+    if (typeof a !== 'object' || typeof b !== 'object') {
+      return a === b;
+    }
+
+    const visitedA = visited.get(a);
```
IlyaKhD: Naming is confusing, I guess, it's `b`:
```suggestion
    const visitedB = visited.get(a);
```

packages/devextreme/js/__internal/core/state_manager/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (visitedA) {
+      return visitedA === b;
+    }
+
+    visited.set(a, b);
```
IlyaKhD: I doubt it works as expected in all cases because of using strict equal in the return statement above `return visitedB === b`
Consider this:



image

packages/devextreme/js/__internal/core/state_manager/utils.ts
```diff
@@ -0,0 +1,108 @@
...
+  function iter(a: unknown, b: unknown, visited = new WeakMap()): boolean {
+    if (a === b) return true;
+
+    if (a === null || b === null || a === undefined || b === undefined) {
+      return a === b;
+    }
+
+    if (typeof a !== 'object' || typeof b !== 'object') {
+      return a === b;
+    }
```
IlyaKhD: At this moment we already know that `a !== b`:
```suggestion
    if (a === null || b === null || a === undefined || b === undefined) {
      return false;
    }

    if (typeof a !== 'object' || typeof b !== 'object') {
      return false;
    }
```

IlyaKhD: Atually, I'd refactor the whole function to make separated object's comparison, like this:

```ts
if (typeof a === 'object' && typeof b === 'object') {
return equalObjects(a,b);
}
```

packages/devextreme/js/__internal/core/state_manager/__tests__/utils.test.ts
```diff
@@ -0,0 +1,183 @@
+import {
```
IlyaKhD: I think it's better to move both test files next to the modules they’re paired with, as done elsewhere in the __internal folder.

The mocks file can still be placed in a folder, but to avoid confusion, it’s better to give it a non-canonical name — like test_helpers or tests, if there are no issues with removing it during build.

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
```diff
@@ -0,0 +1,66 @@
...
+    (controller: IController, id: string) => void>;
+
+  constructor(logger: ILogger) {
+    this.logger = logger;
+    this.controllerRegisteredEmitter = new EventEmitter<(controller: IController,
+    id: string) => void
+      >(
+      'controllerRegistered',
+      logger,
+      );
```
IlyaKhD: It's really hard to read.

```suggestion
    this.controllerRegisteredEmitter = new EventEmitter<(controller: IController, id: string) => void>(
      'controllerRegistered',
      logger,
    );
```

or 

```ts
type ControllerEventEmitter = (controller: IController, id: string) => void;
```

```suggestion
    this.controllerRegisteredEmitter = new EventEmitter<ControllerEventEmitter>(
      'controllerRegistered',
      logger,
    );
```

packages/devextreme/js/__internal/core/state_manager/interfaces.ts
```diff
@@ -0,0 +1,163 @@
+export type IStateManagerActionType = 'UPDATE' | 'INITIALIZE' | 'CLEAR_HISTORY';
+
+export interface IStateChangePayload {
```
IlyaKhD: Please, do not use leading `I` for interfaces. AFAIKS, Hungarian notation isn’t used for interfaces anywhere in the codebase. (I've searched for `(interace|type) I[A-Z]` in the devextreme package, skipping generated code)

```suggestion
export interface StateChangePayload {
```

IlyaKhD: `I` for other types other then intrefaces is also unnecessary, I mean the `IStateManagerActionType`above

IlyaKhD: The file should be renamed to `types.ts` (There is a number of other files in the `__internal` folder)

IlyaKhD: JFYI: to deal with name conflicts you can use aggregated exports:

`export type * as StateManagementTypes from './types'`
or
`import type * as StateManagementTypes from './types'`

packages/devextreme/js/__internal/core/state_manager/interfaces.ts
```diff
@@ -0,0 +1,163 @@
+export type IStateManagerActionType = 'UPDATE' | 'INITIALIZE' | 'CLEAR_HISTORY';
```
IlyaKhD: I'm not sure that each enitity requiers an Interface, this results in too many absratcion that are likely will never get another implementation

packages/devextreme/js/__internal/core/state_manager/interfaces.ts
```diff
@@ -0,0 +1,163 @@
...
+  devToolsConnector: IDevToolsConnector;
+  logger: ILogger;
+}
+
+export interface IStateManagerFactoryOptions extends Partial<IStateManagerConfig> {
+  componentName: string;
+  maxHistorySize?: number;
+  stateContainerManagers?: IStateContainerManager[];
+  stateTracker?: IStateTracker;
+  stateHistory?: IStateHistory;
```
IlyaKhD: These props are already derived from `IStateManagerConfig` :

```suggestion
```

packages/devextreme/js/__internal/core/state_manager/interfaces.ts
```diff
@@ -0,0 +1,163 @@
...
+  logLevel?: ILogLevel;
```
IlyaKhD: Having logLevel in some factory options seem a bit weird, usually it's on application level

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
```diff
@@ -0,0 +1,270 @@
...
+    return this.controllerRegistry.registerController(controller, id);
+  }
+
+  restoreState(state: IState): void {
+    try {
+      this.isRestoring = true;
+      const controllers = this.controllerRegistry.getAllControllers();
+
+      Object.keys(state).forEach((controllerId) => {
+        const controller = controllers[controllerId];
```
IlyaKhD: Why not using the absraction method?
Using entries outside the smart collection breaks incapsulation.

```suggestion
        const controller = this.controllerRegistry.getController(controllerId);
```

IlyaKhD: If there is a need of making a snapshot of the `controllerRegistry` at specific moment, then we should review the case, as it's very unclear and hides some concurrency problems.

In this case we might say that `ControllerRegistry`can't serve its purpose well

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
```diff
@@ -0,0 +1,66 @@
+import { EventEmitter } from './event_emitter';
```
IlyaKhD: Controller Registry seems to be a redundant entity in the scope of  StateManager. It has no state-management specifc logic.

I propose making it busines-logic-agnostic and extracting it in the general `core`scope, like this:



```
class TrackableCollection<T> {
  private items: Record<string, TItem> = {};

  private readonly logger: ILogger;
  private readonly caption: string; // Just for log messages, it's 'conroller' in this case. Not sure if it's reqlly needed
  constructor(logger: ILogger, caption: string) {...}

  add(item: TItem, key: string): string {...}
  get(id: string): TItem | undefined { ... }
  entries(): Record<string, TItem> { ... } // would be good if it's aligned with `Object.entries` (but has better typings than it 🙂)
  
  onAdding(callback: (item: TItem, key: string) => void): void { ... } // or `onAdded` - depends on the moment it's called.
}
```

IlyaKhD: It seems that the `entities` /  `getAllControllers` method is not needed, see another comment in `state_manager.ts` which is the only place I found where the `getAllControllers` method is used

packages/devextreme/js/__internal/core/state_manager/logger.ts
```diff
@@ -0,0 +1,70 @@
+/* eslint-disable no-console */
```
IlyaKhD, line 1: **Queston** **Minor**

Is `Logger` removed from production build?

IlyaKhD: I'm not sure about how we can leverage from using logging in devextreme packages, but if we will - better use some 3rd-party, e.g. [winston](https://www.npmjs.com/package/winston) (it's still quite popular and we use it in a few places)

packages/devextreme/js/__internal/core/state_manager/setup_state_manager.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+const CONTROLLER_SIGN = 'Controller';
+
+function isController(instance: unknown, controllerSign: string): instance is IController {
+  if (instance) {
+    return typeof instance === 'object'
+         && 'constructor' in instance
+         && 'name' in instance.constructor
+         && instance.constructor.name.includes(controllerSign);
```
IlyaKhD, line 26: String-names-based logic is generally a bad practise, it may become really hard to maintain.
Isn't it possible to use `instanceof`? I mean adding some base 'marker' class and force all the controllers to inherit from it

packages/devextreme/js/__internal/core/state_manager/setup_state_manager.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export const setupStateManager = (options: IStateManagerInitializerOptions): void => {
+  const {
+    diContext,
+    componentName,
+    logLevel = 'warn',
+    controllerSign = CONTROLLER_SIGN,
+  } = options;
+
+  const isDevelopmentMode = process.env.NODE_ENV === 'development';
```
IlyaKhD, line 43: I think it's important to extract this to some common core module so that we can track new dependencies on developmentMode

packages/devextreme/js/__internal/core/state_manager/logger.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { ILogger } from './interfaces';
+
+export type ILogLevel = 'debug' | 'info' | 'warn' | 'error';
```
IlyaKhD: There is another `ILogLevel`: https://github.com/DevExpress/DevExtreme/pull/29697/files#diff-f12a289a1cd2c2bc014821dda5a0e463e3dd9c06639643d8d6a80393df3d9c77R78

packages/devextreme/js/__internal/core/state_manager/__tests__/mocks.ts
```diff
@@ -0,0 +1,345 @@
...
+            payload: {
+              ...change.payload,
+              path: `${controllerId}.${change.payload.path}`,
+            },
+          };
+          this.stateChangeCallback(updatedChange);
+        }
+      });
+    }
+  }
```
IlyaKhD: It seems to be overcomplicated for a mock, which should have logic at all...

Isn't it possible to just set up mock static values that mocked enity should return in a specific case?
If it's hard to do, then maybe such things shouldn't be mocked (and should be extracted as a standalone abstraction)

IlyaKhD: Btw, are there any tests for mocked entities?

packages/devextreme/js/__internal/core/state_manager/setup_state_manager.ts
```diff
@@ -0,0 +1,84 @@
...
+      diContext.registerInstance(
+        StateManager,
+        stateManager,
+      );
+
+      let areAnyControllerInitialized = false;
+
+      const registerControllerInStateManager: DecoratorFunction = (instance) => {
+        if (isController(instance, controllerSign)) {
+          const injectedStateManager = diContext.get(StateManager);
```
IlyaKhD: Why getting the instance of StateManager from DI container required?
The setup method is already coupled with the factory and I can't see how another intance of SM can be registered in the container

packages/devextreme/js/__internal/core/state_manager/observable_state_container_manager.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    if (!onChange || typeof onChange !== 'function') {
+      this.logger.error('onChange callback is required');
+      return;
+    }
+
+    // eslint-disable-next-line spellcheck/spell-checker
+    let previousValue = stateContainer.unreactive_get();
+
+    const originalUpdate = stateContainer.update;
```
IlyaKhD: I just have to let it out: I'm still pessimistic about "a hive of atoms" - decentralized state with individual state manupulations. Though I must admit, it looks interesting.

Anyway it's out of this PR scope, I'm just sharing my thoughts.

packages/devextreme/js/__internal/core/state_manager/observable_state_container_manager.ts
```diff
@@ -0,0 +1,189 @@
...
+  IMaybeObservableStateContainer,
+  IObservableStateContainer,
+  IStateChange,
+  IStateChangeCallback,
+  IStateChangePayload,
+  IStateContainerManager,
+} from './interfaces';
+import { areEqual, deepCopy } from './utils';
+
+export class ObservableStateContainerManager implements IStateContainerManager {
```
IlyaKhD: The name confuses me a bit. I mean the `container`. Isn't it a wrapper/adapter or smth like?

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    this.controllers[controllerId] = controller;
+    this.logger.debug(`Registered controller: ${controllerId}`);
+
+    this.controllerRegisteredEmitter.emit(controller, controllerId);
+
+    return controllerId;
+  }
+
+  getController(id: string): IController | null {
```
IlyaKhD: I believe `undefined` is more common in such cases (regarding TS)
```suggestion
  getController(id: string): IController | undefined {
```

packages/devextreme/js/__internal/core/state_manager/interfaces.ts
chaosmirage: The state manager developer will enable it per component for debugging.

packages/devextreme/js/__internal/core/state_manager/logger.ts
chaosmirage: > How can we leverage the use of logging in DevExtreme packages?

Logger is created only for debugging the state manager.
It may, in the future, evolve into something like https://github.com/reatom/reatom/tree/v3/packages/logger.
I don’t think it is worth using a common logger library for it.

> We use it in a few places.

`pnpm list -r winston`

I don’t see it. Do you mean in some other repos?

> Is Logger removed from production build? 

Yes

packages/devextreme/js/__internal/core/state_manager/__tests__/utils.test.ts
chaosmirage: What is the reason for avoiding the common convention of placing tests in the `__tests__` folder?

> I think it's better to move both test files next to the modules they're paired with, as done elsewhere in the __internal folder
>

Why do we have different approaches using different folders?

```jsx
/packages/devextreme-react/src/core/__tests__
/packages/devextreme-react/src/core/configuration/__tests__
/packages/devextreme-react/src/core/configuration/react/__tests__
/packages/devextreme-vue/npm/cjs/core/__tests__
/packages/devextreme-vue/npm/esm/core/__tests__
/packages/devextreme-vue/src/core/__tests__
/packages/devextreme/js/__internal/scheduler/r1/appointment_popup/__tests__
/packages/devextreme/js/__internal/scheduler/r1/filterting/__tests__
/packages/devextreme/js/__internal/scheduler/r1/timezone_calculator/__tests__
/packages/devextreme/js/__internal/scheduler/r1/utils/__tests__
/packages/devextreme/js/__internal/scheduler/workspaces/view_model/__tests__
/packages/devextreme/js/renovation/common/__*tests__*
```

> The mocks file can still be placed in a folder, but to avoid confusion, it's better to give it a non-canonical name — like test_helpers or tests
>

There is no need for it if `__tests__` will be used.

packages/devextreme/js/__internal/core/state_manager/__tests__/mocks.ts
chaosmirage: > Btw, are there any tests for mocked entities?

No.

I think that at the MVP stage, it is sufficient to test high-level logic (binding component logic).

Mocks are fragile. They will interfere with refactoring. By refactoring, I mean rethinking the functions and structure of the system. 

Due to the different possible directions of development of the project, I suppose that a lot can change.

After consumer usage feedback (at least at this point) and stabilization of functions (will it be only an observer, or also a manager, etc.), it will be covered with e2e-like tests.

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
chaosmirage: I set it manually, so the `null` means a concise decision, not an accidentally obtained value.

`null` semantic:

> The **`null`** value represents the intentional absence of any object value.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null

packages/devextreme/js/__internal/core/state_manager/setup_state_manager.ts
chaosmirage: Thank you for your feedback.

I believe the current solution offers a good balance of simplicity and functionality for the MVP stage.

While I agree that the suggested approach would be more robust, I propose we add it to the backlog for future consideration.

packages/devextreme/js/__internal/core/state_manager/setup_state_manager.ts
chaosmirage: `NODE_ENV` is a common convention. 

We can track new dependencies on it (not only for dev mode) with a simple grep search.

Example: [https://github.com/search?q=repo%3Amui%2Fmaterial-ui NODE_ENV&type=code](https://github.com/search?q=repo%3Amui%2Fmaterial-ui%20NODE_ENV&type=code)

packages/devextreme/js/__internal/core/state_manager/observable_state_container_manager.ts
chaosmirage: At least in Preact Signals, it is called a container.

> The `signal` function creates a new signal. A signal is a **container** for a value that can change over time.

https://github.com/preactjs/signals/blob/main/README.md#signalinitialvalue

Why do you think a wrapper or adapter is better? Do you have any examples?

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
chaosmirage: 1. `areEqual`, for example, also doesn’t have state management-specific logic; it is a universal utility, but it is needed for the main functionality.
2. I look at the state manager like a tool that may have its own development lifecycle and be placed as a separate package. Why should it depend on something dx-specific?
It has deps from `import type { DIContext } from '../di';`, but it is a quick solution to avoid duplicate type at this moment 
3. What are the use cases of `TrackableCollection` outside of the state manager? Do we have them?

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
chaosmirage: As I mentioned before, it is an unstable feature, so I added a warning message.

[`The state travel feature is unstable and may cause unexpected behavior.`](https://github.com/DevExpress/DevExtreme/blob/8ec43dd1251ed00e46849a34708d7526f09f1526/packages/devextreme/js/__internal/core/state_manager/state_manager.ts#L226-L230)

My goal is to collect feedback about the necessity of this feature.

If consumers try it and see that it would be great to have, I will improve it; if not, I will remove it.

packages/devextreme/js/__internal/core/state_manager/__tests__/utils.test.ts
IlyaKhD: > What is the reason for avoiding the common convention of placing tests in the __tests__ folder?

Well, I'd say there are two common conventions: `__tests__` directories and `*.test.ts` files — and we're choosing between them.

> Why do we have different approaches using different folders?

This happened because almost all these folders were moved here from other repositories (we used to work without monorepo for a long time)


AFAICS, in the `__internal` folder the `*.test.ts`  convention is more widely used, that's why I think it should be used here.

packages/devextreme/js/__internal/core/state_manager/interfaces.ts
IlyaKhD: 👌🏻

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
IlyaKhD: Consider another analogy, which may be more relevant in this case, because we're speaking about a getter:

- [Array.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) returns `undefined` if no value found
- [Map.get](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/get) returns `undefined` as well

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
IlyaKhD: The `getAllControllers` method breaks the abstraction — that’s why the question arises.
It's about `ControllerRegistry` responsibilities

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
IlyaKhD: > the null means a concise decision, not an accidentally obtained value.

Since `undefined` is not a valid value for controller, there is no real difference for consuming code

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
IlyaKhD: > areEqual, for example, also doesn’t have state management-specific logic; it is a universal utility, but it is needed for the main functionality.

True. That's why I think it should be moved outside state management scope as well (may be not now) or better used as a 3rd party

> I look at the state manager like a tool that may have its own development lifecycle and be placed as a separate package. Why should it depend on something dx-specific?

It makes the domain and the code simpler. I see no good reason for adding one more entity which is just an enhanced object having no state-management related logic
It's the old good Occam's razor 🪒

> What are the use cases of TrackableCollection outside of the state manager? Do we have them?

No we don't. This is intended to reduce unnecessary complexity in the nascent State Management sub-system

packages/devextreme/js/__internal/core/state_manager/observable_state_container_manager.ts
IlyaKhD: > At least in Preact Signals, it is called a container.

For me it's not a `container` in signals, it's just that the word "container" is (mis)used in it's description.

In my experience, a container typically refers to something larger, more complex, and non-scala.
For example, a collection or structure that holds multiple elements as described here: https://en.wikipedia.org/wiki/Container_(abstract_data_type)#:~:text=In%20computer%20science%2C%20a%20container,objects%20(elements)%20it%20contains)

packages/devextreme/js/__internal/core/state_manager/__tests__/mocks.ts
IlyaKhD: I have contradictory thoughts about it. Need some time to figure things out.

packages/devextreme/js/__internal/core/state_manager/interfaces.ts
chaosmirage: Changed

packages/devextreme/js/__internal/core/state_manager/interfaces.ts
chaosmirage: Changed.

I noted that it is preferable to distinguish types and interfaces.

I thought differently because of the eslint rule:

```
Use an `interface` instead of a `type`.eslint[@typescript-eslint/consistent-type-definitions](https://typescript-eslint.io/rules/consistent-type-definitions)
```

packages/devextreme/js/__internal/core/state_manager/interfaces.ts
chaosmirage: changed

packages/devextreme/js/__internal/core/state_manager/logger.ts
chaosmirage: changed

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
chaosmirage: changed

packages/devextreme/js/__internal/core/state_manager/setup_state_manager.ts
chaosmirage: For now, it is indeed not required.

Removed.

packages/devextreme/build/gulp/js-bundles.js
chaosmirage: > Maybe it'd be good to collect all state_manager paths in here:
packages/devextreme/build/gulp/state_manager/constants.js, to explicitly link all the state_manager-related parts together.

Extracted

packages/devextreme/build/gulp/state_manager/transpile_state_manager_prod.js
chaosmirage: Extracted

packages/devextreme/build/gulp/state_manager/transpile_state_manager_prod.js
chaosmirage: Extracted

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
chaosmirage: Good points, changed.

packages/devextreme/js/__internal/core/state_manager/observable_state_container_manager.ts
chaosmirage: > In other words, they store objects in an organized way that follows specific access rules.

> Containers may be classified as either single-value containers or associative containers.

https://en.wikipedia.org/wiki/Container_(abstract_data_type)

Maybe I am missing something, but I don't see the contradiction. 

Are you insisting, or is it an optional recommendation?

chaosmirage: Also, another example of using "container" in the same domain:
https://learn.microsoft.com/en-us/aspnet/core/blazor/state-management?view=aspnetcore-9.0&pivots=server#in-memory-state-container-service

packages/devextreme/js/__internal/core/state_manager/__tests__/utils.test.ts
chaosmirage: Noted. Conventions per folder are allowed (not a one convention for the repo). 
Will change.

packages/devextreme/js/__internal/core/state_manager/__tests__/mocks.ts
chaosmirage: Re-implemented tests

packages/devextreme/js/__internal/core/state_manager/__tests__/utils.test.ts
chaosmirage: Changed

packages/devextreme/js/__internal/core/state_manager/controller_registry.ts
chaosmirage: Removed the controller registry

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
chaosmirage: Removed the controller registry

packages/devextreme/js/__internal/core/di/index.ts
```diff
@@ -38,7 +42,20 @@ export class DIContext {
...
+      (currentInstance, decoratorFn) => decoratorFn(currentInstance),
+      instance,
+    );
+    this.instances.set(id, decoratedInstance);
+  }
+
+  public decorator<T>(decorator: DecoratorFunction<T>): void {
+    this.globalDecorators.push(decorator as DecoratorFunction);
+
+    Array.from(this.instances.entries()).forEach(([id, instance]) => {
```
pomahtri: decorate already existing entities is a risky thing, as they can already be resolved and saved into variable

packages/devextreme/build/gulp/state_manager/transpile_state_manager_prod.js
chaosmirage: changed

packages/devextreme/build/gulp/js-bundles.js
chaosmirage: Changed.

packages/devextreme/js/__internal/core/state_manager/utils.ts
chaosmirage: I removed the `areEqual` function for now. 
I noticed that, for now, we can rely on the responsibility of a value container (preact/signal will do it) to trigger changes only when the current value differs from the previous value.
I suggest we postpone searching for a library until we need it.

packages/devextreme/js/__internal/core/state_manager/utils.ts
chaosmirage: See the [comment](https://github.com/DevExpress/DevExtreme/pull/29697#discussion_r2158613519)

packages/devextreme/js/__internal/core/state_manager/utils.ts
chaosmirage: See the [comment](https://github.com/DevExpress/DevExtreme/pull/29697#discussion_r2158613519)

packages/devextreme/js/__internal/core/di/index.ts
chaosmirage: I agree, changed, thanks.

packages/devextreme/js/__internal/core/state_manager/logger.ts
IlyaKhD: > I don’t see it. Do you mean in some other repos?
Yep, other repos. We're moving toward using a single tool for each functionality, especially in scripts and DevOps

Maybe that's not the case here, this may be a specific logger

packages/devextreme/js/__internal/core/state_manager/observable_state_container_manager.ts
IlyaKhD: Fair enough, you have a point. The Blazor example is certainly relevant, though it feels a bit awkward to me — perhaps due to some bias

packages/devextreme/js/__internal/core/state_manager/redux_dev_tools_connector.ts
```diff
@@ -0,0 +1,132 @@
...
+export class ReduxDevToolsConnector implements StateManagementTypes.DevToolsConnector {
+  private devTools: StateManagementTypes.ReduxDevToolsInstance | null = null;
+
+  private isConnected = false;
+
+  private readonly logger: StateManagementTypes.Logger;
+
+  private readonly componentName: string;
+
+  private readonly externalActionEmitter: EventEmitter<Parameters<StateManagementTypes.DevToolsConnector['onExternalAction']>[0]>;
```
IlyaKhD: **Mnior**
How about this?
```suggestion
  private readonly externalActionEmitter: EventEmitter<StateManagementTypes.DevToolsExternalActionCallback>;
```

packages/devextreme/js/__internal/core/state_manager/production.ts
```diff
@@ -0,0 +1 @@
+export const setupStateManager = (): void => {};
```
IlyaKhD: How about renaming the file to smth like `index.prod.ts` ? So that it's clear which file it's meant to replace?

packages/devextreme/js/__internal/core/state_manager/state_manager_factory.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ReduxDevToolsConnector } from './redux_dev_tools_connector';
+import { StateManager } from './state_manager';
+import type * as StateManagementTypes from './types';
+
+export const makeStateManager = (
+  options: StateManagementTypes.StateManagerFactoryOptions,
+): StateManager => {
+  const logger = options.logger ?? new Logger({ logLevel: options.logLevel, prefix: '[StateManager]' });
+
+  const stateContainerManagers: StateManagementTypes.StateManagerConfig['valueContainerManagers'] = options.valueContainerManagers ?? [ObservableValueContainerManager];
```
IlyaKhD: Consider using the type directly
```suggestion
  const stateContainerManagers: ValueContainerManagerConstructor[] = options.valueContainerManagers ?? [ObservableValueContainerManager];
```

packages/devextreme/js/__internal/core/state_manager/setup_state_manager.ts
```diff
@@ -0,0 +1,80 @@
...
+    diContext,
+    componentName,
+    logLevel = 'warn',
+    controllerSign = CONTROLLER_SIGN,
+  } = options;
+  const logger = new Logger({ logLevel, prefix: '[StateManager]' });
+
+  const isDevelopmentMode = process.env.NODE_ENV === 'development';
+
+  if (isDevelopmentMode) {
```
IlyaKhD: **Minor**
Consider inverting the conditon to reduce nesting:
```suggestion
  if (!isDevelopmentMode) { return undefined; }
```

packages/devextreme/js/__internal/core/state_manager/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+  return statePath.split(separator).filter(Boolean);
+}
+
+export function isValidStatePath(statePath: string): boolean {
+  const parts = splitStatePath(statePath);
+  return parts.length >= 2;
+}
+
+export function deepCopy<T extends object>(inputObject: T): T {
+  function iter(value: unknown, visited = new WeakMap<object, unknown>()): unknown {
```
IlyaKhD: **Minor**
From the `.set` calls it follows that the key is also `unknown`.
```suggestion
  function iter(value: unknown, visited = new WeakMap<unknown, unknown>()): unknown {
```

packages/devextreme/js/__internal/core/state_manager/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+      return regExpCopy;
+    }
+
+    if (Array.isArray(value)) {
+      const arrayCopy: unknown[] = [];
+
+      visited.set(value, arrayCopy);
+
+      value.forEach((item, index) => {
+        arrayCopy[index] = iter(item);
```
IlyaKhD: **Question**
Is it intentional to create a new `visited` map here?

```suggestion
        arrayCopy[index] = iter(item, visited);
```

IlyaKhD: If not, I suggest moving the map to a variable in the outer function scope:

```ts
const visited = new WeakMap<unknown, unknown>();

function iter(value: unknown): unknown { ... }
```

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
```diff
@@ -0,0 +1,147 @@
...
+  private readonly controllerSign: string;
+
+  private readonly valueContainerManagers: StateManagementTypes.ValueContainerManagerConstructor[];
+
+  constructor(
+    config: StateManagementTypes.StateManagerConfig,
+  ) {
+    const {
+      devToolsConnector, logger, valueContainerManagers, controllerSign,
+    } = config;
```
IlyaKhD: **Minor**
JFYI: destructuring breaks the "Find all references" VS Code functionality. For instance, if you try it on`valueContainerManagers`, the IDE will show the destructuring statement, not the assignment below (`this.valueContainerManagers = valueContainerManagers`)

So if some day you're interested in analyzing all the usages of a field (e.g. `valueContainerManagers`), you won’t be able to rely on a single "Find all referencese". Instead you'll have to repeat it for all the cases with destructuring. If you are not OK with that, you could avoid using destructuring for important types and interfaces: `this.valueContainerManagers = config.valueContainerManagers`

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
```diff
@@ -0,0 +1,147 @@
...
+    this.componentState = {};
```
IlyaKhD, line 18: Shouldn't it be a `WeakMap`?

packages/devextreme/js/__internal/core/state_manager/utils.ts
IlyaKhD: By the way, I'm not sure `WeakMap` is necessary here, since its lifetime is scoped to a single `deepCopy` call and all entries are added synchronously.

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
```diff
@@ -0,0 +1,147 @@
...
+      this.logger.error('Controller cannot be null or undefined');
+      return;
+    }
+
+    if (this.componentState[controlledId]) {
+      this.logger.debug(`Controller with ID '${controlledId}' is already tracked. Overwriting.`);
+    }
+
+    Object.entries(controller).forEach(([controllerPropertyName, controllerPropertyValue]) => {
+      if (this.findValueContainerManagerFor(controllerPropertyValue)) {
```
IlyaKhD: This creates a `ValueContainerManager` instance that is not used. Consider adding `hasValueContainerManagerFor` that returns `bool`

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
```diff
@@ -0,0 +1,147 @@
...
+    this.devToolsConnector.onExternalAction((action) => {
+      this.logger.warn(`Handler for the '${action}' action is not implemented`);
+    });
+
+    this.devToolsConnector.connect();
+
+    this.logger.info('StateManager initialized');
+  }
+
+  trackControllerState(controlledId: string, controller: StateManagementTypes.Controller): void {
```
IlyaKhD: **Minor**
To me, the naming is a bit confusing, since controller is actually a collection that's being iterated over using `Object.entries`. It would be easier to read the code if a pluralized or collection-like name was used.

From the outer scope it's a controller, but this function treats as an `entries` collection. Maybe it shouldn't know anything about controllers.

packages/devextreme/js/__internal/core/state_manager/types.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export type ValueContainerManagerConstructor = (
+  new (
+    logger: Logger,
+    controllerSign: string,
+    valueContainer: MaybeValueContainer
+  ) => ValueContainerManager
+) & {
+  canHandle: (valueContainer: MaybeValueContainer) => valueContainer is ValueContainer;
+};
```
IlyaKhD: This would be better as a factory-like structure rather than a class with static methods. If you try 'Find All References' on `canHandle`, the IDE won’t show `ObservableValueContainerManager` because it doesn’t explicitly implement the interface. This is not good for scalability.

```ts
export interface ValueContainerManagerConstructor = {
  canHandle(valueContainer: MaybeValueContainer): valueContainer is ValueContainer;
  create(
    logger: Logger,
    controllerSign: string,
    valueContainer: MaybeValueContainer
  ): ValueContainerManager;
};
```

packages/devextreme/js/__internal/core/state_manager/state_manager.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  beforeEach(() => {
+    jest.resetModules();
+    process.env = { ...originalEnv, NODE_ENV: 'development' };
+  });
+
+  afterEach(() => {
+    process.env = originalEnv;
+  });
+
+  it('Tracks a component state', () => {
```
IlyaKhD: Is it possible to group the assertions into a few smaller tests that share the same setup? That way, it’ll be much easier to debug test faliures

packages/devextreme/js/__internal/core/state_manager/utils.ts
chaosmirage: I don't get it, `.set` calls have narrowed key type.

Also, if I change to `unknown`, I will get `Type 'unknown' does not satisfy the constraint 'object'.`

`interface WeakMap<K extends object, V>`
typescript@4.9.5/node_modules/typescript/lib/lib.es2015.collection.d.ts

Could you please clarify the details?

chaosmirage: Changed to `Map` after https://github.com/DevExpress/DevExtreme/pull/29697#discussion_r2163391357, and now it is ok

packages/devextreme/js/__internal/core/state_manager/redux_dev_tools_connector.ts
chaosmirage: changed

packages/devextreme/js/__internal/core/state_manager/production.ts
chaosmirage: changed

packages/devextreme/js/__internal/core/state_manager/state_manager_factory.ts
chaosmirage: changed

packages/devextreme/js/__internal/core/state_manager/setup_state_manager.ts
chaosmirage: changed

packages/devextreme/js/__internal/core/state_manager/utils.ts
chaosmirage: > Is it intentional to create a new visited map here?

No, fixed, and added a test as well.

> If not, I suggest moving the map to a variable in the outer function scope

Removed the default value to do not forget to pass it. 
In this case, I prefer an explicit parameter.

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
chaosmirage: Changed

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
chaosmirage: No, because:

1.  I need to iterate through a collection, and `WeakMap` cannot iterate.
2.  I do not see an obvious scenario when this may cause memory leaks. If you see something interesting, please share it.
As I expected, a developer works with a component and does not dispose of it in the development environment.
3.  Anyway, just in case I implemented an approach with `WeakRef`, please check it https://github.com/DevExpress/DevExtreme/pull/29697/commits/ebc281dcc025f06d8d38473b52d433863a357f26

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
chaosmirage: Changed

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
chaosmirage: Changed

packages/devextreme/js/__internal/core/state_manager/types.ts
chaosmirage: Changed.

packages/devextreme/js/__internal/core/state_manager/state_manager.test.ts
chaosmirage: Changed.

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
```diff
@@ -0,0 +1,165 @@
...
+            payload: { ...valueContainerChange.payload, path: fullPathToProperty },
+          };
+
+          const currentComponentState = this.getComponentState();
+
+          this.devToolsConnector
+            .sendAction(
+              valueContainerChange.actionType,
+              componentStateChange.payload,
+              currentComponentState,
```
IlyaKhD, line 100: How is it different from `sourceData` that comes to the `trackStateOf` method?

I see how `sourceData` gets wrapped, becomes `this.componentState` and gets unwrapped, but I can't grasp how `this.componentState` can be modified.

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
IlyaKhD: Alright, I see..

Btw, I think it can be made a local `trackStateOf` variable, since there is no need to share its state between `trackStateOf` calls (as I got it)

in this case the `getComponentState` becomes kinda `unwrap` function:

```ts
getComponentState(componentState: StateManagementTypes.ComponentState): StateManagementTypes.ComponentState
```

chaosmirage: The `componentState` contains references to states from multiple different state sources (controllers).

`this.devTools.send(actionObject, currentState);`

The redux devtools expects the entire state for each `send` call.

This is why I need to 
>  iterate through a collection

`trackStateOf()` = "Register this state source (a controller) for tracking"
`getComponentState()` = "Return complete snapshot of ALL tracked sources"

---

```
// First trackStateOf call
stateManager.trackStateOf(columnsControllerInstance, 'ColumnsController');

// Second trackStateOf call  
stateManager.trackStateOf(dataControllerInstance, 'DataController');

// getComponentState() returns ALL tracked state in one object:
const completeState = stateManager.getComponentState();
console.log(completeState);
// Output:
{
  "ColumnsController": {
    "columnsCount": 5,
    "text": "initial"
  },
  "DataController": {
    "pagesCount": 10,
    "pagerConfig": {
      "infinityScrollingEnabled": true
    }
  }
}
```

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
chaosmirage: The redux devtools expects the entire state (all tracked state sources) for each `send` call to compare it with the previous entire state.

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
IlyaKhD: We discussed this on a call. The thing is, Redux DevTools require the state of all component controllers, so it has to be stored within the scope of an SM instance.

packages/devextreme/js/__internal/core/state_manager/state_manager.ts
IlyaKhD: See https://github.com/DevExpress/DevExtreme/pull/29697#discussion_r2174957679
The difference comes down to the scale: sourceData corresponds to a single controller, while the DevTools integration requires the state of all controllers, which is `this.componentState`.

IlyaKhD (approved):

PR #29713, opened 2025-04-25
FileUploader: validation should not consider accept option (T1287250)

Zedwag:


Zedwag:
decided to only aplly fix for 25.1 version

PR #29736, opened 2025-04-29
Scheduler: fix description of appointment with template (WL0wshmQ)

Ambrozy:


packages/devextreme/js/__internal/scheduler/appointments/appointment/m_appointment.ts
```diff
@@ -195,11 +195,12 @@ export class Appointment extends DOMComponent<AppointmentProperties> {
       .then((text) => {
         if (text) {
           const id = `dx-${new Guid()}`;
+          const description$ = $element.find(`.${APPOINTMENT_CONTENT_CLASSES.ARIA_DESCRIPTION}`);
```

If we use ```$``` to make obvious that this var contains jQuery element -> let's use it as a prefix, not a suffix
E.g. ```$description```

Because it sth like convention in jQuery, and ```$``` as suffix usually means reactivity (like ```observable$``` in rxjs)

wdevfx (dismissed):

wdevfx (approved):

PR #29737, opened 2025-04-29
LoadIndicator: Add _animationType option

marker-dao:


r-farkhutdinov (approved):

packages/devextreme/js/__internal/ui/m_load_indicator.ts
```diff
@@ -11,17 +11,31 @@ import Widget from '@ts/core/widget/widget';
...
+export const LOADINDICATOR_SEGMENT_CLASS = 'dx-loadindicator-segment';
+export const LOADINDICATOR_SEGMENT_INNER_CLASS = 'dx-loadindicator-segment-inner';
+export const LOADINDICATOR_IMAGE_CLASS = 'dx-loadindicator-image';
+
+export enum AnimationType {
+  Circle = 'circle',
+}
+
+export const ANIMATION_TYPE_CLASSES = {
+  [AnimationType.Circle]: 'dx-loadindicator-content-circle',
```
ksercs, line 27: Why do we need new class? As far I see we don't have styles for it

packages/devextreme/js/__internal/ui/m_load_indicator.ts
```diff
@@ -113,25 +137,47 @@ class LoadIndicator extends Widget<LoadIndicatorProperties> {
...
+      if (segmentInner) {
+        const $segmentInner = $('<div>').addClass(LOADINDICATOR_SEGMENT_INNER_CLASS);
 
-      if (animatingSegmentInner) {
-        $segment.append($('<div>').addClass(LOADINDICATOR_SEGMENT_INNER_CLASS));
+        $segment.append($segmentInner);
       }
 
-      this._$indicator.append($segment);
+      this._$indicator?.append($segment);
```
ksercs, line 180: Is optional chaining intentional here?

marker-dao: Yes, since TS in this case cannot determine that this._$indicator currently exists. We can either explicitly specify that this._$indicator already exists (but this is forbidden by ESLint), or we can add a check. In this case, I think adding a check is safer.

ksercs (approved):

PR #29742, opened 2025-04-30
🎴 CardView: filter sync - implement

Raushen:


packages/devextreme/js/__internal/grids/new/card_view/header_panel/header_panel.tsx
```diff
@@ -146,4 +151,16 @@ export class HeaderPanel extends Component<HeaderPanelProps> {
...
+  private isFiltered(column: VisibleColumn, syncFilter: unknown): boolean {
+    const { filterType, filterValues } = column;
+
+    let result = filterType === 'exclude'
+      || !!filterValues?.length;
+
+    if (!result && syncFilter) {
+      result = filterHasField(syncFilter, getColumnIdentifier(column));
+    }
+    return result;
```
wdevfx: Minor comment.
Let's also rename ```syncFilter``` -> ```filterSyncValue```

```ts
private isFiltered(column: VisibleColumn, syncFilterPanelValue: unknown): boolean {
  const { filterType, filterValues } = column;
  const hasHeaderFilter = filterType === 'exclude'
        || !!filterValues?.length;
  const columnIdentifier = getColumnIdentifier(column);

  return !hasHeaderFilter && syncFilter
    ? filterHasField(syncFilter, columnIdentifier)
    : hasHeaderFilter;
}
```

packages/devextreme/js/__internal/grids/new/card_view/header_panel/header_panel.tsx
```diff
@@ -57,6 +59,8 @@ export interface HeaderPanelProps {
     columnIndex?: number,
     onMenuCloseCallback?: () => void,
   ) => void;
+
+  syncFilterPanelValue?: unknown;
```
wdevfx: Question [minor about naming].

Maybe it will be better and more obvious to use the ```filterSyncValue``` here?
Because as far as I understand in the future it may be not only value from the filter panel

packages/devextreme/js/__internal/grids/new/card_view/header_panel/item.tsx
```diff
@@ -49,6 +49,7 @@ export interface ItemProps {
   isDragging?: boolean;
   template?: ComponentType<{ column: Column }>;
   cssClass?: string;
+  isFiltered?: boolean;
```
wdevfx: Minor comment.

```isFiltered``` -> ```hasFilters```

File: packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/utils.ts [hunk 4]
```diff
@@ -53,11 +53,16 @@ export const getFilterOperator = (values: unknown, filterType?: FilterType): str
   }
 };
 
+const isFilteringAllowed = (column: Column): boolean => column.allowFiltering
+|| column.allowHeaderFiltering;
+
+export const isColumnFilterable = (column: Column): boolean => isFilteringAllowed(column)
+&& !!column.dataField;
```
wdevfx, line 61: Major comment.

I'm not sure about this condition, because "computed" columns don't have a ```dataField```
By "computed" I mean columns that not represented in the DataSource:
```
columns: [
  { caption: 'computed', name: 'computed', calculateFieldValue: (date) => ... },
]
```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_controller.ts
```diff
@@ -14,29 +15,57 @@ export class FilterController {
...
+
   public readonly filterValueOption = this.options.twoWay('filterValue');
 
+  public readonly filterBuilderPopupOptions = this.options.oneWay('filterBuilderPopup');
+
+  public readonly filterPanelOptions = this.options.twoWay('filterPanel');
+
+  public readonly filterBuilderOptions = this.options.twoWay('filterBuilder');
+
+  public readonly filterSyncEnabledOption = this.options.oneWay('filterSyncEnabled');
```
wdevfx, line 24: As far as I can see, we already have a subscription to all of these options
We have a convention that we subscribe to an option only once in some specific controller (or view)

Therefore, we should inject the controller with desired option subscription to use this option in other places

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_controller.ts
```diff
@@ -14,29 +15,57 @@ export class FilterController {
...
   ] as const;
 
+  public readonly filterSyncEnabled = computed(
+    () => {
+      if (this.filterSyncEnabledOption.value === 'auto') {
+        return !!this.filterPanelVisible.value;
+      }
+      return !!this.filterSyncEnabledOption.value;
+    },
+  );
```
wdevfx: Minor comment.

```ts
public readonly filterSyncEnabled = computed(
    () => this.filterSyncEnabledOption.value === 'auto'
      ? !!this.filterPanelVisible.value
      : !!this.filterSyncEnabledOption.value
  );
```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_controller.ts
```diff
@@ -14,29 +15,57 @@ export class FilterController {
...
+
   public readonly customOperations = computed(
     () => [
-      anyOf(null),
-      noneOf(null),
+      anyOf(this.gridGetter),
+      noneOf(this.gridGetter),
```
wdevfx: Major comment.

Let's not provide the whole CardView widget mock for legacy code
I guess we can pass to these functions an object with the properties they need

As far as I can see, after quick research, the usages is:
- grid.option('headerFilter')
- grid.columnOption()
- grid.getController('headerFilter')

[Link to the custom operations source code](https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_custom_operations.ts)

We also already use this approach in the header filter feature

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_controller.ts
```diff
@@ -51,6 +80,10 @@ export class FilterController {
     private readonly columnsController: ColumnsController,
   ) { }
 
+  public widgetMockGetter = (): WidgetMock | null => null;
+
+  private readonly gridGetter = (): unknown => this.widgetMockGetter();
```
wdevfx: See the comment about custom operations above
Let's try to avoid usage and injecting whole ```WidgetMock``` here

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/controller.ts
```diff
@@ -0,0 +1,113 @@
...
+          }
+
+          const filterConditions = getMatchedConditions(filter, getColumnIdentifier(col as Column));
+          const filterType = getFilterType(filterConditions);
+          // eslint-disable-next-line @stylistic/max-len
+          // eslint-disable-next-line @typescript-eslint/init-declarations, @typescript-eslint/no-explicit-any
+          let filterValues: any[] | undefined;
+          if (filterType) {
+            filterValues = getFilterValues(filterConditions);
+          }
```
wdevfx: Let's change this line and remove ```eslint-disable``` comments

```ts
const filterValues = filterType
  ? getFilterValues(filterConditions)
  : undefined;
```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/controller.ts
```diff
@@ -0,0 +1,113 @@
...
+  private previousComposedHeaderFilter: unknown[] | null = null;
+
+  constructor(
+    private readonly options: OptionsController,
+    private readonly columnsController: ColumnsController,
+    private readonly filterController: FilterController,
+    private readonly headerFilterController: HeaderFilterController,
+  ) {
+    // Sync from FilterPanel to HeaderFilter
+    this.filterController.filterPanelValue.subscribe((filter) => {
```
wdevfx: Question [minor]

Can we use ```effect```here instead of ```subscribe```?
Because this is side-effect and I guess effect more suitable for it (but it's super minor)

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/types.ts
```diff
@@ -0,0 +1,5 @@
+import type { Mode } from '@js/common';
+
+export interface FilterSyncOptions {
+  filterSyncEnabled?: boolean | Mode;
+}
```
wdevfx: Question.

Do we need this interface duplicate?
Can we just use the ```Options``` instead (it's the same interface)?

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/controller.ts
```diff
@@ -0,0 +1,113 @@
...
+
+export class FilterSyncController {
+  public static dependencies = [
+    OptionsController,
+    ColumnsController,
+    FilterController,
+    HeaderFilterController,
+  ] as const;
+
+  private isFirstLoad = true;
```
wdevfx, line 27: Question.

Can we get rid of this ```firstLoad``` flag?
Why is the behavior on the first load different?
As far as I understand, we may have a configuration when we will do just one load at all

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/controller.ts
```diff
@@ -0,0 +1,113 @@
...
+        this.isFirstLoad = false;
+        if (this.filterController.filterPanelValue.value || filter.length === 0) {
+          this.previousComposedHeaderFilter = filter;
+          return;
+        }
+      }
+      if (!this.filterController.filterSyncEnabled.value) {
+        this.previousComposedHeaderFilter = filter;
+        return;
+      }
```
wdevfx: These conditions look complex
It's a little bit hard to understand what's going on here

Can we try to simplify this part of code?

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (!filterCondition) {
+    return undefined;
+  }
+
+  const value = filterCondition[2];
+  const hasArrayValue = Array.isArray(value);
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+  return hasArrayValue ? value : [value];
+};
```
wdevfx: Question.

Are u sure about this filter values extraction method?
As far as I know, the filter may be deeper and more complex
E.g
```
[["OrderNumber",">",100],"and",[["Employee","contains","A"],"or",["SaleAmount","<",1000]]]
```
243E8CF0-EF92-4B3C-AD7F-6AB750F84A1D

packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/utils.ts
```diff
@@ -66,10 +71,15 @@ export const getComposedHeaderFilter = (columns: Column[]): unknown[] => {
     .filter((col) => needCreateHeaderFilter(col));
   filterableColumns
     .forEach((col, index) => {
+      const { filterValues } = col;
+      let normalizedFilterValues = filterValues;
+      if (col.filterValues?.length === 1) {
+        [normalizedFilterValues] = filterValues;
+      }
```
wdevfx: Minor comment.

```ts
const normalizedFilterValues = col.filterValues?.length === 1
  ? filterValues[0]
  : filterValues
```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/view_controller.ts
```diff
@@ -66,34 +68,37 @@ export class HeaderFilterViewController {
...
-              const index = getColumnIndexByName(columns, column.name);
-              const newColumns = [...columns];
-
-              newColumns[index] = {
-                ...newColumns[index],
-                headerFilter: {
-                  ...newColumns[index].headerFilter,
+          if (apply) {
+            apply.call(this);
+          } else {
```
wdevfx: Major comment.

As far as I understand, here we call directly ```apply``` in the real CardView optionManager
Before, the legacy header filter part was isolated from the CardView
(This approach allow us to refactor headerFilter in future and avoid unexpected option updates)

Let's not break this isolation in the header filter and always use the ```else``` condition here :)

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_controller.ts
wdevfx: I missed the point, this subscriptions was moved from other places

packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/view_controller.ts
wdevfx: Decided to rename passed ```apply``` from arguments to the ```filterBuilderApply```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/controller.ts
wdevfx: Sth like that (it's ok to make duplicates here):
```ts
     if (this.isFirstLoad) {
        this.handleFirstLoadFilterSync();
        return;
      } 
      
      this.handleFilterSync();
```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/controller.ts
wdevfx: Yup, we need it

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/utils.ts
wdevfx: Let's comment on top of this util func that it not used with complex filters in filter sync

e2e/testcafe-devextreme/tests/cardView/filterSync/api.functional.ts
```diff
@@ -0,0 +1,362 @@
...
+    { id: 4, gender: undefined },
+    { id: 5, gender: undefined },
+  ],
+};
+
+const expectFilterElementsState = async (t, filteredColumns: string[], active = true) => {
+  const cardView = new CardView('#container');
+
+  // TODO: move to POM, when classes are correct for cardview
+  const filterPanelClearBtn = cardView.element.find('.dx-datagrid-filter-panel-clear-filter');
```
wdevfx: [We can do it in a separate PR]
Let's move it to the ```POM``` now (with ```TODO``` comment) and correct selector classes later

e2e/testcafe-devextreme/tests/cardView/filterSync/api.functional.ts
```diff
@@ -0,0 +1,362 @@
+import CardView from 'devextreme-testcafe-models/cardView';
```
wdevfx, line 1: Big like for functional tests 🙌

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/types.ts
wdevfx: This interface duplicate still here :)

packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/utils.ts
wdevfx: Postponed to fixing computed columns issues

packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/view_controller.ts
```diff
@@ -39,6 +39,8 @@ export class HeaderFilterViewController {
     element: Element,
     column: Column,
     onFilterCloseCallback?: () => void,
+    apply?: () => void,
```
wdevfx: Was discussed in a call, looks like we forgot to leave comment here during previous review, sorry

Let's rename this ```apply``` somehow, now it's shadows the ```apply``` in options below
It's not obvious that this ```apply``` argument is just a custom apply logic with filter sync

So, I suggest to ```apply``` -> ```customApplyAction``` or sth like this

Also, please take a look on the ```customApplyAction``` execution context below

packages/devextreme/js/__internal/grids/new/grid_core/filtering/legacy_filter_custom_operations.ts
```diff
@@ -0,0 +1,177 @@
...
+      };
+
+      column.filterType = 'include';
+      column.filterValues = conditionInfo.value ? conditionInfo.value.slice() : [];
+
+      headerFilterController.showHeaderFilterMenuBase({
+        columnElement: div,
+        column,
+        apply() {
+          setValue(this.filterValues);
```
wdevfx: It's very unclear, that ```this.filterValues``` will be in context of the ```header filter``` legacy code

Let's pass ```filterValues``` as an argument to the ```apply```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/view_controller.ts
```diff
@@ -75,25 +78,29 @@ export class HeaderFilterViewController {
...
-              const index = getColumnIndexByName(columns, column.name);
-              const newColumns = [...columns];
-
-              newColumns[index] = {
-                ...newColumns[index],
-                headerFilter: {
-                  ...newColumns[index].headerFilter,
+          if (apply) {
+            apply.call(this);
+          } else {
```
wdevfx: Minor comment.

Let's reduce complexity here:
```ts
if (customApplyAction) {
  // See comment below about customApplyAction context
  customApplyAction(this.filterValues);
}
```

packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_sync/types.ts
Tucchhaa: Forgot to the delete this file 😵

wdevfx (approved):

PR #29752, opened 2025-04-30
CardView - calculateCellValue should return dates for date-like type columns

tongsonbarbs:


wdevfx:
Overal -> lgtm
But I found one incorrect case

If we have a configuration like this:
```
dataSource: [
  { id: '0', date: new Date('2024-01-01') },
  { id: '1', date: new Date('2024-02-02') },
],
columns: [
  {
    dataField: 'id',
    dataType: 'number',
  },
  {
    dataField: 'date',
    dataType: 'datetime',
  }
],
```

We will see by default ```shortDate``` format, not the ```shortDateShortTime``` (that's [by default](https://github.com/DevExpress/DevExtreme/pull/29752/files#diff-9422b3bd0a74c1aaea1a330577f14d35f704c434b870ad73e95dbbd884b0fc4cR77) for the ```datetime``` type)

@tongsonbarbs, could u please check it and cover this case by test?

wdevfx:
@tongsonbarbs, in terms of this PR, could u please remove the ```calculateFieldValue``` WA from tests and storybook stories?
Because we fix the source of issue and don't need these WA usages anymore

In e2e tests one place:
- [headerFilter visual](https://github.com/DevExpress/DevExtreme/blob/25_1/e2e/testcafe-devextreme/tests/cardView/headerFilter/visual.ts#L104)

In storybook two places in [this file](https://github.com/DevExpress/DevExtreme/blob/25_1/apps/react-storybook/stories/card_view/CardView.stories.tsx)
(U can find them by searching ```calculateFieldValue``` and remove it for the ```dataType: 'date'``` columns)

packages/devextreme/js/__internal/grids/new/grid_core/columns_controller/utils.ts
```diff
@@ -26,7 +26,8 @@ function normalizeColumn(
     ?? columnFromDataOptions?.dataType
     ?? defaultColumnProperties.dataType;
   const columnDataTypeDefaultOptions = defaultColumnPropertiesByDataType[dataType];
-  const columnFormat = column.format ?? columnFromDataOptions?.format;
+  const columnFormat = column.format
+    ?? columnDataTypeDefaultOptions?.format;
```
wdevfx: Let's fall back to the calculated from data format ifthe  default one isn't specified 

Or as extended solution we can specify default formats for all dataTypes and get rid of the format calculated from data (but this is optional enhancement)

```suggestion
  const columnFormat = column.format
    ?? columnDataTypeDefaultOptions?.format
    ?? columnFromDataOptions?.format;
```

wdevfx (approved):

PR #29758, opened 2025-05-02
DataGrid - Validation message gets cut off in Fluent and Material themes (T1285387)

tongsonbarbs:


e2e/testcafe-devextreme/tests/dataGrid/common/validation/cellEditing.ts
```diff
@@ -49,3 +53,45 @@ fixture.disablePageReloads`Validation`
...
   }));
 });
+[
+  Themes.genericLight,
+  Themes.genericDark,
+  Themes.materialBlue,
+  Themes.materialBlueDark,
+  Themes.fluentBlue,
+  Themes.fluentBlueDark,
+].forEach((theme) => {
```
wdevfx: Minor comment.

I don't think it necessary to test ```light``` & ```dark``` theme variants
Just three main themes will be enough here:
```ts
[
  Themes.genericLight,
  Themes.materialBlue,
  Themes.fluentBlue,
].forEach((theme) => {
```

packages/devextreme-scss/scss/widgets/base/_validation.scss
```diff
@@ -65,6 +65,7 @@
     font-size: 0.85em;
     line-height: normal;
     word-wrap: break-word;
+    white-space: normal;
```
wdevfx: The validation's CSS styles are ok

We set the ```white-space``` in the ```DataGrid``` itself by [this class](https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme-scss/scss/widgets/base/_gridBase.scss#L265) in material & generic themes

So, it's not ok to override the default ```white-space``` and set it directly in the validation styles :)

Let's try to find out why we use the provided above ```dx-datagrid-nowrap```
And apply it directly when we need this ```nowrap``` or make an exception for overlays / validation messages in our CSS code

wdevfx (approved):

wdevfx (dismissed):

wdevfx (approved):

PR #29765, opened 2025-05-03
Chat:  add animation on edited text show/hide

Zedwag:


packages/devextreme/js/__internal/ui/chat/messagegroup.ts
```diff
@@ -236,6 +264,10 @@ class MessageGroup extends Widget<Properties> {
     }
   }
 
+  _isAlignmentStart(alignment: string): boolean {
+    return alignment === 'start';
```
marker-dao, line 271: Please also move const { alignment } = this.option(); line to this method: you will reduce the number of copy paste

packages/devextreme/js/__internal/ui/chat/messagegroup.ts
```diff
@@ -216,18 +218,44 @@ class MessageGroup extends Widget<Properties> {
...
+        $editedElement.remove();
+      }, { once: true });
+
+      $editedElement.addClass(CHAT_MESSAGE_EDITED_HIDING_CLASS);
+    };
 
     if ($message.is($firstMessage)) {
       const items = this.option('items');
-      this._renderMessageGroupInformation(items[0]);
+      const $information = this.$element().find(`.${CHAT_MESSAGEGROUP_INFORMATION_CLASS}`);
```
EugeniyKiyashko, line 232: If I add new message and new group is created then after edit of the message 'Edited' status is not added (First message of group)

EugeniyKiyashko: Because message doesn't store the updated isEdited value when you render

packages/devextreme/js/__internal/ui/chat/messagegroup.ts
Zedwag: good point

packages/devextreme/js/__internal/ui/chat/messagegroup.ts
Zedwag: added arg to make sure we set up correct edited state for information element
https://github.com/DevExpress/DevExtreme/pull/29765/commits/e5fc333773465733a5903baebba18c90386edb08

EugeniyKiyashko (approved):

PR #29771, opened 2025-05-05
DataGrid - enable grouping.contextMenuEnabled by default. Add new icons

Alyar666:


packages/devextreme/js/__internal/grids/data_grid/grouping/m_grouping.ts
```diff
@@ -289,15 +294,15 @@ const GroupingDataControllerExtender = (Base: ModuleType<DataController>) => cla
 
   private collapseAll(groupIndex) {
     const dataSource = this._dataSource;
-    if (dataSource && dataSource.collapseAll(groupIndex)) {
+    if (dataSource?.collapseAll(groupIndex)) {
```
tongsonbarbs: I believe these are eslint auto fixes? I think it would be better to revert them..

tongsonbarbs (approved):

PR #29777, opened 2025-05-06
DataGrid - NVDA reads all status messages on page load when multiple components are rendered (T1288423)

tongsonbarbs:


packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -191,13 +191,18 @@ export class ResizingController extends modules.ViewController {
...
-    const columnCount = this._columnsController?._columns?.filter(({ visible }) => !!visible).length ?? 0;
-    const totalItemsCount = Math.max(0, this._dataController.totalItemsCount());
-    const widgetAriaLabel = this._getWidgetAriaLabel();
-    const widgetStatusText = messageLocalization
-    // @ts-expect-error Badly typed format method
-      .format(widgetAriaLabel, totalItemsCount, columnCount);
+  private _setAriaLabel(e?: any): void {
+    let widgetStatusText = '';
+
+    if (!e?.operationTypes?.fullReload) {
```
Alyar666: I'm not sure if you need to use this flag since it will be true when editing and saving data.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -191,13 +191,18 @@ export class ResizingController extends modules.ViewController {
...
+      const totalItemsCount = Math.max(0, this._dataController.totalItemsCount());
+      const widgetAriaLabel = this._getWidgetAriaLabel();
+      widgetStatusText = messageLocalization
+      // @ts-expect-error Badly typed format method
+        .format(widgetAriaLabel, totalItemsCount, columnCount);
+    }
+
     const $ariaLabelElement = this.component.$element().children(`.${GRIDBASE_CONTAINER_CLASS}`);
     // @ts-expect-error Treelist Variable
     const expandableWidgetAriaLabel = messageLocalization.format(this._expandableWidgetAriaId);
```
Alyar666: This is a message for TreeList. So the original user problem will remain for TreeList?

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -1218,6 +1218,7 @@ export class DataController extends DataHelperMixin(modules.Controller) {
   public updateItems(change?, isDataChanged?) {
     change = change || {};
     const that = this;
+    change.isDataValueChanged = that.changed._fired;
```
Alyar666: I recommend using a public fired method. I also suggest renaming the flag to isFirstRender/isFirstChanged:
`change.isFirstRender = !that.changed.fired();`

Alyar666 (changes requested):

Alyar666 (approved):

PR #29780, opened 2025-05-06
Chat: implement Message Editing demo

EugeniyKiyashko:


apps/demos/Demos/Chat/MessageEditing/jQuery/index.js
```diff
@@ -0,0 +1,85 @@
...
+    },
+  });
+
+  $('#allow-deleting').dxSelectBox({
+    items: editingOptions,
+    value: editingOptions[0],
+    displayExpr: 'text',
+    inputAttr: { 'aria-label': 'Allow Deleting' },
+    onValueChanged(data) {
+      chat.option('editing.allowUpdating', data.value);
```
pharret31: allowDeleting

apps/demos/Demos/Chat/MessageEditing/jQuery/index.js
```diff
@@ -0,0 +1,85 @@
...
+        data: {
+          id: new DevExpress.data.Guid(),
+          ...message,
+        },
+      }]);
+    },
+    onMessageDeleted(e) {
+      dataSource.store().push([{
+        type: 'update',
+        key: e.message.id,
```
pharret31: I think it's better to use one approach everywhere, and the one in onMessageEntered where you get message var before working with it looks prettier imho:
```
const { message } = e;
...
```

apps/demos/Demos/Chat/MessageEditing/jQuery/data.js
```diff
@@ -0,0 +1,68 @@
...
+  { text: 'Enabled', value: true },
+  { text: 'Disabled', value: false },
+  {
+    text: 'Only the last message (custom)',
+    value: ({ component, message }) => {
+      const { items, user } = component.option();
+
+      const lastMessage = [...items]
+        .reverse()
+        .find((m) => m.author?.id === user.id && !m.isDeleted);
```
pharret31: Do you want to keep ` && !m.isDeleted`? It means user will be able to edit all messages one by one starting from the last one if they're deleting them in such order

pharret31: E.g. it could be done this way:

```
const lastMessage = [...items]
    .reverse()
    .find((m) => m.author?.id === user.id);

return !lastMessage.isDeleted && lastMessage?.id === message.id;
```

apps/demos/Demos/Chat/MessageEditing/jQuery/index.js
EugeniyKiyashko: fixed

apps/demos/Demos/Chat/MessageEditing/jQuery/index.js
EugeniyKiyashko: Sure, great suggestion!

apps/demos/Demos/Chat/MessageEditing/jQuery/data.js
EugeniyKiyashko: I thought about this and implemented it for demonstration purposes. Otherwise, after a message is deleted, it creates a dead end, and you'd have to change the settings just to see the menu again. I see it as a way to show the menu for the last non-deleted message.

I don't have any objections to choose your option

EugeniyKiyashko: Fixed

pharret31 (approved):

apps/demos/Demos/Chat/MessageEditing/jQuery/data.js
```diff
@@ -0,0 +1,71 @@
...
+  custom: ({ component, message }) => {
+    const { items, user } = component.option();
+
+    const lastMessage = [...items]
+      .reverse()
+      .find((m) => m.author?.id === user.id);
+
+    return !lastMessage.isDeleted && lastMessage?.id === message.id;
+  },
+};
```
ksercs, line 65: 1. I don't think it's data, it's a business logic. In Angular it'll be located in component.ts not in service, right?
2. Last suitable message should be always possible to be edited, even after currently last message remove or adding of a new one. Now it does now work if you remove last message. It was previously discussed with PMs - we here demonstrate a real scenario how some messengers work
3. Currently this code takes O(n) time on every bubble click, actually real time is 3N for every click. I don't think i'd like to implement it this way in my app. I suggest you better keep only `lastEditableMessageId ` and update it on init/renderMessage with O(1) and on last message remove with O(n) (rare scenario) [but anyway it's better to go from last, not destructureize+reverse+find, to improve perf more]

EugeniyKiyashko: 1. right
2. https://github.com/DevExpress/DevExtreme/pull/29780#discussion_r2075728684. I don't remember this decision, thanks for your attention, so will return initial solution back
3. It may do the logic more difficult. Well, will improve for performance purposes

PR #29783, opened 2025-05-06
ButtonGroup/AIDialog: Prevent enter keydown event && Prevent update dialog state on hide

marker-dao:


packages/devextreme/js/__internal/ui/html_editor/ui/m_baseDialog.ts
```diff
@@ -58,23 +58,25 @@ abstract class BaseDialog<T = unknown> {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   show(options?: unknown): Promise<T> | undefined {
     if (this._popup.option('visible')) {
-      return;
+      return undefined;
```
r-farkhutdinov, line 61: Isn't it the same as it was?

r-farkhutdinov (approved):

packages/devextreme/js/__internal/ui/html_editor/ui/m_baseDialog.ts
marker-dao: Not quite, you specified a return type of `Promise<T> | undefined`, but returned `void` in this line.

r-farkhutdinov: Why there was no TS errors in this case?

marker-dao: You can get the ts-error in the `return this.deferred.promise();` line if replace `return undefined` with `return`.

PR #29782, opened 2025-05-06
DataGrid: fix Ctrl+Arrow movement when in the same page more than 1 DataGrid (T1285596)

haksur:

PR #29786, opened 2025-05-06
DataGrid: fix Ctrl+Arrow movement when in the same page more than 1 DataGrid (T1285596) (#29782)

haksur:
…


haksur:
https://github.com/DevExpress/DevExtreme/pull/29782

wdevfx (approved):

PR #29801, opened 2025-05-07
DataGrid: fix height when a row is marked deleted (T1286265)

haksur:


wdevfx:
Checked locally

Initially, this issue reproduces only with fixed columns
So, without fixed columns, all works as expected, and the rows' height stays the same

Unfortunately, current solution mirrors the initial issue 
-> now without fixed columns, the deleted row has 1px less height than expected :( 

See the attached screencast:

https://github.com/user-attachments/assets/f19599f0-b414-4d5a-a98f-ecd87e3c65bd

wdevfx:
Possible fix suggestion

The issue occurs because, for fixed columns, we change the table style ```border-collapse``` ([link to the docs](https://developer.mozilla.org/en-US/docs/Web/CSS/border-collapse)) from ```collapse``` to ```separate```

We do this because of some issues with  borders with ```position:sticky``` style in fixed columns
U can try to manually return ```border-collapse: separate``` in [this demo](https://js.devexpress.com/jQuery/Demos/WidgetsGallery/Demo/DataGrid/FixedAndStickyColumns/MaterialBlueLight/) and see what's wrong with it

Also, reducing the height of the row causes the double borders in deleted rows
649D1339-94D2-4812-A7E4-73448F7045E7_4_5005_c

Without fixed columns, this border is thinner:
FDA67197-D01A-4CBA-A00E-60F761CA2CD9_4_5005_c

Therefore, as a possible solution, after a quick local check, I suggest:

Let's remove the ```border-top``` from the deleted row's cell if the option [showRowLines](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/#showRowLines) is enabled and we have fixed columns (```.dx-datagrid-sticky-columns```)

See attached screencast:

https://github.com/user-attachments/assets/032bf7b8-312d-4038-807a-78370ba9d5ab

But I'm not 100% sure, this suggested solution should be checked in all major themes (```default``` / ```compact``` variants) and with the following options:
- showRowLines
- with/without fixed columns

wdevfx:
About failed CI demos' screenshots checks

0142799A-C5F9-498F-8F62-B1D1D75FFEDB_4_5005_c

If we fix this issue -> it's expected that in some demos (when we take a screenshot that contains a deleted row)
screenshots may be different (because the deleted row height is reduced)

E.g a diff in the ```datagrid_batch_update_request``` demo test
datagrid_batch_update_request_1_desktop_diff

Let's review the failed demos screenshots and replace the affected ones after the fix

**IMPORTANT NOTE:**
In the CI, we have some retry count in testcafe tests (around 10 retries for each)
Unfortunately, the failed screenshots are added for each retry
In other words, if some test fails the first time but passes the second -> in the screenshot archive we found a failed screenshot of the first attempt
**Therefore, we should not just replace all etalons, but review the CI logs and replace etalons only for failed tests**

e2e/testcafe-devextreme/tests/dataGrid/common/markup/T1286265_deletedRowHeight.ts
```diff
@@ -0,0 +1,83 @@
...
+        keyExpr: 'id',
+        height: 300,
+        showBorders: true,
+        columnFixing: { enabled: true },
+        columns: [
+          {
+            dataField: 'id', caption: 'ID', width: 50, fixed: true,
+          },
+          { dataField: 'name', caption: 'Full Name', width: 150 },
+        ],
```
wdevfx: Minor comment.

Let's remove redundant options from columns to simplify component configuration in the e2e test
E.g we don't need a ```caption``` here :)

e2e/testcafe-devextreme/tests/dataGrid/common/markup/T1286265_deletedRowHeight.ts
```diff
@@ -0,0 +1,83 @@
...
+import url from '../../../../helpers/getPageUrl';
+import { changeTheme } from '../../../../helpers/changeTheme';
+import { Themes } from '../../../../helpers/themes';
+
+fixture.disablePageReloads`DataGrid deleted row height consistency T1286265`
+  .page(url(__dirname, '../../../container.html'));
+
+const DATA_GRID_SELECTOR = '#container';
+// check first and second rows, because both can have issue independently each other
+const ROW_INDEXES = [1];
```
wdevfx: Let's remove this ```ROW_INDEXES``` because it contains now just one value

```suggestion
```

e2e/testcafe-devextreme/tests/dataGrid/common/markup/T1286265_deletedRowHeight.ts
```diff
@@ -0,0 +1,83 @@
...
+[
+  Themes.fluentBlue,
+  Themes.fluentBlueCompact,
+  Themes.materialBlue,
+  Themes.materialBlueCompact,
+  Themes.genericLight,
+  Themes.genericLightCompact,
+].forEach((theme) => {
+  ROW_INDEXES.forEach((rowIndex) => {
+    test(`Row height should not change when marked as deleted - ${theme} - row - ${rowIndex}`, async (t) => {
```
wdevfx: Also, let's remove the rowIndex from the test name -> it's redundant now

```suggestion
    test(`Row height should not change when marked as deleted - ${theme}`, async (t) => {
```

e2e/testcafe-devextreme/tests/dataGrid/common/markup/T1286265_deletedRowHeight.ts
```diff
@@ -0,0 +1,83 @@
...
+          { id: 2, name: 'Jane Johnson' },
+          { id: 3, name: 'Mike Wilson' },
+        ],
+        keyExpr: 'id',
+        height: 300,
+        showBorders: true,
+        columnFixing: { enabled: true },
+        columns: [
+          {
+            dataField: 'id', width: 50, fixed: true,
```
wdevfx: Let's write tests with ```DataGrid``` without fixed columns to be sure that in all cases everything works as expected

e2e/testcafe-devextreme/tests/dataGrid/common/markup/T1286265_deletedRowHeight.ts
```diff
@@ -0,0 +1,83 @@
...
+      const deletedRow = dataGrid.getDataRow(rowIndex);
+      const deletedRowHeight = await deletedRow.element.clientHeight;
+
+      // Assert - check if the height remains consistent
+      await t
+        .expect(deletedRowHeight)
+        .eql(initialRowHeight, 'Row height should not change when marked as deleted');
+
+      // Take a screenshot for visual verification
+      await takeScreenshot(`datagrid-deleted-row-height (${theme}) row-${rowIndex}.png`, dataGrid.element);
```
wdevfx:
```suggestion
      await takeScreenshot(`datagrid-deleted-row-height (${theme}).png`, dataGrid.element);
```

packages/devextreme-scss/scss/widgets/fluent/gridBase/_index.scss
```diff
@@ -1142,7 +1141,7 @@ $fluent-grid-base-group-panel-message-line-height: $fluent-button-text-line-heig
     }
 
     &.dx-#{$widget-name}-sticky-columns {
```
wdevfx, line 1143: Question.

Will it be enough to remove ```border-top``` here?
Sth like:
```scss
&.dx-#{$widget-name}-sticky-columns {
  .dx-row-removed > td {
    border-top: none;
  }
}
```

haksur: I think not enough

wdevfx (approved):

PR #29826, opened 2025-05-08
Tests: The utility to help skip tests on unsupported devices was added

pharret31:


packages/devextreme/testing/tests/DevExpress.animation/position.tests.js
```diff
@@ -993,8 +997,11 @@ const testCollision = (name, fixtureName, params, expectedHorzDist, expectedVert
     });
 
     QUnit.test('position should return window.width() if window.outerWidth == window.innerWidth (T939748)', function(assert) {
-        const isPhone = devices.real().deviceType === 'phone';
-        if(isPhone || browser.safari) {
+        if(isRealDeviceTypeNotIn(['desktop', 'tablet'], assert)) {
```
Ambrozy: are we sure that new types will not appear? also it looks unreadable, I would add a separate function `isRealDeviceTypeIn`

packages/devextreme/testing/helpers/device.js
```diff
@@ -0,0 +1,10 @@
+import devices from '__internal/core/m_devices';
+
+export const isRealDeviceTypeNotIn = (deviceTypes, assert = null, message = null) => {
+    const allowedDeviceTypes = Array.isArray(deviceTypes) ? deviceTypes : [deviceTypes];
+    if(!allowedDeviceTypes.includes(devices.real().deviceType)) {
+        assert && assert.ok(true, message || 'Test should be skipped on devices not in list:  ' + deviceTypes.toString());
```
ksercs: It's strange for me to see that we assert here. Your function is named as a predicate => it should basically just return true/false. I believe asserts should be nowhere besides test itself, just for readability and awareness of this special "if" corner cases

packages/devextreme/testing/tests/DevExpress.animation/position.tests.js
ksercs: Agree with Ambrozy. Actually the most readable option to me is to have `isDesktop`, `isPhone` and `isTablet` utils, with no unification

pharret31: New types could appear, that's why I thought it would be better to explicitly specify which devices this test should work on

packages/devextreme/testing/helpers/device.js
pharret31: This test utility is for skipping test cases and it preventing code duplications, so in all the scenarios where this utility should be used it has to do an assert with a message, that's why it was extracted to a specific function

pharret31: Renamed to `shouldSkipTestForDevice`

packages/devextreme/testing/helpers/device.js
```diff
@@ -0,0 +1,12 @@
+import devices from '__internal/core/m_devices';
+
+export const shouldSkipTestForDevice = (allowedDeviceTypes, assert = null, message = null) => {
```
ksercs: `shouldSkipTestForDevice(['desktop'])` = "should I skip test for desktop", but you assert when it's not desktop ¯\_(ツ)_/¯

pharret31: the variable name is allowedDeviceTypes, but function name means skip for a current device

ksercs: >variable name is allowedDeviceTypes

Yeah, function body is well readable.
However, it's call (usage in test) is ambiguous

pharret31: have put all passing params to an object to avoid ambiguity

packages/devextreme/testing/helpers/device.js
```diff
@@ -0,0 +1,16 @@
...
+} = {}) => {
+    if(!Array.isArray(allowedDevices)) {
+        throw new Error('allowedDevices should be an array');
+    }
+    if(!allowedDevices.includes(devices.real().deviceType)) {
+        assert && assert.ok(true, message || 'Test should be skipped on devices not in list:  ' + allowedDevices.toString());
+        return true;
+    }
+    return false;
+};
```
Ambrozy, line 16: The purpose of this function is to skip or pass test according to device type.
Better to not use negatives and say everything directly.
We can use `QUnit.test.if` https://qunitjs.com/api/QUnit/test.if/
```
QUnit.test.if('render', isDesktop() || isTablet(), function (assert) {
  assert.strictEqual(MyApp.render(), '<p>Hello world!</p>');
});
```

packages/devextreme/testing/tests/DevExpress.animation/position.tests.js
```diff
@@ -964,8 +965,11 @@ const testCollision = (name, fixtureName, params, expectedHorzDist, expectedVert
     });
 
     QUnit.test('position should return window.height() if window.outerHeight == window.innerHeight (T939748)', function(assert) {
-        const isPhone = devices.real().deviceType === 'phone';
-        if(isPhone || browser.safari) {
+        if(shouldSkipOnDevice({ allowedDevices: ['desktop', 'tablet'], assert })) {
```
Ambrozy: isDevice(['desktop'], assert)
isNotDevice(['isPhone'])

packages/devextreme/testing/helpers/device.js
```diff
@@ -0,0 +1,16 @@
+import devices from '__internal/core/m_devices';
+
+export const shouldSkipOnDevice = ({
```
ksercs: let's use plural to emphasize array should be passed
```suggestion
export const shouldSkipOnDevices = ({
```

packages/devextreme/testing/helpers/device.js
ksercs: I also think it's better to use negatives because I don't want to remember all existing devices when I want to skip just on phone. I think signature gonna be "{  devices, assert, message } "

ksercs: regarding qunit.test.if - not sure, let's not do it now at least because there will be conflicts with existing wraps like testInActiveWindow

ksercs (approved):

Ambrozy (approved):

PR #29833, opened 2025-05-12
HtmlEditor: Improve aria-attributes support in AIDialog and ai toolbar item

marker-dao:


vladaskorohodova (dismissed):

r-farkhutdinov (dismissed):

packages/devextreme/js/__internal/ui/html_editor/ui/aiDialog.ts
```diff
@@ -719,6 +724,16 @@ export default class AIDialog extends BaseDialog<AIDialogResult> {
     return visible as boolean;
   }
 
+  private _refreshDialogAria(): void {
+    const id = String(new Guid());
+    // @ts-expect-error $overlayContent does not exist on public type
+    const $overlayContent = this._popup.$overlayContent();
```
ksercs: There is a guide in out public docs how to do it without private API, please check

packages/devextreme/js/__internal/ui/html_editor/modules/m_toolbar.ts
```diff
@@ -481,10 +483,15 @@ if (Quill) {
 
       const options = {
         dataSource,
-        onItemClick: (e): void => {
+        disabled: isMenuDisabled,
+        onContentReady: (e: ContentReadyEvent): void => {
+          const $item = $(e.element).find(`.${DX_MENU_ITEM_CLASS}`).first();
+          $item.attr('aria-label', localizationMessage.format('dxHtmlEditor-aiToolbarItem'));
+        },
```
ksercs, line 490: Is it okay that you add a label for a root item only? Items in menu also have `menuitem` role, should they also have a label?

packages/devextreme/js/__internal/ui/html_editor/ui/aiDialog.ts
```diff
@@ -244,6 +246,7 @@ export default class AIDialog extends BaseDialog<AIDialogResult> {
     };
 
     const options = {
+      inputAttr: { 'aria-label': localizationMessage.format('dxHtmlEditor-aiResultTextArea') },
```
ksercs: If we do it for result textArea shouldn't it also be done for askAi textArea?

packages/devextreme/js/localization/messages/fi.json
```diff
@@ -542,6 +542,8 @@
         "dxHtmlEditor-aiTryAgain": "Try again",
         "dxHtmlEditor-aiGenerate": "Generate",
         "dxHtmlEditor-aiStop": "Stop",
+        "dxHtmlEditor-aiToolbarItem": "AI Assistant toolbar item",
+        "dxHtmlEditor-aiResultTextArea": "AI Assistant result",
```
ksercs: afair we usually explicitly specify `label` in the variable name. `aiResultTextArea` looks strange for me. Please discuss with TW

packages/devextreme/js/__internal/ui/html_editor/ui/aiDialog.ts
```diff
@@ -480,6 +484,7 @@ export default class AIDialog extends BaseDialog<AIDialogResult> {
     this._refreshToolbarItems();
     this._refreshLoadIndicator();
     this._refreshInformer();
+    this._refreshDialogAria();
```
ksercs, line 487: Does title depend on state? 
If not, isn't it better to do it single time in popup's `onContentReady`?

packages/devextreme/js/__internal/ui/html_editor/ui/aiDialog.ts
marker-dao: No, because Prompt TextArea has no empty placeholder (placeholder is like aria-label for SR, Axe).

packages/devextreme/js/localization/messages/fi.json
marker-dao: Okay let's use the label postfix.

packages/devextreme/js/__internal/ui/html_editor/ui/aiDialog.ts
marker-dao: No, we can't.

Now that I've looked deeper into the code, I can see that when the state changes, there is a constant update of ToolbarItems: https://github.com/DevExpress/DevExtreme/pull/29833/files#diff-72ca008cf68f852ae0f88cd7cb34d2c59825c068ad1cb7e2c332746cd1f98a86R491 

This causes the Toolbar Items to re-render every time the state changes, which means that the header that was added at the beginning - no longer exists.

In good conscience, we shouldn't design a component this way. But the current redraws have almost no impact on performance.

packages/devextreme/js/localization/messages/fi.json
marker-dao: [a3c6b9246ec499246359f7183cf07e938bb6db6d]

packages/devextreme/js/__internal/ui/html_editor/ui/aiDialog.ts
marker-dao: **content** (dx-popup-content element) and **overlayContent** (dx-overlay-content element) – are not equal.

packages/devextreme/js/__internal/ui/html_editor/modules/m_toolbar.ts
marker-dao: Yes, that's okay. All other elements don't need aria-label, because they have text in them.

packages/devextreme/js/__internal/ui/html_editor/ui/aiDialog.ts
ksercs: Did someone say that it's equal?

ksercs: `$overlayContent = $(e.component.content()).parent()`
https://js.devexpress.com/jQuery/Documentation/Guide/UI_Components/Popup/Accessibility/#Screen_Reader_Support

marker-dao: [77b8b464c86180d4cd32fbb7e26059eb6288691b]

ksercs (approved):

PR #29864, opened 2025-05-14
TagBox - Methods specified in DevExpress.ui.dxSelectBox.defaultOptions are called (T1286533)

tongsonbarbs:


e2e/testcafe-devextreme/tests/dataGrid/common/tagBox.ts
```diff
@@ -67,3 +69,31 @@ fixture.disablePageReloads`Tagbox Columns`.page(
       });
   },
 );
+
+test('TagBox - Methods specified in DevExpress.ui.dxSelectBox.defaultOptions are called (T1286533)', async (t) => {
```
pomahtri: Let's move this file, it tests only tagBox component, it shouldn't be located inside dataGrid folder
Ideally, there should be test for abstract component somewhere inside qunit tests. We already have them, please check

packages/devextreme/js/__internal/core/widget/dom_component.ts
```diff
@@ -142,9 +142,18 @@ class DOMComponent<
 
   // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
   _setOptionsByDevice(instanceCustomRules): void {
-    // @ts-expect-error
-    // eslint-disable-next-line @stylistic/max-len
-    super._setOptionsByDevice([].concat(this.constructor._classCustomRules || [], instanceCustomRules || []));
+    const ctor = this.constructor;
+
+    const hasOwnCustomRules = Object.prototype.hasOwnProperty.call(ctor, '_classCustomRules');
+    const hasOwnDefaultOptions = Object.prototype.hasOwnProperty.call(ctor, 'defaultOptions');
```
pomahtri, line 148: Why we need to check for having own `defaultOptions`?
When can we or customers overwrite it?

packages/devextreme/js/__internal/core/widget/dom_component.ts
```diff
@@ -142,9 +142,18 @@ class DOMComponent<
...
+
+    const ownClassCustomRules = hasOwnCustomRules
+      || hasOwnDefaultOptions
+      // @ts-expect-error
+      ? ctor._classCustomRules
+      : [];
+
+    super._setOptionsByDevice([].concat(ownClassCustomRules || [], instanceCustomRules || []));
```
pomahtri, line 156: Please check how was the behavior before refactoring for native classes: what was in case when we have classes A -> B, and have some default options in A. Were they inherited by B, or this defaultOptions works exactly only for the specified component, not for its descendants?

pomahtri: Also please verify that all such cases are tested or please write tests for them

tongsonbarbs: Hi, I was able to confirm this during my research.

Previously, before we migrated to native ES6 classes, declaring a static `defaultOptions` method on a specific component would only assign `_classCustomRules` to that component itself and not to its descendants. However, after the migration to native classes, calling `defaultOptions` on a subclass (e.g., Class B) now also affects its superclass (e.g., Class A) due to how static properties are inherited and shared in JavaScript.

Specifically, when calling `defaultOptions` in a subclass, the shared static `_classCustomRules` array (inherited from the base class) is used unless explicitly overridden. As a result, default options intended for one class can unintentionally propagate to others in the inheritance chain.

The basic workaround is to explicitly assign `_classCustomRules` to each component to break the inheritance chain. However, this approach is not scalable, as it would require manual initialization across many components.

The proposed solution is to modify the logic that assigns `_classCustomRules`, ensuring that we only initialize or use `_classCustomRules` if it is an _own property_ of the current class (i.e., using `Object.prototype.hasOwnProperty.call(this, '_classCustomRules')`). This prevents shared mutation of inherited static properties.

However, there's a corner case when a class is not directly derived from the expected base class, for example, `dxPopup → PopupFull → Popup`. In [this declaration](https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/ui/popup/m_popup.full.ts#L22), PopupFull defines its own static `defaultOptions` method and manually pulls `_classCustomRules` from its base class. This bypasses the normal inheritance behavior and must be accounted for separately.

pomahtri (approved):

PR #29865, opened 2025-05-14
fix(template manager): Add mounted ref to prevent state modification

ajivanyandev:
Fixes https://github.com/orgs/DevExpress/projects/75/views/9?pane=issue&itemId=110298505&issue=DevExpress%7Cdevextreme-private%7C170

Add mounted ref to the TemplateManager to prevent state modification while unmounted.


VasilyStrelyaev (commented):
Let's try to add a test as we discussed

ajivanyandev:
> Let's try to add a test as we discussed

As the case we are covering is React 17 specific, decided to go without tests.

PR #29868, opened 2025-05-14
CardView: Fix pager after filter was reset

Tucchhaa:


packages/devextreme/js/__internal/grids/new/grid_core/filtering/filter_controller.ts
```diff
@@ -43,7 +43,7 @@ export class FilterController {
   public readonly displayFilter = computed(
     () => gridCoreUtils.combineFilters(
       this.appliedFilterExpressions.value,
-    ),
+    ) || null,
```
pomahtri:
```suggestion
    ) ?? null,
```

pomahtri (approved):

Raushen (approved):

PR #29873, opened 2025-05-14
CardView: fix selection (throw error if needed) (oqSw9T3c)

haksur:

### DIFF
--- a/packages/devextreme/js/__internal/grids/new/grid_core/options_validation/controller.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/options_validation/controller.ts
@@ -1,6 +1,7 @@
 import { isDefined } from '@js/core/utils/type';
-import errors from '@js/ui/widget/ui.errors';
 import { DataController } from '@ts/grids/new/grid_core/data_controller/index';
+
+import { throwError } from './utils';
 
 export class OptionsValidationController {
   public static dependencies = [
@@ -16,7 +17,7 @@ export class OptionsValidationController {
     const keyExpr = this.dataController.dataSource.peek().key();
 
     if (!isDefined(keyExpr)) {
-      throw errors.Error('E1042', 'CardView');
+      throwError('E1042', 'CardView');
     }
   }
 }
--- a/packages/devextreme/js/__internal/grids/new/grid_core/options_validation/controller.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/options_validation/controller.ts
@@ -1,6 +1,6 @@
 import { isDefined } from '@js/core/utils/type';
 import errors from '@js/ui/widget/ui.errors';
-import { DataController } from '@ts/grids/new/grid_core/data_controller';
+import { DataController } from '@ts/grids/new/grid_core/data_controller/index';
 
 export class OptionsValidationController {
   public static dependencies = [
--- a/packages/devextreme/js/__internal/grids/new/grid_core/options_validation/controller.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/options_validation/controller.ts
@@ -1,0 +1,22 @@
+import { isDefined } from '@js/core/utils/type';
+import errors from '@js/ui/widget/ui.errors';
+import { DataController } from '@ts/grids/new/grid_core/data_controller';
+
+export class OptionsValidationController {
+  public static dependencies = [
+    DataController,
+  ] as const;
+
+  constructor(
+    private readonly dataController: DataController,
+  ) {
+  }
+
+  public validateKeyExpr(): void {
+    const keyExpr = this.dataController.dataSource.peek().key();
+
+    if (!isDefined(keyExpr)) {
+      throw errors.Error('E1042', 'CardView');
+    }
+  }
+}
--- a/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
@@ -44,9 +44,11 @@ describe('when keyExpr is missing', () => {
     $(cardView ?? undefined as any)?.dxCardView('dispose');
     document.body.innerHTML = '';
   });
+  
   beforeEach(() => {
     jest.clearAllMocks();
   });
+  
   describe('selection mode single', () => {
     it('shouldn\'t throw E1042 error if keyExpr is missing and selection', () => {
       setup({
--- a/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
@@ -50,7 +50,7 @@ describe('when keyExpr is missing', () => {
   });
   
   describe('selection mode single', () => {
-    it('shouldn\'t throw E1042 error if keyExpr is missing and selection', () => {
+    it('shouldn\'t throw E1042 on initial startup', () => {
       setup({
         dataSource: [{ value: 'test1' }, { value: 'test2' }],
         selection: {
--- a/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
@@ -44,11 +44,11 @@ describe('when keyExpr is missing', () => {
     $(cardView ?? undefined as any)?.dxCardView('dispose');
     document.body.innerHTML = '';
   });
-  
+
   beforeEach(() => {
     jest.clearAllMocks();
   });
-  
+
   describe('selection mode single', () => {
     it('shouldn\'t throw E1042 on initial startup', () => {
       setup({
--- a/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
@@ -103,7 +103,7 @@ describe('when keyExpr is missing', () => {
   });
 
   describe('selection mode multiple', () => {
-    it('shouldn\'t throw E1042 error if keyExpr is missing and selection', () => {
+    it('shouldn\'t throw E1042 on initial startup', () => {
       setup({
         dataSource: [{ value: 'test1' }, { value: 'test2' }],
         selection: {
--- a/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
@@ -1,0 +1,174 @@
+/* eslint-disable spellcheck/spell-checker, @stylistic/max-len */
+import {
+  afterEach, beforeEach, describe, expect, it, jest,
+} from '@jest/globals';
+import $ from '@js/core/renderer';
+import CardView from '@ts/grids/new/card_view/widget';
+import type { Options as GridCoreOptions } from '@ts/grids/new/grid_core/options';
+import { throwError } from '@ts/grids/new/grid_core/options_validation/utils';
+import { rerender } from 'inferno';
+
+const SELECTORS = {
+  cardView: '.dx-cardview',
+  card: '.dx-cardview-card',
+  cardCheckbox: '.dx-checkbox-container',
+  selectAllButton: '[aria-label="Select all"]',
+};
+
+const setup = (options: GridCoreOptions = {}): CardView => {
+  const container = document.createElement('div');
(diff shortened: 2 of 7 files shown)


packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts
```diff
@@ -210,10 +212,18 @@ export class SelectionController {
...
+
+      if (!isDefined(keyExpr)) {
+        throw errors.Error('E1042', 'keyExpr is missing');
+      }
+
       const onSelectionChanged = this.onSelectionChanged.peek();
       const eventArgs = this.getSelectionEventArgs(e);
 
       this.selectedCardKeys.value = [...e.selectedItemKeys];
+      this.itemsController.setSelectionState([...e.selectedItemKeys]);
```
wdevfx: Why do we need this line?

The ```itemsController.setSelectionState``` already called here inside the ```effect```:
https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts#L125

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts
```diff
@@ -210,10 +212,18 @@ export class SelectionController {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private selectionChanged(e: any): void {
     if (e.addedItemKeys.length || e.removedItemKeys.length) {
+      const keyExpr = this.dataController.dataSource.peek().key();
+
+      if (!isDefined(keyExpr)) {
+        throw errors.Error('E1042', 'keyExpr is missing');
+      }
```
wdevfx: The same code already exists in ur other PR:
https://github.com/DevExpress/DevExtreme/pull/29879/files#diff-7cb9f1a76c0baa212b6933a3d0b51772d253c5c7578e4e0c88733b153a3ea7e8R100

Let's move it to some common place, e.g -> ```new/grid_core/options_controller/...```
I see two options for how we can implement it:
- make it a util function that has ```dataSource``` as an argument
- make it as ```OptionValidator``` class and inject ```dataSource``` to this class

So, let's implement this change in one of these PRs (ideally, in the first one and reuse in the second one)

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts
```diff
@@ -210,10 +212,18 @@ export class SelectionController {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private selectionChanged(e: any): void {
     if (e.addedItemKeys.length || e.removedItemKeys.length) {
+      const keyExpr = this.dataController.dataSource.peek().key();
```
wdevfx: Unfortunately, looks like at least one case is missing now

If we create the CardView with the following configuration:
```
dataSource: [{ id: 0, A: 'A_0'}],
selection: { mode: 'multiple' },
```

And change the ```selectedCardKeys``` by API:
```
$('#card-view').dxCardView('instance').option('selectedCardKeys', [0])
```
The error will not be thrown :(

Please note that initial configuration works well (error throws):
```
dataSource: [{ id: 0, A: 'A_0'}],
selection: { mode: 'multiple' },
selectedCardKeys: [0],
```

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts
```diff
@@ -762,6 +762,172 @@ describe('SelectionController', () => {
       });
```
wdevfx, line 762: Let's also add integrations (jest) or e2e (testcafe) test coverage

We would like to check the following cases when ```keyExpr``` is missing at least in:
- selection: 'single' -> click on card
- selection: 'multiple' -> click on card or checkbox
- selection: 'multiple' -> 'selectAll' toolbar btn click
- selection 'single' -> selectedCardKeys set in options initially
- selection 'single' -> selectedCardKeys set in options in runtime (with dxCardView('instance').option('selectedCardKeys', ...) 
- selection 'multiple' -> selectedCardKeys set in options initially
- selection 'multiple' -> selectedCardKeys set in options in runtime (with dxCardView('instance').option('selectedCardKeys', ...)

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts
```diff
@@ -210,10 +212,18 @@ export class SelectionController {
...
+
+      if (!isDefined(keyExpr)) {
+        throw errors.Error('E1042', 'keyExpr is missing');
```
wdevfx: Let's change the error message, now it looks a little bit repetitive:
```Uncaught Error: E1042 - keyExpr is missing requires the key field to be specified.```

File: packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts (hunk 1 above)
Comment (haksur) [reply packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts]: maybe the line in the `effect` not works, because this fixing bug

Comment (haksur) [reply packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts]: Discussed in the call, I will remove this line

File: packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts (hunk 2 above)
Comment (haksur) [reply packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts]: done ✔️

File: packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts (hunk 3 above)
Comment (haksur) [reply packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts]: done ✔️

File: packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts (hunk 5 above)
Comment (haksur) [reply packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts]: done ✔️

File: packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts (hunk 4 above)
Comment (haksur) [reply packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts:762]: done ✔️

File: packages/devextreme/js/__internal/grids/new/grid_core/options_validation/controller.ts [hunk 6]
```diff
@@ -0,0 +1,22 @@
+import { isDefined } from '@js/core/utils/type';
+import errors from '@js/ui/widget/ui.errors';
+import { DataController } from '@ts/grids/new/grid_core/data_controller';
```
wdevfx: QUnit fails because we forget to set ```index``` here
(Unfortunately, QUnit cannot load ```index``` automatically without mention)

```suggestion
import { DataController } from '@ts/grids/new/grid_core/data_controller/index';
```

haksur: done

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts
```diff
@@ -26,6 +26,21 @@ const setup = (config: Options = {}) => {
   };
 };
 
+const checkKeyExprError = (
+  action: () => unknown,
+  selectionController?: SelectionController,
+): void => {
+  try {
+    action();
```
wdevfx: Now tests that uses the ```checkKeyExprError``` is always green
What happened if ```action``` executes without error? :)

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts
```diff
@@ -823,4 +1001,132 @@ describe('SelectionController', () => {
       });
     });
   });
+
+  describe('when keyExpr is missing', () => {
+    describe('selection mode single', () => {
+      it('should throw E1042 error if keyExpr is missing and selection ', () => {
```
wdevfx: This test case should not pass because we don't throw the error without touching the selection actions

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts
```diff
@@ -823,4 +1001,132 @@ describe('SelectionController', () => {
...
+        checkKeyExprError(() => setup({
+          dataSource: [{ value: 'test1' }, { value: 'test2' }],
+          selection: {
+            mode: 'single',
+          },
+        }));
+      });
+
+      it('should throw E1042 error on card click selection', () => {
```
wdevfx: Where is the real ```click``` in this test?
Let's write integration tests with the whole cardview and trigger ```click``` events (and other UI / public API) described in the comments about test cases

Example of how to create the ```CardView``` in jest u can find here:
[Example of CardView setup in jest](https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/options.integration.test.ts#L28)

I suggest creating sa eparate test file to avoid mixing up integration test cases with others in one file
E.g -> ```controller.integration.test.ts```

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts
haksur: done ✔️

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts
haksur: done ✔️

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts
haksur: done ✔️

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
```diff
@@ -0,0 +1,174 @@
...
+  afterEach(() => {
+    const cardView = document.querySelector(SELECTORS.cardView);
+    // @ts-expect-error bad typed renderer
+    $(cardView ?? undefined as any)?.dxCardView('dispose');
+    document.body.innerHTML = '';
+  });
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+  describe('selection mode single', () => {
```
wdevfx: Minor comment.

Let's add blank (empty) lines
```suggestion
  });
  
  beforeEach(() => {
    jest.clearAllMocks();
  });
  
  describe('selection mode single', () => {
```

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    it('shouldn\'t throw E1042 error if keyExpr is missing and selection', () => {
```
wdevfx: Let's rename this test case
Now it looks like unfinished thought/sentence (and selection what? :))

E.g. 
```suggestion
    it('shouldn\'t throw E1042 on initial startup', () => {
```

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.integration.test.ts
```diff
@@ -0,0 +1,174 @@
...
+      });
+
+      cardView.instance().option('selectedCardKeys', [1]);
+
+      checkError();
+    });
+  });
+
+  describe('selection mode multiple', () => {
+    it('shouldn\'t throw E1042 error if keyExpr is missing and selection', () => {
```
wdevfx: Let's rename this test case
Now it looks like unfinished thought/sentence (and selection what? :))

E.g. 
```suggestion
    it('shouldn\'t throw E1042 on initial startup', () => {
```

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.test.ts
```diff
@@ -762,6 +762,171 @@ describe('SelectionController', () => {
...
+          selectedItems: [cardData],
+        };
+
+        try {
+          // Call the private method directly
+          // @ts-expect-error - accessing private method
+          selectionController.selectionChanged(selectionChangedEvent);
+        } catch (error) {
+          expect((error as { __id: string }).__id).toEqual('E1042');
+        }
```
wdevfx: It's better to refactor in the following way:

```ts
expect(() => {
  // @ts-expect-error - accessing private method
  selectionController.selectionChanged(selectionChangedEvent);
}).toThrow(...)
```

Because if we will not throw error -> this test will be green (```catch``` will not be triggered)

haksur: done ✔️

wdevfx (approved):

PR #29877, opened 2025-05-15
CardView: dataController -> paging in the remote OData DataSource not working

pomahtri:
PR consist of two commits:

### fix: bug

Direct fix of the bug described in the card.

Problem: each time some loadOption is changed, we call effect to pass changed values to dataSource and call `.load` if we need it.
In case of filtering, we were passing the `displayFilter.value ?? null` to dataSource, but the displayField could also be `undefined`. In result, we passed each time the 'null' value, but we were comparing it with `undefined` each time.

Because of this, filter was set each time when any loadOption was changed, and there's internal dataSource logic to drop pageIndex after filtering is changed (since it could lead to changes in totalCount)

### fix: getCombinedFilter

I noticed that we didn't have proper logic for getCombinedFilter and fixed it

When we were syncing cardview's display filter to dataSource filter, we were just calling `dataSource.filter(displayFilter)`. However, dataGrid behaves in different way:

Filter, specified in DataSource, is never changed, and is used as 'initial top-level filter'. When we set filter by UI, displayFilter (called additionalFilter in DataGrid) was combined to filter in DataSource, not overwrite it.

I added combineFilter method, used it in `customizeStoreLoadOptions` callback and fixed public method `getCombinedFilter`


packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
```diff
@@ -0,0 +1,57 @@
...
+const setup = (options: Options) => {
+  const context = getContext(options);
+
+  return {
+    optionsController: context.get(OptionsControllerMock),
+    dataController: context.get(DataController),
+  };
+};
+
+describe('regressions', () => {
```
Tucchhaa: Can we rename the description to 'DataController' ? I think current name isn't suitable and doesn't follow the way other testing files have their descriptions.

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
```diff
@@ -211,23 +219,48 @@ export class DataController {
...
+        }
+
+        if (!equalByValue(
+          this.previousDisplayFilter ?? null,
+          displayFilter ?? null,
+          {
+            maxDepth: FILTER_OBJ_COMPARE_DEPTH,
+            strict: true,
+          },
+        )) {
```
Tucchhaa:
```suggestion
        const filterChanged = !equalByValue(
          this.previousDisplayFilter ?? null,
          displayFilter ?? null,
          {
            maxDepth: FILTER_OBJ_COMPARE_DEPTH,
            strict: true,
          },
        );
        
        if (filterChanged) {
```

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/public_methods.test.ts
```diff
@@ -171,4 +167,65 @@ describe('PublicMethods', () => {
...
+        },
+        filter: dataSourceFilter,
+      }),
+      columns: [
+        { dataField: 'a', filterValues: columnFilterValues },
+      ],
+    });
+
+    describe('when displayFilter and filter from dataSource are empty', () => {
+      it('should return empty filter', () => {
```
Tucchhaa, line 186: Maybe just?:

```suggestion
   it('should return empty filter when displayFilter and filter from dataSource are empty', () => {
```

I doubt that there will be any other tests in this describe block.

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/public_methods.test.ts
```diff
@@ -171,4 +167,65 @@ describe('PublicMethods', () => {
...
+      });
+    });
+
+    describe('when displayFilter and filter from dataSource are set', () => {
+      it('should combine filters', () => {
+        const { gridCore } = innerSetup({
+          dataSourceFilter: ['a', '=', 123],
+          columnFilterValues: [1, 2],
+        });
+        expect(gridCore.getCombinedFilter()).toMatchInlineSnapshot(`
```
Tucchhaa: can we use toEqual here too? Just to be consistent

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/public_methods.test.ts
```diff
@@ -171,4 +167,65 @@ describe('PublicMethods', () => {
...
+      });
+    });
+
+    describe('when displayFilter is set and filter from dataSource is empty', () => {
+      it('should return displayFilter filter', () => {
+        const { gridCore } = innerSetup({
+          dataSourceFilter: undefined,
+          columnFilterValues: [1, 2],
+        });
+        expect(gridCore.getCombinedFilter()).toMatchInlineSnapshot('undefined');
```
Tucchhaa: can we use toEqual/toBe here too? Is there any specific reason why toMatchInlineSnapshot was used?

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
pomahtri: I moved the describe into DataController
I want to have 'regressions' to be grouped because they not really good tests (they don't add anything into spec), they are just checks for regressions

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/public_methods.test.ts
pomahtri: It's general rule of https://bettertests.js.org/

Tucchhaa (approved):

Tucchhaa (approved):

PR #29878, opened 2025-05-15
CardView: Fix a state issue with the searchPanel.text property, prevent re-renderings (9GXIgWhP)

haksur:
…


packages/devextreme/js/__internal/grids/new/grid_core/search/view.tsx
```diff
@@ -25,28 +25,37 @@ export class SearchView {
...
+      (component) => {
+        this.searchTextBox = component;
+      },
+    );
     this.toolbarController.addDefaultItem(
-      computed(() => addSearchTextBox(
-        this.getProps().value,
-        (component) => { this.searchTextBox = component; },
-      )),
+      computed(() => toolbarItem),
```
pomahtri: minor
```suggestion
      signal(toolbarItem),
```

wdevfx (approved):

PR #29886, opened 2025-05-16
DashStyle: Extended with string type where it's used

pharret31:


packages/devextreme/js/common/charts.d.ts
```diff
@@ -639,7 +639,7 @@ export interface SeriesLabel {
        * @docid dxChartSeriesTypes.CommonSeries.label.border.dashStyle
        * @default 'solid'
        */
-      dashStyle?: DashStyle | undefined;
+      dashStyle?: DashStyle | string | undefined;
```
ksercs, line 642: It's it better to extend DashStyle type? :)

pharret31: Yep, that was my first attempt :) but there are some different issues with regenerating in all scenarios so I thought that was a restriction of generator and here is my second attempt, I think if the issue with it is resolved then I'll try to extend the original type again

PR #29891, opened 2025-05-16
Gauge: skip core rerender on title change with same height (T1288122)

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/viz/gauges/base_gauge.js
+++ b/packages/devextreme/js/viz/gauges/base_gauge.js
@@ -19,6 +19,8 @@ export const BaseGauge = BaseWidget.inherit({
 
     _themeSection: 'gauge',
 
+    _titleRectCache: null,
+
     _createThemeManager: function() {
         return new themeManagerModule.ThemeManager(this._getThemeManagerOptions());
     },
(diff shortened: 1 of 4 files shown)


packages/devextreme/js/viz/gauges/base_gauge.js
```diff
@@ -118,6 +120,30 @@ export const BaseGauge = BaseWidget.inherit({
...
+            const hasTitleHeightChanged = this._title._boundingRect.height !== this._titleRectCache?.height;
+            const hasTitleYChanged = this._title._boundingRect.y !== this._titleRectCache?.y;
+            const hasVerticalAlignmentChanged = this._title._boundingRect.verticalAlignment !== this._titleRectCache?.verticalAlignment;
+
+            this._titleRectCache = null;
+
+            return hasTitleHeightChanged || hasTitleYChanged || hasVerticalAlignmentChanged;
+        }
+
+        return true;
```
pharret31, line 147: Should we return false if there is no requireCoreUpdate and no title in changes list?

anna-shakhova: Previously we updated core every time _applySize was called, as if this function always returned `true`
Now we return `false` if changes triggered by TITLE only

Refactored a bit to make it more clear

pharret31 (approved):

ksercs (approved):

PR #29893, opened 2025-05-16
Scheduler: Refactor resource data accessor and grouping

Ambrozy:


wdevfx:
The [OnContentReady](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/Configuration/#onContentReady) behavior changed when we have a local resource ```data source```
Previously, because the local resources ```data source``` load was sync -> on the ```OnContentReady``` hook resources was already applied

Now it always work like remote ```data source``` (see attached video)
Let's try to wait the ```resources``` initial loading (if exists) before firing the ```OnContentReady``` event


https://github.com/user-attachments/assets/976583b5-d91f-4235-aa49-677792c68791

packages/devextreme/js/__internal/scheduler/utils/loader/resource_loader.ts
```diff
@@ -0,0 +1,47 @@
...
+  idsGetter: ResourceIdAccessor['idsGetter'];
+
+  idsSetter: ResourceIdAccessor['idsSetter'];
+
+  constructor(
+    config: ResourceConfig,
+    public dataAccessor = new ResourceDataAccessor(config),
+    public allowMultiple = Boolean(config.allowMultiple),
+    public useColorAsDefault = Boolean(config.useColorAsDefault),
+    public resourceIndex = String(getResourceIndex(config)),
```
wdevfx: Let's rename this field to ```resourceId``` because in runtime it's not an index

packages/devextreme/js/__internal/scheduler/appointments/data_provider/m_appointment_filter.ts
```diff
@@ -225,43 +228,13 @@ export class AppointmentFilterBaseStrategy {
...
-
-      for (let i = 0; i < appointmentResourceValues.length; i++) {
-        if (hasResourceValue(resourceData, appointmentResourceValues[i])) {
-          return true;
-        }
-      }
-
-      return false;
-    };
+  _filterAppointmentByResources(appointment, groupsResources: ResourceLoader[]) {
```
wdevfx, line 229: Now, the complexity of this method is:
```O(GroupCount * MaxResourceCount * MaxMultipleResources)```
And we call it for each appointment, so the final complexity will be
```O(AppointmentCount * MaxMultipleResources * GroupCount * MaxResourceCount)```

Where:
- AppointmentCount -> count of displayed appointments
- GroupCount -> group count from scheduler config
- ResourceCount -> maximal count of resources in some resource data source
- MaxMultipleResources -> maximal count of resources for some group

We cannot reduce the ```MaxMultipleResources``` / ```AppointmentCount``` and it's redundant optimization to reduce the ```GroupCount```
But the ```ResourceCount``` can be really high

So, let's add to ```ResourceLoader``` method that will check that the passed resource ids array intersects with resource items and use the hash map (Record or real HashMap) for it inside

After these changes, the complexity will be:
```O(AppointmentCount * MaxMultipleResources * GroupCount)```
(And initially O(GroupCount * MaxResourceCount), but only once after resource loading)

File: packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts [hunk 3]
```diff
@@ -183,7 +192,10 @@ const DEFAULT_WORKSPACE_RENDER_OPTIONS: RenderRWorkspaceOptions = {
   generateNewData: true,
 };
 
-class SchedulerWorkSpace extends WidgetObserver {
+class SchedulerWorkSpace extends WidgetObserver<Omit<dxSchedulerOptions, 'groups'> & {
+  groups: ResourceLoader[];
+  getResourceManager: () => ResourceManager;
+}> {
```
wdevfx: Let's move the ```WidgetObserver``` generic type to a separate type

```ts
type WorkspaceOptionsInternal = ...
class SchedulerWorkSpace extends WidgetObserver<WorkspaceOptionsInternal> {
```

packages/devextreme/js/__internal/scheduler/appointments/data_provider/m_appointment_filter.ts
```diff
@@ -37,6 +38,8 @@ export class AppointmentFilterBaseStrategy {
     this.dataAccessors = this.options.dataAccessors;
   }
 
+  get resourceManager(): ResourceManager { return this.options.getResourceManager(); }
```
wdevfx: Minor comment. 

I see only one usage of this getter in the ```m_appointment_filter_virtual```
Let's remove it and use ```options``` directly to make it more obvious

packages/devextreme/js/__internal/scheduler/appointments/m_settings_generator.ts
```diff
@@ -26,12 +29,14 @@ export class DateGeneratorBaseStrategy {
     this.options = options;
   }
 
+  get resourceManager(): ResourceManager { return this.options.getResourceManager(); }
```
wdevfx, line 32: Minor comment.

Let's get rid of this getter and use the ```options``` directly

e2e/testcafe-devextreme/tests/scheduler/common/api/resourceRequestCount.ts
```diff
@@ -0,0 +1,178 @@
...
+import { createWidget } from '../../../../helpers/createWidget';
+import { resourceApiMock } from '../../apiMocks/resourceApiMock';
+
+fixture.disablePageReloads`Scheduler API - request counting`
+  .page(url(__dirname, '../../../container.html'));
+
+const requestLogger = RequestLogger(/\/api\/data/);
+const updateOption = ClientFunction((name, value) => {
+  ($('#container') as any).dxScheduler('instance').option(name, value);
+});
```
wdevfx: Minor comment.

Let's move this function to the Scheduler POM, it may be helpful in other test cases
Sth like this: https://github.com/DevExpress/DevExtreme/blob/25_1/packages/testcafe-models/dataGrid/index.ts#L493

packages/devextreme/js/__internal/scheduler/appointments/appointment/m_appointment.ts
```diff
@@ -48,8 +48,8 @@ export class Appointment extends DOMComponent<AppointmentProperties> {
     return this.option('data');
   }
 
-  get resourceProcessor(): ResourceProcessor {
-    return this.option('getResourceProcessor')();
+  get resourceManager(): ResourceManager {
```

Let's remove this getter

packages/devextreme/js/__internal/scheduler/appointments/rendering_strategies/m_strategy_agenda.ts
```diff
@@ -29,19 +29,14 @@ class AgendaRenderingStrategy extends BaseRenderingStrategy {
...
-      config,
+    const grouped = groupAppointmentsByGroupLeafs(
+      this.resourceManager.resourceById,
+      this.resourceManager.groupsLeafs,
       appointments,
-      groups,
     );
+
+    // TODO(9): Get rid of it as soon as you can. Unnecessary copy of appointments
+    return grouped.map((group) => group.map((item) => ({ ...item })));
```
wdevfx, line 42: Question.

Can we get rid of it right now?
Because from comment it's not obvious why we cannot get rid of it
Why do we need this copy?

packages/devextreme/js/__internal/scheduler/appointments/rendering_strategies/m_strategy_base.ts
```diff
@@ -37,6 +38,8 @@ class BaseRenderingStrategy {
     this._initPositioningStrategy();
   }
 
+  get resourceManager(): ResourceManager { return this.options.getResourceManager(); }
```
wdevfx: Minor comment.

Let's remove this getter if possible

packages/devextreme/js/__internal/scheduler/utils/promise_to_deferred.ts
```diff
@@ -0,0 +1,17 @@
+import type { DeferredObj } from '@js/core/utils/deferred';
```
wdevfx: Let's use the ```Deferred.fromPromise``` instead of this util func
https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/core/utils/m_deferred.ts#L113

packages/devextreme/js/__internal/scheduler/r1/utils/base.ts
```diff
@@ -461,31 +447,39 @@ export const extendGroupItemsForGroupingByDate = (
...
-      })));
-    }
+  let groupPanelItems = groupResources
+    .map((group) => {
+      const result = [] as GroupRenderItem[];
+      const {
+        resourceName, resourceIndex, items, data,
+      } = group;
+
+      for (let iterator = 0; iterator < repeatCount; iterator += 1) {
```
wdevfx: Minor comment.

Let's rename ```iterator``` -> ```idx``` or other name that u prefer that highlight that it's an index

packages/devextreme/js/__internal/scheduler/utils/is_agenda_workpace_component.ts
```diff
@@ -0,0 +1,6 @@
+const workspaceAgendaComponentName = 'dxSchedulerAgenda';
```
wdevfx, line 1: As far as I can see we have only 1 usage, so let's inline this condition in place where it used

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -957,7 +955,12 @@ class Scheduler extends Widget<any> {
 
     const workspace = this.getWorkSpace();
 
-    if (!this._isAgenda() && this.filteredItems && workspace) {
+    if (
+      !this._isAgenda()
+      && this.filteredItems
+      && workspace
+      && !isAgendaWorkspaceComponent(workspace)
```
wdevfx, line 961: I guess that we can just inline it

packages/devextreme/js/__internal/scheduler/utils/data_accessor/appointment_data_accessor.ts
```diff
@@ -21,10 +21,6 @@ const isDateField = (field: string): boolean => field === 'startDate' || field =
 export class AppointmentDataAccessor extends DataAccessor<Appointment, KnownFields> {
```
wdevfx, line 21: [Can be done in next refactoring]

Let's create index.ts inside the ```utils/data_accessor``` and export only entities that used outside this folder

packages/devextreme/js/__internal/scheduler/utils/loader/loader.test.ts
```diff
@@ -0,0 +1,107 @@
+import {
```
wdevfx, line 1: [Can be done in next refactoring]

Let's create index.ts inside the ```utils/data_accessor``` and export only entities that used outside this folder

packages/devextreme/js/__internal/scheduler/utils/data_accessor/resource_data_accessor.ts
```diff
@@ -0,0 +1,33 @@
+import { compileGetter, compileSetter } from '@js/core/utils/data';
+
+import type { RawResourceData, ResourceConfig, ResourceData } from '../loader/types';
+import { DataAccessor } from './data_accessor';
+import type { DataAccessorGetter, DataAccessorSetter } from './types';
+
+export class ResourceDataAccessor extends DataAccessor<RawResourceData, ResourceData> {
+  constructor(
```
wdevfx, line 14: As far as I can see, we don't pass default arguments to the ctor
So, let's follow the ```YAGNI```practice and remove these arguments
E.g:
```ts
  constructor({ valueExpr, displayExpr, colorExpr }: ResourceConfig) {
    this.idExpr = valueExpr ?? 'id';
  }
```

Also, looks like if we would like to change ```idExpr```, the ```config``` argument is enough

wdevfx (dismissed):

packages/devextreme/js/__internal/scheduler/utils/loader/loader.ts
```diff
@@ -0,0 +1,98 @@
...
+    dataSourceOptions: object,
+    public readonly dataSource = normalizeDataSource<T>(
+      config.dataSource,
+      dataSourceOptions,
+    ),
+    public items: Data[] = [],
+    public data: T[] = [], // TODO(9): probably we dont need it. Used in getGroupPanelData
+    protected readonly isSharedDataSource = config.dataSource instanceof DataSource,
+    protected loadingStatePromise?: Promise<T[]>,
+    protected unsubscribe = noop,
```
wdevfx: Same comment for extra ctor params
Let's follow ```YAGNI``` and remove not used

We will add some later if needed

packages/devextreme/js/__internal/scheduler/utils/loader/resource_loader.ts
```diff
@@ -0,0 +1,46 @@
...
+} from './types';
+
+export class ResourceLoader extends Loader<RawResourceData, ResourceData> {
+  idsGetter: ResourceIdAccessor['idsGetter'];
+
+  idsSetter: ResourceIdAccessor['idsSetter'];
+
+  constructor(
+    config: ResourceConfig,
+    public dataAccessor = new ResourceDataAccessor(config),
```
wdevfx: Same comment for extra ctor arguments

packages/devextreme/js/__internal/scheduler/utils/loader/resource_loader.ts
```diff
@@ -0,0 +1,46 @@
...
+    return items.map((item) => ({
+      id: this.dataAccessor.get('id', item),
+      text: this.dataAccessor.get('text', item),
+      color: this.dataAccessor.get('color', item),
+    }));
+  }
+
+  protected onLoadError(): void {}
+
+  protected onChange(): void {}
```
wdevfx, line 57: Question.

Do we need these methods in the future?
Maybe we can remove them and add them later when they are needed?

packages/devextreme/js/__internal/scheduler/utils/loader/utils.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  if (dataSource.isLoaded()) {
+    return dataSource.items() as T[];
+  }
+
+  return new Promise((resolve, reject) => {
+    dataSource
+      .load()
+      .then(resolve, reject);
+  });
```
wdevfx, line 49: Minor comment.

Let's refactor a little bit:
```ts
if (!forceReload || dataSource.isLoaded()) {
    return dataSource.items() as T[];
  }

  return new Promise((resolve, reject) => {
    dataSource
      .load()
      .then(resolve, reject);
  });
```

packages/devextreme/js/__internal/scheduler/utils/resource_manager/agenda_group_utils.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    return {
+      ...node,
+      children: node.children.length
+        ? node.children
+          .map(filterGroupTree)
+          .filter(Boolean) as GroupNode[]
+        : [],
+    };
+  };
```
wdevfx: Let's move these nested functions to the plain scope of this TS module
(No nested functions)

packages/devextreme/js/__internal/scheduler/utils/resource_manager/agenda_group_utils.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  return groupsTree
+    .map(filterGroupTree)
+    .filter(Boolean) as GroupNode[];
+};
+
+interface DeprecatedGroupNode {
```
wdevfx, line 45: Minor comment.

Let's place this interface on top of this file

packages/devextreme/js/__internal/scheduler/utils/resource_manager/appointment_color_utils.ts
```diff
@@ -0,0 +1,80 @@
...
+  const { groupIndex, itemData } = appointmentConfig;
+  const appointmentGroupValues = getAppointmentGroupValues(itemData, resources);
+  const appointmentGroups = Object.keys(appointmentGroupValues);
+  const paintedResource = getPaintedResource(
+    resources,
+    appointmentGroups,
+    groups,
+  );
+
+  if (paintedResource) {
```
wdevfx: Minor comment.

Let's invert this condition to reduce function complexity
E.g:

```ts
if (!paintedResource) {
  return undefined;
}

// other code
```

packages/devextreme/js/__internal/scheduler/utils/resource_manager/popup_utils.ts
```diff
@@ -0,0 +1,24 @@
+import type { SimpleItem } from '@js/ui/form';
```
wdevfx, line 1: Question.

Should this utility be closer to the popup because it's created for the popup?

packages/devextreme/js/__internal/scheduler/utils/resource_manager/resource_manager.ts
```diff
@@ -0,0 +1,98 @@
...
+  getAppointmentGroupValues,
+  getAppointmentResources,
+} from './appointment_groups_utils';
+import { groupResources } from './group_utils';
+import type { GroupLeaf, GroupNode } from './types';
+
+export class ResourceManager {
+  constructor(
+    config: ResourceConfig[],
+    public resources: ResourceLoader[] = [],
```
wdevfx: Same comment for extra ctor params

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
```diff
@@ -312,6 +324,10 @@ class SchedulerWorkSpace extends WidgetObserver {
     return this._cache;
   }
 
+  get resourceManager(): ResourceManager {
```
wdevfx, line 328: Minor comment.
Let's get rid of this getter

packages/devextreme/testing/helpers/memoryLeaksHelper.js
```diff
@@ -81,6 +81,12 @@
         ]) === -1;
     };
 
+    exports.componentCanBeAsyncInstantiated = function(componentName) {
```
wdevfx, line 84: Question.

Why do we need this change?

packages/devextreme/testing/helpers/scheduler/waitForAsync.js
```diff
@@ -0,0 +1,24 @@
+export const waitAsync = (timeout = 0) => new Promise((resolve) => setTimeout(resolve, timeout));
```
wdevfx: Let's not use real timeouts in tests
We have ```sinon``` with ```fakeTimers``` for this

packages/devextreme/testing/helpers/scheduler/helpers.js
```diff
@@ -74,9 +76,23 @@ export const supportedScrollingModes = ['standard', 'virtual'];
...
+export const initTestMarkup = () => {
+    return $(`#${TEST_ROOT_ELEMENT_ID}`)
+        .html(`<div id="${SCHEDULER_ID}"><div data-options="dxTemplate: { name: 'template' }">Task Template</div></div>`);
+};
 
-export const createWrapper = (option, clock) => new SchedulerTestWrapper($(`#${SCHEDULER_ID}`).dxScheduler(option).dxScheduler('instance'), clock);
+export async function createWrapper(options, selector = `#${SCHEDULER_ID}`) {
+    const instance = $(selector).dxScheduler(options).dxScheduler('instance');
+    const scheduler = new SchedulerTestWrapper(instance);
+    await waitForAsync(() => instance._workSpace);
```
wdevfx, line 87: Let's better resolve the promise in ```onContentReady``` hook

If this approach will not work, the ```await``` of ```Promise.resolve()``` should help us

If both this approaches not work -> it means that we did sth wrong
Because our users wouldn't be able to wait initial Scheduler render too

packages/devextreme/js/__internal/scheduler/appointments/data_provider/m_appointment_filter.ts
Ambrozy: Also this function doesn't work with object resource id

packages/devextreme/js/__internal/scheduler/utils/loader/resource_loader.ts
wdevfx: Discussed, postpone maybe we will find about more correct name in the future

packages/devextreme/js/__internal/scheduler/utils/is_agenda_workpace_component.ts
wdevfx: Discussed, decided to left it here (we already have similar utils)

packages/devextreme/js/__internal/scheduler/utils/data_accessor/appointment_data_accessor.ts
wdevfx: We will discuss question about barrel files laters

packages/devextreme/js/__internal/scheduler/utils/loader/resource_loader.ts
wdevfx: Decided to leave it for future features / refactorings

packages/devextreme/js/__internal/scheduler/utils/resource_manager/agenda_group_utils.ts
wdevfx: It's closer to usage place

packages/devextreme/js/__internal/scheduler/utils/resource_manager/popup_utils.ts
wdevfx: No, it's used insede the root shceduler module, so leave it as is

packages/devextreme/testing/helpers/memoryLeaksHelper.js
wdevfx: Discussed, will notify the components squads about this change

packages/devextreme/js/__internal/scheduler/utils/loader/utils.ts
Ambrozy: forceReload  -> `reload`. otherwise `load`. Different functions are used

packages/devextreme/testing/helpers/scheduler/waitForAsync.js
Ambrozy: task created

wdevfx:
Unfortunately, found the broken case
If Scheduler has groups, and we change the ```resources``` option synchronously after component creation -> appointments will not be loaded

Configuration:
```js
const resourceA = {
  fieldExpr: 'resourceId',
  dataSource: [{
    text: 'A',
    id: 1,
    color: '#1e90ff',
  }],
};
const resourceB = {
  fieldExpr: 'resourceId',
  dataSource: [{
    text: 'B',
    id: 1,
    color: '#f00',
  }],
};
 const appointment = {
  startDate: '2024-01-01T01:00:00',
  endDate: '2024-01-01T02:00:00',
  text: 'Test appt',
  resourceId: 1,
}

const scheduler = $('#scheduler').dxScheduler({
  dataSource: [appointment],
  views: [{
    type: 'workWeek',
    name: 'TestView',
  }],
  currentView: 'TestView',
  currentDate: '2024-01-01',
  groups: ['resourceId'],
  resources: [resourceA],
}).dxScheduler('instance');

// Change resources synchronously
scheduler.option('resources', [resourceB]);
```

In the previous `24.2``` version, this case works well (but with a remote DataSource, we will receive the internal error in both versions)

I checked other similar cases (like similar runtime ```groups``` change, change ```resources``` without groups and etc.),
Looks like other cases works similarly

packages/devextreme/js/__internal/scheduler/appointments/data_provider/m_appointment_filter.ts
wdevfx: Decided to postpone this improvement, because ```id``` or ```item.id``` may be an object
So, we cannot just use hash map here

wdevfx (approved):

packages/devextreme/testing/helpers/scheduler/waitForAsync.js
```diff
@@ -0,0 +1,27 @@
+export const waitAsync = (timeout = 0, clock = undefined) =>
```
wdevfx, line 1: [🚨 Important]

Discussed and decided to get rid of this ```waitAsync``` in terms of a separate tech task

It's better to use the ```sinon.fakeTimers``` instead of the real marco tasks

PR #29932, opened 2025-05-21
CardView - Filtering: can't set filter in filterBuilder if headerFilter.dataSource is present

Raushen:


packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/compatibility.ts
```diff
@@ -36,4 +44,16 @@ export class CompatibilityHeaderFilterController {
   public hideHeaderFilterMenu(): void {
     this.realHeaderFilterViewController.closePopup();
   }
+
+  public getDataSource(column: Column): any {
+    const adapter = this.realDataController.getStoreLoadAdapter();
+    const popupOptions = {
+      column,
```
wdevfx: Let's spread the copy column just in case

packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/compatibility.ts
```diff
@@ -36,4 +44,16 @@ export class CompatibilityHeaderFilterController {
...
+      filterType: column.filterType,
+      filterValues: column.filterValues,
+    } as any;
```
wdevfx: Let's remove the ```any``` from here
(Let's take a look on the ```PopupOptions``` type)

wdevfx (dismissed):

wdevfx (approved):

PR #29964, opened 2025-05-23
CardView - demos - Columns Reordering

pomahtri:


apps/demos/Demos/CardView/ColumnsReordering/jQuery/data.js
```diff
@@ -1,109 +1,298 @@
...
-  State: 'Minnesota',
-  Zipcode: 55403,
-  Phone: '(612) 304-6073',
-  Fax: '(612) 304-6074',
+  OrderNumber: 35703,
+  OrderDate: '2017/04/10',
+  DeliveryDate: '2017/04/13',
+  SaleAmount: 11800,
+  Terms: '15 Days',
+  Location: 'Los Angeles, CA',
```
tongsonbarbs, line 8: In the demo spec, it is mentioned that Location column should be a concatenation of `customer state + customer city`. Are we certain about it because I see that the Orders data array does not have customer/order state and city.. 
 
I just want to be clear that this is expected and there's not a change needed for Orders data..

tongsonbarbs (approved):

apps/demos/Demos/CardView/ColumnsReordering/jQuery/data.js
pomahtri: Discussed in DM, we will let PMs decide

PR #29981, opened 2025-05-26
CardView - Themebuilder

Tucchhaa:


packages/devextreme-scss/scss/widgets/base/cardView/header_panel/item/_variables.scss
```diff
@@ -1,18 +1,42 @@
...
+$cardview-header-panel__item__icon-size: null !default;
+$cardview-header-panel__item__header-filter-icon--selected-color: null !default;
+$cardview-header-panel__item__sort-index-text-color: null !default;
+$cardview-header-panel__item__sort-index-text-weight: null !default;
 
-$cardview-header-item-border-width: null !default;
-$cardview-header-item-border-radius: null !default;
-$cardview-header-item-padding-horizontal: null !default;
-$cardview-header-item-padding-vertical: null !default;
+$cardview-header-panel__item--hovered-bg-color: null !default;
```
pomahtri, line 17: what's the pattern? Why not `$cardview__header-panel__item`?

Tucchhaa: I considered `$cardview-` as a prefix for variables related to CardView, but not as a component. So I didn't replace it to `$cardview__`.

I suggest that we keep current naming, because if I replace in that way, then we will have this kind of names in colors.scss, sizes.scss files:  `cardview__fluent-header-panel__item`. I am not sure if it's a good idea.

GoodDayForSurf (approved):

Alyar666 (approved):

PR #29982, opened 2025-05-26
CardView: sorting from context menu displays "NaN" instead of sort idx

Raushen:


packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.mock.ts
```diff
@@ -1,7 +1,7 @@
 import type { Item as ContextMenuItem } from '@js/ui/context_menu';
 
-import type { ContextInfo, ContextMenuTarget } from '.';
-import { ContextMenuController } from '.';
+import type { ContextInfo, ContextMenuTarget } from './index';
```
wdevfx: I missed this point during ```ContextMenu``` review
@Tucchhaa, why do we need this controller mock?

If we don't need it, @Raushen let's delete it

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.ts
```diff
@@ -65,29 +71,42 @@ export class ContextMenuController
...
+        case 'single':
+          this.sortingController.onSingleModeSortCore(column, true, sortOrder);
+          break;
+        case 'multiple':
+          this.sortingController.onMultipleModeSortCore(column, false, sortOrder);
+          break;
+        default:
+          break;
+      }
     };
```
wdevfx: It's better to get rid off the nested function and pass the ```mode``` with arguments
(We already pass it too, but with ```closure```, it's less obvious)

E.g:
```ts
private handleSortMenuClick(event: ItemClickEvent, mode: SingleMultipleOrNone | undefined): void => {
  // code
}

private getSortingItems(column: Column): ContextMenuItem[] {
  // code
  return [
    // other items
    {
      // ...
      onItemClick: (event: ItemClickEvent) => { this.handleSortMenuClick(event, mode); }
    },
    // other items
  ]
}
```

packages/devextreme/js/__internal/grids/new/card_view/di.ts
```diff
@@ -19,4 +20,5 @@ export function register(diContext: DIContext): void {
 
   diContext.register(ContextMenuController);
   diContext.register(BaseContextMenuController, ContextMenuController);
+  diContext.register(ContextMenuControllerMock, ContextMenuController);
```
wdevfx: Let's not register mocks in the real component ```DI```
Why do we need it here?

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.ts
```diff
@@ -65,29 +71,42 @@ export class ContextMenuController
   }
 
   private getSortingItems(column: Column): ContextMenuItem[] {
+    const mode = this.sortingController.mode.value;
+
     const onItemClick = (e: ItemClickEvent): void => {
```
wdevfx: Minor comment.

Maybe it will be a little more obvious and better rename the ```ItemClickEvent``` type to understand from code that this event from the ```ContextMenu``` public API

E.g
```ts
import type { 
  Item as ContextMenuItem,
  ItemClickEvent as ContextMenuItemPressedEvent, // it's not only click, it also may be KeyboardEvent under the hood
 } from '@js/ui/context_menu';
```

packages/devextreme/js/__internal/grids/new/grid_core/sorting_controller/controller.ts
```diff
@@ -134,6 +135,14 @@ export class SortingController {
       || this.sortedColumns.peek().length > 1;
     const nextSortOrder = getNextSortOrder(column.sortOrder, isCtrl);
 
+    this.onSingleModeSortCore(column, isClearSortingRequired, nextSortOrder);
```
wdevfx, line 138: Minor comment / for future refactoring.

The ```onSingleModeSortClick``` / ```onMultipleModeSortClick``` now accept native events as arguments
And these methods are only used in the ```HeaderPanelView```

From my perspective, it will be better to move event handling logic to the ```HeaderPanelView``` to separate the controller more from the view and related to the view things :)

E.g., we can have a similar interface here as in the ```onSingleModeSortCore``` / ```onMultipleModeSortClick```
And don't touch events in the controller at all

@pomahtri, what do u think about it?

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.ts
Raushen: We can not rename it as it is [a public type](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxMenu/Types/ItemClickEvent/).

packages/devextreme/js/__internal/grids/new/grid_core/sorting_controller/controller.ts
pomahtri: It will be good, but sounds like refactoring outside of the bug

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.mock.ts
Tucchhaa: It was needed for testing. 

Method getItems() in ContextMenuController is protected, but in ContextMenuControllerMock it's public. I didn't want to make this method public to other components.

packages/devextreme/js/__internal/grids/new/card_view/di.ts
Tucchhaa: Maybe we can move registration of ContextMenuControllerMock to the di.test_utils.ts?

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.mock.ts
Tucchhaa: According to the changes in PR looks like we don't need `getItems` to be public. So can we keep this controller? I think it's a good idea to keep methods/properties as private as possible

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.ts
```diff
@@ -65,32 +72,55 @@ export class ContextMenuController
...
       {
         text: this.options.oneWay('sorting.ascendingText').peek(),
         value: 'asc',
-        disabled: column.sortOrder === 'asc',
+        disabled: isDisabled || column.sortOrder === 'asc',
         icon: 'sortuptext',
-        onItemClick,
+        onItemClick: (event: ContextMenuItemPressedEvent): void => {
+          this.handleSortMenuClick(event, mode, column);
+        },
```
Tucchhaa: can we move this function to a variable and not duplicate it for each item?


        
```suggestion
const onItemClick = (event: ContextMenuItemPressedEvent): void => {
    =this.handleSortMenuClick(event, mode, column);
},

return [{
        // ...,
        onItemClick,
        }, {
        // ...,
        onItemClick
        }];
```

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.mock.ts
Raushen: I think it is OK to make public modifier for those methods that need to be tested.

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.ts
wdevfx: I guess it's a minor suggestion, so @Raushen it's up to u :)

wdevfx (dismissed):

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.mock.ts
wdevfx: From my perspective, if method ```private``` or ```protected``` we should not cover it directly by tests
If we need to do it -> it's a design mistake

So, making some method public with an inheritor test class is the same as writing the ```@ts-expect-error``` and calling it directly in tests

So, for now I suggest to leave it as is

packages/devextreme/js/__internal/grids/new/card_view/di.ts
wdevfx: Now the ```ContextMenuControllerMock``` doesn't exists, so I resolve this comment

Tucchhaa (approved):

PR #29989, opened 2025-05-26
TabPanel: Added focus to the selected tab after initialization (T1277409)

pharret31:


packages/devextreme/js/__internal/ui/tab_panel/tab_panel.ts
```diff
@@ -259,6 +259,13 @@ class TabPanel extends MultiView<TabPanelProperties> {
       .addClass(TABPANEL_CONTAINER_CLASS)
       .appendTo($element);
     this._$container.append(this._$wrapper);
+
+    const { focusStateEnabled, selectedIndex } = this.option();
+    if (focusStateEnabled) {
+      // @ts-expect-error ts-error
```
anna-shakhova: Resolve this ts-error, please.
Is it possible for selectedIndex to be undefined? In collection default value is `-1`, how will it work here?

packages/devextreme/js/__internal/ui/tab_panel/tab_panel.ts
```diff
@@ -259,6 +259,13 @@ class TabPanel extends MultiView<TabPanelProperties> {
...
+      const selectedItem = this._tabs.itemElements().get(selectedIndex);
+      this._tabs.option({ focusedElement: selectedItem ?? null });
```
anna-shakhova: Do we need `?? null`? 
According to return type, `.get` returns Element, check, please

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabPanel.tests.js
```diff
@@ -1035,6 +1035,34 @@ QUnit.module('focus policy', {
 
         assert.strictEqual($tabPanel.find(`.${MULTIVIEW_WRAPPER_CLASS}`).hasClass(FOCUS_STATE_CLASS), false);
     });
+
+    QUnit.test('first tab should be focused after initialization', function(assert) {
+        if(shouldSkipOnDevices({ deviceTypes: ['phone', 'tablet'], assert })) {
```
anna-shakhova: it shouldn't be skipped for mobile devices, all that matters is `focusStateEnabled` option, and we can set it to `true` on initialization

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabPanel.tests.js
```diff
@@ -1035,6 +1035,34 @@ QUnit.module('focus policy', {
...
+        const items = Array.from({ length: 20 }, (_, index) => ({
+            title: `Tab ${index + 1}`,
+            text: `Content ${index + 1}`
+        }));
+
+        const $tabPanel = $('#tabPanel').dxTabPanel({ items });
+
+        const $tabs = $tabPanel.find(`.${TABPANEL_TABS_ITEM_CLASS}`);
+        const $firstTab = $tabs.first();
+        assert.strictEqual($firstTab.hasClass(FOCUS_STATE_CLASS), true, 'first tab is focused');
```
anna-shakhova: Let's assert `focusedElement` option, as in second test

packages/devextreme/js/__internal/ui/tab_panel/tab_panel.ts
```diff
@@ -259,6 +259,13 @@ class TabPanel extends MultiView<TabPanelProperties> {
       .addClass(TABPANEL_CONTAINER_CLASS)
       .appendTo($element);
     this._$container.append(this._$wrapper);
+
+    const { focusStateEnabled, selectedIndex } = this.option();
+    if (focusStateEnabled) {
```
ksercs: Why should we do it only when focusStateEnabled? Is it a condition for all places where focusedElement is updated/used?

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabPanel.tests.js
```diff
@@ -1035,6 +1035,34 @@ QUnit.module('focus policy', {
 
         assert.strictEqual($tabPanel.find(`.${MULTIVIEW_WRAPPER_CLASS}`).hasClass(FOCUS_STATE_CLASS), false);
     });
+
+    QUnit.test('first tab should be focused after initialization', function(assert) {
```
ksercs: please add a ticket id how we usually do (and to PR name as well)

packages/devextreme/js/__internal/ui/tab_panel/tab_panel.ts
pharret31: It's possible, also if selectedIndex is not actual value in the _tabs' items the .get() method will return undefined value

packages/devextreme/js/__internal/ui/tab_panel/tab_panel.ts
pharret31: If there's no element it returns undefined, although null is better option

packages/devextreme/js/__internal/ui/tab_panel/tab_panel.ts
pharret31: As far as I understand if focusStateEnabled is false it won't use focus states and there won't be an issue with them the users faced

anna-shakhova (approved):

PR #29997, opened 2025-05-27
fix(inferno): Add logic to avoid class repetition due to ssr

ajivanyandev:


VasilyStrelyaev (approved):

pomahtri:
As I see, `combineClasses` is used only in end-components, not somewhere in the core logic
So any component can just omit this utility and add classes manually, like `className={"${classA} ${classB}"}`
Let's try to find some place which is always called

Can you also please write description with full lifecycle why problem occurs? To make reviewing easier and just for history

ajivanyandev:
> As I see, `combineClasses` is used only in end-components, not somewhere in the core logic So any component can just omit this utility and add classes manually, like `className={"${classA} ${classB}"}` Let's try to find some place which is always called
> 
> Can you also please write description with full lifecycle why problem occurs? To make reviewing easier and just for history

https://wiki.devexpress.devx/en/devextreme/product/new-page

pomahtri (approved):

PR #30002, opened 2025-05-28
TreeList – Chevron/Caret toggle icon is misaligned in the Fluent theme (T1291914)

tongsonbarbs:


e2e/testcafe-devextreme/tests/treeList/markup.ts
```diff
@@ -159,3 +159,29 @@ const tasksT1223168 = [{
...
 });
+
+test('TreeList – Chevron/Caret toggle icon is misaligned in the Fluent theme (T1291914)', async (t) => {
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const treeList = new TreeList('#container');
+
+  await treeList.isReady();
+
+  await t
+    .expect(await takeScreenshot('treelist-expand-collapse-caret-center-fluent-theme', treeList.element))
```
Raushen: There is no file extension '.png' here

e2e/testcafe-devextreme/tests/treeList/markup.ts
```diff
@@ -159,3 +159,29 @@ const tasksT1223168 = [{
...
+    .ok()
+    .expect(compareResults.isValid())
+    .ok(compareResults.errorMessages());
+})
+  .before(async () => {
+    await changeTheme(Themes.fluentBlue);
```
Raushen: How about to test all themes?

packages/devextreme-scss/scss/widgets/fluent/treeList/_sizes.scss
```diff
@@ -3,6 +3,6 @@
 
 // adduse
 
-$treelist-expanded-icon-margin-left: 5px !default;
+$treelist-expanded-icon-margin-left: 3px !default;
 $treelist-collapsed-icon-margin-left: 6px !default;
-$treelist-icon-margin-top: math.div($fluent-base-icon-size, 2) !default;
+$treelist-icon-margin-top: math.div(($fluent-base-icon-size + 4), 2) !default;
```
Raushen: I think we can avoid calculations and just write the correct size as a number

packages/devextreme-scss/scss/widgets/fluent/treeList/_index.scss
```diff
@@ -127,7 +120,7 @@ $treelist-icon-container-padding-right: $fluent-grid-base-checkbox-size + $fluen
 
     .dx-checkbox {
       top: 50%;
-      margin-top: math.div(-$fluent-grid-base-checkbox-size, 2);
+      margin-top: math.div(-$fluent-checkbox-icon-font-size, 4);
```
Raushen: How about fix styles in other themes?

packages/devextreme-scss/scss/widgets/fluent/treeList/_sizes.scss
tongsonbarbs: It's purpose is if we need to change value for `$fluent-base-icon-size`, so whatever the value is, it will always stay in the middle/center..

packages/devextreme-scss/scss/widgets/fluent/treeList/_index.scss
tongsonbarbs: Will try this..

tongsonbarbs: Unfortunately we cannot apply the same changes for generic theme as it mostly follows inherited values for cell and row sizing.

If we apply the same CSS rules for generic, it will change visuals for all TreeList demos which will also have significant/obvious visual change for generic.

packages/devextreme-scss/scss/widgets/fluent/treeList/_sizes.scss
Raushen: Of cause it will not be changed because changing icon size is a great BC

PR #30006, opened 2025-05-28
DataGrid: Fix autoscroll when reordering columns using the keyboard when native scrolling is enabled

Alyar666:


e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/columnReordering.visual.ts
```diff
@@ -1472,162 +1472,165 @@ test('reorder column to left when adaptability is enabled and there are hidden c
...
-  const firstHeaderCell = dataGrid.getHeaders().getHeaderRow(0).getHeaderCell(0);
+  test(`${useNative ? 'Native' : 'Simulated'} scrolling: Auto scroll to the right when column reordering via keyboard when virtual columns are enabled`, async (t) => {
+    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+    const dataGrid = new DataGrid(DATA_GRID_SELECTOR);
+    const firstHeaderCell = dataGrid.getHeaders().getHeaderRow(0).getHeaderCell(0);
 
-  await t
-    .click(firstHeaderCell.element);
+    await t
+      .debug()
```
wdevfx: Let's remove this debug :)

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_headers_keyboard_navigation.ts
```diff
@@ -253,18 +245,60 @@ export class HeadersKeyboardNavigationController extends ColumnKeyboardNavigatio
...
-    super.restoreFocus();
+  public needToFocus(): boolean {
+    return this.needToRestoreFocus;
   }
 }
 
+const columnHeadersView = (
+  Base: ModuleType<ColumnHeadersView>,
+) => class ColumnHeadersViewKeyboardNavigationExtender extends Base {
+  protected handleScroll(e): void {
```
wdevfx: Let's add a type for ```e```

packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
```diff
@@ -1072,6 +1072,14 @@ export class ColumnsView extends ColumnStateMixin(modules.View) {
     }
   }
 
+  protected handleScroll(e): void {
```
wdevfx: Let's add a type for ```e``` here

packages/devextreme/js/ui/shared/accessibility.js
```diff
@@ -148,14 +148,19 @@ export function registerKeyboardAction(viewName, instance, $element, selector, a
         }
         isMouseDown = false;
     };
+    const mouseUpHandler = () => {
+        isMouseDown = false;
+    };
```
wdevfx, line 153: Question.

Why do we need this change?

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_headers_keyboard_navigation.ts
```diff
@@ -253,18 +245,60 @@ export class HeadersKeyboardNavigationController extends ColumnKeyboardNavigatio
...
+    super.handleScroll(e);
+
+    const headersKeyboardNavigation = this.getController('headersKeyboardNavigation');
```
wdevfx: Let's not use the ```getController``` dynamically
It's tough to understand relations between entities if we use it like this

Therefore, move the ```getController``` call to the ```init``` method of base extended class and use ```this.headersKeyboardNavigation``` here instead
E.g. like we use here ```_columnsController```

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation_core.ts
```diff
@@ -101,14 +101,16 @@ export class KeyboardNavigationController extends modules.ViewController {
 
   protected initHandlers(): void {
     const focusedView = this.getFocusedView();
+    const resizeController = this.getController('resizing');
```
wdevfx: Same comment about ```getController```

Let's use it only inside the ```init``` method to set it to the related class field

packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
```diff
@@ -1082,14 +1090,7 @@ export class ColumnsView extends ColumnStateMixin(modules.View) {
...
     }
-
-    eventsEngine.on($scrollContainer, 'scroll', () => {
-      const scrollLeft = $scrollContainer.scrollLeft();
-
-      if (scrollLeft !== this._scrollLeft) {
-        this.scrollChanged.fire({ left: scrollLeft }, this.name);
-      }
-    });
+    eventsEngine.on($scrollContainer, 'scroll', this.handleScroll.bind(this));
```
wdevfx, line 1094: Question.

Is it ok that we don't unsubscribe from the ```scroll``` event?
I mean call of the ```eventsEngine.off```

packages/devextreme/js/ui/shared/accessibility.js
wdevfx: Discussed -> We need this because not after each```contextmenu``` event focus fired

packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
wdevfx: Hm, I re-checked this question
It's better to unsubscribe manually, because native event listeners will be cleaned by GC only if nothing in memory points to the DOM node

So, just in case it's better to call ```off``` somewhere

wdevfx (approved):

PR #30009, opened 2025-05-28
fix: jQuery event types contain excessive props but lack originalEvent

VasilyStrelyaev:
~This PR is planned for merging to 25_2 once it's available (that's why it's a draft so far)~

The latest version does not contain BCs


packages/devextreme/js/events/events.types.d.ts
```diff
@@ -2,6 +2,79 @@
...
+   * @docid
+   * @public
+   */
+  char: any;
+
+  /**
+   * @docid
+   * @public
+   */
+  charCode: number;
```
pomahtri: Please check whether we need to add `?` or just `| undefined`, as some params are suitable only for limited number of events, e.g. here `charCode` only for KeyboardEvent, not for MouseEvent

pomahtri: fixed

mpreyskurantov (dismissed):

packages/devextreme/js/events/events.types.d.ts
```diff
@@ -1,7 +1,92 @@
+type PickFromEvent<TNativeEvent extends Event, TTargetEvents extends Event, PropName extends keyof TTargetEvents> =
+  TNativeEvent extends Pick<TTargetEvents, PropName>
+    ? TTargetEvents[PropName]
+    : undefined;
```
IlyaKhD: I guess it's meant to be `never`:

```suggestion
    : never;
```

pomahtri: `undefined` is better
`never` is assignable to any other type, it will hide errors
https://www.typescriptlang.org/play/?#code/CYUwxgNghgTiAEYD2A7AzgF3lAXPFIAbiDANwBQ5y6WARnpjAJYoDm8AvNhdZonigCuAW1olO3ckA

chaosmirage (approved):

anna-shakhova (approved):

PR #30020, opened 2025-05-29
Toolbar: prevent overflow menu from closing on editor component click (T1287462)

pharret31:


packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
```diff
@@ -10,13 +10,15 @@ const TOOLBAR_MENU_CUSTOM_CLASS = 'dx-toolbar-menu-custom';
...
-  _activeStateUnit?: string;
+  _activeStateUnit!: string;
+
+  _nonActionableComponents!: string[];
 
   _init(): void {
     super._init();
 
     this._activeStateUnit = `.${TOOLBAR_MENU_ACTION_CLASS}:not(.${TOOLBAR_HIDDEN_BUTTON_GROUP_CLASS})`;
+    this._nonActionableComponents = ['dxDropDownButton', 'dxSelectBox', 'dxNumberBox', 'dxColorBox'];
```
anna-shakhova: Is it possible to place in toolbar menu other types of editors? LIke Autocomplete, DateBox, DateRangeBox, Menu, some others?

packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
```diff
@@ -10,13 +10,15 @@ const TOOLBAR_MENU_CUSTOM_CLASS = 'dx-toolbar-menu-custom';
 const TOOLBAR_MENU_LAST_SECTION_CLASS = 'dx-toolbar-menu-last-section';
 const SCROLLVIEW_CONTENT_CLASS = 'dx-scrollview-content';
 export default class ToolbarMenuList extends ListBase {
-  // @ts-expect-error ts-error
-  _activeStateUnit?: string;
+  _activeStateUnit!: string;
+
+  _nonActionableComponents!: string[];
```
anna-shakhova: Do we really need private field to store that list?

packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
```diff
@@ -67,29 +69,44 @@ export default class ToolbarMenuList extends ListBase {
...
+
+    return itemElement;
+  }
+
+  _getItemCssClasses(item): string[] {
+    const location = item.location ?? 'menu';
+    const cssClasses: string[] = [];
+
+    if (item.cssClass) {
+      cssClasses.push(item.cssClass);
```
anna-shakhova: classes sequence was changed - please, check, it won't break styles, if item class rewrites some property from our base classes

packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
```diff
@@ -67,29 +69,44 @@ export default class ToolbarMenuList extends ListBase {
...
     if (this._getItemTemplateName({ itemData: item })) {
-      itemElement.addClass(TOOLBAR_MENU_CUSTOM_CLASS);
+      cssClasses.push(TOOLBAR_MENU_CUSTOM_CLASS);
+    }
+
+    if (this._nonActionableComponents.includes(item.widget)) {
+      return cssClasses;
     }
 
     if (location === 'menu' || item.widget === 'dxButton' || item.widget === 'dxButtonGroup' || item.isAction) {
```
anna-shakhova: I think it is safer to narrow this condition, btw what is `menu` location? According to type we have `after | before |  center`

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.tests.js
```diff
@@ -576,6 +578,49 @@ QUnit.module('toolbar with menu', moduleConfig, () => {
 
         assert.strictEqual($('.dx-state-disabled').length, 0, 'disabled state changed');
     });
+
+    [
+        {
+            component: 'dxSelectBox',
+            cssClass: `.${SELECTBOX_CLASS}`,
```
ksercs: Why do you add box classes as a custom one? Looks like nothing is changed if you don't add it here

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.tests.js
```diff
@@ -576,6 +578,49 @@ QUnit.module('toolbar with menu', moduleConfig, () => {
...
+            options: {
+                displayExpr: 'value',
+                keyExpr: 'id',
+                items: [
+                    { id: 1, value: '1' },
+                    { id: 2, value: '2' },
+                    { id: 3, value: '3' },
+                ],
```
ksercs: these 7 rows can be replaced with `items: ['item 1']`.
Let's simplify tests as much as possible

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.tests.js
```diff
@@ -576,6 +578,49 @@ QUnit.module('toolbar with menu', moduleConfig, () => {
...
+        QUnit.test(`click on ${component} inside the toolbar menu should not close it (T1287462)`, function(assert) {
+            this.instance.option('items', [{
+                locateInMenu: 'always',
+                widget: component,
+                ...options,
+            }]);
+
+            this.overflowMenu.click();
+
+            const $component = this.overflowMenu.instance()._popup.$content().find(cssClass);
```
ksercs: Could we do it without private API?

packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
```diff
@@ -67,29 +66,47 @@ export default class ToolbarMenuList extends ListBase {
...
+    }
+
+    if (item.widget && !actionableComponents.includes(item.widget)) {
+      cssClasses.push(item.cssClass);
+      return cssClasses;
     }
 
-    if (location === 'menu' || item.widget === 'dxButton' || item.widget === 'dxButtonGroup' || item.isAction) {
-      itemElement.addClass(TOOLBAR_MENU_ACTION_CLASS);
+    if (location === 'menu' || item.isAction || actionableComponents.includes(item.widget)) {
```
ksercs: What is `item.isAction`? I don't see where this property is set in our code nor in the API

packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
```diff
@@ -67,29 +66,47 @@ export default class ToolbarMenuList extends ListBase {
...
+    const actionableComponents = this._getActionableComponents();
+
     if (this._getItemTemplateName({ itemData: item })) {
-      itemElement.addClass(TOOLBAR_MENU_CUSTOM_CLASS);
+      cssClasses.push(TOOLBAR_MENU_CUSTOM_CLASS);
+    }
+
+    if (item.widget && !actionableComponents.includes(item.widget)) {
+      cssClasses.push(item.cssClass);
+      return cssClasses;
```
ksercs: It's strange for me to see this assert here - cannot it be refactored to leave only one `return` in this end? Previous algorithm like "add classes for specific cases, then add custom classes and return" looked more intuitive
Or at least we can do `push(item.cssClass` before this `if` to reduce copypaste

Actually looks like we can just remove `location === 'menu'` from the condition below, nah?

packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
```diff
@@ -67,29 +66,47 @@ export default class ToolbarMenuList extends ListBase {
...
     }
 
-    itemElement.addClass(item.cssClass);
+    cssClasses.push(item.cssClass);
+    return cssClasses;
+  }
 
-    return itemElement;
+  _getActionableComponents() {
+    return ['dxButton', 'dxButtonGroup'];
```
anna-shakhova, line 103: What about dxCheckbox and dxSwitch? They are similar to button group

packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
```diff
@@ -67,29 +66,47 @@ export default class ToolbarMenuList extends ListBase {
...
+    const cssClasses: string[] = [];
+    const actionableComponents = this._getActionableComponents();
+
     if (this._getItemTemplateName({ itemData: item })) {
-      itemElement.addClass(TOOLBAR_MENU_CUSTOM_CLASS);
+      cssClasses.push(TOOLBAR_MENU_CUSTOM_CLASS);
+    }
+
+    if (item.widget && !actionableComponents.includes(item.widget)) {
+      cssClasses.push(item.cssClass);
```
anna-shakhova: Duplicated line, perhaps, it is better to rewrite condition below, look like location matters only for items without widget property defined

packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
anna-shakhova:
```suggestion
     const isAction = item.isAction 
     || (location === 'menu' && !item.widget) 
     || actionableComponents.includes(item.widget);
    if (isAction) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.tests.js
```diff
@@ -1236,7 +1281,7 @@ QUnit.module('adaptivity', moduleConfig, () => {
         assert.ok($sections.eq(2).hasClass('dx-toolbar-menu-last-section'), 'border for last section is removed');
     });
 
-    QUnit.test('menu shouldn\'t be closed after click on editors', function(assert) {
+    QUnit.test('menu should not be closed after click on editors', function(assert) {
```
anna-shakhova: This test passed before changes, but we still had an issue. So it's either has incorrect name (why editors, if no `widget` is defined) or configuration (it does not test undefined `location`)
Let's fix it

packages/devextreme/js/__internal/ui/toolbar/internal/m_toolbar.menu.list.ts
ksercs: It's basically editors not just actions, so looks okay for me to leave menu open for them

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.tests.js
pharret31: I used it to find the item inside the toolbar menu but you're right, it seems I can find it by dx-item selector, removed

anna-shakhova (approved):

PR #30022, opened 2025-05-29
CardView - Header Filter: Fix filtering when a custom headerFilter.dataSource is specified as an array of filter expressions

Alyar666:


packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/utils.ts
```diff
@@ -67,22 +68,38 @@ export const needCreateHeaderFilter = (column: Column): boolean => {
...
+
+    if (filterValuesWithoutExpressions.length) {
+      filterExpression = [getFilterExpression(filterValuesWithoutExpressions, column)];
+    }
 
-    filterValue.push([columnName, filterOperator, normalizedFilterValues]);
+    filterValue.push(gridCoreUtils.combineFilters([...filterExpression, ...filterValuesWithExpressions], 'or'));
 
     if (index < filterableColumns.length - 1) {
       filterValue.push('and');
```
wdevfx: Not a mandatory suggestion from my side

Let's get rid of ```let``` variables, we can refactor this code a little bit and make it simpler
Sth like this:
```ts
  filterableColumns.forEach((column, index) => {
    const { filterValues } = column;
    const normalizedFilterValues = Array.isArray(filterValues)
      ? filterValues
      : [filterValues];

    const filterValuesWithExpressions = normalizedFilterValues
      .filter((value) => Array.isArray(value));
    const filterValuesWithoutExpressions = normalizedFilterValues
      .filter((value) => !Array.isArray(value));

    const filterExpression = filterValuesWithoutExpressions.length
      ? [getFilterExpression(filterValuesWithoutExpressions, column)]
      : [];

    filterValue.push(gridCoreUtils.combineFilters([...filterExpression, ...filterValuesWithExpressions], 'or'));

    if (index < filterableColumns.length - 1) {
      filterValue.push('and');
    }
  });
```
---
**Why is it better than ```let``` use ```const```?**

Maybe it's a little bit holy war theme
But from my perspective, we should always keep ```let``` in mind during code reading
I mean that we should track code execution path starting from the ```let``` initialization to the place that we looking at
With ```const``` things more easier -> we just remember how it initialized and thats all, it cannot be changed on the way to place that we looking at

Little example:
```ts
// initialization
let a = 1;

// other code

// change let during code execution somewhere
if (<some condition>) {
  a = 2;
}

// other code

// place that we looking at
// if we see that the a -> let, we should read all code above to understand it's value
console.log(a);
```

e2e/testcafe-devextreme/tests/cardView/headerFilter/common.functional.ts
```diff
@@ -473,3 +473,52 @@ test('FilterBuilder should work with custom headerFilter data source', async (t)
     filterPanel: { visible: true },
   });
 });
+
+test('Filtering should work when a custom data source is specified as an array of filter expressions', async (t) => {
```
wdevfx, line 477: Let's add one more case to e2e functional checks (maybe two cases)

Test scenario:
- Column initially has filter values with filter expression
- Check that header filter items checked correctly
- Check that cards have the right values (filtered)

wdevfx (approved):

PR #30027, opened 2025-05-30
Chat - enter press should add a new line in TextArea on Mobile devices (T1293840)

Zedwag:


packages/devextreme/js/__internal/ui/chat/messagebox.ts
```diff
@@ -179,6 +184,10 @@ class MessageBox extends DOMComponent<MessageBox, Properties> {
     });
   }
 
+  _isMobile(): boolean {
+    return devices.current().deviceType !== 'desktop';
```
ksercs: It's not a component-specific code => it should not be a method. Let's better create a function.
P.s. ideally it should be an util?

Zedwag: https://github.com/DevExpress/DevExtreme/pull/30027/commits/30efd12c909b3171212700b7a781a04a9d079968

ksercs (approved):

PR #30037, opened 2025-06-03
CardView - fix issues with toolbar items order

pomahtri:


packages/devextreme/js/__internal/grids/new/grid_core/toolbar/const.tsx
```diff
@@ -2,9 +2,9 @@
  * used for defining default toolbar items and their default order in header panel
  */
 export const DEFAULT_TOOLBAR_ITEMS = [
```
Raushen, line 4: Let's rename to SORTED_DEFAULT_TOOLBAR_BUTTONS (or ORDERED_) as without the context it is unclear that the item order is important here

Raushen (approved):

PR #30044, opened 2025-06-03
fix: fixed versions assertion

sergey-kras:


packages/devextreme/js/__internal/utils/version.ts
```diff
@@ -35,16 +30,11 @@ export function parseVersion(version: string): Version {
 }
 
 export function assertDevExtremeVersion(packageName: string, version: string): void {
-  assertedVersions.push({
-    packageName,
-    version,
-  });
+  config({ assertedVersions: [...config()?.assertedVersions ?? [], { packageName, version }] });
```
VasilyStrelyaev: What if we add `assertedVersions: []` to the default config object:
https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/core/m_config.ts#L35

Then remove the `?` from the `GlobalConfig` type prop and avoid these `?? []` checks?

Let's try if this works. It may not, because for some reason all the properties in the `GlobalConfig` type are optional. But worth a try, isn't it?

packages/devextreme/js/__internal/utils/version.ts
```diff
@@ -35,16 +30,11 @@ export function parseVersion(version: string): Version {
...
 }
 
 export function clearAssertedVersions(): void {
-  /// #DEBUG
-  assertedVersions.splice(0);
-  /// #ENDDEBUG
+  config({ assertedVersions: [] });
```
VasilyStrelyaev: I'm not sure what these `/// #DEBUG` directives are here for, but I guess Ilya Kharchenko added them (I could not ever), so I guess they are here for a reason. Let's keep them.

Most probably, he marked methods that should only work during testing or something like that

packages/devextreme/js/common.d.ts
```diff
@@ -270,6 +270,24 @@ export type FirstDayOfWeek = 0 | 1 | 2 | 3 | 4 | 5 | 6;
  */
 export type Format = 'billions' | 'currency' | 'day' | 'decimal' | 'exponential' | 'fixedPoint' | 'largeNumber' | 'longDate' | 'longTime' | 'millions' | 'millisecond' | 'month' | 'monthAndDay' | 'monthAndYear' | 'percent' | 'quarter' | 'quarterAndYear' | 'shortDate' | 'shortTime' | 'thousands' | 'trillions' | 'year' | 'dayOfWeek' | 'hour' | 'longDateLongTime' | 'minute' | 'second' | 'shortDateShortTime';
 
+/**
+ * @public
+ * @namespace DevExpress.common
+ * @type object
+ */
```
VasilyStrelyaev: We did not mark the property (line 424) with any jsdoc tags because we don't want it to show up in the docs (as it's internal API)

So, it makes sense to avoid showing in the docs the type for this property.

Let's just remove the js doc tags from here?

VasilyStrelyaev (changes requested):

VasilyStrelyaev (dismissed):

packages/devextreme/js/common.d.ts
```diff
@@ -278,6 +283,7 @@ export type Format = 'billions' | 'currency' | 'day' | 'decimal' | 'exponential'
  * @type object
  */
 export type GlobalConfig = {
+  assertedVersions?: AssertedVersion[];
```
IlyaKhD: Consider renaming:
```suggestion
  versionAssertions?:  VersionAssertion[];
```

packages/devextreme/js/__internal/core/license/license_validation.test.ts
```diff
@@ -507,3 +511,27 @@ describe('internal license check', () => {
...
+    expect(config().assertedVersions).toEqual([{ packageName: 'test-package', version: '1.2.3' }]);
+  });
+
+  it('clearAssertedVersions config().assertedVersions', () => {
+    assertDevExtremeVersion('test-package', '1.2.3');
+    clearAssertedVersions();
+    expect(config().assertedVersions).toEqual([]);
+  });
+
+  it('assertedVersionsCompatible shoild be false', () => {
```
IlyaKhD: Typo
```suggestion
  it('assertedVersionsCompatible should be false', () => {
```

packages/devextreme/js/__internal/core/license/license_validation.test.ts
```diff
@@ -507,3 +511,27 @@ describe('internal license check', () => {
     expect(trialPanelSpy).not.toHaveBeenCalled();
   });
 });
+
+describe('assertedVersions integration', () => {
+  it('assertDevExtremeVersion config().assertedVersions', () => {
```
IlyaKhD: The rest of the file uses `test`, not `it`, so I suggest sticking with `test` to maintain consistency.
Btw, I think that `it` is generally intended for BDD-style assertions and assumes writing test names in [BDD style](https://en.wikipedia.org/wiki/Behavior-driven_development#Behavioral_specifications)

IlyaKhD: Anyway, please make the test names more descriptive. Reading a test name should be enough to understand the use case it covers.

In general, tests are like specifications. Anyone should be able to understand the desired behavior by reading the test names and assertions.

packages/devextreme/js/__internal/core/license/license_validation.test.ts
```diff
@@ -507,3 +511,27 @@ describe('internal license check', () => {
...
+    expect(config().assertedVersions).toEqual([]);
+  });
+
+  it('assertedVersionsCompatible shoild be false', () => {
+    clearAssertedVersions();
+    assertDevExtremeVersion('test-package', '1.2.3');
+    const logSpy = jest.spyOn(errors, 'log').mockImplementation(() => {});
+    const result = assertedVersionsCompatible({ major: 2, minor: 0, patch: 0 });
+    expect(result).toBe(false);
+    expect(logSpy).toHaveBeenCalled();
```
IlyaKhD: Shouldn't the error or warning code be checked here?

Like this: https://github.com/DevExpress/DevExtreme/pull/30044/files#diff-e94c314939ff3413c1348c191eec009e879000634504494658e8d12a9fd8e21eR503

packages/devextreme/js/__internal/core/license/license_validation.test.ts
```diff
@@ -507,3 +511,27 @@ describe('internal license check', () => {
     expect(trialPanelSpy).not.toHaveBeenCalled();
   });
 });
+
+describe('assertedVersions integration', () => {
```
IlyaKhD, line 543: As I understand it, the PR is about adding version assertions to the config.
I would expect the tests to follow a pattern like: 'put something in the config and then assert'.
However, the config isn't used in the **Arrange**' or **Act** parts of the tests. All the tests seem to perform indirect checks that depend on implementation details.

packages/devextreme/js/__internal/core/license/license_validation.test.ts
IlyaKhD: Btw, what about the case when it's true? I think both branches should be tested

IlyaKhD (dismissed):

VasilyStrelyaev (dismissed):

VasilyStrelyaev (approved):

dxvladislavvolkov (approved):

IlyaKhD (approved):

PR #30049, opened 2025-06-03
CardView - support json URL dataSource

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it } from '@jest/globals';
+import { CustomStore } from '@js/common/data';
 
 import { getContext } from '../di.test_utils';
 import type { Options } from '../options';
@@ -64,6 +65,44 @@ describe('DataController', () => {
       await dataController.waitLoaded();
 
       expect(optionsController.oneWay('paging.pageIndex').peek()).toEqual(0);
+    });
+  });
+
+  describe('when totalCount is not specified', () => {
+    it('works with CustomStore', async () => {
+      try {
+        const { dataController } = setup({
+          dataSource: new CustomStore({
+            load: () => generateData(10),
+          }),
+          paging: {
+            pageSize: 3,
+          },
+        });
+
+        await dataController.waitLoaded();
+
+        expect(dataController.dataSource.value.totalCount()).toEqual(3);
+      } catch (err) {
+        expect(false).toBeTruthy();
+      }
+    });
+
+    it('works with json url', async () => {
+      try {
+        const { dataController } = setup({
+          dataSource: 'https://js.devexpress.com/jQuery/Demos/WidgetsGallery/JSDemos/data/customers.json',
+          paging: {
+            pageSize: 3,
+          },
+        });
+
+        await dataController.waitLoaded();
+
+        expect(dataController.dataSource.value.totalCount()).toEqual(3);
+      } catch (err) {
+        expect(false).toBeTruthy();
+      }
     });
   });
 
--- a/packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
@@ -69,7 +69,7 @@ describe('DataController', () => {
   });
 
   describe('totalCount is not specified', () => {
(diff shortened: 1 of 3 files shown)


pomahtri:
Please note that we still need to support case when user provides us custom store where he doesn't calculate totalCount

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
```diff
@@ -67,6 +68,30 @@ describe('DataController', () => {
...
+      });
+
+      await dataController.waitLoaded();
+
+      expect(dataController.dataSource.value.totalCount()).toEqual(10);
+    });
+
+    it('works with json url', async () => {
+      const { dataController } = setup({
+        dataSource: 'https://js.devexpress.com/jQuery/Demos/WidgetsGallery/JSDemos/data/customers.json',
```
Raushen: Please try to import json by relative path.

pomahtri: Better to mock network request
P.S. I've committed some another test with such endpoint, if you could add `TODO` comment there it would be good

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
```diff
@@ -168,13 +168,17 @@ export class DataController {
...
-            e.extra = e.extra || {};
+            e.extra = isPlainObject(e.extra) ? e.extra : {};
 
             if (hasLocalPaging) {
               this._filteredItemCount.value = filteredData.length;
               e.take = take;
               e.skip = skip;
 
+              if (e.storeLoadOptions.requireTotalCount) {
+                e.extra.totalCount = e.data.length;
```
pomahtri: [can be postponed]
Can you please also refactor `_filteredItemCount`? Looks like we don't need it anymore at all

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
Tucchhaa: @pomahtri can you please explain why network request is better?

According to this line: https://github.com/DevExpress/DevExtreme/blob/0a0224d6b17f9ca7b6e3f1508bf1bcfe551b0eea/packages/devextreme/js/__internal/data/data_source/m_utils.ts#L87, it doesn't matter whether it's a url to the remote server or local file. 

It doesn't matter whether the .json file is located on remote server or in the file system.

pomahtri: I doubt that if you will specify local json file, it will load it. It is client-side code, it doesn't have access to local filesystem

Tucchhaa: Yep, just tested. It doesn't work. So let's keep current version

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
Tucchhaa: I debugged and actually we need `_filteredItemCount`.

It's used here: https://github.com/DevExpress/DevExtreme/blob/9f3c8e15d373b690814878d6b58fcdc592df8a1c/packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts#L329.

Without it pager will contain pages for all the loaded data, even if a local filter applied which reduces page count.

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
pomahtri: Anyways, please mock it

wdevfx: It's not good to perform the network request from the Jest environment to the remote resource

We may face network issues, e.g:
- This test will not execute properly without an internet connection
- If the remote server is down -> this test will be red

So, let's replace it with a local JSON file or mock this request

Tucchhaa: Maybe we can delete this test? Since if dataSource is a url, then CustomStore is created:
https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/data/data_source/m_utils.ts#L79

And there is already a test for CustomStore

pomahtri (approved):

PR #30053, opened 2025-06-04
DataGrid - Group row content is scrolled if repaintChangesOnly is enabled and the grid has a fixed column (T1286077)

tongsonbarbs:


e2e/testcafe-devextreme/tests/dataGrid/sticky/common/withGrouping.ts
```diff
@@ -239,3 +239,33 @@ safeSizeTest('The group separator should be visible when dragging a fixed column
...
+
+test('DataGrid - Group row content is scrolled if repaintChangesOnly is enabled and the grid has a fixed column (T1286077)', async (t) => {
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const dataGrid = new DataGrid('#container');
+  const groupPanelToggle = dataGrid.getGroupRow(0).getCell(0).element;
+
+  await dataGrid.isReady();
+
+  await t
+    .doubleClick(groupPanelToggle);
```
Alyar666: Double clicking cannot guarantee that the group has expanded and collapsed correctly. I advise you to separate these actions. And between these actions, it is better to check whether the group has expanded.

tongsonbarbs: @Alyar666 , PR updated, please resolve comment if it is all good :) 

Thanks

Alyar666 (approved):

PR #30061, opened 2025-06-04
CardView - fix memory leaks

pomahtri:
- forgotten unsubscribe in baseView
- forgotten fix from https://github.com/DevExpress/DevExtreme/pull/29667/files
   I have 0 ideas why it was reverted
- fix in overlay


packages/devextreme/js/__internal/ui/overlay/m_overlay.ts
```diff
@@ -781,6 +781,12 @@ class Overlay<
     }
   }
 
+  _destroyTabTerminator(): void {
```
marker-dao, line 784: Please call the _destroyTabTerminator method from _toggleTabTerminator to avoid duplicating functionality here: https://github.com/DevExpress/DevExtreme/pull/30061/files#diff-e6ec2df40e72e9a0136c4a328e07a35bd9787af623bb5cfd698b445a0c85568aR778 

Otherwise, I understand the problem and consider this to be an acceptable and understandable solution.

pomahtri: Fixed (if I understood correctly)
Please note that anyways there's duplication, `const eventName = ...` is calculated two times
Passing it in argument doesn't look good IMO

marker-dao: Agree

wdevfx (dismissed):

marker-dao (approved):

wdevfx (approved):

PR #30068, opened 2025-06-04
🚨 CardView: hotfix filter issues 

wdevfx:


Raushen (dismissed):

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
```diff
@@ -229,6 +229,7 @@ export class DataController {
 
     effect(
       () => {
+        const initialized = this.options.initialized.value;
```
pomahtri, line 232: [possible to leave for future]
let's implement lifecycle controller

wdevfx: Let's postpone it, I don't want to implement it in a rush

pomahtri (approved):

PR #30078, opened 2025-06-05
Stepper Demo: refactor jQuery demo

anna-shakhova:

### DIFF
--- a/apps/demos/Demos/Stepper/FormIntegration/jQuery/index.js
+++ b/apps/demos/Demos/Stepper/FormIntegration/jQuery/index.js
@@ -1,56 +1,44 @@
 $(() => {
-  const stepsCount = initialSteps.length;
   const validationGroups = ['dates', 'guests', 'roomAndMealPlan'];
   let confirmed = false;
   let formData = getInitialFormData();
 
-  const setSelectedIndex = (index) => {
-    stepper.option('selectedIndex', index);
-  };
+  const stepper = $('#stepper').dxStepper({
+    items: steps,
+    onSelectionChanged(e) {
+      const selectedIndex = e.component.option('selectedIndex');
 
-  const confirm = () => {
-    confirmed = true;
+      setSelectedIndex(selectedIndex);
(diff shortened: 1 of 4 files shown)


apps/demos/Demos/Stepper/FormIntegration/jQuery/styles.css
```diff
@@ -65,3 +65,7 @@
     display: flex;
     gap: 8px;
 }
+
+.readonly {
+    pointer-events: none;
+}
```
pharret31, line 71: Shouldn't this be a part of internal stepper style?

anna-shakhova: No, by default stepper allows step selection by click. To make stepper readonly we should set `focusStateEnabled` to `false` and disable pointer events.
In this demo stepper becomes readonly after order confirmation, so we can't disable it in stepper styles, instead we add to its element class on confirm

pharret31 (dismissed):

apps/demos/Demos/Stepper/FormIntegration/jQuery/index.js
```diff
@@ -71,197 +59,243 @@ $(() => {
...
   }).dxButton('instance');
 
-  const getValidationResult = (index) => {
+  function validateStep(index) {
+    const isValid = getValidationResult(index);
+
+    stepper.option(`items[${index}].isValid`, isValid);
+
+    return isValid;
+  }
```
helen-dikareva, line 80: (minor) we can place a line break here, but I doubt that it worths a test restarting

helen-dikareva (dismissed):

apps/demos/Demos/Stepper/FormIntegration/jQuery/index.js
anna-shakhova: fixed, minor but doesn't look nice

pharret31 (approved):

PR #30086, opened 2025-06-06
HtmlEditor: Fix type error when parsing list with non-string Delta insert (T1292587)

r-farkhutdinov:

### DIFF
--- a/packages/devextreme/js/__internal/ui/html_editor/matchers/m_wordLists.ts
+++ b/packages/devextreme/js/__internal/ui/html_editor/matchers/m_wordLists.ts
@@ -1,3 +1,5 @@
+import { isString } from '@js/core/utils/type';
+
 function getListType(matches) {
   const prefix = matches[1];
   return prefix.match(/\S+\./) ? 'ordered' : 'bullet';
@@ -21,10 +23,6 @@ function removeNewLineChar(operations) {
   newLineOperation.insert = newLineOperation.insert.trim();
 }
 
-function isTextOp(op): boolean {
-  return typeof op === 'string';
-}
-
 const getMatcher = (quill) => {
   const Delta = quill.import('delta');
   const msStyleAttributeName = quill.MS_LIST_DATA_KEY;
@@ -34,7 +32,7 @@ const getMatcher = (quill) => {
 
     const insertOperation = ops[0];
 
-    if (!isTextOp(insertOperation.insert)) {
+    if (!isString(insertOperation.insert)) {
       return delta;
     }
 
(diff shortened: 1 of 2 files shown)


packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/multilineIntegration.tests.js
```diff
@@ -65,6 +65,32 @@ export default function() {
...
+                test(`editor should not throw error for ${className} class with allowSoftLineBreak (T1292587)`, function(assert) {
+                    const spy = sinon.spy(console, 'error');
+                    const value = `<p class="${className}"><br><br></p>`;
+                    const $element = $('#htmlEditor');
+                    $element.dxHtmlEditor({
+                        value
+                    }).dxHtmlEditor('instance');
+
+                    assert.strictEqual(spy.called, false, 'No console error was thrown');
+                    spy.restore();
```
ksercs: [not isolated test]
Now after first fail spy will not be restored => next test will be automatically failed because console.error is already spied.
Please use try/catch/finally

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/multilineIntegration.tests.js
```diff
@@ -65,6 +65,32 @@ export default function() {
...
+                });
+
+                test(`editor should preserve double <br> in list paragraph for ${className} class (T1292587)`, function(assert) {
+                    const value = `<p class="${className}"><br><br></p>`;
+                    const $element = $('#htmlEditor');
+                    $element.dxHtmlEditor({
+                        value
+                    }).dxHtmlEditor('instance');
+                    const markup = $element.find(`.${CONTENT_CLASS}`).html();
```
ksercs: Isn't it the same if you just get `htmlEditor.option('value')`? Then it'll be clearer and you don't need the `$element` variable (actually you also don't need it in the test above)

packages/devextreme/js/__internal/ui/html_editor/matchers/m_wordLists.ts
```diff
@@ -21,6 +21,10 @@ function removeNewLineChar(operations) {
   newLineOperation.insert = newLineOperation.insert.trim();
 }
 
+function isTextOp(op): boolean {
```
ksercs:
```suggestion
function isTextOperation(op): boolean {
```
But maybe we don't need this util at all? Actually even an object insert can be a "text operation" .. You basically check only that operation.insert is a string, maybe it's better to state it explicitly

ksercs (approved):

marker-dao (approved):

PR #30087, opened 2025-06-06
ContextMenu: allow using multiple context menus if target is document or undefined (T1295721)

pharret31:


packages/devextreme/testing/tests/DevExpress.ui.widgets/contextMenu.tests.js
```diff
@@ -22,6 +22,7 @@ import 'generic_light.css!';
 QUnit.testStart(() => {
     const markup =
         '<div id="simpleMenu"></div>\
+        <div id="simpleMenu2"></div>\
```
ksercs: It's better to append necessary element to qunit-fixture in your particular test then add it to a base page markup while it's used in the single test only

packages/devextreme/testing/tests/DevExpress.ui.widgets/contextMenu.tests.js
```diff
@@ -2185,6 +2186,36 @@ QUnit.module('Behavior', moduleConfig, () => {
...
+        const documentMenuItemText = 'document menu item 1';
+        new ContextMenu($('#simpleMenu2'), {
+            items: [{ text: documentMenuItemText }],
+        });
+
+        $('#menuTarget').trigger('dxcontextmenu');
+
+        assert.strictEqual(
+            $(`.${DX_MENU_ITEM_CONTENT_CLASS}:contains(${simpleMenuItemText})`).length > 0,
+            true,
```
ksercs: maybe then check that length is strictly equal to 1?

packages/devextreme/js/__internal/ui/context_menu/m_context_menu.ts
```diff
@@ -483,13 +488,19 @@ class ContextMenu extends MenuBase {
...
     } else {
       this._showContextMenuEventHandler = handler;
-      eventsEngine.on(domAdapter.getDocument(), eventName, target, this._showContextMenuEventHandler);
+      eventsEngine.on(
+        domAdapter.getDocument(),
+        eventName,
+        target,
+        this._showContextMenuEventHandler,
+      );
     }
```
ksercs, line 525: If we replace this if-else with only
```javascript
    this._showContextMenuEventHandler = handler;
    eventsEngine.on(
      domAdapter.getDocument(),
      eventName,
      target,
      this._showContextMenuEventHandler,
    );
```
all tests gonna be green. That's because `eventsEngine.off($(target), eventName);` in `_detachShowContextMenuEvents` gonna never be called - we don't unsubscribe all events manually anymore but do it for a particular handler only.

I think it's better to refactor and simplify the code than add a new solution over the old questionable code if it's possible

ksercs: P.s. This code change is not a final proposition but only demonstration of the idea. It's clear it's still better to attach handler directly when target=node to at least make it working with `stopPropagation`

packages/devextreme/testing/tests/DevExpress.ui.widgets/contextMenu.tests.js
pharret31: I thought that it's clearer to use true/false in test results but sure, we can check if it's 1

packages/devextreme/testing/tests/DevExpress.ui.widgets/contextMenu.tests.js
```diff
@@ -1415,40 +1415,26 @@ QUnit.module('Options', moduleConfig, () => {
...
+    [
+        { label: 'selector string', target: () => '#menuTarget' },
+        { label: 'jQuery object', target: () => $('#menuTarget') },
+        { label: 'DOM element', target: () => document.getElementById('menuTarget') },
+        { label: 'document', target: () => document },
+        { label: 'window', target: () => window },
+        { label: 'undefined', target: () => undefined },
+        { label: 'null', target: () => null },
+    ].forEach(({ label, target }) => {
+        QUnit.test(`target option as ${label}`, function(assert) {
```
ksercs: please name the test according to our style guide. What is tested here? "event should be correctly attached if target=${target}"?

packages/devextreme/testing/tests/DevExpress.ui.widgets/contextMenu.tests.js
```diff
@@ -2185,6 +2171,53 @@ QUnit.module('Behavior', moduleConfig, () => {
...
+        $(document).trigger('dxcontextmenu');
+
+        assert.strictEqual(
+            $(`.${DX_MENU_ITEM_CONTENT_CLASS}:contains(${documentMenuItemText})`).length,
+            1,
+            'document menu was shown'
+        );
+    });
+
+    QUnit.test('it should be possible to show menu if stopPropagation on the same event was called', function(assert) {
```
ksercs: if target is a node

packages/devextreme/testing/tests/DevExpress.ui.widgets/contextMenu.tests.js
```diff
@@ -2185,6 +2171,53 @@ QUnit.module('Behavior', moduleConfig, () => {
             assert.ok(false);
         }
     });
+
+    QUnit.test('multiple context menu if one has no target', function(assert) {
```
ksercs: should be correctly displayed

packages/devextreme/js/__internal/ui/context_menu/m_context_menu.ts
```diff
@@ -480,16 +487,24 @@ class ContextMenu extends MenuBase {
       }
     }, { validatingTargetName: 'target' });
 
-    const handler = (e) => contextMenuAction({ event: e, target: $(e.currentTarget) });
+    this._showContextMenuEventHandler = (e): void => contextMenuAction({
+      event: e,
+      target: $(e.currentTarget),
+    });
 
     contextMenuAction = this._createAction(contextMenuAction);
```
ksercs: what is this for? looks useless

packages/devextreme/js/__internal/ui/context_menu/m_context_menu.ts
```diff
@@ -467,7 +473,8 @@ class ContextMenu extends MenuBase {
       return;
     }
 
-    const eventName = addNamespace(showEvent, this.NAME);
+    const eventName = addNamespace(showEvent, this._eventNamespace);
+
     let contextMenuAction = this._createAction((e) => {
```
ksercs: do we still need this additional wrap? I don't see how passed target is used

packages/devextreme/js/__internal/ui/context_menu/m_context_menu.ts
```diff
@@ -454,42 +481,32 @@ class ContextMenu extends MenuBase {
...
-  _attachShowContextMenuEvents() {
-    const target = this._getTarget();
+  _attachShowContextMenuEvents(): void {
     // @ts-expect-error
-    const showEvent = this.getShowEvent(this.option('showEvent'));
+    const { showEvent: showEventOption, disabled } = this.option();
+    const showEvent = this.getShowEvent(showEventOption);
 
-    if (!showEvent) {
+    if (!showEvent || disabled) {
```
ksercs, line 508: Is this change intentional? Previously contextMenu were shown even if disabled=true and it sounds logical while disabled=true should only prevent user interactions with a component.
Did you introduce this change by analogy with a previous code in tooltip? It looks like a BC

packages/devextreme/js/__internal/ui/context_menu/m_context_menu.ts
```diff
@@ -384,6 +388,7 @@ class ContextMenu extends MenuBase {
   }
 
   _renderContentImpl(): void {
+    this._showContextMenuEventHandler = this._createShowContextMenuEventHandler();
```
ksercs: Hm, looks like events will not be unsubscribed now because you recreate the handler before detach => link is already different when we try to off.
p.s. don't forget to add a test

packages/devextreme/js/__internal/ui/context_menu/m_context_menu.ts
```diff
@@ -968,13 +985,16 @@ class ContextMenu extends MenuBase {
...
         break;
+      case 'disabled':
       case 'showEvent':
       case 'position':
       case 'submenuDirection':
         this._invalidate();
         break;
       case 'target':
-        args.previousValue && this._detachShowContextMenuEvents(args.previousValue);
+        if (args.previousValue) {
```
EugeniyKiyashko: const { name, value, previousValue } = args;

Also kindly add typing to _optionChanged() method

packages/devextreme/js/__internal/ui/context_menu/m_context_menu.ts
```diff
@@ -435,17 +440,39 @@ class ContextMenu extends MenuBase {
     }
   }
 
+  _createShowContextMenuEventHandler() {
```
EugeniyKiyashko: Let's add types here and at least in all places that you change

packages/devextreme/js/__internal/ui/context_menu/m_context_menu.ts
```diff
@@ -75,7 +75,9 @@ class ContextMenu extends MenuBase {
 
   _overlayContentId?: string;
 
-  _showContextMenuEventHandler?: (event: unknown) => any;
+  _eventNamespace!: string;
+
+  _showContextMenuEventHandler?: (event: any) => any;
```
EugeniyKiyashko: We mustn't use 'any' type

packages/devextreme/js/__internal/ui/context_menu/m_context_menu.ts
pharret31: Actually it didn't work if disabled=true, as far as I understood the recreation of action (`this.showContextMenuAction = this._createAction(this.showContextMenuAction)`) in addition to _createAction with `validatingTargetName` was done in order to implement such behavior so the ui interactor was able to check if the element or the context menu itself are disabled it won't show the context menu, and because of that recreation wasn't obvious for me and for you I added a solution like we have in tooltip to implement the same behavior as it had

ksercs: Thanks for the explanation. Yeah, now I see ContextMenu previously also was not shown if disabled=true

ksercs (approved):

PR #30093, opened 2025-06-09
Polar Chart Spider Web demo: preload font for material theme

anna-shakhova:

### DIFF
--- a/apps/demos/Demos/Charts/SpiderWeb/Angular/index.html
+++ b/apps/demos/Demos/Charts/SpiderWeb/Angular/index.html
@@ -5,6 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=5.0" />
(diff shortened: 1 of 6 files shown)


apps/demos/Demos/Charts/SpiderWeb/Angular/index.html
```diff
@@ -5,6 +5,8 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=5.0" />
+    <link rel="preload" href="https://fonts.gstatic.com/s/roboto/v48/KFO7CnqEu92Fr1ME7kSn66aGLdTylUAMa3yUBHMdazQ.woff2" as="font" type="font/woff2" crossorigin />
```
EugeniyKiyashko, line 8: I assume we can simplify the links and use our project's built-in Roboto font. Something like this should be sufficient

```
<link rel="preload" href="css/fonts/Roboto-500.woff2" as="font" type="font/woff2" crossorigin />
```

anna-shakhova: It is safer to use same link Roboto css file imports

EugeniyKiyashko (approved):

PR #30098, opened 2025-06-09
DataGrid: Fix the focus recovery issue when ungrouping the column via context menu after leaving the page and returning back

Alyar666:


e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/groupColumnReordering.functional.ts
```diff
@@ -8,6 +9,17 @@ fixture
 
 const DATA_GRID_SELECTOR = '#container';
 
+const triggerVisibilityChange = ClientFunction((isVisible) => {
+  Object.defineProperty(document, 'visibilityState', {
```
pomahtri: Are we sure that it is restored ok? Looks like previously it was getter, now it always return `true` after this test. I doubt that we currently have tests that rely on that, but if we add them in future, we'll be suprised

pomahtri: Or maybe we have API in testcafe that switches between browser tabs natively

pomahtri (approved):

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/groupColumnReordering.functional.ts
pomahtri: Discussed, testcafe doesn't have such API

PR #30116, opened 2025-06-11
Charts: remove deprecated `shift` value from valueAxis.visualRangeUpdateMode

anna-shakhova:


packages/devextreme/js/common/charts.d.ts
```diff
@@ -271,6 +271,12 @@ export type VisualRange = {
  */
 export type VisualRangeUpdateMode = 'auto' | 'keep' | 'reset' | 'shift';
```
pharret31, line 272: Can we rename this to ArgumentAxisVisualRangeUpdateMode?

anna-shakhova: I prefer not to rename type if we haven't changed it. @DevExpress/devextreme-apireviewers, what would you say?

marker-dao: Let's leave it like as it is

ksercs (approved):

marker-dao (approved):

PR #30118, opened 2025-06-11
DropDownButton: Remove content template from Popup integrationOptions (T1294338)

marker-dao:


r-farkhutdinov (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets/dropDownButton.markup.tests.js
```diff
@@ -1095,3 +1095,28 @@ QUnit.module('option change', {}, () => {
         assert.strictEqual(e.value, handler, 'changed option new value is correct');
     });
 });
+
+QUnit.module('content template', () => {
+    QUnit.test('content template render function should be called once (T1294338)', function(assert) {
+        if(!windowUtils.hasWindow()) {
+            assert.ok(true, 'this test is not related to the server side');
```
ksercs: Let's maybe then just place it in regular tests, not here in `.markup`? There is even a module for templates testing :)

packages/devextreme/js/__internal/ui/m_drop_down_button.ts
```diff
@@ -412,6 +414,11 @@ class DropDownButton extends Widget<DropDownButtonProperties> {
         at: `${horizontalAlignment} bottom`,
       },
       _wrapperClassExternal: DROP_DOWN_EDITOR_OVERLAY_CLASS,
+      integrationOptions: {
+        templates: {
+          content: undefined,
```
ksercs: If the reason is "default template name is the same for component and overlay", isn't then this issue actual for another components? Particularly, dropDownBox

packages/devextreme/js/__internal/ui/m_drop_down_button.ts
```diff
@@ -412,6 +414,11 @@ class DropDownButton extends Widget<DropDownButtonProperties> {
         at: `${horizontalAlignment} bottom`,
       },
       _wrapperClassExternal: DROP_DOWN_EDITOR_OVERLAY_CLASS,
+      integrationOptions: {
```
ksercs: If we now have 2 places where template is rendered, isn't it more logical to remove this rendering in ddButton and completely delegate it to Overlay, not vice versa (reimplement in ddButton and restrict behavior on nested component)?

packages/devextreme/testing/tests/DevExpress.ui.widgets/dropDownButton.markup.tests.js
marker-dao: I put this here precisely because it is related to markup rendering. But I have no objection to moving it to the regular tests: [3260f01ca73c2718b6a2ee9d4a0e4a33f6b4268c].

packages/devextreme/js/__internal/ui/m_drop_down_button.ts
marker-dao: No, I already checked that this does not apply to other DropDowns which have content template (only DDBox has) and Popover.

packages/devextreme/js/__internal/ui/m_drop_down_button.ts
marker-dao: I think it will be expensive to rework because:
1. DDButton can use a named template, in which case we will have to make sure that the template name is passed to Overlay;
2. By default, List is rendered there (this mechanism will need to be completely transferred and may be more difficult to manage).

Especially since we are not talking about a traditional rendering method, but rather an extreme case. I believe that the changes are not worth it, and technically I do not see a strong need for them.

The idea is still to completely control the rendering from DropDownButton – this does not contradict the principles.

packages/devextreme/js/__internal/ui/m_drop_down_button.ts
ksercs: Doesn't it mean that ddBox does not render its template but completely delegate to popup? Or why the it's not reproduced there?

marker-dao: ddBox render by itself

packages/devextreme/js/__internal/ui/m_drop_down_button.ts
marker-dao: We decided to replace integrationOptions.templates.content with contentTemplate: null.

ksercs (dismissed):

ksercs (approved):

PR #30130, opened 2025-06-12
fix: fixed shadow dow performance

sergey-kras:


packages/devextreme/testing/tests/DevExpress.core/utils.shadow_dom.tests.js
```diff
@@ -0,0 +1,75 @@
+import { computeStyleSheetsHash, addShadowDomStyles } from '__internal/core/utils/shadow_dom';
```
chaosmirage: https://github.com/DevExpress/DevExtreme/actions/runs/16144761876/job/45622258649?pr=30130

`__internal/core/utils/shadow_dom` => `__internal/core/utils/m_shadow_dom`

`m_` is missed

packages/devextreme/testing/tests/DevExpress.core/utils.shadow_dom.tests.js
```diff
@@ -0,0 +1,75 @@
...
+        await styleSheet.replace(':host { background: red; }');
+        shadow.adoptedStyleSheets = [styleSheet];
+
+        const div = document.createElement('div');
+        shadow.appendChild(div);
+
+        const $div = $(div);
+        addShadowDomStyles($div);
+
+        assert.equal(shadow.adoptedStyleSheets.length, 2, 'Two stylesheets were adopted (global + shadow)');
```
chaosmirage: The `adoptedStyleSheets` property always contains 2 elements even they are empty (contains 0 css rules), because of

```
root.adoptedStyleSheets = [ownerDocumentStyleSheet, currentShadowDomStyleSheet]
```

`packages/devextreme/js/__internal/core/utils/m_shadow_dom.ts`

Style sheets are processed from the `styleSheets` property by `addShadowDomStyles` function, but in the test, it is set to `adoptedStyleSheets`.

```
shadow.adoptedStyleSheets = [styleSheet];
```

So, style sheets in the `shadow.adoptedStyleSheet` property don't contain CSS rules.

How can it be checked correctly (not optimized, just for proof of concept):

```
    QUnit.test('Applies styles to ShadowRoot via adoptedStyleSheets', function(assert) {
        const container = document.createElement('div');
        document.body.appendChild(container);

        const hostStyleElement = document.createElement('style');
        hostStyleElement.innerHTML = '.dx-widget-host { background: red; }';
        document.head.appendChild(hostStyleElement);

        const shadow = container.attachShadow({ mode: 'open' });

        const shadowStyleElement = shadow.ownerDocument.createElement('style');
        shadowStyleElement.innerHTML = '.dx-widget-shadow { background: red; }';
        shadow.appendChild(shadowStyleElement);

        const div = document.createElement('div');
        shadow.appendChild(div);

        const $div = $(div);
        addShadowDomStyles($div);

        assert.equal(shadow.adoptedStyleSheets[0].cssRules[0].cssText, '.dx-widget-host { background: red; }');
        assert.equal(shadow.adoptedStyleSheets[1].cssRules[0].cssText, '.dx-widget-shadow { background: red; }');
    });
```

packages/devextreme/testing/tests/DevExpress.core/utils.shadow_dom.tests.js
```diff
@@ -0,0 +1,75 @@
...
+        shadow.appendChild(div);
+        const $div = $(div);
+
+        addShadowDomStyles($div);
+        const firstCount = shadow.adoptedStyleSheets.length;
+
+        addShadowDomStyles($div);
+        const secondCount = shadow.adoptedStyleSheets.length;
+
+        assert.equal(firstCount, secondCount, 'Stylesheets count unchanged after repeated call');
```
chaosmirage: The `shadow.adoptedStyleSheets` array always contains 2 elements because of 

```
root.adoptedStyleSheets = [ownerDocumentStyleSheet, currentShadowDomStyleSheet]
```
`packages/devextreme/js/__internal/core/utils/m_shadow_dom.ts`

To achieve the goal content of css rules should be compared.

packages/devextreme/js/__internal/core/utils/m_shadow_dom.ts
```diff
@@ -43,22 +44,69 @@ function insertRule(targetStyleSheet, rule, needApplyAllStyles) {
...
+  if (!state) {
+    state = {
+      injectedGlobal: false,
+      lastLocalHash: null,
+    };
+    injectedRootStates.set(root, state);
   }
 
-  if (!ownerDocumentStyleSheet) {
+  if (!ownerDocumentStyleSheet && !state.injectedGlobal) {
```
chaosmirage: As I understood, the `state.injectedGlobal` is not used, because the first `addShadowDomStyles` execution, `ownerDocumentStyleSheet` becomes truthy.

This means `!state.injectedGlobal` will never be re-evaluated.

I don't see any case when `ownerDocumentStyleSheet` is **not** set and `state.injectedGlobal` is set because `state.injectedGlobal` is changed only inside the condition body. 

I suppose the goal was to avoid unnecessary style injection from the host if a shadow DOM already contains injected styles during rendering a component inside the shadow DOM, and as I can see, this goal is already achieved by

`if (state.lastLocalHash === localHash) return;`

The suggestion is to remove the `injectedGlobal` property.

packages/devextreme/js/__internal/core/utils/m_shadow_dom.ts
```diff
@@ -43,22 +44,69 @@ function insertRule(targetStyleSheet, rule, needApplyAllStyles) {
   }
 }
 
+const sheetHashes = new WeakMap();
+export function computeStyleSheetsHash(styleSheets) {
+  let hash = 2166136261;
```
chaosmirage: Please extract the value to a variable, mentioning the algorithm name to identify it in the future, like

`FNV_OFFSET_BASIS`

chaosmirage (changes requested):

packages/devextreme/js/__internal/core/utils/m_shadow_dom.ts
```diff
@@ -43,22 +44,69 @@ function insertRule(targetStyleSheet, rule, needApplyAllStyles) {
...
+    hash ^= localHash;
+  }
+
+  return hash >>> 0;
+}
+
+const injectedRootStates = new WeakMap();
+
 export function addShadowDomStyles($element) {
+  if (config().noCopyStylesToShadowDom) {
```
chaosmirage: I see that in the config, it is called `copyStylesToShadowDom` (without `no` prefix).

Please also cover this functionality with a test.

packages/devextreme/testing/tests/DevExpress.core/utils.shadow_dom.tests.js
chaosmirage: I don't see the `Applies styles to ShadowRoot via adoptedStyleSheets` test. Why was it removed?

As I remember, we discussed moving it somewhere to the testcafe tests to avoid issues related to QUnit CI environment.  

For example like tests for events were created:
https://github.com/chaosmirage/DevExtreme/blob/state-manager/e2e/testcafe-devextreme/tests/common/eventsEngine.ts

e2e/wrappers/package.json
```diff
@@ -44,6 +44,7 @@
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
     "react-router-dom": "^6.22.0",
+    "react-shadow": "^20.6.0",
```
chaosmirage, line 47: It seems like this change and others inside `e2e/wrappers` were added for some experiments and are no longer needed.

packages/devextreme/js/__internal/core/utils/m_shadow_dom.ts
chaosmirage: I don't see the test. Could you please clarify why?

I found an example here: https://github.com/chaosmirage/DevExtreme/blob/state-manager/packages/devextreme/testing/tests/DevExpress.core/component.tests.js#L1710

PR #30138, opened 2025-06-13
fix: double event fire in React wrappers (T1293563)

VasilyStrelyaev:


packages/devextreme-react/src/core/__tests__/props-updating.test.tsx
```diff
@@ -250,35 +250,37 @@ describe('option update', () => {
     expect(Widget.option.mock.calls.length).toBe(1);
     expect(Widget.option.mock.calls[0]).toEqual(['items[0].subItems[0].a', 234]);
   });
-});
 
-it('updates sub-nested collection item within a custom component', () => {
-  const MySetting = (props: any) => {
-    const { value } = props;
+  it('updates sub-nested collection item within a custom component', () => {
```
VasilyStrelyaev, line 254: This test is just brought under `describe()` because it was outside and thus unaffected by clenaup hooks.

packages/devextreme-react/src/core/options-manager.ts
```diff
@@ -257,7 +266,13 @@ class OptionsManager {
 
   private setValue(name: string, value: unknown) {
     if (this.guards[name]) {
-      delete this.guards[name];
+      try {
+        if (compareValues(this.guards[name].latestValue, value)) {
```
chaosmirage, line 270: [optional]


Maybe it is possible to reflect the reason of this condition like

```
const hasWidgetAlreadySetValue = compareValues(this.guards[name].latestValue, value);

if (hasWidgetAlreadySetValue) {
  return;
}
```

chaosmirage (approved):

chaosmirage (approved):

dxvladislavvolkov (approved):

PR #30140, opened 2025-06-16
CardView - demos - design fixes

pomahtri:


apps/demos/Demos/CardView/DataValidation/Angular/app/app.component.html
```diff
@@ -14,8 +10,58 @@
...
+      <dxo-card-view-form-item
+        caption="Personal Data"
+        itemType="group"
+        [colCount]="2"
+        [colSpan]="2"
+      >
+        <dxo-card-view-form-item
+          dataField="firstName"
+        ></dxo-card-view-form-item>
+        <dxo-card-view-form-item dataField="lastName"></dxo-card-view-form-item>
```
Raushen, line 29: There is a difference between this column and the previous. Maybe we can leave something as the following:
```
<dxo-card-view-form-item dataField="firstName"/>
```

apps/demos/Demos/CardView/DataValidation/Angular/app/app.component.html
```diff
@@ -14,8 +10,58 @@
...
+      >
+        <dxo-card-view-form-item
+          dataField="address"
+          [colSpan]="2"
+        ></dxo-card-view-form-item>
+        <dxo-card-view-form-item dataField="city"></dxo-card-view-form-item>
+        <dxo-card-view-form-item dataField="zipcode"></dxo-card-view-form-item>
+        <dxo-card-view-form-item
+          dataField="mobilePhone"
+        ></dxo-card-view-form-item>
```
Raushen, line 65: Here is the different format too

apps/demos/Demos/CardView/DataValidation/Angular/app/app.component.html
pomahtri: Unfortunately we cannot, this is prettier

apps/demos/Demos/CardView/DataValidation/Angular/app/app.component.html
pomahtri: Unfortunately we cannot, this is prettier

Raushen (approved):

PR #30141, opened 2025-06-16
List: prevent radio button from remaining active after double-click if another radio button is selected (T1294715)

pharret31:


packages/devextreme/testing/tests/DevExpress.ui.widgets/listParts/editingUITests.js
```diff
@@ -2595,6 +2595,31 @@ QUnit.module('item select decorator with single selection mode', () => {
         assert.strictEqual(radioButton.option('value'), true, 'item selected');
     });
 
+    QUnit.test('only one item should remain selected after double-clicking on its radio control and then clicking on another item control (T1294715)', function(assert) {
+        const $list = $('#templated-list').dxList({
```
ksercs: [minor]
```suggestion 
        const $list = $('#list').dxList({
```

packages/devextreme/js/__internal/ui/list/m_list.edit.decorator.selection.ts
```diff
@@ -74,6 +76,9 @@ class EditDecoratorSelection extends EditDecorator {
         const isUiClick = !!event;
         if (isUiClick) {
           component._valueChangeEventInstance = undefined;
+          if (this._controlClass === SELECT_RADIO_BUTTON_CLASS && value === false) {
```
ksercs: I see why it works but the code looks to specific to me.
Isn't it better to restore `_valueChangeEventInstance` in RadioButton's level and value change call?
Then we'll not have this specific condition, we'll be sure it'd be not broken in another places and radiobutton's algorithm will be more stable itself

ksercs (approved):

PR #30146, opened 2025-06-16
Button: refactor r1 generated code

anna-shakhova:


packages/devextreme/js/__internal/ui/button/button.tsx
```diff
@@ -44,9 +71,50 @@ const getCssClasses = (model): string => {
...
+  const excludedSet = new Set(excludedKeys);
+
+  return Object.keys(obj).reduce((result, key) => {
+    if (!excludedSet.has(key as K)) {
+      result[key] = obj[key];
+    }
+
+    return result;
+  }, {}) as Omit<T, K>;
+};
```
ksercs, line 91: looks like a good candidate for moving to an utility :)

packages/devextreme/js/__internal/ui/button/button.tsx
```diff
@@ -44,9 +71,50 @@ const getCssClasses = (model): string => {
...
+
+export type ButtonProps = Record<string, unknown> & BaseWidgetProps & ButtonProperties & {
```
ksercs, line 93: hm, so actually this `ButtonProps` type is `Record<string, unknown>`, right? Because it's the most wide from this union. Is it necessary?

packages/devextreme/js/__internal/ui/button/button.tsx
```diff
@@ -218,80 +268,62 @@ export class Button extends InfernoWrapperComponent<ButtonProps> {
...
   }
 
   activate(): void {
-    this.widgetRef.current!.activate();
+    this.widgetRef.current?.activate();
   }
 
   deactivate(): void {
-    this.widgetRef.current!.deactivate();
+    this.widgetRef.current?.deactivate();
```
ksercs, line 311: How can it be possible that we don't have widgetRef?
Intuitively nullish assertions looks more logical here

packages/devextreme/js/__internal/ui/button/props.ts
```diff
@@ -1,50 +0,0 @@
-import type { BaseWidgetProps } from '@ts/core/r1/base_props';
```
ksercs, line 1: Why don't you like the idea to move props to a separate file?)

EugeniyKiyashko (approved):

packages/devextreme/js/__internal/ui/button/button.tsx
anna-shakhova: placed here just because this utility is used in couple legacy Inferno components, if we'll see it required somewhere else - then, it should be moved

packages/devextreme/js/__internal/ui/button/button.tsx
anna-shakhova: it is typing issue in R1 widget, it requires this Record type, so to meet ButtonProps with common Properties some extra code is required.
Also you may see, that I used type here, cause we can't use interface here due to conflicts

packages/devextreme/js/__internal/ui/button/button.tsx
anna-shakhova: it is impossible, but TS compiler not as smart as we are, and we have to add such checks in our code, it is better than say TS to ignore such cases, as it was made before

packages/devextreme/js/__internal/ui/button/props.ts
anna-shakhova: Just because we have another 2 small components in button, and we should have same approach for all components. It was easier to combine props with button, that create 2 more files with props

packages/devextreme/js/__internal/ui/button/button.tsx
ksercs: no problem

PR #30152, opened 2025-06-16
CardView - filter sync - fix

pomahtri:


pomahtri:
went into https://github.com/DevExpress/DevExtreme/pull/30154

PR #30167, opened 2025-06-17
CardView: cardHeader.visible property does not update in runtime

Raushen:


packages/devextreme/js/__internal/grids/new/grid_core/utils/tree/tree.ts
```diff
@@ -129,12 +129,13 @@ export const mergeOptionTrees = (
   pathToMerge: string[],
 ): TreeNodeType => {
   const [lastNodePath] = pathToMerge.slice(-1);
```
wdevfx, line 131: Let's do it simpler

The current solution works the following way:
- We do a shallow copy of the tree path, if path doesn't exist we stop shallow copy
- We try to find the path's parent node
- If the path's parent node doesn't exist -> we create passed path in the tree
- We try to find the path's parent node one more time

But we can do it in just two steps:
- Let's shallow copy or create a passed path (```shallowCopySubtreePath``` func)
- Try to find the path's parent node

So, we can just replace this line in the source code:
```ts
const result = shallowCopySubtreePath(internalTree, pathToMerge);
```

To
```ts
const result = createOrShallowCopySubtreePath(internalTree, pathToMerge);
```

And everything should work as expected :)

packages/devextreme/js/__internal/grids/new/grid_core/utils/tree/tree.test.ts
```diff
@@ -120,7 +120,7 @@ describe('OptionsController', () => {
     });
 
     describe('mergeOptionTrees', () => {
-      it('should only shallow copy root if path contains error at first place', () => {
+      it('should create absent nodes if path contains error at first place', () => {
```
wdevfx, line 123: Let's add at least one more test case:
- tree doesn't have node | updatedTree has node | defaultTree doesn't have node

This case will be the closest to the issue that we are fixing here

Now we checking that passed path created, but don't check that value from ```updatedTree``` copied to the result tree

e2e/testcafe-devextreme/tests/cardView/common/behavior.functional.ts
```diff
@@ -0,0 +1,22 @@
+import CardView from 'devextreme-testcafe-models/cardView';
```
wdevfx, line 1: [Minor comment]

Now we have a mess where to write tests and don't have clear conventions
But in terms of API, it's better to use jest

Because it works 99% the same way in Jest and TestCafe
(In Jest it's ok to check markup that header exist too)

But for now it's ok leave it as is

wdevfx (approved):

PR #30168, opened 2025-06-17
Tabs: add resize observer to reload scrollable on internal item size changes

pharret31:


packages/devextreme/js/__internal/ui/tabs/tabs.ts
```diff
@@ -246,6 +247,11 @@ class Tabs extends CollectionWidget<TabsProperties> {
     this._feedbackHideTimeout = FEEDBACK_HIDE_TIMEOUT;
   }
 
+  _renderContent(): void {
+    super._renderContent();
+    this._observeContentResize(true);
```
EugeniyKiyashko: What scenarios covers by argument in _observeContentResize()? Why is it require?

packages/devextreme/js/__internal/ui/tabs/tabs.ts
```diff
@@ -633,6 +639,18 @@ class Tabs extends CollectionWidget<TabsProperties> {
     }
   }
 
+  _observeContentResize(shouldObserve: boolean): void {
+    if (!this.option('useResizeObserver')) {
+      return;
+    }
+
+    if (shouldObserve) {
+      resizeObserverSingleton.observe(this.$element().get(0), () => { this._dimensionChanged(); });
```
EugeniyKiyashko: There is no test scenario that explains why this ResizeObserver was added

packages/devextreme/js/__internal/ui/tabs/tabs.ts
```diff
@@ -659,6 +677,7 @@ class Tabs extends CollectionWidget<TabsProperties> {
   _clean(): void {
     this._cleanScrolling();
     super._clean();
+    this._observeContentResize(false);
```
EugeniyKiyashko: It is unclear how this part works

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
```diff
@@ -405,6 +408,23 @@ QUnit.module('General', () => {
         assert.strictEqual($element.find(`.${TABS_NAV_BUTTON_CLASS}`).length, 2, 'nav buttons was rendered');
         assert.strictEqual($element.find(`.${TABS_SCROLLABLE_CLASS}`).length, 1, 'scrollable was rendered');
     });
+
+    QUnit.test('resize observer should be connected to the tabs', function(assert) {
+        if(devices.real().platform === 'ios' && compareVersions(devices.real().version, '13.3') <= 0) {
```
EugeniyKiyashko: Was this condition added for jobs that run tests with emulated devices? Why can’t we use the isDeviceDesktop() utility instead?

packages/devextreme/js/__internal/ui/tabs/tabs.ts
pharret31: I wasn't able to test the actual behavior because in qunit environment it renders the dxScrollable in any case even if there is enough space for tabs and I wasn't able to fix that with changing #qunit-fixture/#tabs position and/or size so the test case that tests the ResizeObserver is connected to the element was added.

packages/devextreme/js/__internal/ui/tabs/tabs.ts
pharret31: The argument is needed because we have to unobserve appropriate elements on component dispose to keep the observers map clean.

packages/devextreme/js/__internal/ui/tabs/tabs.ts
pharret31: We have to remove the observer on component clean

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
pharret31: pipeline is using ios10 to test, that's why I added a condition about ios versions - ResizeObserver doesn't exist in ios <= 13.3

packages/devextreme/js/__internal/ui/tabs/tabs.ts
ksercs: Now scrollable is always rendered because this RO callback is always initially triggered after `observe`

packages/devextreme/testing/tests/DevExpress.ui.widgets/tabs.tests.js
ksercs: if we need it it's simpler to check that `window.ResizeObserver` is presented in the environment

EugeniyKiyashko: This likely should be addressed as follows:

1. The widgets component already has defaultOptions that cover iOS versions we no longer support.
2. The CI job is running tests with an emulated, outdated, and unnecessary version of iOS - we probably need to update the user agent to a more recent version.

I assume these tasks(suggests) to improve our work flow can be formulated to avoid unnecessary code, at least in tests.

packages/devextreme/js/__internal/ui/tabs/tabs.ts
pharret31: The same behavior happened without these changes as well - other tests were updated

EugeniyKiyashko (approved):

PR #30169, opened 2025-06-17
CardView - ContextMenu position is wrong after opening from the keyboard

Raushen:


packages/devextreme/js/__internal/grids/new/card_view/header_panel/item.tsx
```diff
@@ -130,4 +130,12 @@ export class Item extends Component<ItemProps> {
       this.props.onFilterClick?.(this.props.elementRef.current);
     }
   };
+
+  private readonly onContextMenuHandler = (event: MouseEvent): void => {
+    event.stopPropagation();
```
wdevfx: Question.

Why do we need ```stopPropagation``` here?

Raushen: There is not an exact purpose, just not to produce additional events on parent elements as we have handled it already.

wdevfx: Discussed
Now we have conventions about events that all events should bubble
Because our users may have global event listeners for their purposes 

E.g: open sidebar after ```contextmenu``` event anywhere on the page (including whole CardView)

So, we decided to remove this ```stopPropagation```

wdevfx (approved):

PR #30190, opened 2025-06-19
DataGrid: Fix quick navigation with Ctrl+End when virtual and simulated scrolling is enabled

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/virtual_columns/m_virtual_columns.ts
wdevfx, line 1: Where are the tests for the fixed issue? :)

wdevfx: Missed that tests already exist, but they were unstable (floating issue)
This fix makes existing tests stable

pomahtri (approved):

PR #30192, opened 2025-06-19
DataGrid: Fix quick navigation when adaptive is enabled

Alyar666:


e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.visual.ts
```diff
@@ -262,6 +262,69 @@ test('Navigate to first cell in the row that contains focus when virtual columns
...
+  // arrange
+  const dataGrid = new DataGrid('#container');
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+
+  await dataGrid.scrollTo(t, { x: 4200 });
+
+  // assert
+  await t
+    .expect(dataGrid.getScrollLeft())
+    .eql(4200);
```
wdevfx, line 303: Question.

Is it safe to compare a scroll value with an exact value?
Can we receive here sth like ```4200.01```?

Because if yes, this test may became unstable and be our pain :)

wdevfx (approved):

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.visual.ts
Alyar666: > Question.
> 
> Is it safe to compare a scroll value with an exact value? Can we receive here sth like `4200.01`?
> 
> Because if yes, this test may became unstable and be our pain :)

There are already such tests. And they have proven themselves to be stable.
The widths of each column are strictly defined in these tests. I think this will become a problem if scrollLeft is different from expected (even by 0.1 px). We will need to fix this.

PR #30231, opened 2025-06-26
Scheduler: add options class

Ambrozy:


packages/devextreme/js/__internal/scheduler/utils/options_validator/core/options_validator_error_handler.ts
```diff
@@ -13,28 +13,31 @@ export abstract class OptionsValidatorErrorHandler<TValidators extends string> {
...
-      const isLastErrorCode = idx === errorCodeArray.length - 1;
+    const resultArray = [...uniqErrorCodes.entries()];
+    resultArray.forEach((value, idx) => {
+      const [errorCode, args = []] = value;
+      const isLastErrorCode = idx === resultArray.length - 1;
 
       // NOTE: For stopping code stack execution and not creating
       // the special error code for this case,
       // we log all errors and throw the last one.
       if (!isLastErrorCode) {
```
Ambrozy, line 72: Always log instead of error for Wxxxx

packages/devextreme/js/__internal/core/m_errors.ts
```diff
@@ -80,7 +80,7 @@ export default errorUtils({
 
   W0007: '\'{0}\' Globalize culture is not defined',
 
-  W0008: 'Invalid view name: \'{0}\'',
+  W0008: 'Invalid view type: \'{0}\'',
```
wdevfx: Let's mention tech writers about this change :)

packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.currentView = getCurrentView(
+      this.option('currentView') ?? '',
+      views,
+    );
+  }
+
+  _initMarkup(): void {
+    // @ts-expect-error
+    super._initMarkup();
+    this._updateViews();
```
wdevfx: Question.

We call the ```this._updateViews``` here and in the ```_init``` method
Looks like it's enough to call it in the ```_init``` method, or is there a reason why we should call it one more time during the ```_initMarkup```?

packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts
```diff
@@ -0,0 +1,116 @@
...
+  _schedulerOptionChanged<K extends keyof SafeSchedulerOptions>(args: {
+    name: K;
+    value: SafeSchedulerOptions[K];
+  }): void {
+    switch (args.name) {
+      case 'currentView':
+      case 'views':
+        this._updateViews();
+        break;
+      default:
```
wdevfx, line 83: ```break``` missing

```suggestion
      default:
        break;
```

packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts
```diff
@@ -0,0 +1,116 @@
...
+export class SchedulerOptionsBaseWidget extends Widget<SafeSchedulerOptions> {
+  protected views: NormalizedView[] = [];
+
+  public currentView!: NormalizedView;
+
+  private _optionsValidator!: SchedulerOptionsValidator;
+
+  private _optionsValidatorErrorHandler!: SchedulerOptionsValidatorErrorsHandler;
+
+  _init(): void {
```
wdevfx: Let's make methods with underscore (```_```) ```protected``` if they not used outside the Scheduler child class

packages/devextreme/js/__internal/scheduler/utils/options/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+  AgendaView, DateOption, NormalizedView, RawViewType, SafeSchedulerOptions, View, ViewType,
+} from './types';
+
+export const getViews = (views: RawViewType[]): NormalizedView[] => views
+  .filter((view) => VIEW_TYPES.includes((isObject(view) ? view.type : view) as ViewType))
+  .map(
+    (view) => (isObject(view)
+      ? extend({}, DEFAULT_VIEW_OPTIONS[view.type as ViewType], view) as NormalizedView
+      : DEFAULT_VIEW_OPTIONS[view]),
+  ).filter(Boolean);
```
wdevfx: I suggest refactoring this util function following way (and remove ```as``` casts)
Also, let's rename ```getView``` -> ```normalizeView``` because this function perform normalization :)

```ts
export const normalizeView = (view: RawViewType): NormalizedView | null => {
  const isObjectConfiguration = isObject(view);
  const viewType = isObjectConfiguration ? view.type : view;

  if (!viewType) {
    return null;
  }

  return isObjectConfiguration
    ? { ...DEFAULT_VIEW_OPTIONS[viewType], ...view }
    : DEFAULT_VIEW_OPTIONS[viewType];
};

export const normalizeViewArray = (viewArray: RawViewType[]): NormalizedView[] => viewArray
  .map((view) => normalizeView(view))
  .filter((view): view is NormalizedView => Boolean(view));
```

packages/devextreme/js/__internal/scheduler/utils/options/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export function getCurrentView(currentView: 'agenda', views: RawViewType[]): AgendaView;
+export function getCurrentView(currentView: string | ViewType, views: RawViewType[]): View;
+export function getCurrentView(
```
wdevfx: Question.

As far as I can see, there only one usage of the ```getCurrentView``` method
Why do we need all of this overloads?

packages/devextreme/js/__internal/scheduler/utils/options/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+): SafeSchedulerOptions[K] => {
+  if (isDateOption(optionName)) {
+    const dateValue = currentOptionValue as SafeSchedulerOptions[DateOption];
+    const date = optionName === 'currentDate'
+      ? parseCurrentDate(dateValue)
+      : parseDateOption(dateValue);
+    return date as SafeSchedulerOptions[K];
+  }
+
+  return currentOptionValue;
```
wdevfx: Minor comment.

Let's invert this ```if``` condition to reduce complexity

```ts
if (!isDateOption(optionName)) {
      return currentOptionValue;
}

const dateValue = currentOptionValue as SafeSchedulerOptions[DateOption];
const date = optionName === 'currentDate'
  ? parseCurrentDate(dateValue)
  : parseDateOption(dateValue);
return date as SafeSchedulerOptions[K];
```

packages/devextreme/js/__internal/scheduler/utils/options/utils.test.ts
```diff
@@ -0,0 +1,197 @@
...
+  describe, expect, it,
+} from '@jest/globals';
+
+import { VIEWS } from './constants_view';
+import {
+  getCurrentView, getViewOption, getViews, parseCurrentDate, parseDateOption,
+} from './utils';
+
+describe('views utils', () => {
+  describe('getCurrentView', () => {
```
wdevfx: Looks like typo in the ```describe```

packages/devextreme/js/__internal/scheduler/utils/options/utils.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    it('should set default for undefined props', () => {
```
wdevfx: Let's better use ```it.each``` or loop to test each array item separately
It's better to check one case (one array item) instead of checking the whole array

packages/devextreme/js/__internal/scheduler/utils/options/utils.test.ts
wdevfx: Missing tests for the next scenarios:
View as string:
- nonexistent string (like ```my-super-day-view```)

View as object:
- default options should not override view options
- view with incorrect type (like ```my-view-type```)
- view as bad object, like array or sth like this

packages/devextreme/js/__internal/scheduler/utils/options/utils.test.ts
```diff
@@ -0,0 +1,197 @@
...
+          intervalCount: 1,
+          name: 'Agenda',
+          type: 'agenda',
+        },
+      ]);
+    });
+  });
+
+  describe('getCurrentView', () => {
+    it('should return string', () => {
```
wdevfx: Looks like wrong test case name, the ```getCurrentView``` doesn't return a ```string``` here

packages/devextreme/js/__internal/scheduler/utils/options/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+): optionName is DateOption => ['currentDate', 'min', 'max'].includes(optionName);
+
+export const getViewOption = <K extends keyof SafeSchedulerOptions>(
+  optionName: K,
+  currentOptionValue: SafeSchedulerOptions[K],
+): SafeSchedulerOptions[K] => {
+  if (isDateOption(optionName)) {
+    const dateValue = currentOptionValue as SafeSchedulerOptions[DateOption];
+    const date = optionName === 'currentDate'
+      ? parseCurrentDate(dateValue)
```
wdevfx: ```parseCurrentDate``` returns ```Date | undefined```, but in the ```SafeSchedulerOptions['currentDate']``` the type is strict ```Date```

packages/devextreme/js/__internal/scheduler/utils/options_validator/core/options_validator_error_handler.ts
```diff
@@ -0,0 +1,61 @@
...
+  ) {}
+
+  handleValidationResult(
+    optionsValidatorResult: OptionsValidatorResult<TValidators>,
+  ): void {
+    if (optionsValidatorResult === true) {
+      return;
+    }
+
+    const uniqErrorCodes = Object.entries(optionsValidatorResult)
```
wdevfx: Let's discuss this logic separately

packages/devextreme/js/__internal/scheduler/utils/options_validator/validator_rules.ts
```diff
@@ -29,9 +32,29 @@ export const cellDurationMustBeLessThanVisibleInterval = createValidatorRule(
...
+    // eslint-disable-next-line no-restricted-syntax
+    for (const view of views) {
+      const viewType = isObject(view) ? view.type : view;
+      const isValidView = Boolean(viewType && VIEW_TYPES.includes(viewType));
+
+      if (!isValidView) {
+        return {
+          message: `The view type "${viewType}" is not supported. Supported types: ${VIEW_TYPES.join(', ')}.`,
+          arguments: [String(viewType)],
+        };
```
wdevfx: Looks too complex
Why do we need ```arguments``` here, if we already has a ```viewType``` in the error message?

Let's discuss it separately

packages/devextreme/js/__internal/scheduler/workspaces/m_cache.ts
```diff
@@ -1,23 +1,25 @@
 import { isDefined } from '@js/core/utils/type';
 
-export class Cache {
-  _cache = new Map();
+export class Cache<K = any, V = any> {
```
wdevfx: Looks like here we get merge conflicts with another change:
https://github.com/DevExpress/DevExtreme/pull/30239

vladaskorohodova (dismissed):

packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts
```diff
@@ -0,0 +1,116 @@
...
+import { SchedulerOptionsValidator, SchedulerOptionsValidatorErrorsHandler } from './utils/options_validator/index';
+
+export class SchedulerOptionsBaseWidget extends Widget<SafeSchedulerOptions> {
+  protected views: NormalizedView[] = [];
+
+  public currentView!: NormalizedView;
+
+  private _optionsValidator!: SchedulerOptionsValidator;
+
+  private _optionsValidatorErrorHandler!: SchedulerOptionsValidatorErrorsHandler;
```
wdevfx: We decided not to use the ```_```
So, let remove this symbol

```suggestion
  private optionsValidator!: SchedulerOptionsValidator;

  private optionsValidatorErrorHandler!: SchedulerOptionsValidatorErrorsHandler;
```

packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  protected _defaultOptionsRules(): SchedulerOptionsRule[] {
+    // @ts-expect-error
+    const rules: SchedulerOptionsRule[] = super._defaultOptionsRules();
+
+    return rules.concat(DEFAULT_SCHEDULER_OPTIONS_RULES);
+  }
+
+  protected _updateViews(): void {
```
wdevfx: Same comment for ```_```
```suggestion
  protected updateViews(): void {
```

packages/devextreme/js/__internal/scheduler/utils/options_validator/validator_rules.ts
```diff
@@ -0,0 +1,54 @@
...
+    startDayHour,
+    endDayHour,
+  }: SafeSchedulerOptions) => {
+    const visibleInterval = (endDayHour - startDayHour) * 60;
+    return lessThan(cellDuration, visibleInterval, false);
+  },
+);
+
+export const allViewsHasCorrectType = createValidatorRule(
+  'views',
```
wdevfx: Let's add more details to this validation rule name
The ```views``` it's too common. It's hard to understand what this validation rule checks if u see only it's name

packages/devextreme/js/__internal/scheduler/utils/options_validator/validator_rules.ts
```diff
@@ -0,0 +1,54 @@
...
+  (views: SafeSchedulerOptions['views']) => {
+    // eslint-disable-next-line no-restricted-syntax
+    for (const view of views) {
+      const viewType = isObject(view) ? view.type : view;
+      const isValidView = Boolean(viewType && VIEW_TYPES.includes(viewType));
+
+      if (!isValidView) {
+        return { arguments: [String(viewType)] };
+      }
```
wdevfx: Question.

Is it expected behavior that we will return on the first invalid view?
So, if we have two incorrect view -> we will see warning only about first one

If it's by design -> let's add a ```jest``` test for this case

packages/devextreme/js/__internal/scheduler/utils/options_validator/validator_rules.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    });
+
+    it('should be the function with the correct name', () => {
+      const func = cellDurationMustBeLessThanVisibleInterval;
+
+      expect(func.name).toBe('cellDurationMustBeLessThanVisibleInterval');
+    });
+  });
+
+  describe('allViewsHasCorrectType', () => {
```
wdevfx, line 145: Let's add same test cases but with string views (like ```week```, ```timelineDay```, etc.)

packages/devextreme/js/__internal/scheduler/utils/options_validator/core/options_validator_error_handler.ts
wdevfx: Let's simplify this code a little bit
I suggest the following way:

1) Let's add a type for logger data
```ts
// NOTE: We can place it in types.ts or on top of this file
interface LogData { errorCode: string; args: string[] }
```

2) Let's add two helper util functions
```ts
// NOTE: We can add this functions in separate utils.ts file or on top of this file
    const getValidatorErrorUniqueKey = (
      errorCode: string,
      validatorError: ValidatorRuleResult,
    ): string => {
      if (typeof validatorError === 'boolean' || !Array.isArray(validatorError.arguments)) {
        return errorCode;
      }

      return `${errorCode}${JSON.stringify(validatorError.arguments)}`;
    };

    const getValidatorErrorArguments = (
      validatorError: ValidatorRuleResult,
    ): string[] => (typeof validatorError === 'boolean' || !Array.isArray(validatorError.arguments)
      ? []
      : validatorError.arguments);
```

3) Refactor this code a little bit and replace reduce:
```ts
  handleValidationResult(
    optionsValidatorResult: OptionsValidatorResult<TValidators>,
  ): void {
    if (optionsValidatorResult === true) {
      return;
    }

    const warningsMap = new Map<string, LogData>();
    const errorsMap = new Map<string, LogData>();

    Object.entries(optionsValidatorResult)
      .forEach(([validatorName, validatorErrorRecord]) => {
        const errorCode: string | undefined = this.validatorNameToErrorCodeMap[validatorName];

        if (!errorCode) {
          return;
        }

        const logMap = errorCode.startsWith('E')
          ? errorsMap
          : warningsMap;

        Object.values(validatorErrorRecord as ValidatorErrors)
          .forEach((validatorError) => {
            const uniqueKey = getValidatorErrorUniqueKey(errorCode, validatorError);
            const args = getValidatorErrorArguments(validatorError);
            logMap.set(uniqueKey, { errorCode, args });
          });
      });

    Array.from(warningsMap).forEach((value) => {
      const [, { errorCode, args }] = value;
      this.globalErrorHandler.logError(errorCode, ...args);
    });

    Array.from(errorsMap).forEach((value, idx) => {
      const [, { errorCode, args }] = value;
      const isLastErrorCode = idx === errorsMap.size - 1;

      // NOTE: For stopping code stack execution and not creating
      // the special error code for this case,
      // we log all errors and throw the last one.
      if (!isLastErrorCode) {
        this.globalErrorHandler.logError(errorCode, ...args);
      } else {
        this.globalErrorHandler.throwError(errorCode, ...args);
      }
    });
  }
```

packages/devextreme/testing/tests/DevExpress.ui/defaultOptions.tests.js
```diff
@@ -1804,8 +1804,6 @@ testComponentDefaults(Scheduler,
     {},
```
wdevfx, line 1804: I found one test case where our validation doesn't work and we receive an internal console.error

Scheduler configuration:
```js
$("#scheduler").dxScheduler({
  dataSource: [],
  views: [
    {
      type: 'blabla',
      name: 'blabla',
      blabla: "incorrect view"
    },
  ],
  currentView: "blabla"
});
```

**Steps:**
1. Open this configuration
2. Check console

**Expected result:**
Our error code or default view is shown with warning in console

**Actual result:**
Internal console error in header:
```
jquery.js:3793 Uncaught TypeError: Cannot read properties of undefined (reading 'type')
```

This case was broken in actual versiton too: [link to the CodePen](https://codepen.io/wdevfx/pen/bNdPopJ?editors=0010)

But I think it will be cool if we cover this case too
@Ambrozy, could u please take a look?

packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts
wdevfx: Still an open question :)

packages/devextreme/js/__internal/scheduler/utils/options_validator/validator_rules.ts
Ambrozy: let's discuss

packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts
Ambrozy: Didn't find better place for this call. Removed form _init(). It's require to sync options before murkup render.

packages/devextreme/js/__internal/scheduler/utils/options_validator/validator_rules.test.ts
wdevfx: Still open comment

packages/devextreme/testing/tests/DevExpress.ui/defaultOptions.tests.js
wdevfx: FIxed ✅

packages/devextreme/js/__internal/scheduler/utils/options_validator/validator_rules.ts
wdevfx: Decided to add all incorrect views' types to the validator's arguments as a string (separated by the ```,```)

wdevfx (approved):

PR #30239, opened 2025-06-30
Scheduler: Add timezones cache (T1296816)

Ambrozy:


e2e/testcafe-devextreme/tests/scheduler/performance/performance.ts
```diff
@@ -0,0 +1,60 @@
+import Scheduler from 'devextreme-testcafe-models/scheduler';
```
wdevfx: Let's follow our existing dev flow and create a PR from our forks (not from the main repository)

e2e/testcafe-devextreme/tests/scheduler/performance/performance.ts
```diff
@@ -0,0 +1,60 @@
...
+import { createWidget } from '../../../helpers/createWidget';
```
wdevfx: I'm not sure about performance tests in the testcafe
On the one hand, it's ok to measure the speed in ```e2e```
And if we will make some change that slow down the Scheduler render significantly -> we will see it in ```e2e``` tests

But on the other hand, speed may depend on runner configuration and may differ between runners
Also, why is the ```+200ms``` threshold to pass the check with timezone calculations?

Maybe it will be better to check the cache with other tests
Let's discuss it separately 

@pomahtri cc

packages/devextreme/js/__internal/scheduler/global_cache.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { Cache } from './global_cache';
+
+describe('global cache', () => {
+  it('Initialization: should be empty', () => {
+    const cache = new Cache();
+    expect(cache.size).toBe(0);
+  });
+
+  it('API: should get and set', () => {
```
wdevfx: It's better to split testing ```get``` and ```set``` in different test cases
Because by the jest ideology -> the test should check only one thing

Or, as a possible solution here -> let's rename this test case to sth like:
```should get existing value```

Comment (wdevfx) [inline packages/devextreme/js/__internal/scheduler/global_cache.test.ts]: Let's get rid of the similar prefix in each test name and move the ```API``` to describe

```ts
describe('API', () => {
  test('should get existing value', () => { ... });
});
```

packages/devextreme/js/__internal/scheduler/global_cache.ts
```diff
@@ -0,0 +1,35 @@
...
+export class Cache {
+  _cache = new Map();
+
+  get size(): number { return this._cache.size; }
+
+  clear(): void {
+    this._cache.clear();
+  }
+
+  get<R>(name: string, defaultValueCallback?: () => R): R {
```
wdevfx: The correct return type here is:
```R | undefined```

Because if we call ```get``` with unexisting key -> we will receive ```undefined```

File: packages/devextreme/js/__internal/scheduler/global_cache.ts [hunk 5]
```diff
@@ -0,0 +1,35 @@
...
+    if (!this._cache.has(name) && defaultValueCallback) {
+      this.set(name, defaultValueCallback());
+    }
+
+    return this._cache.get(name) as R;
```
wdevfx: Let's remove this ```as R```
```suggestion
    return this._cache.get(name);
```

packages/devextreme/js/__internal/scheduler/global_cache.test.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  describe, expect, it,
+} from '@jest/globals';
+
+import { Cache } from './global_cache';
+
+describe('global cache', () => {
```
wdevfx, line 7: Let's add more test cases + split existing ones

From my perspective, test coverage should be:
- empty at initialization
- get nonexistent key
- get existing key
- get nonexisting key + default value callback
- get existing key + default value callback
- set defined key
- set not defined key (e.g. ```undefined``` or ```null```)
- delete nonexisting key
- delete existing key
- clear method test

packages/devextreme/js/__internal/scheduler/global_cache.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  set<R>(name: string, value: R): R {
+    if (isDefined(value)) {
+      this._cache.set(name, value);
+    }
+
+    return value;
```
wdevfx: Why do we return the same value from the ```set``` method?
How may it be used?

packages/devextreme/js/__internal/scheduler/global_cache.ts
```diff
@@ -0,0 +1,35 @@
+import { isDefined } from '@js/core/utils/type';
+
+export class Cache {
+  _cache = new Map();
```
wdevfx: Let's make this field ```private```

wdevfx: [Point to discussion]

About underscores -> let's discuss, in the new ```grids``` code we try not to use the ```_```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -342,29 +344,23 @@ const getTimeZones = (
...
-    }),
-    GET_TIMEZONES_BATCH_SIZE,
-  );
-  timeZoneDataCache = await timeZoneDataCachePromise;
-
-  return timeZoneDataCache;
-};
-
-const getTimeZonesCache = (): TimezoneLabel[] => timeZoneDataCache;
+): Promise<TimezoneLabel[]> => globalCache.timezones.get(
```
wdevfx: I'm a little bit lost in this new logic -> let's discuss it separately

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/appointmentPopup.tests.js
```diff
@@ -1879,7 +1880,7 @@ module.skip('Timezone Editors', moduleOptions, () => {
 
         [startDateTimezoneEditor, endDateTimezoneEditor].forEach(editor => {
             assert.equal(editor.option('value'), 'Europe/Paris', 'value is ok');
-            assert.equal(editor.option('displayValue'), '(GMT +01:00) Europe - Paris', 'displayValue is ok');
+            assert.ok(editor.option('displayValue').endsWith('Europe - Paris'), 'displayValue is ok');
```
wdevfx, line 1883: Why do we change this assertion?

e2e/testcafe-devextreme/tests/scheduler/performance/performance.ts
Ambrozy: let's use jest. write note. QUnit (or jest) check that Intl not run
Add issue to run scheduler in jest
Remove this test

packages/devextreme/js/__internal/scheduler/global_cache.ts
Ambrozy: drop `undefined`

packages/devextreme/js/__internal/scheduler/global_cache.ts
Ambrozy: make set private. rename get to memo

packages/devextreme/js/__internal/scheduler/global_cache.ts
Ambrozy: not use _

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
wdevfx: Discussed, after 25.1 improvements this logic is nice

wdevfx (approved):

PR #30255, opened 2025-07-01
DateBox: fix today button text customization with todayButtonText option

aleksei-semikozov:


packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
```diff
@@ -1198,14 +1201,15 @@ class Calendar<
 
   _renderFooter() {
     const showTodayButton = this.option('showTodayButton');
+    const todayButtonText = this.option('todayButtonText') || messageLocalization.format('dxCalendar-todayButtonText');
```
pharret31: I suppose it's better to use ?? operator here so in that case only nullish values (null, undefined) will not be accepted
So users will be able to put an empty string as a value if somewhy they need that

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
```diff
@@ -600,9 +600,9 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
 
   _isValueChanged(newValue): boolean {
     const oldValue = this.dateOption('value');
-    // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
+
     const oldTime = oldValue && oldValue.getTime();
```
pharret31, line 604: Let's prefer optional chain:

`oldValue?.getTime();`

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
```diff
@@ -600,9 +600,9 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
...
-    // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
+
     const newTime = newValue && newValue.getTime();
```
pharret31, line 605: `newValue?.getTime();`

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/calendar.tests.js
```diff
@@ -1857,6 +1857,16 @@ QUnit.module('Options', {
...
+    QUnit.test('todayButtonText option', function(assert) {
+        const getTodayButton = () => this.$element.find(toSelector(CALENDAR_TODAY_BUTTON_CLASS)).get(0);
+
+        this.calendar.option('showTodayButton', true);
+        this.calendar.option('todayButtonText', 'Custom text');
+
+        const $todayButton = getTodayButton();
+        assert.strictEqual($($todayButton).text(), 'Custom text', 'todayButton text matches the todayButtonText option');
+    });
+
```
pharret31, line 1883: Guess such test should be in the markup tests file
Also as I see there's no tests to check if localization is applied
And the empty string test could be added as well if ?? operator is used

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js
```diff
@@ -6356,6 +6356,13 @@ QUnit.module('valueChanged handler should receive correct event', {
         this.checkEvent(assert, 'dxclick', $todayButton);
         this.testProgramChange(assert);
     });
+
+    QUnit.test('should display custom text on today button from todayButtonText option', function(assert) {
+        this.reinit({ calendarOptions: { showTodayButton: true }, todayButtonText: 'today button text' });
+        const $todayButton = $(this.instance.content()).parent().find(`.${CALENDAR_TODAY_BUTTON_CLASS}`);
+
+        assert.equal($todayButton.text().trim(), 'today button text');
+    });
```
pharret31, line 6372: Guess such test should be in the markup tests file
Also as I see there's no tests to check if localization is applied
And the empty string test could be added as well if ?? operator is used

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
```diff
@@ -81,6 +81,8 @@ export interface CalendarProperties extends Properties {
   viewsCount: number;
 
   currentDate?: Date;
+
+  todayButtonText: string | null;
```
pharret31: I think it could be undefined as well

`todayButtonText?:`

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
aleksei-semikozov: I think null is better
1. Explicitly means “no value” — null clearly signals that the value was intentionally cleared. undefined often means it was never set.
2. Consistency with _getDefaultOptions() — we typically return default values like null, not leave properties undefined.

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
```diff
@@ -1198,14 +1201,15 @@ class Calendar<
 
   _renderFooter() {
     const showTodayButton = this.option('showTodayButton');
+    const todayButtonText = this.option('todayButtonText') ?? messageLocalization.format('dxCalendar-todayButtonText');
```
marker-dao: I think, that we can set messageLocalization.format('dxCalendar-todayButtonText') to the getDefaultOptions method as a default value.

marker-dao:
```suggestion
    const { showTodayButton, todayButtonText } = this.option();
```

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
```diff
@@ -1633,6 +1637,7 @@ class Calendar<
       case 'dateSerializationFormat':
       case 'cellTemplate':
       case 'showTodayButton':
+      case 'todayButtonText':
```
marker-dao, line 1639: In my opinion, redrawing the component because of a single text change is too much. Can't we just update the text in the button?

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
```diff
@@ -600,10 +600,10 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
 
   _isValueChanged(newValue): boolean {
     const oldValue = this.dateOption('value');
-    // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
-    const oldTime = oldValue && oldValue.getTime();
-    // eslint-disable-next-line @typescript-eslint/prefer-optional-chain
-    const newTime = newValue && newValue.getTime();
+
+    const oldTime = oldValue?.getTime();
+
```
marker-dao:
```suggestion
```

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
```diff
@@ -718,6 +718,7 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
       case 'interval':
       case 'disabledDates':
       case 'calendarOptions':
+      case 'todayButtonText':
```
marker-dao, line 720: Same thing: why should we redraw a component because of a minor change? Consider simply passing the new parameters to Calendar.

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
```diff
@@ -745,9 +746,6 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
         super._optionChanged.apply(this, arguments);
         this._formatValidationIcon();
         break;
-      case 'todayButtonText':
```
marker-dao, line 748: Does this still work for datetime?

packages/devextreme/js/__internal/ui/date_box/m_date_box.strategy.calendar.ts
```diff
@@ -112,6 +112,7 @@ class CalendarStrategy extends DateBoxStrategy {
       disabledDates: isFunction(disabledDates) ? this._injectComponent(disabledDates.bind(this.dateBox)) : disabledDates,
       onContouredChanged: this._refreshActiveDescendant.bind(this),
       skipFocusCheck: true,
+      todayButtonText: this.dateBox.option('todayButtonText'),
```
marker-dao: Please use deconstruction.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/calendar.tests.js
marker-dao: I think this is the correct location for the test, since markup tests check the rendering of DOM elements, classes, and other things that are specific to markup. This test checks the functionality of the option.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/calendar.tests.js
```diff
@@ -1857,6 +1857,16 @@ QUnit.module('Options', {
         assert.strictEqual($todayButton, undefined, 'todayButton is not rendered after showTodayButton runtime change to false');
     });
 
+    QUnit.test('todayButtonText option', function(assert) {
+        const getTodayButton = () => this.$element.find(toSelector(CALENDAR_TODAY_BUTTON_CLASS)).get(0);
+
+        this.calendar.option('showTodayButton', true);
```
marker-dao:
```suggestion
        this.calendar.option({
            showTodayButton: true,
            todayButtonText: 'Custom text',
        });
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/calendar.tests.js
```diff
@@ -1857,6 +1857,16 @@ QUnit.module('Options', {
         assert.strictEqual($todayButton, undefined, 'todayButton is not rendered after showTodayButton runtime change to false');
     });
 
+    QUnit.test('todayButtonText option', function(assert) {
```
marker-dao, line 1872: In this test, you checked the runtime change of the option because the component was initialised and rendered in the before hook, after which you changed the value of the option and optionChanged was triggered.

However, you did not test the case where the user initially sets the value of this option. You also need to test the case where the user does not set any text for this option. Finally, runtime option value change is what you have already done in this test. The same applies to DateBox.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js
marker-dao: I think this is the correct test placement. But we need to expand the coverage of scenarios.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js
```diff
@@ -6356,6 +6356,13 @@ QUnit.module('valueChanged handler should receive correct event', {
         this.checkEvent(assert, 'dxclick', $todayButton);
         this.testProgramChange(assert);
     });
+
+    QUnit.test('should display custom text on today button from todayButtonText option', function(assert) {
+        this.reinit({ calendarOptions: { showTodayButton: true }, todayButtonText: 'today button text' });
+        const $todayButton = $(this.instance.content()).parent().find(`.${CALENDAR_TODAY_BUTTON_CLASS}`);
+
+        assert.equal($todayButton.text().trim(), 'today button text');
```
marker-dao, line 6371: Why did you use the .trim() method?

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
marker-dao: Please see the comment: https://github.com/DevExpress/DevExtreme/pull/30255#discussion_r2177630995

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js
aleksei-semikozov: it's more robust. skip white spaces

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
marker-dao: We decided that it was okay to use invalidate in this case to avoid writing unnecessary code, as the option is unlikely to change frequently. This should not affect performance.

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
marker-dao: https://github.com/DevExpress/DevExtreme/pull/30255#discussion_r2179784952

marker-dao (approved):

PR #30257, opened 2025-07-01
TabPanel: add itemIndex to the TitleClickEvent typescript definition (T1296996)

EugeniyKiyashko:


packages/devextreme/js/ui/tab_panel.d.ts
```diff
@@ -49,6 +49,8 @@ export interface TabPanelItemInfo<TItem extends ItemLike> {
     readonly itemData?: TItem;
     /** @docid */
     readonly itemElement?: DxElement;
+    /** @docid */
+    readonly itemIndex: number;
```
pharret31, line 53: Shouldn't it be optional like other properties?

marker-dao (approved):

pharret31 (approved):

vladaskorohodova (approved):

packages/devextreme/js/ui/tab_panel.d.ts
EugeniyKiyashko: I suppose it shouldn't, because itemIndex is always presented in event data

PR #30283, opened 2025-07-06
Demos tests: move skippedTests and accessibilityUnsupported lists to separate files

Zedwag:


r-farkhutdinov (dismissed):

apps/demos/testing/skipped-tests.js
```diff
@@ -0,0 +1,241 @@
+import { THEME } from '../utils/visual-tests/helpers/theme-utils';
```
pomahtri, line 1: Can we make this file `.json` or make other files `.js`?

marker-dao (dismissed):

apps/demos/testing/skipped-tests.js
Zedwag: converted to js

pomahtri (approved):

PR #30290, opened 2025-07-07
Remove type assertion in vue template

GoodDayForSurf:
(this syntax cause error in CodeSandbox)


apps/demos/Demos/FilterBuilder/WithDataGrid/Vue/App.vue
```diff
@@ -30,6 +30,7 @@ import { DataSource, ODataStore } from 'devextreme-vue/common/data';
 import { filter, fields } from './data.ts';
 
 const filterValue = ref(filter);
+const gridColumns = fields as DxDataGridTypes.Column[];
```
pomahtri: Maybe we can instead change type of `fields` inside data.ts?
E.g. 
```
export const fields: (DxFilterBuilderTypes.Field & DxDataGridTypes.Column)[]
```

GoodDayForSurf: I tried it but it gives ts-check error:
error TS2322: Type 'dxFilterBuilderField' is not assignable to type 'dxFilterBuilderField & Column'

GoodDayForSurf: UPD: (DxFilterBuilderTypes.Field & DxDataGridTypes.Column)[] is OK after removing type assertion from fields array in data.ts

thx

pomahtri (approved):

anna-shakhova (approved):

PR #30292, opened 2025-07-07
Add enums to SMD external config

alexslavr:


packages/devextreme-metadata/common/smd.ts
```diff
@@ -18,3 +18,573 @@ export const commonSmdCollectionItems = [
   'ui/tabs:dxTabsItem',
   'ui/tile_view:dxTileViewItem',
 ];
+
+export const enums = {
```
IlyaKhD: I suggest not placing the enums configuration in the `common` folder. Maybe it could go in a separate module, but still scoped to aspnet.

IlyaKhD (approved):

PR #30294, opened 2025-07-07
PivotGrid: fix data area width for virtual scrolling (T1290313)

Ambrozy:


e2e/testcafe-devextreme/tests/pivotGrid/scrolling.ts
```diff
@@ -139,6 +140,55 @@ testFixture()`PivotGrid_scrolling`
...
+    { useNative: true },
+    { useNative: false },
+  ].forEach(({ useNative }) => {
+    test(`Rows content dont hide under vertical scrollbar when scrolling{useNative=${useNative}},height=100% (${height}px) (T1290313)`, async (t) => {
+      await t
+        .expect(await compareScreenshot(
+          t,
+          `PivotGrid rows content height=100%(${height}px),useNative=${useNative}.png`,
+          Selector('#container'),
+        ))
```
wdevfx: The ```PivotGrid``` tested in 3 themes on the CI
It means that we have 3 separate jobs that run all tests in different themes.
Now the latest etalon will override other themes' etalons

Therefore, let's use a special utility function for taking a screenshot:
[testScreenshot](https://github.com/DevExpress/DevExtreme/blob/25_2/e2e/testcafe-devextreme/helpers/themeUtils.ts#L23)

e2e/testcafe-devextreme/tests/pivotGrid/scrolling.ts
```diff
@@ -139,6 +140,55 @@ testFixture()`PivotGrid_scrolling`
...
+        .ok();
+    }).before(async (t) => {
+      await insertStylesheetRulesToPage(`#parentContainer { height: ${height}px; }`);
+      await t.resizeWindow(600, 600);
```
wdevfx: If u'd like to change browser size for some test, it's required to return default browser size after test.
Because with the ```disablePageReload``` test cafe feature, the next test will be executed in the same browser window (600x600) 

We have special util [safeSizeTest](https://github.com/DevExpress/DevExtreme/blob/25_2/e2e/testcafe-devextreme/helpers/safeSizeTest.ts#L56C7-L56C19) for it, let's use it here

e2e/testcafe-devextreme/tests/pivotGrid/scrolling.ts
```diff
@@ -139,6 +140,55 @@ testFixture()`PivotGrid_scrolling`
...
+    test(`Rows content dont hide under vertical scrollbar when scrolling{useNative=${useNative}},height=100% (${height}px) (T1290313)`, async (t) => {
+      await t
+        .expect(await compareScreenshot(
+          t,
+          `PivotGrid rows content height=100%(${height}px),useNative=${useNative}.png`,
+          Selector('#container'),
+        ))
+        .ok();
+    }).before(async (t) => {
+      await insertStylesheetRulesToPage(`#parentContainer { height: ${height}px; }`);
```
wdevfx: If u add some custom styles to the test page, because of the ```disablePageReload``` testcafe feature, it's required to remove these styles from the test page after the test.

Let's do it in the test cafe ```after``` hook

e2e/testcafe-devextreme/tests/pivotGrid/etalons/PivotGrid rows content height=100%(450px),useNative=true.png
wdevfx, line 1: Cannot reproduce it locally on mac device, but this etalon looks like initial bug
Sth wrong with it, please re-check

e2e/testcafe-devextreme/tests/pivotGrid/etalons/PivotGrid rows content height=100%(600px),useNative=false.png
wdevfx, line 1: Question.

Do I understand it right, that with the ```600px``` container's height we test that scroll-bar should not be visible?

Ambrozy: yes, removed

e2e/testcafe-devextreme/tests/pivotGrid/etalons/PivotGrid rows content height=100%(450px),useNative=true.png
Ambrozy: In mac there is different native scrollbar then in windows. Maybe it because of it.

e2e/testcafe-devextreme/tests/pivotGrid/etalons/PivotGrid rows content height=100%(350px),useNative=false (generic-light).png
wdevfx, line 1: I was confused, why only 4 etalons (should be 4 x 3 themes)?
But looks like that ```scrolling``` tested only in one theme o_O
[link to the code](https://github.com/DevExpress/DevExtreme/blob/25_2/e2e/testcafe-devextreme/tests/pivotGrid/scrolling.ts#L15)

So, let's leave it as is, from my perspective, one theme check is enough here

wdevfx (approved):

PR #30298, opened 2025-07-07
Setup CI on branch 25_2

alexslavr:


packages/devextreme/ports.json
```diff
@@ -1,4 +1,4 @@
 {
-  "qunit": "20050",
-  "vectormap-utils-tester": "20051"
+  "qunit": "20060",
+  "vectormap-utils-tester": "20061"
```
pomahtri, line 3: question
Can we stop updating this, since we don't run qunits on farm (or any other  place which require specific ports) and just make them use any free port?

pomahtri (commented):
Should we also update here? https://github.com/pomahtri/DevExtreme/blob/6fae230ab0699b5a6afc9e872841a465439ae30f/packages/devextreme-metadata/make-aspnet-metadata.ts#L10 

~~And here~~
~~https://github.com/pomahtri/DevExtreme/blob/6fae230ab0699b5a6afc9e872841a465439ae30f/packages/devextreme/js/core/version.js#L1~~

~~And package.json~~ 
UPD: oh, I see another PR

packages/devextreme/ports.json
alexslavr: I'm not very familiar with qunits, so I just follow the instruction here ))

pomahtri (approved):

PR #30299, opened 2025-07-07
ExportFromMarkup: fix serialization for self closing tags (T1297767)

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/core/utils/svg.js
+++ b/packages/devextreme/js/core/utils/svg.js
@@ -6,13 +6,14 @@ import { isRenderer, isString } from './type';
 const window = getWindow();
 
 function getMarkup(element, backgroundColor) {
-    const temp = domAdapter.createElement('div');
     const clone = element.cloneNode(true);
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/core/utils/svg.js
```diff
@@ -6,13 +6,14 @@ import { isRenderer, isString } from './type';
 const window = getWindow();
 
 function getMarkup(element, backgroundColor) {
-    const temp = domAdapter.createElement('div');
     const clone = element.cloneNode(true);
+    const serializer = new window.XMLSerializer();
```
pharret31, line 10: do we need window here?

pharret31 (dismissed):

packages/devextreme/js/core/utils/svg.js
anna-shakhova: yes, in 24.1 there is an error without window, as linter could not undestand what is it

pharret31 (approved):

PR #30314, opened 2025-07-08
Demos. Remove type assertion in props value in Vue template

GoodDayForSurf:


pomahtri (approved):

apps/demos/Demos/FilterBuilder/WithDataGrid/Vue/data.ts
```diff
@@ -1,4 +1,5 @@
 import { type DxFilterBuilderTypes } from 'devextreme-vue/filter-builder';
+import { type DxDataGridTypes } from "devextreme-vue/cjs/data-grid";
```
anna-shakhova, line 2: Please, check import path

anna-shakhova (approved):

PR #30315, opened 2025-07-08
ContextMenu: fix flickers when adding more items asynchronously or during runtime (T1247739)

aleksei-semikozov:
We have removed the templatesRenderAsynchronously logic along with the related test because we no longer support AngularJS 1.x.

### DIFF
--- a/packages/devextreme/js/__internal/ui/overlay/m_overlay.ts
+++ b/packages/devextreme/js/__internal/ui/overlay/m_overlay.ts
@@ -136,8 +136,6 @@ class Overlay<
 
   _zIndex!: number;
 
-  _asyncShowTimeout?: ReturnType<typeof setTimeout> | null;
-
   _actions?: any;
 
   _isHidingActionCanceled?: boolean;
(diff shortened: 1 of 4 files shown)


packages/devextreme/js/__internal/ui/overlay/m_overlay.ts
```diff
@@ -570,13 +570,7 @@ class Overlay<
         this._processShowingHidingCancel(showingArgs.cancel, applyShow, cancelShow);
       };
 
-      if (this.option('templatesRenderAsynchronously')) {
-        this._stopShowTimer();
-        // NOTE: T390360, T386038
-        this._asyncShowTimeout = setTimeout(show);
```
ksercs, line 576: Don't you want to remove the rest artifacts related to this timeout? `_stopShowTimer` method, cleaning up and so on

aleksei-semikozov: done

marker-dao (approved):

ksercs (approved):

PR #30337, opened 2025-07-11
FilterBuilder - Fix value formatting - T1284504

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/filter_builder/m_utils.ts
+++ b/packages/devextreme/js/__internal/filter_builder/m_utils.ts
@@ -6,7 +6,7 @@ import { compileGetter } from '@js/core/utils/data';
 import { Deferred, when } from '@js/core/utils/deferred';
 import { extend } from '@js/core/utils/extend';
 import { captionize } from '@js/core/utils/inflector';
-import { isDefined, isFunction } from '@js/core/utils/type';
+import { isBoolean, isDefined, isFunction } from '@js/core/utils/type';
 import formatHelper from '@js/format_helper';
 import filterUtils from '@js/ui/shared/filtering';
 import errors from '@js/ui/widget/ui.errors';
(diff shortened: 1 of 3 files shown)


packages/devextreme/js/__internal/filter_builder/m_utils.ts
```diff
@@ -54,8 +54,21 @@ const FILTER_BUILDER_ITEM_TEXT_PART_CLASS = `${FILTER_BUILDER_ITEM_TEXT_CLASS}-p
 const FILTER_BUILDER_ITEM_TEXT_SEPARATOR_CLASS = `${FILTER_BUILDER_ITEM_TEXT_CLASS}-separator`;
 const FILTER_BUILDER_ITEM_TEXT_SEPARATOR_EMPTY_CLASS = `${FILTER_BUILDER_ITEM_TEXT_SEPARATOR_CLASS}-empty`;
 
-function getFormattedValueText(field, value) {
+function getFormattedValueText(field, value): string {
   const fieldFormat = field.format || DEFAULT_FORMAT[field.dataType];
+
+  if (isBoolean(value)) {
```
pomahtri, line 60: Let's make condition for boolean similar to date. E.g. `dataType === 'boolean'`

packages/devextreme/js/__internal/filter_builder/m_utils.ts
```diff
@@ -54,8 +54,21 @@ const FILTER_BUILDER_ITEM_TEXT_PART_CLASS = `${FILTER_BUILDER_ITEM_TEXT_CLASS}-p
 const FILTER_BUILDER_ITEM_TEXT_SEPARATOR_CLASS = `${FILTER_BUILDER_ITEM_TEXT_CLASS}-separator`;
 const FILTER_BUILDER_ITEM_TEXT_SEPARATOR_EMPTY_CLASS = `${FILTER_BUILDER_ITEM_TEXT_SEPARATOR_CLASS}-empty`;
 
-function getFormattedValueText(field, value) {
+function getFormattedValueText(field, value): string {
```
pomahtri, line 57: Please check if we already have such utility. Check please where `'dxDataGrid-trueText'` is also used

Tucchhaa: TrueText is used here: [ColumnController._updateColumnOptions](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts#L1130) > [getCustomizeTextByDataType](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts#L300) > [customizeTextForBooleanDataType](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts#L291)

Date format is set here: [ColumnController._updateColumnOptions](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts#L1129), but I am not sure how exactly it is displayed in correct date format. I tried to debug to find it out, but it took too much time.

I can dig deeper, if you think that's important. But in my opinion, it's not important in the current context, because looks like even if find it, we wouldn't be able to reuse the helper function.

packages/devextreme/js/__internal/filter_builder/m_utils.ts
Tucchhaa: Actually if we do this, it can be problematic:

E.g.
```js
$('#my_id').dxFilterBuilder({
   fields: [{
      dataField: "X",
      trueText: 'aboba'
   }],
   value: [["X", "=", true]]
});
```

Then in UI the displayed value would be "true", but not "aboba" as specified.

pomahtri: Oh, I see
So we don't have auto calculation of the 'dataType' option, unlike DataGrid

packages/devextreme/js/__internal/filter_builder/m_utils.ts
pomahtri: Ok, let's postpone

pomahtri (approved):

PR #30351, opened 2025-07-11
Drawer: refactor drawer - improve typing

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/ui/drawer/drawer.animation.ts
+++ b/packages/devextreme/js/__internal/ui/drawer/drawer.animation.ts
@@ -1,3 +1,4 @@
+import type { Direction } from '@js/common';
 import type { AnimationConfig, AnimationState, AnimationType } from '@js/common/core/animation';
 import { fx } from '@js/common/core/animation';
 import type { dxElementWrapper } from '@js/core/renderer';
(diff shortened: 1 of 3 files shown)


packages/devextreme/js/__internal/ui/drawer/drawer.ts
```diff
@@ -71,10 +71,8 @@ class Drawer extends Widget<DrawerProperties> {
       ...super._getDefaultOptions(),
       position: 'left',
       opened: false,
-      // @ts-expect-error ts-error
-      minSize: null,
-      // @ts-expect-error ts-error
-      maxSize: null,
+      minSize: undefined,
```
EugeniyKiyashko: According to documentation null value is expected

packages/devextreme/js/__internal/ui/drawer/drawer.animation.ts
```diff
@@ -25,33 +26,26 @@ export interface FadeConfig {
...
+        return undefined;
     }
+  },
+
+  moveTo(config: DrawerMoveAnimationConfig): void {
+    const {
+      $element, position, direction = 'left', duration, complete,
+    } = config;
+    const toConfig = this.getMoveToConfig(direction, position);
+    const animationType: AnimationType = direction === 'right' ? ('custom' as unknown as AnimationType) : 'slide';
```
EugeniyKiyashko:
```suggestion
    const animationType = direction === 'right' ? 'custom' : 'slide';
```

EugeniyKiyashko: It’s better to add ts-expect-error before type: animationType, because it’s an issue in the public types. Once the accepted values for type are extended, the issue will be fixed

EugeniyKiyashko (approved):

PR #30371, opened 2025-07-14
Popup: Move triggering of resize before animation

marker-dao:
## Failed test (only Vue)
https://github.com/DevExpress/DevExtreme/actions/runs/16262938881?pr=30351 
https://github.com/DevExpress/DevExtreme/actions/runs/16262938881/job/45917687197?pr=30351

## Thoughts
Presumably due to the asynchronous mechanism of calling the update() hook. VUE.nextTick() emulation is possible required. The issue may be related to animation.

## Reason
In Vue, due to the nextTick() mechanism, the Toolbar is rendered from the microtask queue (since nextTick is a Promise.then callback: https://github.com/vuejs/core/blob/main/packages/runtime-core/src/scheduler.ts#L50). This happens after all resize events have been triggered, but it should happen before.

I removed the previous resize event call and tied this call to the moment between renderVisibility and animateShowing. This also allowed me to remove unnecessary calls to this event.

### DIFF
--- a/packages/devextreme/testing/tests/DevExpress.ui.widgets/popup.tests.js
+++ b/packages/devextreme/testing/tests/DevExpress.ui.widgets/popup.tests.js
@@ -1583,7 +1583,7 @@ QUnit.module('options changed callbacks', {
             assert.strictEqual(this.resizeEventSpy.callCount, 3, 'event is triggered 3 times during show animation');
         });
 
-        QUnit.test('resize event is triggered on toolbar elements only', function(assert) {
+        QUnit.test('additional resize events are triggered only on toolbar elements', function(assert) {
             this.resizeEventSpy.resetHistory();
             this.reinit({
                 visible: true,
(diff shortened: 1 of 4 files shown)


packages/devextreme/testing/tests/DevExpress.ui.widgets/popup.tests.js
```diff
@@ -1559,10 +1559,29 @@ QUnit.module('options changed callbacks', {
             const $toolbar2 = getTopToolbar();
 
             assert.strictEqual($toolbar2.length, 1, 'top toolbar is rendered');
-            assert.strictEqual(this.resizeEventSpy.callCount, 7, 'event is triggered additional times');
+            assert.strictEqual(this.resizeEventSpy.callCount, 5, 'event is triggered additional times');
 
             assert.strictEqual($toolbar1, $toolbar2, 'toolbar is not rendered twice after toolbarItems update in runtime');
         });
+
```
aleksei-semikozov, line 1566: I think we can rewrite this test more strictly to AAA pattern.
```
QUnit.test('resize event is triggered during animation showing when toolbar is present', function(assert) {
    // Arrange
    this.resizeEventSpy.resetHistory();
    this.reinit({
        visible: false,
        showTitle: true,
        showCloseButton: true,
    });
    const beforeShow = this.resizeEventSpy.callCount;

    // Act
    this.instance.show();

    // Assert
    const topToolbar = this.instance.topToolbar();
    assert.strictEqual(beforeShow, 0, 'event is not triggered when popup is hidden');
    assert.strictEqual(topToolbar.length, 1, 'top toolbar is rendered');
    assert.strictEqual(this.resizeEventSpy.callCount, 3, 'event is triggered 2 times during show animation');
});
```

marker-dao: ### 👍 praise

Good point, will fix) **Upd**. Fixed: [074672e9c276357cb5da711b76c4e579e9d2bb30]

aleksei-semikozov (approved):

Zedwag (approved):

PR #30372, opened 2025-07-14
Diagram: update the demo after making popup showing synchronous

aleksei-semikozov:
After [#30315](https://github.com/DevExpress/DevExtreme/pull/30315), the React Diagram demo started failing because the callback is now called synchronously, while the ref to the Diagram’s container div is still null. Previously, rendering was asynchronous, giving React enough time to assign the ref.

### DIFF
--- a/apps/demos/Demos/Diagram/OperationRestrictions/Angular/app/app.component.ts
+++ b/apps/demos/Demos/Diagram/OperationRestrictions/Angular/app/app.component.ts
@@ -1,12 +1,12 @@
 import {
-  NgModule, Component, ViewChild, enableProdMode,
+  NgModule, Component, enableProdMode,
 } from '@angular/core';
 import { HttpClientModule } from '@angular/common/http';
 import { BrowserModule } from '@angular/platform-browser';
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import notify from 'devextreme/ui/notify';
 import { ArrayStore } from 'devextreme-angular/common/data';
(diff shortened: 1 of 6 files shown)


apps/demos/Demos/Diagram/OperationRestrictions/React/App.tsx
```diff
@@ -50,7 +50,7 @@ export default function App() {
   const diagramRef = useRef(null);
```
ksercs: So, this ref is not used anymore => let's then remove it everywhere?
And let's use this approach across frameworks if we think it's better

aleksei-semikozov: Done

apps/demos/Demos/Diagram/OperationRestrictions/Angular/app/app.component.ts
```diff
@@ -56,7 +54,7 @@ export class AppComponent {
   }
 
   requestEditOperationHandler(e) {
-    const diagram = this.diagram.instance;
+    const diagram = e.component.instance();
```
ksercs, line 57: is this `instance()` needed? e.component should already be an instance

aleksei-semikozov: I double check it. It is really need. In order to take the instance of a component we have to call this function as in other demos

ksercs (approved):

PR #30381, opened 2025-07-15
Scheduler: fix days duplication in workspace (T1252028)

Ambrozy:


packages/devextreme/js/__internal/scheduler/__tests__/santiago_timezone.test.ts
```diff
@@ -0,0 +1,95 @@
...
+        dataSource: priorities,
+        label: 'Priority',
+      }],
+      dataSource,
+    } as any);
+    await new Promise(process.nextTick);
+
+    const workspace = container.querySelector('.dx-scheduler-work-space') as HTMLDivElement;
+
+    expect(workspace).toMatchSnapshot();
```
wdevfx: The result snapshot file is too big (47k lines)
And it will be changed each time that we will change sth in the Scheduler's markup

Let's better select info that we want to check (e.g. header cells content or data cell content in case with a `month` view)
We will make this check more strict and there is a higher chance that this test fail will not be skipped accidentally

packages/devextreme/js/__internal/scheduler/__tests__/santiago_timezone.test.ts
```diff
@@ -0,0 +1,95 @@
...
+      groups: ['priorityId'],
+      currentView: 'week',
+      currentDate: new Date(2024, 8, 8),
+      height: 600,
+      firstDayOfWeek: 1,
+      resources: [{
+        fieldExpr: 'priorityId',
+        dataSource: priorities,
+        label: 'Priority',
+      }],
```
wdevfx: Do we need resources & groups in this test?
If not -> let's remove them

packages/devextreme/js/__internal/scheduler/__tests__/santiago_timezone.test.ts
```diff
@@ -0,0 +1,95 @@
...
+];
+const viewsNames = views.map((view) => view.name);
+
+describe('scheduler', () => {
+  it.each(viewsNames)('should render correct workspace in Santiago DST for view: %s', async () => {
+    const container = document.createElement('div');
+    const scheduler = new Scheduler(container, {
+      views,
+      groups: ['priorityId'],
+      currentView: 'week',
```
wdevfx: Looks like we're not switching view and always test the `week` view here

packages/devextreme/js/__internal/scheduler/__tests__/santiago_timezone.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  {
+    name: 'Timeline Week 8 hour cell duration', type: 'timelineWeek', firstDayOfWeek: 0, cellDuration: 8 * 60,
+  },
+  { name: 'Timeline Day', type: 'timelineDay', intervalCount: 1 },
+  { name: 'Day', type: 'day' },
+];
+const viewsNames = views.map((view) => view.name);
+
+describe('scheduler', () => {
+  it.each(viewsNames)('should render correct workspace in Santiago DST for view: %s', async () => {
```
wdevfx: Let's iterate `views` in `it.each` ->
```ts
  it.each(views)('blabla in $name view', (view) => {
    const scheduler = new Scheduler(container, {
      // blablabla
      views: [view],
      currentView: view,
      // blablabla
    });
  });
```

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
```diff
@@ -195,20 +238,24 @@ export class ViewDataGenerator {
         ? allDayPanelsCount + 1
         : allDayPanelsCount;
 
-      currentViewDataMap[rowIndex].forEach((cell, columnIndex) => {
-        cell.key = keyBase + columnIndex;
-      });
+      currentViewDataMap.push(
```
wdevfx, line 231: Let's discuss it
I understand that better to avoid mutations, but here we have huge matrix and maybe adding key via mutations may  make sense :)

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
```diff
@@ -427,24 +499,29 @@ export class ViewDataGenerator {
...
+      hoursInterval,
       interval,
       firstDayOfWeek,
       intervalCount,
       viewOffset,
     } = options;
+    const cellCountInDay = this.getCellCountInDay(startDayHour, endDayHour, hoursInterval);
 
+    const machineTimezoneName = timeZoneUtils.getMachineTimezoneName();
+    const isSantiagoTimezoneChange = machineTimezoneName === 'America/Santiago';
```
wdevfx: We ***must not*** hardcode timezones names in our source code in conditions

As example this timezones (info from the `Gemini AI`) has shift at midnight too:
- Paraguay
-  Cuba
-  Chile
- Greenland/Denmark
- Jordan
- Lebanon
- Syria
- Palestine
- Iran
- Azores/Portugal

Also, the timezones isn't `const` value and DST time may be changed in any timezone in the future
So, we ***must not*** take into account specific timezone name

Let's think about another fix / another mechanism of detecting issues like this

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator_month.ts
```diff
@@ -20,26 +21,44 @@ export class ViewDataGeneratorMonth extends ViewDataGenerator {
...
     data.today = this.isCurrentDate(startDate, indicatorTime, timeZoneCalculator);
     data.otherMonth = this.isOtherMonth(startDate, this._minVisibleDate, this._maxVisibleDate);
-    data.firstDayOfMonth = isFirstCellInMonthWithIntervalCount(startDate, intervalCount);
+    data.isFirstDayMonthHighlighting = isFirstCellInMonthWithIntervalCount(startDate, intervalCount);
     data.text = monthUtils.getCellText(startDate, intervalCount);
 
     return data;
   }
 
+  getDateByCellIndices(
```
wdevfx: Let's discuss this point
Why do we need special implementation for `month` view model?
As far as I can see, previously we used the `common` one from a base `view_generator`

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator_timeline_month.ts
```diff
@@ -1,14 +1,27 @@
...
 import timezoneUtils from '../../m_utils_time_zone';
+import type { ViewDataProviderExtendedOptions } from './m_types';
 import { ViewDataGenerator } from './m_view_data_generator';
 
 const toMs = dateUtils.dateToMilliseconds;
 
 export class ViewDataGeneratorTimelineMonth extends ViewDataGenerator {
-  _calculateCellIndex(rowIndex, columnIndex, rowCount, columnCount) {
-    return monthUtils.calculateCellIndex(rowIndex, columnIndex, rowCount, columnCount);
+  getDateByCellIndices(
```
wdevfx: Looks like duplication of code, we have same method in month view_generator

packages/devextreme/js/__internal/scheduler/workspaces/view_model/utils/view_generator_utils.ts
```diff
@@ -0,0 +1,40 @@
...
+export const calculateAlignedWeeksBetweenDates = (
+  fromDate: Date,
+  toDate: Date,
+  firstDayOfWeek: number,
+): number => {
+  const alignedFromDate = alignToFirstDayOfWeek(fromDate, firstDayOfWeek);
+  const alignedToDate = alignToLastDayOfWeek(toDate, firstDayOfWeek);
+
+  const weekCount = calculateDaysBetweenDates(alignedFromDate, alignedToDate) / DAYS_IN_WEEK;
+  return Math.max(weekCount, 6);
```
wdevfx: Minor comment.

This code will be just moved, but let's move this `6` into separate `const`
Now it's a little bit unclear why `6` here is used

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
wdevfx: DIscussed, let's avoid mutation here, so now all good

packages/devextreme/js/__internal/scheduler/workspaces/view_model/utils/view_generator_utils.ts
wdevfx: `6` here is a maximal possible rows in one month

packages/devextreme/js/__internal/scheduler/workspaces/view_model/utils/view_generator_utils.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  const weekCount = calculateDaysBetweenDates(alignedFromDate, alignedToDate) / DAYS_IN_WEEK;
+  return Math.max(weekCount, MAX_WEEKS_IN_MONTH);
+};
+
+export const isLocalTimeMidnightDST = (date: Date): boolean => {
+  const startDayDate = new Date(date.getFullYear(), date.getMonth(), date.getDate());
+
+  return startDayDate.getHours() === 1;
+};
```
wdevfx: Let's move this function to other timezone related utils
They placed here: [m_utils_time_zone.ts](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts)

Also, it will be cool to cover it by a few tests

wdevfx (approved):

PR #30385, opened 2025-07-15
HtmlEditor: Get rid of getComputedStyles for table and cell prop fields (T1295949)

marker-dao:
1. BC: no more calculation of values for editors in Cell and Table properties forms (but only hardcoded hor and vert alignment default values). Related fix of cell format getting in devextreme-quill - https://github.com/DevExpress/devextreme-quill/pull/136
2. Fix issue with "nullpx" value of dimension editors after applying empty value
3. Fix parsing of "padding" style for tableLite cells


r-farkhutdinov (dismissed):

packages/devextreme/js/__internal/ui/html_editor/utils/m_toolbar_helper.ts
```diff
@@ -479,32 +486,43 @@ function prepareInsertTableHandler(module) {
...
   let borderColorEditorInstance;
   let backgroundColorEditorInstance;
-  const $table = $element;
+
   const { editorInstance } = module;
-  // eslint-disable-next-line radix
-  const startTableWidth = parseInt(formats.tableWidth) || getOuterWidth($table);
+  const window = getWindow();
+
   const tableStyles = window.getComputedStyle($table.get(0));
```
ksercs: Cannot we completely get rid of this? I don't see a point to look at textAlight, for example. How it can have a meaningful value different from a default one if format is not specified? Same for colors

packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts/tableProperties.tests.js
```diff
@@ -1041,9 +1041,10 @@ module('Table properties forms', {
...
             showTablePropertiesForm(this.instance, $tableElement);
             this.clock.tick(10);
-            const formInstance = this.getFormInstance();
 
+            const formInstance = this.getFormInstance();
             const widthEditor = formInstance.getEditor('width');
+
             widthEditor.option('value', 60);
 
             this.applyFormChanges();
```
ksercs, line 1274: Cannot we add a test for a current failing scenario to keep it in the history?

marker-dao: I think adding tests for all other properties is overkill, so I only added it for border-width: 408877acb4dba859d58b8dc50a73799bf4f7eef9

packages/devextreme/js/__internal/ui/html_editor/utils/m_toolbar_helper.ts
marker-dao: I don't think that's a good idea. Let's look at all the fields for which this is used: 
1. backgroundColor
2. borderStyle
3. borderColor
4. alignment
5. verticalAlignment

We believe that a actual value should be set for them.

We don't know the default value for any of them. It depends on the theme/styles, and the theme/styles can be user-defined. Without obtaining the actual value, we won't be able to set anything, but the actual value is not contained in formats. This way, you will set the default value in the field instead of the actual value.

I suggest leaving this part of the code as it is. There can be no error here.

ksercs: >  It depends on the theme/styles, and the theme/styles can be user-defined.

Same as for properties you have updated :) If we make this BC we presume that we don't consider scenarios when UI is customized not using formats. What's the difference between "borderStyle" and "borderWidth" such that you can set a default value for the second one but no for the first one?
> We don't know the default value for any of them

We know, you can check it here for an alignment: https://www.w3schools.com/css/css_table_align.asp
"By default, the content of <th> elements are center-aligned and the content of <td> elements are left-aligned."

marker-dao: If the user theme/styles specify a border color that differs from our default, then this is a bug: the border color is red, but we are going to display some default color. The same applies to the border style. HtmlEditor will display a `dashed` border, but we will show the default value in the `border style` field, which does not correspond to the current one. **I consider this to be incorrect behavior**.

> If we make this BC we presume that we don't consider scenarios when UI is customized not using formats.

I don't think it should work that way for all properties.

We got rid of getComputedStyles for sizes because they were calculated in an unexpected way. But there is no such problem for other properties: when the page scale changes, the border color or style cannot change.

The screenshot shows that with the current method of setting styles, formats.cellBorderColor is undefined, and the color is now taken from cellStyles.borderTopColor (getComputedStyles). See the example here: https://codepen.io/marker-dao/pen/ogjxweN. I think this way is correct.

image

ksercs: My point is that approaches should be consistent for all props. Now if user sets external `border: 2px red dot` he'll see color and style applied in the form but not width. If we don't leave value computing for dimensions that we mostly should remove it for all props I think

marker-dao: a3d65b5073093e19b064666855daaf84d74c02ca, 3d3b3c7ad84e9f704cd015eaf70e93dc729ad4bd

ksercs (approved):

PR #30386, opened 2025-07-15
TreeList: Fix row alternation when repaintChangesOnly is enabled (T1291705)

Alyar666:


e2e/testcafe-devextreme/tests/treeList/markup.ts
```diff
@@ -159,3 +159,62 @@ const tasksT1223168 = [{
...
+
+// T1291705
+test('The shading should alternate correctly after expanding the node when repaintChangesOnly is enabled', async (t) => {
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const treeList = new TreeList('#container');
+
+  await treeList.apiExpandRow(4);
+  await treeList.apiExpandRow(2);
+
+  await takeScreenshot('row_alternation_after_expanding_node_when_repaintChangesOnly_=_true', treeList.element);
```
Tucchhaa: Can we rename screenshot filename to: "T1291705-row-alternation-after-expanding-node-when-row_alternation_after_expanding_node_when_repaintChangesOnly=true" or something like this, because filenames in the etalons folder use kebab case

e2e/testcafe-devextreme/tests/treeList/markup.ts
```diff
@@ -159,3 +159,62 @@ const tasksT1223168 = [{
...
+}));
+
+test('The shading should alternate correctly after expanding the node when repaintChangesOnly and old fixed columns are enabled', async (t) => {
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const treeList = new TreeList('#container');
+
+  await treeList.apiExpandRow(4);
+  await treeList.apiExpandRow(2);
+
+  await takeScreenshot('row_alternation_after_expanding_node_when_there_is_fixed_column_and_repaintChangesOnly_=_true', treeList.element);
```
Tucchhaa: Can we rename this etalon too?

packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
```diff
@@ -883,8 +886,8 @@ export class RowsView extends ColumnsView {
     return deferred;
   }
 
-  protected _getRows(change) {
-    return change && change.items || this._dataController.items();
+  protected _getRows(change?): Row[] {
```
Tucchhaa, line 889: I think current function name is confusing, I am not sure if it's returning row elements or row item.
What do you think about renaming it to `_getRowItems`?

packages/devextreme/js/__internal/grids/grid_core/views/m_rows_view.ts
```diff
@@ -279,7 +280,9 @@ export class RowsView extends ColumnsView {
         setWatcher({
           element: $row.get(0),
           watch: rowOptions.watch,
-          getter: () => this._isAltRow(row),
+          getter: () => this._isAltRow(
+            gridCoreUtils.getItemByKey(row.key, this._getRows()),
+          ),
```
Tucchhaa: I think we can fix the issue directly in this._isAltRow function:
```js
protected _isAltRow(row) {
    const rowIndex = this._dataController.getRowIndexByKey(row.key);
    return rowIndex % 2 === 1;
 }
 ```
  
  Then we don't need to change this getter and create gridCoreUtils.getItemByKey. What do you think?

Alyar666: dataIndex and rowIndex are different things. For example, when data is grouped, dataIndex will start from 0 within each group.

 ~~But I refactored the code a bit. Now I pass the key to the _isAltRow method, and there I retrieve the required row by the key.~~ - Too many tests failed. At first glance, it's unclear why they failed. I didn’t dig deeper and rolled back this refactoring.

Tucchhaa: I see.

Maybe tests have failed because _isAltRow is used [here](https://github.com/DevExpress/DevExtreme/blob/1495cda0e67e286b7945fe2edc1a06555fa0a475/packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts#L900) too?

I still think that it's better to move the logic of getting the correct row to the _isAltRow. So that whenever this method is used we are sure that we use the correct data. This way we will reduce chances of producing more bugs in the future

Tucchhaa (dismissed):

pomahtri (approved):

PR #30403, opened 2025-07-17
DataGrid - Taps on cells do not activate cell editors in a detail DataGrid if this DataGrid is inside TabPanel on Android tablets (T1289237)

pomahtri:


packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
```diff
@@ -92,6 +97,11 @@ const subscribeToRowEvents = function (that, $table) {
   eventsEngine.on($table, [clickEventName, dblclickEvent, pointerEvents.down].join(' '), '.dx-row', that.createAction((e) => {
     const { event } = e;
 
+    // NOTE: checking for target only for mocks in qunits
+    if (e?.event?.target && !gridCoreUtils.isElementInCurrentGrid(that, $(event.target))) {
```
wdevfx, line 101: Question.

Why do we need this check `e?.event?.target` for `QUnit`?
As far as I can see, we access `event.target` in `if` below (so, it should work in `QUnit`)
Also the `isElementInCurrentGrid` has check that `$element` is defined -> [link to the code](https://github.com/DevExpress/DevExtreme/blob/7f15865230c52e717d937b64e16ef86c9829afab/packages/devextreme/js/__internal/grids/grid_core/m_utils.ts#L591)
But maybe, I miss sth :)

Anyway, I suggest to add some additional checks inside the `isElementInCurrentGrid`
Because we may use this util function in future in other places and it may be helpful

pomahtri: Yep, for qunits
`e.event` sometimes can be `undefined` when we create event mock on our side
Yep, looks like in exactly this place it's unnecessary, I just copied it from handler above (to keep things persistent)
I don't want to move it inside `isElementInCurrentGrid` as it's not requirement how `isElementInCurrentGrid` should work, it's specific of mocks only
We don't want `isElementInCurrentGrid` return `true` if target is `undefined`
But we want this early return not to trigger if it's `undefined`

wdevfx: @pomahtri tried another solution (mock from QUnit side only) -> unfortunately it doesn't work
Here a few failing (specific on CI) tests, so decided to leave it as is for now

wdevfx (approved):

ksercs (approved):

PR #30406, opened 2025-07-17
DataGrid - FilterRow's searchBox's aria-label should be customizable - T1290381

Tucchhaa:

### DIFF
--- a/e2e/testcafe-devextreme/tests/dataGrid/common/filterRow/filterRow.ts
+++ b/e2e/testcafe-devextreme/tests/dataGrid/common/filterRow/filterRow.ts
@@ -259,3 +259,38 @@ test('DataGrid - The `between` filter dropdown sticks to the viewport edge durin
   columnWidth: 400,
   width: 500,
 }));
+
+// T1290381
+test('DataGrid - filter row\'s search-box\'s aria-label should be customizable via localization', async (t) => {
+  const dataGrid = new DataGrid('#container');
+  const filterEditor = dataGrid.getFilterEditor(0, FilterTextBox);
+
+  await dataGrid.isReady();
+
+  await t
+    .expect(filterEditor.menuButton.getAttribute('aria-label'))
+    .eql('custom text');
+
+  // To reset localization messages
+  await t
+    .eval(() => location.reload());
+}).before(async (t) => {
+  await t.eval(() => {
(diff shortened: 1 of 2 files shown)


e2e/testcafe-devextreme/tests/dataGrid/common/filterRow/filterRow.ts
```diff
@@ -259,3 +259,38 @@ test('DataGrid - The `between` filter dropdown sticks to the viewport edge durin
...
+
+  await dataGrid.isReady();
+
+  await t
+    .expect(filterEditor.menuButton.getAttribute('aria-label'))
+    .eql('custom text');
+
+  // To reset localization messages
+  await t
+    .eval(() => location.reload());
```
wdevfx: Minor comment.

Let's place it to the ```after``` test cafe hook
It's not a big deal, but looks like logically this is action after main test logic :)

e2e/testcafe-devextreme/tests/dataGrid/common/filterRow/filterRow.ts
```diff
@@ -259,3 +259,38 @@ test('DataGrid - The `between` filter dropdown sticks to the viewport edge durin
...
+}).before(async (t) => {
+  await t.eval(() => {
```
wdevfx, line 274: Question.

As far as I understand, ```eval``` = ```ClientFunction```, is it right or there are any other reasons why we use ```eval``` instead of the ```ClientFunction``` here?
[eval docs](https://testcafe.io/documentation/402703/reference/test-api/testcontroller/eval)

wdevfx (dismissed):

e2e/testcafe-devextreme/tests/dataGrid/common/filterRow/filterRow.ts
Tucchhaa: There's no specific reason why I used eval. It just looks small and precise 😃

e2e/testcafe-devextreme/tests/dataGrid/common/filterRow/filterRow.ts
Tucchhaa: Ok, I agree with that. I will change

wdevfx (approved):

PR #30407, opened 2025-07-17
TreeView: extend dts types to specify the type of custom item

anna-shakhova:
Docs:
- https://github.com/DevExpress/devextreme-documentation/pull/7754

### DIFF
--- a/packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/tree_view.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/tree_view.ts
@@ -4,7 +4,9 @@ import dxTreeView from '@js/ui/tree_view';
 
 import { InfernoWrapper } from './widget_wrapper';
 
-export class TreeView<TItem extends TreeViewItemProperties> extends InfernoWrapper<
+export class TreeView<
+  TItem extends TreeViewItemProperties = TreeViewItemProperties,
+> extends InfernoWrapper<
   TreeViewProperties,
   dxTreeView<TItem>
 > {
--- a/packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/tree_view.ts
+++ b/packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/tree_view.ts
@@ -4,7 +4,10 @@ import dxTreeView from '@js/ui/tree_view';
 
 import { InfernoWrapper } from './widget_wrapper';
 
-export class TreeView extends InfernoWrapper<TreeViewProperties, dxTreeView> {
+export class TreeView<TItem extends TreeViewItemProperties> extends InfernoWrapper<
+  TreeViewProperties,
+  dxTreeView<TItem>
+> {
   protected getComponentFabric(): typeof dxTreeView {
     return dxTreeView;
   }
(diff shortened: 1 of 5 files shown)


EugeniyKiyashko (dismissed):

packages/devextreme/js/__internal/grids/new/grid_core/inferno_wrappers/tree_view.ts
```diff
@@ -4,7 +4,10 @@ import dxTreeView from '@js/ui/tree_view';
 
 import { InfernoWrapper } from './widget_wrapper';
 
-export class TreeView extends InfernoWrapper<TreeViewProperties, dxTreeView> {
+export class TreeView<TItem extends TreeViewItemProperties> extends InfernoWrapper<
```
pomahtri: Minor

```suggestion
export class TreeView<TItem = TreeViewItemProperties extends TreeViewItemProperties > extends InfernoWrapper<
```

It's strange that you updated code here but we didn't need to update its usage

pomahtri (dismissed):

pomahtri (dismissed):

packages/devextreme/js/ui/tree_view.d.ts
```diff
@@ -34,16 +34,18 @@ import {
 
 import dxScrollable from './scroll_view/ui.scrollable';
 
+type ItemLike = string | Item | any;
```
marker-dao, line 37: ### 💡 suggestion(non-blocking)
Consider using Record:
```suggestion
type ItemLike = string | Item | Record<string, unknown>;
```

packages/devextreme/js/ui/tree_view.d.ts
```diff
@@ -34,16 +34,18 @@ import {
...
+
 /**
  * @docid _ui_tree_view_ItemInfo
  * @hidden
  */
-export interface ItemInfo<TKey = any> {
+export interface ItemInfo<TItem extends ItemLike, TKey= any> {
```
marker-dao: ### 💡 issue(blocking)
Old behavior is breaking down.
```suggestion
export interface ItemInfo<TItem extends ItemLike = Item, TKey= any> {
```

packages/devextreme/js/ui/tree_view.d.ts
```diff
@@ -72,95 +74,95 @@ export type TreeViewExpandEvent = 'dblclick' | 'click';
  * @type object
  * @inherits EventInfo
  */
-export type ContentReadyEvent<TKey = any> = EventInfo<dxTreeView<TKey>>;
+export type ContentReadyEvent<TItem extends ItemLike = any, TKey = any> = EventInfo<dxTreeView<TItem, TKey>>;
```
marker-dao, line 77: ### 💡 issue(blocking)
Old behavior is breaking down.
```suggestion
export type ContentReadyEvent<TItem extends ItemLike = Item, TKey = any> = EventInfo<dxTreeView<TItem, TKey>>;
```
Also, please check similar places.

packages/devextreme/js/ui/tree_view.d.ts
anna-shakhova: No, not every object is Record, we should not force customer to extend its type from Record, also they may want to use numbers or symbols as keys

packages/devextreme/js/ui/tree_view.d.ts
anna-shakhova: Discussed in thread. Suggested keep `any` for now and consider later replacing it with `Item` in all collection based components

marker-dao (dismissed):

mpreyskurantov (approved):

PR #30409, opened 2025-07-17
T1297188 - FileManager inside popup - W1025

Staziz:

### DIFF
--- a/packages/devextreme/testing/tests/DevExpress.ui.widgets/fileManagerParts/markup.tests.js
+++ b/packages/devextreme/testing/tests/DevExpress.ui.widgets/fileManagerParts/markup.tests.js
@@ -1,5 +1,7 @@
 const { test } = QUnit;
 import $ from 'jquery';
+import { extend } from 'core/utils/extend';
+import consoleUtils from '__internal/core/utils/m_console';
 import 'ui/file_manager';
 import fx from 'common/core/animation/fx';
 import windowUtils from 'core/utils/window';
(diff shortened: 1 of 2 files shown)


mpreyskurantov (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets/fileManagerParts/markup.tests.js
```diff
@@ -207,3 +243,31 @@ QUnit.module('Markup rendering', moduleConfig, () => {
...
 });
+
+QUnit.module('fileManager inside popup not causing any warnings in console (T1297188)', moduleConfig_T1297188, () => {
+    test('height applies to the details view (T1297188)', function(assert) {
+        const originalWarn = consoleUtils.logger.warn;
+        const warnings = [];
+        consoleUtils.logger.warn = function(message) {
+            warnings.push(message);
+            originalWarn.apply(consoleUtils.logger, arguments);
+        };
```
wdevfx: Let's use the `sinon.stub` instead here
Example of usage: [link to the code](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/testing/tests/DevExpress.ui.widgets.form/formLayoutManager.markup.tests.js#L2648)

With this stub we can check that no warnings was logged at console at all
(I guess there should be any warnings, not only this specific one)

So, I suggest the following:

1. Use `sinon.stub` to mock `consoleUtils.logger.warn` method
2. Check that this mock wasn't called at all (no calls -> no warnings)
3. Test case name already good (but pay attention that now it checks only specific warning) -> 
`fileManager inside popup not causing any warnings in console (T1297188)`

packages/devextreme/testing/tests/DevExpress.ui.widgets/fileManagerParts/markup.tests.js
```diff
@@ -207,3 +243,31 @@ QUnit.module('Markup rendering', moduleConfig, () => {
     });
 
 });
+
+QUnit.module('fileManager inside popup not causing any warnings in console (T1297188)', moduleConfig_T1297188, () => {
+    test('height applies to the details view (T1297188)', function(assert) {
```
wdevfx: If u'd like additionally check that `height` applied to the `dxDataGrid`, let's write a test that get FileManager's `dxDataGrid` and check the `option('height')` value :)

But it's optional, from my perspective test that we don't have warnings is enough

packages/devextreme/testing/tests/DevExpress.ui.widgets/fileManagerParts/markup.tests.js
Staziz: There may be other warnings in the console. To make the assert more specific, I used sinon.spy with the same string content check of the arguments it was called with

packages/devextreme/testing/tests/DevExpress.ui.widgets/fileManagerParts/markup.tests.js
Staziz: Thanks for the suggestion, I've added this check

wdevfx (dismissed):

wdevfx (approved):

PR #30424, opened 2025-07-18
[WIP] Code owner tool

aleksei-semikozov:


tools/codeowner/utils/git.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+  });
+
+  console.log('-'.repeat(80));
+}
+
+module.exports = {
+  getGitCommits,
+  displayCommits
+};
```
ksercs, line 99: What are these functions for? Look not related to the main idea

PR #30443, opened 2025-07-21
Use devextreme-exceljs-fork

Tucchhaa:


wdevfx:
About demos urls, we missed the point that our demos urls contains `ExcelJS` direct mention
I guess it's okay and no blocker here, but let's clarify with PMs that it's ok 

E.g link to the demo:
https://js.devexpress.com/jQuery/Demos/WidgetsGallery/Demo/DataGrid/ExcelJSOverview/MaterialBlueLight/
And as another example [demo's redirect.json](https://github.com/DevExpress/DevExtreme/blob/914e2091d247fe5526067d8805150e5ea8ca5627/apps/demos/redirects.json#L26)

wdevfx:
Let's also change `exceljs` to our package in the [webpack.config.js](https://github.com/DevExpress/DevExtreme/blob/764a300f4318f0b9f4dd2aa20f058bcc10bfeed6/packages/devextreme/webpack.config.js#L28)

As far as I understand, here is a list of external dependencies that should not be included in the artifacts bundles
So, just in case better to change it too

wdevfx (approved):

PR #30448, opened 2025-07-21
DataGrid(T1281192 & T1298901) - Error occurs on an attempt to group custom columns at runtime

wdevfx:


packages/devextreme/js/__internal/grids/grid_core/utils/selector_comparison.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  if (typeof selector === 'function' && typeof selectorToCompare === 'function') {
+    return compareCallbacks(selector, selectorToCompare);
+  }
+
+  return false;
+};
+
+export const isSelectorEqualWithCallback = (
```
pomahtri, line 47: Why we need this more strict version of `isEqualSelectors`?

pomahtri (approved):

packages/devextreme/js/__internal/grids/grid_core/utils/selector_comparison.ts
wdevfx: It's a little bit separate case

We have two types of comparison here:
- selector with another selector
- selector with some callback (like calculateCellValue / calculateGroupValue / etc)

Selector may be `string` or `function`, and a callback may be `string` or `function` 
As example here is a doc for the [calculateGroupValue option](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/columns/#calculateGroupValue)
But I'd not like to get unexpected match if selector is a `string` and callback is a `string` too
I understand that it's 1/1kk chance to get it, but from my perspective it's better to be more safe here :)

Also, if in the future we'd like to extend one of these comparison -> it will be more convenient to make change only in related function

wdevfx: Discussed & agreed leave it as is :)

PR #30449, opened 2025-07-21
CardView - Fix selectAll when some filter is applied - T1298718

Tucchhaa:


e2e/testcafe-devextreme/tests/cardView/selection/functional.ts
```diff
@@ -968,3 +968,86 @@ test('Switching the showCheckBoxesMode option from always to onClick at runtime
     showCheckBoxesMode: 'always',
   },
 }));
+
+test('should select only cards matching dataSource.filter()', async (t) => {
```
wdevfx: This test cases checks only API calls (no visual testing part)
So, I suggest to move them in ```jest```, because it will be executed faster + we can cover more test scenarios (write more test cases)

About test cases, let's check the following cases:
- [✅ already exist] data source filter
- [✅ already exist] filter panel
- header filter 
- [🟡 partially exist] combination of all of them

And for each case, let's test in case of:
- initial configuration 
  (from options, for data source case we can create data source with filter before creating the CardView)
- runtime api change (how it already implemented)

Tucchhaa: I agree that we should move them to jest and that we need to include headerFilter to tests.

But I am not sure if we need to add tests for setting filters in runtime and in initialization.

In my perspective there should be separate tests that check that 'filterValue' or 'column.filterValues' are set correctly in both runtime and initialization. 

And then integration tests just test the user scenario without worrying about how initialization/runtime option modification works.

In this way I think we can reduce redundancy in our tests and make them more clean.

There's [headerFilter/options.integration.test.ts](https://github.com/DevExpress/DevExtreme/blob/84ddf0fac87ac05573402361b3d97d434172188b/packages/devextreme/js/__internal/grids/new/grid_core/filtering/header_filter/options.integration.test.ts#L1) to test that options are set correctly at the initialization. However there's no tests to check:
- headerFilter options runtime modification
- filterPanel options initialization
- filterPanel options runtime modification

I suggest that we add tests for the above cases too and for this particular bug just set options in runtime or initialization.

wdevfx: > In my perspective there should be separate tests that check that 'filterValue' or 'column.filterValues' are set correctly in both runtime and initialization.
> And then integration tests just test the user scenario without worrying about how initialization/runtime option modification works.

Yup, agree with u, there should not be any difference in terms of this case
So, I suggest the following:
- Let's move existing tests to the jest
- Let's add headerFilter check and maybe some combinations (but one combination of all is enough from my perspective)

> I suggest that we add tests for the above cases too

Let's do it in terms of a separate tech task, could u please create a tech task for it?
There is a lot of nuances and types of filters, so I guess it should be implemented separately

packages/devextreme/js/__internal/grids/new/grid_core/filtering/public_methods.ts
```diff
@@ -9,10 +9,5 @@ export function PublicMethods<T extends Constructor<GridCoreNewBase>>(GridCore:
     public clearFilter(): void {
       this.filterSyncController.clearFilters();
     }
-
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    public getCombinedFilter(): any {
-      return this.filterController.displayFilter.peek();
-    }
```
Tucchhaa, line 16: This code is overriden [here](https://github.com/DevExpress/DevExtreme/blob/5bb5a0f33ee472f3ce76ff5b49b441d99cdd4f71/packages/devextreme/js/__internal/grids/new/grid_core/data_controller/public_methods.ts#L36)

wdevfx (dismissed):

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.ts
```diff
@@ -361,6 +358,7 @@ export class DataController {
     this.pageIndex.value = dataSource.pageIndex();
     this.pageSize.value = dataSource.pageSize();
     this._totalCount.value = dataSource.totalCount();
+    this._dataSourceFilter.value = dataSource.filter();
```
wdevfx: We update the `filter` value from the `DataSource` only after the data will be loaded

I'm not sure that this behavior is ok
Because until data will be loaded -> we will get old value from the `combinedFilter` signal
The problem part only the `dataSource.filter()`, other filter parts is reactive and up-to date

Now it works differently, we always get the actual `dataSource.filter()` part in `combineFilter`
[CodePen example with 25.1](https://codepen.io/wdevfx/pen/NPGNJmy?editors=1000)
U may click on "get combined filter" to see the actual `combinedFilter` value and the `Resolve data` to complete the data request

But if we will launch same example after these changes -> until data will be loaded the `combinedFilter` will be empty

So, this change doubt me a little bit, because we have a delay in state update and may receive the inconsistent filter value in other features and it may cause unexpected behavior & bugs

Unfortunately, as far as I can see, the DataSource doesn't contain suitable hook for update filter :(
(Sth like `onFilterChanged`)

@pomahtri, @Tucchhaa let's discuss this point separately

wdevfx: Decided to rollback reactive changes

wdevfx (approved):

PR #30473, opened 2025-07-23
Core: renderer_base hasClass method needs rework(fixes 1330)

dmalykh-devexpress:
fixed hasClass logic; hasClass checks all Elements


packages/devextreme/testing/tests/DevExpress.core/renderer.tests.js
```diff
@@ -202,6 +202,46 @@ QUnit.test('class should be removed from SVG', function(assert) {
...
+
+    fixture.appendChild($allInnerElements.get(0));
+
+    assert.ok($allInnerElements.hasClass('one'));
+    assert.ok($allInnerElements.hasClass('two'));
+    assert.ok($allInnerElements.hasClass('three'));
+    assert.notOk($allInnerElements.hasClass('five'));
+});
+
+
```
dxvladislavvolkov, line 256: Do we need to add tests for the following cases?
- When the element collection is empty.
- When checking multiple class names (<div class="a b"> with hasClass('x a y')).

packages/devextreme/js/__internal/core/m_renderer_base.ts
```diff
@@ -147,15 +143,19 @@ initRender.prototype.removeClass = function (className) {
...
-      if ((className || '').split(' ').indexOf(classNames[i]) >= 0) return true;
+
+  for (let i = 0; i < this.length; i++) {
+    if (!this[i] || !this[i].className) return false;
+
+    for (let j = 0; j < classNames.length; j++) {
+      if (this[i].classList) {
+        if (this[i].classList.contains(classNames[j])) return true;
+      } else {
+        // IE9
```
dxvladislavvolkov: Do we still need to keep the IE9 support code? Maybe we can remove it?

packages/devextreme/testing/tests/DevExpress.core/renderer.tests.js
dmalykh-devexpress: Good question 

When checking multiple class names (
with hasClass('x a y')).

If it is expected behavior - I'll add it. I would like to remove it but now it is here cause it was here %)
Why to remove? - because in jQuerry, they do NOT support multiple classes as selector 

I didn't find any example of usage with multiple classes in our code base but for some reason it was implemented. So i decided to leave it as it is 

=====

When the element collection is empty.
Agree , i'll do it

packages/devextreme/js/__internal/core/m_renderer_base.ts
dmalykh-devexpress: Removed

chaosmirage (approved):

dxvladislavvolkov (approved):

PR #30485, opened 2025-07-24
Fix wheel event delta normalization to use deltaY or deltaX for common delta

dmalykh-devexpress:
…eltaX

### DIFF
--- a/packages/devextreme/js/__internal/events/core/m_wheel.ts
+++ b/packages/devextreme/js/__internal/events/core/m_wheel.ts
@@ -64,13 +64,6 @@ const wheel = {
     return -delta * DELTA_MUTLIPLIER;
   },
 
-  /* https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event
-  * wheelDelta - is deprecated.
-  * Use deltaX, deltaY, and deltaZ instead.
-  *
-  * jQuery implementation of mousewheel event.
-  * https://github.com/jquery/jquery-mousewheel/blob/main/src/jquery.mousewheel.js#L83
-  */
   _getWheelDelta(deltaY: number, deltaX: number, deltaZ: number) {
     if (deltaY) {
       return deltaY;
--- a/packages/devextreme/js/__internal/events/core/m_wheel.ts
+++ b/packages/devextreme/js/__internal/events/core/m_wheel.ts
@@ -61,7 +61,7 @@ const wheel = {
     }
     // Use multiplier to get rough delta value in px for the LINE or PAGE mode
     // https://bugzilla.mozilla.org/show_bug.cgi?id=1392460
-    return -delta * DELTA_MULTIPLIER;
+    return -DELTA_MULTIPLIER * delta;
   },
 
   _getWheelDelta(deltaY: number, deltaX: number) {
--- a/packages/devextreme/js/__internal/events/core/m_wheel.ts
+++ b/packages/devextreme/js/__internal/events/core/m_wheel.ts
@@ -10,6 +10,19 @@ const NATIVE_EVENT_NAME = 'wheel';
 const PIXEL_MODE = 0;
 const DELTA_MUTLIPLIER = 30;
 
+enum DeltaMode {
+  DOM_DELTA_PIXEL = 0,
+  DOM_DELTA_LINE = 1,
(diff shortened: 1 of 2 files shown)


dxvladislavvolkov (dismissed):

packages/devextreme/js/__internal/events/core/m_wheel.ts
```diff
@@ -40,13 +55,25 @@ const wheel = {
...
 
-  _normalizeDelta(delta, deltaMode = PIXEL_MODE) {
+  _normalizeDelta(delta: number, deltaMode = PIXEL_MODE) {
     if (deltaMode === PIXEL_MODE) {
       return -delta;
     }
     // Use multiplier to get rough delta value in px for the LINE or PAGE mode
     // https://bugzilla.mozilla.org/show_bug.cgi?id=1392460
-    return -DELTA_MUTLIPLIER * delta;
+    return -delta * DELTA_MUTLIPLIER;
```
chaosmirage: What is the principal difference between current implementation `-DELTA_MUTLIPLIER * delta` and the previous one `-delta * DELTA_MUTLIPLIER` ?

packages/devextreme/testing/tests/DevExpress.ui.events/wheel.tests.js
```diff
@@ -79,3 +79,25 @@ QUnit.test('normalize delta for deltaMode LINE and PAGE', function(assert) {
...
+    let eventArgs;
+
+    $element.on(wheelEvent.name, function(e) {
+        eventArgs = e;
+    });
+
+    mouse.wheel(10, { deltaY: 10, deltaX: 5, deltaZ: 0 });
+    assert.strictEqual(eventArgs.delta, -10, 'vertical scroll should use deltaY');
+
+    mouse.wheel(10, { deltaY: 0, deltaX: 5, deltaZ: 0 });
```
chaosmirage: Why is the `10` passed as the first argument? Seems like it was overridden via `deltaY`.

`mouse.wheel(0, { deltaX: 5, deltaZ: 0 })`

Will this work?

packages/devextreme/js/__internal/events/core/m_wheel.ts
```diff
@@ -40,13 +55,25 @@ const wheel = {
...
+  },
+
+  _getWheelDelta(deltaY: number, deltaX: number, deltaZ: number) {
+    if (deltaY) {
+      return deltaY;
+    }
+
+    if (deltaX && deltaZ === 0) {
```
chaosmirage: Please add a purpose explanation via a comment or extract the expression to a variable with a semantic name.

chaosmirage (changes requested):

packages/devextreme/js/__internal/events/core/m_wheel.ts
dmalykh-devexpress: There is no difference. It was made just to highlight the common logic for those lines.

```
      return -delta;
```
      
```
       return -delta * DELTA_MUTLIPLIER;
```   
       
       PS: To be honest I didn't want to commit it :)  
       If it is critical I'll revert it

chaosmirage: Yes, please revert

dmalykh-devexpress: I reverted this line back but fixed typo DELTA_MU**TL**IPLIER -> DELTA_MU**LT**IPLIER

So that this change still there

packages/devextreme/testing/tests/DevExpress.ui.events/wheel.tests.js
dmalykh-devexpress: Yep, you are right

packages/devextreme/js/__internal/events/core/m_wheel.ts
dmalykh-devexpress: Thanks
I checked jQuerry implementation again and noticed that I had mistaken. Now that function is simplified. 

// deltaZ is not needed

chaosmirage (approved):

dxvladislavvolkov (approved):

PR #30499, opened 2025-07-25
DataGrid(T1293295) - Individually selecting all header filter options deselects all options

wdevfx:


EugeniyKiyashko (dismissed):

packages/testcafe-models/list/checkBox.ts
```diff
@@ -11,9 +12,28 @@ export default class CheckBox {
...
+    switch (true) {
+      case isChecked && !isIndeterminate:
+        return 'checked';
+        case !isChecked && !isIndeterminate:
+          return 'unchecked';
+      case !isChecked && isIndeterminate:
+        return 'indeterminate';
+      default:
+        throw Error(`Invalid checkbox state. checked = ${isChecked}, indeterminate = ${isIndeterminate}`);
+    }
```
pharret31: I'd suggest to use if blocks here , since it looks more natural:

```js
if (isChecked && !isIndeterminate) {
  return 'checked';
}
if (!isChecked && !isIndeterminate) {
  return 'unchecked';
}
if (!isChecked && isIndeterminate) {
  return 'indeterminate';
}
throw Error(`Invalid checkbox state. checked = ${isChecked}, indeterminate = ${isIndeterminate}`);
```

packages/testcafe-models/list/checkBox.ts
```diff
@@ -11,9 +12,28 @@ export default class CheckBox {
...
+
+  async getCheckBoxState(): Promise<'checked' | 'unchecked' | 'indeterminate'> {
+    const isChecked = await this.isChecked;
+    const isIndeterminate = await this.isIndeterminate;
+
+    switch (true) {
+      case isChecked && !isIndeterminate:
+        return 'checked';
+        case !isChecked && !isIndeterminate:
+          return 'unchecked';
```
pharret31: Indents have two more spaces than needed

packages/testcafe-models/list/checkBox.ts
wdevfx: > since it looks more natural:

Let's avoid such arguments, because it's a little bit holy war topic :D
From ur perspective `if` is better, from mine `switch(true)` is perfect, from someone else `if -> else if` is perfect option.
But all of these options do the same work and there is no any real difference, it's just a matter of habit and convenience for everyone

So, the better solution here -> have a `code convention` inside a team (or even better on the whole tribe lvl) where one option picked after voting and other options banned by ESLint :)

Since we don't have this one and this `POM` isn't `Grids team` responsibility zone -> let's apply here `if` option

packages/testcafe-models/list/checkBox.ts
wdevfx: Yup, strange that it was passed by `eslint --fix` & auto-formatting, will take a look later

pharret31 (approved):

e2e/testcafe-devextreme/tests/dataGrid/common/headerFilter/headerFilterList.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  return { list, firstListItem, secondListItem };
+};
+
+test('Should has unchecked "Select all" checkbox state if no values is selected', async (t) => {
+  const dataGrid = new DataGrid(GRID_CONTAINER);
+  const { list, firstListItem, secondListItem } = await openHeaderFilterAndGetList(t, dataGrid);
+
+  await t
+    .expect(await list.selectAll.checkBox.getCheckBoxState())
```
Raushen, line 32: The best practice is not to write await in the expect method. Writing so there is only one expectation check and may lead to unstable tests.
Expect method already has awaiting under the hood and available to check passed promise several times until it is true.

Raushen (approved):

PR #30507, opened 2025-07-28
LoadIndicator: Publish animationType option

marker-dao:


packages/devextreme/js/__internal/ui/m_load_indicator.ts
```diff
@@ -114,10 +113,10 @@ class LoadIndicator extends Widget<LoadIndicatorProperties> {
     this.$element().append(this._$wrapper);
   }
 
-  _getAnimationTypeContentClass(): (typeof ANIMATION_TYPE_CLASSES)[AnimationType] {
-    const { _animationType: animationType } = this.option();
+  _getAnimationTypeContentClass(): (typeof ANIMATION_TYPE_CLASSES)[AnimationType] | undefined {
+    const { animationType } = this.option();
```
aleksei-semikozov, line 117: It's can be undefined or null in option?

Zedwag (approved):

packages/devextreme/js/__internal/ui/m_load_indicator.ts
ksercs: If this code is related to the specificity that we're forced to write `| undefined` in type, maybe it's better to enforce `?? 'circle'` here or smth like this that to not worsen types?

marker-dao: > It's can be undefined or null in option?

According to our type system (as it has historically developed), all options are optional, meaning they can be undefined. 

> maybe it's better to enforce ?? 'circle'

The default value is ‘circle’. In my opinion, we are trying to hide the TS problem with JS code. Perhaps it is better that undefined is returned here, because if the user sets the value to undefined, it will allow them not to set the animation CSS class on the component. In any case, let's synchronize our vision on this issue; I am not opposed to either approach.

ksercs (approved):

aleksei-semikozov (approved):

mpreyskurantov (approved):

PR #30542, opened 2025-07-30
DateBox: Fix error on value change when empty string is set as initial value (T1301310)

r-farkhutdinov:

### DIFF
--- a/packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
+++ b/packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
@@ -13,7 +13,6 @@ import { inputType } from '@js/core/utils/support';
 import { isDate as isDateType, isNumeric, isString } from '@js/core/utils/type';
 import { getWindow, hasWindow } from '@js/core/utils/window';
 import type {
-  DateLike,
   DatePickerType, DateType, Properties,
 } from '@js/ui/date_box';
 import type { OptionChanged } from '@ts/core/widget/types';
@@ -152,29 +151,8 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
     ]);
   }
 
-  // eslint-disable-next-line class-methods-use-this
-  _sanitizeValue(value?: DateLike): DateLike | undefined {
-    if (value === '') {
-      return null;
-    }
-
-    return value;
-  }
-
-  _sanitizeValueOption(args: OptionChanged<DateBoxBaseProperties>):
-  OptionChanged<DateBoxBaseProperties> {
-    const sanitizedArgs = { ...args };
-    const sanitizedValue = this._sanitizeValue(sanitizedArgs.value);
-
-    sanitizedArgs.value = sanitizedValue;
-
-    this.option({ value: sanitizedValue });
-
-    return sanitizedArgs;
-  }
-
   _initOptions(options: DateBoxBaseProperties): void {
-    this._userOptions = { ...options, value: this._sanitizeValue(options.value) };
+    this._userOptions = extend({}, options);
     super._initOptions(options);
     this._updatePickerOptions();
   }
@@ -773,9 +751,6 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
       case 'showAnalogClock':
       case '_showValidationIcon':
         break;
-      case 'value':
-        super._optionChanged.apply(this, [this._sanitizeValueOption(args)]);
-        break;
       default:
         // @ts-expect-error ts-error
         super._optionChanged.apply(this, arguments);
@@ -783,7 +758,7 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
   }
 
   _getSerializationFormat() {
-    const value = this.option('value');
+    const { value } = this.option();
 
     if (this.option('dateSerializationFormat') && config().forceIsoDateParsing) {
       return this.option('dateSerializationFormat');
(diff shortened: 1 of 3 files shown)


packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
```diff
@@ -151,8 +152,17 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
     ]);
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  _initValueOption(value?: DateLike): DateLike | undefined {
+    if (isString(value) && !value.length) {
```
aleksei-semikozov: `value === ''`
I think, it clear

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
```diff
@@ -151,8 +152,29 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
...
+  }
+
+  _sanitizeValueOption(args: OptionChanged<DateBoxBaseProperties>):
+  OptionChanged<DateBoxBaseProperties> {
+    const sanitizedArgs = { ...args };
+    const sanitizedValue = this._sanitizeValue(sanitizedArgs.value);
+
+    sanitizedArgs.value = sanitizedValue;
+
+    this.option({ value: sanitizedValue });
```
aleksei-semikozov: I see a recursive call here 
`option(value) -> _optionChanged(switch) -> option(value)`
Is this necessary?

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
```diff
@@ -751,6 +773,9 @@ class DateBox extends DropDownEditor<DateBoxBaseProperties> {
       case 'showAnalogClock':
       case '_showValidationIcon':
         break;
+      case 'value':
+        super._optionChanged.apply(this, [this._sanitizeValueOption(args)]);
```
aleksei-semikozov: In this case we are use only value option. I think it's clear
`this._sanitizeValueOption(args.value)`

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js
```diff
@@ -6164,6 +6164,44 @@ QUnit.module('DateBox number and string value support', {
             assert.ok(true, 'there is no error');
         }
     });
+
+    [true, false].forEach(isRuntime => {
```
aleksei-semikozov, line 6168: Do we really like tests where we create them on the fly?

I don't like them.

If the code gets too big, maybe it's better to make a helper that will help us click.

```
const clickOnApplyButton = () => {
    const $calendarCell = $(`.${CALENDAR_CELL_CLASS}`).eq(0);
    $calendarCell.trigger('dxclick');

    const $applyButton = $(dateBox.content()).parent().find(APPLY_BUTTON_SELECTOR);
    $applyButton.trigger('dxclick');
}
```

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
r-farkhutdinov: Unfortunately, it is as option handling is performed on Component level. This means that even if I sanitize the value, it's still stored as `''` internally. To avoid this, I set it manually here. 

According to recursion: the maximum depth is fixed actually, as the value set here does not fall into condition that checks the same option is passed.

For example, setting `value: ''` will trigger just one extra call. 
```
const dateBox = $('#container1').dxDateBox({
    value: '',
    type: 'datetime',
    onInitialized: () => {
        setTimeout(() => {
            dateBox.option('value', '');
        }, 10);
    }
}).dxDateBox('instance');
```

image

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js
r-farkhutdinov: It's a common practice for tests where the steps are repeated. Even if I use helper functions, the calls and orders will be almost the same. So I would prefer to leave it as is.

packages/devextreme/js/__internal/ui/date_box/m_date_box.base.ts
r-farkhutdinov: It is, but in this case I will have to mutate `args` here in `_optionChanged` handler, so I decided to move all the transformations to a separate function. In the future, if some other edge cases occur, the name can be changed and new checks can be added easily. I would also prefer to leave as is.

Zedwag (approved):

aleksei-semikozov (approved):

marker-dao (approved):

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/datebox.tests.js
ksercs: Actually I'm also a bit worried about non intuitiveness of this approach. It's a common practice to use forEach for iterating through props (states) if the test itself is the same. However, it's strange to see it for uniting of 2 different scenarios to the one test (behavior on init and behavior on runtime change). These scenarios are processed by different code, tested by different actions. Looks like copypaste in this case is better then decreasing of a clarity

P.s. I don't push updating exactly this test, it's a minor

PR #30558, opened 2025-07-31
update fr.json with French translations

DuckTieCorpMember:


AlisherAmonulloev (changes requested):
The document contains formatting issues. We have 4 spaces instead of 2. Ideally, we'd like to see only the updated localization keys

PR #30563, opened 2025-07-31
HTML editor: rework quill module integration types

aleksei-semikozov:

### DIFF
--- a/packages/devextreme/js/__internal/ui/html_editor/converters/m_delta.ts
+++ b/packages/devextreme/js/__internal/ui/html_editor/converters/m_delta.ts
@@ -1,15 +1,16 @@
 import ConverterController from '../m_converterController';
+import type { Delta, QuillInstance } from '../types/quill';
 
 class DeltaConverter {
-  quillInstance: any;
+  quillInstance: QuillInstance | null = null;
 
-  setQuillInstance(quillInstance) {
+  setQuillInstance(quillInstance: QuillInstance): void {
     this.quillInstance = quillInstance;
   }
 
-  toHtml() {
+  toHtml(): string | undefined {
     if (!this.quillInstance) {
-      return;
+      return undefined;
     }
 
     return this._isQuillEmpty()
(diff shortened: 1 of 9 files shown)


packages/devextreme/js/__internal/ui/html_editor/converters/m_delta.ts
```diff
@@ -1,30 +1,40 @@
...
     const delta = this.quillInstance.getContents();
 
-    return delta.length() === 1 && this._isDeltaEmpty(delta);
+    return delta.length() === 1 && DeltaConverter._isDeltaEmpty(delta);
   }
 
-  _isDeltaEmpty(delta) {
-    return delta.reduce((__, { insert }) => insert.indexOf('\n') !== -1);
+  private static _isDeltaEmpty(delta: Delta): boolean {
+    return delta.reduce<boolean>((_, operation) => {
```
ksercs, line 32: I think that `some` is more suitable for this:
```javascript
return delta.some(operation =>
    typeof operation.insert === 'string' && operation.insert.includes('\n')
);
```

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
```diff
@@ -367,10 +366,11 @@ class HtmlEditor extends Editor<Properties> {
     if (isFunction(customizeModules)) {
       customizeModules(modulesConfig);
     }
+    const { placeholder, readOnly, disabled } = this.option();
 
     this._quillInstance = this._getRegistrator().createEditor(this._$htmlContainer[0], {
-      placeholder: this.option('placeholder'),
-      readOnly: this.option('readOnly') || this.option('disabled'),
+      placeholder,
+      readOnly: Boolean(readOnly ?? disabled),
```
ksercs: It's a regression.
Previously, case "readonly=false, disabled=true" worked another way

packages/devextreme/js/__internal/ui/html_editor/types/quill.ts
```diff
@@ -0,0 +1,214 @@
...
+    SILENT: string;
+    USER: string;
+  };
+
+  new (container: string | Element, options?: QuillOptions): QuillInstance;
+}
+
+export type Quill = QuillStatic;
+
+export default Quill;
```
ksercs, line 214: I would say it's better to implement these types directly in devextreme-quill repo + maybe reused existing external types like [these ones](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/quill/index.d.ts). However, I agree it's more rational to do this work separately, so looks good enough for now

packages/devextreme/js/__internal/ui/html_editor/converters/m_delta.ts
aleksei-semikozov: Delta is not Array
https://github.com/slab/delta/blob/dc17ca03e1d68ee729c8cd1ff790ebf96eb0fbec/src/Delta.ts#L195

packages/devextreme/js/__internal/ui/html_editor/html_editor.ts
aleksei-semikozov: My bad. Fix it

ksercs (approved):

PR #30577, opened 2025-08-01
DataGrid - Fix focusedRowKey with grouping, when grouped by 'null' - T1282925

Tucchhaa:


packages/devextreme/js/__internal/grids/data_grid/focus/m_focus.ts
```diff
@@ -56,18 +56,15 @@ const data = (Base: DataControllerBase) => class FocusDataControllerExtender ext
     return true;
   }
 
-  private _getGroupPath(groupItem, groupCount) {
+  private _getGroupPath(groupItems) {
     const groupPath: any[] = [];
-    let items = [groupItem];
+    let item = groupItems[0];
 
-    while (items && items[0] && groupCount) {
```
Alyar666, line 63: I think we shouldn't have removed the check for groupCount. The data may have its own items fields. In this case, groupPath will be generated incorrectly.

packages/devextreme/js/__internal/grids/data_grid/focus/m_focus.ts
```diff
@@ -56,18 +56,15 @@ const data = (Base: DataControllerBase) => class FocusDataControllerExtender ext
...
-      const item = items[0];
-      if (item.key !== undefined) {
-        groupPath.push(item.key);
-      }
-      items = item.items;
-      groupCount--;
+    while (item && item.items) {
+      groupPath.push(item.key);
+      [item] = item.items;
```
Alyar666: Minor comment: [item] = item.items - not very readable code) I would like it to be simple: item = item.items[0]

Alyar666 (changes requested):

Alyar666 (approved):

PR #30595, opened 2025-08-01
Demos. Fix version for rehype-parse

GoodDayForSurf:
(for CodeSandbox compatibility)


mpreyskurantov (changes requested):
What about renovate?
https://github.com/DevExpress/DevExtreme/pull/29729

mpreyskurantov (dismissed):

alexslavr (dismissed):

apps/demos/Demos/HtmlEditor/MarkdownSupport/Vue/App.vue
```diff
@@ -69,6 +69,7 @@ const converter = {
   },
   fromHtml(value) {
     const result = unified()
+      // @ts-expect-error
```
marker-dao: If I understand correctly, your error is caused by a mismatch between the versions of unified and rehypeParse. unified().use() expects you to pass a Plugin of a specific type, but the rehypeParse 8.0.5 version is typed differently. Most likely, to solve this problem, you also need to freeze the unified version (and it would be good if this were consistent with all other plugins).

mpreyskurantov (approved):

alexslavr (approved):

marker-dao (approved):

PR #30601, opened 2025-08-04
DataGrid - Tab key should activate editor with virtual scrolling enabled (T1290811)

Tucchhaa:

### DIFF
--- a/e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/startEditing.functional.ts
+++ b/e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/startEditing.functional.ts
@@ -1,4 +1,5 @@
 import DataGrid from 'devextreme-testcafe-models/dataGrid';
+import { ClientFunction } from 'testcafe';
 import url from '../../../../helpers/getPageUrl';
 import { createWidget } from '../../../../helpers/createWidget';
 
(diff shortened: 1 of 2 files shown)


e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/startEditing.functional.ts
```diff
@@ -36,6 +37,84 @@ test('Editing should start by pressing enter after scrolling content with scroll
...
+    .click(dataGrid.getDataCell(49, 0).element)
+    .expect(dataGrid.getDataCell(49, 0).getEditor().element.focused).ok();
+
+  await t
+    .pressKey('tab')
+    .expect(dataGrid.getDataCell(49, 1).getEditor().element.focused).ok();
+
+  const eventRowKeys = await ClientFunction(() => (window as any).eventRowKeys)();
+
+  await t.expect(eventRowKeys).eql([49, 49, 49, 49, 49, 49]);
```
Alyar666: Why was the allowUpdating callback called 6 times? The check is not obvious. If we don’t care about the number of times this callback is called, maybe it’s enough to check that there are no invalid values among the keys?
Or we could add only unique keys to the eventRowKeys array and check those.

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/startEditing.functional.ts
```diff
@@ -36,6 +37,84 @@ test('Editing should start by pressing enter after scrolling content with scroll
...
+    .click(dataGrid.getDataCell(48, 1).element)
+    .expect(dataGrid.getDataCell(48, 1).getEditor().element.focused).ok();
+
+  await t
+    .pressKey('tab')
+    .expect(dataGrid.getDataCell(49, 0).getEditor().element.focused).ok();
+
+  const eventRowKeys = await ClientFunction(() => (window as any).eventRowKeys)();
+
+  await t.expect(eventRowKeys).eql([48, 48, 48, 48, 49, 49, 49]);
```
Alyar666: The sameSame here. See the comment above.

Alyar666 (changes requested):

Alyar666 (approved):

PR #30603, opened 2025-08-04
Menu: remove toggle focus on hover

anna-shakhova:


ksercs (dismissed):

packages/devextreme/js/__internal/ui/menu/menu.ts
```diff
@@ -412,44 +413,41 @@ class Menu extends MenuBase<MenuProperties> {
     return this._hamburger.$element();
   }
 
-  _toggleTreeView(state: boolean): void {
-    if (isPlainObject(state)) {
-      const { visible } = this._overlay?.option() ?? {};
+  _toggleTreeView(state: boolean | ButtonClickEvent): void {
```
EugeniyKiyashko: I would prefer refactor this part and don't expect ButtonClickEvent object here. It allow us to simplify implementation and remove extra conditions inside

EugeniyKiyashko (approved):

PR #30605, opened 2025-08-04
Move js to ts

dmalykh-devexpress:


packages/devextreme/js/__internal/common/core/animation/presets/m_presets.ts
```diff
@@ -0,0 +1,415 @@
...
+        options: {
+          defaultStaggerAnimationDuration: 350,
+          defaultStaggerAnimationDelay: 50,
+          defaultStaggerAnimationStartDelay: 0,
+        },
+      },
+      { // T254756
+        device: function () {
+          // @ts-expect-error
+          return devices.current().android ?? devices.real.android;
```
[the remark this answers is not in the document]
dmalykh-devexpress: It was changed cause eslint rule. Tests are passed. If someone alive says that we need to revert it back for some reasons I will do it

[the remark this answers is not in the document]
pomahtri: Please do it, we indeed had some problems with such changes previously. I think we will delete this linter rule in future

packages/devextreme/js/core/dom_adapter.d.ts
```diff
@@ -21,6 +21,7 @@ export interface DomAdapter {
   getRootNode: (element: HTMLElement) => Document | DocumentFragment;
   getAttribute: (element, name) => any;
   setAttribute: (element, name, value) => void;
+  createAttribute: (text: any, context?: any) => any;
```
anna-shakhova: is there necessity to use such wide type here? in implementation we define it stricter
```
  createAttribute: (text: string, context?: Document) => Attr;
```

dmalykh-devexpress: Agree, fixed

anna-shakhova (approved):

pomahtri (approved):

packages/devextreme/js/__internal/core/m_dom_adapter.ts
```diff
@@ -88,8 +89,10 @@ const nativeDOMAdapterStrategy: DomAdapter = {
     return context.createTextNode(text);
   },
 
-  createAttribute(text, context) {
+  createAttribute(text: string, context?: Document) {
+    // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
     context = context || this._document;
+    // @ts-expect-error
```
marker-dao, line 95: ### 💡 chore(minor, not necessarily update)
Please in the future leave a small comment why ts-expect-error is used, like here: https://github.com/DevExpress/DevExtreme/pull/30640/files#diff-fbf4669db3de9dab39b424dbcc2b512f2ab25b81ae908b4d282d3a52458b3ba0R151

marker-dao (approved):

aleksei-semikozov (approved):

PR #30608, opened 2025-08-04
Menu & ContextMenu: dts allow customize menu item type

anna-shakhova:


packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
```diff
@@ -791,7 +794,9 @@ class MenuBase<
   }
 
   selectItem(itemElement: Element | dxMenuBaseItem): void {
-    const itemData = itemElement.nodeType ? this._getItemData(itemElement as Element) : itemElement;
+    const isElement = (item: Element | dxMenuBaseItem): item is Element => 'nodeType' in item && !!item.nodeType;
```
EugeniyKiyashko: Can you use domAdapter.isNode() method instead?

packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
```diff
@@ -727,7 +730,7 @@ class MenuBase<
             this._toggleItemSelection(selectedNode, false);
           }
           this._toggleItemSelection(node, true);
-          this._updateSelectedItems();
+          this._changeSelectedItems();
```
EugeniyKiyashko: this._updateSelectedItems() is implemented in parent components. The selection module calls it on SelectionChanged when args.addedItemKeys.length or args.removedItemKeys.length is non-zero. I see that the options for this are not public in Menus, but if you are sure that all processing of selected items happens on the MenuBase side and the selection module shouldn’t perform any updates, we need to override _updateSelectedItems() with an empty function

packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
```diff
@@ -639,8 +641,9 @@ class MenuBase<
 
     const $itemElement = this._getItemElementByEventArgs(event);
     const link = $itemElement?.find(`.${ITEM_URL_CLASS}`)[0];
+    const isUrlItem = (item: Item | dxMenuBaseItem | undefined): item is Item => !!item && 'url' in item && !!item.url;
```
EugeniyKiyashko: Suggestion: extract this function in class method

packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
```diff
@@ -413,7 +415,7 @@ class MenuBase<
 
   // TODO: try to simplify
```
EugeniyKiyashko: I suppose this comment is outdated

packages/devextreme/js/__internal/ui/chat/messagelist.ts
```diff
@@ -290,6 +290,7 @@ class MessageList extends Widget<Properties> {
         icon: 'edit',
         text: editText,
         disabled: isEditActionDisabled(message),
+        // @ts-expect-error ts-error
```
EugeniyKiyashko, line 293: A CollectionItem can include an onClick property (see the implementation in collection.base), but that property isn’t public. I think we should use our internal type for contextMenuItem instead of the public ContextMenuItem here, and our internal type should be extended with onClick property, after this the error will disappear

anna-shakhova: let's extend internal type in separate pr

packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
anna-shakhova: domAdapter.isNode more strict, but it looks is may be used here in future, when it will be transformed in type guard - cause now it does not narrow itemElement type

packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
anna-shakhova: reverted, it is too complicated for public api update pr, will be solved in separate one

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.ts
```diff
@@ -15,6 +15,11 @@ export interface ContextInfo {
   cardIndex?: number;
 }
 
+export interface ExtendedContextMenuItem extends ContextMenuItem {
```
marker-dao: [suggestion]
```suggestion
export interface ControllerContextMenuItem extends ContextMenuItem {
```

packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
```diff
@@ -631,6 +636,10 @@ class MenuBase<
     e._skipHandling = true;
   }
 
+  _isUrlItem(item: Item | dxMenuBaseItem | undefined): item is Item {
+    return !!item && 'url' in item && !!item.url;
```
marker-dao, line 640:
```suggestion
    return Boolean(item?.url);
```

packages/devextreme/js/ui/menu.d.ts
```diff
@@ -333,8 +337,6 @@ export interface MenuBasePlainItem extends CollectionWidgetItem {
    * @public
    */
   visible?: boolean;
-
-  [key: string]: any;
```
marker-dao, line 337: [question]
Is it expected?

anna-shakhova: yes, we remove this WA with spoiling our type, cause we provided better way to use custom Item type

packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
anna-shakhova: This won't work - MenuBaseItem does not have url property
So type guard should be written that way - first we ensure it is defined, then, it has property `url` and its value not empty

packages/devextreme/js/__internal/grids/new/card_view/context_menu/controller.ts
anna-shakhova: renamed using component name where ContextMenu is used

marker-dao (dismissed):

EugeniyKiyashko (approved):

mpreyskurantov (approved):

pomahtri (approved):

Zedwag (approved):

PR #30617, opened 2025-08-05
DataGrid: Fix extra editCellTemplate/onEditorPreparing call after clicking on a cell in a different row and column (T1291087)

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_cell_based.ts
```diff
@@ -235,35 +235,30 @@ const editingControllerExtender = (Base: ModuleType<EditingController>) => class
   }
 
   private _editCellCore(options) {
-    const dataController = this._dataController;
-    const isEditByOptionChanged = isDefined(options.oldColumnIndex) || isDefined(options.oldRowIndex);
+    const editCellOptions: any = this._getNormalizedEditCellOptions(options);
```
wdevfx: Refactoring suggestion

Let's try to add a type for the `_getNormalizedEditCellOptions`
As far as I can see, it [returns object with fields](https://github.com/DevExpress/DevExtreme/blob/f318b82def5601a5a20fcf9bf653941639d59fcf/packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_cell_based.ts#L322) that we can type
(this object has fields with `Column`, `Item` and `number` types)

So, looks like that we can create a return type here easy-peasy and get rid of the `any`

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_cell_based.ts
```diff
@@ -235,35 +235,30 @@ const editingControllerExtender = (Base: ModuleType<EditingController>) => class
   }
 
   private _editCellCore(options) {
```
wdevfx: Refactoring suggestion

Let's add return type to the `_editCellCore`
It's hard to make a right type for `options`, but adding the return type is easy -> it's `boolean | undefined`

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_cell_based.ts
```diff
@@ -324,22 +319,40 @@ const editingControllerExtender = (Base: ModuleType<EditingController>) => class
...
-      params.key = item.key;
-    }
+  private setEditCell(
+    key: unknown,
+    columnIndex: number,
+    silent: boolean,
+  ): void {
+    this._setEditRowKey(key, silent);
+    this._setEditColumnNameByIndex(columnIndex, silent);
+  }
```
wdevfx: Refactoring suggestion.

Let's get rid of this private method
As far as I can see we use it only once (a little below this place)

So, from my perspective it will be easier to read & understand if we just inline calls of the `_setEditRowKey` and `_setEditColumnNameByIndex` instead of creating an extra wrapper above them :)

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_cell_based.ts
```diff
@@ -324,22 +319,40 @@ const editingControllerExtender = (Base: ModuleType<EditingController>) => class
...
+
+  private _prepareEditCell(params): boolean {
+    const {
+      item,
+      column,
+      oldColumn,
+      columnIndex,
+      oldRowIndex,
+    } = params;
```
wdevfx: Refactoring suggestion.

Let's use here a return type from the `_getNormalizedEditCellOptions` (see related comment above)
Btw, we can use `destructuring` feature inline in the method arguments:
```ts
private _prepareEditCell({
  item,
  column,
  oldColumn,
  columnIndex,
  oldRowIndex,
}): boolean {
  // method code
}
```

packages/devextreme/js/__internal/grids/grid_core/validating/m_validating.ts
```diff
@@ -710,12 +710,13 @@ export const validatingEditingExtender = (Base: ModuleType<EditingController>) =
     super._validateEditFormAfterUpdate.apply(this, arguments as any);
   }
 
-  private _prepareEditCell(params) {
+  private _prepareEditCell({ column, item }): boolean {
     // @ts-expect-error
-    const isNotCanceled = super._prepareEditCell.apply(this, arguments as any);
+    const isNotCanceled: boolean = super._prepareEditCell.apply(this, arguments as any);
```
wdevfx: Refactoring suggestion.

Let's get rid of the `@ts-expect-error` and `apply` usage here

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing_cell_based.ts
```diff
@@ -324,22 +319,40 @@ const editingControllerExtender = (Base: ModuleType<EditingController>) => class
...
-    if (!item.isNewRow) {
-      params.key = item.key;
-    }
+  private setEditCell(
+    key: unknown,
+    columnIndex: number,
+    silent: boolean,
+  ): void {
+    this._setEditRowKey(key, silent);
+    this._setEditColumnNameByIndex(columnIndex, silent);
```
wdevfx: Question.

I have one doubt about this change
If we update option silently, would the public `onOptionChanged` triggered?

If not, it may be a potential new issue
The following case:
1. I add `onOptionChange` to the my `DataGrid` configuration
2. In this `onOptionChange` I catch `editing.editRowKey` updates for my application needs
3. I repeat this issue UI actions, and if `onOptionChange` not triggered -> I miss the actual `editing.editRowKey` value

Alyar666: The public event [onOptionChanged](https://github.com/DevExpress/DevExtreme/blob/8cca786b95aa0903bb771e90cd2dfddd2fc9862c/packages/devextreme/js/__internal/core/widget/component.ts#L328) will be triggered (I checked it). Only the call to the base method [_optionChanged](https://github.com/DevExpress/DevExtreme/blob/8cca786b95aa0903bb771e90cd2dfddd2fc9862c/packages/devextreme/js/__internal/core/widget/component.ts#L332C16-L332C30) is rejected. We override this method in the grid [here](https://github.com/DevExpress/DevExtreme/blob/8cca786b95aa0903bb771e90cd2dfddd2fc9862c/packages/devextreme/js/__internal/grids/grid_core/m_widget_base.ts#L50).

wdevfx (approved):

PR #30620, opened 2025-08-05
fix: Multiple root elements are not supported in React elements (T1300588)

VasilyStrelyaev:


ajivanyandev (dismissed):

dxvladislavvolkov (dismissed):

packages/devextreme-react/src/core/template-wrapper.tsx
```diff
@@ -58,23 +57,28 @@ const TemplateWrapperComponent: FC<TemplateWrapperProps> = ({
...
-    if (element.current) {
-      events.off(element.current, DX_REMOVE_EVENT, onTemplateRemoved);
-    }
-
-    if (removalListenerElement.current) {
-      events.off(removalListenerElement.current, DX_REMOVE_EVENT, onTemplateRemoved);
-    }
-
-    onRemoved();
+    Array.prototype.forEach.call(
```
GoodDayForSurf: [].forEach is more readable, let's use that way, pls

packages/devextreme-react/src/core/template-wrapper.tsx
```diff
@@ -83,18 +87,23 @@ const TemplateWrapperComponent: FC<TemplateWrapperProps> = ({
...
+      Array.prototype.forEach.call(
+        [
+          ...elements.current,
+          hiddenNodeElement.current,
+          removalListenerElement.current,
+        ],
+        (el: HTMLElement | undefined) => safeAppend(el),
+      );
+
+      if (elementNodes.length) {
```
GoodDayForSurf, line 99: this "if" is redundant cos foreEach is not called if array is empty

ajivanyandev (approved):

dxvladislavvolkov (approved):

GoodDayForSurf (approved):

packages/devextreme-react/src/core/template-wrapper.tsx
VasilyStrelyaev: it has effect steering to the `else if`s below, and all other options to convey the same logic look less readable

PR #30637, opened 2025-08-06
Gallery: fix ts errors

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/ui/m_gallery.ts
+++ b/packages/devextreme/js/__internal/ui/m_gallery.ts
@@ -27,13 +27,15 @@ import { isDefined, isPlainObject } from '@js/core/utils/type';
 import { hasWindow } from '@js/core/utils/window';
 import type { DxEvent } from '@js/events';
 import type { ClickEvent } from '@js/ui/button';
-import CollectionWidget from '@js/ui/collection/ui.collection_widget.edit';
 import type { Item, Properties } from '@js/ui/gallery';
 import type { OptionChanged } from '@ts/core/widget/types';
 import type { SupportedKeys, WidgetProperties } from '@ts/core/widget/widget';
 import Widget from '@ts/core/widget/widget';
 import type { SwipeEndEvent, SwipeStartEvent, SwipeUpdateEvent } from '@ts/events/m_swipe';
 import type { CollectionWidgetEditProperties } from '@ts/ui/collection/collection_widget.edit';
+import CollectionWidget from '@ts/ui/collection/collection_widget.edit';
+
+import type { CollectionItemKey } from './collection/collection_widget.base';
 
 const GALLERY_CLASS = 'dx-gallery';
 const GALLERY_INDICATOR_VISIBLE_CLASS = 'dx-gallery-indicator-visible';
@@ -60,9 +62,8 @@ const GALLERY_ITEM_DATA_KEY = 'dxGalleryItemData';
 
 const MAX_CALC_ERROR = 1;
 
-export interface GalleryNavButtonProperties extends WidgetProperties {
+export interface GalleryNavButtonProperties extends WidgetProperties<GalleryNavButton> {
   direction?: string;
-
   onClick?: ((e: ClickEvent) => void) | null;
 }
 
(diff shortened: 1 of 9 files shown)


packages/devextreme/js/__internal/ui/m_gallery.ts
```diff
@@ -122,6 +125,8 @@ export interface GalleryProperties extends Properties {
   selectionRequired?: boolean;
 
   selectByClick?: boolean;
+
+  focusedElement?: Element | null;
 }
```
pharret31, line 123: Since Gallery component inherits the CollectionWidgetEdit you can use such construction for defining its props to avoid duplicating them between components

```js
export interface GalleryProperties extends Properties<Item>, Omit<
  CollectionWidgetEditProperties<Gallery, Item>,
  keyof Properties<Item>
  > {
...
}
```

packages/devextreme/js/__internal/ui/m_gallery.ts
```diff
@@ -188,7 +195,7 @@ class Gallery extends CollectionWidget<GalleryProperties> {
           focusStateEnabled: true,
         },
       },
-    ]);
+    ];
   }
 
   ctor(element: Element, options: Properties): void {
```
pharret31: Is it possible to get rid of ctor and change it to constructor?

packages/devextreme/js/__internal/ui/m_gallery.ts
```diff
@@ -453,34 +461,35 @@ class Gallery extends CollectionWidget<GalleryProperties> {
...
+    const offsetRatio = wrapAround ? 0.5 : 0;
     const freeSpace = this._itemFreeSpace();
     const isGapBetweenImages = !!freeSpace;
-    const rtlEnabled = this.option('rtlEnabled');
-    const selectedIndex = this.option('selectedIndex');
     const side = rtlEnabled ? 'Right' : 'Left';
     // @ts-expect-error ts-error
-    this._itemElements().each(function (index) {
+    // eslint-disable-next-line func-names
+    this._itemElements().each(function (index: number): void {
```
pharret31: Seems it could be an arrow function

packages/devextreme/js/__internal/ui/m_gallery.ts
```diff
@@ -558,72 +568,82 @@ class Gallery extends CollectionWidget<GalleryProperties> {
...
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    fx.animate(
+      this._$container.get(0),
+      {
+        type: 'slide',
+        to: {
+          left: targetPosition * (this._elementWidth() ?? 0),
+        },
+        duration: animationDuration,
+        complete() {
```
pharret31, line 573: could be an arrow function as well so we'll be able to get rid of that

packages/devextreme/js/__internal/ui/m_gallery.ts
```diff
@@ -687,24 +708,26 @@ class Gallery extends CollectionWidget<GalleryProperties> {
...
+  _renderSelectedItem(): void {
     const { selectedIndex } = this.option();
 
     this._itemElements()
       .removeClass(GALLERY_ITEM_SELECTED_CLASS)
-      // @ts-expect-error
+      // @ts-expect-error ts-error
       .eq(selectedIndex)
       .addClass(GALLERY_ITEM_SELECTED_CLASS);
   }
```
pharret31: Would that work?

```js
if (isDefined(selectedIndex)) {
this.itemElements().eq(selectedIndex).addClass(GALLERY_ITEM_SELECTED_CLASS);
}
```

packages/devextreme/js/__internal/ui/m_gallery.ts
anna-shakhova: unfortunately, we can't get rid of ctor in our components for now, as they are inherited from Widget, that is inherited from Class

packages/devextreme/js/__internal/ui/m_gallery.ts
anna-shakhova: used in because of `this` but here `this` is `item`, refactored
thanks for a hint here

packages/devextreme/js/__internal/ui/m_gallery.ts
anna-shakhova: `this` is undefined inside arrow function here, and we can't just bind some method to use inside complete, as we have `resolveWith(that)` here, so here we have to use alias

pharret31 (dismissed):

Zedwag (approved):

EugeniyKiyashko (approved):

EugeniyKiyashko (approved):

PR #30656, opened 2025-08-07
Splitter: prevent selection of pane content on double click on resize handler (T1302984)

pharret31:


packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
```diff
@@ -2523,6 +2523,33 @@ QUnit.module('Behavior', moduleConfig, () => {
...
     });
 
+    QUnit.test('resize handler should call the clear selection on double click to prevent content selection (T1302984)', function(assert) {
+        this.reinit({
+            items: [ { collapsible: true }, { collapsible: true } ],
+        });
+
+        const $resizeHandle = this.getResizeHandles().eq(0);
+
+        const originalGetSelection = window.getSelection;
```
EugeniyKiyashko: Use utility method to get Window object

import { getWindow } from '@js/core/utils/window';

const window = getWindow();

packages/devextreme/js/__internal/ui/splitter/resize_handle.ts
```diff
@@ -474,6 +475,8 @@ class ResizeHandle extends Widget<ResizeHandleOptions> {
   _doubleClickHandler(e: DxEvent<PointerEvent | MouseEvent | TouchEvent>): void {
     const { showCollapsePrev, showCollapseNext } = this.option();
 
+    clearSelection();
```
EugeniyKiyashko: I didn't see considered variant in the post, but may be this was missed
As to me, by calling clearSelection(), we’re merely fixing the symptom of the unwanted behavior after it’s already happened. If we decide that double-clicking the resizeHandle element should never select text in the first place, we can instead prevent it via CSS:
```
.dx-resize-handle {
    user-select: none;
}
```
To be attentive, with this rule, text selection will be disabled not only on double-click, but also whenever a pointerDown is performed and you drag the cursor over any text.

EugeniyKiyashko (approved):

PR #30657, opened 2025-08-07
Menu: set items expander stubs width to zero if adaptivityEnabled is set to true (T1302958)

pharret31:


anna-shakhova:
smth went wrong with alignment icon and text in material theme

packages/devextreme/js/__internal/ui/tree_view/tree_view.base.ts
```diff
@@ -95,6 +95,8 @@ export interface TreeViewBaseProperties extends Properties<TreeViewNode>, Omit<
   deferRendering?: boolean;
 
   _supportItemUrl?: boolean;
+
+  showItemStubs?: boolean;
```
EugeniyKiyashko: It’s possible to implement a solution without using a stub element at all, but since this is a bug and would require refactoring the current implementation, we’ll skip that option. 

I’d prefer the variant that simply hides the stub element, as is done in other components that customize TreeView appearance. That way, we only need to add one selector with display none for the TreeView in the menu.

EugeniyKiyashko (approved):

PR #30661, opened 2025-08-08
DataGrid - Refresh selectedItems on dataSource change when repaintChangesOnly=true (T1279797)

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/method-signature-style */
 import ArrayStore from '@js/common/data/array_store';
 import { CustomStore } from '@js/common/data/custom_store';
 import $ from '@js/core/renderer';
@@ -607,7 +606,6 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     this.pushed.fire(changes);
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   public fireError(...args: any[]) {
     this.dataErrorOccurred.fire(errors.Error.apply(errors, args));
   }
@@ -1640,7 +1638,7 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     if (options === true) {
       options = { reload: true, changesOnly: true };
     } else if (!options) {
-      options = { lookup: true, selection: true, reload: true };
+      options = { reload: true, lookup: true };
     }
 
     const that = this;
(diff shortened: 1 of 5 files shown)


packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -1640,7 +1638,7 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     if (options === true) {
       options = { reload: true, changesOnly: true };
     } else if (!options) {
-      options = { lookup: true, selection: true, reload: true };
+      options = { reload: true, lookup: true };
```
Alyar666, line 1641: I'm afraid in this case we can get extra requests. For example, when we have remote data and initially set `selectedRowKeys` with data keys that are not loaded yet, then when calling the `dataGridInstance.refresh(true)` method there will be an extra request for the selected data.

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
```diff
@@ -1855,7 +1855,6 @@ class EditingControllerImpl extends modules.ViewController {
     }
 
     when(dataController.refresh({
-      selection: isFullRefresh,
```
Alyar666: Same as the comment above. For example, when we have remote data, editing.refreshMode = 'reshape' and initially selectedRowKeys are set with data keys that are not yet loaded, then when editing and saving the data, an additional query will be executed for the selected data.

packages/devextreme/js/__internal/grids/grid_core/selection/m_selection.integration.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  const contentReadyHandler = (): void => {
+    resolve({ $container, instance });
+    instance.off('contentReady', contentReadyHandler);
+  };
+
+  instance.on('contentReady', contentReadyHandler);
+});
+
+describe('GridCore master_detail', () => {
```
Alyar666: What does master detail have to do with this?

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
Tucchhaa: Particularly this options.selection doesn't affect anything. It's never read, so it's safe to remove it

packages/devextreme/js/__internal/grids/grid_core/selection/m_selection.integration.test.ts
Tucchhaa: Thank you for noticing this, I will rename this string

packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
Tucchhaa: I have fixed it and added a test

Alyar666 (approved):

PR #30677, opened 2025-08-08
Scheduler: add agenda view model

Ambrozy:


aleksei-semikozov (approved):

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/create_scheduler.ts
```diff
@@ -3,16 +3,71 @@ import Scheduler from '@ts/scheduler/m_scheduler';
...
+        return container.querySelector('.dx-scheduler-appointment');
+      },
+      getAppointments(): NodeListOf<HTMLDivElement> {
+        return container.querySelectorAll('.dx-scheduler-appointment');
+      },
+      getAppointmentColor(view: string): string {
+        return view === 'agenda'
+          ? getAgendaAppointmentColor(container)
+          : getAppointmentColor(container);
+      },
```
wdevfx, line 51: [Super minor comment]

Maybe it will be more convenient to check the Scheduler's `currentView` option instead of passing the view through arguments
Sth like (just an idea):
```ts
      getAppointmentColor(): string {
        return scheduler.option('currentView') === 'agenda'
          ? getAgendaAppointmentColor(container)
          : getAppointmentColor(container);
      },
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/layoutManager.tests.js
```diff
@@ -42,11 +42,6 @@ const moduleOptions = {
...
 QUnit.module('LayoutManager', moduleOptions);
 
-QUnit.test('LayoutManager should be initialized', async function(assert) {
-    await this.createInstance();
-    assert.ok(this.instance.getLayoutManager() instanceof AppointmentLayoutManager, 'AppointmentLayoutManager was initialized');
-});
-
 QUnit.test('RenderingStrategy should be initialized', async function(assert) {
     await this.createInstance();
     assert.ok(this.instance.getLayoutManager().getRenderingStrategyInstance() instanceof BaseAppointmentsStrategy, 'AppointmentLayoutManager was initialized');
```
wdevfx, line 47: Found difference in a behavior with previous version

Config:
```js
dataSource: [
                        {
                            startDate: '2024-01-01T02:00:00',
                            endDate: '2024-01-01T22:00:00',
                            text: 'Appt #0',
                        },
                    ],
                    currentDate: '2024-01-01',
                    currentView: 'agenda',
                    startDayHour: 10,
                    endDayHour: 19,
```

Previous version result:
Appointment `startDate` / `endDate` displayed
A4F2E220-4589-496F-A5F2-E98D40099FE3

Actual version result:
Appointment "cut" by `startDayHour` / `endDayHour`
2A258D8B-5401-42B8-B08C-54BB585B0360_4_5005_c

Additional info:
I don't think that current behavior is "incorrect", it's just different
Formally, it may be a very small `BC`, so, let's discuss it additionally and pay attention to this nuance :)

wdevfx, line 47: Found a bug (also can be reproduced in the previous version)

Config
```js
dataSource: [
                        {
                            startDate: '2024-01-01T02:00:00',
                            endDate: '2024-01-01T22:00:00',
                            text: 'Appt #0',
                            recurrenceRule: 'FREQ=DAILY;COUNT=5',
                            recurrenceException: '20240101T220000Z',
                        },
                    ],
                    currentDate: '2024-01-01',
                    currentView: 'agenda',
                    startDayHour: 10,
                    endDayHour: 19,
```

Steps to reproduce:
1. Open example with Scheduler configuration from the config above
2. Try to delete appointment from recurrence in the `agenda` view (delete appointment, not series)

Expected result
Appointment deleted

Actual result
Appointment still displayed

Additional information
This issue also reproduced in previous version too
Also, the correct `recurrenceException` works well (appointment on TH don't displayed)

packages/devextreme/js/__internal/scheduler/r1/utils/agenda.ts
```diff
@@ -1,7 +1,43 @@
...
 
 export const calculateStartViewDate = (currentDate: Date, startDayHour: number): Date => {
   const validCurrentDate = new Date(currentDate);
 
   return setOptionHour(validCurrentDate, startDayHour);
 };
+
+const getDayStart = (date: Date | number, viewOffset: number): number => {
+  const trimDate = new Date(date).setHours(0, 0, 0, 0);
+  return trimDate + viewOffset;
```
wdevfx, line 12: [Question]

Regarding to [the docs](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/Configuration/#offset) the `agenda` view should not take into account the `offset` option
Why do we need the `viewOffset` calculations here? :)

packages/devextreme/js/__internal/scheduler/view_model/filtering/utils/get_appointment_filter/get_appointment_filter.ts
```diff
@@ -1,11 +1,11 @@
 import type { TimeZoneCalculator } from '../../../../r1/timezone_calculator/calculator';
 import { isAppointmentTakesAllDay } from '../../../../r1/utils/index';
 import type { AppointmentDataItem } from '../../../../types';
+import { isAppointmentMatchedResources } from '../../../../view_model_new/filtration/utils/filter_by_attributes/is_appointment_matched_resources';
```
wdevfx, line 4: [Super minor / suggestion]

Maybe instead of a lot of dots (like `../../../`) it will be better to use `ts alias` like `@ts/scheduler/...`

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/get_filter_options/get_filter_options.test.ts
```diff
@@ -0,0 +1,77 @@
...
+describe('BaseAppointmentLayoutManager', () => {
+  it('should return correct filter options for agenda view', () => {
+    const schedulerStore = getSchedulerStore('agenda');
+    expect(getFilterOptions(schedulerStore)).toEqual({
+      allDayPanelMode: 'allDay',
+      showAllDayPanel: true,
+      supportAllDayPanel: false,
+      resourceManager: schedulerStore.resourceManager,
+      timeZoneCalculator: schedulerStore.timeZoneCalculator,
+      viewOffset: 30 * 60_000,
```
wdevfx, line 22: [Qeustion]

As far as I understand (and after tests), the `offset` option should not take affect in `agenda` view
Why do we have it here?

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/get_filter_options/get_visible_date_time_intervals.ts
```diff
@@ -1,20 +1,18 @@
...
 
 export const getVisibleDateTimeIntervals = ({
   startDayHour,
   endDayHour,
   min,
   max,
-}: CompareOptions, isDateViewOnly: boolean): DateInterval[] => {
-  if (isDateViewOnly || (startDayHour === 0 && endDayHour === 24)) {
+}: CompareOptions, isDateViewOnly: boolean, isSplitByDays = false): DateInterval[] => {
+  if (isDateViewOnly || (!isSplitByDays && (startDayHour === 0 && endDayHour === 24))) {
```
wdevfx, line 12: [Minor comment]

Let's separate right side condition into one separate variable:
```ts
// naming is bad, feel free to choose better one
const shouldBehaveLikeDateOnly = !isSplitByDays && startDayHour === 0 && endDayHour === 24;
```

Btw, for the right side we can drop the additional brackets:
```ts
if (isDateViewOnly || (!isSplitByDays && startDayHour === 0 && endDayHour === 24)) {
```

packages/devextreme/js/__internal/scheduler/view_model/filtering/utils/get_appointment_filter/get_appointment_occurrence_dates.ts
```diff
@@ -67,11 +71,9 @@ export const getAppointmentOccurrenceDates = (
     // -> ['2024-02-01T00:00:00', '2024-02-02T23:59:59']
     case viewOffset === 0:
       return {
-        startDate: dateUtils.trimTime(originalStartDate),
-        endDate: dateUtilsTs.addOffsets(
-          dateUtils.trimTime(originalEndDate),
-          [DAY_WITHOUT_ONE_SECOND_MS],
-        ),
+        startDate: dateUtils.trimTime(new Date(originalStartDate)).getTime(),
+        endDate: dateUtils.trimTime(new Date(originalEndDate)).getTime()
```
wdevfx, line 75: [Suggestion for further improvement]

Let's incapsulate `new Date` creation in the `dateUtils.trimTime`
It's already creates a new date to prevent mutation, but now it looks like `new Date(date.getTime())`

I guess we can make argument type wider in the `dateUtils.trimTime` and accept a milliseconds too :)

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/split_by_parts.ts
```diff
@@ -0,0 +1,82 @@
...
+    .reduce<(T & AppointmentPart)[]>((result, entity) => {
+      const panelOptions = entity.isAllDayPanelOccupied
+        ? allDayPanel
+        : regularPanel;
+      const { intervals, prevIntervalEndDate, nextIntervalStartDate } = panelOptions;
+      const startIndex = intervals.findIndex(({ max }) => entity.startDate < max);
+      let endIndex = intervals.findIndex(
+        (_, index) => index < intervals.length - 1 && entity.endDate <= intervals[index + 1].min,
+      );
+      endIndex = endIndex === -1 ? intervals.length - 1 : endIndex;
```
wdevfx, line 56: As far as I understand, the condition `index < intervals.length - 1` will be true for all array items except the last one

So, we're looking for interval item (except last one) where the next item min date greater or equal the `entity.endDate`
And if we found nothing -> we will take last interval index as `endIndex`

Let's move this logic maybe to a separate function, to make it a little bit easier to read & understand:
```ts
const getEndIndex = (intervals: DateInterval[], endDateMs: number): number => {
  const lastIdx = intervals.length - 1;

  for (let idx = 0; idx < lastIdx; idx++) {
    const nextInterval = intervals[idx + 1];
    
    if (nextInterval.min >= endDate) {
      return idx;
    }
  }
  
  return lastIdx;
} 
```

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/split_by_parts.ts
```diff
@@ -0,0 +1,82 @@
...
+        const parts: (T & AppointmentPart)[] = Array.from({ length: partCount })
+          .map((_, partIndex) => ({
+            ...cropEntityByInterval(entity, intervals[startIndex + partIndex]),
+            partIndex,
+            partCount,
+            reduced: 'body',
+          }));
+        parts[0].reduced = isStartOnPrevInterval ? 'body' : 'head';
+        parts[parts.length - 1].reduced = isEndOnNextInterval ? 'body' : 'tail';
+        result.push(...parts);
```
wdevfx, line 78: [Minor comment]

Let's simplify it a little bit and get rid of the `parts` array mutations
For achieving this, let's add one more util func `getReduced` or sth like this

```ts
const getReduced = (
 isFirstItem: boolean,
 isLastItem: boolean,
 isStartOnPrevInterval: boolean,
 isEndOnNextInterval: boolean,
): 'head' | 'body' | 'tail' => {
  switch (true) {
    case isFirstItem && !isStartOnPrevInterval:
      return 'head';
    case isLastItem && !isEndOnNextInterval:
      return 'tail';
    default:
      return 'body';
  }
}

// in this code part
const parts = Array.from({ length: partCount })
  .map((_, partIndex) => {
    const isFirstIdx = partIndex === 0;
    const isLastIdx = partIndex === partCount - 1;
    
    
    return {
      ...cropEntityByInterval(entity, intervals[startIndex + partIndex]),
      partIndex,
      partCount,
      reduced: getReduced(isFirstIdx, isLastIdx, isStartOnPrevInterval, isEndOnNextInterval),
    }
  });
```

wdevfx, line 78: [Minor comment]

Just jfyi, instead of `push` + `spread` we can use a `concat` here :)

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/split_by_recurrence/split_by_recurrence.ts
```diff
@@ -0,0 +1,28 @@
...
+    };
+    const occurrences = getAppointmentRecurrenceOccurrences<T>(
+      appointment,
+      { firstDayOfWeek, interval: recurrenceInterval },
+      timeZoneCalculator,
+    );
+
+    occurrences.forEach((occurrence) => {
+      acc.push(occurrence);
+    });
```
wdevfx, line 25: [Minor comment]

Maybe `concat` here will be better?

packages/devextreme/js/__internal/scheduler/view_model_new/m_appointments_layout_manager_bridge.ts
```diff
@@ -0,0 +1,93 @@
...
+class AppointmentLayoutManagerBridge {
+  preparedItems: any[] = [];
+
+  filteredItems: any[] = [];
+
+  oldLayoutManager: AppointmentLayoutManagerDeprecated;
+
+  get _positionMap() {
+    return this.oldLayoutManager._positionMap;
+  }
```
wdevfx, line 21: As far as I can see, wee need this `_positionMap` getter only for `QUnit` tests
Please correct me if I wrong

We have a build feature, that can `cut off` some code for production build
[Link to the usage example](https://github.com/DevExpress/DevExtreme/blob/c4266301e4719b6577a08d29092a1c2fac8501e1/packages/devextreme/js/__internal/utils/version.ts#L41)
The code block between comments is deleted from the production bundles / packages

So, maybe we should use it here to highlight that we need this getter only in tests? :)

packages/devextreme/js/__internal/scheduler/view_model_new/preparation/utils/replace_incorrect_end_date.ts
```diff
@@ -0,0 +1,44 @@
...
+): SafeAppointment[] => {
+  if (!rawAppointments) {
+    return [];
+  }
+
+  return rawAppointments.reduce<SafeAppointment[]>((result, rawAppointment) => {
+    const startDate = dataAccessors.get('startDate', rawAppointment);
+    const endDate = dataAccessors.get('endDate', rawAppointment);
+
+    // NOTE: error E1032
```
wdevfx, line 23: Let's write more rich comment
Now it isn't fully clear what does it mean, e.g:
- Is this error should be thrown in other code during dates validation?
- Or we should add error here later?

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/create_scheduler.ts
Ambrozy: I don't think so. It's not clear logic. If something will go wrong you have to know that this function works like this, but if you pass parameter you are sure how it works.

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/layoutManager.tests.js
Ambrozy: You are right. It's BC. You can see that I replaced screenshots because of this. Let's discuss

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/get_filter_options/get_filter_options.test.ts
Ambrozy: I have already removed this file in next task. Don't mind

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/split_by_parts.ts
Ambrozy: Absolutely no! concat creates each time new array, but push just add items to existed. concat much much longer then push.

packages/devextreme/js/__internal/scheduler/view_model_new/m_appointments_layout_manager_bridge.ts
Ambrozy: I will delete oldLayoutManager at the end, so no need to think about now

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/split_by_parts.ts
Ambrozy: Sure

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/split_by_parts.ts
Ambrozy: Sure. Will add in next PR

packages/devextreme/js/__internal/scheduler/view_model_new/filtration/utils/get_filter_options/get_visible_date_time_intervals.ts
Ambrozy: It's bad API here, but I don't know how to fix it to make it better

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/layoutManager.tests.js
Ambrozy: This BC will cover it: https://github.com/DevExpress/devextreme-private/issues/1625

Ambrozy: We have task 1625 to resolve it

PR #30678, opened 2025-08-08
T1280020 - DataGrid - The 'row' parameter in the FocusedRowChanged event refers to a non-focused row if the grid height is small

Raushen:


packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
```diff
@@ -166,12 +167,14 @@ export class FocusController extends core.ViewController {
     }
   }
 
-  private _focusRowByKey(key) {
+  private _focusRowByKey(
+    key: any,
+  ): DeferredObj<number | unknown> {
     if (!isDefined(key)) {
       this._resetFocusedRow();
```
pomahtri, line 175: Question
Is `_resetFocusedRow()` sync? If not and there are some async ops inside, won't it be misleading that we don't wait for it?

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
```diff
@@ -205,10 +208,13 @@ export class FocusController extends core.ViewController {
     if (!this.isAutoNavigateToFocusedRow()) {
       this.option('focusedRowIndex', -1);
     }
-    return this._navigateToRow(key);
+    return this._navigateToRow(key, false);
   }
 
-  public _navigateToRow(key, needFocusRow?) {
+  public _navigateToRow(
+    key: any,
```
pomahtri: Minor
Afair, there was some `export type Key = any`
It will not change anything for TS, but let's still use it for better readability

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
```diff
@@ -205,10 +208,13 @@ export class FocusController extends core.ViewController {
     if (!this.isAutoNavigateToFocusedRow()) {
       this.option('focusedRowIndex', -1);
     }
-    return this._navigateToRow(key);
+    return this._navigateToRow(key, false);
```
pomahtri, line 212: Question
Why? Isn't it required to focus row when `autoNavigateToFocusedRow` is true?

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
```diff
@@ -166,12 +167,14 @@ export class FocusController extends core.ViewController {
     }
   }
 
-  private _focusRowByKey(key) {
+  private _focusRowByKey(
+    key: any,
```
pomahtri: Minor
Let's use `Key` type (see another comment below)

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
Raushen: It is just a refactoring, I've added this argument to make method typing easier.
I think a row can be focused only if the row focusing is enabled. Also this method is related to the scrolling and can be used without focusing.

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
Raushen: The method is sync

e2e/testcafe-devextreme/tests/dataGrid/common/scrolling.ts
```diff
@@ -1824,3 +1824,29 @@ test('DataGrid - Scrolling position is reset to far right on an attempt to scrol
...
+    { id: '0' },
+    { id: '1' },
+    { id: '2' },
+  ],
+  scrolling: { mode: 'virtual' },
+  keyExpr: 'id',
+  focusedRowEnabled: true,
+  onFocusedRowChanged(e) {
+    const data = e.row?.data;
+    $('#otherContainer').text(data.id);
```
Alyar666, line 1850: Why do we need a container to output data? Can't we remember the row object in the window? And check it as a whole to make sure it's valid?

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
```diff
@@ -205,10 +209,13 @@ export class FocusController extends core.ViewController {
...
       this.option('focusedRowIndex', -1);
     }
-    return this._navigateToRow(key);
+    return this._navigateToRow(key, false);
   }
 
-  public _navigateToRow(key, needFocusRow?) {
+  public _navigateToRow(
+    key: Key,
+    needFocusRow: boolean,
```
Alyar666, line 217: I suggest setting this argument to a default value of false right away

Raushen: Why? It is a required argument now and its value defined at calling the method. [Here is passing false](https://github.com/DevExpress/DevExtreme/pull/30678/files#diff-59b9c2648b45ecdc8ea26a98ba3c536f9e833b4b15634c71a094025ff190a4dfR212).

e2e/testcafe-devextreme/tests/dataGrid/common/scrolling.ts
Raushen: It is more easier way for me. Surely you can use the window object but it will require more code,

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
Alyar666: Yeah, I'd like to suggest leaving this argument optional with a default value of false. That way you don't have to explicitly pass false when calling this method.
But I won't insist on this.

e2e/testcafe-devextreme/tests/dataGrid/common/scrolling.ts
Alyar666: Well, okay, let's leave it as is.

Alyar666 (dismissed):

Alyar666 (approved):

PR #30684, opened 2025-08-08
Scheduler:  fix appointment of the last day of month occupy two cells in timeline view instead of one

aleksei-semikozov:


packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/appointment.dataProcessor.tests.js
```diff
@@ -1057,6 +1057,32 @@ module('Client side after filtering', () => {
         });
     });
 
+    test('Appointment should be filtered if endDate is at the edge of the trimmed end view date', function(assert) {
```
wdevfx, line 1060: I guess it doesn't matter in the `24.2` version
But in the future, let's try to write `jest` tests instead of `QUnit` ones

Because in the future we'd like to get rid of the old `QUnit` tool and it's better not to increase `QUnit` tests count :)

wdevfx (approved):

PR #30686, opened 2025-08-08
Scheduler: refactor entry point of view model

Ambrozy:


packages/devextreme/js/__internal/scheduler/appointments/data_provider/m_appointment_data_provider.ts
```diff
@@ -1,154 +0,0 @@
-import config from '@js/core/config';
```
wdevfx, line 1: We lost `git history` for this file
Now for git it looks like that we delete the old one and add a new one
This history is important for us and we deep dive into it sometimes, because lack of documentation
So, it's extremely useful to understand when and why this code was added to the component :)

65E11491-09B9-444E-ADC4-8208706F06E7_4_5005_c

### How to move file and 100% keep it version history

1. Use `git mv` for moving files (sometimes IDE do wrong actions unfortunately)
2. Move files in a separate commit without any other changes (if u'd like to modify files -> just do it in another commit)
3. Merge PR without squash (squash will lost files history)

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -31,16 +31,15 @@ import { custom as customDialog } from '@js/ui/dialog';
 import type { Appointment, AppointmentTooltipShowingEvent } from '@js/ui/scheduler';
 import errors from '@js/ui/widget/ui.errors';
 import { dateUtilsTs } from '@ts/core/utils/date';
+import { AppointmentDataProvider } from '@ts/scheduler/view_model/3.generate_view_model/data_provider/m_appointment_data_provider';
```
wdevfx: Let's avoid numbers in imports

The `generate_view_model` is enough
Here only 4 steps, so I think it's not required to mark these steps order in folder names

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -154,12 +149,6 @@ const RECURRENCE_EDITING_MODE = {
 };
 
 class Scheduler extends SchedulerOptionsBaseWidget {
-  // NOTE: Do not initialize variables here, because `_initMarkup` function runs before constructor,
```
wdevfx, line 157: Question.

Is this `NOTE` not actual anymore?

packages/devextreme/js/__internal/scheduler/view_model/4.detect_changes/get_repainted_appointments.ts
```diff
@@ -0,0 +1,153 @@
...
+
+const isSettingChanged = (
+  settings: Settings[],
+  sourceSetting: Settings[],
+): boolean => {
+  if (settings.length !== sourceSetting.length) {
+    return true;
+  }
+
+  const createSettingsToCompare = (
```
wdevfx: As far as I can see, it's pure function, let's move it to the module scope outside the `isSettingChanged`
(Like `isAppointmentShouldAppear` function)

packages/devextreme/js/__internal/scheduler/view_model/dataProcessor.test.ts
```diff
@@ -0,0 +1,1220 @@
+import {
```
wdevfx: It's QUnit -> Jest migration, but I want to highlight that this file history is important, let's not lost it

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
Ambrozy: I've left the NOTE just in case

wdevfx (approved):

PR #30599, opened 2025-08-01
Calendar and Navigator: improve types

Zedwag:

PR #30688, opened 2025-08-11
Calendar: improve Views and Selection strategies types

Zedwag:


packages/devextreme/js/__internal/ui/date_range_box/m_date_range_box.ts
```diff
@@ -149,7 +149,7 @@ class DateRangeBox extends Editor<Properties> {
       value: [null, null],
       valueChangeEvent: 'change',
       _internalValidationErrors: [],
-      _currentSelection: 'startDate',
```
ksercs, line 152: why if we don't publish it?

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
```diff
@@ -1725,10 +1750,12 @@ class Calendar<
...
         if (!isSameValue) {
-          this._selectionStrategy.processValueChanged(value, previousValue);
+          this._selectionStrategy.processValueChanged(
+            value as (Date | null)[],
+            previousValue as (Date | null)[],
+          );
         }
-
-        this._setSubmitValue(value);
+        this._setSubmitValue(value as DateLike | DateLike[] | undefined);
```
ksercs, line 1763: DateLike should already have `undefined`, no?

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
```diff
@@ -179,8 +192,7 @@ class BaseView<
     const { value } = this.option();
 
     const localizedWidgetName = this._getLocalizedWidgetName();
-    // @ts-expect-error ts-error
-    const [startDate, endDate] = value;
+    const [startDate, endDate] = value as [Date | undefined, Date | undefined];
```
ksercs: `dateLike[]`?
Maybe we can even extract a new type like `DateRange`? Because I see you use this custom type in several places.
I remember I've tried to do it in .d.ts some time ago but didin't manage to do this because IT does not support ts tuples. But maybe we can use it for the code at least

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
```diff
@@ -1318,10 +1328,16 @@ class Calendar<
       .appendTo(this.$element());
 
     const { value } = this.option();
+    // if (!this._isArrayValue('value', value)) {
```
ksercs: ?

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
Zedwag: no, null only

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
Zedwag: just some artifacts

packages/devextreme/js/__internal/ui/date_range_box/m_date_range_box.ts
Zedwag: I thought at some point we alredy strarted to name private options without _ prefix as well. So since such a naming is not ok for current lint rules, I decided to rename it instead of changing lint rules

marker-dao: I would prefer it to remain as it was (mb with eslint comment to fix it in the future)

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
marker-dao: just a joke)

packages/devextreme/js/__internal/core/utils/m_date.ts
```diff
@@ -425,33 +425,33 @@ function getDateIntervalByString(intervalString) {
...
 
-function sameHoursAndMinutes(date1, date2) {
+function sameHoursAndMinutes(date1, date2): boolean {
   return date1 && date2 && date1.getHours() === date2.getHours() && date1.getMinutes() === date2.getMinutes();
 }
 
-const sameDecade = function (date1, date2) {
-  if (!isDefined(date1) || !isDefined(date2)) return;
+const sameDecade = function (date1, date2): boolean {
+  if (!isDefined(date1) || !isDefined(date2)) return false;
```
marker-dao:
```suggestion
  if (!isDefined(date1) || !isDefined(date2)) {
      return false;
  }
```

packages/devextreme/js/__internal/core/utils/m_date.ts
```diff
@@ -425,33 +425,33 @@ function getDateIntervalByString(intervalString) {
...
   const startDecadeDate1 = date1.getFullYear() - date1.getFullYear() % 10;
   const startDecadeDate2 = date2.getFullYear() - date2.getFullYear() % 10;
 
   return date1 && date2 && startDecadeDate1 === startDecadeDate2;
 };
 
-const sameCentury = function (date1, date2) {
-  if (!isDefined(date1) || !isDefined(date2)) return;
+const sameCentury = function (date1, date2): boolean {
+  if (!isDefined(date1) || !isDefined(date2)) return false;
```
marker-dao: same

packages/devextreme/js/__internal/core/utils/m_date.ts
```diff
@@ -545,23 +545,23 @@ function getDayWeekNumber(date, firstDayOfWeek) {
...
+function getWeekNumber(date, firstDayOfWeek, rule): number {
   const firstWeekDayInYear = getDayWeekNumber(getFirstDateInYear(date.getFullYear()), firstDayOfWeek);
   const lastWeekDayInYear = getDayWeekNumber(getLastDateInYear(date.getFullYear()), firstDayOfWeek);
   const daysInFirstWeek = DAYS_IN_WEEK - firstWeekDayInYear + 1;
 
   let weekNumber = Math.ceil((getDayNumber(date) - daysInFirstWeek) / 7);
   switch (rule) {
     case 'fullWeek': {
-      if (daysInFirstWeek === DAYS_IN_WEEK) { weekNumber++; }
+      if (daysInFirstWeek === DAYS_IN_WEEK) { weekNumber += 1; }
```
marker-dao:
```suggestion
      if (daysInFirstWeek === DAYS_IN_WEEK) {
          weekNumber += 1;
      }
```

packages/devextreme/js/__internal/core/utils/m_date.ts
```diff
@@ -545,23 +545,23 @@ function getDayWeekNumber(date, firstDayOfWeek) {
...
       if (weekNumber === 0) {
         const lastDateInPreviousYear = getLastDateInYear(date.getFullYear() - 1);
         return getWeekNumber(lastDateInPreviousYear, firstDayOfWeek, rule);
       }
       return weekNumber;
     }
     case 'firstDay': {
-      if (daysInFirstWeek > 0) { weekNumber++; }
+      if (daysInFirstWeek > 0) { weekNumber += 1; }
```
marker-dao: same

packages/devextreme/js/__internal/core/utils/m_date.ts
```diff
@@ -570,7 +570,7 @@ function getWeekNumber(date, firstDayOfWeek, rule) {
       return weekNumber;
     }
     case 'firstFourDays': {
-      if (daysInFirstWeek > 3) { weekNumber++; }
+      if (daysInFirstWeek > 3) { weekNumber += 1; }
```
marker-dao: same

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
```diff
@@ -31,23 +51,25 @@ class CalendarMultiSelectionStrategy extends CalendarSelectionStrategy {
...
     return this._getLowestDateInArray(dates);
   }
 
-  restoreValue() {
+  restoreValue(): void {
     this.calendar.option('value', []);
   }
 
-  _weekNumberClickHandler({ rowDates, event }) {
+  _weekNumberClickHandler({ rowDates, event }: { rowDates: Date[]; event: DxEvent }): void {
```
marker-dao: It's better to describe an interface in the top of the file

packages/devextreme/js/__internal/ui/calendar/m_calendar.range.selection.strategy.ts
```diff
@@ -1,16 +1,33 @@
...
     super(component);
     this.NAME = 'RangeSelection';
   }
 
-  getViewOptions() {
+  dateOption(optionName: 'value'): (Date | null)[];
+  dateOption(optionName: 'min' | 'max'): Date | null;
+  dateOption(optionName: 'min' | 'max' | 'value'): Date | null | (Date | null)[] {
+    if (optionName === 'value') {
+      return this.calendar._getDateOption('value') as (Date | null)[];
```
marker-dao: It's better to add a type guard here instead of `as`. Or may be fix typing in _getDateOption

packages/devextreme/js/__internal/ui/calendar/m_calendar.range.selection.strategy.ts
```diff
@@ -19,77 +36,84 @@ class CalendarRangeSelectionStrategy extends CalendarSelectionStrategy {
...
-    previousValue ??= [];
-
+  updateAriaSelected(val?: (Date | null)[] | null, previousVal?: (Date | null)[] | null): void {
+    const value = val ?? this._getValue();
+    const previousValue = previousVal ?? [];
     super.updateAriaSelected(value, previousValue);
   }
 
-  processValueChanged(value, previousValue) {
+  processValueChanged(value: (Date | null)[], previousValue: (Date | null)[]): void {
```
marker-dao, line 87: You use this type `(Date | null)[]` non in one place, mb describe it in the top?)
And I see we can get the typing from m_calendar: this is the value option (for both)

packages/devextreme/js/__internal/ui/calendar/m_calendar.range.selection.strategy.ts
```diff
@@ -105,53 +129,60 @@ class CalendarRangeSelectionStrategy extends CalendarSelectionStrategy {
...
-    // @ts-expect-error
-    const rangeStartDate = new Date(Math.max(firstDateInViews, startDate));
-    // @ts-expect-error
-    const rangeEndDate = new Date(Math.min(lastDateInViews, endDate));
-
-    return [...dateUtils.getDatesOfInterval(rangeStartDate, rangeEndDate, DAY_INTERVAL), rangeEndDate];
+    const firstDateInViews = dateUtils.getFirstMonthDate(
+      currentDate,
+      isAdditionalViewDate ? -2 : -1,
+    ) as Date;
```
marker-dao, line 148: It's better to use // ts-expect-error comment here and describe getFirstMonthDate correctly in the feature)

packages/devextreme/js/__internal/ui/calendar/m_calendar.range.selection.strategy.ts
```diff
@@ -105,53 +129,60 @@ class CalendarRangeSelectionStrategy extends CalendarSelectionStrategy {
...
+    const lastDateInViews = dateUtils.getLastMonthDate(
+      currentDate,
+      isAdditionalViewDate ? 1 : viewsCount,
+    ) as Date;
```
marker-dao, line 152: same

packages/devextreme/js/__internal/ui/calendar/m_calendar.range.selection.strategy.ts
```diff
@@ -105,53 +129,60 @@ class CalendarRangeSelectionStrategy extends CalendarSelectionStrategy {
...
   }
 
-  _getRange() {
+  _getRange(): Date[] {
     const [startDate, endDate] = this._getValue();
     return this._getDaysInRange(startDate, endDate);
   }
 
-  _getDaysInRange(startDate, endDate) {
+  _getDaysInRange(startDate: Date | null, endDate: Date | null): Date[] {
```
marker-dao, line 138: Can we get return type from calendar?

packages/devextreme/js/__internal/ui/calendar/m_calendar.range.selection.strategy.ts
```diff
@@ -105,53 +129,60 @@ class CalendarRangeSelectionStrategy extends CalendarSelectionStrategy {
...
+    const rangeEndDate = new Date(Math.min(lastDateInViews.getTime(), endDate.getTime()));
+
+    return [
+      ...dateUtils.getDatesOfInterval(rangeStartDate, rangeEndDate, DAY_INTERVAL),
+      rangeEndDate,
+    ];
   }
 
-  _cellHoverHandler(e) {
+  _cellHoverHandler(e: { value: Date }): void {
```
marker-dao: I believe we can describe this event in m_calendar.base_view.ts.
Check the files, please, there may be similar places where you can describe types in the same way.

packages/devextreme/js/__internal/ui/calendar/m_calendar.single.selection.strategy.ts
```diff
@@ -1,46 +1,56 @@
...
-  constructor(component) {
+  constructor(component: Calendar) {
     super(component);
     this.NAME = 'SingleSelection';
   }
 
-  getViewOptions() {
+  dateOption(optionName: 'min' | 'max' | 'value'): Date | null {
+    if (optionName === 'value') {
+      return this.calendar._getDateOption('value') as Date;
```
marker-dao: Are you sure that the method will not return null at this place? Or another type

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
```diff
@@ -83,6 +90,14 @@ export interface BaseViewProperties extends Properties {
   range: Date[];
 
   hoveredRange: Date[];
+
+  cellTemplate?: template | (
+    (itemData: CellTemplateData, itemIndex: number, itemElement: Element) => dxElementWrapper
+  );
+
+  onCellClick?: (e: { event: DxEvent; value: Date }) => void;
```
marker-dao: copypast

packages/devextreme/js/__internal/ui/calendar/m_calendar.multiple.selection.strategy.ts
marker-dao: Or in the base file

packages/devextreme/js/__internal/ui/date_range_box/m_date_range_box.ts
ksercs: > we alredy strarted to name private options without _ prefix as well

Didn't know about this 🤔
Maybe you're about private class props not options? It sounds more reasonable because modificators can be used. But for options I would prefer to differ private and public ones, don't u?

Zedwag: no, i mean options. that is why in my previous pr I used this [hack](https://github.com/DevExpress/DevExtreme/pull/30599/files#diff-44adb655963a7077017a2c45bf1e09d8d010b3addd794d15f9aa3cb54bc1ad8aR704), otherwise there is eslint warning on const started with prefix

I see u both prefer keep private options with prefix, I'll revert these name and add lint ignore comment. Later we may consider disabling this rule

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
```diff
@@ -114,7 +129,6 @@ class BaseView<
 
   _$selectedCells!: dxElementWrapper;
```
EugeniyKiyashko:
```suggestion
  _$selectedCells!: dxElementWrapper[];
```

EugeniyKiyashko: It fixes three ts-errors

EugeniyKiyashko: _$rangeCells, _$hoveredRangeCells has the same type. It fixes 7 ts-errors

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
```diff
@@ -668,23 +692,24 @@ class BaseView<
...
+    ) ?? date;
+
+    return new Date(min && firstAvailableDate < min ? min : firstAvailableDate);
   }
 
   // @ts-expect-error ts-error
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  _getCellByDate(contouredDate): dxElementWrapper {
+  _getCellByDate(contouredDate: Date | undefined): dxElementWrapper {
     Class.abstract();
```
EugeniyKiyashko: If you don't use Class.abstract in other methods, may be it better return empty dxElementWrapper as follows: return $();

packages/devextreme/js/__internal/ui/calendar/m_calendar.base_view.ts
Zedwag: it's more like local hack here, tbh i think we can even remove undefined, updated the pr. I don't think we should extract a separate type, since it's only actual in couple place
>dateLike[]?

no, in views we don't use dateLike type, since value is already converted from string or number when it passed to Views

packages/devextreme/js/__internal/ui/calendar/m_calendar.range.selection.strategy.ts
Zedwag: nah, don't think we should, it's more like local for range selection

packages/devextreme/js/__internal/ui/calendar/m_calendar.range.selection.strategy.ts
Zedwag: type is ok for _getDateOption since it consider 3 different strategies. but I don't think type guard will make each of strategies better than as

packages/devextreme/js/__internal/ui/calendar/m_calendar.ts
Zedwag: just to clarify: export type DateLike = Date | number | string | null;

packages/devextreme/js/__internal/ui/calendar/m_calendar.single.selection.strategy.ts
Zedwag: good point, added null in all strategies. and no other types are expected

packages/devextreme/js/__internal/ui/calendar/m_calendar.range.selection.strategy.ts
Zedwag: this function is already typed and seems like have correct types for common case, but should be ok when we sure date is defined (and we're sure for current date)

EugeniyKiyashko (approved):

r-farkhutdinov (approved):

marker-dao (approved):

PR #30698, opened 2025-08-11
Toolbar: close menu when a focused item is selected via the enter/space key (T1301705)

juleari:


packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
```diff
@@ -282,8 +282,11 @@ export default class DropDownMenu extends Widget<DropDownMenuProperties> {
       dragEnabled: false,
       showTitle: false,
       fullScreen: false,
+      ignoreChildEvents: false,
       _fixWrapperPosition: true,
     });
+    this._popup.registerKeyHandler('space', this._popupKeyHandler.bind(this));
```
EugeniyKiyashko:
```suggestion
    this._popup.registerKeyHandler('space', (e) => { 
         this._popupKeyHandler(e);
    }
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
```diff
@@ -595,6 +595,74 @@ QUnit.module('widget sizing render', moduleConfig, () => {
...
+            const popup = this.overflowMenu.popup();
+            assert.equal(popup.option('visible'), true, 'popup is visible');
+
+            this.keyboard
+                .keyDown('tab')
+                .keyDown('space');
+            assert.equal(popup.option('visible'), true, 'space on dxCheckbox does not hide popup');
+
+            this.instance.option('closeOnClick', false);
+            this.keyboard
```
EugeniyKiyashko: target element is overflow menu button, not button in list. So we need concrete on what element we trigger space or enter

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
```diff
@@ -322,13 +325,7 @@ export default class DropDownMenu extends Widget<DropDownMenuProperties> {
...
       noDataText: '',
       itemTemplate,
-      onItemClick: (e: ItemClickEvent<ListItem>): void => {
-        const { closeOnClick } = this.option();
-        if (closeOnClick) {
-          this.option('opened', false);
-        }
-        this._itemClickAction?.(e);
-      },
+      onItemClick: (e) => this._itemClickHandler(e),
```
EugeniyKiyashko:
```suggestion
      onItemClick: (e) => {
          this._itemClickHandler(e);
      }
```

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
```diff
@@ -359,7 +376,7 @@ export default class DropDownMenu extends Widget<DropDownMenuProperties> {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   _getKeyboardListeners(): any[] {
-    return super._getKeyboardListeners().concat([this._list]);
+    return super._getKeyboardListeners().concat([this._list, this._popup]);
```
EugeniyKiyashko: Seems this change is excessive; I assume that with ignoreChildEvents, all handlers of the overlay will be triggered

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
```diff
@@ -282,8 +282,11 @@ export default class DropDownMenu extends Widget<DropDownMenuProperties> {
       dragEnabled: false,
       showTitle: false,
       fullScreen: false,
+      ignoreChildEvents: false,
```
EugeniyKiyashko, line 280: After switching this option, the Esc key will also close the popup when the focus is inside the overflow menu popup. Seems okay, but several thinks are missing:
-- Test that the Esc key closes the overflow menu popup.
-- When the dropdown is open and the Esc key is pressed, both the dropdown and the overflow menu will close immediately. Is that okay?
-- Which element should have focus after the overflow menu is closed?

Perhaps the current fix does not aim to make the Esc key work as part of this bug and should be processed separately?

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
juleari: @EugeniyKiyashko 
is there style guide, that suggests it?

all one-line functions in this file don't have such wrapper:
https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts#L247
https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts#L249
https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts#L269
https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts#L416

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
juleari: added to https://github.com/DevExpress/devextreme-private/issues/1050

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
```diff
@@ -595,6 +599,85 @@ QUnit.module('widget sizing render', moduleConfig, () => {
...
+                ]
+            );
+
+            const checkBox = this.overflowMenu.$popupContent().find(`.${CHECK_BOX_CLASS}`).first();
+            const button = this.overflowMenu.$popupContent().find(`.${BUTTON_CLASS}`).first();
+
+            this.instance.option({ focusStateEnabled: true });
+            this.overflowMenu.list().option({ focusStateEnabled: true });
+
+            // checkBox.focusin();
```
EugeniyKiyashko:
```suggestion
            // checkBox.focusin();
```

Excess comments

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
EugeniyKiyashko: What about exclude this excess behavior from current fix?

EugeniyKiyashko: Right now, all the handlers that exist for the Popup will trigger, even if the Popup is not currently focused. One of them is for the Esc key, possibly some others as well. From the current fix, it’s clear that we don’t want any handlers to trigger except for space and enter, so we need to try disabling the others that might trigger

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
ksercs: I'd say it's a minor because such things are better to be handled with a linter. However, I agree with Eugeniy's proposition because:
- proposed code more clearly shows we don't expect any return value from `_itemClickHandler`
- it's more comfortable to set a debugger in this handler

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
```diff
@@ -337,6 +340,26 @@ export default class DropDownMenu extends Widget<DropDownMenuProperties> {
     });
   }
 
+  _popupKeyHandler(e: DxEvent<KeyboardEvent>): void {
+    if (
+      this._list && $(e.target).closest(this._list._activeStateUnit).length
```
ksercs: Cannot we achieve the goal without using of a list private API? It's not very good to use a private API out of the component because it'll need additional refactorings if the private api is changed.

Also I'd propose to move this stuff to a variable with a meaningful name to make this `if` more clear and declarative + simplify future debugging

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
```diff
@@ -32,6 +34,8 @@ const DROP_DOWN_MENU_POPUP_CLASS = 'dx-dropdownmenu-popup';
 const DROP_DOWN_MENU_POPUP_WRAPPER_CLASS = 'dx-dropdownmenu-popup-wrapper';
 const LIST_ITEM_CLASS = 'dx-list-item';
 const SCROLLVIEW_CONTENT_CLASS = 'dx-scrollview-content';
+const CHECK_BOX_CLASS = 'dx-checkbox';
+const BUTTON_CLASS = 'dx-button';
```
pharret31, line 39: Please export class constants from check_box.tsx and button.tsx files to be sure we use the same class names across the project

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
```diff
@@ -595,6 +599,85 @@ QUnit.module('widget sizing render', moduleConfig, () => {
...
+                        template: () => $('<div>').dxButton({ text: 'Button' })
+                    },
+                ]
+            );
+
+            const checkBox = this.overflowMenu.$popupContent().find(`.${CHECK_BOX_CLASS}`).first();
+            const button = this.overflowMenu.$popupContent().find(`.${BUTTON_CLASS}`).first();
+
+            this.instance.option({ focusStateEnabled: true });
+            this.overflowMenu.list().option({ focusStateEnabled: true });
```
ksercs: You already skipped this test for mobile devices, so maybe we don't need this anymore?

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
```diff
@@ -595,6 +599,85 @@ QUnit.module('widget sizing render', moduleConfig, () => {
...
+                        widget: 'dxCheckBox',
+                    },
+                    {
+                        template: () => $('<div>').dxButton({ text: 'Button' })
+                    },
+                ]
+            );
+
+            const checkBox = this.overflowMenu.$popupContent().find(`.${CHECK_BOX_CLASS}`).first();
+            const button = this.overflowMenu.$popupContent().find(`.${BUTTON_CLASS}`).first();
```
ksercs: We usually use `$` prefix to clearly show it's a jquery element and not a native element or component instance
```suggestion
            const $button = this.overflowMenu.$popupContent().find(`.${BUTTON_CLASS}`).first();
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
```diff
@@ -595,6 +599,85 @@ QUnit.module('widget sizing render', moduleConfig, () => {
...
+
+            this.instance.option({ focusStateEnabled: true });
+            this.overflowMenu.list().option({ focusStateEnabled: true });
+
+            // checkBox.focusin();
+            keyboardMock(checkBox)
+                .keyDown('space');
+            assert.equal(popup.option('visible'), true, 'space on dxCheckbox does not hide popup');
```
ksercs: Please let's try to keep tests as atomic as possible. For this the following practices can help:
1. AAA pattern (arrange - act - assert)
2. describing the whole test in the name. In your case the correct name for this test is smth like "popup should not be closed after space key press on a list item nested checkbox but should be closed after press on a list item nested button"

Now it looks like there are 2 different scenarios which are supposed to be tested separately.
And maybe there are more like "not close on nested selectBox opening"?

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
```diff
@@ -595,6 +599,85 @@ QUnit.module('widget sizing render', moduleConfig, () => {
...
+
+            const checkBox = this.overflowMenu.$popupContent().find(`.${CHECK_BOX_CLASS}`).first();
+            const button = this.overflowMenu.$popupContent().find(`.${BUTTON_CLASS}`).first();
+
+            keyboardMock(checkBox)
+                .keyDown('enter');
+            assert.equal(popup.option('visible'), true, 'enter on dxCheckbox does not hide popup');
+
+            keyboardMock(button)
+                .keyDown('enter');
```
ksercs: This test is completely the same as the test above except the key we press? Then we can reduce copypaste if we iterate with forEach. It'll be smth like
```javascript
['space', 'enter'].forEach(key => {
            QUnit.test(`popup should be closed after ${key} press on a list item nested button`, function(assert) {
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
juleari: there are no such exports in the moment. are you suggesting to add it and got through the project to use it everywhere?
it will be nice to do. but i'm not sure that this is an appropriate pr for that

EugeniyKiyashko: The Button and CheckBox components are renovated and use class definitions with methods, so this approach is not sufficient here. For other components, we should use the approach suggested by Andrey. 
Summary: No changes are needed in this PR; declaring the corresponding constant is enough

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
juleari: are you sure that writing non explicit test cases is a good idea?
they are the same now, but it may change in the future.
for example space and enter keys have a different effect on checkbox itself.

ksercs: Sure i'm against writing not explicit tests but moving key to a variable does not add non-intuitivity to me - it still should be well readable. For example, check [this TagBox test](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js#L2476) out - isn't it good to you?

> for example space and enter keys have a different effect on checkbox itself.

I would say this probability is very low. But anyway now it's only good to emphasize that behavior is the same for both keys

P.s. I still don't insist if you more like to have 2 different tests. But you will need to apply other refactoring notes in 2 places then

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
```diff
@@ -282,8 +282,28 @@ export default class DropDownMenu extends Widget<DropDownMenuProperties> {
...
+    ) => {
+      this._popupKeyHandler(e);
+    });
+    this._popup.registerKeyHandler('escape', (
+      e: DxEvent<KeyboardEvent>,
+    ) => {
+      // @ts-expect-error
+      if (this._popup?.$overlayContent().is($(e.target))) {
+        this.option('opened', false);
+      }
```
ksercs, line 298: Hm, why do we need this? Looks like a default popup handler 🤔

EugeniyKiyashko: Yes, we need this to minimize changes in behavior after the fix. The task to improve this behavior was moved to UI/UX Toolbar Enhancements, and such a change is better implemented only in the new version

EugeniyKiyashko (dismissed):

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
ksercs: Okay I gonna trust you

EugeniyKiyashko (approved):

PR #30710, opened 2025-08-11
Chat: support emptyViewTemplate option

Zedwag:


r-farkhutdinov (approved):

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.tests.js
```diff
@@ -758,6 +760,81 @@ QUnit.module('Chat', () => {
             });
         });
 
+        QUnit.module('emptyViewTemplate', () => {
```
ksercs, line 763: Hm, why do you test it here, not in the `messageList` tests? Looks like here we should test only passing to MessageList because the rest is implemented not here. no?

ksercs (approved):

packages/devextreme/testing/tests/DevExpress.ui.widgets/chatParts/chat.tests.js
Zedwag: I decided to keep similar to `messageTemplate` tests in the same place. Currenlty it tested on chat level

pomahtri (approved):

PR #30716, opened 2025-08-11
DataGrid(T1278804):  bounces to the top on an attempt to scroll down when an adaptive row is expanded and virtual scrolling is enabled

wdevfx:


e2e/testcafe-devextreme/tests/dataGrid/common/scrolling.ts
```diff
@@ -1824,3 +1824,72 @@ test('DataGrid - Scrolling position is reset to far right on an attempt to scrol
...
+      const hasSmallScrollValues = scrollOffsets.some((offset) => offset < scrollValuesThreshold);
+      await t.expect(hasSmallScrollValues).notOk();
+    },
+  ).before(async () => {
+    const columns = new Array(100)
+      .fill(null)
+      .map((_, idx) => `col_${idx}`);
+
+    await createWidget('dxDataGrid', {
+      dataSource: new Array(3)
```
Tucchhaa: Why not use `getData(3, 100).map((item, index) => ({ ...item, id: index }))`?
I think it's more readable than reduce.

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -3006,7 +3006,7 @@ const adaptiveColumns = (Base: ModuleType<AdaptiveColumnsController>) => class A
   protected _hideVisibleColumnInView({ view, isCommandColumn, visibleIndex }) {
     super._hideVisibleColumnInView({ view, isCommandColumn, visibleIndex });
     if (view.name === ROWS_VIEW) {
-      this._rowsView.renderFocusState(null);
+      this._rowsView.renderFocusState({ preventScroll: true });
```
Raushen: I noticed there is [the util method ](https://github.com/DevExpress/DevExtreme/blob/a4d2c571d1e67f888a72f6183334af864fd100b1/packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation_utils.ts#L64)checking should we prevent scrolling or not, taking into account virtual scrolling mode. Will it be useful here?
If so, It might be the common implementation for DataGrid. And it can be implemented on the base level by default - to not forget to perform such calculations the next time this method used.

e2e/testcafe-devextreme/tests/dataGrid/common/scrolling.ts
wdevfx: Yup, thank u for a catch! 🙌

I totally missed this helper function
Unfortunately, `getData` has some disadvantage (it create unique object keys, we need the same for different rows)
I fixed it too and it shouldn't affect other tests because in these tests only one row is created in data :)

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
wdevfx: Yup, thank u for a great suggestion!

I re-checked and added usage of this helper utils function here
But for base common implementation -> let's research and implement it in terms of another task
I'm afraid that it may takes time and maybe here additional nuances that we should take into account :)

Tucchhaa (approved):

Raushen (approved):

PR #30717, opened 2025-08-12
Chart: Update Pie Label Overlap demo description

arman-boyakhchyan:


apps/demos/Demos/Charts/PieResolveLabelOverlap/description.md
```diff
@@ -1,2 +1,4 @@
-In&nbsp;the PieChart, series may include a&nbsp;large number of&nbsp;points, which may result in&nbsp;point label overlapping. This demo illustrates the **resolveLabelOverlapping** property that allows you to&nbsp;specify how the component must behave when point labels overlap.
-<!--split-->
\ No newline at end of file
+DevExtreme PieChart can automatically resolve label overlapping issues when displaying a large number of points. To implement this functionality, specify the [resolveLabelOverlapping](/Documentation/ApiReference/UI_Components/dxPieChart/Configuration/#resolveLabelOverlapping) property.
```
khamlyuk: This introduction sounds negative, as if our implementation was buggy. Let's add a positive angle :)

```suggestion
When a DevExtreme Pie Chart contains a lot of data points or they are crowded, point labels may overlap. Use the [resolveLabelOverlapping](/Documentation/ApiReference/UI_Components/dxPieChart/Configuration/#resolveLabelOverlapping) property to resolve overlapping labels.
```

apps/demos/Demos/Charts/PieResolveLabelOverlap/description.md
```diff
@@ -1,2 +1,4 @@
...
+<!--split-->
+
+PieChart supports two **resolveLabelOverlapping** modes: *"hide"* and *"shift"*. To test these modes in this demo, select an item in the SelectBox below the PieChart.
```
khamlyuk: As far as I understand, this part of the description will appear below the demo. It could be inconvenient to scroll down to read about demo functionality.
Plus, in my opinion, mode names are self-explanatory and visual changes in demo will say everything for us. All in all we have two options:
1. We can remove this paragraph completely.
2. We can try to express the same idea in one sentence and move it to the first paragraph. For example, "In this demo, you can hide or shift overlapping labels".

arman-boyakhchyan: Yeah, everything after the  tag appears below the demo container. I'll add your second suggestion point, I think it will give the description a nice conclusion

pharret31 (approved):

PR #30723, opened 2025-08-12
DropDownEditor: Add fieldAddons to .d.ts

marker-dao:


packages/devextreme/js/ui/drop_down_editor/ui.drop_down_list.d.ts
```diff
@@ -45,7 +45,10 @@ export interface SelectionChangedInfo<T = any> {
  * @docid
  * @hidden
  */
-export interface dxDropDownListOptions<TComponent> extends DataExpressionMixinOptions<TComponent>, dxDropDownEditorOptions<TComponent> {
+export interface dxDropDownListOptions<TComponent> extends DataExpressionMixinOptions<TComponent>, Omit<
```
ksercs, line 48: Why? We need this API in SelectBox and TagBox inherited from ddList.
Look at generation in frameworks - there is no new API for these components now - [link](https://github.com/DevExpress/DevExtreme/pull/30723/files#diff-4c797649be6ce2ad6ad606625607620f63965ca71c01c688f46fa54390cfe950R203)

packages/devextreme/js/ui/drop_down_editor/ui.drop_down_editor.d.ts
```diff
@@ -86,6 +100,13 @@ export interface dxDropDownEditorOptions<TComponent> extends Omit<dxTextBoxOptio
      * @public
      */
     dropDownButtonTemplate?: template | ((buttonData: DropDownButtonTemplateDataModel, contentElement: DxElement) => string | UserDefinedElement);
+    /**
+     * @docid
+     * @default null
+     * @type object
```
ksercs: Is it required? It degrades types - [link](https://github.com/DevExpress/DevExtreme/pull/30723/files#diff-5b5792fb1577c537f54f5a3f6edaf6f5dc74cad5f5e408bb70568a2f1a137aa2R169)

packages/devextreme/js/ui/drop_down_editor/ui.drop_down_list.d.ts
marker-dao: > Why? We need this API in SelectBox and TagBox inherited from ddList.

We don't need this option in all other ddLists. Therefore, it was added separately for SelectBox: e3f3a6143349e9ee56cca0b9c07ad7f779000e68

packages/devextreme/js/ui/drop_down_editor/ui.drop_down_editor.d.ts
marker-dao: fixed

r-farkhutdinov (dismissed):

packages/devextreme/js/ui/drop_down_editor/ui.drop_down_list.d.ts
ksercs: It looks for me like it'd be more clear if we implement it by explicit including in necessary components how it was earlier done for fieldTemplate. Doesn't it look better to u?

ksercs: Discussed in voice - we think it's better to have types specifed on the same level where implementation is

ksercs (dismissed):

ksercs:
Looks strange to me that we don't have changes in react wrappers - I'd expect to see at least generated aliases like for [fieldTemplate](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme-react/src/select-box.ts#L55) 🤔 But let's keep tuning this part with api reviewers

marker-dao:
> Looks strange to me that we don't have changes in react wrappers - I'd expect to see at least generated aliases like for [fieldTemplate](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme-react/src/select-box.ts#L55) 🤔 But let's keep tuning this part with api reviewers

I have dealt with this matter.

ksercs (approved):

pomahtri (approved):

PR #30727, opened 2025-08-12
DropDownEditor: Deprecate fieldTemplate option in .d.ts

marker-dao:


packages/devextreme/js/ui/color_box.d.ts
```diff
@@ -197,6 +197,7 @@ export interface dxColorBoxOptions extends dxDropDownEditorOptions<dxColorBox> {
      * @default null
      * @type_function_return string|Element|jQuery
      * @public
+     * @deprecated
```
ksercs: Shouldn't we add a deprecation note?
```suggestion
     * @deprecated Use fieldTemplates instead
```

marker-dao: Comments on the @deprecated tag are not possible for object fields.

ksercs (dismissed):

packages/devextreme/js/ui/color_box.d.ts
ksercs: Aren't these places the same?
- DataGrid rowTemplate - [link](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/ui/data_grid.d.ts#L1385)
- Scheduler dropDownAppointmentTemplate - [link](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/ui/scheduler.d.ts#L602)

marker-dao: The comment cannot be arbitrary:
the current interface + replacement option must be specified.

r-farkhutdinov (approved):

Zedwag (approved):

anna-shakhova (approved):

PR #30732, opened 2025-08-13
Scheduler: refactor view model output and sorted Index

Ambrozy:


wdevfx (dismissed):

packages/devextreme/js/__internal/scheduler/__tests__/appointments.test.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  describe, expect, it,
+} from '@jest/globals';
+
+import { createScheduler } from './__mock__/create_scheduler';
+import { setupSchedulerTestEnvironment } from './__mock__/m_mock_scheduler';
+
+describe('Appointments', () => {
+  it('All-day appointment should not be resizable if current view is "day"', async () => {
```
aleksei-semikozov, line 9: Can you clarify?  `appointment should not be resizable` but in assertion block appointment.classList.contains('dx-resizable') should be resizeble.
Same in test name - `Appointment should **not** re-rendered on window resize when width and height have percent value (T1139566)` but in the assertion block we are not checking whether it was re-rendered or not. Why was the test renamed this way

packages/devextreme/js/__internal/scheduler/appointments/utils/get_arrays_diff.test.ts
```diff
@@ -0,0 +1,205 @@
...
+  describe, expect, it,
+} from '@jest/globals';
+
+import { getArraysDiff, isNeedToAdd, isNeedToRemove } from './get_arrays_diff';
+
+interface Obj { id: number; name: string }
+
+const compare = (a: Obj, b: Obj): boolean => a.id === b.id && a.name === b.name;
+
+const getOperations = <T>(items: ReturnType<typeof getArraysDiff<T>>): string => items
```
aleksei-semikozov, line 11: This is a great finding — it looks like these methods are not only useful for testing, but could also define the API between submodules.
Currently, isNeedToAdd and isNeedToRemove seem somewhat scattered and are used only in tests.
An API that directly states whether an item should be added, left unchanged, or removed would be much more convenient than passing separate flags like needToAdd: true, etc.
As a suggestion for future refactoring, it might be worth applying this approach both in tests and in the submodules’ API to make it consistent and easier to work with.

packages/devextreme/js/__internal/scheduler/appointments/utils/get_arrays_diff.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export function getArraysDiff<A, B = A>(
+  a: A[],
+  b: B[],
+  equal: (x: A, y: B) => boolean,
+): DiffItem<A, B>[] {
+  const n = a.length;
+  const m = b.length;
+
+  const dp: number[][] = Array.from({ length: n + 1 }, () => new Array<number>(m + 1).fill(0));
```
aleksei-semikozov, line 26: An `n*m` array can be quite memory-intensive (and thus impact performance). We could potentially reduce memory usage here.

Ambrozy: The performance tests result:
1. 1000 of recurrence appointments in timeline view without virtual scrolling. Before: 5.57s, After: 1.5s
2. 100 appointments for 100 resources with virtual scrolling. Before: 1.73s, After: 1.326s

packages/devextreme/js/__internal/scheduler/__tests__/appointments.test.ts
Ambrozy: The condition is correct `!appointment.classList.contains('dx-resizable')`. if appointment resizable it has `dx-resizable` class. Otherwise no class.

According to `Appointment should **not** re-rendered on window resize when width and height have percent value (T1139566)`. Because new change detection mechanism is better track changes, this cases not trigger rerender anymore. In the ticket user complained to often rerender. S it's  become better.

wdevfx (approved):

PR #30739, opened 2025-08-13
MenuBase: refactor _updateSelectedItems method

juleari:
_updateSelectedItems in menu_base had 1purpose to call _fireSelectionChangeEvent if needed
since it does not take collections as arguments, its name was changed to _updateSelectedItem and the order of old/new items was changed, to correspond to the same params in _fireSelectionChangeEvent method

added tests to ensure that resulted onSelectionChanged is called with the same arguments as before this changes

### DIFF
--- a/packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
+++ b/packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
@@ -737,7 +737,6 @@ class MenuBase<
             this._toggleItemSelection(selectedNode, false);
           }
           this._toggleItemSelection(node, true);
-          this._updateSelectedItems();
         }
         break;
       }
@@ -779,25 +778,24 @@ class MenuBase<
     return result;
   }
 
-  // @ts-expect-error ts-error
-  _updateSelectedItems(
-    oldSelection?: Item,
-    newSelection?: Item | null,
+  _updateSelectedItem(
+    addedItem?: Item | null,
+    removedItem?: Item,
   ): void {
-    if (oldSelection || newSelection) {
-      this._fireSelectionChangeEvent(newSelection, oldSelection);
+    if (addedItem || removedItem) {
+      this._fireSelectionChangeEvent(addedItem, removedItem);
     }
   }
 
   _fireSelectionChangeEvent(
-    addedSelection: Item | null | undefined,
-    removedSelection: Item | undefined,
+    addedItem?: Item | null,
+    removedItem?: Item,
   ): void {
     this._createActionByOption('onSelectionChanged', {
       excludeValidators: ['disabled', 'readOnly'],
     })({
-      addedItems: [addedSelection],
-      removedItems: [removedSelection],
+      addedItems: [addedItem],
+      removedItems: [removedItem],
     });
   }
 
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/ui/context_menu/menu_base.ts
```diff
@@ -779,25 +778,24 @@ class MenuBase<
     return result;
   }
 
-  // @ts-expect-error ts-error
-  _updateSelectedItems(
```
EugeniyKiyashko, line 783: this._updateSelectedItems() is implemented in parent components. The selection module calls it on SelectionChanged when args.addedItemKeys.length or args.removedItemKeys.length is non-zero. I see that the options for this are not public in Menus, but if you are sure that all processing of selected items happens on the MenuBase side and the selection module shouldn’t perform any updates, we need to override _updateSelectedItems() with an empty function

packages/devextreme/testing/tests/DevExpress.ui.widgets/menuBase.tests.js
```diff
@@ -1031,6 +1031,57 @@ QUnit.module('Selection', () => {
...
+        ];
+        const selectionChangedHandler = sinon.stub();
+        const menuBase = createMenu({
+            items: items,
+            selectionMode: 'single',
+            selectByClick: true,
+            onSelectionChanged: selectionChangedHandler,
+        });
+
+        const $item2 = menuBase.element.find('.' + DX_MENU_ITEM_CLASS).eq(1);
```
EugeniyKiyashko: Use template literals instead

packages/devextreme/testing/tests/DevExpress.ui.widgets/menuBase.tests.js
```diff
@@ -1031,6 +1031,57 @@ QUnit.module('Selection', () => {
...
+        $item2.trigger('dxclick');
+        assert.equal(selectionChangedHandler.callCount, 1, 'onSelectionChanged was called for the first time');
+        assert.deepEqual(selectionChangedHandler.args[0][0].addedItems, [{ text: 'item2', selected: true }], 'onSelectionChanged is called with selected item as added item');
+        assert.deepEqual(selectionChangedHandler.args[0][0].removedItems, [null], 'onSelectionChanged first called with null as removed item');
+
+        const $item1 = menuBase.element.find('.' + DX_MENU_ITEM_CLASS).eq(0);
+        $item1.trigger('dxclick');
```
EugeniyKiyashko, line 1056: You need to split the arrange, act, and assert sections inside the test to improve its readability. Add an empty line between them.

packages/devextreme/testing/tests/DevExpress.ui.widgets/menuBase.tests.js
```diff
@@ -1031,6 +1031,57 @@ QUnit.module('Selection', () => {
...
+        const menuBase = createMenu({
+            items: items,
+            selectionMode: 'single',
+            selectByClick: true,
+            onSelectionChanged: selectionChangedHandler,
+        });
+
+        const $item2 = menuBase.element.find('.' + DX_MENU_ITEM_CLASS).eq(1);
+        $item2.trigger('dxclick');
+        assert.equal(selectionChangedHandler.callCount, 1, 'onSelectionChanged was called for the first time');
```
pharret31: Please prefer the strictEqual checks, we aim to change all equal to strictEqual in code to be consistent and strict on checks

EugeniyKiyashko (approved):

PR #30754, opened 2025-08-14
DataGrid: Fix updating of the focused row index after a row update when the repaintChangesOnly option is enabled (T1292991)

Alyar666:


e2e/testcafe-devextreme/tests/dataGrid/common/focus/focusedRow/focusedRow.ts
```diff
@@ -987,3 +987,56 @@ test('Group should expand when focusedRowKey is set and data items have \'items\
   height: 400,
   focusedRowEnabled: true,
 }));
+
+// T1292991
+test('Focused row index should be updated after updating a row when repaintChangesOnly is enabled', async (t) => {
```
Tucchhaa: Suggestion: since the fix was quite complex with taking into account previous tickets and refactoring, can we add matrix tests that will cover the issue more widely? I think this will help us to reduce chances of receiving again similar issues from customers.

Like: one loop for repaintChangesOnly: [true, false] and another loop for changeType: ['insert', 'remove', 'update']. And in the test we check that focusedRowKey and focusedRowIndex are correct and that focusedRowChanged, focusedRowChanging events are called.

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
```diff
@@ -572,20 +560,23 @@ const data = (Base: ModuleType<DataController>) => class FocusDataControllerExte
...
-      if (this._needToUpdateFocusedRowByIndex) {
-        this._needToUpdateFocusedRowByIndex = false;
-        this._focusController._focusRowByIndex();
+      if (forceUpdateFocusedRow && this.isEmpty()) {
+        this._focusController._resetFocusedRow();
       } else if (e.changeType === 'refresh' && e.items.length || isPartialUpdateWithDeleting) {
         this._updatePageIndexes();
-        this._updateFocusedRow(e);
+        this._updateFocusedRowIfNeeded(e, forceUpdateFocusedRow);
       } else if (e.changeType === 'append' || e.changeType === 'prepend') {
```
Tucchhaa, line 576: question: do we need to update focusesRow if changeType is 'append' or 'prepend'?

e2e/testcafe-devextreme/tests/dataGrid/common/focus/focusedRow/focusedRow.ts
Tucchhaa: Also I suggest to rewrite the test to jest, because this test is fully functional

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
Alyar666: No, that’s not necessary. The changeType with the value 'append'/'prepend' appears during virtual scrolling in legacy mode (scrolling.legacyMode = true). Since the data does not change during virtual scrolling and focusedRowIndex stores the global row index, there is no need to update focusedRow.

Tucchhaa (approved):

PR #30761, opened 2025-08-14
LoadPanel: Improve typing

marker-dao:


packages/devextreme/js/__internal/ui/m_load_panel.ts
```diff
@@ -212,7 +220,8 @@ class LoadPanel extends Overlay<LoadPanelProperties> {
   _cleanPreviousContent(): void {
     this.$content().find(`.${LOADPANEL_MESSAGE_CLASS}`).remove();
     this.$content().find(`.${LOADPANEL_INDICATOR_CLASS}`).remove();
-    delete this._$indicator;
+
+    this._$indicator = undefined;
```
r-farkhutdinov, line 224: [Minor]

Manual setting of `undefined`. Can we set `null` instead?

r-farkhutdinov (approved):

packages/devextreme/js/__internal/ui/m_load_panel.ts
marker-dao: According to the type system, only undefined is allowed.

PR #30768, opened 2025-08-14
Selection: improve TS typing

juleari:


packages/devextreme/js/__internal/ui/collection/collection_widget.edit.ts
```diff
@@ -224,6 +225,7 @@ class CollectionWidget<
     const { itemsGetter } = this._editStrategy;
     const { selectionMode, maxFilterLengthInRequest } = this.option();
 
+    // @ts-expect-error TItem
     this._selection = new Selection({
```
anna-shakhova:
```suggestion
    this._selection = new Selection<TItem>({
```
this solves ts issues in callbacks and in other properties as well

packages/devextreme/js/__internal/ui/collection/collection_widget.edit.ts
```diff
@@ -479,7 +485,9 @@ class CollectionWidget<
 
         const { grouped } = this.option();
 
+        // @ts-expect-error items inside TItem
```
anna-shakhova: we can use type guard here, smth like 
```suggestion
        const hasSubItems = (item: TItem): item is TItem & { items: TItem[] } => isObject(item) && 'items' in item && Array.isArray(item.items);

        if (grouped && hasSubItems(normalizedSelection)) {
```

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.deferred.ts
```diff
@@ -3,35 +3,47 @@ import type { DeferredObj } from '@js/core/utils/deferred';
...
     this._loadFilteredData(this.options.selectionFilter, null, select).done((items) => {
-      // @ts-expect-error
-      const keys = items.map((item) => that.options.keyOf(item));
+      const keys = (Array.isArray(items) ? items : []).map(getKey);
 
       d.resolve(keys);
-    // eslint-disable-next-line @typescript-eslint/no-misused-promises
-    }).fail(d.reject);
+    }).fail((error) => {
+      // @ts-expect-error error
```
anna-shakhova, line 30: minor: no need to rewrite 
```
    .fail(d.reject)
```
with 
```
    .fail((e) => {
        d.reject(e);
    })
```
for me 1st syntax more readable as it is shorter

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.deferred.ts
```diff
@@ -148,31 +165,46 @@ export default class DeferredStrategy extends SelectionStrategy {
...
+      filter = [filter];
     }
-    if (selectionFilter.length) {
-      selectionFilter.push(filterOperator);
+    if (Array.isArray(filter) && filter.length) {
+      filter.push(filterOperator);
     }
-    return selectionFilter;
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
anna-shakhova, line 193: I would use type assertion here, as SelectionFilter is in fact `any[]` in public types. You can keep this solution if you think this is better way

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
```diff
@@ -58,32 +72,36 @@ export default class StandardStrategy extends SelectionStrategy {
...
       const key = keyOf(item);
       if (isDeselect) {
-        keyIndex = this.removeSelectedItem(key, keyIndicesToRemoveMap, item?.disabled);
-        if (keyIndicesToRemoveMap && keyIndex >= 0) {
+        const keyIndex = this.removeSelectedItem(
+          key,
+          keyIndicesToRemoveMap,
+          item && typeof item === 'object' && 'disabled' in item ? !!item.disabled : false,
+        );
+        if (keyIndicesToRemoveMap && typeof keyIndex === 'number' && keyIndex >= 0) {
```
anna-shakhova:
```suggestion
        if (keyIndicesToRemoveMap && isNumeric(keyIndex) && keyIndex >= 0) {
```

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
```diff
@@ -58,32 +72,36 @@ export default class StandardStrategy extends SelectionStrategy {
...
           keyIndicesToRemoveMap[keyIndex] = true;
         }
       } else {
         this.addSelectedItem(key, item);
       }
-    }
+    });
 
     if (isBatchDeselect) {
```
anna-shakhova, line 91:
```suggestion
    if (isBatchDeselect && keyIndicesToRemoveMap) {
      this._batchRemoveSelectedItems(keyIndicesToRemoveMap);
    }
```

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
```diff
@@ -135,50 +172,48 @@ export default class StandardStrategy extends SelectionStrategy {
...
+    isDeselect?: boolean,
+    updatedKeys?: TKey[],
+  ): RequestData<TItem, TKey> {
+    let selectedItems: TKey[] = [];
     const deselectedItems = isDeselect ? keys : [];
 
     if (updatedKeys) {
       selectedItems = updatedKeys;
     } else {
+      // @ts-expect-error removeDuplicates
```
anna-shakhova, line 216: let's add type `any[]`to `removeDuplicates` method

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
```diff
@@ -189,75 +224,121 @@ export default class StandardStrategy extends SelectionStrategy {
...
-      currentKeys = removeDuplicates(keys.concat(this._lastRequestData?.addedItems), this._lastRequestData?.removedItems);
+    if (
+      this._isMultiSelectEnabled()
+      && this._shouldMergeWithLastRequest
+      && !isDeselect
+      && !isSelectAll
+    ) {
+      currentKeys = removeDuplicates(
+        // @ts-expect-error removeDuplicates
+        keys.concat(this._lastRequestData?.addedItems),
```
anna-shakhova: spread operator allows concatenate arrays with different item types

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
```diff
@@ -153,42 +170,53 @@ export default class SelectionStrategy {
     };
   }
 
-  _loadFilteredData(remoteFilter, localFilter?: any, select?: any, isSelectAll?: boolean) {
-    const filterLength = encodeURI(JSON.stringify(this._removeTemplateProperty(remoteFilter))).length;
-    const needLoadAllData = this.options.maxFilterLengthInRequest && (filterLength > this.options.maxFilterLengthInRequest);
-    const deferred = Deferred();
+  _loadFilteredData(
+    remoteFilter: SelectionFilter,
+    localFilter?: unknown,
```
anna-shakhova: From usage
```
    localFilter?: Function | null,
    select?: Function | string[] | null,
```

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
```diff
@@ -250,34 +278,57 @@ export default class SelectionStrategy {
...
+    updatedKeys?: TKey[],
+    // eslint-disable-next-line  @typescript-eslint/no-unused-vars
+    forceCombinedFilter?: boolean,
+  ): DeferredObj<TItem[]> {
+    throw new Error('selectedItemKeys method should be overriden');
+  }
+
+  isItemKeySelected(itemKey: TKey | TItem): boolean;
+  // eslint-disable-next-line  @typescript-eslint/no-unused-vars, class-methods-use-this
+  isItemKeySelected(itemKey: TKey | TItem, options: { checkPending?: boolean } = {}): boolean {
```
anna-shakhova, line 312: option type repeated in several placed, let's create type alias or interface for it

packages/devextreme/js/__internal/ui/selection/m_selection.ts
```diff
@@ -1,30 +1,40 @@
...
+export default class Selection<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  TItem extends SelectionItem = any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  TKey extends string | number = any,
+  TDeferred extends boolean = boolean,
+> {
+  options: SelectionOptions<TItem, TKey, TDeferred>;
+
+  _selectionStrategy: SelectionStrategy<TItem, TKey, TDeferred>;
```
anna-shakhova, line 25:
```suggestion
  _selectionStrategy: DeferredStrategy<TItem, TKey> | StandardStrategy<TItem, TKey>;
```

packages/devextreme/js/__internal/ui/selection/m_selection.ts
```diff
@@ -33,8 +43,9 @@ export default class Selection {
     }
   }
 
-  _getDefaultOptions() {
-    return {
+  // eslint-disable-next-line class-methods-use-this
+  _getDefaultOptions(): DefaultOptions<TItem, TKey, false> {
```
anna-shakhova, line 49: _getDefaultOptions method returns component options, here it should be SelectedOptions<...>

packages/devextreme/js/__internal/ui/selection/m_selection.ts
```diff
@@ -1,30 +1,40 @@
...
+  _selectionStrategy: SelectionStrategy<TItem, TKey, TDeferred>;
 
   _focusedItemIndex: number;
 
-  _shiftFocusedItemIndex!: number;
+  _shiftFocusedItemIndex?: number;
 
-  constructor(options) {
+  constructor(options: Partial<SelectionOptions<TItem, TKey, TDeferred>>) {
     this.options = extend(this._getDefaultOptions(), options, {
```
anna-shakhova, line 32: extend looses typings,
here you can use spread operator to create options objects
and then issues with this.option has type `any` will be solved
```suggestion
    this.options = {
      ...this._getDefaultOptions(),
      ...options,
      selectedItemKeys: options.selectedKeys ?? [],
    };
```

packages/devextreme/js/__internal/ui/selection/m_selection.ts
```diff
@@ -43,109 +54,134 @@ export default class Selection {
...
   }
 
-  clearSelection() {
+  clearSelection(): DeferredObj<TItem[]> {
     return this.selectedItemKeys([]);
   }
 
-  _addSelectedItem(itemData, key) {
+  _addSelectedItem(itemData: TItem, key: TKey): void {
+    // @ts-expect-error addSelectedItem
```
anna-shakhova, line 150: there type assertion will work
```suggestion
    (this._selectionStrategy as StandardStrategy).addSelectedItem(key, itemData);
```
better: create type guard and add check
```suggestion
      _isStandardStrategy(strategy: StandardStrategy | DeferredStrategy): strategy is StandardStrategy {
    return this.options.deferred;
  }

  _addSelectedItem(itemData: TItem, key: TKey): void {
    if (this._isStandardStrategy(this._selectionStrategy)) {
      this._selectionStrategy.addSelectedItem(key, itemData);
    }
  }
```

packages/devextreme/js/__internal/ui/selection/m_selection.ts
```diff
@@ -43,109 +54,134 @@ export default class Selection {
...
+    itemIndex: number,
+    keys: { control?: boolean; shift?: boolean } = {},
+    setFocusOnly?: boolean,
+  ): boolean | undefined {
+    let isSelectedItemsChanged = false;
     const items = this.options.plainItems();
     const item = items[itemIndex];
+    let focusedItemIndex = itemIndex;
+    // eslint-disable-next-line @typescript-eslint/init-declarations
     let deferred;
```
anna-shakhova:
```suggestion
    let deferred: DeferredObj<unknown>;
```

packages/devextreme/js/__internal/ui/selection/m_selection.ts
```diff
@@ -43,109 +54,134 @@ export default class Selection {
...
     const { isVirtualPaging } = this.options;
     const allowLoadByRange = this.options.allowLoadByRange?.();
     const { alwaysSelectByShift } = this.options;
+    // eslint-disable-next-line @typescript-eslint/init-declarations
     let indexOffset;
```
anna-shakhova:
```suggestion
    let indexOffset: number;
```

packages/devextreme/js/__internal/ui/selection/m_selection.ts
```diff
@@ -43,109 +54,134 @@ export default class Selection {
...
     let focusedItemNotInLoadedRange = false;
     let shiftFocusedItemNotInLoadedRange = false;
 
-    const itemIsNotInLoadedRange = (index) => index >= 0 && !items.filter((it) => it.loadIndex === index).length;
+    const itemIsNotInLoadedRange = (index: number): boolean => index >= 0 && !items.filter(
```
anna-shakhova, line 185: nitpick: difficult to read, worth to create a variable for second part with some good name

packages/devextreme/js/__internal/ui/selection/m_selection.ts
```diff
@@ -263,15 +318,20 @@ export default class Selection {
       this._shiftFocusedItemIndex = this._focusedItemIndex;
     }
 
-    let data;
+    let itemIndexStep = 0;
+    // eslint-disable-next-line @typescript-eslint/init-declarations
     let itemKey;
```
anna-shakhova:
```suggestion
    let itemKey: TKey;
```

packages/devextreme/js/__internal/ui/selection/types.ts
```diff
@@ -0,0 +1,125 @@
...
+import type { DeferredObj } from '@js/core/utils/deferred';
+import type DeferredStrategy from '@ts/ui/selection/m_selection.strategy.deferred';
+import type StandardStrategy from '@ts/ui/selection/m_selection.strategy.standard';
+
+export type SelectionFilter = FilterDescriptor[];
+
+type Filter = () => SelectionFilter | undefined;
+export interface SelectionItem {
+  disabled?: boolean;
+  loadIndex: number;
```
anna-shakhova, line 13: is it required? looks like collection item use selection well without it

anna-shakhova (changes requested):

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.deferred.ts
juleari: e may be short for event as well as for error. that's why i think it's better to not use e.
but maybe there is no need to pass error to reject at all. i suggested this investigation in the refactoring issue

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.deferred.ts
juleari: i think this one is better, because when we change the type to more accurrate -- this comment will become unnecessary and will be shown as warning.
in case of assertion -- no warning will be made and will probably stay that way.

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
juleari: are you sure this is the right pr to do that?
looks like a lot of methods from m_array needs type improvement. and i don't think they should be made in pr which not about arrays

packages/devextreme/js/__internal/ui/selection/m_selection.ts
juleari: this may change the actual behavior of this method so i suggested to explore this in the next issue (refactoring card has this as one of the items)

packages/devextreme/js/__internal/ui/selection/m_selection.ts
juleari: the first part would not make sense if the second part would be a variable -- since going through an array would not be needed if index will be -1;
but  i can refactor it a bit using `some` instead of `filter...length`. but since this is a refactoring-refactoring i added it to the refactor card

packages/devextreme/js/__internal/ui/selection/types.ts
juleari: it is required for changeItemSelection method

packages/devextreme/js/__internal/ui/collection/collection_widget.edit.ts
juleari: fixed in: https://github.com/DevExpress/DevExtreme/pull/30768/commits/7dfa3fcfb4071bac6bb77c47615c9dd41725df4d

packages/devextreme/js/__internal/ui/collection/collection_widget.edit.ts
juleari: fixed in https://github.com/DevExpress/DevExtreme/pull/30768/commits/bac2839415ee638ae839913f6cb7856a12d253e5

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
juleari: fixed in https://github.com/DevExpress/DevExtreme/pull/30768/commits/7dfa3fcfb4071bac6bb77c47615c9dd41725df4d#diff-ef5e271ee54eec3cb1e6db61637108ca986c490bfa149fe9d10a8869b6e18eba

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
juleari: fixed in https://github.com/DevExpress/DevExtreme/pull/30768/commits/7dfa3fcfb4071bac6bb77c47615c9dd41725df4d#diff-ef5e271ee54eec3cb1e6db61637108ca986c490bfa149fe9d10a8869b6e18eba

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.standard.ts
juleari: fixed in https://github.com/DevExpress/DevExtreme/pull/30768/commits/7dfa3fcfb4071bac6bb77c47615c9dd41725df4d#diff-ef5e271ee54eec3cb1e6db61637108ca986c490bfa149fe9d10a8869b6e18eba

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
juleari: fixed in https://github.com/DevExpress/DevExtreme/pull/30768/commits/7dfa3fcfb4071bac6bb77c47615c9dd41725df4d#diff-68e40718c36f32e88e9c05867aa5e4935c0a16c1df8093833076a0723e142e03

packages/devextreme/js/__internal/ui/selection/m_selection.strategy.ts
juleari: fixed in https://github.com/DevExpress/DevExtreme/pull/30768/commits/7dfa3fcfb4071bac6bb77c47615c9dd41725df4d#diff-68e40718c36f32e88e9c05867aa5e4935c0a16c1df8093833076a0723e142e03

packages/devextreme/js/__internal/ui/selection/m_selection.ts
juleari: as discussed, fixed in https://github.com/DevExpress/DevExtreme/pull/30768/commits/7dfa3fcfb4071bac6bb77c47615c9dd41725df4d#diff-68e40718c36f32e88e9c05867aa5e4935c0a16c1df8093833076a0723e142e03, but left SelectionStrategy type in case we will want it in the future

packages/devextreme/js/__internal/ui/selection/m_selection.ts
juleari: as we discussed, i left different types for default and settable options, so some options MUST be setted on instance creation.

packages/devextreme/js/__internal/ui/selection/m_selection.ts
juleari: fixed in https://github.com/DevExpress/DevExtreme/pull/30768/files#diff-20c3dbd6cf08c2c28dabf97cda749aab84e653438cd15f496c85e6b3370a61fc

packages/devextreme/js/__internal/ui/selection/m_selection.ts
juleari: fixed in https://github.com/DevExpress/DevExtreme/pull/30768/files#diff-20c3dbd6cf08c2c28dabf97cda749aab84e653438cd15f496c85e6b3370a61fc

packages/devextreme/js/__internal/ui/selection/m_selection.ts
juleari: fixed in https://github.com/DevExpress/DevExtreme/pull/30768/files#diff-20c3dbd6cf08c2c28dabf97cda749aab84e653438cd15f496c85e6b3370a61fc

packages/devextreme/js/__internal/ui/selection/m_selection.ts
juleari: aaand reverted this changes because it broke too many random grid tests. will investigate it in refactoring pr

anna-shakhova (dismissed):

packages/devextreme/js/__internal/grids/grid_core/selection/m_selection.ts
```diff
@@ -299,6 +299,7 @@ export class SelectionController extends modules.Controller {
   private _createSelection() {
     const options = this._getSelectionConfig();
 
+    // @ts-expect-error TKey
     return new Selection(options);
```
wdevfx, line 303: Let's change this `ts-expect-error` a little bit
The problem that `this._getSelectionConfig` returned object missed some required properties from the 
`Selection` -> `DefaultOptions` type

So, from my perspective better to highlight this type inconsistency and move `ts-expect-error` to the `options` line:
```suggestion
    // @ts-expect-error inconsistent types with SelectionOptions and this._getSelectionConfig
    const options: SelectionOptions = this._getSelectionConfig();
    
    return new Selection(options);
```

As a further TS improvement steps in the future, from Grids team side we will try to make this types consistent and remove this `ts-expect-error` :)

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts
```diff
@@ -365,6 +365,7 @@ export class SelectionController {
   }
 
   public getSelectedCardsData(): DataObject[] {
+    // @ts-expect-error selectedItems can return a promise for deferred selection
     return this.selectionHelper?.peek()?.getSelectedItems();
   }
```
wdevfx: Only the `DataGrid` component can have the `deferred` selection strategy
So, in the `CardView` and `TreeList` it's always not `deferred` one (as far as I know)

[Link to the DataGrid public API docs](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/selection/#deferred)
If u will try to find selection.deferred for `CardView`/`TreeList` -> there is nothing in public API about it :)

So, let's just add a type for `this.selectionHelper` instead of this `ts-expect-error`
We can do it here -> [link to the code](https://github.com/DevExpress/DevExtreme/blob/b73e7cb41ef823838adfa41e1857fe7545a58cb3/packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts#L48)

It may be sth like (because it's always not a `deferred` one type):
```ts
  private readonly selectionHelper: ReadonlySignal<Selection<any, any, false> | undefined>;
```

packages/devextreme/js/__internal/grids/tree_list/selection/m_selection.ts
```diff
@@ -543,6 +543,7 @@ const selection = (Base: ModuleType<SelectionController>) => class SelectionCont
     const selectedKeys = this.getSelectedRowKeys(mode) || [];
     const selectedRowsData: any[] = [];
 
+    // @ts-expect-error if _selection is deferred, then selectedKeys will be a Promise
```
wdevfx: Could I ask u a small favor if changes will be pushed in this PR? :)

Let's change this `ts-expect-error` a little bit to sth like this:
```ts
// @ts-expect-error selection may be deferred only in DataGrid,
//  we need to improve GridCore types to take it into account
```

Because in the `TreeList` (regarding to the public API) as far as I can see, selection cannot be in a `deferred` mode
And in the future it will be very helpful for Grids team to understand context of this `ts-expect-error` better :)

packages/devextreme/js/__internal/grids/grid_core/selection/m_selection.ts
juleari: Right now SelectionOptions, that are passed to the constructor are all Partial and the error is not about missing properies.
It is about `selectedKeys` property, which has type `unknown[] | undefined`, that's because type of TKey is `unknown`. and the expected type of TKey is `string | number`

Also, default props should not all be required for passing here. Because they will fallback to default ones. There may be some required props, that are not in default (this will be established in refactoring issue).

wdevfx (approved):

packages/devextreme/js/__internal/grids/grid_core/selection/m_selection.ts
wdevfx: Ok, let's leave it as is for now then :)

packages/devextreme/js/__internal/grids/new/grid_core/selection/controller.ts
```diff
@@ -365,6 +366,7 @@ export class SelectionController {
...
 
   public getSelectedCardsData(): DataObject[] {
+    // @ts-expect-error undefined is not assignable to DataObject[]
```
wdevfx, line 369: This method `getSelectedCardsData` used only as a `public API` method
In the `public API` of the `CardView` component we have return type `DataObject[]`
[Link to the one usage point in public API](https://github.com/DevExpress/DevExtreme/blob/b73e7cb41ef823838adfa41e1857fe7545a58cb3/packages/devextreme/js/__internal/grids/new/grid_core/selection/public_methods.ts#L19) | [Link to the public API d.ts](https://github.com/DevExpress/DevExtreme/blob/b73e7cb41ef823838adfa41e1857fe7545a58cb3/packages/devextreme/js/ui/card_view.d.ts#L1969)

We can change the public API type (`array | undefined`), but just formally it will be a very small `BC`
Could u please just normalize return type here the following way and remove the `ts-expect-error`? :)
```ts
return this.selectionHelper?.peek()?.getSelectedItems() ?? [];
```

If u don't want to add additional changes to this PR, feel free to resolve this comment -> we will do it from the `Grids` team side as a separate PR after ur changes will be merged :)

juleari: i can do it in a separate pr, but not in this one)
it will change current behavior of method in grids and this pr should not be the place for that)

anna-shakhova (approved):

PR #30771, opened 2025-08-15
Scheduler: fix all-day appointment rendered in the next resource

aleksei-semikozov:


packages/devextreme/js/__internal/scheduler/__tests__/views.test.ts
```diff
@@ -37,4 +37,46 @@ describe('views', () => {
...
+          { title: 'Luke', id: 3 },
+        ],
+        displayExpr: 'title',
+        fieldExpr: 'resourceId',
+        valueExpr: 'id',
+      }],
+    });
+
+    scheduler.option('width', 800);
+    scheduler.option('width', 730);
```
Ambrozy: I a bit confused about the test case.
In 25.1 and 24.2 we can see the wrong resources without resizing.
And also the issue is that when we click next button in day view there is no appointment tail there. But on week view we see 2 parts of appointment 2nd and 3rd.
So the test should do:

```
expect(appointments.length).toBe(1);
nextButton.click()
expect(appointments.length).toBe(1);
```
without scheduler resizing

If you would like to check that for small cells width it's okay, then do it in this way:
```
[
  { width: 200.5, title: 'wide cells' },
  { width: DEFAULT_CELL_WIDTH, title: 'small cells' },
].forEach(({}) => {
   it(`...`)
})

```

packages/devextreme/js/__internal/scheduler/__tests__/views.test.ts
```diff
@@ -37,4 +37,46 @@ describe('views', () => {
     const appointment = container.querySelector('.dx-scheduler-appointment');
     expect(appointment !== null).toBe(true);
   });
+
+  it('should not render all-day appointment in wrong resource (T1297021)', async () => {
+    setupSchedulerTestEnvironment({ width: 250.4 });
```
Ambrozy: Also it's not clear why do we need such cell width. Please, write it in the test description like.
`should not render all-day appointment in wrong resource for wide cells (T1297021)`

packages/devextreme/js/__internal/scheduler/__tests__/santiago_timezone.test.ts
```diff
@@ -103,7 +103,7 @@ const getTexts = (
 
 describe('scheduler', () => {
   it.each(views)('should render correct workspace in Santiago DST for view: $view.name', async ({ view, result }) => {
-    setupSchedulerTestEnvironment(true);
+    setupSchedulerTestEnvironment({ height: 450 });
```
Ambrozy: Can you please make a constant in `packages/devextreme/js/__internal/scheduler/__tests__/__mock__/m_mock_scheduler.ts`
`const DEFAULT_TIMELINE_CELL_HEIGHT = 450`

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/m_mock_scheduler.ts
```diff
@@ -3,25 +3,32 @@ import DOMComponent from '@ts/core/widget/dom_component';
...
+interface SetupSchedulerTestEnvironmentOptions {
+  width?: number;
+  height?: number;
+}
 
-  (DOMComponent.prototype as any)._isVisible = jest.fn().mockReturnValue(true);
-  SchedulerWorkSpace.prototype._createCrossScrollingConfig = () => ({
+export const setupSchedulerTestEnvironment = ({
+  width = 250,
+  height = 80,
```
Ambrozy:
```
const DEFAULT_CELL_WIDTH = 250
const DEFAULT_CELL_HEIGHT = 80
```

Ambrozy (changes requested):

packages/devextreme/js/__internal/scheduler/__tests__/views.test.ts
```diff
@@ -37,4 +37,39 @@ describe('views', () => {
...
+          { title: 'Luke', id: 2 },
+        ],
+        displayExpr: 'title',
+        fieldExpr: 'resourceId',
+        valueExpr: 'id',
+      }],
+    });
+
+    scheduler.option('width', 800);
+    scheduler.option('width', 730);
```
Ambrozy: lets remove width changes because it doesn't effect on the view in jest. We mock cell width at the start of the test

Ambrozy (approved):

PR #30778, opened 2025-08-15
DataGrid: Fix incorrect display of headers and filter row when there are fixed, banded, and grouped columns (T1276277)

Alyar666:


e2e/testcafe-devextreme/tests/dataGrid/sticky/common/withBandColumns.ts
```diff
@@ -0,0 +1,52 @@
...
+[false, true].forEach((rtlEnabled) => {
+  // T1279722
+  safeSizeTest(`Headers and filter row should display correctly after scrolling to the max right position when there is a grouped column (rtl=${rtlEnabled})`, async (t) => {
+    const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+    const dataGrid = new DataGrid(DATA_GRID_SELECTOR);
+
+    await t.expect(dataGrid.isReady()).ok();
+
+    await dataGrid.scrollTo(t, { x: rtlEnabled ? 0 : 10000 });
+    await takeScreenshot(`T1279722_band_sticky_columns-headers_with_filter_row_and_grouped_column_(rtl=${rtlEnabled}).png.png`, dataGrid.element);
```
wdevfx: Let's remove duplication of file extension in etalon name
```suggestion
    await takeScreenshot(`T1279722_band_sticky_columns-headers_with_filter_row_and_grouped_column_(rtl=${rtlEnabled}).png`, dataGrid.element);
```

packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
```diff
@@ -613,17 +613,26 @@ export class ColumnsController extends modules.Controller {
...
+        headerCellTemplate: null,
+        fixed: !isDefined(column.groupIndex) || !isFixedFirstGroupColumn ? isColumnFixing : true,
+        fixedPosition: rtlEnabled ? 'right' : 'left',
+      },
+      expandColumn,
+      {
+        index: column.index,
+        type: column.type || GROUP_COMMAND_COLUMN_NAME,
+      },
+    ));
```
wdevfx, line 635: Two questions / suggestions here

Can we get rid of `extend` or it's special logic required here?
I mean sth like this:
```suggestion
    expandColumn = {
      ...column,
      ownerBand: undefined,
      visibleWidth: null,
      minWidth: null,
      cellTemplate: !isDefined(column.groupIndex) ? column.cellTemplate : null,
      headerCellTemplate: null,
      fixed: !isDefined(column.groupIndex) || !isFixedFirstGroupColumn ? isColumnFixing : true,
      fixedPosition: rtlEnabled ? 'right' : 'left',
      ...expandColumn,
      index: column.index,
      type: column.type || GROUP_COMMAND_COLUMN_NAME,
    };
```

If we need extend, let's don't mix up spread and extend, but it's a minor
```suggestion
    expandColumns = map(expandColumns, (column) => extend(
      {},
      column,
      {
        ownerBand: undefined,
        visibleWidth: null,
        minWidth: null,
        cellTemplate: !isDefined(column.groupIndex) ? column.cellTemplate : null,
        headerCellTemplate: null,
        fixed: !isDefined(column.groupIndex) || !isFixedFirstGroupColumn ? isColumnFixing : true,
        fixedPosition: rtlEnabled ? 'right' : 'left',
      },
      expandColumn,
      {
        index: column.index,
        type: column.type || GROUP_COMMAND_COLUMN_NAME,
      },
    ));
```

Alyar666: Hi,

Initially, I did exactly as you suggested at the beginning. But then the test started failing. The failure was caused by the fact that some properties get overwritten if the last property value is `undefined`. In our implementation of extend, an `undefined` value is simply ignored. That’s why I decided to combine the two approaches.

I also can’t set `ownerBand = null`. In our code, there are too many checks for `ownerBand`, and everywhere we assume that ownerBand has some value or is undefined.

Alyar666: I can add a comment explaining this behavior. @wdevfx Do you think I should add it?

wdevfx: No, I think it's alright without a comment
Ok, now I got it, let's leave it as is and think about how to simplify in in the future :)

wdevfx (approved):

PR #30779, opened 2025-08-15
Toast: Improve typing

marker-dao:


packages/devextreme/js/__internal/ui/toast/m_hide_toasts.ts
```diff
@@ -1,25 +1,37 @@
 import $ from '@js/core/renderer';
+import type Toast from '@ts/ui/toast/m_toast';
+import { TOAST_CLASS } from '@ts/ui/toast/m_toast';
 
-const TOAST_CLASS = 'dx-toast';
-
-function hideAllToasts(container): void {
+function hideAllToasts(container: Element): void {
```
Zedwag:
```suggestion
function hideAllToasts(container?: dxElementWrapper | Element): void {
  const toasts = $(`.${TOAST_CLASS}`).toArray();
  if (!container) {
    toasts.forEach((toast) => {
      Toast.getInstance(toast)?.hide();
    });
    return;
  }
```

let's refactor it this way. also i see it will break one test, but it's ok to update the test (pass jq el instead of htmlElement), minor change in behavior (check on undefined instead of arguments length) is ok, since function arg is not public and we have never shown this possibility to users

packages/devextreme/js/__internal/ui/toast/m_hide_toasts.ts
```diff
@@ -1,25 +1,37 @@
...
   const toasts = $(`.${TOAST_CLASS}`).toArray();
+
   if (!arguments.length) {
-    // @ts-expect-error
-    toasts.forEach((toast) => { $(toast).dxToast('hide'); });
+    toasts.forEach((toast) => {
+      // @ts-expect-error does not exist on type 'dxElementWrapper'
+      $(toast).dxToast('hide');
```
Zedwag: same here and below

packages/devextreme/js/__internal/ui/toast/m_hide_toasts.ts
marker-dao: I'm sure it should work a little differently, give me a moment.

marker-dao: ### Bottom line
1. The parameters of the function are not described in [**public types**](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/ui/toast/hide_toasts.d.ts#L7), and the documentation [**does not mention**](https://js.devexpress.com/Vue/Documentation/ApiReference/Common/Utils/ui/Methods/#hideToasts) that this function can accept any parameters, so container should be optional (you have correctly indicated this).

2. The container must be of union type Element | dxElementWrapper (not HtmlElement), since jQuery and dxElementWrapper do not exist in React and other frameworks (same).

3. The `arguments` is used when it is important for us to check that the function was called without any parameters at all. This is because it is possible to call the function with a parameter that may have the value undefined (this is confirmed by tests). But I agree that this behavior should be reviewed. In any case, it is logical that if the container is an empty string, not passed, null/undefined, then we should hide everything.

**So** I left the arguments, added a container check, and changed the test to make it more explicit.

> Toast.getInstance(toast)?.hide();

4. I believe that using optional chaining here is inappropriate, as this behavior is unsafe.

Let me explain: in this section of code, we need to be sure that the toasts variable contains all the elements from which a Toast instance can be obtained.

If you use optional chaining here, you will not notice a potential error. After all, if the getInstance method does not return an instance where it should, then it means that it is not working correctly and we need to fix it. If there is no instance here, we will notice it (because we will get a JS error) and fix it.

> static getInstance<T = any>(element: Element | dxElementWrapper): T {

5. Let's ask @EugeniyKiyashko to take a look [**this change**](https://github.com/DevExpress/DevExtreme/pull/30779/commits/732aca344ff80d38b0675f658133c3d5652d32f3#diff-d1c95929184a964f10dc6f0e71558cdda917da73a397d5cf4d63dad6a20917c3R46). We are temporarily using any here (it would be more correct to use unknown, but this causes some TS errors that can be fixed in the future). I believe that Generic should be used here, since TS is unable to calculate the type of the returned object here and it is really unknown.

r-farkhutdinov (approved):

anna-shakhova (approved):

PR #30784, opened 2025-08-17
Map: improve types

Zedwag:


packages/devextreme/js/__internal/ui/map/m_map.ts
```diff
@@ -190,17 +202,22 @@ class Map extends Widget<MapProperties> {
...
-    // @ts-expect-error ts-error
-    this._rendered[option] = value.slice();
+  _saveRendered(option: 'markers' | 'routes'): void {
+    if (option === 'markers') {
+      const { markers = [] } = this.option();
+      this._rendered.markers = [...markers];
+    } else if (option === 'routes') {
+      const { routes = [] } = this.option();
+      this._rendered.routes = [...routes];
+    }
```
ksercs: Does really TypeScript force us to introduce this duplication? 🤔

packages/devextreme/js/__internal/ui/map/m_map.ts
```diff
@@ -294,19 +324,22 @@ class Map extends Widget<MapProperties> {
 
         const prevValue = this._rendered[name];
         this._saveRendered(name);
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         this._queueAsyncAction(
-          `update${titleize(name)}`,
-          changeBag ? changeBag.removed : prevValue,
+          `${name === 'markers' ? 'updateMarkers' : 'updateRoutes'}`,
+          // @ts-expect-error ts-error
+          changeBag?.removed ? changeBag.removed : prevValue,
```
ksercs: Hm, is logic really not changed?

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.bing.ts
```diff
@@ -216,8 +235,8 @@ class BingProvider extends DynamicProvider {
     return Promise.resolve();
   }
 
-  updateMapType() {
-    const type = this._option('type');
+  updateMapType(): Promise<void> {
+    const type = this._option('type') ?? 'roadmap';
```
ksercs: I see this guard in a bunch of places - don't u think it's a good idea to create a getter somewhere in a base provider?

packages/devextreme/js/__internal/ui/map/m_provider.google_static.ts
```diff
@@ -2,17 +2,17 @@
...
 import Provider from './m_provider';
+import type { LocationOption, MarkerOptions, RouteOptions } from './m_provider.dynamic';
 
 let GOOGLE_STATIC_URL = 'https://maps.google.com/maps/api/staticmap?';
 
 class GoogleStaticProvider extends Provider {
-  _locationToString(location) {
+  _locationToString(location: LocationOption): string {
     const latLng = this._getLatLng(location);
-    return latLng ? `${latLng.lat},${latLng.lng}` : location.toString().replace(/ /g, '+');
```
ksercs, line 15: hm, is omit of `.toString` indended here?

packages/devextreme/js/__internal/ui/map/m_map.ts
```diff
@@ -334,12 +366,16 @@ class Map extends Widget<MapProperties> {
...
     this._queueAsyncAction('updateDimensions');
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  _queueAsyncAction(name, markers?, routers?) {
-    const options = [].slice.call(arguments).slice(1);
+  _queueAsyncAction(
+    name: string,
+    markers?: MapProperties['markers'],
+    routers?: MapProperties['routes'],
```
r-farkhutdinov: Shouldn't it be `routes`?

packages/devextreme/js/__internal/ui/map/m_map.ts
```diff
@@ -334,12 +366,16 @@ class Map extends Widget<MapProperties> {
...
+  ): Promise<void> {
+    const markerAndRoutes = [markers, routers].filter((option) => option !== undefined);
```
r-farkhutdinov: Do we want to check exactly non-undefined values or just truthy is enough?
```
[markers, routers].filter(Boolean);
```

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
```diff
@@ -1,50 +1,57 @@
...
       satellite: 'satellite',
       hybrid: 'satellite_road_labels',
     };
-    return mapTypes[type] || mapTypes.roadmap;
+    return mapTypes[type] ?? mapTypes.roadmap;
   }
 
-  _movementMode(type) {
-    const movementTypes = {
+  _movementMode(type: RouteMode | string = ''): string {
```
r-farkhutdinov, line 57: Default value should be: 
```suggestion
  _movementMode(type: RouteMode | string = movementTypes.driving): string {
```

And early return above can be removed

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
```diff
@@ -56,20 +63,22 @@ class AzureProvider extends DynamicProvider {
     return movementTypes[type] ?? type;
   }
 
-  _resolveLocation(location) {
+  _resolveLocation(location: LocationOption | null | undefined): Promise<AzureLocation> {
```
r-farkhutdinov:
```suggestion
  _resolveLocation(location?: LocationOption | null): Promise<AzureLocation> {
```

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
```diff
@@ -56,20 +63,22 @@ class AzureProvider extends DynamicProvider {
...
       const latLng = this._getLatLng(location);
       if (latLng) {
         resolve(new atlas.data.Position(latLng.lng, latLng.lat));
       } else {
-        this._geocodeLocation(location).then((geocodedLocation) => {
-          resolve(geocodedLocation);
-        });
+        this._geocodeLocation(location as string).then((geocodedLocation) => {
+          resolve(geocodedLocation as AzureLocation);
+        }).catch(() => {});
```
r-farkhutdinov: Empty catch, is it added to resolve a linter warning? If so, maybe we can just ignore it?

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
```diff
@@ -129,20 +143,22 @@ class AzureProvider extends DynamicProvider {
...
     });
   }
 
-  _loadMapResources() {
+  _loadMapResources(): Promise<void> {
     return Promise.all([
       this._loadMapScript(),
       this._loadMapStyles(),
-    ]);
+    ]).then(() => {});
```
r-farkhutdinov, line 163: Same here, I think it's better to just ignore the linter warning instead of adding empty then or catch

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
```diff
@@ -300,9 +323,10 @@ class AzureProvider extends DynamicProvider {
     return Promise.resolve();
   }
 
-  _renderMarker(options) {
-    return this._resolveLocation(options.location).then((location) => {
-      const markerOptions: any = {
+  _renderMarker(options: MarkerOptions): Promise<MarkerObject> {
+    const markerLocation = options.location;
```
r-farkhutdinov:
```suggestion
    const { location: markerLocation } = options;
```

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
```diff
@@ -342,14 +367,15 @@ class AzureProvider extends DynamicProvider {
     });
   }
 
-  _renderTooltip(location, options) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  _renderTooltip(location: AzureLocation, options: MarkerOptions['tooltip']): any {
     if (!options) {
-      return;
+      return undefined;
```
r-farkhutdinov, line 379: Why undefined should be specified explicitly here?

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
```diff
@@ -358,14 +384,14 @@ class AzureProvider extends DynamicProvider {
...
-    if (options.visible) {
+    if (parsedOptions.visible) {
       popup.open();
     }
 
     return popup;
   }
 
-  _destroyMarker(marker) {
+  _destroyMarker(marker: { marker: unknown; popup: unknown; handler: () => {} }): void {
```
r-farkhutdinov:
```suggestion
  _destroyMarker(marker: { marker: unknown; popup: unknown; handler: () => void }): void {
```

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.bing.ts
```diff
@@ -1,20 +1,26 @@
+/* eslint-disable @typescript-eslint/prefer-nullish-coalescing */
```
r-farkhutdinov: Let's add this "on demand" or are there too many places?

packages/devextreme/js/__internal/ui/map/m_map.ts
Zedwag: hm, don't remember y i decided to do it, but let's revert

packages/devextreme/js/__internal/ui/map/m_map.ts
Zedwag: yep. but now i've decided to better extend  this._rendered types

packages/devextreme/js/__internal/ui/map/m_map.ts
Zedwag: lol, didn't even noticed that, will update

packages/devextreme/js/__internal/ui/map/m_map.ts
Zedwag: yep, truthy should be enough, ty

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
Zedwag: no, i'd prefer keep it this way since even though I used RouteMode | string type here, it still possible someone pass smth like null which won't be handled in args default value

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
Zedwag: updated

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
Zedwag: yep it's for lenter. tbh i like this way more than just ignore lint warnings

Zedwag: was thinking a little more and decided to replace with ingnore comments

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
Zedwag: nah, it's not the same case here, this one is needed to to make sure we return promise void after promise.all call

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
Zedwag: because otherwise there will be lint warning, for example
```
_someFunction(flag?: boolean): unknown {
    if (!flag) {
      return;
    }

    return true;
  }
```
  
  will log `Method '_someFunction' expected no return value.eslint[consistent-return](https://eslint.org/docs/latest/rules/consistent-return)`

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.bing.ts
Zedwag: done

packages/devextreme/js/__internal/ui/map/m_provider.google_static.ts
Zedwag: well i didn't find a scenario when it's not string but let revert just in case

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.bing.ts
Zedwag: ok, I'll just handle undefined inside _mapType() to get rid of this

packages/devextreme/js/__internal/ui/map/m_provider.dynamic.azure.ts
Zedwag: done

r-farkhutdinov (approved):

PR #30811, opened 2025-08-20
AIIntegration: add SmartPaste command

pharret31:


packages/devextreme/js/__internal/core/ai_integration/commands/smartPaste.ts
```diff
@@ -0,0 +1,54 @@
...
+  protected parseResult(response: string): SmartPasteCommandResult {
+    const result: SmartPasteCommandResult = [];
+
+    response.split(';;;').forEach((data: string) => {
+      const [name, ...values] = data.split(':::');
+      const value = values.length === 1 ? values[0] : values;
+
+      if (value) {
+        try {
+          result.push({
```
anna-shakhova: can't imagine why we expect error on push
looks like, this try/catch should be moved in form, where we will use this result to update form values

packages/devextreme/js/common/ai-integration.d.ts
```diff
@@ -106,6 +106,23 @@ export type TranslateCommandParams = {
   lang: string;
 };
 
+/**
+ * @hidden
+ */
+export type FieldInfo = {
```
anna-shakhova, line 112: specify @namespace for type, please, no it is in common instead of aiIntegration

packages/devextreme/js/__internal/core/ai_integration/commands/smartPaste.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  describe('buildPromptData', () => {
+    it('should form PromptData with text and fields info', () => {
+      // @ts-expect-error Access to protected property for a test
+      const promptData: PromptData = command.buildPromptData(params);
+      // @ts-expect-error Access to private property for a test
+      const fieldsInfo = command.generateFieldsInstructions(params.fields);
+
+      expect(promptData).toStrictEqual({
+        user: { text: USER_TEXT, fields: fieldsInfo },
```
anna-shakhova: this test will pass always, even we change `generateFieldsInstructions` implementation, because we test, that prompt data have the result from `generateFieldsInstructions` but we do not check this result is what is expected

anna-shakhova (approved):

PR #30830, opened 2025-08-20
DataGrid: Fix the focus() method that focuses the interactive element inside a cell instead of the row when focusedRowEnabled is true (T1293309)

Alyar666:
Also fixed the focus() method for TreeList ([T1294363](https://isc.devexpress.com/internal/ticket/details/T1294363))

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
@@ -88,6 +88,7 @@ import {
   isDetailRow,
   isEditForm,
   isEditorCell,
+  isEditRow,
   isElementDefined,
   isGroupFooterRow,
   isGroupRow,
@@ -1533,6 +1534,7 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
         isHighlighted,
       );
       $element = args.$newCellElement;
+
       if (isRowFocusType && !args.isHighlighted) {
         this.setRowFocusType();
       }
@@ -1540,7 +1542,10 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
 
     if (!args.cancel) {
       this._focus($element, !args.isHighlighted);
-      this._focusInteractiveElement($element);
+
+      if (this._getElementType($element) !== 'row' || isEditRow($element)) {
+        this._focusInteractiveElement($element);
+      }
     }
   }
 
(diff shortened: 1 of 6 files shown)


packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1483,10 +1483,10 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
     const isHighlighted = this._isCellElement($(element));
 
     if (!element) {
-      activeElementSelector = '.dx-datagrid-rowsview .dx-row[tabindex]';
+      activeElementSelector = `.${this.addWidgetPrefix(ROWS_VIEW_CLASS)} .dx-row[tabindex]`;
```
Raushen, line 1486: Minor: I'd extract this variable as rowsViewSelector...

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1533,14 +1534,18 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
         isHighlighted,
       );
       $element = args.$newCellElement;
+
```
Raushen, line 1537: Minor: extra space/change

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1533,14 +1534,18 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
...
       if (isRowFocusType && !args.isHighlighted) {
         this.setRowFocusType();
       }
     }
 
     if (!args.cancel) {
       this._focus($element, !args.isHighlighted);
-      this._focusInteractiveElement($element);
+
+      if (this._getElementType($element) !== 'row' || isEditRow($element)) {
```
Raushen, line 1546: Minor: I prefer to extract variables from conditions so it would be easier to debug later.

Raushen (approved):

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
Alyar666: I agree, it could have been extracted into a separate variable.

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
Alyar666: No, I added it on purpose. Usually, we put an indent before a conditional block.

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
Alyar666: I think it’s unnecessary in this case:

- the method is called only once
- the method is simple
- the condition is easy to read

In other cases, I would agree with you.

PR #30846, opened 2025-08-21
DropDownBox, SelectBox, ColorBox, TagBox: Log fieldTemplate deprecation warning

r-farkhutdinov:

### DIFF
--- a/apps/demos/testing/known-warnings.js
+++ b/apps/demos/testing/known-warnings.js
@@ -1,5 +1,13 @@
+export const dropDownEditorsFieldTemplateWarnings = [
+  'W0001 - dxSelectBox - \'fieldTemplate\'',
+  'W0001 - dxColorBox - \'fieldTemplate\'',
+  'W0001 - dxDropDownBox - \'fieldTemplate\'',
+  'W0001 - dxTagBox - \'fieldTemplate\'',
+];
+
 export const knownWarnings = [
   'W0019 -',
   'W0022 -',
   'W2108 -',
+  ...dropDownEditorsFieldTemplateWarnings,
 ];
--- a/packages/devextreme/js/__internal/ui/drop_down_editor/m_drop_down_editor.ts
+++ b/packages/devextreme/js/__internal/ui/drop_down_editor/m_drop_down_editor.ts
@@ -56,6 +56,13 @@ const isIOs = devices.current().platform === 'ios';
 
 type HideOnOutsideClickEvent = DxEvent<MouseEvent | PointerEvent | TouchEvent>;
 
+export const DROP_DOWN_EDITOR_DEPRECATED_OPTIONS = {
+  fieldTemplate: {
+    since: '25.2',
+    message: 'Use the \'fieldAddons\' option instead',
+  },
+};
+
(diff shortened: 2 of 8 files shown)


packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/dropDownEditors.deprecation.tests.js
```diff
@@ -0,0 +1,59 @@
...
+            QUnit.test(
+                `${name}: fieldTemplate should raise a deprecation warning if set ${isSetDynamically ? 'dynamically' : 'on initialization'}`,
+                function(assert) {
+                    sinon.spy(errors, 'log');
+
+                    const fieldTemplate = () => $('<div>').dxTextBox();
+                    const options = isSetDynamically ? undefined : { fieldTemplate };
+                    const instance = $(selector)[name](options)[name]('instance');
+
+                    if(isSetDynamically) {
```
ksercs: Looks strange for me to change part of tests depending on iterated variable - looks like it'd be clearer to just write 2 tests

apps/demos/testing/known-warnings.js
```diff
@@ -1,5 +1,14 @@
...
+  'W0001 - dxDropDownBox - \'fieldTemplate\'',
+  'W0001 - dxTagBox - \'fieldTemplate\'',
+];
+
 export const knownWarnings = [
   'W0019 -',
   'W0022 -',
   'W2108 -',
+  'W0001 -',
+  ...dropDownEditorsFieldTemplateWarnings,
```
ksercs, line 12: Nope, we should update demos to not use deprecated API instead of test skip)

packages/devextreme/js/__internal/ui/drop_down_editor/m_drop_down_editor.ts
```diff
@@ -1143,6 +1165,9 @@ class DropDownEditor<
         break;
       case 'fieldAddons':
```
ksercs: [mistake]
Now the warning will be logged even after `fieldAddons` runtime option change
p.s. maybe we should also cover it with tests if we decide to implement this custom logging?

Zedwag (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/dropDownEditors.deprecation.tests.js
```diff
@@ -0,0 +1,70 @@
...
+
+    $('#qunit-fixture').html(markup);
+});
+
+QUnit.module('Deprecated fieldTemplate', {
+    beforeEach() {
+        this.errorsSpy = sinon.spy(errors, 'log');
+
+        this.assertWarningIsCorrect = (assert, name) => {
+            assert.deepEqual(
```
ksercs, line 45: I noticed during manual testing that excess warnings are logged now. For example, 5 warning on init.

Please update the assert to additionally check that only one warning is logged

Also please consider if it's helpful to test that warning is not logged on value change (when we get the value of a field template)

ksercs (approved):

PR #30848, opened 2025-08-21
Form: create .d.ts properties for the Smart Paste

pharret31:


packages/devextreme/js/ui/form.d.ts
```diff
@@ -152,6 +185,12 @@ export type SimpleItemLabelTemplateData = SimpleItemTemplateData & { text: strin
  * @docid
  */
 export interface dxFormOptions extends WidgetOptions<dxForm> {
+    /**
+     * @docid
+     * @default null
+     * @public
+     */
+    aiIntegration?: AIIntegration;
```
anna-shakhova: in htmlEditor we use AIIntegration | undefined with undefined as default, let's make here the same way

packages/devextreme/js/ui/form.d.ts
```diff
@@ -365,6 +420,13 @@ export default class dxForm extends Widget<dxFormOptions> {
      * @public
      */
     reset(editorsData?: Record<string, any>): void;
+    /**
+     * @docid
+     * @publicName smartPaste(text)
+     * @param1 text:object
```
anna-shakhova:
```suggestion
     * @param1 text:string|undefined
```

packages/devextreme/js/ui/form.d.ts
```diff
@@ -61,6 +63,23 @@ export type FormItemType = 'empty' | 'group' | 'simple' | 'tabbed' | 'button';
 export type LabelLocation = 'left' | 'right' | 'top';
 /** @public */
 export type FormLabelMode = 'static' | 'floating' | 'hidden' | 'outside';
+/** @public */
+export type FormPredefinedButtonItem = 'reset' | 'submit' | 'smartPaste';
+
+/** @public */
+export type AIResult = Record<string, string | string[]>;
```
ksercs, line 70: This looks strange to me - I'm just worried there will be conflicts after we add more AI features - this "ai result" can be just a string or maybe even another type.

The type looks very general - shouldn't we reuse some type from aiIntegration core types? 🤔

PR #30851, opened 2025-08-22
DataGrid(T1267471): ColumnChooser doesn't immediately reflect the changes made to column options

wdevfx:


e2e/testcafe-devextreme/tests/dataGrid/common/columnChooser.ts
```diff
@@ -192,3 +192,44 @@ test('Check the behavior of pressing the Esc button when dragging a column from
     mode: 'dragAndDrop',
   },
 }));
+
+test(
+  'Should take into account column options change during general option change [T1267471]',
```
pomahtri: very minor
```suggestion
  'Should take into account column options change during general option change (T1267471)',
```

wdevfx: Yup, ur brackets more consistent with other test case names -> agree
FIxed

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -376,25 +376,45 @@ export class ColumnChooserView extends ColumnsView {
...
 
-      if (needUpdate) {
-        this._updateItemsSelection(columnIndices);
+    const columnIndices = isDefined(changes.columnIndex)
+      ? [changes.columnIndex]
+      : changes.columnIndices;
+    const needUpdate: boolean = COLUMN_OPTIONS_USED_IN_ITEMS
+      .some((
+        optionName,
+      ) => optionNames[optionName]) || (changes.changeTypes.columns && optionNames.all);
```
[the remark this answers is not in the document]
wdevfx: It's not correct, if I apply this advice -> `ESLint` warning about line length will appear here :)

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -376,25 +376,45 @@ export class ColumnChooserView extends ColumnsView {
...
+      this._updateItems();
+    }
+  }
+
+  private isColumnVisibilityOnlyUpdated(
+    optionNames: { length: number } & Record<string, unknown>,
+  ): boolean {
+    const optionKeys = Object
+      .keys(optionNames ?? {})
+      .filter((key) => key !== 'length');
```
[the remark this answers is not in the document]
wdevfx: About readability -> it's a matter of taste
I prefer and vote to leave it as is because different lines read easier that one big line from my perspective

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -376,25 +376,45 @@ export class ColumnChooserView extends ColumnsView {
...
+    if (!needUpdate) {
+      return;
     }
+
+    this._updateItemsSelection(columnIndices);
+    if (!onlyVisibleChanged) {
+      this._updateItems();
+    }
+  }
+
```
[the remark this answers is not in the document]
wdevfx: I guess this comment duplicates the code and describe it's simple logic
So, I don't think that adding obvious comment like this may be useful :)

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -376,25 +376,45 @@ export class ColumnChooserView extends ColumnsView {
...
+  private isColumnVisibilityOnlyUpdated(
+    optionNames: { length: number } & Record<string, unknown>,
```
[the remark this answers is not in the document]
wdevfx: I prefer to leave this type inline, because unfortunately we don't have a good type for columnOption changes
In future, it should be replaced by this general `changes` type (like in `_optionChanged`)

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -376,25 +376,45 @@ export class ColumnChooserView extends ColumnsView {
...
+  ): boolean {
+    const optionKeys = Object
+      .keys(optionNames ?? {})
+      .filter((key) => key !== 'length');
```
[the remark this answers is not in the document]
wdevfx: From my perspective, it's a `too much` optimization for such a simple case
Have some `object` or `Map` for caching here -> overkill

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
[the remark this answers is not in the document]
wdevfx: Already answered above

In future, we'd like to change this inline internal type for one common type of `_columnOptionChange` arguments
While we don't have it, inline type looks ok here :)

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -376,25 +376,45 @@ export class ColumnChooserView extends ColumnsView {
...
 
-  protected _columnOptionChanged(e) {
-    super._columnOptionChanged(e);
+  protected _columnOptionChanged(changes): void {
+    super._columnOptionChanged(changes);
 
+    const { optionNames } = changes;
     const isSelectMode = this.isSelectMode();
+    const onlyVisibleChanged = this.isColumnVisibilityOnlyUpdated(optionNames);
+    const isOnlyColumnVisibilityUpdated = this._isUpdatingColumnVisibility
```
Tucchhaa, line 385: this._isUpdatingColumnVisibility flag appears useless to me now.
In the place where [it's set](https://github.com/DevExpress/DevExtreme/blob/951dc01f6835619559c9d94b0fe7d5d60bb178ca/packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts#L342) updateColumnVisibility() function is called, which only updates 'visible' option of the column.

onlyVisibleChanged and needUpdate checks should be enough here.

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -376,25 +376,45 @@ export class ColumnChooserView extends ColumnsView {
...
-        }
-      }
+    if (!needUpdate) {
+      return;
     }
+
+    this._updateItemsSelection(columnIndices);
+    if (!onlyVisibleChanged) {
+      this._updateItems();
+    }
```
Tucchhaa, line 407: If we don't need this._isUpdatingColumnVisibility anymore, then let's remove this if statement:
```suggestion
  this._updateItems();
  ```

File: e2e/testcafe-devextreme/tests/dataGrid/common/columnChooser.ts [hunk 9]
```diff
@@ -192,3 +192,44 @@ test('Check the behavior of pressing the Esc button when dragging a column from
...
+    const columnChooserBtn = dataGrid.getColumnChooserButton();
+
+    await t.click(columnChooserBtn);
+
+    const columnChooser = dataGrid.getColumnChooser();
+    const lastItemCheckbox = columnChooser.getCheckbox(1);
+
+    await t.click(lastItemCheckbox);
+
+    await t.expect(columnChooser.isCheckboxDisabled(0)).ok();
```
Tucchhaa, line 212: I suggest to check that all columns have expected visibility just in case. 

Minor. Also I would suggest to take screenshot of columnChooser (or get treeview's item through api), because DataGrid's columns state and treeView's items state could desync.

pomahtri (dismissed):

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
Tucchhaa: Discussed in DM. Actually we still need this flag, since user can use .columnOption(0, 'visible', true/false) and then we still need to call _updateItemsSelection

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
Tucchhaa: Not relevant, see previous comment

e2e/testcafe-devextreme/tests/dataGrid/common/columnChooser.ts
Tucchhaa: Discussed in DM. Decided that we shouldn't check columns visibility, because the test would be too vague. And there already other tests to check column visibility. However decided to check other checkbox's disabled state too.

wdevfx: Yup, other checkboxes checks added

Tucchhaa (approved):

PR #30853, opened 2025-08-22
Scheduler: Add Eslint rules according to codestyle

sjbur:
Added following rules to scheduler eslint config:
- Use Boolean instead of !!
- File names in snake case
- Use Promise instead of Deferred
- Don't use underscore for variable naming (_cache → cache)
- Use this construction
```
switch(true) {
  case fCondition:
    return foo;
  case sCondition:
    return bar;
  default:
    return null;
}
```
- [no-extra-boolean-cast](https://eslint.org/docs/latest/rules/no-extra-boolean-cast)


packages/devextreme/eslint.config.mjs
```diff
@@ -524,6 +526,53 @@ export default [
...
+            sourceType: 'script',
+            parserOptions: {
+                project: './tsconfig.json',
+                tsconfigRootDir: `${__dirname}/js/__internal`,
+            },
+        },
+        rules: {
+            'no-implicit-coercion': ['error', {
+                boolean: true,
+            }],
```
Tucchhaa: Minor suggestion: let's remove `{ boolean: true }` since it's enabled in [the rule by default](https://eslint.org/docs/latest/rules/no-implicit-coercion).

P.S. The rule also disallows by default implicit conversion for numbers and strings. I think that's ok

packages/devextreme/eslint.config.mjs
```diff
@@ -524,6 +526,53 @@ export default [
...
+            'no-extra-boolean-cast': 'error',
+            'unicorn/filename-case': ['error', { case: 'snakeCase' }],
+            '@typescript-eslint/naming-convention': [
+                'error',
+                {
+                    selector: ['variable', 'function', 'parameter'],
+                    format: null,
```
Tucchhaa, line 549: I suggest to set: `format: 'camelCase'` and check if there are not too many linter errors. If not, then I think it's a good idea to discuss it with the squad and set it.

But if there are a lot of errors, then let's postpone it

packages/devextreme/eslint.config.mjs
```diff
@@ -524,6 +526,53 @@ export default [
...
+            'devextreme-custom/prefer-switch-true': ['error', { minBranches: 3 }],
+        },
+    },
+    // Allow Deferred in m_* scheduler files only
+    {
+        files: ['js/__internal/scheduler/**/m_*.ts?(x)'],
+        rules: {
+            'devextreme-custom/no-deferred': 'off',
+        },
+    },
```
Tucchhaa, line 573: I have a question. I noticed that Deferred is used only in m_*.ts files, so do we plan to remove this rule in the future?

sjbur: I was told that in newer files we use only Promises and that Deferred is deprecated

packages/devextreme/js/__internal/scheduler/r1/utils/base.ts
```diff
@@ -331,7 +331,7 @@ export const getKeyByGroup = (
   groupIndex: number | undefined,
   isVerticalGrouping: boolean,
 ): string => {
-  if (isVerticalGrouping && !!groupIndex) {
+  if (isVerticalGrouping && groupIndex !== undefined) {
```
aleksei-semikozov, line 334: This change affects behavior for value = 0
```
// Example:
//   groupIndex = 0
!!0               // false
0 !== undefined   // true
```
Are you sure this behavior change is intentional?

aleksei-semikozov (changes requested):

packages/devextreme/eslint.config.mjs
sjbur: I am afraid that this change adds a lot of errors. One of examples is a conflict with rule "A constructor name should not start with a lowercase letter".

packages/devextreme/eslint.config.mjs
sjbur: Made this change

packages/devextreme/js/__internal/scheduler/r1/utils/base.ts
sjbur: I chose groupIndex !== undefined instead of !!groupIndex because it makes the condition explicit.
When groupIndex === 0, both versions still return '0', but they do it through different branches:
	•	with !!groupIndex, the if condition fails (since 0 is falsy) and '0' is returned from the else,
	•	with !== undefined, the condition passes (since 0 !== undefined) and '0' is returned inside the if.

So while the current result is the same, using !== undefined makes it clear that 0 is treated as a valid index rather than as an “absent” value.

Tucchhaa (approved):

aleksei-semikozov (approved):

PR #30863, opened 2025-08-25
Demos - Migrate ESLint from v8 to v9

Tucchhaa:


apps/demos/eslint.config.mjs
```diff
@@ -0,0 +1,491 @@
...
+      'bundles',
+      'coverage',
+      'publish-demos',
+      'Demos/**/testcafe-test-code.js',
+      'shared/empty-file.ts',
+      'Demos/**/config.js',
+      'Demos/**/Vue/**/*.html',
+      'utils', // TODO unignore this
+
+      '**/*.{png,json,mjs,css,html,md}',
```
Alyar666, line 46: Is this really necessary? Does Eslint check these files? (except mjs)

apps/demos/tsconfig.eslint.json
```diff
@@ -0,0 +1,21 @@
...
+    "module": "ESNext",
+    "moduleResolution": "Node",
+    "allowImportingTsExtensions": true,
+    "jsx": "react-jsx",
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    "lib": ["es2022", "DOM", "DOM.Iterable"]
+  },
+  "include": [
+    "./Demos/**/React/**/*.ts",
```
Alyar666, line 16: Is it possible to simplify this expression?
For example, like this - `"./Demos/**/{React,Vue,Angular}/**/*.{ts,tsx}"` ?
Or even simpler - `"./Demos/**/*.{ts,tsx}"` ?

apps/demos/eslint.config.mjs
Tucchhaa: Yes, it's actually necessary. Eslint recursively check all files in the root directory and if some file matches the pattern in any block that have some rule, eslint will apply these rules.

In the current config we have a[ block](https://github.com/DevExpress/DevExtreme/pull/30863/files#diff-e4fa5497e7595e7fd43766b86f8c96e9bc2fff148c304cc324dcf7cddd0cccc2R90) with general rules that are applied to all files. Also eslint:recommended and devextreme/spell-check [are applied](https://github.com/DevExpress/DevExtreme/pull/30863/files#diff-e4fa5497e7595e7fd43766b86f8c96e9bc2fff148c304cc324dcf7cddd0cccc2R77) to all files.

apps/demos/tsconfig.eslint.json
Tucchhaa: I have tried to do that, but it didn't work

Alyar666 (dismissed):

Ambrozy (dismissed):

marker-dao (dismissed):

apps/demos/Demos/CardView/CardTemplate/React/App.tsx
```diff
@@ -24,7 +24,7 @@ const App = () => {
   }, []);
 
   return (
-    <>
+    <React.Fragment>
```
pharret31, line 27: Usually we use the full syntax if we need to pass some arguments like <React.Fragment key={}> but in general <> is enough

apps/demos/Demos/CardView/CardTemplate/Vue/App.vue
```diff
@@ -6,9 +6,9 @@
     :card-min-width="240"
     card-template="cardTemplate"
   >
-    <DxHeaderFilter :visible="true" />
-    <DxSearchPanel :visible="true" />
-    <DxPaging :pageSize="12" />
+    <DxHeaderFilter :visible="true"/>
+    <DxSearchPanel :visible="true"/>
+    <DxPaging :pageSize="12"/>
```
pharret31, line 11: There should actually be a space before the slash?

Tucchhaa: Maybe, but I can find more than thousand of such slashes in our demos by this regexp: `[^ ]/>`, so I think we should not to modify this in terms of this migration PR. We can consider to add such rule in the future

apps/demos/Demos/CardView/CardTemplate/React/App.tsx
Tucchhaa: In the old eslint config there was [this rule](https://github.com/DevExpress/DevExtreme/pull/30863/files#diff-3e720c91070111f0c1bfd7475e9170d36502d464de0b681a5599733bb12650e1L477) to enforce usage of `<React.Fragment>`

I decided to keep this rule since React.Fragment was used almost everywhere, but `<>` was used only in several demos. You can see which demos had `<>` changed to `React.Fragment` in [this commit](https://github.com/DevExpress/DevExtreme/pull/30863/commits/e1f4d148a2345f19a971c5a61917facb5ce2a384)

Ambrozy (approved):

apps/demos/Demos/CardView/ColumnChooser/Vue/App.vue
```diff
@@ -101,7 +101,8 @@
 <script setup lang="ts">
 import { ref } from 'vue';
 import {
-  DxCardView, DxColumn, DxCardCover, DxSearchPanel, DxColumnChooser, DxColumnChooserSearch, DxColumnChooserSelection,
+  DxCardView, DxColumn, DxCardCover, DxSearchPanel, DxColumnChooser,
+  DxColumnChooserSearch, DxColumnChooserSelection,
```
pharret31, line 105: Should they be placed in a column?

apps/demos/Demos/CardView/DataValidation/React/App.tsx
```diff
@@ -146,13 +146,13 @@ const App = () => (
       dataField="birthDate"
       dataType="date"
     >
-      <RequiredRule/>
+      <RequiredRule />
```
pharret31, line 149: And here spaces were added 🤔

Alyar666 (approved):

apps/demos/Demos/CardView/DataValidation/React/App.tsx
Tucchhaa: Because it's a React file and the old config had [react specific rule](https://github.com/DevExpress/DevExtreme/pull/30863/files/56caddbed42b1bc217723301093200e77b689796#diff-3e720c91070111f0c1bfd7475e9170d36502d464de0b681a5599733bb12650e1L486) for tag spacing, but config for vue files didn't have a rule for that.

apps/demos/Demos/DropDownBox/MultipleSelection/React/App.tsx
```diff
@@ -48,7 +86,7 @@ function App() {
         onItemSelectionChanged={treeViewItemSelectionChanged}
       />
     ),
-    [treeDataSource],
+    [syncTreeViewSelection, treeViewItemSelectionChanged],
```
ksercs, line 89: Is this intended?

Tucchhaa: Yes, it was intended, because there was a warning from 'react-hooks/exhaustive-deps' [rule](https://github.com/DevExpress/DevExtreme/pull/30863/files/56caddbed42b1bc217723301093200e77b689796#diff-e4fa5497e7595e7fd43766b86f8c96e9bc2fff148c304cc324dcf7cddd0cccc2R358).

apps/demos/Demos/CardView/ColumnChooser/Vue/App.vue
Tucchhaa: Maybe, but in the scope of migration I only fixed or disabled eslint rules, with only minor additions to the rules. There was no rule for that in the old config, I didn't consider adding new rules and I suggest to do that as a separate task.

I splitted this line because of the 'vue/max-len' rule in [this commit](https://github.com/DevExpress/DevExtreme/pull/30863/commits/3b5b85fe6ef04c8861d7aa932f1bb358c38c7d65)

apps/demos/Demos/Stepper/FormIntegration/ReactJs/App.js
```diff
@@ -95,31 +96,46 @@ export default function App() {
...
+      />
+    ),
+    [formData],
+  );
+  const renderAdditionalForm = useCallback(
+    () => <AdditionalForm formData={formData} />,
+    [formData],
+  );
+  const renderConfirmation = useCallback(
+    () => (
```
ksercs, line 131: Is this really more expected formatting for React?
I see it's better for `if` conditions but this particular code looks to be less clear now

ksercs (approved):
LGTM for updates in Components demos

marker-dao (approved):

pharret31 (approved):

apps/demos/Demos/Stepper/FormIntegration/ReactJs/App.js
Tucchhaa: Since it's a ReactJs demos, this file is generated automatically by ts to js converter. We need stricter eslint rules [here](https://github.com/DevExpress/DevExtreme/pull/30863/files/56caddbed42b1bc217723301093200e77b689796#diff-e4fa5497e7595e7fd43766b86f8c96e9bc2fff148c304cc324dcf7cddd0cccc2R445-R457) for reactjs demos to fix this code style.

Tucchhaa: I will create a card to fix this, thank you for pointing this out :)

alexslavr (approved):

PR #30864, opened 2025-08-25
Scheduler: Test for grouping css auto width

aleksei-semikozov:


e2e/testcafe-devextreme/tests/scheduler/common/grouping/groupHeaderLongNamesCss.ts
```diff
@@ -0,0 +1,172 @@
...
+    flex: 0 0 auto;
+  }
+`;
+
+const createScheduler = async (options = {}): Promise<void> => createWidget('dxScheduler', extend(DEFAULT_OPTIONS, options));
+
+fixture.disablePageReloads`Scheduler: Group Header CSS for Long Resource Names`
+  .page(url(__dirname, '../../../container.html'));
+
+test('Group header CSS should work with vertical grouping and long resource names', async (t) => {
```
Ambrozy, line 102: The code in tests is very similar.
I suggest to squash it into one test wrapped with `forEach`.
Then after a while if it will be necessary to change tests we could change only in one place. Otherwise it's often easy to forget change in all places.

```
[
   { currentView: 'Vertical Grouping' },
   { currentView: 'Horizontal Grouping' },
   { currentView: 'Group By Date', groupByDate: true },
   { currentView: 'agenda' },
].forEach((options) => {
   test('Group header CSS should work with ${options.currentView} and long resource names', async (t) => {
   ...
   });
});

```

Ambrozy (approved):

PR #30884, opened 2025-08-26
Form: create .d.ts properties for the Smart Paste

pharret31:


pharret31 (commented):
@ksercs
> This looks strange to me - I'm just worried there will be conflicts after we add more AI features - this "ai result" can be just a string or maybe even another type.
> 
> The type looks very general - shouldn't we reuse some type from aiIntegration core types? 🤔

This AIResult only applies to the form representing name-value structure after parsing actual AI response that we're going to show to user without typification
From the form side we won't return back other types since we have only fields and their values in here
Maybe instead it could be better to rename it like FormAIResult?

anna-shakhova (approved):

PR #30890, opened 2025-08-26
chore: add subscription name to the trial panel

VasilyStrelyaev:


ajivanyandev (dismissed):

packages/devextreme/js/__internal/core/license/trial_panel.client.ts
```diff
@@ -185,12 +186,14 @@ class DxLicense extends SafeHTMLElement {
...
     const contentContainer = document.createElement('div');
     contentContainer.style.cssText = this._contentStyles;
+    const subscriptions = this.getAttribute(attributeNames.subscriptions);
     contentContainer.append(
       this._createSpan('For evaluation purposes only. Redistribution prohibited. Please '),
       this._createLink('register', this.getAttribute(attributeNames.licensingDoc) as string),
       this._createSpan(' an existing license or '),
       this._createLink('purchase a new license', this.getAttribute(attributeNames.buyNow) as string),
       this._createSpan(` to continue use of DevExpress product libraries (v${this.getAttribute(attributeNames.version)}).`),
+      this._createSpan(subscriptions ? ` Included in Subscriptions: ${subscriptions}.` : ''),
```
chaosmirage: [optional]

Render an element only if `subscriptions` is not empty to avoid rendering an empty element like `<span></span>`


```
const elements = [
      this._createSpan('For evaluation purposes only. Redistribution prohibited. Please '),
      this._createLink('register', this.getAttribute(attributeNames.licensingDoc) as string),
      this._createSpan(' an existing license or '),
      this._createLink('purchase a new license', this.getAttribute(attributeNames.buyNow) as string),
      this._createSpan(` to continue use of DevExpress product libraries (v${this.getAttribute(attributeNames.version)}).`),
];

if (subscriptions) {
  elements.push(this._createSpan(` Included in Subscriptions: ${subscriptions}.`));
}

contentContainer.append(...elements);
```

chaosmirage: Maybe we don't need this condition if the `subscriptions` value will be a required parameter (doesn't have a default value `''`). Do we have a case when the `subscriptions` value is allowed to be `undefined`?

chaosmirage (approved):

ajivanyandev (approved):

PR #30901, opened 2025-08-27
fix(demos): replace external process.js with window.process

chaosmirage:


chaosmirage:
re-opened here https://github.com/DevExpress/DevExtreme/pull/30902

PR #30909, opened 2025-08-27
TreeView: localize selectAll aria-label (T1299342)

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/ui/tree_view/tree_view.base.ts
+++ b/packages/devextreme/js/__internal/ui/tree_view/tree_view.base.ts
@@ -1496,7 +1496,7 @@ class TreeViewBase extends HierarchicalCollectionWidget<TreeViewBaseProperties, 
     const isAllSelected = this._dataAdapter.isAllSelected();
     this._createComponent(this._$selectAllItem, CheckBox, {
       value: isAllSelected,
-      elementAttr: { 'aria-label': 'Select All' },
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/ui/tree_view/tree_view.base.ts
```diff
@@ -1496,7 +1496,7 @@ class TreeViewBase extends HierarchicalCollectionWidget<TreeViewBaseProperties,
     const isAllSelected = this._dataAdapter.isAllSelected();
     this._createComponent(this._$selectAllItem, CheckBox, {
       value: isAllSelected,
-      elementAttr: { 'aria-label': 'Select All' },
+      elementAttr: { 'aria-label': messageLocalization.format('dxList-selectAll') },
```
juleari, line 1499: maybe we should use this.options.selectAllText?
i see one in defaultOptions

EugeniyKiyashko (approved):

EugeniyKiyashko (approved):

packages/devextreme/js/__internal/ui/tree_view/tree_view.base.ts
anna-shakhova: user may redefine text for selectAllText, at the same time aria-label should be clear enough and smth that is expected by users, they still can read custom text with their reader

juleari (approved):

PR #30917, opened 2025-08-28
Form: add indication for Smart Past feature

EugeniyKiyashko:


packages/devextreme/js/__internal/ui/form/form.ts
```diff
@@ -1794,6 +1823,7 @@ class Form extends Widget<FormProperties> {
     this._abort?.();
     this._abort = undefined;
     this._currentAICommand = undefined;
+    this._hideLoadPanel();
```
anna-shakhova, line 1826: _proceesCommandCompletion is called on aiIntegration option change or on smartPaste trigger during previous call still in progress
Hiding panel here may cause blinking

packages/devextreme/js/__internal/ui/form/form.ts
```diff
@@ -165,9 +170,21 @@ class Form extends Widget<FormProperties> {
     this._itemsRunTimeInfo = new FormItemsRunTimeInfo();
     this._groupsColCount = [];
 
+    this._initFormLoadPanel();
```
anna-shakhova: this will affect all existing forms, perhaps, we can init it only if aiIntegration is set initially or using option changed

packages/devextreme/js/__internal/ui/form/form.ts
```diff
@@ -1846,8 +1880,14 @@ class Form extends Widget<FormProperties> {
       instruction: item.aiOptions?.instruction,
     }));
 
+    const clipboardText = text ?? await navigator.clipboard?.readText();
+    if (!clipboardText) {
```
anna-shakhova: if no text is passed and clipboard is empty, panel will blink, but, perhaps, too quickly (getting field data not so slow) to understand that we started process and ended due to data
we can show loadPanel after all preparations OR set some timer to ensure it is noticable

packages/devextreme/js/__internal/ui/form/form.ts
```diff
@@ -1846,8 +1880,14 @@ class Form extends Widget<FormProperties> {
...
+      this._hideLoadPanel();
```
juleari: why do we silently bury the request? without even a warning in console? or showing something to a user?

packages/devextreme/js/__internal/ui/form/form.ts
```diff
@@ -1846,8 +1880,14 @@ class Form extends Widget<FormProperties> {
       instruction: item.aiOptions?.instruction,
     }));
 
+    const clipboardText = text ?? await navigator.clipboard?.readText();
```
juleari: since it's not necessarily a clipboardText, can we name this variable more accurate?
like `smartPasteText`?

packages/devextreme/js/__internal/ui/form/form.ts
juleari: don't we wont to give user an option to use this loadPanel on submit?

packages/devextreme/js/__internal/ui/form/form.ts
EugeniyKiyashko: This should be a separate task when will add warnings

packages/devextreme/js/__internal/ui/form/form.ts
EugeniyKiyashko: We always import LoadPanel during implementation. Another option is to ask the customer to add the import manually. Let's discuss separatelly

packages/devextreme/js/__internal/ui/form/form.ts
EugeniyKiyashko: I hope this part, and the existence of aiIntegration, should be at the top of the method along with the warning processing

packages/devextreme/js/__internal/ui/form/form.ts
EugeniyKiyashko: Fixed

packages/devextreme/js/__internal/ui/form/form.ts
EugeniyKiyashko: Fixed

packages/devextreme/js/__internal/ui/form/form.ts
EugeniyKiyashko: Reworked

packages/devextreme/js/__internal/ui/form/form.ts
EugeniyKiyashko: Will check this scenario during testing

anna-shakhova (dismissed):

juleari (approved):

PR #30957, opened 2025-09-01
Form Smart Paste: Handle possible issues in smartPaste method

juleari:


packages/devextreme/js/__internal/ui/form/form.layout_manager.ts
```diff
@@ -734,6 +734,25 @@ class LayoutManager extends Widget<LayoutManagerProperties> {
         ...config,
         ...item.buttonOptions ?? {},
       };
+      if (item.name === 'smartPaste') {
+        const form = this._getFormOrThis();
+        const { aiIntegration } = form.option();
+        if (!aiIntegration && item.buttonOptions && !isDefined(item.buttonOptions.disabled)) {
+          item.buttonOptions.disabled = true;
+        }
```
pharret31: Seems unobvious, so if user sets disabled to false then it would be enabled and we'll still should handle that case?

packages/devextreme/js/__internal/ui/form/form.ts
```diff
@@ -1879,12 +1882,32 @@ class Form extends Widget<FormProperties> {
...
     };
-    this._abort = aiIntegration?.[command](params, callbacks);
+    this._abort = aiIntegration[command](params, callbacks);
+  }
+
+  private _updateFieldWithSmartPasteValue(dataField: string, value: SmartPasteCommandResult[number]['value']): void {
+    const { formData } = this.option();
+
+    if (isDefined(formData)) {
+      let resultValue: string | string[] | boolean = value;
```
pharret31: What do you think about extracting this type somewhere to make it named?

packages/devextreme/js/__internal/ui/form/form.ai.utils.ts
```diff
@@ -23,6 +27,71 @@ const getEditorTypeInfo = (editorType: FormItemComponent | undefined): string =>
   }
 };
 
+export const parseResultForEditorType = (
+  dataField: string,
+  editorType: FormItemComponent | undefined,
+  value: SmartPasteCommandResult[number]['value'],
+): string | string[] | boolean => {
+  const errorValue = Array.isArray(value) ? `[${value}]` : `'${value}'`;
```
pharret31: JSON.stringify?

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.ai.utils.tests.js
```diff
@@ -80,3 +81,364 @@ QUnit.module('getItemFormatInfo', () => {
         assert.strictEqual(format, expectedFormat, 'list of acceptable values is correct, custom values are allowed');
     });
 });
+
+QUnit.module('parseResultForEditorType', () => {
+    QUnit.test('should return correct value for dxAutocomplete', function(assert) {
+        const value = 'string';
+
+        assert.strictEqual(parseResultForEditorType('', 'dxAutocomplete', value), value, 'string is correct value for dxAutocomplete');
```
pharret31: I would put value into the description so if it's changed no need to rewrite it in two places:
```assert.strictEqual(parseResultForEditorType('', 'dxAutocomplete', value), value, `${value} is correct value for dxAutocomplete`);```
And for other places where applicable as well

File: packages/devextreme/js/__internal/ui/form/form.layout_manager.ts (hunk 1 above)
Comment (anna-shakhova) [reply packages/devextreme/js/__internal/ui/form/form.layout_manager.ts]: we discussed that default smartPaste button should be always enabled, regardless of aiIntegration option

Comment (juleari) [reply packages/devextreme/js/__internal/ui/form/form.layout_manager.ts]: if developer sets disabled to false, then it is his desicion. and the button will not be disabled. smartPaste method will throw an error, when user clicks this button.

File: packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.ai.utils.tests.js [hunk 5]
```diff
@@ -80,3 +81,364 @@ QUnit.module('getItemFormatInfo', () => {
...
+            parseResultForEditorType('dataField', 'dxCalendar', value);
+            assert.ok(false, `parsing should fail for '${value}' for dxCalendar`);
+        } catch(error) {
+            assert.ok(true, `parsing failed for '${value}' for dxCalendar`);
+            assert.strictEqual(errorsSpy.calledOnce, true, 'one error was thrown');
+            assert.strictEqual(errorsSpy.lastCall.args[0], 'E1064', 'the error had correct number');
+            assert.strictEqual(errorsSpy.lastCall.args[1], 'dataField', 'the error had correct dataField');
+            assert.strictEqual(errorsSpy.lastCall.args[2], `'${value}'`, 'the error had correct value');
+            assert.strictEqual(errorsSpy.lastCall.args[3], 'date', 'the error had expected type');
+        }
```
pharret31: you can use assert.throws in such places

https://qunitjs.com/api/assert/throws/

packages/devextreme/js/__internal/ui/form/form.ts
```diff
@@ -1879,12 +1882,32 @@ class Form extends Widget<FormProperties> {
...
+
+      const dataItems = this._itemsRunTimeInfo.getItemsForDataExtraction();
```
anna-shakhova: we have O(n2) difficulty here also it is a bit complicated to read, just to get edtiorType, better map editorType by dataField once

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.ai.utils.tests.js
```diff
@@ -80,3 +81,364 @@ QUnit.module('getItemFormatInfo', () => {
...
+        } catch(error) {
+            assert.ok(true, `parsing failed for [${value}] for dxRangeSlider`);
+            assert.strictEqual(errorsSpy.calledOnce, true, 'one error was thrown');
+            assert.strictEqual(errorsSpy.lastCall.args[0], 'E1064', 'the error had correct number');
+            assert.strictEqual(errorsSpy.lastCall.args[1], 'dataField', 'the error had correct dataField');
+            assert.strictEqual(errorsSpy.lastCall.args[2], `[${value}]`, 'the error had correct value');
+            assert.strictEqual(errorsSpy.lastCall.args[3], 'number range', 'the error had expected type');
+        }
+        errorsSpy.restore();
+    });
```
pharret31: Actually it seems all these tests can be rewritten using forEach and one test case, doesn't it?
You can use the same syntax as in test below and pass to test objects like:
```ts
[{ value, component, errorType }, ...].forEach(data => QUnit.test('', function(assert) {...}));
```

juleari: it can, i thought about it, but desided against since it will be way too big array of objects. and it will stop looking like "simple tests".

but I can rewrite it, if you insist.

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.aiIntegration.tests.js
```diff
@@ -36,6 +44,69 @@ QUnit.module('aiIntegration option', () => {
         assert.strictEqual(smartPaste2.calledOnce, true, 'new smartPaste invoked after update');
         assert.deepEqual(smartPaste1.getCall(0).args, smartPaste1.getCall(0).args, 'new smartPaste invoked with same text and fields data');
     });
+
+    QUnit.test('Default smartPaste button should be enabled if aiIntegration is set', function(assert) {
```
anna-shakhova: these test can be combined with forEach, using editorType, value, expected format variables

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.ai.utils.tests.js
pharret31: Not bigger than bunch of similar tests I believe :) I think it would be even easier to read when you have all cases that you want to test with the same logic in one place so if you want to check what's going on there you can just look around maybe even without scrolling a page and see that all these cases are handled :)
It could look like this:
https://github.com/DevExpress/DevExtreme/blob/5ec817918840aaf05b72e8a0af9f62d966e7e379/packages/devextreme/testing/tests/DevExpress.ui.widgets/listParts/commonTests.js#L1032

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.aiIntegration.tests.js
juleari: the tests you referring to don't use editorType or value
they are about default smart button

at the end of file there are dxCheckBox, dxSwitch, dxHtmlEditor and dxDateRangeBox tests.
dxCheckBox and dxSwitch are similar between each other (and differ from others because the resulted value is not string|[string]), but they also different from each other since one of them have default value of undefined, and other false.
dxHtmlEditor has html wrapper to its value. so it is also different from the others.

mistakes are similar though, but to me it does not look right to group mistakes, and not editors

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.ai.utils.tests.js
juleari: I created this monster)

packages/devextreme/js/__internal/ui/form/form.ai.utils.ts
```diff
@@ -5,6 +5,8 @@ import type { FormItemComponent, SimpleItem } from '@js/ui/form';
 import errors from '@js/ui/widget/ui.errors';
 import { dateUtilsTs } from '@ts/core/utils/date';
 
+export type ParsedAiValue = string | string[] | boolean;
```
pharret31: We agreed to use ai letters in code as either AI or ai, so the type should be ParsedAIValue

juleari: renamed it

anna-shakhova (approved):

pharret31 (approved):

PR #30959, opened 2025-09-01
Scheduler: avoid double date transform out of view model to display date

Ambrozy:

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.test.ts
@@ -14,11 +14,11 @@ import { getTargetedAppointment } from './get_targeted_appointment';
 
 const dataAccessor = new AppointmentDataAccessor(mockUppercaseFieldExpressions, true);
 const timeZoneCalculator = createTimeZoneCalculator('America/Los_Angeles');
-const appointment = {
+const appointment: any = {
   StartDate: new Date(200, 0, 0),
   EndDate: new Date(200, 0, 1),
 };
-const appointmentRecurrence = {
+const appointmentRecurrence: any = {
   ...appointment,
   RecurrenceRule: 'FREQ=DAILY',
 };
@@ -54,7 +54,7 @@ describe('getTargetedAppointment', () => {
       {
         info: getInfo(),
         groupIndex: 0,
-      },
+      } as any,
       dataAccessor,
       timeZoneCalculator,
       getResourceManagerMock(),
@@ -74,8 +74,8 @@ describe('getTargetedAppointment', () => {
       appointmentRecurrence,
       {
         info: getInfo(),
-        groupIndex: 5,
-      },
+        groupIndex: 5, // 0,1; 0,2; 0,3; 0,4; 1,1; 1,2; <- 5
+      } as any,
       dataAccessor,
       timeZoneCalculator,
       resourceManager,
@@ -97,7 +97,7 @@ describe('getTargetedAppointment', () => {
         isAgendaModel: true,
         info: getInfo(),
         groupIndex: 0,
-      },
+      } as any,
       dataAccessor,
       timeZoneCalculator,
       getResourceManagerMock(),
@@ -117,7 +117,7 @@ describe('getTargetedAppointment', () => {
         isAgendaModel: true,
         info: getInfo(),
         groupIndex: 0,
-      },
+      } as any,
       dataAccessor,
       timeZoneCalculator,
       getResourceManagerMock(),
@@ -136,15 +136,15 @@ describe('getTargetedAppointment', () => {
       {
         isAgendaModel: true,
         info: getInfo(),
-        groupIndex: 5,
-      },
+        groupIndex: 3, // 0,1; 0,2; 0,3; 0,4; <- 3
+      } as any,
       dataAccessor,
       timeZoneCalculator,
       resourceManager,
     )).toEqual({
       ...appointment,
-      assigneeId: [2],
-      roomId: 1,
+      assigneeId: [4],
+      roomId: 0,
       displayStartDate: appointment.StartDate,
       displayEndDate: appointment.EndDate,
     });
--- a/packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.test.ts
@@ -1,0 +1,152 @@
+/**
+ * @timezone Europe/Belgrade
+ */
+
+import {
+  describe, expect, it,
+} from '@jest/globals';
+import { getResourceManagerMock } from '@ts/scheduler/__mock__/resource_manager.mock';
+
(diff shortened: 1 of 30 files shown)


packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.ts
```diff
@@ -0,0 +1,96 @@
...
+  timeZoneCalculator: TimeZoneCalculator,
+  resourceManager: ResourceManager,
+): T & {
+  displayStartDate: Date;
+  displayEndDate: Date;
+} => {
+  const targetedAdapter = new AppointmentAdapter(
+    rawAppointment,
+    dataAccessor,
+  ).clone();
```
aleksei-semikozov: 🤔 This seems strange to me. Why are we cloning a newly created instance? Is this really necessary, or perhaps the name of the method should better reflect its purpose?

packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.test.ts
```diff
@@ -0,0 +1,152 @@
+/**
+ * @timezone Europe/Belgrade
```
aleksei-semikozov, line 2: 🤔 This confuses me a bit — the @timezone Europe/Belgrade annotation doesn’t seem to apply here, since below we’re explicitly creating a timeZoneCalculator('America/Los_Angeles'). Is this annotation actually needed?

packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.ts
```diff
@@ -0,0 +1,96 @@
...
+  return {
+    ...rawTargetedAppointment,
+    displayStartDate: new Date(info.appointment.startDate),
+    displayEndDate: new Date(info.appointment.endDate),
+  };
+};
+
+export const getTargetedAppointment = <T extends Appointment>(
+  rawAppointment: T,
+  settings: Pick<AppointmentItemViewModel, 'info' | 'groupIndex'>
```
aleksei-semikozov: The settings type here feels a bit confusing — it’s a union of different Pick<>s and a full model. It’s not clear what shape we should expect. We only use getTargetedAppointment twice in the main code, and every time I need to re-check what settings actually is. For tests it’s even a third shape. Maybe we could introduce a clearer, dedicated type (e.g. TargetedAppointmentSettings) to make the intent easier to follow?

packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.ts
```diff
@@ -0,0 +1,96 @@
...
+  dataAccessor: AppointmentDataAccessor,
+  resourceManager: ResourceManager,
+): T & {
+  displayStartDate: Date;
+  displayEndDate: Date;
+} => {
+  const targetedAdapter = new AppointmentAdapter(
+    rawAppointment,
+    dataAccessor,
+  ).clone();
```
aleksei-semikozov: I wrote about this bellow

packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.ts
```diff
@@ -0,0 +1,96 @@
...
+  resourceManager: ResourceManager,
+): void => {
+  const { groups, resourceById, groupsLeafs } = resourceManager;
+  if (groups.length) {
+    const cellGroups = getLeafGroupValues(groupsLeafs, settings.groupIndex);
+    setAppointmentGroupValues(rawAppointment, resourceById, cellGroups);
+  }
+};
+
+export const getTargetedAppointmentFromInfo = <T extends Appointment>(
```
aleksei-semikozov: 🤔 I’m not sure the generic <T> adds much value here. We never explicitly pass a type parameter, and in practice we always end up with something that already looks like a TargetedAppointment (with displayStartDate/displayEndDate)
Instead of a generic, maybe we could narrow the signature and directly return a TargetedAppointment. That would make the contract clearer and avoid the impression of unnecessary abstraction.

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/adapt_agenda_settings.ts
```diff
@@ -0,0 +1,43 @@
+import type { TimeZoneCalculator } from '../../r1/timezone_calculator';
+import { AppointmentAdapter } from '../../utils/appointment_adapter/appointment_adapter';
+import type { AppointmentDataAccessor } from '../../utils/data_accessor/appointment_data_accessor';
+import { plainViewModel } from './plain_view_model';
```
aleksei-semikozov, line 4: /// I think we can rewrite `plain_view_model` a little bit simple
```import type { SafeAppointment } from '@ts/scheduler/types';

import type {
  AgendaViewModelSettingsInternal,
  AppointmentViewModelInternal,
  AppointmentViewModelSettingsInternal,
} from './types';

type PlainViewModelItem =
  AppointmentViewModelSettingsInternal &
  AgendaViewModelSettingsInternal & {
    itemData: SafeAppointment;
  };

export const plainViewModel = (
  viewModel: AppointmentViewModelInternal[],
): PlainViewModelItem[] => viewModel.flatMap((appointment) => appointment.settings.map(
  (setting) => ({
    ...setting,
    itemData: appointment.itemData,
  }),
));
```

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/add_collector.ts
```diff
@@ -73,12 +74,23 @@ const processVirtualAppointment = (
...
 export const addCollector = (
   viewModel: AppointmentViewModelInternal[],
+  timeZoneCalculator: TimeZoneCalculator,
 ): AppointmentViewModelPlain[] => {
   const settings = plainViewModel(viewModel);
   const result: AppointmentViewModelPlain[] = [];
   const virtualAppointments: Record<string, AppointmentCollectorViewModel> = {};
 
-  settings.forEach((setting) => {
+  settings.map((setting) => ({
```
aleksei-semikozov: 💡 Maybe we should rename settings here — it’s not really an array of settings, but rather flattened view model items. Something like viewModelItems or appointmentItems would make the intent clearer.

packages/devextreme/js/__internal/scheduler/utils/get_targeted_appointment.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      {
+        info: getInfo(),
+        groupIndex: 5,
+      },
+      dataAccessor,
+      timeZoneCalculator,
+      resourceManager,
+    )).toEqual({
+      ...appointmentRecurrence,
+      assigneeId: [2],
```
aleksei-semikozov, line 84: 🤔 I don’t fully understand why we expect assigneeId: [2] and roomId: 1 here. There’s no hint in this file about how these values are produced. I think it comes from the getResourceManagerMock, but it’s not clear.

It might be clearer if in arrange step we set the mock data directly

aleksei-semikozov (approved):

PR #30975, opened 2025-09-02
Chart: type pointInfo in tooltip.contentTemplate

anna-shakhova:


packages/devextreme/js/viz/chart.d.ts
```diff
@@ -397,6 +397,325 @@ export interface baseLabelObject {
...
+  /**
+   * @docid
+   * @public
+   */
+  rangeValue2Text?: string;
+  /**
+   * @docid
+   * @public
+   */
+  seriesName?: any;
```
anna-shakhova, line 610: In `baseSeriesObject` `name` typed as `any`, so this corresponding field typed same way

EugeniyKiyashko (dismissed):

pharret31 (dismissed):

packages/devextreme/js/viz/chart.d.ts
```diff
@@ -397,6 +397,325 @@ export interface baseLabelObject {
...
+  /**
+   * @docid
+   * @public
+   */
+  rangeValue1Text?: string;
+  /**
+   * @docid
+   * @public
+   */
+  rangeValue12?: string | number | Date;
```
IlyaKhD: Shouldn't it be [rangeValue2](https://js.devexpress.com/React/Documentation/ApiReference/UI_Components/dxChart/Series_Types/RangeAreaSeries/#rangeValue2Field)?

```suggestion
  rangeValue2?: string | number | Date;
```

packages/devextreme/js/viz/chart.d.ts
```diff
@@ -397,6 +397,325 @@ export interface baseLabelObject {
...
+   * @docid
+   * @public
+   */
+  originalValue?: string | number | Date;
+};
+
+/**
+ * @public
+ */
+export type BaseChartPointInfo = BasePointInfo<chartPointObject>;
```
IlyaKhD: Generally we avoid making *base* types public, preferring union instead. Can we keep it private?

packages/devextreme/js/viz/chart.d.ts
```diff
@@ -397,6 +397,325 @@ export interface baseLabelObject {
...
+
+/**
+ * public
+ */
+export type PointInfo =
+  | BaseChartPointInfo
```
IlyaKhD:
```suggestion
```
`PointInfo` already includes `BasePointInfo<chartPointObject>` through `StackedPointInfo` and `BubblePointInfo`

packages/devextreme/js/viz/chart.d.ts
anna-shakhova: typo, thanks!

packages/devextreme/js/viz/chart.d.ts
```diff
@@ -397,6 +397,325 @@ export interface baseLabelObject {
...
+  originalValue?: string | number | Date;
+};
+
+/**
+ * @public
+ */
+export type BaseChartPointInfo = BasePointInfo<chartPointObject>;
+
+/**
+ * public
```
IlyaKhD:
```suggestion
 * @public
```

packages/devextreme/js/viz/chart.d.ts
anna-shakhova: It is not base in meaning of it should be hidden, it is shared part between different series types, that also may be used as is
renamed

packages/devextreme/js/viz/chart.d.ts
anna-shakhova: fixed typo

EugeniyKiyashko (dismissed):

alexslavr (approved):

EugeniyKiyashko (approved):

PR #30980, opened 2025-09-03
Form Smart Paste: fix AIIntegration tests

juleari:


anna-shakhova (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.aiIntegration.tests.js
```diff
@@ -1189,14 +1217,19 @@ QUnit.module('SmartPaste', () => {
...
-                    assert.strictEqual(form.getEditor(dataField).option('value'), value, 'text value was set');
+                    setTimeout(() => {
+                        assert.strictEqual(form.getEditor(dataField).option('value'), value, 'text value was set');
+                        done();
+                    }, 10);
                 } catch(error) {
                     assert.ok(false, `failed with error: ${error}`);
+                    done();
                 }
             });
```
pharret31, line 1234: why do we even need the try catch blocks here if we don't test throw behavior? I mean if it fails it will fail without them as well

juleari: it will. in this case we will see an error in the same way, as it in other tests in this file something like this: https://github.com/DevExpress/DevExtreme/blame/25_2/packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.aiIntegration.tests.js#L358-L362

anna-shakhova (approved):

PR #31009, opened 2025-09-05
Editor: unify value and reset(value) types (T1304359)

marker-dao:


packages/devextreme/js/ui/calendar.d.ts
```diff
@@ -236,9 +236,9 @@ export default class dxCalendar extends Editor<dxCalendarOptions> {
      * @docid
      * @publicName reset(value)
      * @public
-     * @param1 value:Date|number|string|Array<Date|number|string|null>|null
+     * @param1 value:Date | number | string | Array<Date | number | string | null> | null | undefined
```
ksercs: `undefined` is not a default for Calendar's `value` => we don't need to state it explicitly in value type to fix Angular generation => don't see a reason to specify it here as well

Or do you add it for some another reason?

p.s. I see it in some other places, please check it as well

packages/devextreme/js/ui/switch.d.ts
```diff
@@ -111,8 +111,9 @@ export default class dxSwitch extends Editor<dxSwitchOptions> {
      * @docid
      * @publicName reset(value)
      * @public
+     * @param1 value:boolean | undefined
```
ksercs: as fr as I understand this doccoment should be specified when we have a difficult type and generator cannot understand it (like DateLike). But for simple case it looks to be excess

ksercs (dismissed):
Left notes about doccomments, the rest ts par lgtm

packages/devextreme/js/ui/calendar.d.ts
marker-dao: Reverted: [391236de65f6ad03a6c4e10bbcf405a1b436d440].

packages/devextreme/js/ui/switch.d.ts
marker-dao: Reverted: [391236de65f6ad03a6c4e10bbcf405a1b436d440].

ksercs (dismissed):

vladaskorohodova (approved):

alexslavr (approved):

Zedwag (approved):

r-farkhutdinov (approved):

PR #31023, opened 2025-09-09
Editor: unify value and reset(value) types (T1304359)

marker-dao:


packages/devextreme/js/ui/editor/editor.d.ts
```diff
@@ -112,13 +112,24 @@ export interface EditorOptions<TComponent> extends WidgetOptions<TComponent> {
      */
     readonly isDirty?: boolean;
 }
+
+/**
+ * @docid
+ * @hidden
+ */
+export interface EditorValuableObject {
```
vladaskorohodova: I don't really understand this name. Why is this object valuable?

marker-dao: Because it contains value. This utility interface with `@hidden` doctag ensures that all types that `extend EditorValuableObject` will contain a value field.

vladaskorohodova: Okay. I may not understand this to full extent, but believe me when I say that the current name is not it, even if it will be hidden.

I have a little pile of names for you to choose from. If you think one of them is okay, go on and rename. If not, I will come up with something else:
`EditorHasValueObject`
`EditorWithValueObject`
`EditorValueHolderObject`

marker-dao: `EditorOptionsWithValue`

vladaskorohodova (approved):

alexslavr (approved):

Zedwag (approved):

r-farkhutdinov (approved):

PR #31024, opened 2025-09-09
pnpm: add 'frozen-lockfile' flag

wdevfx:


e2e/devextreme-bundler/package.json
```diff
@@ -0,0 +1,10 @@
+{
```
pomahtri, line 1: [Can be postponed for future]
The right way of creating new e2e is to add to package.json:

- all deps in "devDeps" or "deps" instead of installing them on CI. Here it is devextreme package
- "build" script that will build the package. Here it can be empty because it doesn't need any building in first place
- "test" script that will trigger testing actions. Here it will be calling "devextreme-init-bundle" and other stuff

After that, there's no any additional scripts in github workflow needed. All projects with "test" target are check in "Default Workflow"

.github/workflows/build_all.yml
```diff
@@ -64,22 +63,26 @@ jobs:
...
+        run: mv ./npm/devextreme-2*.tgz ./e2e/devextreme-bundler-test/devextreme.tgz
 
       - name: Setup deps
+        working-directory: ./e2e/devextreme-bundler-test
         run: |
-          npm init -y
-          npm i webpack terser-webpack-plugin
-          npm i ./devextreme.tgz
+          pnpm install --filter devextreme-bundler-test --frozen-lockfile
+          pnpm add i ./devextreme.tgz
```
pomahtri: I don't see syntax 'pnpm add i' on pnpm docs. Is it ok?
https://pnpm.io/cli/add
image

wdevfx: It's a typo, thank u for u catch -> fixed

e2e/devextreme-bundler/package.json
wdevfx: Yup, agree with u here
But let's postpone this improvement for a separate task :)

pomahtri (approved):

PR #31033, opened 2025-09-10
Form Smart Paste: trim results after getting a value from AI

pharret31:


packages/devextreme/js/__internal/ui/form/form.ai.utils.ts
```diff
@@ -34,61 +34,68 @@ export const parseResultForEditorType = (
...
+      return normalizedValue;
     case 'dxHtmlEditor':
-      if (Array.isArray(value)) {
+      if (Array.isArray(normalizedValue)) {
         throw errors.Error('E1064', dataField, errorValue, 'string');
       }
-      return value;
+      return normalizedValue;
     default:
-      return value;
```
juleari: are we sure, that we want to trim everything? even textarea values?

pharret31: I guess so, AI will remove the spaces around any text by itself in any case so currently it seems that if it adds a space somewhere it's more of an artifact

juleari (dismissed):

EugeniyKiyashko (dismissed):

packages/devextreme/js/__internal/core/ai_integration/commands/smartPaste.ts
```diff
@@ -29,9 +29,15 @@ export class SmartPasteCommand extends BaseCommand<
       }
 
       const [name, ...values] = data.split(':::');
```
anna-shakhova:
```suggestion
      const [name, ...rowValues] = data.split(':::');
      const values = rowValues.map((value) => value.trim());
      const value = values.length <= 1 ? values[0] : values;
```

anna-shakhova (approved):

PR #31042, opened 2025-09-11
Toolbar Menu: ignore preventScrollEvent deprecation (T1306077)

pharret31:


packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
```diff
@@ -244,6 +244,7 @@ export default class DropDownMenu extends Widget<DropDownMenuProperties> {
       },
       deferRendering: false,
       preventScrollEvents: false,
+      _ignorePreventScrollEventsDeprecation: true,
```
anna-shakhova, line 247: shouldn't we just remove option, deprecated in 23.1?

anna-shakhova (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets/toolbar.menu.tests.js
```diff
@@ -192,6 +192,12 @@ QUnit.module('render with popup', moduleConfig, () => {
 
         assert.ok(overlayTop >= buttonBottom);
     });
+
+    QUnit.test('should not throw a warning about the preventScroll event option is deprecated on popup opening (T1306077)', function(assert) {
+        const logDeprecationSpy = sinon.spy(Popup.prototype, '_logDeprecatedOptionWarning');
```
ksercs: You can spy for `errors.log`, please check how it's done in other places. Then you will not need to use a private API

Also it's better to restore the spy if you're spying for a method of a real instance

packages/devextreme/js/__internal/ui/toolbar/internal/toolbar.menu.ts
ksercs: We cannot because it overrides default behavior. This option is a fallback and does not have an alternative. 
So I recommend you do it this way for now by analogy with Calendar and FilterBuilder

EugeniyKiyashko (approved):

PR #31032, opened 2025-09-10
Form Smart Paste: created a screenshot test for loader

juleari:

PR #31044, opened 2025-09-11
Form: eslint fix Smart Paste React demo

ksercs:


ksercs:
Is fixed in https://github.com/DevExpress/DevExtreme/pull/31032

PR #31051, opened 2025-09-11
SpeechToText: Add .d.ts types

r-farkhutdinov:


packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+   * @default 'stopfilled'
+   * @public
+   */
+  stopIcon?: string;
+
+  /**
+   * @docid dxSpeechToTextOptions.webSpeechApiConfig
+   * @public
+   */
+  webSpeechApiConfig?: WebSpeechApiConfig | { [key: string]: any } | null;
```
ksercs: do we really need null here, isn't `{}` enough?

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  /**
+   * @docid dxSpeechToTextOptions.customSpeechRecognizer
+   * @public
+   */
+  customSpeechRecognizer?: CustomSpeechRecognizer;
```
ksercs: I'm not sure if we need `default` and maybe `type` doccoments here, please recheck

r-farkhutdinov: Re-checked. Looks like it's not required to build successfully

ksercs (dismissed):

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
+import {
+    EventInfo,
```
marker-dao: ### ⚠ chore(blocking)
extra spase
```suggestion
  EventInfo,
```

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+   * @docid dxSpeechToTextOptions.webSpeechApiConfig
+   * @public
+   */
+  webSpeechApiConfig?: WebSpeechApiConfig | { [key: string]: any };
+
+  /**
+   * @docid dxSpeechToTextOptions.customSpeechRecognizer
+   * @public
+   */
+  customSpeechRecognizer?: CustomSpeechRecognizer;
```
marker-dao: ### 💡 suggestion(non-blocking)
Let's use alphabetical order for properties (except for events, which should be at the end, as you have done).
In any way it's up to you.

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+
+import Widget from './widget/ui.widget';
+import type { Properties as dxButtonOptions } from './button';
+
+/**
+ * @docid
+ * @namespace DevExpress.ui.dxSpeechToText
+ * @public
+ */
+export type WebSpeechApiConfig = {
```
marker-dao, line 16: ### ❕ thought(non-blocking)
I am sure we should not base our types on this package, but I will leave this link here as useful material.
https://www.npmjs.com/package/@types/webspeechapi?activeTab=code

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+   * @default 'stopfilled'
+   * @public
+   */
+  stopIcon?: string;
+
+  /**
+   * @docid dxSpeechToTextOptions.webSpeechApiConfig
+   * @public
+   */
+  webSpeechApiConfig?: WebSpeechApiConfig | { [key: string]: any };
```
marker-dao: ### 🎉 praise
Actually, I think this is the best solution as long as we have generator restrictions when using Record<string, unknown>.

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+ * @inherits EventInfo,ChangedOptionInfo
+ */
+export type OptionChangedEvent = EventInfo<dxSpeechToText> & ChangedOptionInfo;
+
+/**
+ * @public
+ * @docid dxSpeechToTextOptions
+ * @type object
+ */
+export type Properties = Omit<dxButtonOptions, 'template' | 'useSubmitBehavior' | 'validationGroup' | 'icon' | 'text' | 'onClick'> & {
```
marker-dao: ### 💡 issue(blocking)

Actually, I think there is an architectural error here, because SpeechToText is inherited not from Button, but from Widget. And Button is used as an internal component. Accordingly, it would be more correct to inherit SpeechToText options not from dxButtonOptions, but from WidgetOptions. Check out how these types are inherited in this part of my PR and notice how Button types are used in this interface: 

https://github.com/DevExpress/DevExtreme/blob/ee5ff51f0853e88ed55c0bf6bf4efed92847a122/packages/devextreme/js/__internal/ui/speech_to_text/types/properties.ts#L13

Source PR: https://github.com/DevExpress/DevExtreme/pull/31061

marker-dao: ### 💡 issue(blocking)

We also have an agreement to use dxComponentOptions and override it in type Properties, as shown here:
https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/ui/chat.d.ts#L526

```suggestion
export interface dxSpeechToTextOptions extends WidgetOptions<dxSpeechToText>
```

I asked the API Review group why we should do this, and as soon as I receive a response, I will share it with you.

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+  webSpeechApiConfig?: WebSpeechApiConfig | { [key: string]: any };
+
+  /**
+   * @docid dxSpeechToTextOptions.customSpeechRecognizer
+   * @public
+   */
+  customSpeechRecognizer?: CustomSpeechRecognizer;
+
+  /**
+   * @docid dxSpeechToTextOptions.onStartClick
```
marker-dao, line 194:
```suggestion
   * @docid dxSpeechToTextOptions.onStartClick
   * @default undefined
```

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+   * @type_function_param1 e:{ui/speech_to_text:StartClickEvent}
+   * @action
+   * @public
+   */
+  onStartClick?: ((e: StartClickEvent) => void);
+
+  /**
+   * @docid dxSpeechToTextOptions.onStopClick
```
marker-dao, line 203:
```suggestion
   * @docid dxSpeechToTextOptions.onStopClick
   * @default undefined
```

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+   * @type_function_param1 e:{ui/speech_to_text:StopClickEvent}
+   * @action
+   * @public
+   */
+  onStopClick?: ((e: StopClickEvent) => void);
+
+  /**
+   * @docid dxSpeechToTextOptions.onResult
```
marker-dao, line 212:
```suggestion
   * @docid dxSpeechToTextOptions.onResult
   * @default undefined
```

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+   * @type_function_param1 e:{ui/speech_to_text:ResultEvent}
+   * @action
+   * @public
+   */
+  onResult?: ((e: ResultEvent) => void);
+
+  /**
+   * @docid dxSpeechToTextOptions.onError
```
marker-dao, line 221:
```suggestion
   * @docid dxSpeechToTextOptions.onError
   * @default undefined
```

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+  isListening?: boolean;
+};
+
+/**
+ * @docid _ui_speech_to_text_StartClickEvent
+ * @public
+ * @type object
+ * @inherits NativeEventInfo
+ */
+export type StartClickEvent = NativeEventInfo<dxSpeechToText, KeyboardEvent | MouseEvent | PointerEvent>;
```
marker-dao: ### ❕ issue(blocking)

You can create a PointerLikeEvent as shown here:
https://github.com/DevExpress/DevExtreme/blob/3c369f9a0c3f7a44a1794cf4e0d859fd865623e7/packages/devextreme/js/__internal/ui/speech_to_text/types/events.ts#L5

because you also need `TouchEvent`, and this will help you avoid copy-pasting.

Source PR: https://github.com/DevExpress/DevExtreme/pull/31061.

```suggestion
export type StartClickEvent = NativeEventInfo<dxSpeechToText, PointerLikeEvent>;
```

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+
+/**
+ * @docid _ui_speech_to_text_StopClickEvent
+ * @public
+ * @type object
+ * @inherits NativeEventInfo
+ */
+export type StopClickEvent = NativeEventInfo<dxSpeechToText, KeyboardEvent | MouseEvent | PointerEvent>;
```
marker-dao: ### ❕ issue(blocking)

```suggestion
export type StopClickEvent = NativeEventInfo<dxSpeechToText, PointerLikeEvent>;
```

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+
+/**
+ * @docid _ui_speech_to_text_ResultEvent
+ * @public
+ * @type object
+ * @inherits NativeEventInfo
+ */
+export type ResultEvent = NativeEventInfo<dxSpeechToText, Event>;
```
marker-dao: ~~There is no a `result` property here~~
### Upd.
Here, we should not use the additional result property, as [event](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition/result_event#event_properties) will contain this property.

packages/devextreme/js/ui/speech_to_text.d.ts
```diff
@@ -0,0 +1,274 @@
...
+
+/**
+ * @docid _ui_speech_to_text_ErrorEvent
+ * @public
+ * @type object
+ * @inherits NativeEventInfo
+ */
+export type ErrorEvent = NativeEventInfo<dxSpeechToText, Event>;
```
marker-dao: ~~There is no a `error` property here~~
### Upd.
Here, we should not use the additional result property, as [event](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition/error_event#event_type) will contain this property.

packages/devextreme/js/ui/speech_to_text.d.ts
ksercs: The linked package is deprecated (
`lib.dom` also does not include types for WebSpeech now

packages/devextreme/js/ui/speech_to_text.d.ts
ksercs: Internal implementation (inherit in JS / does not inherit in JS) is not what we should rely on when implementing types I think. If we use composition everywhere across codebase instead of inheriting it would not force us to extend all components properties types from WIdget, right?

Semantically STT is the Button with additional functionality => when we designing API we say "we take all Button properties except 4 of this and add these several options" =>  it's better for user to see this relationship explicitly

Also there are couple of advantages of this approach for us - not duplicating prop types (like stylingMode/type) and easier support of new Button props (if we add new UI props in design system project for example).

Examples of similar approach in other components:
- DateRangeBox inherits Widget but uses DateBox inside - [.d.ts](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/ui/date_range_box.d.ts#L178)
- CheckBox inherits Widget but uses Editor through composition - [.d.ts](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/ui/check_box.d.ts#L58)
- ...

So this approach looks good to me. Please elaborate more if you see any concerns

packages/devextreme/js/ui/speech_to_text.d.ts
ksercs: If we do this, we also need to state "undefined" explicitly in the type (for Angular metadata generation).
If we really need to specify default isn't it better to use `null` as an explicit empty value?

packages/devextreme/js/ui/speech_to_text.d.ts
ksercs: If we do this, it should be a core types reused in all such places so let's better do it in a separate PR to not stop

packages/devextreme/js/ui/speech_to_text.d.ts
marker-dao: > At first they called it dx***Options everywhere, then they decided to rename it to Properties within the framework of the renaming "widget -> component, option -> property" and within the framework of the guideline "no dx*** prefixes"
> 
> But the internal utilities are tied to the dx***Options pattern, so all we can do is continue adding this type to .d.ts, but immediately deprecate it and alias it

packages/devextreme/js/ui/speech_to_text.d.ts
ksercs: P.s. don't forget to add a touch event

packages/devextreme/js/ui/speech_to_text.d.ts
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/31051/commits/79e10928847d536434ca3b379aa2296deb32de78

packages/devextreme/js/ui/speech_to_text.d.ts
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/31051/commits/79e10928847d536434ca3b379aa2296deb32de78

packages/devextreme/js/ui/speech_to_text.d.ts
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/31051/commits/79e10928847d536434ca3b379aa2296deb32de78

packages/devextreme/js/ui/speech_to_text.d.ts
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/31051/commits/79e10928847d536434ca3b379aa2296deb32de78

packages/devextreme/js/ui/speech_to_text.d.ts
marker-dao: Regarding this approach, I am not sure that this is how it should be, despite the fact that this is how it is done in some other components.

My concern is that we don't automatically support passing all parameters from STT to Button. Based on the types, it may seem to the user that everything in the button will be available in STT, but the reality is different—it won't be available unless we implement it in STT.

Otherwise, in JS, we will have to pass all STT options to the button, removing the 4 that we decided not to support. But that's not how it works now.

marker-dao: At least, my code does not display the transfer of all options from STT to Button, but only some of them:

https://github.com/DevExpress/DevExtreme/blob/df79eaa3bc66ea447001fcd4e851d7f903ec5392/packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts#L88

marker-dao: We decided to try using Pick. No changes on JS part.

ksercs: Discussed in voice - now we think that the safest approach it to extend widget but pick necessary props from buttonOptions

ksercs: We rechecked that current generation and generation if we use deprecated Options is the same, so gonna try using of this approach

ksercs (dismissed):

ksercs (dismissed):

ksercs (approved):

marker-dao (approved):

PR #31059, opened 2025-09-12
ButtonGroup (Inferno): branch component

anna-shakhova:


packages/devextreme/js/__internal/ui/button_group_inferno/inferno_widget.tsx
```diff
@@ -0,0 +1,50 @@
+import { infernoRenderer } from '@ts/core/m_inferno_renderer';
```
juleari, line 1: shouldn't this inferno_widget be moved outside of button_group?
it looks like it is more general class, than just a part of button_group

anna-shakhova: It would be later, when we finalize our solution. For now it placed here just to preserve all changes in one place and do not interfere to the other squad's area

EugeniyKiyashko (approved):

PR #31060, opened 2025-09-12
Menu: dx-state-focused class should not be added if focusStateEnabled is set to false (T1306949)

EugeniyKiyashko:


packages/devextreme/testing/tests/DevExpress.ui/collectionWidget.tests.js
```diff
@@ -1970,7 +1972,47 @@ module('focus policy', {
...
+
+        instance.option('focusedElement', $item.get(0));
+
+        assert.strictEqual($item.hasClass(FOCUSED_STATE_CLASS), false, `${FOCUSED_STATE_CLASS} class is not applied when focusStateEnabled is false`);
+        assert.strictEqual($element.attr('aria-activedescendant'), String(focusedItemId), 'aria-activedescendant is set on container');
+        assert.strictEqual(!!$item.attr('id'), true, 'focused item has id attribute');
+        assert.strictEqual($item.attr('id').includes(focusedItemId), true, 'focused item id contains focusedItemId');
+    });
+
+    test(`${FOCUSED_STATE_CLASS} class is applied when focusStateEnabled is true`, function(assert) {
```
pharret31: Maybe it's better to use something like "focused state class is applied when forusStateEnabled is true" just in case of simplify search? That won't rely on the actual class name but will show the purpose of test

pharret31 (dismissed):

pharret31 (approved):

r-farkhutdinov (approved):

PR #31061, opened 2025-09-12
SpeechToText: Create component

marker-dao:


packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+class SpeechToText extends Widget<SpeechToTextProperties> {
+  private _button!: Button;
+
+  private _state!: SpeechToTextState;
+
+  private _actions!: SpeechToTextActions;
+
+  _getDefaultOptions(): SpeechToTextProperties {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return extend(super._getDefaultOptions(), {
```
r-farkhutdinov: Should we use extend here? If possible, let's use spread instead

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+      type: 'default',
+      startIcon: DEFAULT_INITIAL_ICON,
+      stopIcon: DEFAULT_LISTENING_ICON,
+      startText: '',
+      stopText: '',
+      customSpeechRecognizer: {
+        enabled: false,
+        isListening: false,
+      },
+      onStartClick: undefined,
```
r-farkhutdinov, line 44: I believe these 4 callbacks don't exist in properties, so it should be fine if we don't explicitly set `undefined` value. Omitting these will not change the logic.

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+        this._handleCustomEngineState();
+        break;
+
+      case 'stylingMode':
+      case 'type':
+      case 'disabled':
+      case 'width':
+      case 'height':
+        this._button?.option(name, value);
+        if (name === 'disabled') {
```
r-farkhutdinov: Let's remove this `if` and put `_setState` under `disabled` case with 1 line of duplication of `this._button?.option(name, value);`

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
ksercs: I'd prefer to use approach which is used across codebase

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  private _state!: SpeechToTextState;
+
+  private _actions!: SpeechToTextActions;
+
+  _getDefaultOptions(): SpeechToTextProperties {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return extend(super._getDefaultOptions(), {
+      stylingMode: 'contained',
+      type: 'default',
```
ksercs: Hm, should really the default `type` be default? (blue button)
It's `normal` it button.
P.s. let's not duplicate inherited props if default value is the same as in parent (at least stylingMode)

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
ksercs: I think it's better to state it explicitly at least for us to clearly see what new options we declare at this level

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  private _getCurrentText(): string {
+    const { startText, stopText } = this.option();
+
+    return this._isListening()
+      ? stopText ?? ''
+      : startText ?? '';
+  }
+
+  private _buttonClickHandler(e: ClickEvent): void {
```
ksercs: let's state with a verb as in other places
```suggestion
  private _handleButtonClick(e: ClickEvent): void {
```

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+    const { customSpeechRecognizer } = this.option();
+
+    if (!customSpeechRecognizer?.enabled) {
+      this._setState(SpeechToTextState.LISTENING);
+    }
+
+    this._actions.onStartClick?.({
+      component: this,
+      element: this.$element().get(0) as HTMLElement,
+      event: e.event,
```
ksercs, line 150: if `e` is a `ClickEvent`, shouldn't it here be proxied as `event: e`? 🤔

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+  private _handleStartClick(e: ClickEvent): void {
+    const { customSpeechRecognizer } = this.option();
+
+    if (!customSpeechRecognizer?.enabled) {
+      this._setState(SpeechToTextState.LISTENING);
+    }
+
+    this._actions.onStartClick?.({
+      component: this,
+      element: this.$element().get(0) as HTMLElement,
```
ksercs: [blocking]
nope, it's still should be a jQuery element if jquery is imported to the page
```suggestion
      element: this.$element(),
```

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+      return;
+    }
+
+    this._state = newState;
+    this._updateButtonState();
+    this._updateCssClasses();
+  }
+
+  private _updateButtonState(): void {
+    if (!this._button) {
```
ksercs: How it can be nullish if we always create it on init?

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+    if (this._state === newState) {
+      return;
+    }
+
+    this._state = newState;
+    this._updateButtonState();
+    this._updateCssClasses();
+  }
+
+  private _updateButtonState(): void {
```
ksercs, line 178: Let's make the naming more strict? It does not look like we gonna extend this method
```suggestion
  private _updateButtonIconAndText(): void {
```

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+        break;
+
+      case 'stylingMode':
+      case 'type':
+      case 'disabled':
+      case 'width':
+      case 'height':
+        this._button?.option(name, value);
+        if (name === 'disabled') {
+          this._setState(value ? SpeechToTextState.DISABLED : SpeechToTextState.INITIAL);
```
ksercs: So if I'm in listening state with a custom engine and change height the state will be changed to "initial"?

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: No, because it's a ButtonClickEvent:
```
export type ClickEvent = NativeEventInfo<dxButton, KeyboardEvent | MouseEvent | PointerEvent> & {
    /**
     * @docid _ui_button_ClickEvent.validationGroup
     * @type object
     */
    validationGroup?: any;
};
```

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: Not quite: the button is initialized in `_initMarkup`, not in the constructor, so it doesn't exist for a while.

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: No because of `if (name === 'disabled') {` condition.

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: Hmm, yes, you're right, we actually have a utility for selecting which element to return: `getPublicElement`. I'll use it, it automatically resolves the element type.

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: I don't think this is a good idea, because then depending on the state we will update the `aria-label` directly in this method.

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
ksercs: but it's anyway always exists when this method is called, nah?
I also see u use type assertion `_button!`  so ts should not be worried it's undefined

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -0,0 +1,619 @@
...
+            startIcon: 'micoutline',
+            stopIcon: 'stopfilled',
+            startText: '',
+            stopText: '',
+        };
+
+        Object.entries(expectedDefaults).forEach(([key, value]) => {
+            assert.strictEqual(this.instance.option(key), value, `${key} has correct default value`);
+        });
+    });
```
ksercs: No need to test it here - it's already checked in `defaultOptions.tests.js`
Same about "component should have proper customSpeechRecognizer default" test below

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -0,0 +1,619 @@
...
+        this.instance._clean();
+
+        assert.ok(!this.$element.hasClass(SPEECH_TO_TEXT_LISTENING_CLASS), 'state reset to initial');
+    });
+
+    QUnit.test('should dispose actions on _dispose', function(assert) {
+        const actionsBeforeDispose = Object.keys(this.instance._actions).length;
+        assert.ok(actionsBeforeDispose > 0, 'actions exist before dispose');
+
+        this.instance._dispose();
```
ksercs, line 590: we can use a public `dispose` method here

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -0,0 +1,619 @@
...
+
+        this.reinit({ onStartClick: onStartClickSpy });
+        this.getButton().trigger('dxclick');
+
+        assert.ok(onStartClickSpy.calledOnce, 'event works after first init');
+
+        this.reinit({ onStartClick: onStartClickSpy });
+        this.getButton().trigger('dxclick');
+
+        assert.ok(onStartClickSpy.calledTwice, 'event works after reinit');
```
ksercs: Hm, how can it be not called twice if you passed the same handler? It's not clear to me what this test is for. Did you want to check that previous event handler is unsubscribed on reinit? Then it would be more intuitive to pass a new spy on reinit and check explicitly that new one is called and old one is not

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -0,0 +1,619 @@
...
+
+        this.reinit({ onStopClick: onStopClickSpy });
+
+        const $button = this.getButton();
+
+        $button.trigger('dxclick');
+        $button.trigger('dxclick');
+
+        assert.ok(onStopClickSpy.calledOnce, 'onStopClick event fired exactly once');
+    });
```
ksercs, line 213: Would be also good to check that events can be subscribed through `on` method

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -0,0 +1,619 @@
...
+
+    QUnit.test('should pass correct event arguments to onStopClick', function(assert) {
+        assert.expect(4);
+
+        this.reinit({
+            onStopClick: (e) => {
+                assert.ok(e.component instanceof SpeechToText, 'component is correct');
+                assert.strictEqual(e.element, this.$element.get(0), 'element is provided correctly');
+                assert.ok(e.event, 'original event is provided');
+                assert.strictEqual(typeof e.event, 'object', 'event is an object');
```
ksercs, line 238: Test would be more strict if you trigger not just a click but trigger an event directly `$.Event('click')` - then there will be a single assert that event is passed correctly

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -0,0 +1,619 @@
...
+        this.instance.option('customSpeechRecognizer', {
+            isListening: true,
+        });
+
+        assert.ok(!this.$element.hasClass(SPEECH_TO_TEXT_LISTENING_CLASS), 'missing enabled handled');
+    });
+});
+
+QUnit.module('Options', moduleConfig, () => {
+    QUnit.test('should update button icon and text based on state', function(assert) {
```
ksercs: This test is duplicated on "State management" module - "should update button options when state changes"

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -0,0 +1,619 @@
...
+            height: 50,
+        });
+
+        const buttonInstance = this.getButtonInstance();
+
+        assert.strictEqual(buttonInstance.option('width'), 200, 'width applied to button');
+        assert.strictEqual(buttonInstance.option('height'), 50, 'height applied to button');
+    });
+
+    QUnit.test('should handle stylingMode changes at runtime', function(assert) {
```
ksercs, line 508: Why do you test `stylingMode` but not test `type` for example? Or width and height but not focusStateEnabled?
Maybe it's better to write a single test like "button-related option should be passed to button at runtime change"?

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    return {
+      stylingMode,
+      type,
+      disabled,
+      width,
+      height,
+      activeStateEnabled,
+      focusStateEnabled,
+      hoverStateEnabled,
```
ksercs, line 105: If we don't need to pass `rtlEnabled` because it's well synced at `createComponent` level then we mostly don't also need to pass `*StateEnabled` props, no?
Also it's strange to see that we pass these properties here but don't handle its runtime change;
I think we should not look at it at all if it's well synced on core level OR handle both initial and runtime change how you did for width/height

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: [9e2a7752de838dac87a3a6aa040186168a64c19c]

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: [9e2a7752de838dac87a3a6aa040186168a64c19c]

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: [9e2a7752de838dac87a3a6aa040186168a64c19c]

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: [9e2a7752de838dac87a3a6aa040186168a64c19c]

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: Used `this.element()`: [9e2a7752de838dac87a3a6aa040186168a64c19c]

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: optionChanged on init calls this method
> u use type assertion

already refactored it: [9e2a7752de838dac87a3a6aa040186168a64c19c]

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
marker-dao: Yes, already refactored it here: [9e2a7752de838dac87a3a6aa040186168a64c19c]

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
marker-dao: No, because they have different functions. Please investigate it.

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/31061/commits/c2171e1e32e5be072ef7feacc7b28b1ccf0c58c6

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/31061/commits/c2171e1e32e5be072ef7feacc7b28b1ccf0c58c6

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/31061/commits/c2171e1e32e5be072ef7feacc7b28b1ccf0c58c6

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/31061/commits/c2171e1e32e5be072ef7feacc7b28b1ccf0c58c6

ksercs (approved):

r-farkhutdinov (approved):

PR #31064, opened 2025-09-14
Demos: updated add-demo script

juleari:


apps/demos/utils/interactive/add-demo.js
```diff
@@ -4,25 +4,40 @@ const path = require('path');
...
+const menuMetaUtils = require('../shared/menu-meta-utils');
 const menuMetaData = require('../../menuMeta.json');
 
 const existingApproaches = ['jQuery', 'Angular', 'React', 'Vue'];
-
-// const descriptionFileName = 'description.md';
+const extraModules = [
+  'signalr',
+  'devextreme-aspnet-data-nojquery',
+  'globalize',
```
pharret31, line 14: I see that in menuMeta.json this module is used like `globalize--vue&react', what for this is used and is it possible to support such syntax or is it not needed?

apps/demos/utils/shared/menu-meta-utils.js
```diff
@@ -0,0 +1,155 @@
...
+  menuMetaData.push(categoryMeta);
+};
+
+const addGroup = (menuMetaData, pathParts, groupName) => {
+  const groupMeta = { Name: groupName, Equivalents: '', Demos: [] };
+  const groupParent = getByPath(menuMetaData, pathParts);
+  groupParent.push(groupMeta);
+};
+
+const addDemo = (menuMetaData, pathParts, demoName, widgetName, equivalents) => {
```
pharret31, line 13: Also I noticed that this command removes the last empty line in the menuMeta.json file 🤔

apps/demos/utils/interactive/add-demo.js
```diff
@@ -4,25 +4,40 @@ const path = require('path');
...
+  'devextreme-exceljs-fork&file-saver',
+  'jspdf',
+  'jspdf&jspdf-autotable',
```
pharret31, line 17: don't really get it, is & works as , or is it something else? if so then perhaps it's possible to remove jspdf from the second line?

pharret31: currently if I use both modules they are added like this: `Modules: ...jspdf,jspdf&jspdf-autotable...`

apps/demos/utils/interactive/add-demo.js
juleari: modules from `menuMeta.json` are used to create config in `prepare-bundles` script.
on `update-config` step of `prepare-budles` configs for each approach are copied from relative templates from https://github.com/juleari/DevExtreme/tree/a2c96e9ebcc4db985033b4c97561c10fd4dde5eb/apps/demos/configs

modules from `menuMeta.json` splited by `,` and then sent to 'removeModules' function (https://github.com/juleari/DevExtreme/blob/a2c96e9ebcc4db985033b4c97561c10fd4dde5eb/apps/demos/utils/internal/create-config.js#L32-L36) which removes every module except ones in the list.

`globalize--vue&react` is not present in template configs. but `globalize` is.

apps/demos/utils/interactive/add-demo.js
juleari: since it was used only once (affects only one Demo, I'll change default [configs](https://github.com/juleari/DevExtreme/blob/a2c96e9ebcc4db985033b4c97561c10fd4dde5eb/apps/demos/configs/React/config.js#L83-L92) to make them cleaner. but maybe the point was to not allow jspdf-autotable without jspdf)

juleari: from https://www.npmjs.com/package/jspdf-autotable/v/3.0.0-alpha.3 it looks like `jspdf-autotable` can't exist without `jspdf`

so 2 options:
1. we remove `jspdf&jspdf-autotable from` configs, and check if `jspdf` is selected when `jspdf-autotable` is selected
2. we leave `jspdf&jspdf-autotable` and check unselect jspdf if jspdf-autotable is selected.

the second one looks safer, and it can be done not just for jspdf, but for any `&`-separated modules. so i'll do the second one

pharret31 (approved):

pomahtri (approved):

PR #31074, opened 2025-09-16
Scheduler: refactor integration appointment form popup test

aleksei-semikozov:

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/popup.ts
+++ b/packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/popup.ts
@@ -1,0 +1,136 @@
+export class PopupModel {
+  element: HTMLDivElement;
+
+  constructor(element: HTMLDivElement) {
+    this.element = element;
+  }
+
+  getLabelIdByText = (labelText: string): string => {
+    const labels = Array.from(this.element.querySelectorAll('label'));
+
+    const label = labels.find((l) => l?.textContent?.trim()?.startsWith(labelText));
+
+    if (!label) {
+      throw new Error(`Label with text "${labelText}" not found`);
+    }
+
+    const forId = label.getAttribute('for');
+    if (!forId) {
+      throw new Error(`Label with text "${labelText}" has no "for" attribute`);
+    }
+    return forId;
+  };
+
+  getInputByLabel = (labelText: string): HTMLInputElement => {
+    const forId = this.getLabelIdByText(labelText);
+
+    const input = this.element.querySelector(`input#${forId}`) as HTMLInputElement;
+
+    if (!input) {
+      throw new Error(`Input with id "${forId}" not found`);
+    }
+
+    return input;
+  };
+
+  setInputValueByLabel = (labelText: string, value: string): HTMLInputElement => {
+    const input = this.getInputByLabel(labelText);
+    if (!input) {
+      throw new Error(`Input with label "${labelText}" not found`);
+    }
+    input.value = '';
+
+    value.split('').forEach((char) => {
+      input.value += char;
+      input.dispatchEvent(new Event('input', { bubbles: true }));
+      input.dispatchEvent(new KeyboardEvent('keydown', { bubbles: true, key: char }));
+      input.dispatchEvent(new KeyboardEvent('keypress', { bubbles: true, key: char }));
+      input.dispatchEvent(new KeyboardEvent('keyup', { bubbles: true, key: char }));
+    });
+
+    input.dispatchEvent(new Event('change', { bubbles: true }));
+    input.dispatchEvent(new Event('input', { bubbles: true }));
+
+    return input;
+  };
+
+  getSwitchByName = (name: string): HTMLInputElement => {
+    const hiddenInput = this.element.querySelector<HTMLInputElement>(`input[name=${name}]`);
+
+    if (!hiddenInput) {
+      throw new Error(`Switch with name "${name}" not found`);
+    }
+
+    return hiddenInput;
+  };
+
+  selectRadio = (value: string): Element | null => {
+    const group = this.element.querySelector('[role="radiogroup"]');
+    if (!group) throw new Error('Radiogroup not found');
+
+    const radios = Array.from(group.querySelectorAll('[role="radio"]'));
+
+    const target = radios.find((radio) => {
+      const label = radio.getAttribute('aria-label')?.trim();
+      const text = radio.textContent?.trim();
+      return label === value || text === value;
+    });
+
+    if (!target) throw new Error(`Radio with value "${value}" not found`);
+
+    radios.forEach((r) => {
+      r.setAttribute('aria-checked', 'false');
+      r.classList.remove('dx-item-selected', 'dx-radiobutton-checked');
+    });
+
+    target.setAttribute('aria-checked', 'true');
+    target.classList.add('dx-item-selected', 'dx-radiobutton-checked');
+
+    target.dispatchEvent(new MouseEvent('click', { bubbles: true }));
+
+    return target;
+  };
+
+  getSelectedRadio = (): HTMLElement | null => this.element.querySelector('[role="radio"][aria-checked="true"]');
+
+  getSelectedRadioValue = (): string | null => {
+    const selected = this.getSelectedRadio();
+    return selected?.getAttribute('aria-label') ?? selected?.textContent?.trim() ?? null;
+  };
+
+  getForm = (): HTMLElement | null => this.element.querySelector('.dx-form');
+
+  getTitle = (): HTMLElement | null => document.querySelector('.dx-popup-title');
+
+  getDoneButton = (): HTMLButtonElement => {
+    const doneButton = this.element.querySelector('.dx-button.dx-popup-done') as HTMLButtonElement;
+    if (!doneButton) {
+      throw new Error('Done button not found');
+    }
+    return doneButton;
+  };
+
+  getCancelButton = (): HTMLButtonElement => {
+    const cancelButton = this.element.querySelector('.dx-button.dx-popup-cancel') as HTMLButtonElement;
+    if (!cancelButton) {
+      throw new Error('Cancel button not found');
+    }
+    return cancelButton;
+  };
+
+  getFormEditor = (fieldName: string): HTMLElement | null => {
+    const form = this.getForm();
+    if (form === null) {
+      return null;
+    }
+    return form.querySelector(`[data-field="${fieldName}"]`);
+  };
+
+  getEditSeriesButton = (): HTMLElement => {
+    const editSeriesButton = document.querySelector('[aria-label="Edit series"]') as HTMLElement;
+    if (!editSeriesButton) {
+      throw new Error('Edit series button not found');
+    }
+    return editSeriesButton;
+  };
+}
--- a/packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
@@ -1,0 +1,804 @@
+import {
+  afterEach, beforeEach, describe, expect, it, jest,
+} from '@jest/globals';
+
+import fx from '../../../common/core/animation/fx';
+import { createScheduler } from '../__tests__/__mock__/create_scheduler';
+import { setupSchedulerTestEnvironment } from '../__tests__/__mock__/m_mock_scheduler';
+
+const defaultData = [
+  {
+    text: 'recurrent-app',
+    startDate: new Date(2017, 4, 1, 9, 30),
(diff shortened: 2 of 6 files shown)


sjbur (dismissed):

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+  afterEach, beforeEach, describe, expect, it, jest,
+} from '@jest/globals';
+
+import fx from '../../../common/core/animation/fx';
+import { createScheduler } from '../__tests__/__mock__/create_scheduler';
+import { setupSchedulerTestEnvironment } from '../__tests__/__mock__/m_mock_scheduler';
+import type { AppointmentModel } from '../__tests__/__mock__/model/appointment';
+import type { SchedulerModel } from '../__tests__/__mock__/model/scheduler';
+
+function getLabelIdByText(
```
Ambrozy: A lot of helper functions in test file. Move it to POM in model directory. And create own file for Popup model. You can also check POM of testcafe for inspiration https://github.com/DevExpress/DevExtreme/blob/f34ec6dc2597d4d60399dd4d07f7d47353acf717/packages/testcafe-models/scheduler/appointment/popup.ts 

It's better to keep test files clean and simple. When you face with issue it's good to understand test fast instead of waste of time trying figure out what's going on.

Object better then single functions:
[not clear where] getLabelIdByText()
[clear where] popup.getLabelIdByText()
Object sets context.

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+function getInputByLabel(
+  labelText: string,
+  root: Document | HTMLElement = document,
+): HTMLInputElement {
+  const forId = getLabelIdByText(labelText, root);
+
+  const input = root.querySelector(`input#${forId}`) as HTMLInputElement;
+
+  if (!input) {
+    throw new Error(`Input with id "${forId}" not found`);
```
Ambrozy: I like the Idea of throwing reasonable errors

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+}
+
+function getSelectedRadioValue(container: HTMLElement): string | null {
+  const selected = getSelectedRadio(container);
+  return selected?.getAttribute('aria-label') ?? selected?.textContent?.trim() ?? null;
+}
+
+describe('Appointment popup form', () => {
+  beforeEach(() => {
+    fx.off = true;
```
Ambrozy, line 35: Question: what does this construction for? `fx.off` can we remove it?

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+    setupSchedulerTestEnvironment();
+  });
+
+  afterEach(() => {
+    fx.off = false;
+    document.body.innerHTML = '';
+    jest.useRealTimers();
```
Ambrozy, line 42: should we create function `cleanSchedulerTestEnvironment` and move this code there?

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+      },
+      height: 600,
+    });
+
+    scheduler.showAppointmentPopup();
+
+    const form = POM.getForm() as HTMLElement;
+    setInputValueByLabel('Subject', 'qwerty', form);
+
+    POM.clickDoneButton();
```
Ambrozy: Better use object model instead of plain function.
For example:
`setInputValueByLabel('Subject', 'qwerty', form);`
`POM.clickDoneButton();`
=>
```
const popup = POM.getPopup();
popup.getInput('Subject').setValue('qwerty')
popup.getDoneButton().click()

```
It require more work with object model, but it's more readable

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+    fx.off = false;
+    document.body.innerHTML = '';
+    jest.useRealTimers();
+  });
+
+  it('Original appointment\'s fields shouldn\'t fill if used fieldExpr', async () => {
+    const data: Record<string, unknown>[] = [];
+    const textExpValue = 'Subject';
+
+    let newAppointment: Record<string, unknown> | null = null;
```
Ambrozy: I think better just use `any` and avoid of `@ts-expect` below. Complicated type makes visual noise, but not the type is the purpose of test.

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+    setupSchedulerTestEnvironment();
+  });
+
+  afterEach(() => {
+    fx.off = false;
+    document.body.innerHTML = '';
+    jest.useRealTimers();
+  });
+
+  it('Original appointment\'s fields shouldn\'t fill if used fieldExpr', async () => {
```
Ambrozy, line 45: Let's rename it in positive way. `Should set appointment text using textExpr`

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+        customRecurrenceRule: 'FREQ=WEEKLY;BYDAY=TU,FR;COUNT=10',
+      }],
+      views: ['month'],
+      currentView: 'month',
+      currentDate: new Date(2017, 4, 25),
+      recurrenceRuleExpr: 'customRecurrenceRule',
+      height: 600,
+    });
+
+    openPopupByDblClick(POM);
```
Ambrozy: `openPopupByDblClick(POM)` => `POM.dblClickByCell()`

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+
+    const editSeriesButton = document.querySelector('[aria-label="Edit series"]') as HTMLElement;
```
Ambrozy: You use it several times. Let's create small object model for this dialog. As in testcafe https://github.com/DevExpress/DevExtreme/blob/a7e5a8bdcf7a6e420185e2813c248abc451a9271/packages/testcafe-models/scheduler/appointment/dialog.ts

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -0,0 +1,952 @@
...
+    editSeriesButton.click();
+
+    const { value: repeatInputRecurrence } = getSwitchByName('repeat');
+
+    scheduler.hideAppointmentPopup();
+    scheduler.showAppointmentPopup();
```
Ambrozy, line 98: I don't understand why an what do you check:
```
    scheduler.hideAppointmentPopup();
```
and why expect of `repeatInputRecurrence` is at the end of the test. I would like read test line by line without jumping up and down.

Maybe you have to rename test. It's unclear what is going on here. show, hide, show, hide, expect.

Ambrozy: Okey, I see, it's old tests. It's require to keep git history for these tests. Are you sure you want to move it to the jest and not just keep in QUnits?
So how do we do this:
commit 1: move QUnit file into jest directory
commit 2: replace all QUnit tests code into jest

Then ask Devops team to help merge it with rebase. It will save the git history.

By the way did you use AI?

Ambrozy (changes requested):

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: We need this setting so that we don't have to wait for the animation to draw the popup.
To remove it, we need to use useFakeTimers. This solution is much better, and I like it more.
But I couldn't easily solve this problem because of Promise issues.

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: I think we should do this after we implement the testing library, it does part of what happens here
(clearing the screen).

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: Ok, I do it in next iteration of refactor this tests

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: I use the AI for copy test cases, and create several helpers. AI didn't do a good job rewriting it.

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/popup.ts
```diff
@@ -0,0 +1,128 @@
...
+  getSelectedRadioValue = (): string | null => {
+    const selected = this.getSelectedRadio();
+    return selected?.getAttribute('aria-label') ?? selected?.textContent?.trim() ?? null;
+  };
+
+  getForm = (): HTMLElement | null => this.element.querySelector('.dx-form');
+
+  clickDoneButton = (): void => {
+    const doneButton = this.element.querySelector('.dx-button.dx-popup-done') as HTMLElement;
+    doneButton.click();
```
Ambrozy: Please, instead of action return button. I saw in test you have select .dx-button.dx-popup-done. It will be more clear to see `doneButton.click()` and `expect(doneButton).not.toBe(null)`

And the same with cancel button. Click is simple action. more flexibility for test writers will be better.

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/popup.ts
```diff
@@ -0,0 +1,136 @@
...
+  element: HTMLDivElement;
+
+  constructor(element: HTMLDivElement) {
+    this.element = element;
+  }
+
+  getLabelIdByText = (labelText: string): string => {
+    const labels = Array.from(this.element.querySelectorAll('label'));
+
+    const label = labels.find((l) => l?.textContent?.trim()?.startsWith(labelText));
```
Ambrozy, line 11: nitpick: one letter variables not easy to select by mouse or touchpad. In my perspective, better just use general word like `item`

No need to fix. Just thoughts

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/popup.ts
```diff
@@ -0,0 +1,136 @@
...
+  getForm = (): HTMLElement | null => this.element.querySelector('.dx-form');
+
+  getTitle = (): HTMLElement | null => document.querySelector('.dx-popup-title');
+
+  getDoneButton = (): HTMLButtonElement => {
+    const doneButton = this.element.querySelector('.dx-button.dx-popup-done') as HTMLButtonElement;
+    if (!doneButton) {
+      throw new Error('Done button not found');
+    }
+    return doneButton;
```
Ambrozy, line 110: I understand why in `selectRadio` we throw an error, because the last part of the function dependent on element existing. So it's not about the test, but test utils.

In this case we have not throw an error because it just return value. Let test writer decide what to do with this. For example, `getSelectedRadioValue` return null without error, why this function throw an error?

I think better not throw error on return value.

Ambrozy (approved):

PR #31078, opened 2025-09-16
Grids - CardView: Default values set in onInitNewCard are not populated in the new card's popup editor form

dmlvr:


packages/testcafe-models/cardView/index.ts
```diff
@@ -102,6 +107,15 @@ export default class CardView extends GridCore {
     return this.element.find(`.${CLASS.cardViewContent}`).hasClass(`${CLASS.selectCheckBoxesHidden}`);
   }
 
+  isReady(): Promise<boolean> {
+    const { getInstance } = this;
+
+    return ClientFunction(
+      () => (getInstance() as any).isReady(),
```
siarhei-arzamasau, line 114: @dmlvr minor: could you avoid usage of type assertion operator here?

siarhei-arzamasau (dismissed):

packages/testcafe-models/cardView/index.ts
dmlvr: @siarhei-arzamasau, I can't fix it only here. For fix it I need to change global type for `getInstance()` here - https://github.com/DevExpress/DevExtreme/blob/25_2/packages/testcafe-models/internal/widget.ts#L40

I tried to fix it by checking ⬇️, but it still doesn't work for typescript compiler, only for runtime

```
  isReady(): Promise<unknown> {
    const { getInstance } = this;

    return ClientFunction(
      () => {
        const instance = getInstance();
        if (typeof instance === 'object' && instance !== null && 'isReady' in instance && typeof instance.isReady === 'function') {
          return instance.isReady();
        }
      },
      { dependencies: { getInstance } },
    )();
  }
```

siarhei-arzamasau: ok, then I'm resolving comment

pomahtri (approved):

PR #31082, opened 2025-09-16
Scheduler: View switcher text is not translated by localization in v25.1.4

sjbur:


packages/devextreme/js/__internal/scheduler/header/m_view_switcher.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        };
+        return localizationMap[key];
+      });
+
+      const ruLocaleResult = getTabViewSwitcher(mockSchedulerHeader, {});
+
+      expect(ruLocaleResult.options.items).toEqual([
+        { type: 'day', name: 'Day', text: 'Day' },
+        { type: 'week', name: 'Week', text: 'Week' },
+        { type: 'month', name: 'Month', text: 'Month' },
```
Ambrozy: Looks like localization didn't change. `text` is still in English. And in test below

packages/devextreme/js/__internal/scheduler/header/m_view_switcher.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  describe('localization change', () => {
+    it('should update view names when locale changes from English to Russian', () => {
```
Tucchhaa: In this file we have a lot of mocks, that we will need to support in the future. For example if we will need to change viewswticher a lot, then this test will fail, because mocks will become irrelevant.

So I suggest to implement integration test instead. In the neighbouring PR, I have created an [integration test file](https://github.com/DevExpress/DevExtreme/pull/31039/files#diff-398803b0bbcb8004d6f9bf902562012827a5376260d1a800c121ce86958aac10R1) for view switcher. 

What do you think?

Tucchhaa (approved):

PR #31091, opened 2025-09-17
ButtonGroup (Inferno): use existing Button in ButtonCollection

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/ui/button_group_inferno/button_collection.item.component.tsx
+++ b/packages/devextreme/js/__internal/ui/button_group_inferno/button_collection.item.component.tsx
@@ -1,7 +1,56 @@
+import { isDefined } from '@js/core/utils/type';
+import type { Properties as ButtonProperties } from '@js/ui/button';
 import type { Item } from '@js/ui/button_group';
+import { BaseInfernoComponent } from '@ts/core/r1/runtime/inferno/index';
+import { Button } from '@ts/grids/new/grid_core/inferno_wrappers/button';
+import { BUTTON_GROUP_CLASS } from '@ts/ui/button_group_inferno/button_collection.component';
 import type { CollectionItemProps } from '@ts/ui/collection_inferno/collection.item.component';
 import { ItemComponent } from '@ts/ui/collection_inferno/collection.item.component';
 
-export type ButtonCollectionItemProps = CollectionItemProps<Item>;
+const BUTTON_GROUP_ITEM_CLASS = `${BUTTON_GROUP_CLASS}-item`;
+const BUTTON_GROUP_FIRST_ITEM_CLASS = `${BUTTON_GROUP_CLASS}-first-item`;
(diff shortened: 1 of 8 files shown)


packages/devextreme/js/__internal/ui/button_group_inferno/button_collection.item.component.tsx
```diff
@@ -1,7 +1,56 @@
...
+      collectionItemClassName,
+      BUTTON_GROUP_ITEM_CLASS,
+      SHAPE_STANDARD_CLASS,
+      elementAttr?.class ?? null,
+      isFirst ? BUTTON_GROUP_FIRST_ITEM_CLASS : null,
+      isLast ? BUTTON_GROUP_LAST_ITEM_CLASS : null,
+      hasWidth ? BUTTON_GROUP_ITEM_HAS_WIDTH : null,
+    ];
+
+    return classes.filter(isDefined).join(' ');
```
juleari: maybe it's better to use combineClasses from our utils?
`import { combineClasses } from '@ts/core/utils/combine_classes';`

packages/devextreme/js/__internal/ui/button_group_inferno/button_collection.widget.tsx
```diff
@@ -26,6 +28,24 @@ export class ButtonCollection extends InfernoCollectionWidget<
...
 
+  protected getItemProps(
+    item: ButtonCollectionItemProps,
+    index: number,
+  ): CollectionItemProps<ButtonCollectionItemProps> {
+    return {
+      ...super.getItemProps(item, index),
+      isFirst: index === 0,
+      isLast: index === (this.props?.items?.length ?? 0) - 1,
+      hasWidth: isDefined(this.props?.width),
```
juleari, line 43: is it intentional that hasWidth would be true for `width == 0`?

packages/devextreme/js/__internal/ui/collection_inferno/collection.item.component.tsx
```diff
@@ -11,13 +11,14 @@ export class ItemComponent<
   TProps extends CollectionItemProps,
 > extends BaseInfernoComponent<TProps> {
   render(): JSX.Element {
-    const { text, isSelected } = this.props;
-    const className = `dx-item ${isSelected ? ' dx-item-selected' : ''}`;
+    const { isSelected } = this.props;
+    const className = `dx-item${isSelected ? ' dx-item-selected' : ''}`;
```
juleari: can't there be more in the future? like disabled/focused/active?
if so, maybe we should use combineClasses here as well?

packages/devextreme/js/__internal/ui/button_group_inferno/button_collection.widget.tsx
anna-shakhova: this behavior is from legacy button collection, kept the same for backward compatibility

PR #31104, opened 2025-09-18
✨AI Column: New column type / Unsupported Sorting properties

Raushen:


packages/devextreme/js/__internal/grids/grid_core/ai_column/m_ai_column_controller.ts
```diff
@@ -0,0 +1,23 @@
+import type { ColumnsController } from '../columns_controller/m_columns_controller';
+import modules from '../m_modules';
```
Alyar666: Minor comment: I think it's better to use destructuring here - `import { Controller } from '../m_modules'`

packages/devextreme/js/__internal/grids/grid_core/ai_column/unsupported_base_properties.integration.test.ts
```diff
@@ -0,0 +1,108 @@
...
+        dataSource,
+        showBorders: true,
+        columns: [
+          'id',
+          {
+            caption: 'AI',
+            type: 'ai',
+            name: 'AItest',
+
+            allowSorting: true,
```
Alyar666, line 71: This test is too redundant – it checks too many things at once. It's better to split it into several tests:
- default render: check that the required attributes are present.
- allowSorting: check that clicking on the header doesn't trigger aria-sort and sorting isn't applied.
- sortOrder, sortIndex: check that aria-sort isn't present.
- calculateSortValue: check that sorting isn't applied by this value when sortOrder is specified.

packages/devextreme/js/__internal/grids/grid_core/ai_column/unsupported_base_properties.integration.test.ts
```diff
@@ -0,0 +1,108 @@
...
+        columns: [
+          'id',
+          {
+            caption: 'AI',
+            type: 'ai',
+            name: 'AItest',
+          },
+        ],
+      });
+      instance.columnOption('AItest', 'allowSorting', true);
```
Alyar666, line 96: Same thing - the test is too redundant.

packages/devextreme/js/__internal/grids/grid_core/ai_column/m_ai_column_controller.ts
```diff
@@ -0,0 +1,23 @@
...
+export class AiColumnController extends modules.Controller {
+  private _columnsController!: ColumnsController;
+
+  public init(): void {
+    this._columnsController = this.getController('columns');
+    this.addAiCommandColumn();
+  }
+
+  private addAiCommandColumn(): void {
+    const aiColumnOptions = getAiCommandColumnOptions();
```
Alyar666, line 14: Minor comment: Do we need this variable? It's used only once. Plus, the method itself is small.
I understand that it seems easier to debug with a variable. However, this variable is never used or modified anywhere else. So, this advantage is questionable in this case.

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/ai_column/m_ai_column_controller.ts
Raushen: Let's postpone it for the next requests as it will be changed

Alyar666 (approved):

PR #31118, opened 2025-09-19
SpeechToText: Add Web Speech API integration support

r-farkhutdinov:

### DIFF
--- a/packages/devextreme/js/__internal/core/speech_recognition_adapter.ts
+++ b/packages/devextreme/js/__internal/core/speech_recognition_adapter.ts
@@ -1,0 +1,76 @@
+import { getWindow } from '@js/core/utils/window';
+import type { SpeechRecognitionConfig } from '@js/ui/speech_to_text';
+
+type SpeechRecognition = SpeechRecognitionConfig & {
+  start: () => void;
+  stop: () => void;
+} & Record<string, unknown>;
+
+interface SpeechRecognitionEvents {
+  onResult: (event: Event) => void;
+  onError: (event: Event) => void;
+  onEnd: () => void;
+}
+
+const SR_CONFIG_IGNORED_PROPERTIES = ['onresult', 'onerror', 'onend'];
+export const SPEECH_RECOGNITION_ERROR_TYPE = 'speechrecognitionerror';
+export const UNSUPPORTED_BROWSER_ERROR = 'unsupported-browser';
+export const UNSUPPORTED_BROWSER_ERROR_MESSAGE = 'The browser does not support the SpeechRecognition API.';
+
+export class SpeechRecognitionAdapter {
+  private _speechRecognition?: SpeechRecognition | null;
+
+  constructor(config: SpeechRecognitionConfig, events: SpeechRecognitionEvents) {
+    const window = getWindow();
+    // @ts-expect-error SpeechRecognition API is not supported in TS
+    const SpeechRecognitionConstructor = window.SpeechRecognition || window.webkitSpeechRecognition;
+
+    if (!SpeechRecognitionConstructor) {
+      const errorEvent = new ErrorEvent(SPEECH_RECOGNITION_ERROR_TYPE, {
+        error: UNSUPPORTED_BROWSER_ERROR,
+        message: UNSUPPORTED_BROWSER_ERROR_MESSAGE,
+      });
+
+      events.onError(errorEvent);
+
+      return;
+    }
+
+    this._speechRecognition = new SpeechRecognitionConstructor();
+
+    this.applyConfig(config);
+    this._attachEvents(events);
+  }
+
+  private _attachEvents(events: SpeechRecognitionEvents): void {
+    if (!this._speechRecognition) {
+      return;
+    }
+
+    // eslint-disable-next-line spellcheck/spell-checker
+    this._speechRecognition.onend = events.onEnd;
+    // eslint-disable-next-line spellcheck/spell-checker
+    this._speechRecognition.onresult = events.onResult;
+    this._speechRecognition.onerror = events.onError;
+  }
+
+  applyConfig(config: SpeechRecognitionConfig = {}): void {
+    Object.entries(config).forEach(([key, value]) => {
+      if (this._speechRecognition && !SR_CONFIG_IGNORED_PROPERTIES.includes(key)) {
+        this._speechRecognition[key] = value;
+      }
+    });
+  }
+
+  start(): void {
+    this._speechRecognition?.start();
+  }
+
+  stop(): void {
+    this._speechRecognition?.stop();
+  }
+
+  dispose(): void {
+    this._speechRecognition = null;
+  }
+}
--- a/packages/devextreme/js/__internal/core/speech_recognition_adapter.ts
+++ b/packages/devextreme/js/__internal/core/speech_recognition_adapter.ts
@@ -12,7 +12,7 @@ interface SpeechRecognitionEvents {
   onEnd: () => void;
 }
 
-const SR_CONFIG_IGNORED_PROPERTIES = ['onresult', 'onerror', 'onend'];
+const EVENT_NAMES = ['onresult', 'onerror', 'onend'];
 export class SpeechRecognitionAdapter {
   private _speechRecognition?: SpeechRecognition | null;
 
@@ -45,7 +45,7 @@ export class SpeechRecognitionAdapter {
 
   applyConfig(config: SpeechRecognitionConfig = {}): void {
     Object.entries(config).forEach(([key, value]) => {
-      if (this._speechRecognition && !SR_CONFIG_IGNORED_PROPERTIES.includes(key)) {
+      if (this._speechRecognition && !EVENT_NAMES.includes(key)) {
         this._speechRecognition[key] = value;
       }
     });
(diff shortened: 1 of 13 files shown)


packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -19,6 +22,14 @@ enum SpeechToTextState {
 }
 
 type SpeechToTextActions = Pick<SpeechToTextProperties, 'onStartClick' | 'onStopClick' | 'onResult' | 'onError'>;
+type PointerLikeEvent = KeyboardEvent | MouseEvent | PointerEvent | TouchEvent;
+
+type SpeechRecognitionLike = WebSpeechApiConfig & {
```
marker-dao: ### 💡 thought(non-blocking)

Let's rename `SpeechRecognitionLike` to `SpeechRecognition`. But first, let's consider renaming `WebSpeechApiConfig` to `SpeechRecognitionConfig` because, in fact, Web Speech API is not a constructor, but a category of browser API for working with speech, which already includes several constructors: SpeechRecognition (which you are familiar with), [SpeechSynthesisUtterance](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis) (speech synthesis from text), and several others.

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -52,11 +65,54 @@ class SpeechToText extends Widget<SpeechToTextProperties> {
...
+    // eslint-disable-next-line spellcheck/spell-checker
+    this._speechRecognition.onend = this._handleSpeechRecognitionEnd.bind(this);
+    // eslint-disable-next-line spellcheck/spell-checker
+    this._speechRecognition.onresult = this._handleSpeechRecognitionResult.bind(this);
+    this._speechRecognition.onerror = this._handleSpeechRecognitionError.bind(this);
+  }
+
+  _initSpeechRecognition(): void {
+    // @ts-expect-error SpeechRecognition API is not supported in TS
+    const SpeechRecognitionConstructor = window.SpeechRecognition || window.webkitSpeechRecognition;
```
marker-dao: ### 🟥 issue(non-blocking)

Hi, I would create a separate class in a new file to work with the Web Speech API instead of keeping all the logic in the base SpeechToText class.

All browser APIs are wrapped in adapters for testing and cross-platform compatibility and are stored at the `__internal/core/` level. See how this is done for `m_dom_adapter.ts`, `m_http_request.ts`, `m_resize_observer.ts`, 
`m_ajax.ts`.

marker-dao: ### 💡 issue(blocking)

It needs to use getWindow function instead of just global window:

```
import { getWindow } from '@js/core/utils/window';
const window = getWindow();
```

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -52,11 +65,54 @@ class SpeechToText extends Widget<SpeechToTextProperties> {
...
+
+    if (this._isCustomSpeechRecognitionEnabled() || !SpeechRecognitionConstructor) {
+      return;
```
marker-dao: ### ❔ suggestion(non-blocking)

Should we notify the user about it via onError if SpeechRecognitionConstructor is unavailable?

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
```diff
@@ -129,6 +185,14 @@ class SpeechToText extends Widget<SpeechToTextProperties> {
       : startText ?? '';
   }
 
+  private _emitNativeEvent<K extends keyof Pick<SpeechToTextActions, 'onResult' | 'onError'>>(name: K, event: Event): void {
```
marker-dao, line 160: ### 💡 issue(non-blocking)

Copy past: _emitNativeEvent and _emitDxEvent are the same.

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -22,6 +23,22 @@ QUnit.testStart(() => {
 
 const moduleConfig = {
     beforeEach: function() {
+        this.originalSpeechRecognition = window.SpeechRecognition;
+
+        class SpeechRecognitionMock {
```
marker-dao: ### ❕ thought(non-blocking)

I am sure that tests for SpeechRecognition should be separate for the SpeechRecognition utility, and there should be integration tests for this utility in SpeechToText.

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -22,6 +23,22 @@ QUnit.testStart(() => {
...
+                this.stop = sinon.spy();
+                this.onresult = null;
+                this.onerror = null;
+                this.onend = null;
+            }
+        }
+
+        window.SpeechRecognition = SpeechRecognitionMock;
+
+        this.getSpeechRecognition = () => this.instance._speechRecognition;
```
marker-dao: ### 💡 suggestion(non-blocking)

Move this function after the init call for consistency.

packages/devextreme/testing/tests/DevExpress.ui.widgets/speechToText.tests.js
```diff
@@ -610,3 +630,144 @@ QUnit.module('Component Lifecycle', moduleConfig, () => {
...
 });
+
+QUnit.module('Web Speech API', moduleConfig, () => {
+    QUnit.test('should initialize SpeechRecognition when available', function(assert) {
+        this.reinit();
+
+        const speechRecognition = this.getSpeechRecognition();
+
+        assert.ok(speechRecognition, 'speech recognition initialized');
+        assert.strictEqual(typeof speechRecognition.start, 'function', 'start method exists');
```
marker-dao: ### ❓ issue(non-blocking)

In SpeechToText, we do not create start and stop methods. Why check this? These methods exist because you added them in beforeEach. Most likely, this test will always be green 😊

r-farkhutdinov: Agree, this is synthetic, actually this checks that the constructor was called. We can omit checking methods and only leave the first assertion.

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
r-farkhutdinov: I think we can do this, but need to think about the error format.

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
r-farkhutdinov: True, the difference is in type signature. As these methods are really small, I would keep it as is, it can't be typed easily to support both types.

packages/devextreme/js/__internal/ui/speech_to_text/speech_to_text.ts
marker-dao: ### UPD
I discussed with Anton the principle of moving such utilities. The conclusion: anything that is used in several places should be moved; in this case, moving it may not be necessary because the API is only used in one place. But from my point of view, this remains a consistent and architecturally sound solution. Do as you see fit.

packages/devextreme/js/__internal/core/speech_recognition_adapter.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  stop(): void {
+    this._speechRecognition?.stop();
+  }
+
+  dispose(): void {
+    this._speechRecognition = null;
+  }
+
+  isAvailable(): boolean {
```
marker-dao: ### 🤔 thought(non-blocking)

This method is not used anywhere in the code except in 1 test)

packages/devextreme/js/__internal/core/speech_recognition_adapter.ts
```diff
@@ -0,0 +1,65 @@
...
+      }
+    });
+  }
+
+  start(): void {
+    this._speechRecognition?.start();
+  }
+
+  stop(): void {
+    this._speechRecognition?.stop();
```
ksercs, line 59: We can use [null object pattern](https://refactoring.guru/introduce-null-object) to avoid duplicated safe if / optional chaining. Similar approach is used for ResizeObserver - [link](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/core/m_resize_observer.ts#L22)

packages/devextreme/js/__internal/core/speech_recognition_adapter.ts
```diff
@@ -0,0 +1,65 @@
...
+  stop: () => void;
+} & Record<string, unknown>;
+
+interface SpeechRecognitionEvents {
+  onResult: (event: Event) => void;
+  onError: (event: Event) => void;
+  onEnd: () => void;
+}
+
+const SR_CONFIG_IGNORED_PROPERTIES = ['onresult', 'onerror', 'onend'];
```
ksercs: then there will be no custom abbreviation and it's clearer why it's ignored in config applying
```suggestion
const EVENT_NAMES = ['onresult', 'onerror', 'onend'];
```

ksercs (approved):

marker-dao (approved):

PR #31127, opened 2025-09-22
Scheduler: Add Appointment Form Customization  to scheduler.d.ts

sjbur:


packages/devextreme/playground/jquery copy.html
```diff
@@ -0,0 +1,300 @@
+<!DOCTYPE html>
```
aleksei-semikozov: remove this file, plz

aleksei-semikozov (changes requested):

aleksei-semikozov (dismissed):

packages/devextreme-react/src/scheduler.ts
```diff
@@ -213,30 +423,226 @@ const Editing = Object.assign<typeof _componentEditing, NestedComponentMeta>(_co
...
+const GroupItem = Object.assign<typeof _componentGroupItem, NestedComponentMeta>(_componentGroupItem, {
+  componentType: "option",
+});
+
+// owners:
+// TabPanelOptions
+// Form
 // Toolbar
 // Options
 type IItemProps = React.PropsWithChildren<{
```
anna-shakhova, line 561: Is it expected that toolbar item props have been merged with form item props?
Looks like it can cause inadequate props typing

alexslavr: Yes, this is expected behavior. In metadata we merge complex options (configuration components) with same option name (item). I agree, it looks bad here. For better typing specific form items are available (SimpleItem, GroupItem, TabbedItem, etc)

mpreyskurantov (approved):

PR #31133, opened 2025-09-22
Vitest - fix the `window.getComputedStyle` error on Windows OS

artem-kurchenko:
- Add try-catch block to handle potential errors in theme marker reading
- Return null on error to prevent exceptions from breaking theme detection


artem-kurchenko:
closed as dup of  https://github.com/DevExpress/DevExtreme/pull/31987

PR #31166, opened 2025-09-25
DataGrid - AI Column: implement .d.ts

pomahtri:


packages/devextreme/js/common/grids.d.ts
```diff
@@ -1373,6 +1498,36 @@ export type HeaderFilterTexts = {
...
+   * @publicName clearAIColumn(columnName)
+   * @public
+   */
+  clearAIColumn(columnName: string): void;
+  /**
+   * @docid
+   * @publicName getAIColumnText(key)
+   * @public
+   */
+  getAIColumnText(key: TKey): string;
```
Raushen: There are should be two args: columnName and key. Sorry if API changed and I did not notify you.

packages/devextreme/js/common/grids.d.ts
```diff
@@ -70,6 +73,61 @@ import {
...
+   */
+  additionalInfo?: Record<string, any>;
+  /**
+   * @docid
+   */
+  useCache?: boolean;
+  /**
+   * @docid
+   */
+  onAIColumnResponseReceived?: (e: AIColumnResponseReceivedInfo) => void;
```
Raushen: Could you please move it on the Grid root level as the onAIColumnRequestCreating?

pomahtri: Oh, when reading the spec I thought like it's inside `e` of `onAIColumnRequestCreating` and like it is set per-request

packages/devextreme/js/common/grids.d.ts
pomahtri: Oh, nope, my bad

Raushen (dismissed):

mpreyskurantov (approved):

Raushen (approved):

PR #31169, opened 2025-09-26
Form Smart Paste: move parsing to AiIntegration and extend sendRequest parameters

pharret31:


packages/devextreme/js/__internal/ui/form/form.ai.utils.ts
```diff
@@ -29,66 +26,29 @@ const getEditorTypeInfo = (editorType: FormItemComponent | undefined): string =>
...
-        || value.length > 2
-        || value.some((item) => isNaN(parseFloat(item)))
-      ) {
-        throw errors.Error('E1064', dataField, errorValue, 'number range');
-      }
-      return value;
+      return 'numberRange';
+    case 'dxColorBox':
+      return 'color';
     case 'dxHtmlEditor':
```
anna-shakhova: excessive case, already handled in default

packages/devextreme/js/__internal/ui/form/form.ai.utils.ts
```diff
@@ -1,11 +1,8 @@
-import Color from '@js/color';
-import type { SmartPasteCommandResult } from '@js/common/ai-integration';
+import type {
```
anna-shakhova, line 1: looks like all utils should be common

packages/devextreme/js/__internal/ui/form/form.ts
```diff
@@ -1984,6 +1975,7 @@ class Form extends Widget<FormProperties> {
     const fields = dataItems.map((item) => ({
       name: item.dataField,
       format: getItemFormatInfo(item),
+      type: getFieldType(item.editorType),
```
anna-shakhova, line 1981: type is not enough for prompt building, for editors with values list we also require this list and information whether custom values are allowed
after adding values list `format` is excessive as it is duplicated by other fields

mpreyskurantov (approved):

anna-shakhova (approved):

PR #31170, opened 2025-09-26
dxFileManager: fix 'First item is selected when right-clicking on an empty area' (T1307139)

miherlosev:


e2e/testcafe-devextreme/tests/fileManager/common.ts
```diff
@@ -24,3 +24,30 @@ test('Custom DropDown width for Material and Fluent themes', async (t) => {
   fileSystemProvider: [],
   height: 450,
 }));
+
+test('Thumbnails view - Should not focus the first item on right click (T1307139)', async (t) => {
```
Staziz: I'd suggest writing this as a qunit test, similar to other fileManager tests. They're essentially the same e2e tests, just not written in TestCafe. And TestCafe tests, if I understand correctly, are mostly screenshot tests.

packages/devextreme/js/ui/file_manager/ui.file_manager.item_list.thumbnails.js
```diff
@@ -140,6 +154,12 @@ class FileManagerThumbnailsItemList extends FileManagerItemListBase {
         });
     }
 
+    _onItemListFocusedElement(args) {
+        // Do not focus the element when the right mouse button is pressed.
+        args.shouldFocusElement = this._lastMouseButton !== 3;
```
Staziz: The behavior is also reproducible with the LMB click

Staziz (commented):
In single selection mode, the component should synchronize the selection and focus state. This is the original intention that we've put into it. Perhaps it's a bit unusual, but nevertherless. 
And I believe we need to mimic the details view behavior in the single selection mode, namely, not focusing any items during `_focusInHandler`. (This is a rough estimate that requires verification). The focus frame should appear only after an element is selected, not when we simply focus the thumbnails view with a mouse click or context menu

Staziz (approved):

PR #31172, opened 2025-09-26
Scheduler: AllDay panel cells are misaligned with the main table on MacOS if the scrollbar is visible by default

sjbur:


packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
```diff
@@ -909,6 +909,38 @@ class SchedulerWorkSpace extends Widget<WorkspaceOptionsInternal> {
...
+      this._$headerPanelContainer.css('paddingRight', `${scrollbarWidth}px`);
+    }
+  }
+
+  _getScrollbarWidth() {
+    if (!this._dateTableScrollable) {
+      return 0;
+    }
+
+    const $container = this._dateTableScrollable.$element().find('.dx-scrollable-container');
```
Tucchhaa: Scrollable has container() function: https://github.com/DevExpress/DevExtreme/blob/45525faa3bd834615ab7406c114121374fce698e/packages/devextreme/js/__internal/ui/scroll_view/scrollable.ts#L471

Let's use it here, because it's better than finding elements in DOM

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
```diff
@@ -909,6 +909,38 @@ class SchedulerWorkSpace extends Widget<WorkspaceOptionsInternal> {
...
+    if (!$container.length) {
+      return 0;
+    }
+
+    const containerElement = $container.get(0);
+    if (!containerElement) {
+      return 0;
+    }
+
+    const hasVerticalScrollbar = containerElement.scrollHeight > containerElement.clientHeight;
```
Tucchhaa: Also Scrollable has `scrollHeight` and `clientHeight` methods, that we could possibly use here. And so we can make `_getScrollbarWidth` function to be very small 

https://github.com/DevExpress/DevExtreme/blob/45525faa3bd834615ab7406c114121374fce698e/packages/devextreme/js/__internal/ui/scroll_view/scrollable.ts#L493

Tucchhaa: If it wouldn't be too hard, I would suggest to add Scrollable type to `_dateTableScrollable`

e2e/testcafe-devextreme/tests/scheduler/common/layout/horizontalScroll/horizontalScroll.ts
```diff
@@ -0,0 +1,169 @@
...
+
+const scrollHorizontallyToEnd = ClientFunction(() => {
+  const scheduler = document.querySelector('.dx-scheduler');
+  if (!scheduler) return { success: false, message: 'Scheduler not found' };
+
+  const dateTableScrollable = scheduler.querySelector('.dx-scheduler-date-table-scrollable');
+  if (!dateTableScrollable) return { success: false, message: 'Date table scrollable not found' };
+
+  const scrollableContainer = dateTableScrollable.querySelector('.dx-scrollable-container');
+  if (!scrollableContainer) return { success: false, message: 'Scrollable container not found' };
```
Tucchhaa, line 104: I doubt that we need such checks for elements. Because if scheduer, dateTableScrollable or scrollableContainer are not present on the page, that means that scheduler doesn't render and that will lead to many other tests fail.

e2e/testcafe-devextreme/tests/scheduler/common/layout/horizontalScroll/horizontalScroll.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  await scheduler.repaint();
+
+  await t.wait(100);
+
+  await t.expect(
+    await takeScreenshot('scheduler-horizontal-scroll-before.png', scheduler.element),
+  ).ok();
+
+  const scrollResult = await scrollHorizontallyToEnd();
```
Tucchhaa: Scheduler POM already has dateTableScrollable: https://github.com/DevExpress/DevExtreme/blob/45525faa3bd834615ab7406c114121374fce698e/packages/testcafe-models/scheduler/index.ts#L65.

Can we scroll scheduler without ClientFunction? Like here: https://github.com/DevExpress/DevExtreme/blob/45525faa3bd834615ab7406c114121374fce698e/e2e/testcafe-devextreme/tests/scheduler/common/grouping/smoothCellLines.ts#L19?

I think it's not necessary to calculate `maxScrollLeft`, we can just set some big enough value for scrollimg.

e2e/testcafe-devextreme/tests/scheduler/common/layout/horizontalScroll/horizontalScroll.ts
```diff
@@ -0,0 +1,169 @@
...
+  await t
+    .expect(scrollResult.success).ok(`Scroll failed: ${scrollResult.message}`)
+    .expect(scrollResult.hasHorizontalScroll).ok('No horizontal scroll available')
+    .expect(scrollResult.finalScrollLeft)
+    .gt(0, 'Scroll position should be greater than 0');
+
+  await t.wait(500);
+
+  await t.expect(
+    await takeScreenshot('scheduler-horizontal-scroll-after.png', scheduler.element),
```
Tucchhaa: Can we rename screenshot name according to the testing guidelines that Alyar has created recently on the wiki?

e2e/testcafe-devextreme/tests/scheduler/common/layout/horizontalScroll/horizontalScroll.ts
```diff
@@ -0,0 +1,169 @@
...
+  return {
+    success: true,
+    initialScrollLeft,
+    maxScrollLeft,
+    finalScrollLeft: scrollableContainer.scrollLeft,
+    hasHorizontalScroll: maxScrollLeft > 0,
+  };
+});
+
+test('Scheduler horizontal scroll screenshot test', async (t) => {
```
Tucchhaa: Let's give this test more meaningful name, I think current name is very generic. Also we can mention the ticket number in the test name

e2e/testcafe-devextreme/tests/scheduler/common/layout/horizontalScroll/horizontalScroll.ts
```diff
@@ -0,0 +1,169 @@
+import { createScreenshotsComparer } from 'devextreme-screenshot-comparer';
```
Tucchhaa: I suggest to move this test to [this file](https://github.com/DevExpress/DevExtreme/blob/45525faa3bd834615ab7406c114121374fce698e/e2e/testcafe-devextreme/tests/scheduler/common/header/header.ts#L46), because we add `padding-right` for scheduler's header

aleksei-semikozov (approved):

PR #31201, opened 2025-10-01
Splitter: prevent exceeding maximum size on expanding (T1308778)

pharret31:


packages/devextreme/testing/tests/DevExpress.ui.widgets/splitter.tests.js
```diff
@@ -882,6 +882,15 @@ QUnit.module('Pane sizing', moduleConfig, () => {
                 { targetButton: 'prev', resizeHandleIndex: 0, expectedLayout: ['10.1626', '44.9187', '44.9187'] },
                 { targetButton: 'next', resizeHandleIndex: 0, expectedLayout: ['27.5407', '27.5407', '44.9187'] },
             ]
+        },
+        {
+            items: [ { collapsible: true, collapsedSize: 50, size: 100, minSize: 100, maxSize: 100 }, { collapsible: true, size: 100 }],
```
EugeniyKiyashko, line 887: I would prefer to add extra configuration with three panes, like the sample in the original issue, to be sure that distribution between panes works correctly

packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -844,7 +844,7 @@ class Splitter extends CollectionWidgetLiveUpdate<Properties> {
     );
 
     this._itemRestrictions.forEach((pane) => {
-      pane.maxSize = undefined;
+      if (item.collapsed) pane.maxSize = undefined;
```
EugeniyKiyashko:
```suggestion
      if (item.collapsed) {
          pane.maxSize = undefined;
      }
```

EugeniyKiyashko: We reset maxSize in two places, including _itemVisibleOptionChanged. Should this change be made there as well? Kindly, check it

EugeniyKiyashko (approved):

packages/devextreme/js/__internal/ui/splitter/splitter.ts
pharret31: When visible of item is false it doesn't have collapsed size so it won't be added to the delta that is taken from cache so it works well without such adjusting

PR #31093, opened 2025-09-17
DataGrid - Column reordering doesn't work when there are many columns and columnRenderingMode is 'virtual' (T1304328, T1305380)

pomahtri:

PR #31256, opened 2025-10-08
DataGrid - Column reordering doesn't work when there are many columns and columnRenderingMode is 'virtual' (T1304328, T1305380) 

pomahtri:
Cherry-pick of https://github.com/DevExpress/DevExtreme/pull/31093


Alyar666 (approved):

pomahtri:
Both issues doesn't occur in 24.2

PR #31275, opened 2025-10-08
✨AI Column: Methods and handlers

Raushen:


dmlvr (dismissed):

vladaskorohodova:
lgtm

AlisherAmonulloev:
LGTM

marker-dao:
Lgtm

pomahtri (dismissed):

packages/devextreme/js/__internal/core/ai_integration/commands/generateGridColumn.ts
```diff
@@ -20,9 +20,24 @@ export class GenerateGridColumnCommand extends BaseCommand<
...
+      };
+    }
+
+    const data = typeof response.data === 'string'
+      ? JSON.parse(response.data)
+      : response.data;
+
+    return {
+      data,
+      additionalInfo: response.additionalInfo,
```
Alyar666, line 39: Shouldn't 'response.additionalInfo' be parsed too?

Alyar666 (changes requested):

packages/devextreme/js/__internal/core/ai_integration/commands/generateGridColumn.ts
Raushen: We expected this field will be read-to-use in the onAIColumnResponseReceived handler as its value came not from AI, but from an endpoint. So yes, it was planned it is already dictionary.
However we gonna to avoid using this handler and prefer do all necessary operations on the aiIntegration.sendRequest level. In the future there will be a task removing this field.

Alyar666 (approved):

mpreyskurantov (approved):

PR #31279, opened 2025-10-09
TreeView: support selection by shift+arrows/home/end when showCheckBoxesMode='none'

pharret31:


packages/devextreme/testing/tests/DevExpress.ui.widgets/treeViewParts/keyboardNavigation.js
```diff
@@ -185,18 +200,33 @@ QUnit.test('shift+end key pressing extends selection up to the last node', funct
...
+
+            const selectedNodes = $treeView.dxTreeView('instance').getSelectedNodes();
+
+            if(selectByClick) {
+                assert.strictEqual(
+                    selectedNodes.length,
+                    selectionMode === 'multiple' ? DATA[0].length : 1
+                );
+            } else {
+                assert.deepEqual(selectedNodes, []);
```
ksercs: `if/else` for asserts in tests is a bad practice because it complicated reading although tests should be as simple as possible. If you need it in means that matrix iteration is not a good fit for this case and it's better to split test.

I recommend you create a module like `kbn when showCheckBoxesMode="none"` and solve copypaste problem by moving initialization to `beforeEach`. Additional good point that these tests will be aggregated together

packages/devextreme/testing/tests/DevExpress.ui.widgets/treeViewParts/keyboardNavigation.js
```diff
@@ -827,3 +826,82 @@ QUnit.testInActiveWindow('First list item should be focused on the tab key press
...
+        };
+    },
+    afterEach: function() {
+        this.clock.restore();
+        fx.off = false;
+    }
+});
+
+['single', 'multiple'].forEach(function(selectionMode) {
+    QUnit.test(`shift+home key pressing, selectionMode=${selectionMode}`, function(assert) {
```
ksercs: Let's rearrange tests to be inside the module

packages/devextreme/testing/tests/DevExpress.ui.widgets/treeViewParts/keyboardNavigation.js
```diff
@@ -827,3 +826,82 @@ QUnit.testInActiveWindow('First list item should be focused on the tab key press
...
+        const $treeView = this.createTreeView(selectionMode);
+        const $item = $treeView.find('.' + internals.ITEM_CLASS).eq(-1);
```
ksercs:
```suggestion
        const $item = $treeView.find(`.${internals.ITEM_CLASS}`).eq(-1);
```

packages/devextreme/testing/tests/DevExpress.ui.widgets/treeViewParts/keyboardNavigation.js
```diff
@@ -827,3 +826,82 @@ QUnit.testInActiveWindow('First list item should be focused on the tab key press
...
+        const selectedNodes = $treeView.dxTreeView('instance').getSelectedNodes();
+
+        assert.strictEqual(
+            selectedNodes.length,
+            selectionMode === 'multiple' ? DATA[0].length : 1
+        );
+    });
+});
+
+QUnit.test('shiftDown key test', function(assert) {
```
ksercs: Let's make all these tests naming more meaningul? "key press should ..."

packages/devextreme/testing/tests/DevExpress.ui.widgets/treeViewParts/keyboardNavigation.js
```diff
@@ -827,3 +826,82 @@ QUnit.testInActiveWindow('First list item should be focused on the tab key press
...
+            });
+        };
+    },
+    afterEach: function() {
+        this.clock.restore();
+        fx.off = false;
+    }
+});
+
+['single', 'multiple'].forEach(function(selectionMode) {
```
ksercs: why tests below are in a separate duplicated forEach?

ksercs (approved):

PR #31287, opened 2025-10-09
CI optimization: optimized TestCafe accessibility tests (Editors)

juleari:


e2e/testcafe-devextreme/tests/accessibility/button.ts
```diff
@@ -12,7 +12,6 @@ fixture.disablePageReloads`Accessibility`
 const optionsWithSimpleItems: Options<Properties> = {
   useSubmitBehavior: [true, false],
   disabled: [true, false],
-  hint: [undefined, 'hint'],
```
pharret31, line 15: Don't we need to test hint attr?

marker-dao: I am sure that there can be no accessibility issues here, so there is no need to test it.

marker-dao (approved):

r-farkhutdinov (approved):

pharret31 (approved):

EugeniyKiyashko (approved):

PR #31297, opened 2025-10-10
CI optimization: reduce the number of scenarios in TestCafe testing (Editors)

juleari:


e2e/testcafe-devextreme/runner.ts
anna-shakhova, line 1: Non-related change from other PR

PR #31356, opened 2025-10-15
CI optimization: move editors related tests to editors folder

pharret31:


e2e/testcafe-devextreme/tests/chat/data/index.ts
```diff
@@ -29,16 +29,10 @@ export const getLongText = (useLineBreaks = false, length = 1): string => {
   const longItemArray = Array(length).fill(longItem);
   const longString = longItemArray.join('');
 
-  const result = `${UUID}:${useLineBreaks ? lineBreaks : ''}${longString}`;
```
marker-dao, line 32: ### 💡 minor(non-blocking)
We use the same style of code writing as before. But here it is not critical)

marker-dao (approved):

e2e/testcafe-devextreme/tests/chat/data/index.ts
pharret31: My linter just asked me to fix that but okay, I'll remember that )

r-farkhutdinov (approved):

PR #31375, opened 2025-10-16
Caret Utils: Remove body contains input check (T1309092)

marker-dao:


packages/devextreme/testing/tests/DevExpress.utils/utils.caret.tests.js
```diff
@@ -28,15 +28,39 @@ testModule('caret', () => {
...
+            set: function() {
+                setterStub();
+                throw new Error('Cannot set selection');
+            },
+        };
+
+        Object.defineProperty(HTMLInputElement.prototype, 'selectionStart', { ...originalDescriptors.selectionStart, ...throwingDescriptor });
+        Object.defineProperty(HTMLInputElement.prototype, 'selectionEnd', { ...originalDescriptors.selectionEnd, ...throwingDescriptor });
+
         const input = document.createElement('input');
```
ksercs: The test looks overcomplicated to me.
- Why do we need to touch global HTMLInputElement descriptors? Looks like it's enough to overload this `input` descriptions and that's it - in this case we also don't need to retrieve overload back
- Why do we need to stub `getActiveElement`? Caret util works with the passed `input` parameter => looks like it's enough to stub its descriptors

I didn't test but guess smth like this should work the same way
```javascript
    const input = document.createElement('input');
    Object.defineProperty(input, 'selectionStart', {
        set() { throw new Error('Cannot set selection'); }
    });
    try {
        caret(input, { start: 1, end: 2 });
        assert.ok(true, 'no exception is thrown for detached input');
    } catch (e) {
        assert.ok(false, 'exception was thrown');
    }
```

r-farkhutdinov (dismissed):

packages/devextreme/testing/tests/DevExpress.utils/utils.caret.tests.js
marker-dao: Fixed [30e73e61adf28557ebc8957bbb9a7a3730ab500a]

> Why do we need to stub getActiveElement?

The previous test is always green because the condition in the caret function `domAdapter.getActiveElement(input) !== input` is always true. Code: https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/__internal/ui/text_box/m_utils.caret.ts#L55C44-L55C88. And we don't enter the setCaret function at all. But in those cases where there were bugs, we did enter it.

packages/devextreme/testing/tests/DevExpress.utils/utils.caret.tests.js
```diff
@@ -80,9 +93,9 @@ testModule('caret', () => {
             assert.ok(true, 'exception is not thrown');
         } catch(e) {
             assert.ok(false, 'exception is thrown');
+        } finally {
+            Object.defineProperty(HTMLInputElement.prototype, 'selectionStart', initialDescriptor);
```
ksercs: forgot to remove

marker-dao: Refactored: [b6edefb13df45d4e2c427d233c9a4a9a7da81058]

r-farkhutdinov (approved):

PR #31399, opened 2025-10-20
Scheduler: Form customization

aleksei-semikozov:

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
@@ -3,6 +3,7 @@ import {
 } from '@jest/globals';
 import $ from '@js/core/renderer';
 import type dxDateBox from '@js/ui/date_box';
+import type { GroupItem, Item as FormItem } from '@js/ui/form';
 import { toMilliseconds } from '@ts/utils/toMilliseconds';
 
 import fx from '../../../common/core/animation/fx';
@@ -1444,4 +1445,339 @@ describe('Timezone Editors', () => {
   it.todo('timeZone editors should have correct options');
   it.todo('timeZone editor should have correct display value for timezones with different offsets');
   it.todo('dataSource of timezoneEditor should be filtered');
+});
+
+describe('Customize form items', () => {
+  beforeEach(() => {
+    fx.off = true;
+    setupSchedulerTestEnvironment();
+  });
+
+  afterEach(() => {
+    fx.off = false;
+    document.body.innerHTML = '';
+    jest.useRealTimers();
+  });
+
+  describe('Basic form customization', () => {
+    it('should use default form when editing.items is not set', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+
+      expect(formItems).toBeDefined();
+      expect(formItems?.length).toBeGreaterThan(0);
+    });
+
+    it('should show empty form when editing.items is empty array', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: [],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+
+      expect(formItems?.length ?? 0).toBe(0);
+    });
+
+    it('should show mainGroup when specified in string array', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: ['mainGroup'],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+
+      expect(formItems?.length).toBe(1);
+      expect(formItems?.[0]?.name).toBe('mainGroup');
+    });
+
+    it('should hide group when visible is false', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: [{ name: 'mainGroup', visible: false }],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+
+      expect(formItems?.length).toBe(1);
+      expect(formItems?.[0]?.visible).toBe(false);
+    });
+
+    it('should show group when visible is true', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: [{ name: 'mainGroup', visible: true }],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+
+      expect(formItems?.length).toBe(1);
+      expect(formItems?.[0]?.visible).toBe(true);
+    });
+
+    it('should filter children when items array is specified', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: [{
+              name: 'mainGroup',
+              visible: true,
+              items: ['subjectGroup'],
+            }],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      const mainGroup = formItems?.[0] as GroupItem;
+
+      expect(formItems?.length).toBe(1);
+      expect(mainGroup?.items?.length).toBe(1);
+      expect(mainGroup?.items?.[0]?.name).toBe('subjectGroup');
+    });
+
+    it('should handle non-existent groups gracefully', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: ['nonExistentGroup'],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+
+      expect(formItems?.length ?? 0).toBe(1);
+    });
+  });
+
+  describe('Form customization with editing.items', () => {
+    it('should handle empty items array', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: [],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      expect(formItems?.length).toBe(0);
+    });
+
+    it('should handle string array configuration', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: ['mainGroup'],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      expect(formItems?.length).toBe(1);
+      expect((formItems?.[0] as GroupItem)?.name).toBe('mainGroup');
+    });
+
+    it('should handle object configuration with visible false', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: [{ name: 'mainGroup', visible: false }],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      expect(formItems?.length).toBe(1);
+      expect(formItems?.[0]?.visible).toBe(false);
+    });
+
+    it('should handle object configuration with custom items', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: [{
+              name: 'mainGroup',
+              items: ['subjectGroup', 'dateGroup'],
+            }],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      const mainGroup = formItems?.[0] as GroupItem;
+      expect(mainGroup?.items?.length).toBe(2);
+      expect((mainGroup?.items?.[0] as GroupItem)?.name).toBe('subjectGroup');
+      expect((mainGroup?.items?.[1] as GroupItem)?.name).toBe('dateGroup');
+    });
+
+    it('should handle non-existent group names', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: ['nonExistentGroup'],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      expect(formItems?.length).toBe(1);
+    });
+
+    it('should handle undefined items', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: undefined,
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      expect(formItems?.length).toBeGreaterThan(0);
+    });
+
+    it('should handle mixed configurations', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: [
+              'mainGroup',
+              { name: 'mainGroup', visible: false },
+            ],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      expect(formItems?.length).toBe(2);
+      expect((formItems?.[0] as any)?.name).toBe('mainGroup');
+      expect((formItems?.[1] as any)?.name).toBe('mainGroup');
+      expect(formItems?.[1]?.visible).toBe(false);
+    });
+
+    it('should handle empty items array in object config', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          form: {
+            items: [{
+              name: 'mainGroup',
+              items: [],
+            }],
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      const mainGroup = formItems?.[0] as any;
+      expect(mainGroup?.items?.length).toBe(0);
+    });
+  });
 });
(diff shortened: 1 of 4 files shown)


packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -192,7 +195,10 @@ export class AppointmentForm {
     this.setStylingModeToEditors(mainGroup, showMainGroupIcons);
     this.setStylingModeToEditors(recurrenceGroup, showRecurrenceGroupIcons);
 
-    this.createForm(items);
+    const editingConfig = this.scheduler.getEditingConfig();
+    const handledItems = customizeFormItems(items, editingConfig?.form?.items);
```
Tucchhaa: minor: I think `customizedItems` is better name for the variable

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -280,6 +286,7 @@ export class AppointmentForm {
     const { textExpr } = this.scheduler.getDataAccessors().expr;
 
     return {
+      name: 'subjectGroup',
```
Tucchhaa: I suggest to move this string to constant variables like all other group names

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -288,11 +295,13 @@ export class AppointmentForm {
       },
       items: [
         {
+          name: 'subjectIcon',
           colSpan: 1,
           cssClass: CLASSES.formIcon,
           template: createFormIconTemplate('isnotblank'),
         },
         {
+          name: 'subject',
```
Tucchhaa: I suggest to move `name: 'subjectIcon'` and `name: 'subject'` to constant variables. like:
```js
const SUBJECT_ICON_NAME = 'subjectIcon';

const SUBJECT_EDITOR_NAME = 'subject';
```

Tucchhaa: Can you apply this to all other names in the code please? (like timezone editor's name, etc.)

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -877,10 +903,10 @@ export class AppointmentForm {
...
 
     this.dxForm.beginUpdate();
-    this.dxForm.itemOption(`${MAIN_GROUP_NAME}.${EDITOR_NAMES.startDate}`, 'colSpan', visible ? 1 : 2);
-    this.dxForm.itemOption(`${MAIN_GROUP_NAME}.${EDITOR_NAMES.startTime}`, 'visible', visible);
-    this.dxForm.itemOption(`${MAIN_GROUP_NAME}.${EDITOR_NAMES.endDate}`, 'colSpan', visible ? 1 : 2);
-    this.dxForm.itemOption(`${MAIN_GROUP_NAME}.${EDITOR_NAMES.endTime}`, 'visible', visible);
+    this.dxForm.itemOption(`${MAIN_GROUP_NAME}.${DATE_GROUP_NAME}.${START_DATE_GROUP_NAME}.${START_DATE_EDITOR_NAME}`, 'colSpan', visible ? 1 : 2);
+    this.dxForm.itemOption(`${MAIN_GROUP_NAME}.${DATE_GROUP_NAME}.${START_DATE_GROUP_NAME}.${START_TIME_EDITOR_NAME}`, 'visible', visible);
+    this.dxForm.itemOption(`${MAIN_GROUP_NAME}.${DATE_GROUP_NAME}.${END_DATE_GROUP_NAME}.${END_DATE_EDITOR_NAME}`, 'colSpan', visible ? 1 : 2);
+    this.dxForm.itemOption(`${MAIN_GROUP_NAME}.${DATE_GROUP_NAME}.${END_DATE_GROUP_NAME}.${END_TIME_EDITOR_NAME}`, 'visible', visible);
```
Tucchhaa: Minor: That's a very long string here :) I suggest to create a variable for editor's name, e.g.:

```js
const startDateItemName = `${MAIN_GROUP_NAME}.${DATE_GROUP_NAME}.${START_DATE_GROUP_NAME}.${START_DATE_EDITOR_NAME}`;

this.dxForm.itemOption(startDateItemName, 'colSpan', visible ? 1 : 2);
```

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -257,7 +258,7 @@ describe('Appointment Popup Form', () => {
       expect(POM.popup.getSwitchByName('allDay').value).toBe('false');
     });
 
-    it('should hide time editors when switched on', async () => {
+    it.skip('should hide time editors when switched on', async () => {
```
Tucchhaa: I checked locally, it seems that this test passes. Should we unskip it?

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -275,7 +276,7 @@ describe('Appointment Popup Form', () => {
       expect(POM.popup.endTime).toBeNull();
     });
 
-    it('should show time editors when switched off', async () => {
+    it.skip('should show time editors when switched off', async () => {
```
Tucchhaa: I checked locally, it seems that this test passes. Should we unskip it?

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
```diff
@@ -0,0 +1,115 @@
...
+): void => {
+  Array.from(itemsMap.values()).forEach((group) => {
+    if (isGroupItem(group) && group.items) {
+      group.items = [...group.items.filter((item) => item.name !== itemName)];
+    }
+  });
+};
+
+const processGroupItems = (newItem: FormItem, itemsMap: Map<string, FormItem>): FormItem => (
+  isGroupItem(newItem) && newItem.items
```
Tucchhaa: Probably we can replace this line by
```js
isGroupItem(newItem)
```

because isGroupItem() already checks that newItems.items is defined

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
```diff
@@ -0,0 +1,115 @@
...
+  },
+  map,
+);
+
+const removeItemFromGroups = (
+  itemName: string,
+  itemsMap: Map<string, FormItem>,
+): void => {
+  Array.from(itemsMap.values()).forEach((group) => {
+    if (isGroupItem(group) && group.items) {
```
Tucchhaa, line 41: Probably we can replace this line by
```js
if(isGroupItem(newItem)) {
```

because isGroupItem() already checks that newItems.items is defined

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
```diff
@@ -0,0 +1,115 @@
...
+  configItem: ConfigItem,
+  itemsMap: Map<string, FormItem>,
+): FormItem => {
+  if (typeof configItem !== 'object') {
+    return targetItem;
+  }
+
+  const merged = { ...targetItem, ...configItem };
+
+  return isGroupItem(configItem) && configItem.items && isGroupItem(merged)
```
Tucchhaa: Don't we just need to check that isGroupItem(merged) is true here?

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
```diff
@@ -0,0 +1,115 @@
...
+const mergeConfigIntoItem = (
+  targetItem: FormItem,
+  configItem: ConfigItem,
+  itemsMap: Map<string, FormItem>,
+): FormItem => {
+  if (typeof configItem !== 'object') {
+    return targetItem;
+  }
+
+  const merged = { ...targetItem, ...configItem };
```
Tucchhaa: If configItem is:
```js

{
   // ...
   editorOptions: {
      placeholder: 'custom placeholder'
   }
}
```

then targetItem.editorOptions would be completely overriden. I suggest to extend item like this:
```js
extend(true, {}, targetItem, configItem)
```
this will do a deep merge of both items into a new object

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  const map = itemsMap ?? buildFormItemsMap(items);
+
+  return userConfig
+    .map((option) => applySingleCustomization(map, option))
+    .filter((item): item is FormItem => item !== undefined);
+};
+
+export { customizeFormItems };
+export type { ConfigItem, FormItem };
```
Tucchhaa, line 96: While reading through this file, I noticed that several functions call each other recursively:
customizeFormItems -> 
applySingleCustomization -> 
processGroupItems -> 
customizeFormItems

and

customizeFormItems -> 
applySingleCustomization -> 
mergeConfigIntoItem -> 
customizeFormItems

It was a hard time for me to trace the flow.

Moreover `itemsMap` is passed across all functions, which implies to me that probably we need to wrap these functions into a class.

So, I was wondering if we could explore an alternative structure. For example, wrapping the logic into a class and handling traversal in a single DFS-style function.

I have taken your approach as foundation and prepared a draft. Please share your opinion about this 🙂

```ts
class Customizer {
  constructor(defaultItems) {
    this.defaultItems = defaultItems;
    this.defaultItemsMap = this.buildItems(defaultItems);
  }

  customize(userItems) {
    const result = [];

    // depth-first search
    userItems.forEach(userItem => {
      const children = getChildren(userItem);

      if(children) {
        userItem.items = customize(userItem.items);
      }

      const itemName = typeof userItem === 'string' ? userItem.name : itemName;
      const defaultItem = this.defaultItemsMap.get(itemName);
      const shouldExtend = isDefined(itemName) && defaultItem

      if(shouldExtend) {
        result.push(extend(true, defaultItem, userItem));
        this.defaultItemsMap.delete(itemName);
      } else {
        result.push(typeof userItem === 'string'
          ? this.createBasicItem(itemName)
          : userItem
        );
      }
    });

    return result;
  }

  private buildItems() { ... }

  private getChildren() { ... }

  private createBasicItem(name: string) { ... }
}
```

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  const existingItem = itemsMap.get(itemName);
+
+  if (!existingItem) {
+    const newItem = createFormItemFromConfig(configItem);
+    return processGroupItems(newItem, itemsMap);
+  }
+
+  removeItemFromGroups(itemName, itemsMap);
+  return mergeConfigIntoItem({ ...existingItem }, configItem, itemsMap);
```
Tucchhaa: Maybe we can move ```{ ...existingItem }``` inside the `mergeConfigIntoItem` function?

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
Tucchhaa: P.S. this approach may not be very accurate, I've prepared it just to show the idea

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
aleksei-semikozov: isGroupItem is type guard, return  `items?` prop type, We should check items again

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
aleksei-semikozov: isGroupItem is type guard, return `items?` prop type, We should check items again

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1009,3 +1010,311 @@ describe('Timezone Editors', () => {
...
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+
+      expect(formItems?.length).toBe(1);
+      expect(formItems?.[0] && 'name' in formItems[0] ? formItems[0].name : null).toBe('mainGroup');
```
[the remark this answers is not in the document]
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1009,3 +1010,311 @@ describe('Timezone Editors', () => {
...
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
+      const { form } = POM.popup;
+      const formItems = form.option('items') as FormItem[];
+      const mainGroup = formItems?.[0] as GroupItem;
+
+      expect(formItems?.length).toBe(1);
+      expect(mainGroup?.items?.length).toBe(1);
+      expect(mainGroup?.items?.[0] && 'name' in mainGroup.items[0] ? mainGroup.items[0].name : null).toBe('subjectGroup');
```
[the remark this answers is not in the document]
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
aleksei-semikozov: done. We are should check `isGroupItem(configItem) && isGroupItem(merged)`

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
aleksei-semikozov: Thank you for the proposed class-based approach! I've carefully reviewed it, but I'd like to explain why the current implementation might be preferable:
 **Recursion is natural and shallow here**
- We're processing a tree structure (forms), where recursion is the natural pattern
- Recursion depth is limited by the form structure (typically 2-3 levels max)
**Advantages of the current functional approach:**

**Simplicity of use:**
```
// Current approach - simple function call
const result = customizeFormItems(defaultItems, userConfig);

// Proposed approach - instance creation + method call
const customizer = new Customizer(defaultItems);
const result = customizer.customize(userConfig);
```
**Classes aren't needed for pure functions:**
Classes are justified when you need to store instance state and methods that modify that state
In our case, this is a pure transformation function: input → output

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
aleksei-semikozov: Sure, I can do it 😄

I just don’t fully see the benefit in this case: these names are used only once. When there’s reuse — I agree and I extracted constants already. But here it feels like extra noise without added value.

If you still prefer it — I’ll apply it everywhere.

Tucchhaa: I see 3 points why would we need to move it constants:

1. No magic variables. Usually in the code we try to avoid putting plain strings. It's better to move this string to a constant to give it some name.

2. This point may apply only to me, but I see it would be useful to see all group, editor and icon names as constants at the top. This way it's very clear what kind of customizable entities the form has

3. Consistency. Right now 50% of the names are formatted as constants and other 50% as plain strings.It's a minor point, but that imo would definitely improve code style

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
Tucchhaa: `Recursion is natural and shallow here `

Yes, I know. The problem of current approach is that current  `customizeFormItems` **does not look recursive by the first glance**.  I had to deeply investigate this function to find out that the recursion consists of 3 functions (!) calling each other

DFS is a standard and a simple algorithm that traverses a graph (a tree in our case) that's perfect for our case. The point of the alternative approach I suggested is that by looking at `Customizer.customize` function it's very clear that the function is recursive.

Another flaw that I found of current approach is that we have to disable these eslint rules:
- https://github.com/DevExpress/DevExtreme/pull/31399/files#diff-4a2a869cda126d4b13a03719a772072ed5224504e9027d3a34422e9338015479R56
- https://github.com/DevExpress/DevExtreme/pull/31399/files#diff-4a2a869cda126d4b13a03719a772072ed5224504e9027d3a34422e9338015479R73

Yes, I know the code works, but what I mean is that eslint rules are there for a reason...


`Simplicity of use`
That's kinda right, but in my opinion that doesn't make much of a difference if have 1 more or less lines of code :)

Maybe I wasn't clear enough, but my main point was not to say "let's use classes", but was to show that there's a way to simplify current complexity of the code (multiple functions recursion, itemsMap passed everywhere). 

I was hoping that you would take a deeper look at what's going on inside `customize` function.

P.S. It's possible to do the same by using functions:

```js

const customizeFormItems = (defaultItems, customItems) => {
  const buildItems = () => { /* ... */ };
  const getChildren = (item) => { /* ... */ };
  const createBasicItem = (itemName) => { /* ... */ };

  const defaultItemsMap = buildItems(defaultItems);

  const customize = (customItems) => {
    const result = [];

    // depth-first search
    customItems.forEach(customItem => {
      const children = getChildren(customItem);

      if (children) {
        customItem.items = customize(customItem.items);
      }

      const itemName = typeof customItem === 'string' ? customItem.name : itemName;
      const defaultItem = defaultItemsMap.get(itemName);
      const shouldExtend = isDefined(itemName) && defaultItem;

      if (shouldExtend) {
        result.push(extend(true, defaultItem, customItem));
        defaultItemsMap.delete(itemName);
      } else {
        result.push(typeof customItem === 'string'
          ? createBasicItem(itemName)
          : customItem
        );
      }
    });

    return result;
  };

  return customize(customItems);
};

```

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
Tucchhaa: Another point that I have noticed is that you didn't change anything in m_recurrence_form.ts. Will recurrence form customization be implemented later?

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
```diff
@@ -0,0 +1,116 @@
...
+  map,
+);
+
+const removeItemFromGroups = (
+  itemName: string,
+  itemsMap: Map<string, FormItem>,
+): void => {
+  Array.from(itemsMap.values()).forEach((group) => {
+    if (isGroupItem(group) && group.items) {
+      group.items = [...group.items.filter((item) => item.name !== itemName)];
```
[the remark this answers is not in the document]
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/m_customize_form_items.ts
Tucchhaa: A little bit more staightforward approach:

```js
    const itemName = getItemName(customItem);
    const defaultItem = itemName ? defaultItemsMap.get(itemName) : undefined;

    if (defaultItem) {
      removeItemFromGroups(defaultItem.name as string, defaultItemsMap);

      return typeof customItem === 'string'
        ? defaultItem
        : extend(true, {}, defaultItem, customItem) as FormItem;
    }

    return createFormItemFromConfig(customItem);
```

Tucchhaa: Looks like we don't need to use spread operator here

Tucchhaa, line 29: Looks like we don't need to use spread operator here

Tucchhaa: `buildFormItemsMap` is called only for defaultItems, which means that we will never have tab items here.

So I suggest:
```js
if(isGroupItem(item)) {
  return buildFormItemsMap(item.items ?? [], accumulator)
}

return accumulator;
```

Tucchhaa: I don't see a point of wrapping `resolveItem` into `baseResolveItem` if all the `baseResolveItem` is that it puts `map` to a closure.

I suggest to do one of these:
1. remove baseResolveItem wrapper and pass `map` directly to resolveItem function call
2. move `baseResolveItem` and `removeItemFromGroups` inside `customizeFormItems`, so that they already have `map` is default closure. (like it's done [here](https://github.com/DevExpress/DevExtreme/pull/31399/files#r2488835722))

Tucchhaa: By the way this will not work if user has tabItem that uses a default item inside. But I suggest to postpone implementing this

Tucchhaa: Looks like we don't to check that `tabs` property does not present in item. just check for `items` should be enough

Tucchhaa, line 92: I suggest to use a consistent naming in this **file** (yes, not only this function), because the same entities may have different names throughout the file which may be very confusing. For example, currently we have:

`userConfig`, `config`, `configure` which represent a custom item from user.

or variable name `map` doesn't tell anything about the variable except it's type, which I can already see by typescript.

So my suggestion is to rename entities like this:
`map` -> `defaultItemsMap`
`configure` -> `customItems` or `userItems`
`existingItem` -> `defaultItem`
`config` -> `customItem` or `userItem`
`userConfig` -> `customItemsRoot` or `userItemsRoot`

aleksei-semikozov: done

Tucchhaa (approved):

PR #31422, opened 2025-10-22
FileUploader: implement _hideCancelButtonOnUpload, _showFileIcon, _cancelButtonPosition options and onCancelButtonClick event

Zedwag:


packages/devextreme/js/__internal/ui/file_uploader/file_uploader.ts
```diff
@@ -177,11 +180,20 @@ interface FileUploaderChunkUploadResponse {
   error?: string;
 }
 
+type CancelButtonClickEvent = NativeEventInfo<KeyboardEvent | MouseEvent | PointerEvent> & {
```
r-farkhutdinov: Can we use `InteractionEvent type` here? It's a new utility type added [here](https://github.com/DevExpress/DevExtreme/blob/25_2/packages/devextreme/js/events/events.types.d.ts#L119).

```suggestion
type CancelButtonClickEvent = NativeEventInfo<InteractionEvent> & {
```

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.ts
```diff
@@ -1399,17 +1426,38 @@ class FileUploader extends Editor<FileUploaderProperties> {
...
+    const { _showFileIcon } = this.option();
 
-    if ($uploadButton) {
-      $container.append($uploadButton);
+    if (!_showFileIcon) {
+      return;
     }
+
+    $('<div>')
+      .addClass(`${FILEUPLOADER_FILE_ICON_CLASS} dx-icon dx-icon-${getFileIconName(fileName)}`)
```
r-farkhutdinov: Let's move `dx-icon` to a constant too 
```suggestion
      .addClass(`${FILEUPLOADER_FILE_ICON_CLASS} ${ICON_CLASS} ${ICON_CLASS}-${getFileIconName(fileName)}`)
```

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.ts
```diff
@@ -177,11 +180,20 @@ interface FileUploaderChunkUploadResponse {
...
 
+type CancelButtonClickEvent = NativeEventInfo<KeyboardEvent | MouseEvent | PointerEvent> & {
+  readonly file?: File;
+};
 export interface Properties extends PublicProperties {
   _buttonStylingMode?: ButtonStyle;
 
   _uploadButtonType?: ButtonType;
 
+  _hideCancelButtonOnUpload?: boolean;
```
r-farkhutdinov, line 192: I would prefer to add `_showCancelButtonOnUpload` instead as: 
1. It's hidden by default, so currently we have to set a default value to `true` so that it doesn't change existing behavior. So `_showCancelButtonOnUpload` will not require a default value
2. In the condition below, we can remove negative check

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.utils.ts
```diff
@@ -0,0 +1,52 @@
+export const getFileIconName = (filename = ''): string => {
+  const lastDotIndex = filename.lastIndexOf('.');
```
r-farkhutdinov, line 42: We can just have something like:

```
const extension = filename.split('.').pop()?.toLowerCase();
```

No additional check is needed (if it's undefined, `file` will be returned as default)

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.utils.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  if (lastDotIndex === -1 || lastDotIndex === filename.length - 1) {
+    return 'file';
+  }
+
+  const extension = filename.slice(lastDotIndex + 1);
+
+  switch (extension) {
```
r-farkhutdinov: This switch can be refactored and I see two options that make all extensions to be stored separately for easy maintenance: 
1. Create a map like `EXTENSION_ICON_MAP`, where we will have `{ jpg: 'image', mp4: 'video' ... }` and then just `return EXTENSION_ICON_MAP[extension ?? ''] ?? 'file';`.
2. Create arrays like `EXTENSIONS_IMAGE = ['jpg', 'jpeg'...]` and 7 if statements with return.

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.utils.ts
Zedwag: initially i did it this way, but it doesn't suite for case when filename is smth like `mp3` without extention but will be considered as music

r-farkhutdinov: No, as it's split using `.`, or maybe you could provide such example?

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.ts
Zedwag: it's shown in both cases when uploading started, but then only in one case we hide it, in another we don't. So hide is correct here. And it's ok to have a default value as true

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.utils.ts
Zedwag: image

Zedwag: i expect file.png returns 'image', but just 'png' should return 'file'

r-farkhutdinov: I see, thank you

packages/devextreme-scss/scss/widgets/base/_fileUploader.scss
```diff
@@ -195,6 +195,13 @@
   font-size: 10px;
 }
 
+.dx-fileuploader-file-icon {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-right: 6px;
```
marker-dao, line 202: ### suggestion(minor)

Usually we use vars, but here up to you

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.ts
```diff
@@ -177,11 +181,20 @@ interface FileUploaderChunkUploadResponse {
   error?: string;
 }
 
+type CancelButtonClickEvent = NativeEventInfo<InteractionEvent> & {
+  readonly file?: File;
+};
```
marker-dao, line 185: ### chore(minor)
Add space

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.ts
```diff
@@ -813,6 +828,8 @@ class FileUploader extends Editor<FileUploaderProperties> {
 
   _beforeSendAction?: (event?: Record<string, unknown>) => void;
 
+  _cancelButtonClickAction?: (event?: Record<string, unknown>) => void;
```
marker-dao: ### question(non-blocking)
Why Record<>, not CancelButtonClickEvent?

```suggestion
  _cancelButtonClickAction?: (event?: Partical<CancelButtonClickEvent>) => void;
```

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.ts
```diff
@@ -17,17 +17,20 @@ import { extend } from '@js/core/utils/extend';
...
 import type { Properties as PublicProperties } from '@js/ui/file_uploader';
 import ProgressBar from '@js/ui/progress_bar';
 import { current, isFluent, isMaterial } from '@js/ui/themes';
 import { fromPromise } from '@ts/core/utils/m_deferred';
+import { ICON_CLASS } from '@ts/core/utils/m_icon';
 import type { OptionChanged } from '@ts/core/widget/types';
 import type { EditorProperties, UnresolvedEvents } from '@ts/ui/editor/editor';
 import Editor from '@ts/ui/editor/editor';
 
+import { getFileIconName } from './file_uploader.utils';
```
marker-dao:
```suggestion
import { getFileIconName } from '@ts/ui/file_uploader/file_uploader.utils';
```

packages/devextreme-scss/scss/widgets/base/_fileUploader.scss
Zedwag: it's more like temporar for now, higly likely will redefine it on chat lvl later

packages/devextreme/js/__internal/ui/file_uploader/file_uploader.ts
```diff
@@ -1265,6 +1287,10 @@ class FileUploader extends Editor<FileUploaderProperties> {
     this._dropZoneLeaveAction = this._createActionByOption('onDropZoneLeave');
   }
 
+  _createCancelButtonClickAction(): void {
+    this._cancelButtonClickAction = this._createActionByOption('onCancelButtonClick', { excludeValidators: ['readOnly'] });
```
marker-dao, line 1291: ### question(non-blocking)
Also ‘disabled’, not only ‘readOnly’?

marker-dao (approved):

r-farkhutdinov (approved):

PR #31458, opened 2025-10-24
fix: React two template roots regression (T1310108)

VasilyStrelyaev:


ajivanyandev (approved):
Creative solution! Looks good to me!

dxvladislavvolkov (approved):

PR #31491, opened 2025-10-28
Appointment Form Redesign: Change Popup title depending on editing or adding

sjbur:


packages/devextreme/js/localization/messages/en.json
```diff
@@ -274,7 +274,7 @@
         "dxScheduler-appointmentAriaLabel-recurring": "Recurring appointment",
 
         "dxScheduler-appointmentListAriaLabel": "Appointment list",
-
+        "dxScheduler-newAppointmentPopupTitle": "New Appointment",
```
arman-boyakhchyan: To keep the new variable naming consistent with the existing "editPopupTitle", I propose the following:
```suggestion
        "dxScheduler-newPopupTitle": "New Appointment",
```

sjbur: Applied this change.

Tucchhaa (dismissed):

PR #31497, opened 2025-10-28
Appointment Form Redesign: Demos: Update resource icons

sjbur:


apps/demos/package.json
```diff
@@ -93,6 +93,7 @@
         "@babel/eslint-parser": "catalog:",
         "@babel/preset-env": "7.23.9",
         "@babel/preset-react": "^7.26.3",
+        "@babel/preset-typescript": "^7.28.5",
```
Tucchhaa: why did you add this dependency?

apps/demos/Demos/Scheduler/Adaptability/Angular/app/app.component.html
```diff
@@ -13,6 +13,7 @@
     fieldExpr="priorityId"
     label="Priority"
     [dataSource]="priorities"
+    icon="tag"
```
Tucchhaa: Icon with name 'tag' doesn't exist. You should use icon with name 'tags'. 
Please don't forget to run demos locally and check that demos are working properly. 

It would be advisable for you to check all changed demos in all frameworks. It wouldn't take a lot of your time, but that will help to avoid mistakes like this

Tucchhaa: Wrong icon name is the reason why the screenshot test has failed. Please pay more attention to such details

Tucchhaa (changes requested):

apps/demos/package.json
sjbur: I was responsible for moving demos to typescript. In this PR I encountered a problem regarding this, demos linting failed because of typescript errors. By installing this depedency the problem is solved

apps/demos/Demos/Scheduler/Adaptability/Angular/app/app.component.html
Tucchhaa: This is the guide for launching demos locally: https://github.com/DevExpress/DevExtreme/blob/25_2/apps/demos/README.md#L1

apps/demos/package.json
Tucchhaa: Do you know what was the exact reason why linting was failing? 

I have found [the failing job](https://github.com/DevExpress/DevExtreme/actions/runs/18880455155/job/53882588891). It looks like it was something wrong with [Scheduler/VirtualScrolling/Angular/app/app.component.html](https://github.com/DevExpress/DevExtreme/pull/31497/files#diff-2fe759e07b418da1eb3a44d875cdcaa0d3ae42edfc2ef772aaaf726d373e280bR19) file. 

So demos linting was failing because of the changes of this PR?

sjbur: I made a separate PR without this change, where you can see a failing job: https://github.com/DevExpress/DevExtreme/actions/runs/18911905424/job/53985763307?pr=31513

If we try to remove the babel typescript plugin, the build still fails. By installing this depedency, this issue is resolved and builds fine.

Tucchhaa (approved):

PR #31506, opened 2025-10-29
Appointment Form Redesign: Common Layout changes after designer review

sjbur:


e2e/testcafe-devextreme/runner.ts
```diff
@@ -248,7 +248,7 @@ createTestCafe(TESTCAFE_CONFIG)
     }
 
     const runOptions: RunOptions = {
-      quarantineMode: { successThreshold: 1, attemptLimit: 2 },
```
Tucchhaa: do we need changes in this file?

sjbur: No, added this by a mistake.

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -473,6 +473,7 @@ describe('Appointment Popup Form', () => {
 
       expect(mainGroup.hasClass(CLASSES.mainGroupHidden)).toBe(false);
       expect(recurrenceGroup.hasClass(CLASSES.recurrenceGroupHidden)).toBe(true);
+      expect(recurrenceGroup.attr('tabindex')).toBe('-1');
```
Tucchhaa, line 477: Suggestion: Let's check both of the groups tabindex everywhere, like we check classes. Like:

```js
expect(mainGroup.attr('tabindex')).toBeUndefied('-1');
expect(recurrenceGroup.attr('tabindex')).toBe('-1');
```

also I think it's better to check `toBeUndefined` rather than `not.toBe('-1')`, because if an element has tabindex=1 the test will pass. but that's clearly not what we want

sjbur: Made changes to tests

Tucchhaa (approved):

PR #31509, opened 2025-10-29
Chat: implement FileUploader integration

Zedwag:


packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -11,23 +11,33 @@ import type {
...
+const TEXT_AREA_ATTACHMENTS = 'dx-textarea-attachments';
+const TEXT_AREA_ATTACH_BUTTON = 'dx-textarea-attach-button';
 
 const isMobile = (): boolean => devices.current().deviceType !== 'desktop';
 
 type EnterKeyEvent = NativeEventInfo<ChatTextArea, KeyboardEvent>;
 
 export type SendEvent = ClickEvent | EnterKeyEvent;
 
+type FileToUpload = Attachment & {
```
marker-dao: ### suggestion(minor): namig

```suggestion
type FileToSend = Attachment & {
```

or

```
_filesToSend —> _filesToUpload
```

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -200,6 +229,98 @@ class ChatTextArea extends TextArea<Properties> {
     return configuration;
   }
 
+  _renderFileUploader(): void {
+    let $fileUploader = this._fileUploader?.$element();
+    this._fileUploader?.dispose();
```
marker-dao: ### suggestion(non-blocking)
The same code can be used in _dispose(). If you decide to leave the line with `$fileUploader?.remove();` then please move it to a separate method.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -112,9 +136,13 @@ class ChatTextArea extends TextArea<Properties> {
   _initMarkup(): void {
     super._initMarkup();
     this._renderToolbar();
+    this._renderFileUploader();
   }
 
   _renderToolbar(): void {
+    this._toolbar?.dispose();
```
marker-dao: ### suggestion(non-blocking)
The same code already used in _dispose(). Can we move it to a separate method?

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -112,9 +136,13 @@ class ChatTextArea extends TextArea<Properties> {
   _initMarkup(): void {
     super._initMarkup();
     this._renderToolbar();
+    this._renderFileUploader();
```
marker-dao: ### suggestion
I would prefer to render elements in the order they appear in the DOM: FileUploader area, text input, toolbar: we had problems with the toolbar in Popup due to non sequence rendering, so let's stick to the sequence.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -200,6 +229,98 @@ class ChatTextArea extends TextArea<Properties> {
     return configuration;
   }
 
+  _renderFileUploader(): void {
+    let $fileUploader = this._fileUploader?.$element();
+    this._fileUploader?.dispose();
+    $fileUploader?.remove();
```
marker-dao: Actually, you don't have to delete this element here. Then you can use `.appendTo(this.$element())` (instead of `insertBefore`) once during the initial rendering.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -200,6 +229,98 @@ class ChatTextArea extends TextArea<Properties> {
...
+      .addClass(TEXT_AREA_ATTACHMENTS)
+      .insertBefore(this._$textEditorContainer);
+
+    this._fileUploader = this._createComponent(
+      $fileUploader,
+      FileUploader,
+      this._getFileUploaderOptions(),
+    );
+
+    this._filesToSend = new Map<File, FileToUpload>();
```
marker-dao: ### suggestion
I don't think this is the best place to initialize this property: the method is responsible for rendering FileUploader. Please adhere to the principle of single responsibility (SOLID).

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -200,6 +229,98 @@ class ChatTextArea extends TextArea<Properties> {
...
+      visible,
+      onValueChanged,
+      onUploadStarted,
+      onUploaded,
+      onCancelButtonClick,
+    };
+  }
+
   _toggleButtonDisableState(state: boolean): void {
     this._sendButton?.option('disabled', state);
```
marker-dao: ### suggestion
Could we reduce the cases where we use `const shouldButtonBeDisabled = !this._isMessageCanBeSent();` with adding here:
```
const idDisable = state ?? !this._isMessageCanBeSent();
```

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -200,6 +229,98 @@ class ChatTextArea extends TextArea<Properties> {
...
+      }
+
+      const shouldButtonBeDisabled = !this._isMessageCanBeSent();
+      this._toggleButtonDisableState(shouldButtonBeDisabled);
+    };
+
+    return {
+      ...fileUploaderOptions,
+      uploadMode: 'instantly',
+      dialogTrigger: $(`.${TEXT_AREA_ATTACH_BUTTON}`).get(0),
```
marker-dao: ### suggestion(blocking)
It is better to use Guid here because there may be several Chat components on the page. This will be more safe. Or at least search for `this.$element()`, that should narrow down the search area. Right now, it's the entire page.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -200,6 +229,98 @@ class ChatTextArea extends TextArea<Properties> {
...
+      _hideCancelButtonOnUpload: false,
+      _showFileIcon: true,
+      _cancelButtonPosition: 'end',
+      multiple,
+      visible,
```
marker-dao: ### suggestion(minor): codestyle
Please follow this guide: https://github.com/airbnb/javascript?tab=readme-ov-file#objects--grouped-shorthand

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -200,6 +229,98 @@ class ChatTextArea extends TextArea<Properties> {
...
+      onValueChanged,
+      onUploadStarted,
+      onUploaded,
+      onCancelButtonClick,
```
marker-dao: ### suggestion(non-blocking)
Can we make the code more readable and create methods named like `onEventHandler`? Now this method has 62 lines — it is hard to quick understanding / debugging, etc)

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -252,20 +373,55 @@ class ChatTextArea extends TextArea<Properties> {
...
+    }
+
+    return Array.from(this._filesToSend.values())
+      .every((file) => file.readyToSend);
+  }
+
+  _isMessageCanBeSent(): boolean {
+    const hasText = this._isValuableTextEntered();
+    const hasReadyFiles = this._areFilesReadyToSend();
+    const hasUnreadyFiles = this._filesToSend && Array.from(this._filesToSend.values())
```
marker-dao: ### question
How do these variables differ: hasReadyFiles and hasUnreadyFiles? Can we simplify the code here?

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -200,6 +229,98 @@ class ChatTextArea extends TextArea<Properties> {
...
+      this._toggleButtonDisableState(shouldButtonBeDisabled);
+
+      fileUploaderOptions.onUploadStarted?.(e);
+    };
+    const onUploaded = (e: UploadedEvent): void => {
+      const { file } = e;
+      const fileInfo = this._filesToSend?.get(file);
+
+      if (fileInfo) {
+        fileInfo.readyToSend = true;
```
marker-dao: ### suggestion(minor): principle of immutability

```suggestion
        this._filesToSend.set(file, {
            ...fileInfo,
            readyToSend: true,
        });
```

Minor, up to you. But it would be better to get used to it right away before switching to React)

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -252,20 +373,55 @@ class ChatTextArea extends TextArea<Properties> {
...
 
       case 'fileUploaderOptions':
+        this._handleFileUploaderOptionsChange(args);
+        break;
       default:
         super._optionChanged(args);
     }
   }
 
+  _handleFileUploaderOptionsChange(args: OptionChanged<Properties>): void {
```
marker-dao, line 390: Can you explain what's going on here? XD

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -39,18 +49,32 @@ class ChatTextArea extends TextArea<Properties> {
...
+
+  _filesToSend?: Map<File, FileToUpload>;
+
   _sendButton?: Button;
 
   _sendAction?: (e: SendEvent) => void;
 
+  getAttachments(): Attachment[] | undefined {
+    if (!this._filesToSend?.size) {
+      return undefined;
```
marker-dao, line 64: ### question(minor)
Is it required to return undefined here?

```
return Array
      .from(this._filesToSend?.values() ?? [])
      .map(({ name, size }) => ({ name, size }));
```
will returns [].

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
marker-dao: In the future, we will fix the ESLint configuration so that we don't have to do this manually.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
Zedwag: I decided it's more safe, just in case we won't properly handle case when pass message we attachment field, but with empty array

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
Zedwag: well, it's a good point, but it will require additional actions after we render tollbar, since we need to pass toolbar button as triggerDialog element

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
Zedwag: we need to remove it in case of fileUploader runtime disable, otherwise we'll have an empty element which will affect flex container

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
marker-dao: If you want, there is another way: subscribe to Toolbar.onContentReady and set the DialogTrigger button there. Okay, it's up to you.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -200,6 +229,98 @@ class ChatTextArea extends TextArea<Properties> {
...
+  }
+
+  _getFileUploaderOptions(): FileUploaderProperties {
+    const { fileUploaderOptions = {} } = this.option();
+    const multiple = fileUploaderOptions.multiple ?? true;
+    const visible = this._shouldHideFileUploader(fileUploaderOptions.value);
+    const onValueChanged = (e: ValueChangedEvent): void => {
+      const { value, component } = e;
+
+      component.option('visible', this._shouldHideFileUploader(value));
```
r-farkhutdinov:
```suggestion
      component.option({ visible: this._shouldHideFileUploader(value) });
```

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
Zedwag: discussed in voice

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
Zedwag: decided to keep as it, it's more like matter of taste + way with string just a little bit better in performance

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
Zedwag: let's leave as it for now, if we find any problems during testing (especially in other platforms), we know what to do

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
marker-dao: ### issue(blocking)
You did not take into account the height of your block in this function. This is the reason why Scroll may/will not work correctly.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
Zedwag: extracted array creating to another function, but otherwise i don't see a good refactoring here

marker-dao (dismissed):

marker-dao (dismissed):

r-farkhutdinov (approved):

marker-dao (approved):

marker-dao (approved):

PR #31526, opened 2025-10-30
Core Animation: move files to TS

anna-shakhova:


packages/devextreme/js/__internal/common/core/animation/frame.ts
```diff
@@ -0,0 +1,59 @@
...
+
+let cancel = function (requestID: number): void {
+  clearTimeout(requestID);
+};
+
+const setAnimationFrameMethods = callOnce(() => {
+  const nativeRequest = window.requestAnimationFrame
+            || window.webkitRequestAnimationFrame
+            || window.mozRequestAnimationFrame
+            || window.oRequestAnimationFrame
```
pharret31, line 33: What is oRAF?
And do we need all these checks?
According to https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame
All the browsers now supports the default method

packages/devextreme/js/__internal/common/core/animation/frame.ts
```diff
@@ -0,0 +1,59 @@
...
+            || window.msRequestAnimationFrame;
+
+  const nativeCancel = window.cancelAnimationFrame
+            || window.webkitCancelAnimationFrame
+            || window.mozCancelAnimationFrame
+            || window.oCancelAnimationFrame
```
pharret31, line 39: Same: https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame

packages/devextreme/js/__internal/common/core/animation/frame.ts
anna-shakhova: I suppose, it is for Opera, I have not refactored code, just added some types and fixed obvious eslint issues. I think better do all refactoring in separate PR

pharret31 (approved):

EugeniyKiyashko (approved):

dmlvr (approved):

PR #31530, opened 2025-10-30
Scheduler: Appointment popup Customization

sjbur:


packages/devextreme/js/__internal/scheduler/appointment_popup/m_popup.ts
```diff
@@ -93,14 +93,16 @@ export class AppointmentPopup {
...
+    const userPopupOptions = editingConfig?.popup ?? {};
+
+    const defaultPopupConfig = {
       height: 'auto',
       maxHeight: '90%',
       showCloseButton: false,
       showTitle: false,
       preventScrollEvents: false,
       enableBodyScroll: false,
-      _ignorePreventScrollEventsDeprecation: true,
```
aleksei-semikozov, line 103: I am sure that we should leave this private option so that we don't get unnecessary warnings.

aleksei-semikozov (dismissed):

packages/devextreme/js/__internal/scheduler/appointment_popup/m_popup.ts
```diff
@@ -115,6 +118,11 @@ export class AppointmentPopup {
       onShowing: (e): void => this._onShowing(e),
       wrapperAttr: { class: APPOINTMENT_POPUP_CLASS },
     };
+
+    return {
+      ...defaultPopupConfig,
+      ...userPopupOptions,
+    };
```
Tucchhaa: If user has this configuration:
```js
editing.popup: {
   onHiding: myOnHiding,
   onShowing: myOnShowing
}
```

then our `onHiding` and `onShowing` will never be called and this may break popup's behavior (our onShowing callback updates the form). So it's critical for us to call our onHiding and onShowing.

So to make our callbacks and user's callback to run we need to wrap users callback like this:

```ts
const defaultPopupConfig = {
  // ...
  onShowing: (e): void => {
    this._onShowing(e);
    userPopupOptions?.onShowing?.(e);
  },
  // do the same for onHiding
}
``` 

However we don't need to do that for contentTemplate. But we should add a note to our docs that user's can't override contentTemplate. It's the same way as [DataGrid has](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/editing/#popup).

Tucchhaa: We need a deep extend of options here:
```js
extend(true, defaultPopupConfig, userPopupOptions);
```

Reasons are the same as [here](https://github.com/DevExpress/DevExtreme/pull/31555#discussion_r2486312229), but in this case because of [wrapperAttr](https://github.com/DevExpress/DevExtreme/pull/31530/files/6cfc79f52c07f637a2ad82cf753b1965ee7fb1f4#diff-1b3c3b914f817852cdf156470fdf611795b3a65fc4411fa7a8fba800603a937dR119)

packages/devextreme/js/__internal/scheduler/appointment_popup/m_popup.ts
```diff
@@ -93,7 +93,10 @@ export class AppointmentPopup {
   }
 
   _createPopupConfig() {
-    return {
+    const editingConfig = this.scheduler.getEditingConfig();
+    const userPopupOptions = editingConfig?.popup ?? {};
```
Tucchhaa: optional: let's rename to customPopupOptions

packages/devextreme/js/__internal/scheduler/appointment_popup/m_popup.ts
Tucchhaa: Please implement separate tests for both onShowing and onHiding to check that if these callbacks are set, then popup is working properly and user's callback is also working.

Tucchhaa (changes requested):

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1386,6 +1386,89 @@ describe('Appointment Popup Form', () => {
...
+
+    it('should allow overriding default popup options', async () => {
+      const { scheduler, POM } = await createScheduler({
+        ...getDefaultConfig(),
+        editing: {
+          allowAdding: true,
+          allowUpdating: true,
+          popup: {
+            showCloseButton: true,
+            enableBodyScroll: true,
```
Tucchhaa, line 1473: Please add this config to popup:
```js
wrapperAttr: {
  id: 'test'
}
```

and then check that popup class has correct wrapperAttr.class and wrapperAttr.id

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/common.options.tests.js
```diff
@@ -588,7 +588,8 @@ QUnit.module('Options', () => {
             legacyForm: false,
             form: {
                 iconsShowMode: 'main'
-            }
+            },
+            popup: {}
```
anna-shakhova, line 592: In dts we set `undefined` as default value

packages/devextreme/js/ui/scheduler.d.ts
```diff
@@ -665,6 +665,12 @@ export interface dxSchedulerOptions extends WidgetOptions<dxScheduler> {
          */
         iconsShowMode?: AppointmentFormIconsShowMode;
       } | undefined;
+      /**
+       * @docid
+       * @default undefined
+       * @public
+       */
+      popup?: Properties;
```
anna-shakhova: Is it intended to be Scheduler Properties, not Popup one?

packages/devextreme-angular/src/ui/scheduler/nested/editing.ts
```diff
@@ -89,6 +89,14 @@ export class DxoSchedulerEditingComponent extends NestedOption implements OnDest
         this._setOption('form', value);
     }
 
+    @Input()
+    get popup(): any {
```
anna-shakhova: Inconsistent with dts type

packages/devextreme-react/src/scheduler.ts
```diff
@@ -403,6 +403,7 @@ type IEditingProps = React.PropsWithChildren<{
     onCanceled?: ((formData: any) => void);
     onSaved?: ((formData: any) => void);
   };
+  popup?: any;
```
anna-shakhova: Inconsistent with dts type

packages/devextreme-vue/src/scheduler.ts
```diff
@@ -675,7 +676,8 @@ const DxEditingConfig = {
     allowResizing: Boolean,
     allowTimeZoneEditing: Boolean,
     allowUpdating: Boolean,
-    form: Object as PropType<Record<string, any>>
+    form: Object as PropType<Record<string, any>>,
+    popup: {}
```
anna-shakhova: Inconsistent with dts type

packages/devextreme/js/ui/scheduler.d.ts
sjbur: My mistake, changed to Popup properties

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/common.options.tests.js
sjbur: Changed the .d.ts popup default value to {}

packages/devextreme/js/ui/scheduler.d.ts
```diff
@@ -31,7 +31,7 @@ import {
 import dxDraggable from './draggable';
 
 import dxForm, { Item as FormItem } from './form';
-import dxPopup from './popup';
+import dxPopup, { Properties as dxSchedulerPopupOptions } from './popup';
```
anna-shakhova: nitpick: usually we do not use dx prefix for imported type

packages/devextreme/js/ui/scheduler.d.ts
```diff
@@ -665,6 +665,12 @@ export interface dxSchedulerOptions extends WidgetOptions<dxScheduler> {
          */
         iconsShowMode?: AppointmentFormIconsShowMode;
       } | undefined;
+      /**
+       * @docid
+       * @default {}
```
anna-shakhova: Do we need default there? I looked through code base, grid base and card view also have editing config, and they do not set default value. Check their inplementation - similar options should be consistent among all components

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1386,6 +1386,152 @@ describe('Appointment Popup Form', () => {
...
+          allowAdding: true,
+          allowUpdating: true,
+          popup: {
+            onShowing,
+          },
+        },
+      });
+
+      scheduler.showAppointmentPopup(commonAppointment);
+
```
Tucchhaa, line 1521: In this test we need to check that default `onShowing` is also called when custom `onShowing` is set. 

I see that 'onAppointmentFormOpening' callback [is fired](https://github.com/DevExpress/DevExtreme/blob/1df0ea71a2513af24610397b4488c164858f8d68/packages/devextreme/js/__internal/scheduler/appointment_popup/m_popup.ts#L148) by default onShowing callback. So I suggest to add onAppointmentFormOpening callback to the scheduler config and check if it was called too. 
It's just a possible way to do that, it's possible that there could be a better way

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1386,6 +1386,152 @@ describe('Appointment Popup Form', () => {
...
+      expect(onHiding).not.toHaveBeenCalled();
+
+      scheduler.hideAppointmentPopup();
+
```
Tucchhaa, line 1549: In this test we need to check that default `onHiding` is also called when custom `onHiding` is set. 

I see that 'scheduler.focus' [is fired](https://github.com/DevExpress/DevExtreme/blob/1df0ea71a2513af24610397b4488c164858f8d68/packages/devextreme/js/__internal/scheduler/appointment_popup/m_popup.ts#L109). Can you please check what this function does? For example we could check that correct element was focused after popup was closed.

sjbur: Added focus spy to test

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
sjbur: Added onAppointmentFormOpening callback check

packages/devextreme/js/ui/scheduler.d.ts
sjbur: Removed the prefix

Tucchhaa (dismissed):

mpreyskurantov (dismissed):

aleksei-semikozov (approved):

Tucchhaa (approved):

anna-shakhova (approved):

PR #31538, opened 2025-10-30
Chat FU: Hide download button of File when onDownload is not passed

r-farkhutdinov:

### DIFF
--- a/packages/devextreme/js/__internal/ui/chat/chat.ts
+++ b/packages/devextreme/js/__internal/ui/chat/chat.ts
@@ -108,6 +108,7 @@ class Chat extends Widget<Properties> {
       onMessageEntered: undefined,
       onTypingEnd: undefined,
       onTypingStart: undefined,
+      onAttachmentDownload: undefined,
     };
   }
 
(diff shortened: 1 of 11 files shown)


packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -195,11 +195,13 @@ class Chat extends Widget<Properties> {
       dayHeaderFormat,
       messageTimestampFormat,
       typingUsers = [],
+      onAttachmentDownload,
```
marker-dao: ### issue(blocking)
This parameter is not defined in the _getDefaultOptions method.

packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -217,6 +219,7 @@ class Chat extends Widget<Properties> {
       messageTimestampFormat,
       typingUsers,
       isLoading,
+      showAttachmentDownloadButton,
```
marker-dao: ### issue(blocking): extra code
I don't see the point of this option. You can achieve the same functionality by passing undefined as the value for onAttachmentDownload.

marker-dao (approved):

PR #31555, opened 2025-11-03
Scheduler: Appointment form customization

aleksei-semikozov:


packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -257,6 +259,7 @@ export class AppointmentForm {
         this._dxForm = e.component;
         this._recurrenceForm.dxForm = this.dxForm;
       },
+      ...userFormOptions,
```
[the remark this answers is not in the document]
Tucchhaa: I agree with copilot. If user overrides onInitialized or onFieldDataChanged then those callbacks will not work.

A possible solution is to call user's form.onInitialized and form.onFieldDataChanged inside corresponding function. E.g.:

```ts
this.scheduler.createComponent(element, dxForm, {
   // ...
   onInitialized: (e): void => {
        this._dxForm = e.component;
        this._recurrenceForm.dxForm = this.dxForm;
        
        userFormOptions.onInitialized(e);
   },
   // same for onFieldDataChanged
})
```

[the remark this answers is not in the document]
Tucchhaa: Please implement separate tests for both onInitialized and onFieldDataChanged to check that if these callbacks are set, then form is working properly and user's callback is also working.

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -257,6 +259,7 @@ export class AppointmentForm {
         this._dxForm = e.component;
```
Tucchhaa, line 292: By looking at [this line](https://github.com/DevExpress/DevExtreme/pull/31555/files#diff-e2184e9321386d96f057b770cc362ad0d417f9be815ab4e9ac68f5b977f54009R224) I see that, if user has this configuration:

```js
editing.form: {
   elementAttr: {
       someAttr: 'someValue'
   }
}
```

then `elementAttr` prop will be overriden and the form element will not have `class: CLASSES.form`. This could result in a bug report in the future.

We should use a deep extend in this case. We can do that by:
```js
extend(true, defaultOptions, editingConfig.form);
```

but don't forget about onInitialized and onFieldDataChanged callbacks

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -207,6 +207,8 @@ export class AppointmentForm {
 
   private createForm(items: FormProperties['items']): dxForm {
     const element = $('<div>');
+    const editingConfig = this.scheduler.getEditingConfig();
+    const userFormOptions = editingConfig?.form ?? {};
```
Tucchhaa: optional: let's rename userFormOptions to customFormOptions

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1327,6 +1327,28 @@ describe('Appointment Popup Form', () => {
     });
   });
 
+  describe('Form custimization', () => {
```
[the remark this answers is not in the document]
Tucchhaa: Let's apply copilot's suggestion

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
aleksei-semikozov: I made deepMerge options. Regarding the exclusion of the default option, I planned to do the same as in dxGrid and dxCardView, describing which options could break the behavior for the user, but leaving it up to them.
P.S. There are no deep merge options in dxCardView and dxGrid.
https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxCardView/Configuration/editing/#form
I suggest not processing these options in any special way and leaving it up to the customer.

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
Tucchhaa: Please remove `items` property from `customFormOptions`, otherwise form will have only users items without customization

aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
aleksei-semikozov: done

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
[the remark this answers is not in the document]
aleksei-semikozov: Answered bellow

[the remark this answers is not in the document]
aleksei-semikozov: https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxCardView/Configuration/editing/#form
I suggest not processing these options in any special way and leaving it up to the customer.

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -285,13 +289,22 @@ export class AppointmentForm {
       onInitialized: (e): void => {
         this._dxForm = e.component;
         this._recurrenceForm.dxForm = this.dxForm;
+
+        if (onInitialized) {
+          onInitialized.call(this, e);
```
Tucchhaa: Question: why do we need to change context of onInitialized to `this` (which will be internal `AppointmentForm` class)?

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -285,13 +289,22 @@ export class AppointmentForm {
       onInitialized: (e): void => {
         this._dxForm = e.component;
         this._recurrenceForm.dxForm = this.dxForm;
+
+        if (onInitialized) {
```
Tucchhaa: minor: let's use `?.` operator like it's done with onContentReady call

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1339,6 +1339,28 @@ describe('Appointment Popup Form', () => {
     });
   });
 
+  describe('Form customization', () => {
+    it('should propagate editing.form options to the form instance', async () => {
```
Tucchhaa, line 1373: I suggest to add a test that if user's custom form options has `elementAttr.id="someValue"`, then the form has correct values for both 'elementAttr.id' and 'elementAttr.class' (this is a default property) values.

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
Tucchhaa: Maybe we can just call it?: `onInitialized?.(e)`

aleksei-semikozov: We should do this because the customer may want to refer to the existing instance in their handler.
For example:
```
function onInitialized() {
  console.log(this);
}

createScheduler({
  ...getDefaultConfig(),
  ...{
    editing: {
      form: {
        onInitialized,
      },
    },
  },
});
 ```
`this` is a form instance 
Without this call customer should get form instance self

[It is common approach in Devextreme](https://github.com/aleksei-semikozov/DevExtreme/blob/6e90af807b74a7322815dfbeeef1af04e8326b57/packages/devextreme/js/__internal/grids/grid_core/adaptivity/m_adaptivity.ts#L738)

File: packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts (hunk 6 above)
Comment (aleksei-semikozov) [reply packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts]: done

File: packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts (hunk 7 above)
Comment (aleksei-semikozov) [reply packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts:1373]: done

File: packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts (hunk 5 above)
Comment (Tucchhaa) [reply packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts]: I checked and turns out that `this` is an `AppointmentForm` instance. I really doubt that we need to expose this instance to the user. User can use `e.component` to access dxForm instance

By looking at [the blame](https://github.com/aleksei-semikozov/DevExtreme/blame/35b68d84fa6792d9941b77a07b1895ec1d1af5f3/js/ui/grid_core/ui.grid_core.adaptivity.js#L635) of line of code you mentioned, it seems that this code was written more than 8 years ago. So maybe this approach is just obsolete right now.

**Review by Tucchhaa — DISMISSED** [review]

File: packages/devextreme/js/ui/scheduler.d.ts [hunk 8]
```diff
@@ -633,17 +633,10 @@ export interface dxSchedulerOptions extends WidgetOptions<dxScheduler> {
...
        */
-      form?: {
-        /**
-         * @docid
-         * @type Array<dxFormSimpleItem | dxFormGroupItem | dxFormTabbedItem | dxFormEmptyItem | dxFormButtonItem>
-         * @default []
-         * @public
-         */
-        items?: Array<FormItem>;
+      form?: FormProperties & {
```
anna-shakhova: Looks like smth went wrong, we define form type as union, but type doc-tag says that it is dxFormOptions
As result types in frasmeworks differ - in react and vue they removed additional properties and left only dxFormOption fields, while in angular it is dxFormOptions or added properties

anna-shakhova: the construction `FormProperties & {...} | undefined` difficult to read and undestand - it is not clear in what consiquence operators are applied. Better to create intermediate type, and then here use single operand

packages/devextreme/js/ui/scheduler.d.ts
```diff
@@ -79,6 +79,34 @@ export type CellAppointmentsLimit = 'auto' | 'unlimited';
 export type RecurrenceEditMode = 'dialog' | 'occurrence' | 'series';
 /** @public */
 export type AppointmentFormIconsShowMode = 'both' | 'main' | 'recurrence' | 'none';
+
+/**
+ * @docid
+ * @public
+ * @inherits dxFormOptions
+ */
+export type AppointmentFormOptions = FormProperties & {
```
mpreyskurantov: AppointmentFormOptions  -> AppointmentFormProperties

mpreyskurantov (changes requested):

mpreyskurantov (approved):

PR #31572, opened 2025-11-04
Scheduler: T1308596 — "Uncaught TypeError: Cannot read properties of undefined (reading 'getBoundingClientRect')" is thrown when dragging an appointment while an onAppointmentUpdating promise is pending

sjbur:


e2e/testcafe-devextreme/tests/scheduler/common/dragAndDrop/T1308596.ts
aleksei-semikozov, line 1: I noticed that we have e2e tests named after ticket numbers. I’m not fully convinced this naming convention is ideal — it makes it hard to understand what the test actually does without referring to an external ticket.

Regarding this specific test: it feels a bit overengineered to me. It’s testing a negative scenario — something that ideally shouldn’t even happen. In e2e tests we usually want to focus on the positive, expected path, not exceptions.

In my opinion, this test belongs in the code `workspace` tests instead. I’d suggest moving it there.

packages/devextreme-scss/scss/widgets/base/scheduler/appointment/regular/_index.scss
```diff
@@ -164,6 +164,7 @@ $reduced-icon-offset: null !default;
 
   &.dx-scheduler-appointment-drag-source {
     opacity: 0.7;
+    pointer-events: none;
```
aleksei-semikozov: I have a concern about using this CSS property — it can be risky. In most cases we should be able to solve the underlying interaction problem without disabling pointer events entirely

e2e/testcafe-devextreme/tests/scheduler/common/dragAndDrop/T1308596.ts
sjbur: I moved test to a different file, e2e/testcafe-devextreme/tests/scheduler/common/dragAndDrop/dragEvents.ts

I think it is related to drag & drop, because the error occurs only during dragging the appointment.

sjbur: Also I optimized the test, hope it looks a bit better now

aleksei-semikozov (approved):

PR #31589, opened 2025-11-04
Chat FU: attach button should be invisible and fileUploader should reset value after message editing started

Zedwag:


packages/devextreme/js/__internal/ui/chat/chat.ts
```diff
@@ -330,6 +330,8 @@ class Chat extends Widget<Properties> {
       messageEditingStartArgs.cancel,
       () => {
         this._messageBox.option('text', e.message.text);
+        this._messageBox.resetFileUploader();
```
marker-dao, line 333: ### question(non-blocking)
Don't you think that's unnecessary? Can't we move it to _optionChanged case ‘text’ in MessageBox?

marker-dao (approved):

r-farkhutdinov (approved):

packages/devextreme/js/__internal/ui/chat/chat.ts
Zedwag: well, this seems like a better place for me (since toggle visibility also happens here), but in general it's not so important

PR #31605, opened 2025-11-06
Scheduler - Update Templates Demo for jQuery

Tucchhaa:


packages/devextreme/js/__internal/scheduler/appointment_popup/m_popup.ts
```diff
@@ -155,7 +155,6 @@ export class AppointmentPopup {
       if (canceled) {
         e.cancel = true;
       } else {
-        this.updateToolbarForMainGroup();
```
aleksei-semikozov, line 158: We are really should remove this line?

Tucchhaa: Yep, that's a duplicate invocation of `updateToolbarForMainGroup`. Because already have this callstack:
_onShowing -> 
_updateForm -> 
showMainGroup -> 
updateToolbarForMainGroup

Sorry for the unrelated change in the PR, but I have noticed it and was afraid that I would forgot about that.

aleksei-semikozov (dismissed):

aleksei-semikozov (approved):

PR #31627, opened 2025-11-07
Chat FU: rework styles after design review

Zedwag:
- added bg color for while attachments inside messageBox
- removed gap in case when there is no text in a bubble, only attachments
- added hint for download button
- reworked margins for cancel button
- added same font-size for all icons in filiview and fileuploader
- used based font-color for file icons
- font-weght: 500 for filename in material


ksercs (approved):

packages/devextreme-scss/scss/widgets/fluent/fileUploader/_sizes.scss
```diff
@@ -5,11 +5,13 @@
 $fluent-fileuploader-file-container-padding: null !default;
 $fluent-fileuploader-file-container-text-margin: null !default;
 $fluent-fileuploader-file-container-button-margin: null !default;
+$fluent-fileuploader-file-cancel-button-margin-left: 4px !default;
```
marker-dao, line 8:
```suggestion
$fluent-fileuploader-file-cancel-button-margin-start: 4px !default;
```

PR #31634, opened 2025-11-10
Scheduler: Add form customization stories

aleksei-semikozov:


sjbur (dismissed):

apps/react-storybook/stories/scheduler/form-customization.css
```diff
@@ -0,0 +1,10 @@
+.icon-button {
```
Tucchhaa: question: why do we need these styles?

I ask because I had experience with storybook that styles in different stories affected other stories (because selectors where too broad) which was a problem during the design review.

can you please check this style doesn't affect other stories? Maybe the problem was solved, otherwise we need a more specific selector here

aleksei-semikozov: Styles are necessary for customizeItem Story. I add a specific id and a more specific class to avoid the behavior you described before 😊

Tucchhaa (approved):

PR #31674, opened 2025-11-12
CardView: Fix the multiple re-render issue (T1306983, T1309423)

dmirgaev:


packages/devextreme/js/__internal/grids/new/card_view/di.ts
```diff
@@ -24,6 +26,8 @@ export function register(diContext: DIContext): void {
   diContext.register(HeaderPanelView);
   diContext.register(ContextMenuView);
 
+  diContext.register(ItemsController);
+  diContext.addAlias(BaseItemsController, ItemsController);
```
Alyar666: Why do we need this alias? It doesn't seem to be used anywhere.

packages/devextreme/js/__internal/grids/new/card_view/items_controller/items_controller.ts
```diff
@@ -0,0 +1,24 @@
...
+    () => {
+      // NOTE: We should trigger computed by search options change,
+      // But all work with these options encapsulated in SearchHighlightTextProcessor
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+      this.searchController.highlightTextOptions.value;
+
+      return this.dataController.items.value.map(
+        (item, itemIndex) => this.createCardInfo(
+          item,
+          this.columnsController.visibleColumns.peek(),
```
Alyar666: Why can't we fix this at a basic level?

Alyar666 (changes requested):

Alyar666 (approved):

PR #31698, opened 2025-11-13
DataGrid: Fix Tab key navigation to the next editable cell in virtual column display mode if the next column is not in the viewing area (T1287665)

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -566,26 +612,38 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
...
+    const focusNextCell = (): void => {
       const $cell = this._getNextCell(directionCode);
+
       if (isElementDefined($cell)) {
-        this._arrowKeysHandlerFocusCell($event, $cell, directionCode);
+        this._arrowKeysHandlerFocusCell(originalEvent, $cell, directionCode);
       }
+    };
 
-      $event && $event.preventDefault();
```
dmirgaev, line 587: Are we sure we can remove this? This may cause unexpected horizontal scrolling.

Alyar666: I didn't delete it. I moved the preventDefault call up. See line 625.

dmirgaev (approved):

PR #31739, opened 2025-11-17
Scheduler: Add "(No subject)" title to appointments without text

sjbur:


aleksei-semikozov (dismissed):

packages/devextreme/js/__internal/scheduler/m_subscribes.ts
```diff
@@ -133,14 +134,14 @@ const subscribes = {
...
-      text: adapter.text,
+      text: adapter.text || messageLocalization.format('dxScheduler-noSubject'),
       formatDate: formatDates(startDate, endDate, formatType),
     };
   },
 
   _createAppointmentTitle(data) {
     if (isPlainObject(data)) {
-      return data.text;
+      return data.text || messageLocalization.format('dxScheduler-noSubject');
```
Tucchhaa: Why did you change this method? 

I couldn't find any usages of `_createAppointmentTitle` in the entire codebase... Should we remove it?

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1687,6 +1687,28 @@ describe('Appointment Popup Form', () => {
       expect(toolbarItems?.[0]?.text).toBe('Custom Toolbar');
     });
   });
+
+  it('should display "(No subject)" for appointments without title', async () => {
```
Tucchhaa: I suggest to add a visual test to check that "(No subject)" is displayed correctly.

I know that there are already some screenshots that you've updated, but it seems that they are not directly related to appointments without subject.

What do you think?

Tucchhaa (changes requested):

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1687,6 +1687,28 @@ describe('Appointment Popup Form', () => {
...
+      dataSource: [appointmentWithoutTitle],
+    });
+
+    const appointment = POM.getAppointment();
+    expect(appointment.getText()).toBe('(No subject)');
+
+    scheduler.showAppointmentPopup(appointmentWithoutTitle);
+
+    const titleInput = POM.popup.getInputByLabel('Subject');
+    expect(titleInput.value).toBe('');
```
Tucchhaa: I don't think that this BC is somehow related to the Appointment Form.

AppointmentForm uses raw appointment data. Which will never have '(No subject)' text (unless it's in user's data).

In my perspective what we need is:
1. Check that appointment element text is '(No subject)' ✅
2. Check that appointment data in datasource has empty subject field

Checking that subject field is empty in the form's input is not necessary in my opinion, because:
1.  '(No subject') text is inserted when we generate formatted text to render
2. Appointment form uses appointment's data and doesn't take any data from html.

If you agree with this, then should we move this test to a better file, since it's not related to appointment form?

packages/devextreme/js/__internal/scheduler/m_subscribes.ts
sjbur: Yes, I agree to remove it. Wasn't sure about this, seemed that those two methods about same functionality

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
sjbur: Since it is a BC it is a good idea. For such small functionality in other scenario I would consider it as overkill

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
sjbur: I moved this test to a different file. Now I just check (No subject) for appointment without a title. Considering we have visual test now, we may remove this test completely

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
Tucchhaa: I just realized that we don't have visual test for appointment without subject. Maybe add a visual or jest test for appointment tooltip too?

Tucchhaa (dismissed):

vladaskorohodova (dismissed):

Tucchhaa (approved):

vladaskorohodova (approved):

PR #31776, opened 2025-11-19
Scheduer: Appointment Form Redesign: Fix Recurrence Form tests

sjbur:


packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/popup.ts
```diff
@@ -27,6 +27,22 @@ export class PopupModel {
     return this.element.querySelector('.dx-scheduler-form-recurrence-group');
   }
 
+  isMainGroupVisible(): boolean {
+    const group = this.mainGroup;
+    if (!group) return false;
+    const $group = $(group);
+    return !$group.hasClass('dx-scheduler-form-main-group-hidden')
```
Tucchhaa: let's check only class of the element. because it's tabindex doesn't directly affect element's visibility

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
Tucchhaa: We need to replace creating new scheduler by usage of `Scheduler.option()` to update firstDayOfWeek.

packages/testcafe-models/scheduler/index.ts
```diff
@@ -223,4 +223,25 @@ export default class Scheduler extends Widget {
...
+  ): Promise<AppointmentPopup> {
+    const { getInstance } = this;
+
+    await ClientFunction((appointmentData) => {
+      (getInstance() as any).showAppointmentPopup(appointmentData);
+    }, {
+      dependencies: { appointment, getInstance },
+    })(appointment);
+  
+
```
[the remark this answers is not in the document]
Tucchhaa: Looks like we should add eslint rule for that. If it's not hard for you, can you please do it? Or let's just remove extra line and that's it

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/popup.ts
sjbur: Changes applied

Tucchhaa (approved):

PR #31780, opened 2025-11-20
✨ AI Column: fix jQuery, React, Angular demos on the AZ server

dmlvr:


apps/demos/Demos/DataGrid/AIColumns/React/App.tsx
```diff
@@ -90,13 +95,8 @@ export default function App() {
           fixed={true}
           fixedPosition="right"
           cssClass="ai__cell"
-        >
-          <Ai
```
Raushen, line 94: Do we have a confirmation from Platforms we do not have other way and must use an object as a config, not the component?

Raushen (approved):

PR #31798, opened 2025-11-21
Scheduler - Appointment Form - Reimplement legacy tests for the new form - Part 1

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
@@ -130,23 +130,23 @@ describe('Appointment Popup Form', () => {
 
       scheduler.showAppointmentPopup(commonAppointment);
 
-      const subjectEditor = ($(POM.popup.subjectInput) as any).dxTextBox('instance');
-      const descriptionEditor = ($(POM.popup.descriptionTextArea) as any).dxTextArea('instance');
-      const startTimeZoneEditor = ($(POM.popup.startTimeZone) as any).dxSelectBox('instance');
-      const endTimeZoneEditor = ($(POM.popup.endTimeZone) as any).dxSelectBox('instance');
+      const subjectEditor = POM.popup.form.getEditor('subjectEditor');
+      const descriptionEditor = POM.popup.form.getEditor('descriptionEditor');
+      const startTimeZoneEditor = POM.popup.form.getEditor('startTimeZoneEditor');
+      const endTimeZoneEditor = POM.popup.form.getEditor('endTimeZoneEditor');
 
-      descriptionEditor.option('value', 'temp');
-      startTimeZoneEditor.option('value', 'America/Los_Angeles');
-      endTimeZoneEditor.option('value', 'America/Anchorage');
+      descriptionEditor?.option('value', 'temp');
+      startTimeZoneEditor?.option('value', 'America/Los_Angeles');
+      endTimeZoneEditor?.option('value', 'America/Anchorage');
 
       POM.popup.getSaveButton().click();
 
       scheduler.showAppointmentPopup();
 
-      expect(subjectEditor.option('value')).toBe('');
-      expect(descriptionEditor.option('value')).toBe('');
-      expect(startTimeZoneEditor.option('value')).toBeNull();
-      expect(endTimeZoneEditor.option('value')).toBeNull();
+      expect(subjectEditor?.option('value')).toBe('');
+      expect(descriptionEditor?.option('value')).toBe('');
+      expect(startTimeZoneEditor?.option('value')).toBeUndefined();
+      expect(endTimeZoneEditor?.option('value')).toBeUndefined();
     });
   });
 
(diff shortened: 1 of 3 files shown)


packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -297,60 +203,154 @@ describe('Appointment Popup Form', () => {
       expect(POM.popup.endTime).toBeDefined();
     });
 
-    it.each(['day', 'week', 'month'])('should set correct dates when switching on then off in %p view', async (view) => {
+    it.each(
+      ['day', 'week', 'workWeek', 'timelineDay', 'timelineWeek', 'timelineWorkWeek', 'agenda'],
```
aleksei-semikozov: I don't think our form/popup has anything to do with how our appointments are displayed. Let's remove this each, it's not reflected in either the business logic or the code.

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -118,129 +118,35 @@ describe('Appointment Popup Form', () => {
...
 
-      await POM.popup.component.hide();
       scheduler.showAppointmentPopup();
-      expect(POM.popup.component.option('toolbarItems')).toMatchObject(toolbarWithSaveButton);
+
+      expect(subjectEditor?.option('value')).toBe('');
+      expect(descriptionEditor?.option('value')).toBe('');
+      expect(startTimeZoneEditor?.option('value')).toBeUndefined();
+      expect(endTimeZoneEditor?.option('value')).toBeUndefined();
     });
```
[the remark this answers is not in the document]
aleksei-semikozov: I agree with copilot. During internal refactoring, links to editors may change after opening/closing a popup. The test may not be stable.

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -297,60 +203,154 @@ describe('Appointment Popup Form', () => {
...
+      expect(POM.popup.getInputValue('startDateEditor')).toBe('5/9/2017');
+      expect(POM.popup.getInputValue('startTimeEditor')).toBe('9:00 AM');
+      expect(POM.popup.getInputValue('endDateEditor')).toBe('5/10/2017');
+      expect(POM.popup.getInputValue('endTimeEditor')).toBe('12:00 AM');
     });
 
-    it('should show correct dates after switching on allDay and canceling changes (T832711)', async () => {
-      const { scheduler, POM } = await createScheduler(undefined);
+    it.each(
+      ['month', 'timelineMonth'],
```
aleksei-semikozov: The same. I think currentView not affect popup

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
Tucchhaa: It's related to how startDate and endDate values are set. 

If you open all day appt, then you toggle 'all day' switch, then startDate will be equal to startDateHour. And endDate will depend on what's the current view. If it's these views, then endDate will be startDate + half an hour, if view is 'month' then end date is the end of the day.

So that's why we need this each.

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
Tucchhaa: Explained in the prev comment

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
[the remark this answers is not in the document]
Tucchhaa: Thanks, I have applied it

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -400,9 +395,11 @@ export class AppointmentForm {
       editorType: 'dxSwitch',
       editorOptions: {
         onValueChanged: (e) => {
+          this.updateDateTimeEditorsVisibility();
+
           const { startDate } = this;
 
-          if (!startDate) {
+          if (!startDate || e.event === undefined) {
```
aleksei-semikozov, line 402: Why should we separate dxSwitch event handling into native and non-native?

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: If we want to check how an appointment opens in different views, then we should use POM.openPopupByDblClick. Since now it is called via API and does not depend on currentView.
Do we really want to lock this behaviour? To me, it seems a bit strange:
1. Open time 11:30-12:00
2. Double click on allDay switch
3. Time is startDayHour + half hour

sjbur (dismissed):

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
Tucchhaa: Discussed in private. We decided to remove .each() and redundant expects

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
Tucchhaa: Discussed in private. This was needed to fix a bug in how all day switch works

aleksei-semikozov (approved):

PR #31802, opened 2025-11-21
✨AI Column: Test unsupported Binding properties

Raushen:


packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/unsupported_base_properties.integration.test.ts
```diff
@@ -693,4 +693,184 @@ describe('Unsupported properties', () => {
...
+              const { name } = value as { name: string };
+              result[key] = `Response ${name}`;
+            });
+            resolve(JSON.stringify(result));
+          }),
+          abort: (): void => {},
+        };
+      },
+    });
+    it('Should not bind AI column to dataField (first load)', async () => {
```
dmirgaev, line 713: I really don't understand what this test does (= Looks like there are no scenarios for the test fails, it will always be green

packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/unsupported_base_properties.integration.test.ts
```diff
@@ -693,4 +693,184 @@ describe('Unsupported properties', () => {
...
+            name: 'AItest',
+            ai: {
+              prompt: 'Provide name for item with value {value}',
+              aiIntegration,
+            },
+          },
+        ],
+      });
+      await Promise.resolve();
+      instance.columnOption('AItest', 'dataField', 'name');
```
dmirgaev, line 756: But we've already bound AI Column value on "name" field, maybe better use another fields, "value" for example?

packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/unsupported_base_properties.integration.test.ts
```diff
@@ -693,4 +693,184 @@ describe('Unsupported properties', () => {
...
+        showBorders: true,
+        keyExpr: 'id',
+        columns: [
+          'id',
+          {
+            caption: 'AI',
+            type: 'ai',
+            name: 'AItest',
+            dataField: 'name',
+            calculateCellValue: (data) => data.name,
```
dmirgaev, line 774: And again, even if the column reacts on calculateCellValue, we will never know that, as it's already bound on "name" field. Let's use another field instead

packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/unsupported_base_properties.integration.test.ts
```diff
@@ -693,4 +693,184 @@ describe('Unsupported properties', () => {
...
+            ai: {
+              prompt: 'Provide name for item with value {value}',
+              aiIntegration,
+            },
+          },
+        ],
+      });
+      await Promise.resolve();
+      instance.columnOption('AItest', 'dataField', 'name');
+      instance.columnOption('AItest', 'calculateCellValue', (data) => data.name);
```
dmirgaev, line 807: Here the same problem, we already use "name" field in `sendRequest`, such a test looks useless

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/exportController.tests.js
```diff
@@ -139,6 +140,59 @@ QUnit.module('ExportController', {
...
+                    }
+                    resolve(JSON.stringify(result));
+                }),
+                abort: () => {},
+            };
+        },
+    });
+
+    [true, false].forEach((allowExporting) => {
+        QUnit.test(`Get columns from data provider when visible columns has AI column and allowExporting=${allowExporting}`, function(assert) {
```
dmirgaev, line 163: Regardless of the `allowExporting` value passed: true or false - the AI ​​column won't be rendered if the `allowExporting` option is passed at all, right?

dmirgaev (changes requested):

dmirgaev (approved):

PR #31804, opened 2025-11-21
React - useImperativeHandle dependencies fix and useCallbacks' cleanup (T1308983)

ajivanyandev:


packages/devextreme-react/src/core/extension-component.tsx
```diff
@@ -36,6 +36,12 @@ const ExtensionComponent = forwardRef<ComponentBaseRef, any>(
       }
     }, []);
 
+    const imperativeCreateWidget = useRef(createWidget);
```
VasilyStrelyaev: let's maybe rename these entities just `createWidgetRef` here and everywhere - it will be clearer than `imperativeXXX`

VasilyStrelyaev (dismissed):

packages/devextreme-react/src/core/extension-component.tsx
ajivanyandev: Done

VasilyStrelyaev (approved):

PR #31817, opened 2025-11-24
Grids: T1313619 - CardView - OptionChangedEvent type is missing

dmlvr:


packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -715,6 +716,14 @@ export type CardHoverChangedEvent = EventInfo<dxCardView> & WithCardInfo & {
     eventType: string;
 };
 
+/**
+ * @docid _ui_card_view_OptionChangedEvent
+ * @public
+ * @type object
+ * @inherits EventInfo,ChangedOptionInfo
+ */
+export type OptionChangedEvent<TCardData = any, TKey = any> = EventInfo<dxCardView<TCardData, TKey>> & ChangedOptionInfo;
```
dmirgaev: I believe it would be better to define the parameters as `unknown` instead of `any`

packages/devextreme/js/ui/card_view.d.ts
dmirgaev, line 1: After changing this file you need to launch `pnpm run regenerate-all` from the monorepo root
Check carefully whether Vue files were generated successfully as they might be taken from the cache (remote or local)

packages/devextreme/js/ui/card_view.d.ts
dmlvr: changed `any` to `unknown`

mpreyskurantov (changes requested):
add a related `onOptionChanged` prop and link it with a new event handler type like here:
https://github.com/DevExpress/DevExtreme/pull/31791

mpreyskurantov (approved):

dmirgaev (approved):

PR #31831, opened 2025-11-25
DataGrid Deferred Selection Demo: Replace OData with Web API

markallenramirez:


apps/demos/Demos/DataGrid/DeferredSelection/React/App.tsx
```diff
@@ -34,8 +30,8 @@ const App = () => {
   const calculateStatistics = useCallback(async () => {
     const selectedItems = await dataGrid.getSelectedRowsData();
 
-    const totalDuration = selectedItems.reduce((currentValue: number, item: { Task_Due_Date: number; Task_Start_Date: number; }) => {
-      const duration = item.Task_Due_Date - item.Task_Start_Date;
+    const totalDuration = selectedItems.reduce((currentValue: number, item: { Task_Due_Date: string; Task_Start_Date: string; }) => {
+      const duration = +new Date(item.Task_Due_Date) - +new Date(item.Task_Start_Date);
```
Alyar666: I doubt that using unary plus in the demo example is appropriate. It makes the code harder to read. You could use the [getTime](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime) method. But I don't think there is a need to convert time to milliseconds. This will be done automatically when subtracting.

markallenramirez: Unfortunately, TS doesn't allow arithmetic on Date objects directly. Without the unary plus, the follow errors occur:
> The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.
> The right-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.

The `getTime` approach is one of the possible solutions. I realize now that it's more readable than the initial solution. I will make the necessary changes shortly. Thanks.

Alyar666 (approved):

PR #31836, opened 2025-11-25
DataGrid - AI Column: Column reordering does not work if allowColumnReordering and columnFixing.enabled turned on

dmlvr:


packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts
```diff
@@ -915,7 +915,15 @@ export const mergeColumns = (that: ColumnsController, columns, commandColumns, n
   return result;
 };
 
-export const isColumnFixed = (that: ColumnsController, column) => (isDefined(column.fixed) || !column.type ? column.fixed && column.fixedPosition !== StickyPosition.Sticky : that._isColumnFixing());
+export const isColumnFixed = (that: ColumnsController, column) => {
+  const isNotCommandColumn = !column.type || isDefined(column.fixed) || column.type === AI_COLUMN_NAME;
```
Raushen: Let's try to invert the condition and omit extra check for column.fixed. Something like this:
```suggestion
const isFixedCommandColumn = column.type && column.type !== AI_COLUMN_NAME;
if(!isFixedCommandColumn) {
    return column.fixed && column.fixedPosition !== StickyPosition.Sticky;
}
```

dmlvr: Did it and tested. It works, thank you!

Added fix in next commit

Raushen (approved):

PR #31848, opened 2025-11-26
DataGrid - AI Column: Endless prompt editor loading when edit prompt and rollback changes

dmlvr:


packages/devextreme/js/__internal/grids/grid_core/ai_prompt_editor/ai_prompt_editor.ts
```diff
@@ -48,6 +48,12 @@ export class AIPromptEditor {
       height: 110,
       stylingMode: 'outlined',
       onValueChanged: (e): void => {
+        if (e.value === this.prompt) {
```
Alyar666: I think you can write it more simply:
`this.updateButtonOption(APPLY_BUTTON_INDEX, 'disabled', !e.value || e.value === this.prompt);`
`this.updateButtonOption(REGENERATE_DATA_BUTTON_INDEX, 'disabled', !e.value || e.value !== this.prompt);`

Alyar666: You can also move the `e.value === this.prompt` condition to a separate variable. For example, isValueChanged.

dmlvr: thank you! Did it

Alyar666 (approved):

PR #31877, opened 2025-11-27
DataGrid – Resize indicator is moved when resizing a grouped column if showWhenGrouped is set to true (T1314667)

Raushen:


packages/devextreme/js/__internal/grids/grid_core/column_headers/m_column_headers.ts
```diff
@@ -483,25 +482,30 @@ export class ColumnHeadersView extends ColumnContextMenuMixin(ColumnsView) {
...
-    const rowCount = that.getRowCount();
+    const columnsController = this._columnsController;
+    const rowCount = this.getRowCount();
 
-    if (that.option('showColumnHeaders')) {
+    if (this.option('showColumnHeaders')) {
       if (rowCount > 1 && (!isDefined(index) || isDefined(bandColumnIndex))) {
         const result: any[] = [];
-        const visibleColumns = isDefined(bandColumnIndex) ? columnsController.getChildrenByBandColumn(bandColumnIndex, true) : columnsController.getVisibleColumns();
+        let visibleColumns = columnsController.getVisibleColumns();
```
dmirgaev: I suppose It's better to use a ternary operator or an if-else condition. Otherwise, we will execute "getVisibleColumns" method unnecessarily

dmirgaev (approved):

PR #31909, opened 2025-12-02
DataGrid - AI Column is empty in the adaptive row

dmlvr:


dmirgaev (dismissed):

packages/testcafe-models/dataGrid/adaptiveDetailRow.ts
```diff
@@ -11,4 +15,9 @@ export default class AdaptiveDetailRow {
   getAdaptiveCell(index: number): AdaptiveCell {
     return new AdaptiveCell(this.element, index);
   }
+
+  getAdaptiveCellByName(name: string): AdaptiveCell {
```
dmirgaev: Looks like its "cell by text", not "cell by name" (=

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/adaptivity.functional.ts
```diff
@@ -1,12 +1,106 @@
...
+    (window as any).aiResolve = null;
+  }
+});
+
+const deleteGlobalVariables = ClientFunction((): void => {
+  delete (window as any).aiResponseData;
+  delete (window as any).aiResolve;
+});
+
+test('The AI column should have value in the adaptive detail row', async (t) => {
```
dmirgaev: What was the reason to move this test up? I wonder if it can be a problem for future us (=

dmirgaev (changes requested):

packages/testcafe-models/dataGrid/adaptiveDetailRow.ts
dmlvr: agree, renamed it

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/adaptivity.functional.ts
dmlvr: This is not key change for test. The key change is `fixture` without `disablePageReloads` in the test file.

I moved test to the bottom of the file and it's green.

dmirgaev (approved):

PR #31911, opened 2025-12-02
 Scheduler: update react storybook for legacy popup layout

sjbur:


apps/react-storybook/stories/scheduler/SchedulerFormCustomization.stories.tsx
```diff
@@ -332,21 +334,117 @@ export const RTL: Story = {
 export const LegacyPopup: Story = {
   args: {
     ...baseConfig,
-    "editing.form.iconsShowMode": "none",
+    "editing.form.iconsShowMode": "both",
   } as Properties,
   argTypes: iconsShowModeArgType,
   render: (args) => {
+    let form: any = null;
```
siarhei-arzamasau: what about `unknown`?

sjbur: Removed "any" type and used `InstanceType<typeof dxForm> | null` for better typing

apps/react-storybook/stories/scheduler/SchedulerFormCustomization.stories.tsx
```diff
@@ -332,21 +335,109 @@ export const RTL: Story = {
...
+                        editorType: "dxSwitch",
+                        dataField: "repeat",
+                        editorOptions: {
+                          onValueChanged: (e) => {
+                            if (e.value === true) {
+                              form?.option("colCount", 2);
+
+                              const recurrenceRule = form?.option('formData')?.recurrenceRule;
+
+                              form?.option('formData', {
```
Tucchhaa: Minor:
It's possible to access recurrence rule by: `form.option(formData.recurrenceRule)`.

So we can replace it to:
form?.option('formData.recurrenceRule', recurrenceRule || "FREQ=DAILY");

packages/devextreme/js/__internal/scheduler/appointment_popup/m_recurrence_form.ts
```diff
@@ -628,7 +621,6 @@ export class RecurrenceForm {
         break;
     }
 
-    this.dxForm.endUpdate();
```
Tucchhaa, line 631: Question: why do we need to remove beginUpdate/endUpdate?

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1081,6 +1081,30 @@ describe('Appointment Form', () => {
...
+      expect(weekDayButtons).toBeTruthy();
+
+      const firstButton = weekDayButtons?.querySelector('.dx-button');
+      expect(firstButton).toBeTruthy();
+
+      // @ts-expect-error
+      const buttonInstance = $(firstButton).dxButton('instance');
+      expect(buttonInstance?.option('disabled')).toBe(true);
+    });
+
```
Tucchhaa, line 1107: I suggest to add more tests:
- We have added [code](https://github.com/DevExpress/DevExtreme/pull/31911/files#diff-e2184e9321386d96f057b770cc362ad0d417f9be815ab4e9ac68f5b977f54009R674-R677) to preserve values in recurrence form for repeat end editors. So we need to test case when:
1. User opens recurrence form
2. Changes value for repeat end editors (both count and until)
3. User goes back to main form
4. User selects different frequency in repeat editor (not 'never')
5. User open recurrence form again
6. Check that values in count, until and repeat end values are preserved

- We added [code](https://github.com/DevExpress/DevExtreme/pull/31911/files#diff-891dbd01e6233e19cef741c5aa7391803af4f9feb6919ca06c95d17dafe68826R541) to set 'disabled' to repeat end editors on content ready. We need to cover it with tests

packages/devextreme/js/__internal/scheduler/appointment_popup/m_recurrence_form.ts
sjbur: Discussed in DM, returned this field

apps/react-storybook/stories/scheduler/SchedulerFormCustomization.stories.tsx
sjbur: Done

packages/devextreme/js/__internal/scheduler/appointment_popup/m_recurrence_form.ts
```diff
@@ -368,8 +367,11 @@ export class RecurrenceForm {
                 this.recurrenceRule.byDay.push(item.key);
               }
 
-              // todo: update state only of the current button and remove this.updateWeekDaysButtons
-              this.updateWeekDaysButtons();
+              const button = this._weekDayButtons[item.key];
+              const newIsSelected = this.recurrenceRule.byDay.includes(item.key);
+
+              button.option('stylingMode', newIsSelected ? 'contained' : 'outlined');
+              button.option('type', newIsSelected ? 'default' : 'normal');
```
Tucchhaa: minor: it's better to move this code into [the condition](https://github.com/DevExpress/DevExtreme/pull/31911/commits/2e3de5c72fd48df74d172bde008350c164972a50#diff-891dbd01e6233e19cef741c5aa7391803af4f9feb6919ca06c95d17dafe68826R363-R368) above:
```js
if (isSelected) {
  const index = this.recurrenceRule.byDay.indexOf(item.key);
  this.recurrenceRule.byDay.splice(index, 1);
  
  e.component.option('stylingMode', 'outlined');
  e.component.option('type', 'normal');
} else {
  this.recurrenceRule.byDay.push(item.key);

  e.component.option('stylingMode', 'contained');
  e.component.option('type', 'default');
}
```

That's because we do one more array traversal [here](https://github.com/DevExpress/DevExtreme/pull/31911/commits/2e3de5c72fd48df74d172bde008350c164972a50#diff-891dbd01e6233e19cef741c5aa7391803af4f9feb6919ca06c95d17dafe68826R371), which seems redundant. 

But that's a minor comment

apps/react-storybook/stories/scheduler/SchedulerFormCustomization.stories.tsx
```diff
@@ -382,6 +388,12 @@ export const LegacyPopup: Story = {
           form: {
             onContentReady: function (e) {
               form = e.component;
+
+              form.on('fieldDataChanged', (e) => {
+                if (e.dataField === 'recurrenceRule') {
+                  form?.option('formData.repeat', !!e.value)
+                }
+              });
```
Tucchhaa, line 394: it's better to move this handler to form.onInitialized. onContentReady will be called each time form is rendered. Form can be rerendered, for example when recurrenceGroup.visible is changed

packages/devextreme/js/__internal/scheduler/appointment_popup/m_recurrence_form.ts
sjbur: Applied this change

apps/react-storybook/stories/scheduler/SchedulerFormCustomization.stories.tsx
sjbur: Done

apps/react-storybook/stories/scheduler/SchedulerFormCustomization.stories.tsx
```diff
@@ -332,21 +335,117 @@ export const RTL: Story = {
 export const LegacyPopup: Story = {
   args: {
     ...baseConfig,
-    "editing.form.iconsShowMode": "none",
+    "editing.form.iconsShowMode": "both",
   } as Properties,
   argTypes: iconsShowModeArgType,
-  render: (args) => {
```
aleksei-semikozov, line 338: editing.form.iconsShowMode not working you are not use args

Tucchhaa (approved):

apps/react-storybook/stories/scheduler/SchedulerFormCustomization.stories.tsx
Tucchhaa: fixed, thanks

PR #31923, opened 2025-12-04
 Scheduler: Appointment Form — fix 'Save' button visibility in recurrence form for disabled appointment

sjbur:


packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -2013,6 +2078,16 @@ describe('Appointment Popup', () => {
         expect(POM.popup.component.option('toolbarItems')).toMatchObject(toolbarWithCancelButton);
       }
 
+      POM.popup.selectRepeatValue('daily');
```
aleksei-semikozov, line 2052: Let's put this in a separate test (with a different name):
We are use  [AAA](https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80) pattern

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1959,6 +1959,71 @@ describe('Appointment Popup', () => {
         options: { stylingMode: 'outlined' },
       },
     ];
+    const recurrenceToolbarWithSaveButton = [
```
aleksei-semikozov: I think we can rewrite it more readable
```
const [arrow, label, saveButton, cancelButton] = [
  {
    toolbar: "top",
    location: "before",
    widget: "dxButton",
    options: {
      icon: "arrowleft",
      stylingMode: "text",
      onClick: expect.any(Function),
    },
  },
  {
    toolbar: "top",
    location: "before",
    cssClass: "dx-toolbar-label",
  },
  {
    toolbar: "top",
    location: "after",
    widget: "dxButton",
    options: {
      text: "Save",
      stylingMode: "contained",
      type: "default",
      onClick: expect.any(Function),
    },
  },
  {
    toolbar: "top",
    location: "after",
    widget: "dxButton",
    options: {
      text: "Cancel",
      stylingMode: "outlined",
      onClick: expect.any(Function),
    },
  },
];
```

And use it in test

```
const toolbarItems = POM.popup.component.option("toolbarItems");
if (shouldHaveSaveButton) {
  expect(toolbarItems).toMatchObject([arrow, label, saveButton, cancelButton]);
} else {
  expect(toolbarItems).toMatchObject([arrow, label, cancelButton]);
}
```

packages/devextreme/js/__internal/scheduler/appointment_popup/m_popup.ts
```diff
@@ -438,20 +442,21 @@ export class AppointmentPopup {
...
+
+    toolbarItems.push({
+      toolbar: 'top',
+      location: 'after',
+      widget: 'dxButton',
+      options: {
+        text: messageLocalization.format('Cancel'),
+        stylingMode: 'outlined',
+        onClick: (): void => {
+          this.hide();
```
Tucchhaa: I have noticed that for the toolbar of the main form `shortcut` property is used for  'cancel' and 'save' buttons ([here](https://github.com/sjbur/DevExtreme/blob/fe2172fec444363a43e597b8b43745ec6c32ef31/packages/devextreme/js/__internal/scheduler/appointment_popup/m_popup.ts#L499)).

I would like to ask to check we can use 'shortcut' for 'cancel' and 'save' for recurrence toolbar buttons too. Thus we could remove some properties from 'cancel' and 'save' buttons config

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1959,6 +1959,71 @@ describe('Appointment Popup', () => {
...
+        toolbar: 'top',
+        location: 'after',
+        widget: 'dxButton',
+        options: {
+          text: 'Cancel',
+          stylingMode: 'outlined',
+          onClick: expect.any(Function),
+        },
+      },
+    ];
```
Tucchhaa: I think we should not check the whole toolbar items configuration, because if we want to change something in its config this test will fail. So basically we have a duplicate config of toolbar items in both tests and implementation.

I would suggest to check that toolbar items order by checking item's text, cssClass or 'shortcut' property value.

I know that there are already test that check the entire config, but now I think it's an incorrect approach and it's better to refuse it.

But I don't want to push this suggestion and I would like to know your opinion regarding this.

cc @aleksei-semikozov

aleksei-semikozov: I agree, we can change the approach.
The only way to check whether the configured items are displayed correctly is to look at the result with your own eyes, and can't it in jest tests

aleksei-semikozov (approved):

PR #31967, opened 2025-12-11
DataGrid - AI Column loading spinner placement is incorrect on the first load

Raushen:


dmlvr (commented):
Minor: we have different names for property, with we use for min-height.

#gridContainer
#treeList
#employees
.ai__grid

When I see it together it looks a little strange, but our customers will see demos separately. Thats why this only minor comment.

dmlvr (approved):

Raushen:
> When I see it together it looks a little strange, but our customers will see demos separately. Thats why this only minor comment.

Yep, let's leave it as is for now.

PR #31987, opened 2025-12-16
Fix: Vitest, the "window.getComputedStyle" error is not defined when running tests

ajivanyandev:


packages/devextreme/js/__internal/ui/themes.ts
```diff
@@ -60,6 +60,8 @@ function readThemeMarker(): string | null {
       return null;
     }
     return result.substr(THEME_MARKER_PREFIX.length);
+  } catch (e) {
+    return null;
```
VasilyStrelyaev: I'm not sure I like this fix.

First, are we saying that it's normal that under `vitest` the `window.getComputedStyles` method is not mocked? Assuming we are in `jsdom` here, I can see it is in fact mocked:
https://github.com/jsdom/jsdom/blob/main/lib/jsdom/browser/Window.js#L908

and probably that's why this bug is not reproduced on macOS (and even on Parallels' Windows for ARM).

What's different on Windows x64?

And even if we think that it's too expensive to investigate this (do we?), okay, in this case we may want to support the case when there is no `getComputedStyles` in our environment. But is it a good way to do this with a blanket `catch`? Not checking for specific condition?

ajivanyandev: Thanks! Let me investigate a little more, and come back

ajivanyandev: Thanks! Let me investigate a little more, and come back

VasilyStrelyaev (approved):

PR #31996, opened 2025-12-17
CardView, FilterBuilder, Sortable, TreeList: fix ts issues in strict mode

anna-shakhova:


apps/demos/Demos/TreeList/ColumnChooser/React/App.tsx
```diff
@@ -27,6 +27,22 @@ const App = () => {
...
+    setAllowSelectAll(!!value);
+  }, []);
+
+  const onSelectByClickChange = useCallback((value: boolean | null | undefined) => {
+    setSelectByClick(!!value);
+  }, []);
+
+  const onRecursiveChange = useCallback((value: boolean | null | undefined) => {
+    setRecursive(!!value);
+  }, []);
```
pharret31: Why do you use useCallbacks just for calling useState? It seems redundant

apps/demos/Demos/CardView/ColumnChooser/React/App.tsx
```diff
@@ -28,11 +28,27 @@ type ColumnChooserMode = typeof columnChooserModes[number];
 
 const App = () => {
   const [columnChooserMode, setColumnChooserMode] = useState<ColumnChooserMode>('select');
-  const [searchEnabled, setSearchEnabled] = useState(true);
+  const [searchEnabled, setSearchEnabled] = useState<boolean>(true);
   const [allowSelectAll, setAllowSelectAll] = useState(true);
   const [selectByClick, setSelectByClick] = useState(true);
   const [allowColumnReordering, setAllowColumnReordering] = useState(false);
```
pharret31: I'd add type definitions for all useStates, now it's strange that only one has it but all of them have the same type

apps/demos/Demos/TreeList/ColumnChooser/React/App.tsx
anna-shakhova: Checkbox onValueChange type annotation assumes it may be boolean | null | undefined, and setState allows only boolean value here, that's why wrap is needed

apps/demos/Demos/CardView/ColumnChooser/React/App.tsx
anna-shakhova: not necessary, it is recognized as boolean from initial value

Alyar666 (approved):

PR #32007, opened 2025-12-18
Upload SBOM to gha artifacts, remove from devextreme-artifacts package

alexslavr:

### DIFF
--- a/.github/workflows/packages_publishing.yml
+++ b/.github/workflows/packages_publishing.yml
@@ -99,7 +99,7 @@ jobs:
       - name: Upload SBOM artifact
         uses: actions/upload-artifact@v4
         with:
-          name: SBOM
+          name: sbom
           path: packages/sbom/dist
           retention-days: 7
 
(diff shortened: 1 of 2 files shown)


.github/workflows/packages_publishing.yml
```diff
@@ -96,7 +96,15 @@ jobs:
       - name: Build artifacts package
         run: npx ts-node tools/scripts/make-artifacts-package
 
-      - uses: actions/upload-artifact@v4
+      - name: Upload SBOM artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: SBOM
```
IlyaKhD:
```suggestion
          name: sbom
```

Let's name it in lowercase to match another workflow artifact: [sbom](https://github.com/DevExpress/DevExtreme/blob/25_2/.github/workflows/build_all.yml#L75)

IlyaKhD: Or we can make both uppercase 🤔

IlyaKhD (approved):

PR #32050, opened 2025-12-25
QUnit tests: ensure compatibility with Fluent and drop Generic theme testing (Editors)

Zedwag:


marker-dao (approved):

packages/devextreme/testing/tests/DevExpress.knockout/fieldset_bundled.fluent.tests.js
```diff
@@ -0,0 +1,4 @@
+require('fluent_blue_light.css!');
+
+const testFieldsetsOnPlatform = require('./fieldsetParts/shared.js');
+testFieldsetsOnPlatform('fluent.blue.light', { testSwitchBaseline: false });
```
pharret31, line 4: I guess this file is no longer needed since we now have only one fluent testing so it should be removed, and the shared.js should be moved to the upper level with the name fieldset_bundled.tests.js

PR #32054, opened 2025-12-26
ArrayStore - onPush TypeScript type does not match actual return value (T1316896)

ajivanyandev:


e2e/compilation-cases/T1316896.ts
```diff
@@ -0,0 +1,54 @@
+import ArrayStore from 'devextreme/data/array_store';
```
mpreyskurantov: devextreme/common ...

packages/devextreme/js/data/store.d.ts
```diff
@@ -166,7 +173,7 @@ export class Store<
      * @param1 changes:Array<any>
```
mpreyskurantov, line 174: Array vs Array<any> in Docs

VasilyStrelyaev (approved):

mpreyskurantov (approved):

PR #32073, opened 2025-12-30
Scheduler: T1310544 — scrollTo does not take offset into account

sjbur:


aleksei-semikozov (commented):
I see two remote test files.
How I see them:
workspace.base.test.ts - unit test.
workspace.recalculation.test.ts - integration test where we create the scheduler

I don't think this test should be in one file
I would rename workspace.recalculation.test.ts -> workspace.test.ts and write your test case in this file

sjbur:
@aleksei-semikozov applied changes you asked

aleksei-semikozov (dismissed):

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
```diff
@@ -1865,10 +1868,14 @@ class SchedulerWorkSpace extends Widget<WorkspaceOptionsInternal> {
   }
 
   _isValidScrollDate(date, throwWarning = true) {
+    const viewOffset = this.option('viewOffset') as number;
     const min = this.getStartViewDate();
     const max = this.getEndViewDate();
 
-    if (date < min || date > max) {
+    const extendedMin = new Date(min.getTime() - viewOffset);
+    const extendedMax = new Date(max.getTime() + viewOffset);
```
Tucchhaa: Question: shouldn't min and max dates be offseted in the same direction?

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
```diff
@@ -1411,13 +1411,16 @@ class SchedulerWorkSpace extends Widget<WorkspaceOptionsInternal> {
...
+    if (viewOffset === 0) {
+      if (hours < startDayHour) {
+        hours = startDayHour;
+      }
 
-    if (hours >= endDayHour) {
-      hours = endDayHour - 1;
+      if (hours >= endDayHour) {
+        hours = endDayHour - 1;
+      }
```
Tucchhaa: I notice that we validate scrollTo date [here](https://github.com/DevExpress/DevExtreme/blob/37a82c0cbc9a284898895756c2f2ea93a0561d4d/packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts#L1830) before calculating scrollTo coordinates. 

I suspect that this normalization may be redundant now, since we will throw a warning if scrollTo date is out of range. But I am not fully sure.

Can you investigate it please?

Tucchhaa (changes requested):

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
Tucchhaa: I got it. We need this normalization for views larger than day, e.g. timelineWeek, timelineMonth, etc.

However we should not just ignore this normalization if viewOffset is set. We need to adjust this normalization to include viewOffset into it.

For example [this case](https://codepen.io/FufelShmertz/pen/qENZKvv?editors=0010) (I have changed currentView and viewOffset). The scheduler is scrolled to 5 PM, however it should scroll to the last cell of the day (7 PM).

Please adjust the normalization and write a test for it.

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
sjbur: Yes, you're right

PR #32104, opened 2026-01-07
Demos: improve linting

pharret31:


dmlvr (dismissed):

r-farkhutdinov (dismissed):

dmirgaev (dismissed):

r-farkhutdinov (dismissed):

dmlvr (dismissed):

apps/demos/eslint.config.mjs
```diff
@@ -459,7 +468,11 @@ export default [
   // testcafe tests
   ...compat.extends('devextreme/testcafe').map(config => ({
     ...config,
-    rules: changeRulesToStylistic(config.rules || {}),
+    rules: {
+      // ...changeRulesToStylistic(config.rules || {}),
+      ...config.rules,
+      'no-only-tests/no-only-tests': 0,
```
Tucchhaa: Are you sure about disabling this rule?

pharret31: yes, this is required to make 'only' tests works, it's related to testcafe helpers, not to testcafe tests themselves

apps/demos/eslint.config.mjs
```diff
@@ -487,5 +500,33 @@ export default [
   ...compat.extends('plugin:jest/recommended', 'plugin:jest/style').map(config => ({
     ...config,
     files: ['utils/tests/**/*.*'],
+    rules: {
+      'jest/expect-expect': 0,
```
Tucchhaa: Seems that rule checks that jest tests should have at least one `expect`. Why do we need to disable this rule?

pharret31: there was an error if there's this rule, it seems like it's been provoked by inconsistency in dependencies versions of eslint/ts, in order to not to dig it deeper it was disabled, I'd suggest to consider this in another task

apps/demos/eslint.config.mjs
Tucchhaa: Is it possible to use `// eslint-disable-next-line-` where needed? I think it's not good idea to disable this check for the entire demo's repo

apps/demos/eslint.config.mjs
Tucchhaa: I would suggest to use `// eslint-disable-next-line` if there are not many errors and if it's possible to fix this way, or mark this rule with comment like `// TODO enable this rule` to not forget about it 🙂

r-farkhutdinov (dismissed):

dmlvr (dismissed):

dmirgaev (dismissed):

Tucchhaa (dismissed):

r-farkhutdinov (dismissed):

dmlvr (dismissed):

r-farkhutdinov (approved):

dmirgaev (approved):

Tucchhaa (approved):

dmlvr (approved):

PR #32162, opened 2026-01-12
Material & Fluent themes: ensure compliance with color-contrast requirements according to a11y standards

EugeniyKiyashko:


Zedwag (approved):

marker-dao (approved):

alexslavr (approved):

dmlvr (approved):

Raushen (approved):

aleksei-semikozov (approved):

apps/demos/Demos/DataGrid/BatchEditing/Angular/app/app.component.css
```diff
@@ -5,7 +5,6 @@
 .options {
   margin-top: 20px;
   padding: 20px;
-  background: #f5f5f5;
```
dmirgaev, line 8: Maybe better use some CSS variable instead of relying on a transparent background?

apps/demos/Demos/DataGrid/StatePersistence/Angular/app/app.component.css
```diff
@@ -4,7 +4,6 @@
 }
 
 ::ng-deep #descContainer a {
-  color: #f05b41;
```
dmirgaev, line 7: The color could be inherited from another rule for `a` element, therefore I suppose it'd be better to use a variable instead of relying on default browser settings

apps/demos/Demos/TreeList/FilterModes/Angular/app/app.component.css
```diff
@@ -5,7 +5,6 @@
 .options {
   margin-top: 20px;
   padding: 20px;
-  background: #f5f5f5;
```
dmirgaev, line 8: The same: better use a CSS variable than rely on a transparent background

apps/demos/Demos/TreeList/StatePersistence/Angular/app/app.component.css
```diff
@@ -4,7 +4,6 @@
 }
 
 ::ng-deep #descContainer a {
-  color: #f05b41;
```
dmirgaev, line 7: Here we go again: the color could be inherited from another rule for a element, therefore I suppose it'd be better to use a variable instead of relying on default browser settings

apps/demos/Demos/DataGrid/StatePersistence/Angular/app/app.component.css
EugeniyKiyashko: The $base-color-link color will be applied. It is a high-level selector defined in the theme typography settings for <a> tags.

apps/demos/Demos/TreeList/StatePersistence/Angular/app/app.component.css
EugeniyKiyashko: Answered here:https://github.com/DevExpress/DevExtreme/pull/32162#discussion_r2765549175

apps/demos/Demos/TreeList/FilterModes/Angular/app/app.component.css
EugeniyKiyashko: Thank you for your remark, you are right. This style defines the background of the options section, and in dark themes a specific color is used, as defined in the Widget Gallery styles. We will align on the corresponding variable for this purpose with the designer and fix it in the next PR.

apps/demos/Demos/DataGrid/BatchEditing/Angular/app/app.component.css
EugeniyKiyashko: Answered here: https://github.com/DevExpress/DevExtreme/pull/32162#discussion_r2765599478

apps/demos/Demos/TreeList/FilterModes/Angular/app/app.component.css
EugeniyKiyashko: PR: https://github.com/DevExpress/DevExtreme/pull/32405

PR #32186, opened 2026-01-13
T1311329 - DataGrid - Column chooser hides a banded column on using search and recursive selection

Raushen:


packages/devextreme/js/__internal/grids/grid_core/column_chooser/__tests__/column_chooser.integration.test.ts
```diff
@@ -0,0 +1,226 @@
...
+  $container.remove();
+  jest.clearAllMocks();
+  jest.useRealTimers();
+};
+
+describe('Bugs', () => {
+  beforeEach(beforeTest);
+  afterEach(afterTest);
+
+  describe('T1311329 - DataGrid - Column chooser hides a banded column on using search and recursive selection', () => {
```
anna-shakhova, line 65: I see in docs that band columns support hierarchies of any nesting level, please, add test case with hiding/searching on deep levels

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -288,11 +295,53 @@ export class ColumnChooserView extends ColumnsView {
...
+      const columnA = this._columnsController.columnOption(a.itemData.id);
+      const columnB = this._columnsController.columnOption(b.itemData.id);
+      const isBandA = columnA?.hasColumns ? 1 : 0;
+      const isBandB = columnB?.hasColumns ? 1 : 0;
+      // Band columns should be updated after regular columns
+      return isBandA - isBandB;
+    });
+  }
+
+  private updateColumnVisibility(nodes: Node[]): void {
```
anna-shakhova: we update all columns visibility every time user selects column in TreeView, it also has `onItemSelectionChanged` event with item data, it can be more efficient to use it and update only effected column instead of all

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
anna-shakhova, line 367: This line should select all TreeView nodes that a not allowed to be hidden, on TreeView selection changes
But these nodes are disabled, and if they are selected on 1st render nothing should happen with them. Do we really need this logic on each node click?

Raushen: As I see Select All check box changes selection of disabled items too: https://codepen.io/Raushen-the-styleful/pen/YPWNwNX
Did I miss some option to prevent this?

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
Raushen: As I said [in the comment below](https://github.com/DevExpress/DevExtreme/pull/32186/changes#r2692305947) we have to fix TreeView behavior.
For this purpose we use beginUpdate/endUpdate mechanism to make all changes at once.
If we will handle all events separately and apply changes immediately, user will see flicks of grid columns as they can have complex calculations for example with adaptivity or autowidth properties.
For now we have one entry point and handle all changes at once.

packages/devextreme/js/__internal/grids/grid_core/column_chooser/__tests__/column_chooser.integration.test.ts
Raushen: Added test with nesting level = 3

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
anna-shakhova: Got it, disabled children also change their state with click on parent

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -288,46 +301,75 @@ export class ColumnChooserView extends ColumnsView {
...
+      result.push({ node, level });
+
+      if (node.children.length) {
+        getNodeLevelPairsRecursive(node.children, result, level + 1);
+      }
+
+      return result;
+    }, flatNodesArray);
+
+    const flatNodes = getNodeLevelPairsRecursive(nodes, [], 0)
```
anna-shakhova: sorting logic looks excessive, after recursively getting nodes, we will have them in order like: [1, 2, 3, 3_1, 3_2, 3_2_1, 3_2_2, 3_3, 4, 5...]
If we just reverse the order, we will process them in correct order: children first, then their parent, then parents' parent

packages/devextreme/js/__internal/grids/grid_core/column_chooser/m_column_chooser.ts
```diff
@@ -288,46 +296,70 @@ export class ColumnChooserView extends ColumnsView {
...
+    nodes.forEach((node) => {
+      const columnIndex = node.itemData.id;
+      const isVisible = this._getColumnVisibility(columnIndex, node.selected);
+
+      this._columnsController.columnOption(columnIndex, 'visible', isVisible);
+    });
+  }
+
+  private _getSortedFlatNodes(nodes: NodeInternal[]): NodeInternal[] {
+    const getNodeLevelPairsRecursive = (
```
anna-shakhova: this utility not returns node-level pairs any more, let's rename it to smth like `flattenNodesDepthFirst` to show the order is important

also method `_getSortedFlatNodes` has incorrect name too - we do not sort any more

anna-shakhova (approved):

PR #32190, opened 2026-01-14
Scheduler: add new scroll api method

aleksei-semikozov:


packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -2022,8 +2031,30 @@ class Scheduler extends SchedulerOptionsBaseWidget {
...
+    let allDayValue;
+    let align: 'start' | 'center' = 'center';
+
+    if (this._isScrollOptionsObject(groupValuesOrOptions)) {
+      groupValues = groupValuesOrOptions.group;
+      allDayValue = groupValuesOrOptions.allDay;
+      align = groupValuesOrOptions.align ?? 'center';
+    } else {
+      groupValues = groupValuesOrOptions;
+      allDayValue = allDay;
```
Tucchhaa, line 2041: I have a question. Will we deprecate this old scrollTo API in the future? Should show a warning if the user uses old API?

packages/devextreme/js/__internal/scheduler/__tests__/workspace.base.test.ts
```diff
@@ -70,3 +101,95 @@ describe('scheduler workspace', () => {
...
+
+  it('should scroll to all day panel when allDay is true', () => {
+    const { workspace, container } = createWorkspace(SchedulerTimelineDay, 'timelineDay');
+
+    const scrollableElement = container.querySelector('.dx-scheduler-date-table-scrollable') as HTMLElement;
+    const scrollableContainer = scrollableElement.querySelector('.dx-scrollable-container') as HTMLElement;
+
+    workspace.scrollTo(new Date(2017, 4, 25, 22, 0), undefined, true);
+
+    expect(scrollableContainer.scrollLeft).toBeCloseTo(11125);
```
Tucchhaa, line 164: In the neighbor [PR](https://github.com/DevExpress/DevExtreme/pull/32160/changes#diff-b03a67923390a307b8c5301fb8dae454648aa46fe7149bff86e1b68bcfac1c67R301-R306), we also implemented tests for scrollTo, but in testcafe. Instead of checking magic numbers for scroll we check the cell data in scrollable's viewport to see if the scroll worked correctly.

I think that maybe in this test we could use the same approach (I mean to check  the cell data instead of scroll position, because it appears more reliable). 

What do you think?

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -90,6 +91,14 @@ import SchedulerWorkSpaceMonth from './workspaces/m_work_space_month';
...
 import SchedulerWorkSpaceWorkWeek from './workspaces/m_work_space_work_week';
 
+interface ScrollToOptions {
+  group?: RawGroupValues | GroupValues;
+  allDay?: boolean | undefined;
+  align?: 'start' | 'center';
+}
+
+type ScrollToGroupValuesOrOptions = RawGroupValues | GroupValues | ScrollToOptions | undefined;
+
```
Tucchhaa: Maybe [types.ts file](https://github.com/DevExpress/DevExtreme/blob/12ebcb44da59156fe9d47f674888320a2e0962c7/packages/devextreme/js/__internal/scheduler/types.ts#L1) is a more appopriate place for this types?

sjbur: I agree that we should move this type somewhere, however, types.ts looks messy to me, but it is better than have this type here

packages/devextreme/js/__internal/scheduler/__tests__/workspace.base.test.ts
aleksei-semikozov: I think for unit tests and e2e tests we need different strategies.

## E2E vs Unit Tests

For **e2e tests** (like in your testcafe PR #32160) - yes, checking cell data is better approach! We test like user, high-level:

```javascript
scheduler.scrollTo(date);
expect(getSchedulerCenterViewport(scheduler)).toEqual(date) // Good for e2e
```

But for **unit tests** - I think direct checks is better. Let me explain why.

## Problem with Test Helpers in Unit Tests

When test fails with helper function like `getWorkspaceCenterViewport()`, it's not clear what broken:
- Is problem in workspace code?
- Or problem in helper logic?
- What exactly changed?

Helper function adds one more abstraction layer. In unit test we already testing small piece, so we don't need extra abstraction. It makes harder to understand what really happens.

Also helpers can grow complex over time and check more things than we actually need for this specific test.

## Why "Magic Numbers" Can Be OK

About magic numbers - I can make them more clear with calculation:

```javascript
const cellWidth = 60;
const cellCount = 6; 
const padding = 18;

expect(scrollLeft).toEqual(cellWidth * cellCount - padding * 2); // = 324
```

This way it's not really "magic" - you see what we checking and why this number.

When test fails, I immediately see: "oh, scroll is 334 instead of 324, so +10px difference" - very quick to understand what broken.

With helper I need dig into helper code to understand what it calculates and why test fails.

## Main Point

For unit tests:
- Test breaks fast when something change ✓
- Easy to fix (just update number) ✓ 
- Clear what exactly broken ✓

For e2e tests:
- Your approach with checking cell data is correct ✓
- More stable when refactoring ✓

So I think both approaches good, just for different test types.

What you think?

Tucchhaa: Are these tests unit tests? For me it appears more like integration tests, but with just workspace rendered. Or may wrong here? 🤔

Nevertheless, I don't quite understand your point about test helpers. I think `getCellDataAtViewportCenter` is more clear than 11125.

I think magic numbers are bad because of two reasons: 
1. It's not clear, what result we get. Yeah, workspace scrolled, but did it scroll to the correct date? We can't know, unless we launch the test manually in playground and see the result by our eyes.

2. Hard coded logic. What if some sizes in scheduler changes? For example some 1px border was added, then all tests with magic numbers will fail. If we have a lot of them, then we will need to fix all the magic numbers.

Also because of this exact reason, now we have to update a large amount of qunit tests to make it work with fluent theme instead of generic. With e2e approach we will save ourselves from the pain in the future, if for example we would want to test in different theme, then we wouldn't need to fix these tests.

What I suggest?:
Since we basically test the same thing in two PRs, but in different ways. I suggest to find a common ground and write similar tests, either in jest or in testcafe. My main point is to remove magic numbers. Since your PR tests two aligns of scrollTo 'start' and 'center', I think we could modify the `getCellDataAtViewportCenter` to work with both alignments.

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
aleksei-semikozov: Yes. we are should, fix it

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
aleksei-semikozov: Ok, thank you. Move it

packages/devextreme/js/__internal/scheduler/__tests__/workspace.base.test.ts
aleksei-semikozov: Unit test or no unit test? 
If we imagine that a workspace can be a separate entity that communicates according to specific contracts, then this entity can be called a module, and this module can be covered by unit tests. This is a holy war topic. In our code, I see three levels of testing:
— modular (unit): form+popup, workspace, viewmodel (appointment collection), timeZoneCalculator
— integration: the work of all modules in conjunction with the scheduler
— e2e: testcafe environment, real browser
We could call modular (unit) testing integration testing. But I think that would add confusion.

Regarding your comments:
1. The name of the test indicates what we should obtain.
2. This environment is locked, css does not affect them, only js. If we logically decide to add padding in js, then all checks will need to be rewritten.

As I wrote above, using test helpers, which hide behaviour for unit tests, does not seem very good in my opinion.

sjbur (dismissed):

Tucchhaa (dismissed):

packages/devextreme/js/ui/scheduler.d.ts
```diff
@@ -1308,6 +1308,20 @@ export default class dxScheduler extends Widget<dxSchedulerOptions> {
      * @public
      */
     scrollTo(date: Date, group?: object, allDay?: boolean): void;
+    /**
+     * @docid
+     * @publicName scrollTo(date, options)
+     * @param2 options:Object|undefined
+     * @param2_field group:Object|undefined
+     * @param2_field allDay:Boolean|undefined
+     * @param2_field align:Enums.SchedulerScrollToAlign|undefined
```
anna-shakhova:
```suggestion
     * @param2_field alignInView:Enums.SchedulerScrollToAlign|undefined
```

anna-shakhova: We do not have type SchedulerScrollToAlign, either add this union as type (preferrable), or use `string | undefined`signature

aleksei-semikozov: fixed

aleksei-semikozov: fixed

anna-shakhova (approved):

sjbur (approved):

PR #32196, opened 2026-01-14
React: Add Activity support

ajivanyandev:


packages/devextreme-react/src/core/component-base.tsx
```diff
@@ -340,6 +368,25 @@ const ComponentBase = forwardRef<ComponentBaseRef, any>(
...
+        // If element is still connected, it is likely Activity hide -> keep instance
+        if (el?.isConnected) return;
+
+        disposedRef.disposed = true;
+        disposeWidgetNow();
+      };
+
+      // Run in both microtask and macrotask to avoid timing edge cases
+      queueMicrotask(checkAndDispose);
+      setTimeout(checkAndDispose, 0);
```
EvgeniiVarnavskii, line 387: This sequence is pointless: the microtask will always run before any timeout.

PR #32226, opened 2026-01-16
DataGrid: Fix the header filter\'s aria-label when column's caption has a "\n" character (T1318766). Normalization of aria attributes of elements.

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/m_modules.ts
```diff
@@ -145,14 +145,15 @@ export class ModuleItem {
     return this._actions[actionName];
   }
 
-  public setAria(name, value, $target) {
+  public setAria(name: string, value: any, $target: dxElementWrapper) {
```
anna-shakhova:
```suggestion
  public setAria(name: string, value: string | number | boolean | undefined, $target: dxElementWrapper) {
```

packages/devextreme/js/__internal/grids/grid_core/__tests__/grid.integration.test.ts
```diff
@@ -0,0 +1,82 @@
...
+          { id: 2, name: 'Name 2', value: 20 },
+        ],
+        columns: ['id', { dataField: 'name', caption: 'Test\nName' }, 'value'],
+        showBorders: true,
+        headerFilter: {
+          visible: true,
+        },
+      });
+
+      expect(component.getHeaderCellFilter(1).attr('aria-label')).not.toMatch(/\n/);
```
anna-shakhova: let's also add positive assertion for expected attr value

anna-shakhova (approved):

PR #32258, opened 2026-01-21
Scheduler: refactor Header module

sjbur:


packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -205,74 +202,82 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
...
+    return extend(true, {}, item) as ToolbarItem;
   }
 
-  _callEvent(event, arg) {
+  _callEvent(event: string, arg: unknown): void {
     if (this.eventMap.has(event)) {
       const events = this.eventMap.get(event);
-      events.forEach((event) => event(arg));
+      if (events) {
+        events.forEach((EventMapHandler) => EventMapHandler(arg));
```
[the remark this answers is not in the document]
aleksei-semikozov: I agree

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
aleksei-semikozov: You are forget here step: Step

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
```diff
@@ -71,56 +72,34 @@ const getDateAfterWorkWeek = (workWeekStart: Date): Date => {
...
-} => {
-  const startDate = getIntervalStartDate(options);
-  const endDate = getIntervalEndDate(startDate, options);
-
-  return { startDate, endDate };
-};
-
-const getPeriodEndDate = (currentPeriodStartDate: Date, step: Step, agendaDuration: number): Date => {
-  let date;
+const getPeriodEndDate = (
```
aleksei-semikozov, line 97: we can rewrite simple without eslint issues and without create redundant default value;
`let date: Date = new Date();`

```
const getPeriodEndDate = (
  currentPeriodStartDate: Date,
  step: Step,
  agendaDuration: number,
): Date => {
  const calculators: Record<Step, () => Date> = {
    day: () => nextDay(currentPeriodStartDate),
    week: () => nextWeek(currentPeriodStartDate),
    month: () => nextMonth(currentPeriodStartDate),
    workWeek: () => getDateAfterWorkWeek(currentPeriodStartDate),
    agenda: () => nextAgendaStart(currentPeriodStartDate, agendaDuration),
  };

  return subMS(calculators[step]());
};
```

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
sjbur: Wrote about this in DM

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
```diff
@@ -104,23 +114,26 @@ export default class SchedulerCalendar extends Widget<dxSchedulerOptions> {
...
-      focusStateEnabled: this.option('focusStateEnabled'),
-      onValueChanged: this.option('onValueChanged'),
+      value,
+      min,
+      max,
+      firstDayOfWeek,
+      focusStateEnabled,
+      tabIndex,
+      onValueChanged,
+      // @ts-expect-error skipFocusCheck is an internal Calendar property
```
Tucchhaa, line 130: Another way to do this is to declare two functions in SchedulerCalendar:

```ts

  setOption<K extends keyof HeaderCalendarOptions>(
    optionName: K,
    value: HeaderCalendarOptions[K],
  ): void {
    this.option(optionName, value);
  }

  getOption<K extends keyof HeaderCalendarOptions>(
    optionName: K,
  ): HeaderCalendarOptions[K] {
    return this.option(optionName) as unknown as HeaderCalendarOptions[K];
  }
```

and use them in the code instead (also in other modules like `m_header.ts`), this way `Component.option()` will be used as intended (there's some caching logic behind optionManger), though, through a wrapper functions. 

However I don't push applying this changes, but rather I would like to discuss how it should be done in the future.

From my perspective, it's a problem that basic `Component` doesn't support typed values that should be resolved. Otherwise it feels a little bit messy to refactor code to ts, with structures like this. 

But to push this change to Component may require some time, and we need to decide what approach to use temporarily.

What do you think? @aleksei-semikozov @sjbur

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
```diff
@@ -67,24 +71,28 @@ export default class SchedulerCalendar extends Widget<dxSchedulerOptions> {
...
+
+    if (isMobileLayout) {
+      this._overlay = this._createComponent(this.$element(), Popup, overlayConfig);
+    } else {
+      this._overlay = this._createComponent(this.$element(), Popover, overlayConfig);
+    }
   }
 
-  _createOverlayContent() {
+  _createOverlayContent(): dxElementWrapper {
```
Tucchhaa: I would like to ask to add function modifiers like `public` and `private`, if you don't mind

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
sjbur: Will apply this change to function

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
```diff
@@ -1,62 +1,66 @@
...
-  hide() {
-    this._overlay.hide();
+  public async hide(): Promise<void> {
+    await this._overlay?.hide();
   }
 
-  _keyboardHandler(opts): void {
+  public _keyboardHandler(opts: KeyboardKeyDownEvent): boolean {
     this._calendar?._keyboardHandler(opts);
+    return true;
```
Tucchhaa: Looks like it should be this:
```suggestion
return this._calendar?._keyboardHandler(opts);
```

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
```diff
@@ -94,7 +102,9 @@ export default class SchedulerCalendar extends Widget<dxSchedulerOptions> {
     return result;
   }
 
-  _optionChanged({ name, value }) {
+  public _optionChanged(
+    { name, value } : { name: string; value: DateLike | DateLike[] },
```
Tucchhaa: I think parameter type should be the same as in the Widget class ([Widget._optionChanged](https://github.com/DevExpress/DevExtreme/blob/5ecd11c4060b743e434da4fe3c6190d763511692/packages/devextreme/js/__internal/core/widget/widget.ts#L597)).

It looks strange to me to have function different type in child's class. Moreover if another option is handled is _optionChanged, thenthe parameter type is not correct anymore.

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -65,50 +69,49 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
     };
   }
 
-  _getDefaultOptions() {
-    // @ts-expect-error
+  public _getDefaultOptions(): HeaderOptions & { _useShortDateFormat: boolean } {
     return extend(super._getDefaultOptions(), {
       _useShortDateFormat: !devices.real().generic || devices.isSimulator(),
-    });
+    }) as HeaderOptions & { _useShortDateFormat: boolean };
```
Tucchhaa: Can we add `_useShortDateFormat` to HeaderOptions?

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -65,50 +69,49 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
...
 
-  _addEvent(name, event) {
+  public _addEvent(name: string, event: EventMapHandler): void {
     if (!this.eventMap.has(name)) {
       this.eventMap.set(name, []);
     }
 
     const events = this.eventMap.get(name);
-    this.eventMap.set(name, [...events, event]);
+    if (events) {
```
Tucchhaa: this is a redundant if-statement, because `this.eventMap.get(name)` must be defined according to the code above.

We can rewrite this function to:

```ts
  public _addEvent(name: string, event: EventMapHandler): void {
    const events = this.eventMap.get(name) ?? [];
    this.eventMap.set(name, [...events, event]);
  }
```

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -65,50 +69,49 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
...
   }
 
-  _optionChanged(args) {
+  public _optionChanged(args: OptionChanged<HeaderOptions>): void {
     const { name, value } = args;
 
     if (this.eventMap.has(name)) {
       const events = this.eventMap.get(name);
-      events.forEach((event) => {
+      events?.forEach((event) => {
```
Tucchhaa: redundant if statement. Let's rewrite to the function to this:
```ts
  public _optionChanged(args: OptionChanged<HeaderOptions>): void {
    const { name, value } = args;

    const events = this.eventMap.get(name);

    events?.forEach((event) => {
      event(value);
    });
  }

```

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -119,53 +122,47 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
...
 
-    // @ts-expect-error
     this._toolbar = this._createComponent(toolbarElement, Toolbar, config);
   }
 
-  _toggleVisibility(): void {
-    const toolbarOptions = this.option('toolbar') as any;
-    const isHeaderShown = toolbarOptions.visible
-      || (toolbarOptions.visible === undefined && toolbarOptions.items.length);
+  public _toggleVisibility(): void {
```
Tucchhaa, line 155: isn't it private method?

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -119,53 +122,47 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
...
+    const { toolbar } = this.option();
+    const isHeaderShown = toolbar.visible
+      ?? (toolbar.visible === undefined && toolbar.items.length);
```
Tucchhaa: Wrong if-statement. Check copilot's comment. I suggest to rewrite to:
`const isHeaderShown = toolbar.visible ?? toolbar.items.length;`

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -205,74 +202,82 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
...
   }
 
-  _callEvent(event, arg) {
+  private _callEvent(event: string, arg: unknown): void {
     if (this.eventMap.has(event)) {
       const events = this.eventMap.get(event);
-      events.forEach((event) => event(arg));
+      if (events) {
+        events.forEach((eventMapHandler) => eventMapHandler(arg));
+      }
```
Tucchhaa: redundant if-statement, I suggest to rewrite it to this:
```ts
  private _callEvent(event: string, arg: unknown): void {
    const events = this.eventMap.get(event);

    events?.forEach((eventMapHandler) => eventMapHandler(arg));
  }
```

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -281,28 +286,27 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
     return this.getIntervalOptions(date);
   }
 
-  _getCaption() {
+  public _getCaption(): DateNavigatorTextInfo {
+    const { customizeDateNavigatorText } = this.option();
     const options = this._getCaptionOptions();
-    const customizationFunction = this.option('customizeDateNavigatorText');
     const useShortDateFormat = this.option('_useShortDateFormat');
```
Tucchhaa: After `_useShortDateFormat` is added to the HeaderOptions, will it be possible to use `this.option()._useShortDateFormat` instead?

packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -281,28 +286,27 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
...
 
-  _showCalendar(e) {
-    this._calendar.show(e.element);
+  public async _showCalendar(e: ItemClickEvent): Promise<void> {
+    await this._calendar?.show(e.element);
   }
 
-  _hideCalendar() {
-    this._calendar.hide();
+  private async _hideCalendar(): Promise<void> {
```
Tucchhaa: is this function used anywhere?

packages/devextreme/js/__internal/scheduler/header/m_header.ts
Tucchhaa: Also there some other `.option('value')` usages, I suggest to replace them all:
- https://github.com/DevExpress/DevExtreme/blob/5ecd11c4060b743e434da4fe3c6190d763511692/packages/devextreme/js/__internal/scheduler/header/m_header.ts#L193
- https://github.com/DevExpress/DevExtreme/blob/5ecd11c4060b743e434da4fe3c6190d763511692/packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts#L39
- https://github.com/DevExpress/DevExtreme/blob/5ecd11c4060b743e434da4fe3c6190d763511692/packages/devextreme/js/__internal/scheduler/header/m_date_navigator.ts#L53

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
sjbur: Yes, will fix this

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
sjbur: Yes, will fix this

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Yes, will add

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Legit

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Legit

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Yes, this one should be changed to private

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: I think not, we can remove it

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Will change

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Will change

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Legit

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Ah, I see.  I get error, because _toggleVisibility is not specified as private in base Widget class.

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: I added the type to HeaderOptions and noticed that we need to pass this property in m_scheduler.ts I decided to remove override default props and just pass _useShortDateFormat directly from m_scheduler.ts 

What do you think about this approach?

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/header/m_header.ts
sjbur: Removed

packages/devextreme/js/__internal/scheduler/header/m_header.ts
Tucchhaa: Can we make this property optional in HeaderOptions type and set its value in m_header.ts? This would be better as m_scheduler doesn't need to know about this option

sjbur: OK, Done.

Tucchhaa (approved):

PR #32263, opened 2026-01-21
DataGrid: prevent column resizing in band area (T1317039)

anna-shakhova:


packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_row.ts
```diff
@@ -897,7 +897,9 @@ const columnsResizer = (Base: ModuleType<ColumnsResizerViewController>) => class
       if (overlayInstance) {
         const cellIndex = overlayInstance.$element().closest('td').index();
 
-        if (cellIndex === that._targetPoint.columnIndex || cellIndex === that._targetPoint.columnIndex + 1) {
+        if (cellIndex === that._targetPoint?.columnIndex
```
Alyar666: **Minor comment**: This code could be refactored as follows:
`
.....

    super._startResizing.apply(that, arguments);

    // @ts-expect-error
    const overlayInstance = that._columnHeadersView.getFilterRangeOverlayInstance();

    if (!overlayInstance || !that.isResizing() || !that._targetPoint) {
        return;
    }

    const cellIndex = overlayInstance.$element().closest('td').index();

    if (cellIndex === that._targetPoint.columnIndex || cellIndex === that._targetPoint.columnIndex + 1) {
        overlayInstance.$content().hide();
    }
.....
`

e2e/testcafe-devextreme/tests/dataGrid/common/columnResizing/functional.ts
```diff
@@ -0,0 +1,152 @@
...
+  )();
+};
+
+// T1317039
+test('DataGrid – Columns should not be resized from band area (T1317039)', async (t) => {
+  const dataGrid = new DataGrid('#container');
+
+  await t.expect(dataGrid.isReady()).ok();
+
+  await tryResizeHeaderInBandArea(dataGrid, 3, 30);
```
Alyar666, line 108: **Minor comment**: I think you can use the existing model method for resizing headers - `dataGrid.resizeHeader`.
This really needs some work. For example, you can add a 4th argument - initialOffset: { top: number, left: number }

Alyar666 (dismissed):

e2e/testcafe-devextreme/tests/dataGrid/common/columnResizing/functional.ts
anna-shakhova: it is not quite clear to be 4th argument here, it should be 2nd, or maybe 3rd, but this will cause update usages everywhere just for a single case

Alyar666 (approved):

PR #32283, opened 2026-01-22
T1319739 - DataGrid - Columns are misaligned after adding a column at runtime

Raushen:


packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -488,6 +488,20 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     e.group = columnsController.getGroupDataSourceParameters(!dataSource.remoteOperations().grouping);
   }
 
+  private needFullRepaint(optionNames, changeTypes) {
```
dmirgaev: Don't we have an agreement to name private methods with `_`?

Raushen: The modern code we write without underscore.
Earlier we wrote underscore to show the method is private but now we have Typescript and just the modifier "private")

Alyar666 (approved):

PR #32286, opened 2026-01-23
FilterBuilder: fix invalid date formatting (T1319193, T1311486)

anna-shakhova:


Alyar666 (dismissed):

packages/testcafe-models/dataGrid/index.ts
```diff
@@ -245,6 +248,15 @@ export default class DataGrid extends GridCore {
   getFilterRangeOverlay(): Selector {
     return this.getHeadersContainer().child(`.${this.addWidgetPrefix(CLASS.filterRangeOverlay)}`);
   }
+
+  getFilterRangeStartEditor(): DateBox {
```
Alyar666: The filter range start/end editor can also be NumberBox

Alyar666 (changes requested):

Alyar666 (approved):

PR #32290, opened 2026-01-23
T1308327 - DataGrid - Cell value is not restored after canceling changes in cell editing mode if repaintChangesOnly is enabled

Raushen:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/validating/__tests__/validating.integration.test.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/validating/__tests__/validating.integration.test.ts
@@ -1,62 +1,11 @@
 import {
   afterEach, beforeEach, describe, expect, it, jest,
 } from '@jest/globals';
-import type { dxElementWrapper } from '@js/core/renderer';
-import $ from '@js/core/renderer';
-import type { Properties as DataGridProperties } from '@js/ui/data_grid';
-import DataGrid from '@js/ui/data_grid';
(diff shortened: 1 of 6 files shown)


packages/devextreme/js/__internal/grids/grid_core/validating/__tests__/validating.integration.test.ts
```diff
@@ -0,0 +1,72 @@
+import {
+  afterEach, beforeEach, describe, expect, it, jest,
+} from '@jest/globals';
+import {
+  afterTest,
+  beforeTest,
+  createDataGrid,
+} from '@ts/grids/grid_core/__tests__/__mock__/helpers/utils';
+
+describe('Bugs', () => {
```
Alyar666: We typically use the "should... when..." pattern.
For example, in your case it would be like this:

```
describe('DataGrid Cell Editing', () => {
    describe('when showEditorAlways and repaintChangesOnly is enabled', () => {
        it('should restore the value after canceling changes with validation error (T1308327)', async () => {
            .....
        });
    });
});
```

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/validating/__tests__/validating.integration.test.ts
dmirgaev: done

Alyar666 (approved):

PR #32292, opened 2026-01-23
GridCore: fix data controller odata test

anna-shakhova:


dmirgaev (dismissed):

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/data_controller.test.ts
Raushen: It seems we need to use sample url like "https://expample.com/odata" and mock store method returning data. After that we can leave all checks as is.

mpreyskurantov: +1, You can find some examples at:
- https://github.com/DevExpress/devextreme-documentation/pull/8324

mpreyskurantov (approved):

Raushen (approved):

PR #32334, opened 2026-01-28
List: fix border cases on dragging first or last items using KBN (T1320189)

dmlvr:


packages/devextreme/testing/tests/DevExpress.ui.widgets/listParts/editingTests.js
```diff
@@ -319,6 +319,53 @@ QUnit.module('keyboard navigation', {
         assert.deepEqual(list.option('items'), items, 'items were reordered');
     });
 
+    QUnit.test('shift+arrowUp on first item should not throw error', function(assert) {
```
pharret31: I don't see the test that would check if error isn't thrown, could you please add it to make the test title and its body connected? As well as for the second test, to make that I suppose you can use sinon methods smth like:
```
const consoleSpy = sinon.spy(console, 'error');
...
assert.strictEqual(consoleSpy.callCount, 0, 'no error messages in console');
```
because currently it seems that old behaviour would pass these tests as well since before items order wasn't changed too, isn't it?
also you can actually combine them in one using forEach syntax

or

if this behaviour actually throws an error it could be handled by standard QUnit methods

dmlvr: Thank you! The test crashed without fix, but I wasn't checking for errors.

Rewrote test, now it directly checks for errors.

pharret31 (approved):

PR #32348, opened 2026-01-29
T1318244 - DataGrid - The "Certain ARIA roles must be contained by particular parents" error occurs if the grid has no data

Raushen:


e2e/testcafe-devextreme/package.json
```diff
@@ -11,7 +11,7 @@
     "@babel/plugin-transform-runtime": "7.28.5",
     "@testcafe-community/axe": "3.5.0",
     "@types/jquery": "catalog:",
-    "axe-core": "4.10.3",
+    "axe-core": "4.11.1",
```
anna-shakhova: please, update axe-core version in other packages as well
actually, we can add it to catalog to keep in always in sync

anna-shakhova (approved):

Alyar666 (approved):

PR #32387, opened 2026-02-03
Storybook: Update WF parameters

r-farkhutdinov:

### DIFF
--- a/.github/workflows/pr-storybook-deploy-manual.yml
+++ b/.github/workflows/pr-storybook-deploy-manual.yml
@@ -42,6 +42,12 @@ jobs:
           ref: refs/pull/${{ inputs.pr_number }}/head
           fetch-depth: 1
 
+      - name: Checkout base repo
+        if: inputs.action == 'remove'
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-storybook-deploy-manual.yml
```diff
@@ -42,6 +42,12 @@ jobs:
           ref: refs/pull/${{ inputs.pr_number }}/head
           fetch-depth: 1
 
+      - name: Checkout base repo
```
chaosmirage, line 45: Do we need it? 

Seems like it is already done under the hood https://github.com/rossjrw/pr-preview-action/blob/6f3a89d9ff1b752c6e781d91eacd27d9952188b0/action.yml#L157 by https://github.com/JamesIves/github-pages-deploy-action

r-farkhutdinov: Yes, we do. Without checkout, removing will not work properly (I believe it's related to the access rights)

chaosmirage (dismissed):

chaosmirage (approved):

PR #32397, opened 2026-02-04
Scheduler: mark scrollTo method as deprecated

aleksei-semikozov:


mpreyskurantov:
Let's better do the proposed changes within original lines numbers, to minimize diffs

aleksei-semikozov:
> Let's better do the proposed changes within original lines numbers, to minimize diffs

You need to swap the api definitions. If you don't do this, ts will issue a warning on the scrollTo(date) call.
In this case, there will be no problems.

sjbur (approved):

anna-shakhova (approved):

PR #32399, opened 2026-02-04
Chat: Add STT integration types

r-farkhutdinov:
Moved to:
- #32517


packages/devextreme/js/ui/chat.d.ts
```diff
@@ -432,6 +434,12 @@ export interface dxChatOptions extends WidgetOptions<dxChat> {
      * @public
      */
     alerts?: Array<Alert>;
+    /**
+     * @docid
+     * @default null
```
marker-dao: null is not equal to string or undefined) Same for fileUploaderOptions and speechToTextOptions.

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -475,6 +483,13 @@ export interface dxChatOptions extends WidgetOptions<dxChat> {
      * @public
      */
     showMessageTimestamp?: boolean;
+    /**
+     * @docid
+     * @default null
```
marker-dao: Also, we should set undefined as the default value here, since null is a set value, and we don't set anything by default.

By the way, we are still discussing whether we should use @default undefined. Let's also ask Anna to review this PR.

packages/devextreme/js/ui/chat.d.ts
```diff
@@ -432,6 +434,12 @@ export interface dxChatOptions extends WidgetOptions<dxChat> {
      * @public
      */
     alerts?: Array<Alert>;
+    /**
+     * @docid
+     * @default null
+     * @public
+     */
+    messageBoxValue?: string;
```
marker-dao: This type must match the one used in TextArea.

marker-dao: Also, I'm confused by the name of this option. But let's wait for the review from technical writers)

r-farkhutdinov: For TextArea, it's `value`. We can try using `textAreaValue`, but on chat abstraction level it's still not clear. I would suggest to keep it as is or discuss the naming later.

packages/devextreme/js/ui/chat.d.ts
r-farkhutdinov: Agree with this point, however, we need to keep everything consistent and changing this in one place will require changing all other. Once final decision is made, we can use `@default undefined/''`, etc.

packages/devextreme/js/ui/chat.d.ts
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/32399/changes/23ddc4a69e09cdc22ed613104ba3699bd7274ddc

packages/devextreme/js/ui/chat.d.ts
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/32399/changes/23ddc4a69e09cdc22ed613104ba3699bd7274ddc

packages/devextreme/js/ui/chat.d.ts
r-farkhutdinov: https://github.com/DevExpress/DevExtreme/pull/32399/changes/23ddc4a69e09cdc22ed613104ba3699bd7274ddc

packages/devextreme-vue/src/chat.ts
```diff
@@ -537,6 +579,108 @@ const DxMessageTimestampFormat = defineComponent(DxMessageTimestampFormatConfig)
...
+    "update:onResult": null,
+    "update:onStartClick": null,
+    "update:onStopClick": null,
+    "update:rtlEnabled": null,
+    "update:speechRecognitionConfig": null,
+    "update:startIcon": null,
+    "update:startText": null,
+    "update:stopIcon": null,
+    "update:stopText": null,
+    "update:stylingMode": null,
```
marker-dao, line 635: Omit doesn't work here and for type prop?

packages/devextreme-react/src/chat.ts
```diff
@@ -351,6 +372,78 @@ const MessageTimestampFormat = Object.assign<typeof _componentMessageTimestampFo
...
+  onResult?: ((e: ResultEvent) => void) | undefined;
+  onStartClick?: ((e: StartClickEvent) => void) | undefined;
+  onStopClick?: ((e: StopClickEvent) => void) | undefined;
+  rtlEnabled?: boolean;
+  speechRecognitionConfig?: Record<string, any> | SpeechToTextSpeechRecognitionConfig;
+  startIcon?: string;
+  startText?: string;
+  stopIcon?: string;
+  stopText?: string;
+  stylingMode?: ButtonStyle;
```
marker-dao, line 424: Same here. Need to check.

packages/devextreme-angular/src/ui/chat/nested/speech-to-text-options.ts
```diff
@@ -0,0 +1,301 @@
...
+    @Input()
+    get stopText(): string {
+        return this._getOption('stopText');
+    }
+    set stopText(value: string) {
+        this._setOption('stopText', value);
+    }
+
+    @Input()
+    get stylingMode(): ButtonStyle {
```
marker-dao, line 229: And here.

mpreyskurantov (changes requested):
Let's merge #32246 first

PR #32402, opened 2026-02-04
Scheduler: delete Appointment enhancements

sjbur:


packages/devextreme-scss/scss/widgets/generic/list/_colors.scss
```diff
@@ -89,10 +89,10 @@ $list-item-chevron-icon-color: $base-icon-color !default;
   $list-item-selected-bg: color.change($base-accent, $alpha: 0.48) !default; // #f05b41 => #ffb5a8
   $list-item-selected-hover-bg: color.change($base-accent, $alpha: 0.3) !default;
   $list-item-focused-selected-bg: color.change($base-focus-bg, $alpha: 0.7) !default;
-  $list-focused-bg: $base-focus-bg !default;
+  $list-focused-bg: $list-item-hover-bg !default;
```
Tucchhaa: I see that color is changed for the dxList. Were these changes discussed with designer and Navigation squad?

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
```diff
@@ -72,6 +72,21 @@ export class TooltipStrategyBase {
         this.hide();
         this._tooltip.option('target').focus();
       });
+      this._list.registerKeyHandler?.('del', (e) => {
+        const { focusedElement } = this._list.option();
+
+        if (!focusedElement || !this._options.allowDelete) {
```
Tucchhaa: If I have an appt with 'disabled' field, e.g:
```js
{
    disabled: true,
    text: 'Website Re-Design Plan',
    startDate: new Date('2021-04-26T16:30:00.000Z'),
    endDate: new Date('2021-04-26T18:30:00.000Z'),
  }
```

Then this appointment can be deleted using KBN, even though it shouldn't because delete icon isn't displayed.

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
```diff
@@ -72,6 +72,21 @@ export class TooltipStrategyBase {
...
+          return;
+        }
+
+        const { appointment, targetedAppointment } = this._list._getItemData(focusedElement);
+
+        if (appointment) {
+          e.preventDefault();
+          this.hide();
```
Tucchhaa: Was hiding a tooltip after appt is deleted using KBN discussed with the designer?
For me it feels like it shouldn't, so I would like to know what the designer thinks

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
```diff
@@ -72,6 +72,21 @@ export class TooltipStrategyBase {
...
+        const { focusedElement } = this._list.option();
+
+        if (!focusedElement || !this._options.allowDelete) {
+          return;
+        }
+
+        const { appointment, targetedAppointment } = this._list._getItemData(focusedElement);
+
+        if (appointment) {
+          e.preventDefault();
```
Tucchhaa: I have a question. Why preventDefault is needed here?

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
Tucchhaa: Take a look at this [line of code](https://github.com/DevExpress/DevExtreme/blob/bfd020bf9c3a2e7334369d2c452379528f0e19f8/packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts#L207). 

Instead of passing `_options.allowDelete` from m_scheduler, it would be better to use the existing logic. I would suggest to move that line of code into a separate function and use it in both places

sjbur: I will take a look

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
sjbur: I will ask designer

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
sjbur: Yes, I took a look here and now it seems redundant to me. I looked other cases of using registerKeyHandler and there are no e.preventDefault use cases

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
sjbur: Implemented same logic as in that line

packages/devextreme-scss/scss/widgets/generic/list/_colors.scss
sjbur: As part of this task we needed to fix contrast within background and foreground in appointment item collector. Appointment collector item is a list item. I agree that maybe we should not change this, but in this case I am not sure that can fix contrast issue on our side

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
sjbur: The designer suggested not to hide the collector neither by keyboard deletion neither by mouse. Hide collector only if it does not have any appointments left

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
Tucchhaa: Ok, but can we move this condition to a separate function? I think, it's better to keep such logic in one place. Or if you have objections, let's discuss it

packages/devextreme-scss/scss/widgets/generic/list/_colors.scss
Tucchhaa: If the contrast issue can reproduced only with dxList, then, yeah, it should be fixed here. But it's still strange that `$list-focused-bg` variable is assigned a value of `$list-item-hover-bg`. These two variables should used for different scenarios.

If the contrast issue doesn't reproduce on dxList and only on scheduler's side, then we should override some of the styles on Scheduler's side.

If you have some questions, let's discuss them :)

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
Tucchhaa: Also, please, add a test for the case of not allowing deletion of disabled appointment by keyboard

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
Tucchhaa: Ok, can you please add tests for scenarios when tooltip should/shouldn't close?

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
sjbur: Extracted to separate method

sjbur: Added test

packages/devextreme-scss/scss/widgets/generic/list/_colors.scss
sjbur: Tried to fix the contrast issue on our side. Added screenshot test for generic, material and fluent theme for appointment in collector focused state.

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
sjbur: Implemented this behavior and added tests

packages/devextreme/build/bundle-templates/dx.custom.js
```diff
@@ -0,0 +1,194 @@
+"use strict";
```
Tucchhaa: Please remove this file from the PR

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_tooltip_strategy_base.ts
```diff
@@ -35,43 +35,79 @@ export class TooltipStrategyBase {
...
+    const newDataList = currentItems.filter((item) => item.appointment !== appointment);
+
+    if (newDataList.length === 0) {
+      this.hide();
+      return;
+    }
+
+    const $target = this._tooltip?.option('target');
+    this.show($target, newDataList, this._extraOptions);
+  }
```
Tucchhaa: This method makes the tooltip to blink (quickly hide and show) after appointment deletion. I don't think this is a good UI. Please fix it.

Moreover we shouldn't compute newDataList here. We should use dataList that is computed on m_scheduler's side. And we should update tooltip, only **after** deletion is completed.

That's because of this case:
When appt is deleted, the [onAppointmentDeleting](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/Configuration/#onAppointmentDeleting) event is fired.
If user sets `e.cancel=true`, then the appt will not be deleted. However due to this logic, it will be remove from the tooltip.

Tucchhaa (changes requested):

e2e/testcafe-devextreme/tests/scheduler/common/tooltipBehaviour/tooltipBehavior.ts
```diff
@@ -125,6 +126,123 @@ test.meta({ browserSize: [600, 400] })('The tooltip should hide after manually s
...
+    text: 'Text3',
+    startDate: new Date(2017, 4, 22, 9, 30, 0, 0),
+    endDate: new Date(2017, 4, 22, 10, 30, 0, 0),
+  }],
+  views: [{ type: 'month', maxAppointmentsPerCell: 1 }],
+  currentView: 'month',
+  currentDate: new Date(2017, 4, 22),
+}));
+
+test.meta({ browserSize: DEFAULT_BROWSER_SIZE })('Collector tooltip should hide when last item is deleted', async (t) => {
```
Tucchhaa: can we move this test to jest?

e2e/testcafe-devextreme/tests/scheduler/common/deleteAppointments.ts
```diff
@@ -100,3 +100,84 @@ test('Common appointments should be deleted by click on \'delete\' button and pr
...
+    endDate: new Date(2017, 4, 22, 10, 30, 0, 0),
+    disabled: true,
+  }],
+  views: [{
+    type: 'month',
+    maxAppointmentsPerCell: 1,
+  }],
+  currentView: 'month',
+  currentDate: new Date(2017, 4, 22),
+}));
```
Tucchhaa: Is it possible to move this testcafe tests to jest?

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_desktop_tooltip_strategy.ts
```diff
@@ -8,47 +8,47 @@ const APPOINTMENT_TOOLTIP_WRAPPER_CLASS = 'dx-scheduler-appointment-tooltip-wrap
 const MAX_TOOLTIP_HEIGHT = 200;
 
 export class DesktopTooltipStrategy extends TooltipStrategyBase {
-  _prepareBeforeVisibleChanged(dataList) {
+  protected override prepareBeforeVisibleChanged(dataList) {
```
Tucchhaa, line 11: Thank you for doing this refactoring, but I kindly ask you to do refactoring in the separate PR next time or in the single first commit of the PR.

Right now this refactoring adds a mess to the changes and it is hard to focus on the fix itself.

e2e/testcafe-devextreme/tests/scheduler/common/deleteAppointments.ts
sjbur: Moved

e2e/testcafe-devextreme/tests/scheduler/common/tooltipBehaviour/tooltipBehavior.ts
sjbur: Moved

packages/devextreme/js/__internal/scheduler/tooltip_strategies/m_desktop_tooltip_strategy.ts
sjbur: OK, sorry. I think I saw a message by Aleksei suggesting to do such refactoring when is possible

Tucchhaa:
I have got the idea of your implementation:
1. show tooltip 
2. subscribe to appointments dataSource change
3. when appt is deleted, get new appointments in the collector from m_appointments_collection by passing sortIndex of the collector element
4. update tooltip list's items

There are several problems with this approach:
1 . Tooltip list's items are collected from SchedulerAppointments.option('items') on dataChange event. But SchedulerAppointments.option('items') are updated on dataChange event too. See [scheduler._renderAppointments](https://github.com/DevExpress/DevExtreme/blob/4439294b4566d952f2d577572dcadb166bb8f0e1/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L886) is called in [onDataSourceChangedChandler](https://github.com/DevExpress/DevExtreme/blob/4439294b4566d952f2d577572dcadb166bb8f0e1/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L853)

While this is works, because onDataSourceChangedHandler callback is subscribed before, the tooltip's dataChange handler, I consider this as a bad approach, because there's no clear data flow and it is very hard to trace such dependencies. Moreover we cannot guarantee that these callbacks will always be called in this order.

2 . We can not rely on the collectorSortIndex to get list of appointments. Here's the example scenario:
Put appointments like this:
image

1. Focus the first collector
2. Delete all items in the tooltip using keyboard

**Result:**
Second tooltip is shown for some reason, while it shouldn't

That happens, because when all items are deleted the collector element doesn't exist anymore, but the code uses its  sortIndex to get list items. Thus tooltip is shown again instead of closing.

PR #32469, opened 2026-02-09
Scheduler - Appointments Collection - Refactor KBN

Tucchhaa:


sjbur (dismissed):

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
```diff
@@ -0,0 +1,118 @@
...
+      tab: this._tabHandler.bind(this),
+    };
+  }
+
+  public resetTabIndex($appointment: dxElementWrapper): void {
+    this.$focusableItems().attr('tabIndex', -1);
+    $appointment.attr('tabIndex', this._collection.option('tabIndex'));
+  }
+
+  private _tabHandler(e): void {
```
aleksei-semikozov: The leading underscore is redundant here — all three private methods (`_tabHandler`, `_delHandler`, `_escHandler`) already have the `private`. The `_` prefix is a JS-era convention and shouldn't be mixed with TS access modifiers.

```suggestion
private tabHandler(e): void {
```

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
```diff
@@ -0,0 +1,118 @@
...
+export class AppointmentsKeyboardNavigation {
+  private readonly _collection: SchedulerAppointments;
+
+  public $focusedItem: dxElementWrapper | null = null;
+
+  constructor(collection: SchedulerAppointments) {
+    this._collection = collection;
+  }
+
+  public $focusableItems(): dxElementWrapper {
```
aleksei-semikozov: 1. **Missing `get`** — this method takes no arguments and returns a computed value, which is a classic getter pattern.
2. **`$` prefix** — the `$` convention signals a jQuery object. Since the return type is `dxElementWrapper` (not jQuery), 
```suggestion
public get focusableItems(): dxElementWrapper {
```

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
```diff
@@ -0,0 +1,118 @@
...
+    const collectors = this._collection.$element().find('.dx-scheduler-appointment-collector');
+
+    return appts.add(collectors);
+  }
+
+  public $focusableItemBySortedIndex(sortedIndex: number): dxElementWrapper {
+    const $items = this.$focusableItems();
+
+    // @ts-expect-error
+    return $items.filter((_, itemElement: Element) => {
```
aleksei-semikozov: Can we remove `@ts-expect-error`? The `dxElementWrapper.filter()` callback likely expects `(index: number, element: HTMLElement) => boolean`, but `Element` is passed instead. Let's fix the types rather than silence the error:

```suggestion
    return $items.filter((_index: number, itemElement: HTMLElement) => {
```

packages/devextreme/js/__internal/scheduler/appointments/m_appointment_collection.ts
```diff
@@ -48,11 +47,11 @@ import type {
 import { AgendaAppointment } from './appointment/agenda_appointment';
 import { Appointment } from './appointment/m_appointment';
 import { createAgendaAppointmentLayout, createAppointmentLayout } from './m_appointment_layout';
+import { AppointmentsKeyboardNavigation } from './m_appointmetns_kbn';
```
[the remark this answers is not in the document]
aleksei-semikozov: Agree

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
Tucchhaa: Nope, it's needed here. Otherwise there's a type error. .filter() callbacks expects a `selector: string`.

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
Tucchhaa: 1. Usually getters return some private fields, I think `$focusableItems` has quite a complicated logic, so I made it a function. Anyway, I think it's a minor comment. If you insist I can replace it to a getter

2. In devextreme we don't directly use jQuery types, instead we use dxElementWrapper type, which implements the same functionality as jQuery. That is why I added `$` here. If you take a look at the function impl, it actually returns jQuery object

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
Tucchhaa: On one side, I agree with you. On another side having leading `_`, makes it very obvious that the method is private when reading a code. Otherwise, you would need to navigate to the method to see its modifier.

In my perspective this is a minor comment, so I would like to skip it. But if you insist I can remove the prefixes

packages/devextreme/js/__internal/scheduler/appointments/m_appointment_collection.ts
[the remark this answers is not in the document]
Tucchhaa: fixed

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
aleksei-semikozov: 1. This name is inconvenient to write and copy, so I would rewrite it. Or at least rename it to getFocusableItems() if it is a method, because the name does not reflect what it actually is.
2. If jQuery is present on the page, then yes. If not, we use our own implementation.

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
Tucchhaa: Done

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
Tucchhaa: done

packages/devextreme/js/__internal/scheduler/appointments/m_appointmetns_kbn.ts
Tucchhaa: Done

sjbur (approved):

PR #32525, opened 2026-02-11
DataGrid: fix skipped qunit tests

anna-shakhova:

### DIFF
--- a/packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/columnFixing.integration.tests.js
+++ b/packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/columnFixing.integration.tests.js
@@ -130,8 +130,7 @@ QUnit.module('Fixed columns', baseModuleConfig, () => {
         assert.notOk($(dataGrid.getCellElement(1, 1)).hasClass('dx-col-fixed'), 'dx-col-fixed');
     });
 
-    // TODO jsdmitry: wait fix T381435
-    QUnit.skip('Columns hiding - do not hide fixed columns', function(assert) {
+    QUnit.test('Columns hiding - do not hide fixed columns', function(assert) {
         // arrange
         $('#container').width(150);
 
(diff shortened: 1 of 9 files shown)


packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/columnFixing.integration.tests.js
```diff
@@ -130,46 +130,52 @@ QUnit.module('Fixed columns', baseModuleConfig, () => {
...
         $('#container').width(150);
 
         const dataGrid = $('#dataGrid').dxDataGrid({
             loadingTimeout: null,
             columnHidingEnabled: true,
             dataSource: [{ firstName: 'Blablablablablablablablablabla', lastName: 'Psy', age: 40 }],
-            columns: [{ dataField: 'firstName', fixed: true, fixedPosition: 'left' }, 'lastName', 'age'],
+            columns: [
+                { dataField: 'firstName', fixed: true, fixedPosition: 'left' },
+                { dataField: 'lastName', hidingPriority: 0 },
```
Alyar666, line 143: Are we sure about these changes? If you specify the hidingPriority property for at least one column, the default hiding priorities are canceled. In this case, only one column will be hidden - lastName. The adaptive behavior will not apply to other columns.

Alyar666 (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/columnFixing.integration.tests.js
anna-shakhova: Yep, I've changed the scenario here. This test was skipped for more than 8 years, looks like core logic was changed since then. Test expected that second column will be hidden before 3rd.
But is our docs: Each column has a unique default hiding priority. The rightmost column has the priority of 0. This value is incremented by 1 for columns from right to left; the column with the lowest priority is hidden first.

So I decided to add hidingPriority here to test this case. Without that property both columns were hidden, and as I remember, this case was already covered above

Alyar666 (approved):

PR #32529, opened 2026-02-11
Chat: implement send and stt buttons state update

Zedwag:


packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -72,10 +75,29 @@ const ERRORS = {
   fileLimit: messageLocalization.format('dxChat-fileLimitReachedWarning', MAX_ATTACHMENTS_COUNT),
 };
 
-const STT_INITIAL_STATE = {
-  stylingMode: 'text' as const,
+interface ButtonState {
```
marker-dao: It would be better if all types were above constants in the code, because a constant may require a specific type in advance.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -72,10 +75,29 @@ const ERRORS = {
...
+  disabled?: boolean;
+  stylingMode: ButtonStyle;
+  type: ButtonType;
+}
+
+export const STT_INITIAL_STATE: ButtonState = {
+  stylingMode: 'text',
+  type: 'normal',
+};
+export const STT_LISTENING_STATE: ButtonState = {
```
marker-dao, line 89: Add spaces between constants.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -299,10 +323,14 @@ class ChatTextArea extends TextArea<Properties> {
       hoverStateEnabled,
       ...speechToTextOptions,
       ...STT_INITIAL_STATE,
+      startIcon: 'micoutline',
+      stopIcon: 'micfilled',
       onEnd: (e: EndEvent): void => {
         this._initialInputText = '';
 
         speechToTextOptions?.onEnd?.(e);
+        this._speechToTextIsListening = false;
```
marker-dao: onEnd is called after the SR Engine has finished working, i.e., the SR API has already returned to its normal state https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition#events , so it makes sense to follow the same principle: first perform all actions in our component, and then execute the user callback onEnd.

The same applies to other calls.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -503,9 +536,23 @@ class ChatTextArea extends TextArea<Properties> {
     this._addFileToMap(file);
   }
 
-  _toggleButtonDisableState(state?: boolean): void {
-    const shouldDisable = state ?? !this._isMessageCanBeSent();
-    this._sendButton?.option('disabled', shouldDisable);
+  _updateButtonsState(): void {
+    if (this._speechToTextIsListening === true && this._speechToTextButton) {
```
marker-dao:
```suggestion
    if (this._speechToTextIsListening === true && speechToTextEnabled) {
```

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -503,9 +536,23 @@ class ChatTextArea extends TextArea<Properties> {
...
+      this._speechToTextButton?.option(STT_LISTENING_STATE);
```
marker-dao, line 547: You have already checked this above.
```suggestion
      this._speechToTextButton.option(STT_LISTENING_STATE);
```

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -105,6 +127,8 @@ class ChatTextArea extends TextArea<Properties> {
 
   _speechToTextButton?: dxSpeechToText;
 
+  _speechToTextIsListening?: boolean;
```
marker-dao: Non-standard order: boolean variables usually start with is/has/should.

packages/devextreme/js/__internal/ui/chat/message_box/chat_text_area.ts
```diff
@@ -569,6 +616,9 @@ class ChatTextArea extends TextArea<Properties> {
 
       case 'speechToTextEnabled':
         this._toolbar?.option({ items: this._getToolbarItems() });
+        this._speechToTextButton = undefined;
```
marker-dao, line 625: Let's clean the link also in _dispose, to avoid memory leaks.

marker-dao (approved):

PR #32570, opened 2026-02-13
Scheduler: fix timeline demos with too wide cells

sjbur:


apps/demos/Demos/Scheduler/CurrentTimeIndicator/Vue/styles.css
```diff
@@ -0,0 +1,3 @@
+.dx-scheduler-timeline .dx-scheduler-cell-sizes-horizontal {
```
Tucchhaa: I suggest to move this styles to CurrentTimeIndicator/Vue/App.vue file and remove changes from index.html

Tucchhaa (approved):

PR #32583, opened 2026-02-16
PivotGrid - ♿️ Handle Accessibility restrictions

aleksei-semikozov:

### DIFF
--- a/packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
+++ b/packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
@@ -121,16 +121,8 @@ export class FieldChooserBase extends mixinWidget {
     };
   }
 
-  protected _setAriaSortAttribute(column, ariaSortState, $rootElement) {
-    $rootElement.removeAttr('aria-roledescription');
-
-    if (column.isGrouped && isDefined(column.sortOrder)) {
-      const description = column.sortOrder === 'asc'
-        ? this.localize('dxDataGrid-ariaSortedAscendingColumn')
-        : this.localize('dxDataGrid-ariaSortedDescendingColumn');
-      this.setAria('roledescription', description, $rootElement);
-    }
-  }
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected _setAriaSortAttribute(_column, _ariaSortState, _$rootElement) { }
 
   _init() {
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
```diff
@@ -121,6 +121,17 @@ export class FieldChooserBase extends mixinWidget {
...
+  protected _setAriaSortAttribute(column, ariaSortState, $rootElement) {
+    $rootElement.removeAttr('aria-roledescription');
+
+    if (column.isGrouped && isDefined(column.sortOrder)) {
+      const description = column.sortOrder === 'asc'
+        ? this.localize('dxDataGrid-ariaSortedAscendingColumn')
+        : this.localize('dxDataGrid-ariaSortedDescendingColumn');
+      this.setAria('roledescription', description, $rootElement);
+    }
+  }
```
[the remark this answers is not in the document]
aleksei-semikozov: The correctness of aria attribute setting is confirmed in the demo

sjbur (dismissed):

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
```diff
@@ -121,6 +121,17 @@ export class FieldChooserBase extends mixinWidget {
     };
   }
 
+  protected _setAriaSortAttribute(column, ariaSortState, $rootElement) {
+    $rootElement.removeAttr('aria-roledescription');
+
+    if (column.isGrouped && isDefined(column.sortOrder)) {
```
Tucchhaa: PivotGrid does not have column grouping, so this if statement will never pass. 

As I understand we just need to prevent parent's _setAriaSortAttribute method to set the aria-sort attribute, so looks like we need make this function empty:

```js
protected _setAriaSortAttribute(column, ariaSortState, $rootElement) { }
```

Tucchhaa (changes requested):

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
aleksei-semikozov: fixed

Tucchhaa (approved):

PR #32584, opened 2026-02-16
PivotGrid - Fix unstable tests - QUnit - dataController.tests.js, pivotgrid.tests.js, fieldsArea.tests.js

aleksei-semikozov:

### DIFF
--- a/packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/dataController.tests.js
+++ b/packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/dataController.tests.js
@@ -6709,47 +6709,46 @@ QUnit.module('Remote paging', {
     });
 
     QUnit.test('load with CustomStore', function(assert) {
-        const done = assert.async();
         const that = this;
+
+        const CustomPivotStore = Class.inherit({
+            ctor: function() {},
+            getFields: function() {
+                return $.Deferred().resolve([]).promise();
+            },
+            supportPaging: function() {
+                return true;
+            },
+            load: function(loadOptions) {
+                that.loadArgs.push(loadOptions);
+                return $.Deferred().resolve({
+                    rows: [
+                        { value: 'row 1', text: 'row 1', index: 1 },
+                        { value: 'row 2', text: 'row 2', index: 2 }
+                    ],
+                    columns: [],
+                    values: [[null], [null], [null]],
+                    grandTotalRowIndex: 0,
+                    grandTotalColumnIndex: 0
+                }).promise();
+            }
+        });
 
         const dataController = that.setup({
             paginate: true,
-            store: {
-                load: function(loadOptions) {
-                    that.loadArgs.push(loadOptions);
-                    return $.Deferred().resolve({
-                        rows: [
-                            { value: 'row 1', text: 'row 1', index: 1 },
-                            { value: 'row 2', text: 'row 2', index: 2 }
-                        ],
-                        columns: [],
-                        values: [[null], [null], [null]],
-                        grandTotalRowIndex: 0,
-                        grandTotalColumnIndex: 0
-                    });
-                },
-                getFields: function() {
-                    return $.Deferred().resolve([]);
-                },
-                supportPaging: function() {
-                    return true;
-                }
-            },
+            store: new CustomPivotStore(),
             fields: [{ dataField: 'row', area: 'row' }, { area: 'data' }]
         });
 
-        setTimeout(function() {
-            assert.strictEqual(that.loadArgs.length, 1, 'one load');
-            assert.ok(that.loadArgs[0].rows, 'has rows in load options');
-            assert.strictEqual(that.loadArgs[0].rowSkip, 0, 'rowSkip is 0');
-            assert.strictEqual(that.loadArgs[0].rowTake, 2, 'rowTake is 2');
+        assert.strictEqual(that.loadArgs.length, 1, 'one load');
+        assert.ok(that.loadArgs[0].rows, 'has rows in load options');
+        assert.strictEqual(that.loadArgs[0].rowSkip, 0, 'rowSkip is 0');
+        assert.strictEqual(that.loadArgs[0].rowTake, 2, 'rowTake is 2');
 
-            assert.deepEqual(dataController.getRowsInfo(), [
-                [{ dataSourceIndex: 1, text: 'row 1', path: ['row 1'], type: 'D', isLast: true }],
-                [{ dataSourceIndex: 2, text: 'row 2', path: ['row 2'], type: 'D', isLast: true }]
-            ]);
-            done();
-        }, 0);
+        assert.deepEqual(dataController.getRowsInfo(), [
+            [{ dataSourceIndex: 1, text: 'row 1', path: ['row 1'], type: 'D', isLast: true }],
+            [{ dataSourceIndex: 2, text: 'row 2', path: ['row 2'], type: 'D', isLast: true }]
+        ]);
     });
 
 });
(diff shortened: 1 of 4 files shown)


packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/pivotGrid.tests.js
```diff
@@ -6199,10 +6199,16 @@ QUnit.module('Vertical headers', {
...
+        this.clock.tick(10);
 
         const columnsWidth = grid._columnsArea.getColumnsWidth();
         const dataWidth = grid._dataArea.getColumnsWidth();
-        assert.deepEqual(columnsWidth, dataWidth);
+
+        assert.strictEqual(columnsWidth.length, dataWidth.length, 'arrays have same length');
+
+        for(let i = 0; i < columnsWidth.length; i++) {
+            assert.roughEqual(columnsWidth[i], dataWidth[i], 2, `column ${i} width`);
```
Tucchhaa: Looks strange to me that column cells and data cells may have different widths. This assert contradicts the test name. Maybe we should try to find the source of the issue and fix it?

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/fieldsArea.tests.js
```diff
@@ -225,17 +256,25 @@ QUnit.module('Hamburger', {
...
         this.renderArea();
         const button = $(this.area.tableElement()[0].rows[0].cells[0]).children(0).dxButton('instance');
-
-        button.element().trigger('dxclick');
         const popup = this.area.tableElement().find('.dx-fields-area-popup').dxPopup('instance');
 
-        assert.ok(popup);
-        assert.ok(popup.option('visible'));
-        assert.strictEqual(popup.content().find('.dx-pivotgrid-fields-area-head').length, 1);
-        assert.strictEqual(popup.content().find('.dx-pivotgrid-fields-area-head').find('.dx-area-field').length, 2);
```
Tucchhaa, line 238: I tried to run this test locally. I have noticed that popup's content rendered with some delay. Maybe if we add wait time after button click to make sure that popup's content is rendered?

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/dataController.tests.js
```diff
@@ -6708,34 +6708,42 @@ QUnit.module('Remote paging', {
...
+        const dataController = that.setup({
+            paginate: true,
+            store: new CustomPivotStore(),
+            fields: [{ dataField: 'row', area: 'row' }, { area: 'data' }]
+        });
+
+        assert.strictEqual(that.loadArgs.length, 1, 'one load');
+        assert.ok(that.loadArgs[0].rows, 'has rows in load options');
+        assert.strictEqual(that.loadArgs[0].rowSkip, 0, 'rowSkip is 0');
+        assert.strictEqual(that.loadArgs[0].rowTake, 2, 'rowTake is 2');
```
Tucchhaa: Can we use deepEqual with all fields of `that.loadArgs[0]` here like it was in the previous test?

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/pivotGrid.tests.js
aleksei-semikozov: This is sub-pixel rounding because of CSS zoom 1.35. Browser rounds values differently for columns area and data area elements. Max difference I found is ~0.5px. This is browser behavior, not PivotGrid bug. Changed tolerance to 0.5px.

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/fieldsArea.tests.js
aleksei-semikozov: I tried clock.tick() approach but it doesn't work.

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/dataController.tests.js
aleksei-semikozov: It's looked as
```
       assert.deepEqual(that.loadArgs[0], {
            reload: true,
            rows: [{
                dataField: 'row',
                area: 'row',
                index: 0,
                _initProperties: { caption: undefined },
                caption: 'Row',
                areaIndex: 0
            }],
            columns: [],
            values: [{
                area: 'data',
                index: 1,
                _initProperties: { caption: undefined },
                caption: '',
                areaIndex: 0
            }],
            filters: [],
            columnExpandedPaths: [],
            rowExpandedPaths: [],
            pageSize: 40,
            rowSkip: 0,
            rowTake: 2,
            columnSkip: 0,
            columnTake: 2
        }, 'load args');

```

too noisli for me

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/pivotGrid.tests.js
Tucchhaa: ok, thank you

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/fieldsArea.tests.js
Tucchhaa: Hm, strange. Maybe try to make this test an integration test?  I see there's filedChooser.markup.tests.js file, which contains integration tests

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/dataController.tests.js
Tucchhaa: How about we check deepEqual with this?:

```
            rows: [{
                dataField: 'row',
            }],
            columns: [],
            values: [{
                area: 'data',
            }],
            filters: [],
            columnExpandedPaths: [],
            rowExpandedPaths: [],
            pageSize: 40,
            rowSkip: 0,
            rowTake: 2,
            columnSkip: 0,
            columnTake: 2
```

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/fieldsArea.tests.js
aleksei-semikozov: Found the root cause! When contentTemplate returns a <table> element directly, normalizeTemplateElement in DevExtreme extracts tbody.contents() from it:

https://github.com/DevExpress/DevExtreme/blob/25_1/packages/devextreme/js/__internal/core/utils/m_dom.ts#L96-L97

```
} else if ($element.is('table')) {
  $element = $element.children('tbody').contents();
}
```
Since our table only has \<thead> (no \<tbody>), the content was being stripped to empty. Fixed by wrapping the table in a \<div>.

Now the test actually verifies that popup content renders correctly after click, not just that contentTemplate function exists.

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/dataController.tests.js
aleksei-semikozov: Added assertions for all the fields you requested.

deepEqual requires strict equality of entire objects - since loadOptions contains additional internal fields (_initProperties, caption etc.), we can't use it directly with a partial expected object

Tucchhaa (approved):

PR #32585, opened 2026-02-16
Form: publish getScrollable() method in .d.ts

pharret31:


EugeniyKiyashko (dismissed):

packages/devextreme/js/ui/form.d.ts
```diff
@@ -390,6 +391,13 @@ export default class dxForm extends Widget<dxFormOptions> {
      * @public
      */
     getEditor(dataField: string): Editor | undefined;
+    /**
+     * @docid
+     * @publicName getScrollable()
+     * @public
+     * @return Scrollable | undefined
```
marker-dao: We don't need to specify the return type here. This should only be done when wrappers cannot be generated correctly. Try without it. Most likely, nothing will change in all.d.ts and wrappers.

EugeniyKiyashko (approved):

marker-dao (approved):

PR #32622, opened 2026-02-18
Scheduler - Resource's icon option does not support FontAwesome/SVG (T1321514)

aleksei-semikozov:


packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1186,6 +1186,33 @@ describe('Appointment Form', () => {
...
+          icon: 'fas fa-home',
+          dataSource: [{ text: 'Room 1', id: 1 }, { text: 'Room 2', id: 2 }],
+        }],
+      });
+      const dataSource = (scheduler as any).getDataSource();
+      const appointment = dataSource.items()[0];
+
+      scheduler.showAppointmentPopup(appointment);
+
+      const $iconElement = $('.dx-scheduler-form-resources-group .dx-icon');
```
Tucchhaa: Maybe move it to POM?
We have [PopupModel.subjectIcon](https://github.com/DevExpress/DevExtreme/blob/ff712543770d5f6c39c71a33e0f9d854a552d679/packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/popup.ts#L43), so I think we could add something like resourceIcon

Tucchhaa (dismissed):

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
sjbur: +1, makes sense in this case

sjbur (dismissed):

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: fixed

Tucchhaa (approved):

PR #32626, opened 2026-02-19
Scheduler - Refactor ViewModel before getOccurrence() implementation

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/m_scheduler.ts
+++ b/packages/devextreme/js/__internal/scheduler/m_scheduler.ts
@@ -275,7 +275,7 @@ class Scheduler extends SchedulerOptionsBaseWidget {
         this._updateOption('header', name, dateValue);
         this._updateOption('header', 'startViewDate', this.getStartViewDate());
         this._appointments.option('items', []);
-        this._filterAppointmentsByDate();
+        this._setRemoteFilterIfNeeded();
 
         this._postponeDataSourceLoading();
         break;
@@ -286,7 +286,7 @@ class Scheduler extends SchedulerOptionsBaseWidget {
 
         this._postponeResourceLoading().done(() => {
           this.appointmentDataSource.setDataSource(this._dataSource);
-          this._filterAppointmentsByDate();
+          this._setRemoteFilterIfNeeded();
           this._updateOption('workSpace', 'showAllDayPanel', this.option('showAllDayPanel'));
         });
         break;
@@ -318,7 +318,7 @@ class Scheduler extends SchedulerOptionsBaseWidget {
         this._postponeResourceLoading().done(() => {
           this._refreshWorkSpace();
           this._header?.option(this._headerConfig());
-          this._filterAppointmentsByDate();
+          this._setRemoteFilterIfNeeded();
           this._appointments.option('allowAllDayResize', value !== 'day');
         });
         // NOTE:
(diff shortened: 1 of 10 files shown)


packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -580,38 +580,26 @@ class Scheduler extends SchedulerOptionsBaseWidget {
     return this.option('showAllDayPanel') && this._layoutManager.hasAllDayAppointments();
   }
 
-  _filterAppointmentsByDate() {
-    if (!this._workSpace) {
+  _setRemoteFilterIfNeeded(): void {
```
aleksei-semikozov, line 583: lease remove the underscore prefix and use TypeScript private modifier instead:


// Current:
_setRemoteFilterIfNeeded(): void {

// Should be:
private setRemoteFilterIfNeeded(): void {

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/integration.dateNavigator.tests.js
```diff
@@ -381,13 +381,13 @@ QUnit.module('Integration: Date navigator', moduleConfig, function() {
     QUnit.test('Tasks should be rerendered after click on next/prev button', async function(assert) {
```
aleksei-semikozov, line 381: Current test spies on internal method _setRemoteFilterIfNeeded - this tests implementation, not behavior. Better approach: use CustomStore and verify that loadOptions.filter is passed correctly to load(). This way the test won't break on refactoring.


like this, but i think better remove this test from Qunit and make it in jest
```
QUnit.test('Remote filter should be passed to CustomStore.load on date navigation', async function(assert) {
    let capturedLoadOptions;
    
    const customStore = new CustomStore({
        key: 'id',
        load: (loadOptions) => {
            capturedLoadOptions = loadOptions;// or make sinin spy and call him with loadOption
            return [];
        },
    });
    
    await this.createInstance({
        currentDate: new Date(2015, 1, 24),
        dataSource: { store: customStore },
        remoteFiltering: true,
    });
    
    capturedLoadOptions = null;  // reset after initial load
    
    $('.dx-scheduler-navigator-previous').trigger('dxclick');
    
    assert.ok(capturedLoadOptions?.filter, 'filter was passed to load');
    assert.ok(Array.isArray(capturedLoadOptions.filter), 'filter is an array');
});
```

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -580,38 +580,26 @@ class Scheduler extends SchedulerOptionsBaseWidget {
...
+    if (!this._workSpace || !remoteFiltering || !dataSource) {
       return;
     }
 
     const dateRange = this._workSpace.getDateRange();
-
     const startDate = this.timeZoneCalculator.createDate(dateRange[0], 'fromGrid');
     const endDate = this.timeZoneCalculator.createDate(dateRange[1], 'fromGrid');
 
-    this.setRemoteFilter(
```
aleksei-semikozov, line 593: The behavior has changed - two methods (_filterAppointmentsByDate + setRemoteFilter) were merged into one with combined early return logic. The only test change is renaming the spy. Please add tests

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
sjbur: I was also thinking about this. +1

aleksei-semikozov (approved):

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/integration.dateNavigator.tests.js
Tucchhaa: Discussed in DM. I will create a separate issue for refactoring such tests

PR #32633, opened 2026-02-19
Scheduler: delete Appointment enhancements

sjbur:


packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/scheduler.ts
```diff
@@ -55,6 +57,59 @@ export class SchedulerModel {
...
+    const keyEvent = new KeyboardEvent('keydown', {
+      key: 'Delete',
+      bubbles: true,
+    });
+    const keyboardOptions = {
+      keyName: 'del',
+      originalEvent: keyEvent,
+    };
+    listInstance._keyboardHandler(keyboardOptions);
+  }
```
Tucchhaa: I think we don't need this method. It introduces abstraction for simple user actions: like focus item and pressing delete. So tests with this function become a little unclear.

Moreover, focusing item using `list.option()` is not preferred in the integration tests. It's better to use simulate 'tab' presses or 'focus' events.

I think this `keydown` helper function ([here](https://github.com/DevExpress/DevExtreme/blob/84f2e6ca9030d008189f32bfbafc57957e56935a/packages/devextreme/js/__internal/scheduler/__tests__/__mock__/create_scheduler.ts#L23)) in scheduler's model may help you.

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/scheduler.ts
```diff
@@ -55,6 +57,59 @@ export class SchedulerModel {
...
+
+    return collectors[index];
+  }
+
+  isTooltipVisible(): boolean {
+    return document.querySelector('.dx-overlay-wrapper.dx-scheduler-appointment-tooltip-wrapper') !== null;
+  }
+
+  getTooltipDeleteButton(index = 0): HTMLElement {
+    const buttons = document.querySelectorAll<HTMLElement>('.dx-tooltip-appointment-item-delete-button');
```
Tucchhaa: Can we use testing library to get elements?

packages/devextreme/js/__internal/scheduler/__tests__/appointment_tooltip.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      currentView: 'month',
+      currentDate: new Date(2017, 4, 22),
+      height: 600,
+    });
+
+    POM.getCollectorButton().click();
+    POM.pressDeleteOnTooltipItem();
+
+    expect(POM.isTooltipVisible()).toBe(true);
+  });
```
Tucchhaa, line 165: Please add tests for these scenarios:
- Delete button is not focusable using tab
- Appointment can't be deleted when editing.allowDeleting=false
- Appointment can be deleted on delete icon click (probably there's already a qunit test for that, but it would better if we had it in jest too)

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/scheduler.ts
```diff
@@ -55,6 +57,59 @@ export class SchedulerModel {
     return getTexts(collectors);
   }
 
+  getCollectorButton(index = 0): HTMLElement {
```
Tucchhaa, line 63: Can we create a separate file for tooltip's POM? And in scheduler POM just have a function `getTooltip()` or something like that

packages/devextreme/js/__internal/scheduler/__tests__/collector_tooltip.test.ts
```diff
@@ -0,0 +1,73 @@
+import {
+  afterEach, beforeEach, describe, expect, it, jest,
+} from '@jest/globals';
+import fx from '@js/common/core/animation/fx';
+
+import { createScheduler } from './__mock__/create_scheduler';
+import { setupSchedulerTestEnvironment } from './__mock__/m_mock_scheduler';
+
+describe('Collector tooltip behavior', () => {
```
Tucchhaa: Can we rename it to: 'Appointment Tooltip' as this term is used scheduler instead of 'Collector Tooltip'? also rename the file too, pls

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/scheduler.ts
sjbur: I tried just calling keydown event on tooltip / list for focus, but it did not work. That's why I tried to use _keyboardHandler on the list instance. But I agree that having a separate method is not the best idea.

packages/devextreme/js/__internal/scheduler/__tests__/appointment_tooltip.test.ts
sjbur: OK. I will add

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/scheduler.ts
sjbur: Sure

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/scheduler.ts
sjbur: Sure

packages/devextreme/js/__internal/scheduler/__tests__/collector_tooltip.test.ts
sjbur: OK

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/scheduler.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/__tests__/collector_tooltip.test.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/__tests__/appointment_tooltip.test.ts
sjbur: Done

aleksei-semikozov (dismissed):

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/scheduler.ts
sjbur: I tried one more time and managed to get it working. I need to call keydown event on .dx-scrollview-content of .dx-scrollable element.

Removed the method and list logic as well

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/scheduler.ts
sjbur: I thought it would be easier. Where I could, I used Testing Library, but no everywhere.

packages/devextreme/js/__internal/scheduler/__tests__/appointment_tooltip.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      },
+    });
+
+    POM.getCollectorButton().click();
+
+    const tooltipScrollableContent = POM.tooltip.getScrollableContent();
+    tooltipScrollableContent?.dispatchEvent(new FocusEvent('focusin', { bubbles: true }));
+    tooltipScrollableContent?.dispatchEvent(new KeyboardEvent('keydown', { key: 'Delete', bubbles: true }));
+
+    expect(POM.tooltip.isVisible()).toBe(true);
```
Tucchhaa: can we check that appointment is not deleted from dataSource instead of checking that tooltip is visible? I thought we have a task, to try not hide tooltip on appt delete. So, it would be less problems for us in the future if we remove this visibility check

packages/devextreme/js/__internal/scheduler/__tests__/appointment_tooltip.test.ts
```diff
@@ -0,0 +1,162 @@
...
+      height: 600,
+    });
+
+    POM.getCollectorButton().click();
+
+    const tooltipScrollableContent = POM.tooltip.getScrollableContent();
+    tooltipScrollableContent?.dispatchEvent(new FocusEvent('focusin', { bubbles: true }));
+    tooltipScrollableContent?.dispatchEvent(new KeyboardEvent('keydown', { key: 'Delete', bubbles: true }));
+
+    expect(POM.tooltip.isVisible()).toBe(true);
```
Tucchhaa: Can we check datasource data instead of tooltip visibility here too?

packages/devextreme/js/__internal/scheduler/__tests__/appointment_tooltip.test.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/__tests__/appointment_tooltip.test.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/__tests__/appointment_tooltip.test.ts
```diff
@@ -127,24 +129,26 @@ describe('Appointment tooltip behavior', () => {
     tooltipScrollableContent?.dispatchEvent(new FocusEvent('focusin', { bubbles: true }));
     tooltipScrollableContent?.dispatchEvent(new KeyboardEvent('keydown', { key: 'Delete', bubbles: true }));
 
-    expect(POM.tooltip.isVisible()).toBe(true);
+    expect(scheduler.option('dataSource')).toEqual(dataSource);
```
Tucchhaa: This condition will always be true, because `scheduler.option('dataSource')` and `dataSource` are the same object.

Tucchhaa: This assertion will always pass because `scheduler.option('dataSource')` and `dataSource` reference the same object.

Instead, please verify the actual data inside the data source (by checking the items returned by it).

See:
- [Scheduler.getDataSource()](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/Methods/#getDataSource)
- [DataSource.items()](https://js.devexpress.com/jQuery/Documentation/ApiReference/Data_Layer/DataSource/Methods/#items)
- [Example test](https://github.com/DevExpress/DevExtreme/blob/cd20f51877d54a28e314bdee1d5d679a651cbcad/packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts#L75)

sjbur: Done

Tucchhaa (approved):

aleksei-semikozov (approved):

PR #32673, opened 2026-02-23
Scheduler: Add testing library helper

aleksei-semikozov:
Cherry-pick of #32516 into 25_1.


sjbur (approved):

sjbur (changes requested):
Fix tests, please

packages/devextreme/js/__internal/scheduler/__tests__/__mock__/model/popup.ts
```diff
@@ -1,14 +1,92 @@
+import type { dxElementWrapper } from '@js/core/renderer';
```
Tucchhaa: 25.1 doesn't have new appointment form, so many added methods and properties are not relevant.

Like: mainGroup, recurrenceGroup, subjectIcon, backButton, etc. I suggest to revert changes from popup.ts for 25.1

sjbur (approved):

PR #32677, opened 2026-02-24
Scheduler - getOccurrences() - Implement d.ts

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/ui/scheduler.d.ts
+++ b/packages/devextreme/js/ui/scheduler.d.ts
@@ -1286,8 +1286,6 @@ export default class dxScheduler extends Widget<dxSchedulerOptions> {
     /**
      * @docid
      * @publicName getOccurrences(startDate, endDate, appointments)
-     * @param1 startDate:Date
-     * @param2 endDate:Date
      * @param3 appointments:Array<dxSchedulerAppointment>
      * @public
      */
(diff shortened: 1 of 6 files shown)


packages/devextreme/ts/dx.all.d.ts
```diff
@@ -26359,6 +26367,24 @@ declare module DevExpress.ui {
      */
     visible?: boolean;
   } & Record<string, any>;
+  /**
+   * @deprecated Use the Scheduler's Occurrence type instead
+   * @deprecated Attention! This type is for internal purposes only. If you used it previously, please submit a ticket to our {@link https://supportcenter.devexpress.com/ticket/create Support Center}. We will check if there is an alternative solution.
```
aleksei-semikozov: I think this line doesn't make any sense. I think the first line is enough.

packages/devextreme/js/ui/scheduler.d.ts
```diff
@@ -1283,6 +1283,15 @@ export default class dxScheduler extends Widget<dxSchedulerOptions> {
      * @public
      */
     getEndViewDate(): Date;
+    /**
+     * @docid
+     * @publicName getOccurrences(startDate, endDate, appointments)
+     * @param1 startDate:Date
+     * @param2 endDate:Date
+     * @param3 appointments:Array<dxSchedulerAppointment>
```
aleksei-semikozov, line 1289: I think, the description of these parameters is not necessary in this case

packages/devextreme/ts/dx.all.d.ts
Tucchhaa: This if file is auto-generated by `pnpm run regenerate-all` command and we can't edit it

packages/devextreme/js/ui/scheduler.d.ts
Tucchhaa: All other methods in the d.ts have these descriptions, so I written them too. Let's wait for API reviewers review, I think they can clarify this

mpreyskurantov (approved):

sjbur (approved):

PR #32682, opened 2026-02-24
Scheduler: refactor Shaders module

sjbur:


Tucchhaa (dismissed):

packages/devextreme/js/__internal/scheduler/shaders/m_current_time_shader_vertical.ts
```diff
@@ -33,34 +33,57 @@ class VerticalCurrentTimeShader extends CurrentTimeShader {
     }
   }
 
-  _renderShaderParts(groupCount, shaderHeight, maxHeight, isSolidShader) {
```
Tucchhaa, line 36: I have a question. Don't we need to set modifiers: public, private, protected, and remove the '_' from the function name?

sjbur: Yes, will push changes soon

Tucchhaa (approved):

PR #32690, opened 2026-02-25
T1322380 - DataGrid - repaint grid layout after unfixing a column via the context menu

dmirgaev:


e2e/testcafe-devextreme/tests/dataGrid/common/fixedColumns/visual.ts
```diff
@@ -0,0 +1,384 @@
...
+    Department: 'Human Resources',
+    MobilePhone: '(818) 555-6546',
+  }],
+  keyExpr: 'ID',
+  columnAutoWidth: true,
+  showBorders: true,
+  repaintChangesOnly: true,
+  columnFixing: { enabled: true },
+  columns: [
+    {
```
anna-shakhova: Looks like there is no need in so many columns and unbound columns with calculateCellValue set
let's simplify configuration and dataSource

e2e/testcafe-devextreme/tests/dataGrid/common/fixedColumns/visual.ts
```diff
@@ -0,0 +1,384 @@
...
+// T1322380
+test('The grid layout should be correct after unfixing a column via the context menu', async (t) => {
+  const dataGrid = new DataGrid(DATA_GRID_SELECTOR);
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const contextMenu = dataGrid.getContextMenu();
+
+  await t.expect(dataGrid.isReady()).ok();
+
+  // act: fix the first 4 columns (Employee, Position, Address, City)
+  await ClientFunction(() => {
```
anna-shakhova: I've reproduced the issue when columns were fixed on initialization, not after 1st rendering. Looks like it is not important for case
Let's simplify test code and fix them from the start instead of using ClientFunction

anna-shakhova (approved):

PR #32694, opened 2026-02-25
DataGrid/TreeList: Split basic cell styles

Alyar666:


packages/devextreme-scss/scss/widgets/base/dataGrid/layout/cell.scss
```diff
@@ -0,0 +1,286 @@
...
+// (0,8,1)
+.dx-rtl .dx-datagrid .dx-datagrid-sticky-columns .dx-datagrid-content .dx-datagrid-table .dx-row.dx-column-lines > td.dx-datagrid-first-header {
+  border-right: none;
+}
+
+// (0,9,1)
+.dx-rtl .dx-datagrid .dx-datagrid-sticky-columns .dx-datagrid-content .dx-datagrid-table .dx-row.dx-column-lines > td.dx-datagrid-first-header.dx-datagrid-sticky-column-border-left {
+  border-left: 2px solid;
+  border-left-color: $datagrid-border-color;
+}
```
Raushen, line 215: Why we do not use scss? These rules can be refactored, now they are too heavy. There is the style duplication everywhere: 
`.dx-rtl .dx-datagrid .dx-datagrid-sticky-columns .dx-datagrid-content .dx-datagrid-table .dx-row.dx-column-lines...`

Alyar666: > Why we do not use scss? These rules can be refactored, now they are too heavy. There is the style duplication everywhere: `.dx-rtl .dx-datagrid .dx-datagrid-sticky-columns .dx-datagrid-content .dx-datagrid-table .dx-row.dx-column-lines...`

I agree that we can combine styles into shared selectors in some places. But let's do that later. I’ve added it to my to-do list.

Raushen (approved):

PR #32703, opened 2026-02-26
Scheduler - Resolve Time Conflicts Demo

aleksei-semikozov:


apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.css
```diff
@@ -0,0 +1,57 @@
...
+::ng-deep .dx-scheduler-form-main-group .dx-item:last-child .dx-field-item-content,
+::ng-deep .dx-scheduler-form-main-group .dx-item:last-child .dx-item:last-child,
+::ng-deep .dx-scheduler-form-main-group .dx-item:last-child .dx-item:last-child .dx-field-item-content {
+  overflow: visible;
+}
+
+::ng-deep .dx-scheduler-form-recurrence-group.dx-scheduler-form-recurrence-group-hidden,
+::ng-deep .dx-scheduler-form-main-group.dx-scheduler-form-recurrence-group-hidden {
+  top: 41px;
+}
```
Tucchhaa: This style need to applied only when conflict informer is shown. Also I think this style is overriden by the default style of hidden group, so it is actually not applied.

P.S. I know that I provided this style as a solution, but it was only for reference since I disabled default style in the browser and wanted to quickly explain the solution. I should have mentioned that

Tucchhaa: Moreover, I have noticed that this doesn't work universally for all themes, because paddings between form items maybe different.

I would propose this solution:
1. make sure that height of the informer in all themes and both in main group and recurrence group is the same (just set height to the informer is enough)
2. Remove padding-top from . dx-scheduler-form-main-group and dx-scheduler-form-recurrence-group, because this padding is different in themes
3. Add some padding-bottom to the informer
4. Set the correct `top` value (top = informer height + informer padding-bottom)

And I think we would need to get an approve from designer

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Vue/App.vue
```diff
@@ -0,0 +1,329 @@
...
+function getEndDate(occurrence: DxSchedulerTypes.Occurrence): Date {
+  return (occurrence.appointmentData as Appointment).allDay
+    ? getNextDay(occurrence.startDate as Date)
+    : occurrence.endDate as Date;
+}
+
+function isOverlapping(a: DxSchedulerTypes.Occurrence, b: DxSchedulerTypes.Occurrence): boolean {
+  const aEnd = getEndDate(a);
+  const bEnd = getEndDate(b);
+  if ((a.startDate as Date) >= bEnd || (b.startDate as Date) >= aEnd) return false;
```
Tucchhaa: a.startDate already has type `Date`, the same with b.startDate. Please remove all such type conversions

apps/demos/Demos/Scheduler/ResolveTimeConflicts/React/App.tsx
```diff
@@ -0,0 +1,265 @@
...
+  </div>
+);
+
+const conflictInformerRender = () => (
+  <div className="conflict-informer">This time slot conflicts with another appointment.</div>
+);
+
+const App = () => {
+  const popupRef = useRef<NonNullable<PopupTypes.InitializedEvent['component']> | null>(null);
+  const formRef = useRef<NonNullable<FormTypes.InitializedEvent['component']> | null>(null);
```
Tucchhaa: Why not use `dxForm` and `dxPopup` types like in the Vue demo?: [link](https://github.com/DevExpress/DevExtreme/blob/cb45e8acaebfc7372e151b1e99ede18e99813b22/apps/demos/Demos/Scheduler/ResolveTimeConflicts/Vue/App.vue#L104)

apps/demos/Demos/Scheduler/ResolveTimeConflicts/React/App.tsx
```diff
@@ -0,0 +1,265 @@
...
+  const showConflictErrorRef = useRef(false);
+  const overlappingRuleRef = useRef('sameResource');
```
Tucchhaa, line 108: Minor: please add an empty line after this line, so that there would be a separation between variables and functions. It will slightly improve code's readability

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Vue/App.vue
```diff
@@ -0,0 +1,329 @@
...
+        message: 'Time conflict',
+        ignoreEmptyValue: true,
+        reevaluate: true,
+        validationCallback: () => !showConflictError,
+      },
+    ];
+  }
+};
+
+const onOverlappingRuleChanged = (e: any) => {
```
Tucchhaa: Can you replace `any` with the actual type?

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.html
```diff
@@ -0,0 +1,27 @@
...
+  [dataSource]="appointmentsData"
+  [views]="views"
+  currentView="week"
+  [startDayHour]="9"
+  [endDayHour]="19"
+  [currentDate]="currentDate"
+  [height]="600"
+  [showAllDayPanel]="false"
+  [resources]="resources"
+  [editing]="editingOptions"
```
Tucchhaa: I think it is better to define `resources` and `editing` using tags, like in React and Vue.

Here's [the example](https://github.com/DevExpress/DevExtreme/blob/cb45e8acaebfc7372e151b1e99ede18e99813b22/apps/demos/Demos/Scheduler/Templates/Angular/app/app.component.html#L21) of how it was done in Templates/Anguler demo.

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
```diff
@@ -0,0 +1,258 @@
...
+    icon: 'user',
+    allowMultiple: true,
+  }];
+
+  overlappingRuleItems = [
+    { value: 'sameResource', text: 'Allow across resources' },
+    { value: 'allResources', text: 'Disallow all overlaps' },
+  ];
+
+  private popup: any;
```
Tucchhaa: Can you replace usages of `any` in Angular demo with actual types? Here I would suggest to use `dxPopup`, like in [Vue demo](https://github.com/DevExpress/DevExtreme/blob/cb45e8acaebfc7372e151b1e99ede18e99813b22/apps/demos/Demos/Scheduler/ResolveTimeConflicts/Vue/App.vue#L105)

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
```diff
@@ -0,0 +1,258 @@
...
+                {
+                  name: 'assigneeId',
+                  isRequired: true,
+                  editorOptions: {
+                    onValueChanged: (e: any) => {
+                      if (e.value.length > 1) {
+                        e.component.option('value', [e.value[e.value.length - 1]]);
+                      }
+                    },
+                    tagTemplate: (tagData: Assignee) => {
```
Tucchhaa: I suggest to use *dxTemplate approach ([example](https://github.com/DevExpress/DevExtreme/blob/cb45e8acaebfc7372e151b1e99ede18e99813b22/apps/demos/Demos/Scheduler/Templates/Angular/app/app.component.html#L95)) instead of creating nodes manually in a function, because it is more readable

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
```diff
@@ -0,0 +1,258 @@
...
+  private getEndDate(occurrence: DxSchedulerTypes.Occurrence): Date {
+    return (occurrence.appointmentData as Appointment).allDay
+      ? this.getNextDay(occurrence.startDate)
+      : occurrence.endDate;
+  }
+
+  private isOverlapping(a: DxSchedulerTypes.Occurrence, b: DxSchedulerTypes.Occurrence): boolean {
+    const aEnd = this.getEndDate(a);
+    const bEnd = this.getEndDate(b);
+    if ((a.startDate) >= bEnd || (b.startDate) >= aEnd) return false;
```
Tucchhaa: Remove extra braces:
`a.startDate >= bEnd || b.startDate >= aEnd`

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
```diff
@@ -0,0 +1,258 @@
...
+                    },
+                  },
+                },
+              ],
+            },
+          ],
+        },
+        'recurrenceGroup',
+      ],
+      customizeItem: (item: any) => {
```
Tucchhaa: change `any` to actual type

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
```diff
@@ -0,0 +1,258 @@
...
+
+  onAppointmentAdding(e: DxSchedulerTypes.AppointmentAddingEvent): void {
+    this.alertConflictIfNeeded(e, e.appointmentData as Appointment);
+  }
+
+  onAppointmentUpdating(e: DxSchedulerTypes.AppointmentUpdatingEvent): void {
+    this.alertConflictIfNeeded(e, { ...e.appointmentData, ...e.newData } as Appointment);
+  }
+
+  onOverlappingRuleChanged(e: any): void {
```
Tucchhaa: change `any` to actual type

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const assignees: Assignee[] = [
+  { id: 1, text: 'Samantha Bright', color: '#A7E3A5' },
+  { id: 2, text: 'John Heart', color: '#CFE4FA' },
+  { id: 3, text: 'Todd Hoffman', color: '#F9E2AE' },
+  { id: 4, text: 'Sandra Johnson', color: '#F1BBBC' },
+];
+
+const appointments: Appointment[] = [
+  { id: 1, text: 'Website Re-Design Plan', startDate: new Date(2026, 1, 9, 9, 30), endDate: new Date(2026, 1, 9, 11, 30), assigneeId: [2] },
```
Tucchhaa: Can you format data objects such that each property is on the line, like in other demos?

apps/demos/Demos/Scheduler/ResolveTimeConflicts/React/App.tsx
aleksei-semikozov: This is a problem in Vue, I fixed it there. We cannot import devextreme components without the framework name in demos.
wrong
```
import dxForm from 'devextreme/ui/form.js';
```

apps/demos/Demos/Scheduler/ResolveTimeConflicts/React/App.tsx
aleksei-semikozov: fix

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Vue/App.vue
aleksei-semikozov: fix

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Vue/App.vue
aleksei-semikozov: fix

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.service.ts
aleksei-semikozov: fix

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.html
aleksei-semikozov: changed

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
aleksei-semikozov: fix

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
aleksei-semikozov: move to *dxTemplate

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
aleksei-semikozov: did it

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
aleksei-semikozov: did it

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
aleksei-semikozov: fix

apps/demos/Demos/Scheduler/ResolveTimeConflicts/jQuery/styles.css
```diff
@@ -0,0 +1,68 @@
...
+.option {
+  margin-top: 10px;
+  display: flex;
+  align-items: center;
+  gap: 10px;
+}
+
+#overlapping-rule {
+  width: 200px;
+}
```
Tucchhaa: I think we don't need this style. Other frameworks don't have it, and because of it the text in select box is cropped:

Image

apps/demos/Demos/Scheduler/ResolveTimeConflicts/jQuery/styles.css
```diff
@@ -0,0 +1,68 @@
...
+
+.dx-dialog:has(#conflict-dialog) .dx-toolbar-center,
+.dx-dialog:has(#conflict-dialog) .dx-button {
+  width: 100%;
+}
+
+.dx-scheduler-form-main-group .dx-item:last-child,
+.dx-scheduler-form-main-group .dx-item:last-child .dx-field-item-content {
+  overflow: visible;
+}
```
Tucchhaa: This selector is different from [React's](https://github.com/DevExpress/DevExtreme/blob/e3e7bef6d34e9a7fd1b590c50c513fe98a83a877/apps/demos/Demos/Scheduler/ResolveTimeConflicts/React/styles.css#L55) and [Vue's](https://github.com/DevExpress/DevExtreme/blob/e3e7bef6d34e9a7fd1b590c50c513fe98a83a877/apps/demos/Demos/Scheduler/ResolveTimeConflicts/Vue/App.vue#L329), which one is correct? Why do we need this style?

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
```diff
@@ -0,0 +1,213 @@
...
+  constructor(service: Service) {
+    this.appointmentsData = service.getAppointments();
+  }
+
+  setConflictError(show: boolean): void {
+    this.showConflictError = show;
+  }
+
+  onFormInitialized(e: DxFormTypes.InitializedEvent): void {
+    this.form = e.component;
```
Tucchhaa, line 89: This callback is never called for some reason. So changing 'startDate' for example doesn't hide conflict-informer if it was shown

I think we need rewrite [this](https://github.com/DevExpress/DevExtreme/blob/e3e7bef6d34e9a7fd1b590c50c513fe98a83a877/apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.html#L28) to [this](https://github.com/DevExpress/DevExtreme/blob/e3e7bef6d34e9a7fd1b590c50c513fe98a83a877/apps/demos/Demos/Scheduler/Templates/Angular/app/app.component.html#L32)

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  assignees: Assignee[] = assignees;
+
+  overlappingRuleItems = [
+    { value: 'sameResource', text: 'Allow across resources' },
+    { value: 'allResources', text: 'Disallow all overlaps' },
+  ];
+
+  assigneeIdEditorOptions = {
+    onValueChanged: (e: { value: unknown[]; component: { option: (k: string, v: unknown) => void } }) => {
```
Tucchhaa: Very strange parameter type, please change it to the type like in React and Vue

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
```diff
@@ -0,0 +1,213 @@
...
+      if (e.value.length > 1) {
+        e.component.option('value', [e.value[e.value.length - 1]]);
+      }
+    },
+    tagTemplate: 'assigneeTagTemplate',
+  };
+
+  popupOptions = {
+    onInitialized: (e: { component: dxPopup }) => { this.popup = e.component; },
```
Tucchhaa: Change parameter type to the similar type like in React and Vue

apps/demos/Demos/Scheduler/ResolveTimeConflicts/React/App.tsx
```diff
@@ -0,0 +1,270 @@
...
+  onValueChanged: (e: TagBoxTypes.ValueChangedEvent) => {
+    if (e.value.length > 1) {
+      e.component.option('value', [e.value[e.value.length - 1]]);
+    }
+  },
+  tagRender: (itemData: Assignee) => tagTemplate(itemData),
+};
+
+const tagTemplate = (itemData: Assignee) => (
+  <div className="dx-tag-content" style={{ backgroundColor: itemData.color, borderColor: itemData.color }}>
```
Tucchhaa, line 94: tag background color is not applied in react:

Image

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
```diff
@@ -0,0 +1,213 @@
...
+  assignees: Assignee[] = assignees;
+
+  overlappingRuleItems = [
+    { value: 'sameResource', text: 'Allow across resources' },
+    { value: 'allResources', text: 'Disallow all overlaps' },
+  ];
+
+  assigneeIdEditorOptions = {
+    onValueChanged: (e: { value: unknown[]; component: { option: (k: string, v: unknown) => void } }) => {
+      if (e.value.length > 1) {
```
Tucchhaa: For some reason e.value can be undefined and we get this error:

Image

It happens only in Angular, so I suggest to rewrite it to:
`if (e.value?.length > 1) {`

The error is shown in this case:
1. Open any appointment
2. Close the form
3. Click on cell to create new appointment
4. The error is shown to the console

apps/demos/Demos/Scheduler/ResolveTimeConflicts/jQuery/styles.css
aleksei-semikozov: Ok, removed it

apps/demos/Demos/Scheduler/ResolveTimeConflicts/jQuery/styles.css
aleksei-semikozov: I forget remove this, delete it

apps/demos/Demos/Scheduler/ResolveTimeConflicts/Angular/app/app.component.ts
aleksei-semikozov: fixed

apps/demos/Demos/Scheduler/ResolveTimeConflicts/jQuery/styles.css
```diff
@@ -0,0 +1,64 @@
...
```
Tucchhaa: Please remove this styles from jquery too

apps/demos/Demos/Scheduler/ResolveTimeConflicts/React/App.tsx
```diff
@@ -0,0 +1,275 @@
...
+  const div = document.createElement('div');
+  div.className = 'dx-tag-content';
+  div.style.backgroundColor = itemData.color;
+  div.style.borderColor = itemData.color;
+  div.textContent = itemData.text;
+  const removeBtn = document.createElement('div');
+  removeBtn.className = 'dx-tag-remove-button';
+  div.appendChild(removeBtn);
+  return div;
+}
```
Tucchhaa: It's possible to use JSX, check how I have done it before in [this commit](https://github.com/DevExpress/DevExtreme/pull/32703/changes/1dc2da147600b7c435cb79e9c4cca67e1518460c#diff-e57bda88a7817122e7c411ed936aa990860f5dcc1dc82873a0aa8765de349b35L50).

I suggest to use JSX as it is more "React way". The approach is similar to Vue's and Angular's in terms that we use \<Template\> tag

Tucchhaa (approved):

PR #32755, opened 2026-03-03
DataGrid - decompose _getNextCell method in keyboard navigation controller

dmirgaev:


packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -2103,77 +2103,98 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
...
-        isVerticalNavigation
-        && isElementDefined($cell)
-        && !this._isCellValid($cell)
-      ) {
-        const visibleColumns = this._columnsController.getVisibleColumns(null, true);
-        const newFocusedColumnIndex = newFocusedCellPosition.columnIndex;
-        const newFocusedColumn = visibleColumns[newFocusedColumnIndex];
+    const newPosition = this._getNewPositionByCode(focusedPosition, elementType, keyCode);
+    const $cell = $(this._getCell(newPosition));
+    const $adjustedCell = this._adjustCellOnVerticalNav($cell, newPosition, keyCode);
```
anna-shakhova: We can move line
```
  const $cell = $(this._getCell(newPosition));
```
inside _adjustCellOnVerticalNav and call it like getCellByPositionAndCode or smth like this
```
  const $adjustedCell = this.getCellByPositionAndCode(newPosition, keyCode);
```

anna-shakhova: Let's use the same order in position/keyCode between different methods for convinience

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -2103,77 +2103,98 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
...
+    return this._getNextCellSkippingHiddenRow($validCell, keyCode, focusedPosition);
+  }
 
-      const isLastCellOnDirection = keyCode === 'previous'
-        ? this._isFirstValidCell(newFocusedCellPosition)
-        : this._isLastValidCell(newFocusedCellPosition);
+  // T1322130, T1322440: During vertical navigation, command cells (e.g. expand)
+  // may be invalid because their column inherits groupIndex from the grouped data column.
+  // Find the first valid data cell in the target row instead of recursing to the next row.
+  private _adjustCellOnVerticalNav(
```
anna-shakhova: in our current convention we do not add _ prefix for private method, just use key word `private` instead

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -2103,77 +2103,98 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
...
+    if (!this._isCellInRow(position, includeCommandCells) || isBoundaryCell) {
+      return $cell;
+    }
+
+    return isRowFocus
+      ? this.getFirstValidCellInRow($cell.parent(), position.columnIndex)
+      : this._getNextCell(keyCode, 'cell', position);
+  }
+
+  private _getNextCellSkippingHiddenRow(
```
anna-shakhova:
```suggestion
  private resolveHiddenRowCell(
```

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/types.ts
```diff
@@ -1 +1,5 @@
 export type NavigationDirection = 'next' | 'previous' | 'nextInRow' | 'previousInRow';
+
+export type NavigationDirectionExtended = NavigationDirection | 'upArrow' | 'downArrow';
```
anna-shakhova: Nitpick: We have some mix of direction and codes here, a bit confusing
perhaps, change name to
export type NavigationKeyCode = NavigationDirection | 'upArrow' | 'downArrow';

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
dmirgaev: In my opinion, this will only make it worse readable: what code means?
I agree, that it could be shorter, but I totally don't like such a method name (=
Gonna think, how to name it

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/types.ts
dmirgaev: But it's not a key code, it's exactly direction, and I've really racked my brains over how to call it (=
Okay, let it be navigation key code... even though it's not a key code T___T

anna-shakhova (approved):

PR #32802, opened 2026-03-05
DataGrid: Fix the banded Columns' vertical borders (T1318812)

Alyar666:


packages/devextreme-scss/scss/widgets/base/dataGrid/layout/cell.scss
```diff
@@ -192,24 +199,24 @@ $datagrid-focused-border-color: null !default;
   border-left-color: $datagrid-border-color;
 }
 
+// (0,6,1)
+.dx-rtl > .dx-gridbase-container .dx-datagrid-content .dx-datagrid-table .dx-row > td.dx-datagrid-first-cell {
+  border-right: none;
```
anna-shakhova, line 204: suggestion: use where possible `border-inline-start` property and similar to avoid duplications for rtl and non-rtl styles

Alyar666: Yeah, makes sense 👍
Right now it would require quite a lot of refactoring, so probably not worth doing in this scope.

But agreed — we should switch to properties like border-inline-start in the future to avoid duplicating RTL/LTR styles.

e2e/testcafe-devextreme/tests/dataGrid/common/bandColumns/functional.ts
```diff
@@ -0,0 +1,211 @@
...
+// After resize (width: 275px, field1 hidden — hidingPriority: 0):
+//
+// Header Rows:
+// ┌───────────────────────┬──────────┬──────────┐
+// │  Band 1 (colspan=2)   │  field3  │ adaptive │
+// │     [FIRST ✓]         │(rowspan2)│(rowspan2)│
+// ├───────────┬───────────┤          │          │
+// │  field1   │  field2   │          │          │
+// │ [HIDDEN]  │ [FIRST ✓] │          │          │
+// └───────────┴───────────┴──────────┴──────────┘
```
Raushen, line 99: So cool scheme!

packages/devextreme/js/__internal/grids/grid_core/adaptivity/m_adaptivity.ts
```diff
@@ -1317,25 +1326,54 @@ const resizing = (Base: ModuleType<ResizingController>) => class AdaptivityResiz
...
+    oldHiddenColumns: Column[],
+    hiddenColumns: Column[],
+  ): boolean {
+    if (oldHiddenColumns.length !== hiddenColumns.length) {
+      return true;
+    }
+
+    const oldIndices = new Set(oldHiddenColumns.map((col) => col.index));
+
+    return hiddenColumns.some((col) => !oldIndices.has(col.index));
```
Raushen, line 1339: Minor: looks like a util method areArraysEqual

Raushen (approved):

PR #32816, opened 2026-03-06
GridCore: make expand column header rowspan equal to header rows count (T1317623)

anna-shakhova:


packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
```diff
@@ -758,18 +752,18 @@ export class ColumnsController extends modules.Controller {
...
     const bandColumnsCache = this.getBandColumnsCache();
 
-    const positiveIndexedColumns: any = [];
-    const negativeIndexedColumns: any = [];
+    const positiveIndexedColumns: IndexedColumns['positiveIndexedColumns'] = [];
+    const negativeIndexedColumns: IndexedColumns['negativeIndexedColumns'] = [];
 
     for (let i = 0; i < rowCount; i += 1) {
-      negativeIndexedColumns[i] = [{}];
+      negativeIndexedColumns[i] = {};
```
anna-shakhova, line 766: Change initial value to align types for positive and negative indexes.
Previously negative columns were like `[{}, [{...-1column}], [{...-2column}]]`,
now: `{ 1: [{...-1column}], 2: [{...-2column}]}`

this inconsistency also resulted in necessity to use unshift.apply in _getVisibleColumnsFromIndexed, as columns had different type for different indexes and without apply caused runtime error here
```result[rowIndex].unshift.apply(result[rowIndex], columns);```

**Review by Alyar666 — CHANGES_REQUESTED** [review]
There is a missing border in RTL mode:

Image

File: packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts [hunk 2]
```diff
@@ -859,26 +860,22 @@ export const getFixedPosition = function (that: ColumnsController, column) {
...
-
-  if (rowspan > 1) {
-    expandColumnsByType = map(expandColumnsByType, (expandColumn) => extend({}, expandColumn, { rowspan }));
-  }
-  expandColumnsByType.unshift.apply(expandColumnsByType, isDefined(customColumnIndex) ? [customColumnIndex, 1] : [columnIndex, 0]);
-  columns.splice.apply(columns, expandColumnsByType);
-
-  return rowspan || 1;
+export const processExpandColumns = (
+  columns: Column[],
```
Alyar666, line 864: Minor comment: Maybe we can pass arguments as an object?

Alyar666 (approved):

PR #32862, opened 2026-03-11
Splitter: size of panes should be updated taking into account their limitations on splitter resize

pharret31:


packages/devextreme/js/__internal/ui/splitter/splitter.ts
```diff
@@ -252,8 +270,11 @@ class Splitter extends CollectionWidgetLiveUpdate<Properties> {
     this._updateResizeHandlesResizableState();
     this._updateResizeHandlesCollapsibleState();
 
+    this._initialPaneSizes = items.map((item: Item): string | number | undefined => item.size);
```
dmlvr, line 273: minor: Is type `Item['size']` better for this place?

```this._initialPaneSizes = items.map((item: Item): Item['size'] => item.size);```

File: packages/devextreme/js/__internal/ui/splitter/splitter.ts [hunk 2]
```diff
@@ -106,6 +108,7 @@ export interface Properties<
     keyof PublicProperties<TItem, TKey>
   > {
   _renderQueue?: RenderQueueItem[];
+  _ignoreSizeConstraints?: boolean;
```
dmlvr, line 111: minor: maybe `_ignoreSizeRestrictions` better?

We have `_itemRestrictions`, it looks like the same

dmlvr (approved):

PR #32921, opened 2026-03-16
DataGrid - Performance is slow if a large number of summaries is used (T1316562)

dmirgaev:


packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -490,32 +491,96 @@ const data = (Base: ModuleType<DataController>) => class SummaryDataControllerEx
...
   }
 
-  private _calculateSummaryCells(summaryItems, aggregates, visibleColumns, calculateTargetColumnIndex, isGroupRow?) {
-    const that = this;
+  // The map is built once per _processItems cycle (via options) and discarded after.
+  private _buildColumnLookupMap(): ColumnMap {
+    const columnMap: ColumnMap = new Map();
+    const allColumns = [
+      ...this._columnsController.getColumns(),
+      ...(this._columnsController._commandColumns ?? []),
```
Alyar666: Even though the _commandColumns property has a public modifier, it should actually be private. Perhaps we should implement a public getCommandColumns method at the columnsController level? We could also add a public getAllColumns method to the columnController

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -490,32 +491,96 @@ const data = (Base: ModuleType<DataController>) => class SummaryDataControllerEx
...
+    ];
+
+    for (const column of allColumns) {
+      const copiedColumn = { ...column };
+      // The method registers each column under a few keys: index, name, dataField, and caption.
+      // This is because the developer can specify summaryItem.column (and summaryItem.showInColumn)
+      // in any of these forms — number for column index and string for all the rest.
+      const keys = [
+        column.index,
```
Alyar666, line 543: Do we really need an index?

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -490,32 +491,96 @@ const data = (Base: ModuleType<DataController>) => class SummaryDataControllerEx
...
+      for (const key of keys) {
+        columnMap.set(key, copiedColumn);
+      }
+    }
+
+    return columnMap;
+  }
+
+  private _calculateSummaryCells(
+    summaryItems: SummaryItem[],
```
Alyar666: Minor comment: There are too many arguments. Can we group some of them into an object?

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -471,9 +472,9 @@ const data = (Base: ModuleType<DataController>) => class SummaryDataControllerEx
   private _processGroupItem(groupItem, options) {
     const that = this;
 
-    if (!options.summaryGroupItems) {
-      options.summaryGroupItems = that.option('summary.groupItems') || [];
-    }
+    options.summaryGroupItems ??= that.option('summary.groupItems') || [];
```
anna-shakhova: looks like no reason to use alias for `this` here, same in line 524
tip: if use `const { summary: { groupItems } } = this.option()` it will not lose types when options typed, while usage with argument loses

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
dmirgaev: Discussed: column index could be used by user as `column` or `showInColumn`, even though it's not described in docs

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
dmirgaev: Yes, you're right, we even don't need command columns here, as they serve no functional purpose for summary computation

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -490,32 +489,93 @@ const data = (Base: ModuleType<DataController>) => class SummaryDataControllerEx
...
+
+  private _calculateSummaryCells({
+    summaryItems,
+    aggregates,
+    visibleColumns,
+    calculateTargetColumnIndex,
+    isGroupRow,
+    columnMap,
+  }: CalculateSummaryCellsArgs) {
     const summaryCells: any = [];
```
anna-shakhova: Lets add typing for summaryCells, if it is not too big for this task

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -490,32 +489,93 @@ const data = (Base: ModuleType<DataController>) => class SummaryDataControllerEx
...
     const summaryCellsByColumns = {};
```
anna-shakhova: from line 610
```suggestion
    const summaryCellsByColumns: Record<number, SummaryCell[] | undefined> = {};
```

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -532,7 +592,9 @@ const data = (Base: ModuleType<DataController>) => class SummaryDataControllerEx
             valueFormat = gridCore.getFormatByDataType(column && column.dataType);
           }
           summaryCellsByColumns[columnIndex].push(extend({}, summaryItem, {
```
anna-shakhova, line 595: `extend({}, someObj, {...someProps})` is the same as 
```
{
  ...someObj,
  ...someProps
}
```

let's replace it here for readability

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -490,32 +489,93 @@ const data = (Base: ModuleType<DataController>) => class SummaryDataControllerEx
...
-      const column = that._columnsController.columnOption(summaryItem.column);
-      const showInColumn = summaryItem.showInColumn && that._columnsController.columnOption(summaryItem.showInColumn) || column;
+      const column = columnMap
+        ? getColumnFromMap(summaryItem.column, columnMap)
+        : this._columnsController.columnOption(summaryItem.column);
+      const showInColumn = (summaryItem.showInColumn
+        && (columnMap
+          ? getColumnFromMap(summaryItem.showInColumn, columnMap)
+          : this._columnsController.columnOption(summaryItem.showInColumn)))
+        || column;
```
anna-shakhova:
```suggestion
      const getColumnByKey = (key) => columnMap
        ? getColumnFromMap(key, columnMap)
        : this._columnsController.columnOption(key);
      const column = getColumnByKey(summaryItem.column);
      const showInColumn = summaryItem.showInColumn ? getColumnByKey(summaryItem.showInColumn) : undefined;
        const columnIndex = calculateTargetColumnIndex(summaryItem, showInColumn ?? column);
```

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -532,7 +592,9 @@ const data = (Base: ModuleType<DataController>) => class SummaryDataControllerEx
             valueFormat = gridCore.getFormatByDataType(column && column.dataType);
           }
           summaryCellsByColumns[columnIndex].push(extend({}, summaryItem, {
-            value: isString(aggregate) && column && column.deserializeValue ? column.deserializeValue(aggregate) : aggregate,
+            value: isString(aggregate) && column && column.deserializeValue
```
anna-shakhova, line 596: can we replace it with shorter
```suggestion
            value: isString(aggregate) && column?.deserializeValue
```

Alyar666 (dismissed):

anna-shakhova (dismissed):

anna-shakhova (approved):

PR #32936, opened 2026-03-17
Scheduler: Provide the SnapToCellsMode option — update .d.ts

sjbur:


packages/devextreme/js/ui/scheduler.d.ts
```diff
@@ -1145,6 +1152,10 @@ export interface dxSchedulerOptions extends WidgetOptions<dxScheduler> {
        * @default "all"
        */
        allDayPanelMode?: AllDayPanelMode;
+       /**
+        * @docid
+        */
```
aleksei-semikozov, line 1157: Should we add a @default annotation for snapToCellsMode in the d.ts? Currently the docs show null as default.

aleksei-semikozov (approved):

PR #32944, opened 2026-03-18
Scheduler - Appointment Form - Fix popup jump on group switching

Tucchhaa:


sjbur (dismissed):

e2e/testcafe-devextreme/tests/scheduler/common/appointmentForm/form.functional.ts
```diff
@@ -60,3 +67,73 @@ test('Recurrence start date editor should have focus after opening recurrence se
...
+            items: ['repeatGroup'],
+          },
+          'recurrenceGroup',
+        ],
+      },
+    },
+  });
+});
+
+test('Popup should not change dimensions when switching groups and main group height is larger', async (t) => {
```
aleksei-semikozov, line 106: This test passes without the fix — I verified it by running only the test changes against the unfixed code (see #32948). What specific regression scenario does it guard against?

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -896,8 +896,6 @@ export class AppointmentForm {
   }
 
   showMainGroup(): void {
```
aleksei-semikozov, line 899: Instead of comments explaining the call order, consider extracting the logic into named steps — this makes the sequence self-documenting:

```
showMainGroup(): void {
  this.restorePopupHeight();
  this.switchGroupVisibility('main');
  this._popup.updateToolbarForMainGroup();
}

\\ ...

showRecurrenceGroup(): void {
  this.lockPopupHeight();
  this.switchGroupVisibility('recurrence');
  this._popup.updateToolbarForRecurrenceGroup();
}
```
When the method reads as three clear steps, the ordering constraint becomes obvious from the code itself — no comment needed.

Tucchhaa: I don't think the ordering constraint is obvious in the provided codesnippet. Because, it's not clear why I can't update toolbar before switching group visibility. I think the comment still will be needed.

The unobvious thing is that toolbar update causes the position recalculation, which comment makes obvious

e2e/testcafe-devextreme/tests/scheduler/common/appointmentForm/form.functional.ts
Tucchhaa: Thanks for the draft PR.

While this test doesn't fail without the fix, it's there just to cover different scenarios. Like if the case when "recurrence group height is larger", why not also cover the mirrored case when "main group height is larger".

So this is an e2e test and just checks that basic user scenario works without taking into account the implementation details

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
aleksei-semikozov: In a 20-line method, a comment is easy to miss during refactoring.

With named steps like `restorePopupHeight → switchGroupVisibility → updateToolbar`, the ordering reads naturally — like `prepare → execute → finalize`. You wouldn't move finalize before execute. And if a comment is still wanted, it's far more visible on a 3-line method than buried in a long one.

e2e/testcafe-devextreme/tests/scheduler/common/appointmentForm/form.functional.ts
aleksei-semikozov: This test doesn't catch any regression — it passes with and without the fix. It doesn't cover any special edge case in the code either. It exists only as a mirror of the first test.

A testcafe test is expensive to run and maintain. Adding one in a bugfix PR should be motivated: it either reproduces the bug or covers a known edge case. "Just in case" is not enough reason. Can you explain what specific problem this test would catch in the future?

if we add tests for every possible configuration "just in case", where do we stop? The number of user scenarios with custom editing.form.items is practically unlimited.

sjbur (approved):

aleksei-semikozov (approved):

PR #32955, opened 2026-03-18
Scheduler — Fix slide animation overlapping items outside mainGroup

aleksei-semikozov:


packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -1065,6 +1066,22 @@ export class AppointmentForm {
     this.dxForm.endUpdate();
   }
 
+  private updateAnimationOffset(): void {
+    if (!this._$mainGroup) {
+      return;
+    }
+
+    const formElement = this.dxForm.$element()[0];
```
Tucchhaa: minor: we can use `this.dxForm.element()` instead

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
```diff
@@ -1065,6 +1066,22 @@ export class AppointmentForm {
...
+    const mainGroupElement = this._$mainGroup[0];
+
+    if (formElement && mainGroupElement) {
```
Tucchhaa: Are there cases when formElement does not exist?

Tucchhaa: Also I see the existence of main group is already checked here:
https://github.com/DevExpress/DevExtreme/pull/32955/changes#diff-e2184e9321386d96f057b770cc362ad0d417f9be815ab4e9ac68f5b977f54009R1070

Can we remove this if statement then?

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
```diff
@@ -1594,6 +1594,18 @@ describe('Appointment Form', () => {
...
+    it('should set animation offset CSS variable when switching to recurrence form', async () => {
+      const { scheduler, POM } = await createScheduler(getDefaultConfig());
+
+      scheduler.showAppointmentPopup();
+
+      POM.popup.selectRepeatValue('weekly');
+
+      const formElement = POM.popup.dxForm.$element()[0];
+      const animationTop = formElement.style.getPropertyValue('--dx-scheduler-animation-top');
+      expect(animationTop).toBe('50px');
```
Tucchhaa, line 1613: It wasn't clear to me why animationTop should be 50px if there's no form customization, until I saw that getBoundingClientRect is mocked with some special values.

I suggest to write a functional testcafe test with form customization, so that we wouldn't need to mock getBoundingClient rect and the test would become more clear.

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
aleksei-semikozov: did it

packages/devextreme/js/__internal/scheduler/appointment_popup/m_form.ts
aleksei-semikozov: formElement element() always returns the mounted DOM node when the form is initialized
Fixed — _$mainGroup is already guarded above and element() is always available at that point. The inner check was redundant, remove it

packages/devextreme/js/__internal/scheduler/appointment_popup/appointment_popup.test.ts
aleksei-semikozov: Prefer Jest here — CSS animation timing makes TestCafe tests flaky + any changed css or fonts break TestCafe test. 
The test verifies the --dx-scheduler-animation-top value is calculated correctly, which is the actual behavior we care about.
The mock makes the expected value deterministic. If you feel the mock setup needs more clarity, happy to add a comment explaining the mocked rect values.

I suggest extending setupSchedulerTestEnvironment with classRects option so the rect values are explicit right in the test:

```
it('should set animation offset when custom item is placed before mainGroup', async () => {
  setupSchedulerTestEnvironment({
    height: 600,
    classRects: {
      'dx-scheduler-form-main-group': { top: 50, height: 50, bottom: 100 },
    },
  });

  const { scheduler, POM } = await createScheduler(getDefaultConfig());

  POM.popup.selectRepeatValue('weekly');

  const animationTop = POM.popup.dxForm.element().style.getPropertyValue('--dx-scheduler-animation-top');
  expect(animationTop).toBe('50px'); // mainGroup starts 50px from form top
});
```

Tucchhaa (dismissed):

Tucchhaa (approved):

PR #32969, opened 2026-03-19
DataGrid - Fix lookup filter changing to "(All)" after searching twice in another column (T1284002)

markallenramirez:


packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_row.ts
```diff
@@ -769,6 +769,13 @@ const columnHeadersView = (Base: ModuleType<ColumnHeadersView>) => class ColumnH
           || !equalByValue(editorDataSource.__dataGridSourceFilter || null, filter);
 
         if (shouldUpdateFilter) {
+          const selectedItem = editor.option('selectedItem');
```
anna-shakhova: minor: it is preferable to use destructurization from options() instead of getting option by name
in that case it will not lose types if editor properly typed
```
const { selectedItem, items = [] } = editor.option();
```

Raushen (approved):

anna-shakhova (approved):

PR #32994, opened 2026-03-20
Scheduler - Qunit tests - remove watchers of private methods

aleksei-semikozov:


packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/common.options.tests.js
```diff
@@ -959,10 +958,10 @@ QUnit.module('Options', () => {
             'views[2].intervalCount': 2,
             'views[2].startDate': new Date(),
         });
-        await waitForAsync(() => count === 2);
+        await waitForAsync(() => loadCount === 2);
+        await waitAsync(200);
```
sjbur, line 962: Why did you change the logic in waitAsync?

sjbur: I mean the await waitAsync(200);

aleksei-semikozov: The `waitAsync(200)` is added on purpose. Here is why:

1. `waitForAsync(() => loadCount === 2)` waits until two loads finish
2. After that we need to check that no third load happens
3. `waitAsync(200)` gives enough time for a third load to fire if it exists (the setTimeout inside CustomStore is 100ms, so 200ms is enough)
4. Then `assert.equal(loadCount, 2)` catches the problem if a third load happened

sjbur (approved):

PR #33004, opened 2026-03-20
Refactor test

Raushen:


e2e/testcafe-devextreme/tests/dataGrid/sticky/common/withGrouping.ts
```diff
@@ -291,31 +291,15 @@ test('DataGrid - Group row content is scrolled if repaintChangesOnly is enabled
   // T1284612
   test.meta({
     browserSize: [900, 800],
```
anna-shakhova: Check, please, whether we need custom browserSize for these test.
We set width and height for DataGrid in its configuration, then make screenshot of the grid only, looks like custom window size not used

e2e/testcafe-devextreme/tests/dataGrid/sticky/common/withGrouping.ts
```diff
@@ -370,3 +356,61 @@ test('DataGrid - Group row content is scrolled if repaintChangesOnly is enabled
...
+    await t
+      .expect(compareResults.isValid())
+      .ok(compareResults.errorMessages());
+  }).before(async () => createWidget('dxDataGrid', {
+    ...defaultConfig,
+    rtlEnabled,
+    customizeColumns(columns) {
+      columns[2].groupIndex = 0;
+      columns[1].visible = false;
+      columns[4].width = 150;
```
anna-shakhova, line 378: Suggestion: return back previous logic with changing grid option in runtime to change `summary.totalItelm` and `column[index].width`. This will save us time on initial page render

Raushen: I do not agree. It is difficult to create one config for 4 cases. Also I do not think initial render takes a lot of time.

anna-shakhova (approved):

PR #33014, opened 2026-03-23
Scheduler - Appointments Refactoring - Rendering

Tucchhaa:


packages/devextreme/js/__internal/scheduler/types.ts
```diff
@@ -13,11 +13,9 @@ export type RenderStrategyName = 'horizontal' | 'horizontalMonth' | 'horizontalM
 export type FilterItemType = Record<string, string | number> | string | number;
 export type HeaderCellTextFormat = string | ((date: Date) => string);
 
-export interface SafeAppointment extends Appointment {
-  startDate: Date | string;
-  endDate: Date | string;
```
Tucchhaa, line 18: SafeAppointment type actually doesn't have this properties, so I have removed them

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
```diff
@@ -0,0 +1,167 @@
...
+  targetedAppointmentData: TargetedAppointment;
+  appointmentTemplate: SchedulerProperties['appointmentTemplate'];
+
+  onAppointmentRendered: (e: AppointmentRenderedEvent) => void;
+
+  getResourceManager: () => ResourceManager;
+  getDataAccessor: () => AppointmentDataAccessor;
+}
+
+export class BaseAppointment<
```
aleksei-semikozov: We talked about this in our meeting — in the Scheduler codebase "Appointment" is the source data (what user creates). What we render on the view is an "Occurrence" — we already have the `Occurrence` type and `getOccurrences()` method for this.

This is not a nitpick. This is the wrong name. These classes render items on a view, not the source data. Right now we have `SafeAppointment`, `AppointmentItemViewModel`, `BaseAppointment` — three different things all called "Appointment". This makes the code harder to read.

Please rename to `BaseOccurrenceView` / `GridOccurrenceView` / `AgendaOccurrenceView`, or at least `AppointmentView`.

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
```diff
@@ -0,0 +1,167 @@
...
+  TProperties extends BaseAppointmentProperties = BaseAppointmentProperties,
+> extends DOMComponent<BaseAppointment<TProperties>, TProperties> {
```
aleksei-semikozov: I looked at what `DOMComponent` features are actually used here:

- [`$element()`](https://github.com/DevExpress/DevExtreme/blob/69ef155fc7b9ceb6a0535d53ee24720897352f2a/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts#L78) — just a jQuery ref
- [`option()`](https://github.com/DevExpress/DevExtreme/blob/69ef155fc7b9ceb6a0535d53ee24720897352f2a/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts#L43) — just a props object
- [`_getTemplateByOption()`](https://github.com/DevExpress/DevExtreme/blob/69ef155fc7b9ceb6a0535d53ee24720897352f2a/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts#L140) — one call: `template.render({ container, model })`. Parent can do this.
- [`_createActionByOption()`](https://github.com/DevExpress/DevExtreme/blob/69ef155fc7b9ceb6a0535d53ee24720897352f2a/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts#L61) — could be a direct callback

The only place that really needs `DOMComponent` is [`_createComponent` in AppointmentCollector](https://github.com/DevExpress/DevExtreme/blob/69ef155fc7b9ceb6a0535d53ee24720897352f2a/packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts#L78) — it creates a Button and handles `disabled`/`rtl` cascading.

But that does not mean every appointment needs to be a `DOMComponent`. The Collector can get a button factory from the parent instead.

Also: [`focus()` is an empty stub](https://github.com/DevExpress/DevExtreme/blob/69ef155fc7b9ceb6a0535d53ee24720897352f2a/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts#L90). We lost KBN and accessibility that `CollectionWidget` gave us for free.

I think each appointment should be a simple object (jQuery element + position + classes), not a full `DOMComponent`. Template rendering should stay in the parent — like how `CollectionWidget` does it for `List`, `Gallery`, `TileView`.

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/grid_appointment.test.ts
```diff
@@ -0,0 +1,261 @@
...
+import type { ResourceConfig } from '@ts/scheduler/utils/loader/types';
+import type { AppointmentItemViewModel } from '@ts/scheduler/view_model/types';
+
+import fx from '../../../common/core/animation/fx';
+import { getMockedBaseAppointmentProperties, mockGridViewModel } from '../__mock__/appointment_properties';
+import { APPOINTMENT_CLASSES, APPOINTMENT_TYPE_CLASSES } from '../const';
+import type { GridAppointmentProperties } from './grid_appointment';
+import { GridAppointment } from './grid_appointment';
+
+const getGridAppointmentProperties = (options: {
```
aleksei-semikozov: The test for "show correct title" (line 95-105) needs:
1. `getGridAppointmentProperties()` → `mockGridViewModel()` (40 lines, 15+ fields)
2. `getMockedBaseAppointmentProperties()` → mocked ResourceManager, DataAccessor
3. `createGridAppointment()` → DOMComponent + `@ts-expect-error` + `await process.nextTick`

All this to check that a div has text "Test title".

Simpler version:
```typescript
const $el = renderAppointment({ text: "Test title", startDate, endDate });
expect($el.find(".dx-scheduler-appointment-title").text()).toBe("Test title");
```

The rendering logic is ~30 lines of jQuery. But we need a 126-line mock file to test it. These tests check **how the component is connected** (DOMComponent lifecycle, template manager, resource manager) — not **what it does** (render a title in a div). If `AppointmentItemViewModel` adds a new field, the mock breaks even though title rendering did not change.

I think this happened because `AppointmentItemViewModel` is too complex and has too many things inside. It mixes rendering data, positioning, view state, and data access all in one object. If the view model was simpler and focused only on what the component needs to render, the tests would be simpler too.

packages/devextreme/js/__internal/scheduler/appointments_new/utils/get_view_model_diff.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+    : result;
+};
+
+const isAppointmentDataChanged = (
+  appointmentData: SafeAppointment,
+  appointmentDataSource: AppointmentDataSource,
+): boolean => {
+  const updatedAppointmentData = appointmentDataSource.getUpdatedAppointment();
+
```
aleksei-semikozov, line 51: minor: Just a thought — this is classic LCS with O(n*m) time and space. With virtual scrolling we can have 50-100+ visible appointments, and this runs on every scroll.

Since appointments use absolute positioning (top/left), DOM order does not matter. A simple `Map<itemData, component>` would give O(n+m) — iterate both lists, find added/removed/kept.

Not critical, you can keep it as is. But think about it — a map-based approach would be simpler and faster here because we do not need to preserve order.

packages/devextreme-scss/scss/widgets/fluent/scheduler/_index.scss
```diff
@@ -209,25 +209,25 @@ $fluent-scheduler-agenda-time-panel-cell-padding: 8px;
...
-        .dx-item-content.dx-scheduler-appointment-content {
+        .dx-scheduler-appointment-content {
           padding-right: $fluent-scheduler-appointment-20min-recurrence-padding-right;
         }
       }
     }
 
-    .dx-item-content.dx-scheduler-appointment-content {
+    .dx-scheduler-appointment-content {
       @container (max-height: #{$fluent-scheduler-appointment-25min-height}) {
```
aleksei-semikozov, line 231: minor: This removes `.dx-item-content` from the selector, which changes CSS specificity. This is the only change that affects production (old implementation). The etalon screenshots changed — so there is a visual diff.

This feels out of place in this PR. Maybe move to a separate small PR so we can test it on its own?

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -310,18 +316,31 @@ class Scheduler extends SchedulerOptionsBaseWidget {
         this.updateOption('header', name, value);
         break;
       case 'currentView':
-        this._appointments.option({
-          items: [],
-          allowDrag: this.allowDragging(),
-          allowResize: this.allowResizing(),
-          itemTemplate: this.getAppointmentTemplate('appointmentTemplate'),
-        });
+
```
aleksei-semikozov, line 319: 8 `if/else` blocks for `_newAppointments` in one file. This is the part I worry about most.

Every future change to the Scheduler needs to handle both paths. If someone changes the old path and forgets the new one — we get a bug that only shows up when the flag is turned on.

I think we should replace parts one by one instead. Each PR removes old code and adds new code that works right away. No dead code, no parallel paths. For example:
1. Extract `get_date_text.ts` to replace old `m_text_utils.ts`
2. Extract `get_targeted_appointment.ts`
3. Replace `CompactAppointmentsHelper` with new `AppointmentCollector`
4. Then migrate the rendering itself

aleksei-semikozov:
Good work on isolating the appointment components and the diff-based rendering idea.

I left inline comments, but here is the summary:

**1. Feature flag:** We are adding ~2800 lines behind `_newAppointments` that is not enabled and not tested with real Scheduler. This creates 8 if/else branches in m_scheduler.ts.

I think we can do this step by step — replace old code with new code directly, no flag needed:
- `get_date_text.ts` can replace `m_text_utils.ts` + `text_utils.ts` right now (same logic, cleaner code)
- `get_targeted_appointment.ts` can replace the old one in `scheduler/utils/`
- `AppointmentCollector` (107 lines) can replace `CompactAppointmentsHelper` (205 lines)
- Then migrate rendering

Each step is small, tested in production, and removes old code instead of adding dead code.

**2. No integration tests:** 77 unit tests, but no test with a real Scheduler. We do not know if this works end-to-end.

**3. DOMComponent per appointment:** Each appointment is a full DOMComponent — with template manager, action system, options, lifecycle. But it is just a rectangle with text. This is too much. If we moved away from CollectionWidget, we should use something simpler, not something equally heavy. See inline comments.

Let's discuss the approach before merging.

packages/devextreme-scss/scss/widgets/fluent/scheduler/_index.scss
sjbur: I am not sure if the screenshots changed because of this, but I would like to know if it was because of css or not.

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/agenda_appointment.ts
```diff
@@ -0,0 +1,112 @@
...
+  ): dxElementWrapper {
+    this.renderMarker($container);
+    this.renderInfo($container);
+
+    return $container;
+  }
+
+  private renderMarker($container: dxElementWrapper): void {
+    const $leftContainer = $('<div>')
+      .addClass('dx-scheduler-agenda-appointment-left-layout')
```
sjbur, line 53: Can we move those classes this to AGENDA_APPOINTMENT_CLASSES?

packages/devextreme/js/__internal/scheduler/types.ts
sjbur: Then we should remove SafeAppointment? Now there is no point in having it

packages/devextreme/js/__internal/scheduler/m_subscribes.ts
```diff
@@ -117,24 +117,26 @@ const subscribes = {
...
     // pull out time zone converting from appointment adapter for knockout (T947938)
     const startDate = targetedAppointment.displayStartDate || this.timeZoneCalculator.createDate(adapter.startDate, 'toGrid');
     const endDate = targetedAppointment.displayEndDate || this.timeZoneCalculator.createDate(adapter.endDate, 'toGrid');
-    const formatType = format ?? getFormatType(startDate, endDate, adapter.allDay, this.currentView.type !== 'month');
+    const formatType = format ?? getDateFormatType(startDate, endDate, adapter.allDay, this.currentView.type);
 
     return {
       text: adapter.text || messageLocalization.format('dxScheduler-noSubject'),
-      formatDate: formatDates(startDate, endDate, formatType),
+      formatDate: getDateText(startDate, endDate, formatType as any),
```
sjbur, line 139: If you specify "format" argument in createFormattedDateText as of type DateFormatType, then we don't need to cast formatType as any.

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
Tucchhaa: Applied new AppointmentView name

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
Tucchhaa: Very good point, but I would suggest to postpone removal of DOMComponent. But right now it's a familiar mechanism to pass properties and use common functions. 

DOMComponent doesn't introduce a lot of overheads to the appointments and I believe it will be easy to remove DOMComponent in the future.

If you don't mind I would postpone this

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/grid_appointment.test.ts
Tucchhaa: Simplified properties that are passed to the appointments or collector. Now mocks in this tests are simpler.

packages/devextreme/js/__internal/scheduler/appointments_new/utils/get_view_model_diff.ts
Tucchhaa: I also thought about this solution, but we can't use this approach, because recurring appointments have the same itemData, so we don't have  a unique key for the appointments

packages/devextreme-scss/scss/widgets/fluent/scheduler/_index.scss
Tucchhaa: @sjbur definitely screenshots changed because of this. But it seems that now it's better, because all day appts have the same padding-right as the common appts.

That happened because of .dx-item-content selector, some styles weren't applied to the all day appts.

packages/devextreme/js/__internal/scheduler/appointments_new/utils/get_view_model_diff.ts
```diff
@@ -0,0 +1,153 @@
...
+    : {
+      allDay: item.allDay,
+      groupIndex: item.groupIndex,
+      direction: item.direction,
+      reduced: item.reduced,
+      partIndex: item.partIndex,
+      partTotalCount: item.partTotalCount,
+      rowIndex: item.rowIndex,
+      columnIndex: item.columnIndex,
+    };
```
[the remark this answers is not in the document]
Tucchhaa: old impl didn't have this props, so I consider that they maybe omitted

aleksei-semikozov (commented):
LGTM

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -0,0 +1,320 @@
...
+    return getViewModelDiff(
+      normalizedOldViewModel,
+      newViewModel,
+      this.option().getAppointmentDataSource(),
+    );
+  }
+
+  private renderAgendaAppointments(appointments: AppointmentViewModelPlain[]): void {
+    const commonFragment = domAdapter.createDocumentFragment();
+
```
aleksei-semikozov, line 145: Memory leak: `appointmentBySortIndex` is never reset before re-rendering. Each call to `renderAgendaAppointments` appends new entries but stale references from previous renders remain. Add `this.appointmentBySortIndex = {};` before the `forEach`.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -0,0 +1,320 @@
...
+    return getViewModelDiff(
+      normalizedOldViewModel,
+      newViewModel,
+      this.option().getAppointmentDataSource(),
+    );
+  }
+
+  private renderAgendaAppointments(appointments: AppointmentViewModelPlain[]): void {
+    const commonFragment = domAdapter.createDocumentFragment();
```
aleksei-semikozov, line 144: Only `$commonContainer` is cleared here but `$allDayContainer` is not. If a previous non-agenda render placed elements into `$allDayContainer`, switching to agenda view will leave those orphaned DOM nodes. Add `this.$allDayContainer?.empty();` here as well.

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    return Boolean(allDay);
+  }
+
+  private renderContentTemplate(): void {
+    const $content = $('<div>')
+      .addClass(APPOINTMENT_CLASSES.CONTENT)
+      .appendTo(this.$element());
+
+    const template = this.option().appointmentTemplate instanceof EmptyTemplate
```
aleksei-semikozov, line 119: minor: `instanceof EmptyTemplate` check looks fragile. Maybe better to use flag or null check? Not blocker.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -0,0 +1,320 @@
...
+
+    this.$commonContainer.empty();
+
+    appointments.forEach((appointmentViewModel) => {
+      const appointment = this.renderAppointment(commonFragment, appointmentViewModel);
+      this.appointmentBySortIndex[appointmentViewModel.sortedIndex] = appointment;
+    });
+
+    this.$commonContainer.get(0).appendChild(commonFragment);
+  }
```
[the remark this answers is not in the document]
aleksei-semikozov: Yes, copilot I think here is right

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -0,0 +1,320 @@
...
+      : oldViewModel as AppointmentItemViewModel[] | AppointmentCollectorViewModel[];
+
+    return getViewModelDiff(
+      normalizedOldViewModel,
+      newViewModel,
+      this.option().getAppointmentDataSource(),
+    );
+  }
+
+  private renderAgendaAppointments(appointments: AppointmentViewModelPlain[]): void {
```
aleksei-semikozov, line 141: minor: No test for grid → agenda switch with allDay appointments. Good to have for cleanup verification.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: Thank you for notificing, I have applied the fix and added a test

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: Done

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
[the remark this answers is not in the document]
Tucchhaa: Done

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: Added the test

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
Tucchhaa: EmptyTemplate is provided by default: https://github.com/DevExpress/DevExtreme/blob/9a9defe7c26dbbb5cd1cc2922d31598eea7e01b6/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts#L66

If user sets custom template function, then it should be applied.

If user sets null or undefined, then template should be empty according to the old impl ([codepen](https://codepen.io/FufelShmertz/pen/RNGMbGO?editors=0010)). 

If [default templates](https://github.com/DevExpress/DevExtreme/blob/9a9defe7c26dbbb5cd1cc2922d31598eea7e01b6/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts#L65) is null or undefined, we would need to somehow distinguish between users null or undefined and default null or undefined.

So I think current way is the simplest to support custom templates.

aleksei-semikozov (approved):

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -0,0 +1,323 @@
...
+          if (isRepaintAll) {
+            break;
+          }
+
+          this.getAppointmentElement(sortedIndex).remove();
+          break;
+        }
+        case diffItem.needToAdd: {
+          const fragment = allDay ? allDayFragment : commonFragment;
+          const appointment = this.renderAppointment(fragment, diffItem.item);
```
aleksei-semikozov, line 186: This will break for collectors.

When a collector aggregates all-day appointments, it gets `allDay: true` from the view model. So it goes into `allDayFragment` → `$allDayContainer`. But collectors are compact overflow buttons with absolute positioning — they should always go into `$commonContainer`.

```ts
const fragment = (allDay && !isAppointmentCollectorViewModel(diffItem.item))
  ? allDayFragment : commonFragment;
```

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -0,0 +1,323 @@
...
+
+          newAppointmentBySortedIndex[sortedIndex] = appointment;
+          break;
+        }
+        case diffItem.needToResize: {
+          const appointment = this.appointmentBySortIndex[sortedIndex];
+          appointment.option('geometry', {
+            height: diffItem.item.height,
```
aleksei-semikozov, line 194: This assigns numeric `width`/`height` to geometry. But `AgendaAppointmentViewModel.width` is a string (`'100%'`).

Right now agenda does not enter this path — it uses `renderAgendaAppointments`. But if it ever does, this will silently pass wrong types to `resize()`.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -0,0 +1,323 @@
...
+  public focus(): void { /* TODO: legacy compatibility */ }
+
+  public _renderAppointmentTemplate(): void { /* TODO: legacy compatibility */ }
+
+  private getAppointmentElement(sortedIndex: number): dxElementWrapper {
+    return this.appointmentBySortIndex[sortedIndex].$element();
+  }
+
+  private getViewModelDiff(
+    newViewModel: AppointmentItemViewModel[] | AppointmentCollectorViewModel[],
```
aleksei-semikozov, line 125: This takes `(newViewModel, oldViewModel)`, but the utility `getViewModelDiff` takes `(oldViewModel, newViewModel)`. You swap them inside.

This is confusing. If someone reads this method signature and calls the utility directly — they will swap old/new. Consider matching the utility order `(old, new)` here too.

Tucchhaa: I agree, looks like small refactoring, but I will apply it in the future PR

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: Ok, thank you. I will fix this in the future. I am sure, that not all of the cases are covered in this PR, so I expect that we will notice more and more such cases and fix them pointwise

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: I can't imagine a situation when it could ever go this way. But I can suggest to implement a test to cover this case. What do you think?

aleksei-semikozov: You're right, this path is unreachable with the current code. Let's skip it

PR #33021, opened 2026-03-24
Scheduler: Provide the SnapToCellsMode option

sjbur:


packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/options/get_view_model_options.test.ts
aleksei-semikozov, line 1: These tests don't check any logic — they just verify that hardcoded values in configByView

aleksei-semikozov: Just remove it

sjbur: OK

packages/devextreme/js/__internal/scheduler/types.ts
```diff
@@ -9,6 +9,7 @@ export type Direction = 'vertical' | 'horizontal';
 export type GroupOrientation = 'vertical' | 'horizontal';
 export type ViewType = 'agenda' | 'day' | 'month' | 'timelineDay' | 'timelineMonth' | 'timelineWeek' | 'timelineWorkWeek' | 'week' | 'workWeek';
 export type AllDayPanelModeType = 'all' | 'allDay' | 'hidden';
+export type SnapToCellsModeType = 'always' | 'auto' | 'never';
```
Tucchhaa: Small suggestion: we can import this type directly from Scheduler's d.ts (like Appointment type is imported), instead of duplicating the type here.

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/options/get_view_model_options.ts
```diff
@@ -1,27 +1,50 @@
...
+    isTimelineView: true, isMonthView: false, viewOrientation: 'horizontal', snapToCellsMode: 'never',
+  },
+  timelineWorkWeek: {
+    isTimelineView: true, isMonthView: false, viewOrientation: 'horizontal', snapToCellsMode: 'never',
+  },
+  timelineMonth: {
+    isTimelineView: true, isMonthView: true, viewOrientation: 'horizontal', snapToCellsMode: 'always',
+  },
+  agenda: {
+    isTimelineView: false, isMonthView: false, viewOrientation: 'vertical', snapToCellsMode: 'always',
```
Tucchhaa: question: can we remove 'agenda' view from this config as it was before? Agenda view doesn't have snapping at all, and so agenda.snapToCellsMode value will never be read.

Tucchhaa: As I see in the `generateViewModel` function [here](https://github.com/DevExpress/DevExtreme/blob/7980fd56725a8a21f33db5e79aac196c9efcf5e6/packages/devextreme/js/__internal/scheduler/view_model/appointments_layout_manager.ts#L76), when generating view model for agenda, the option manager is never used. So we really don't need value for 'agenda' here

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/options/get_view_model_options.ts
```diff
@@ -1,27 +1,50 @@
...
-}> = {
-  day: { isTimelineView: false, isMonthView: false, viewOrientation: 'vertical' },
-  week: { isTimelineView: false, isMonthView: false, viewOrientation: 'vertical' },
-  workWeek: { isTimelineView: false, isMonthView: false, viewOrientation: 'vertical' },
-  month: { isTimelineView: false, isMonthView: true, viewOrientation: 'horizontal' },
-  timelineDay: { isTimelineView: true, isMonthView: false, viewOrientation: 'horizontal' },
-  timelineWeek: { isTimelineView: true, isMonthView: false, viewOrientation: 'horizontal' },
-  timelineWorkWeek: { isTimelineView: true, isMonthView: false, viewOrientation: 'horizontal' },
-  timelineMonth: { isTimelineView: true, isMonthView: true, viewOrientation: 'horizontal' },
+  snapToCellsMode: SnapToCellsModeType;
```
Tucchhaa: I suggest to move default snapToCellsMode values to [contant_view.ts](https://github.com/DevExpress/DevExtreme/blob/7980fd56725a8a21f33db5e79aac196c9efcf5e6/packages/devextreme/js/__internal/scheduler/utils/options/constants_view.ts#L21).

That file already defines specific default options for each view.

 If you move default values there, then we won't need to get [defaultSnapToCellsMode here](https://github.com/DevExpress/DevExtreme/pull/33021/changes#diff-07e798c9ab3ec21ccb4c0ef3730dfa80f9c142724f345b4e581f27dbcc7a52f3R77), because `schedulerStore.getViewOption('snapToCellsMode')` will be enough

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/generate_grid_view_model.ts
```diff
@@ -40,9 +41,11 @@ export const sortAppointments = (
...
       sortByGroupIndex(innerStep1);
       const innerStep2 = addPosition(innerStep1, optionManager.getCells(panelName));
-      const innerStep3 = isMonthView || panelName === 'allDayPanel'
-        ? snapToCells(innerStep2, optionManager.getCells(panelName))
-        : innerStep2;
+      const innerStep3 = snapToCells(
+        innerStep2,
+        optionManager.getCells(panelName),
+        panelName === 'allDayPanel' ? 'always' : snapToCellsMode,
+      );
```
Tucchhaa, line 48: Can we add tests that check that scheduler.snapToCellsMode and scheduler.view.snapToCellsMode options actually have effect?

I mean if I have this config:
```js
$().dxScheduler({
   snapToCellsMode: 'always',
   currentView: 'day'
})
```
or this config:
```js
$().dxScheduler({
   view: [{ type: 'day', snapToCells: 'always' }]
   currentView: 'day'
})
```
then check that the option is applied

packages/devextreme/js/__internal/scheduler/types.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/options/get_view_model_options.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/generate_grid_view_model.ts
sjbur: Added integration test which tests 3 cases for day view
1. Checks default behaviour without setting snapToCellsMode
2. Checks behaviour when root snapToCellsMode is set
3. Checks behaviour when snapToCellsMode is set via views config

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/options/get_view_model_options.ts
sjbur: I tried putting the default in constants_view.ts, but that merges into each view object, so getViewOption('snapToCellsMode') always sees a value on the view and never falls back to the root option. So defaults for snapToCellsMode can’t live in DEFAULT_VIEW_OPTIONS the same way as other view fields if we want root-level snapToCellsMode to work.

packages/devextreme/js/__internal/scheduler/__tests__/snap_to_cells_mode.test.ts
```diff
@@ -0,0 +1,86 @@
...
+        text: 'short',
+        startDate: new Date(2026, 2, 15, 10, 0),
+        endDate: new Date(2026, 2, 15, 10, 10),
+      }],
+      snapToCellsMode: 'always',
+    });
+
+    const appH = POM.getAppointment('short').getGeometry().height;
+
+    expect(appH).toBeGreaterThan(DEFAULT_CELL_HEIGHT * 0.85);
```
Tucchhaa: can we check that the appointment height is exactly the cell height. This way it would more intuitive than multiplying by magic number 0.85:
```js
expect(appointmentHeight).toEqual(DEFAULT_CELL_HEIGHT)

// or to roughly check
expect(appointmentHeight).toBeCloseTo(DEFAULT_CELL_HEIGHT)
```

Tucchhaa: the same applies to [the other test](https://github.com/DevExpress/DevExtreme/pull/33021/changes/6e4fccc0cf23c5ea17db6118c465c51fa7163234#diff-39538db5c5b56b5e220f6796d8ceb8ac082d396a368f450c5d60b436160a42a1R65)

Tucchhaa (approved):

PR #33022, opened 2026-03-24
DataGrid - Fix incorrect position of the 'between' filter dropdown when columns are resized (T1303927)

markallenramirez:


e2e/testcafe-devextreme/tests/dataGrid/common/filterRow/visual.ts
```diff
@@ -125,3 +125,34 @@ test('DataGrid - The `between` filter dropdown sticks to the viewport edge durin
...
+  // arrange
+  const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
+  const dataGrid = new DataGrid('#container');
+  const filterCell = dataGrid.getFilterCell(0);
+
+  // assert
+  await t.expect(dataGrid.isReady()).ok();
+
+  // act
+  await t.click(filterCell.element);
```
Alyar666: We should verify that the popup opens on click

e2e/testcafe-devextreme/tests/dataGrid/common/filterRow/visual.ts
```diff
@@ -125,3 +125,34 @@ test('DataGrid - The `between` filter dropdown sticks to the viewport edge durin
...
+  await dataGrid.resizeHeader(1, 30);
+
+  // assert
+  await testScreenshot(t, takeScreenshot, 'filter-row-filter-range-position-on-resize.png');
```
Alyar666: Question: Won’t the test be unstable? I know that hiding/showing the popup is asynchronous (because of the animation). Will we always manage to take the correct screenshot?

Alyar666 (changes requested):

e2e/testcafe-devextreme/tests/dataGrid/common/filterRow/visual.ts
markallenramirez: animation is disabled for the filterRangeOverlay (see [filter_row.ts#L305](https://github.com/DevExpress/DevExtreme/blob/26_1/packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_row.ts#L305))

markallenramirez: we decided to change the behavior. test is also updated - visual test is no longer needed

Alyar666 (approved):

PR #33046, opened 2026-03-25
Regenerate wrappers. Use explicit type exports and remake reexports

GoodDayForSurf:


apps/demos/utils/bundle/index.js
```diff
@@ -187,6 +188,28 @@ const prepareConfigs = (framework) => {
     ];
   }
 
+  if (['react', 'vue'].includes(framework)) {
```
ajivanyandev, line 191: Is this if block meant to be here?(looks like a change related to react/vue)

mpreyskurantov (dismissed):

apps/demos/utils/bundle/index.js
GoodDayForSurf: yes it's actual for react and vue for legacy systemjs-bundler we use in demos. After remake reexports systemjs-bundler can't correctly resolve imports in common reexports. Angular is bundeled different way.

ajivanyandev (approved):

PR #33075, opened 2026-03-27
DataGrid: fix rows on collapse group and virtual scroll (T1325181)

anna-shakhova:


packages/devextreme/js/__internal/grids/data_grid/grouping/m_grouping_expanded.test.ts
Alyar666, line 1: Can we write the tests in a simpler way? They look complicated, and I’m worried they’ll be hard to maintain in the future.

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/data_grid/grouping/m_grouping_expanded.test.ts
anna-shakhova: refactored tests, simplify and move helpers to separate files

Alyar666 (approved):

PR #33080, opened 2026-03-27
T1306840 - DataGrid - TagBox in the filter row doesn't show selected tags if values are numbers

Raushen:


packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_row.ts
```diff
@@ -124,9 +125,20 @@ const getColumnSelectedFilterOperation = function (that, column) {
   }
 };
 
-const isValidFilterValue = function (filterValue, column) {
-  if (column && BETWEEN_OPERATION_DATA_TYPES.includes(column.dataType) && Array.isArray(filterValue)) {
-    return false;
+const hasMultiselectEditor = function ($editorContainer): boolean {
+  const editor = getEditorInstance($editorContainer);
+  // @ts-expect-error
```
anna-shakhova: change import Editor to `import Editor from '@ts/ui/editor/editor';`
internal types contain NAME property, and it would be possible to remove `@ts-expect-error`

packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_row.ts
```diff
@@ -124,9 +125,20 @@ const getColumnSelectedFilterOperation = function (that, column) {
...
+  return !editor || MULTISELECT_EDITOR_NAMES.includes(editor.NAME);
+};
+
+const isValidFilterValue = function (filterValue, column, $editorContainer): boolean {
```
anna-shakhova: add types to arguments, please

anna-shakhova (approved):

PR #33088, opened 2026-03-30
DataGrid - Fix the issue with the focus method not setting the focus to the grid (T1308919)

markallenramirez:


e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -6576,3 +6576,57 @@ test('The last cell should be focused after changing the page size (T1063530)',
...
   });
 });
+
+test('Focus should be set to the grid to allow keyboard navigation when the focus method is called (T1308919)', async (t) => {
+  // arrange
+  const dataGrid = new DataGrid('#container');
+  const firstDataCell = dataGrid.getDataCell(0, 0);
+  const firstRow = dataGrid.getDataRow(0);
+  const secondRow = dataGrid.getDataRow(1);
+  const focusButton = ClientFunction(() => {
```
Raushen: Let's use [Button model](https://github.com/DevExpress/DevExtreme/blob/1710f03a019cd0f7a195ca7e1c3c341dc37d7eec/packages/testcafe-models/button.ts#L10) and add there apiFocus method like in the grid.

markallenramirez: I modified the test to use the Button model. No need to add 'apiFocus' because it already exists in the base Widget class ([focus()](https://github.com/DevExpress/DevExtreme/blob/26_1/packages/testcafe-models/internal/widget.ts#L88)).

Raushen (approved):

PR #33118, opened 2026-04-01
GridCore: add methods to HeaderPanel to add/remove toolbar items

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
@@ -41,8 +41,8 @@ export class HeaderPanel extends ColumnsView {
     this.createAction('onToolbarPreparing', { excludeValidators: ['disabled', 'readOnly'] });
   }
 
-  public addToolbarItem(name: string, item: ToolbarItem): void {
-    this._registeredToolbarItems.set(name, item);
+  public setToolbarItem(name: string, item: ToolbarItem): void {
+    this._registeredToolbarItems.set(name, { ...item, name });
 
     if (this._$element) {
       this._invalidate();
--- a/packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
@@ -7,7 +7,8 @@ import type { Properties as ToolbarProperties } from '@js/ui/toolbar';
 import Toolbar from '@js/ui/toolbar';
 import type { EditingController } from '@ts/grids/grid_core/editing/m_editing';
 import type { HeaderFilterController } from '@ts/grids/grid_core/header_filter/m_header_filter';
-import { normalizeToolbarItems } from '@ts/grids/new/grid_core/toolbar/utils';
+import type { DefaultToolbarItem, ToolbarItem } from '@ts/grids/new/grid_core/toolbar/types';
+import { isDefaultToolbarItem, normalizeToolbarItems } from '@ts/grids/new/grid_core/toolbar/utils';
 
 import type { ModuleType } from '../m_types';
 import { ColumnsView } from '../views/m_columns_view';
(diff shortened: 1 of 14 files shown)


packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1227,8 +1227,7 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
 
   private _ctrlFKeyHandler(eventArgs) {
     if (this.option('searchPanel.visible')) {
-      // @ts-expect-error
-      const searchTextEditor = this._headerPanel.getSearchTextEditor();
+      const searchTextEditor = this.getController('searchPanel').getSearchTextEditor();
```
Alyar666: Usually, we initialize controllers in init.
And by the way, do we need _headerPanel now?

packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
```diff
@@ -25,22 +26,49 @@ export class HeaderPanel extends ColumnsView {
...
     this._headerFilterController = this.getController('headerFilter');
+
     this.createAction('onToolbarPreparing', { excludeValidators: ['disabled', 'readOnly'] });
   }
 
+  public setToolbarItem(name: string, item: ToolbarItem): void {
+    this._registeredToolbarItems.set(name, { ...item, name });
+
+    if (this._$element) {
+      this._invalidate();
```
Alyar666, line 67: Why do we need to call the _invalidate method here? This method is only used in optionChanged.

packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
```diff
@@ -25,22 +26,49 @@ export class HeaderPanel extends ColumnsView {
...
+    }
+  }
+
+  public removeToolbarItem(name: string): void {
+    if (this._registeredToolbarItems.has(name)) {
+      this._registeredToolbarItems.delete(name);
+
+      if (this._$element) {
+        this._invalidate();
```
Alyar666: Why do we need to call the _invalidate method here? This method is only used in optionChanged.

packages/devextreme/js/__internal/grids/grid_core/search/m_search.ts
```diff
@@ -151,68 +165,89 @@ const headerPanel = (
...
+      }
+    } else {
+      this._headerPanel.removeToolbarItem(SEARCH_PANEL_ITEM_NAME);
+    }
+  }
+
+  private _getSearchPanelToolbarItem(): ToolbarItem | null {
+    const { searchPanel: searchPanelOptions } = this.option();
+
+    if (this._headerPanel && searchPanelOptions && searchPanelOptions.visible) {
```
Alyar666: It seems to me that this condition is unnecessary here. I think the _getSearchPanelToolbarItem method should always return a configuration.

packages/devextreme/js/__internal/grids/grid_core/search/m_search.ts
```diff
@@ -151,68 +165,89 @@ const headerPanel = (
...
+      return {
+        template: (data, index, container: dxElementWrapper | Element): void => {
+          if (this._headerPanel) {
```
Alyar666: This also seems redundant. If we’re not sure whether _headerPanel exists somewhere, we can use the optional chaining operator (?.) where needed.

packages/devextreme/js/__internal/grids/grid_core/search/m_search.ts
```diff
@@ -151,68 +165,89 @@ const headerPanel = (
...
   }
 
-  private getSearchTextEditor() {
-    const that = this;
-    const $element = that.element();
-    const $searchPanel = $element.find(`.${that.addWidgetPrefix(SEARCH_PANEL_CLASS)}`).filter(function () {
-      return $(this).closest(`.${that.addWidgetPrefix(HEADER_PANEL_CLASS)}`).is($element);
-    });
+  public getSearchTextEditor(): TextBox | null {
+    if (!this._headerPanel) {
```
Alyar666: This also seems redundant. If we’re not sure whether _headerPanel exists somewhere, we can use the optional chaining operator (?.) where needed.

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
anna-shakhova: we need it there to redraw toolbar after item remove, the 1st line just removes item from registered but do not update headerPanel state

packages/devextreme/js/__internal/grids/grid_core/header_panel/m_header_panel.ts
anna-shakhova: Agree, not necessary to redraw the whole headerPanel when we just update some property of button/editor
replaced it with more complex logic - of item already exists, we just update its options, otherwise redraw header panel to add item

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
anna-shakhova: fixed

packages/devextreme/js/__internal/grids/grid_core/search/m_search.ts
anna-shakhova: those condtions were added to remove ts errors, but in fact it will always return `true`, updated a bit to remove them

packages/devextreme/js/__internal/grids/grid_core/search/m_search.ts
anna-shakhova: I prefer keep it here to ensure we append template to existing headerPanel (not before 1st render) 
Also this way it is cleaner than multiple `?` and skip-errors

packages/devextreme/js/__internal/grids/grid_core/search/m_search.ts
anna-shakhova: fixed

packages/devextreme/js/__internal/grids/grid_core/search/m_search.ts
```diff
@@ -151,68 +167,80 @@ const headerPanel = (
...
+  }
 
-      items.push(toolbarItem);
-    }
+  public getSearchTextEditor(): TextBox | null {
+    const $element = this.headerPanel?.element();
 
-    return items;
-  }
+    if (!this.headerPanel || !$element) {
```
anna-shakhova, line 227: kept check here for cleaner TS checks

Alyar666 (dismissed):

Alyar666 (dismissed):

Alyar666 (approved):

PR #33158, opened 2026-04-03
Scheduler - Appointments Refactoring - Custom Templates

Tucchhaa:

PR #33159, opened 2026-04-03
Scheduler - Appointments Refactoring - Custom Templates

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
@@ -84,6 +84,43 @@ describe('New Appointments', () => {
         expect(isTemplateApplied(POM)).toBe(true);
       });
 
+      it('should apply default template if current view does not have it', async () => {
+        const defaultTemplate = jest.fn();
+
+        const { POM } = await createScheduler({
+          ...config,
+          views: [{ type: 'day' }],
+          [templateName]: defaultTemplate,
+        });
+
+        expect(defaultTemplate).toHaveBeenCalled();
+        expect(isTemplateApplied(POM)).toBe(false);
+      });
+
+      it('should apply default template after .option() change with default value', async () => {
+        const defaultValue = templateName === 'appointmentCollectorTemplate'
+          ? 'appointmentCollector'
+          : 'appointment';
+
+        const { POM, scheduler } = await createScheduler({
+          ...config,
+          [templateName]: templateFunction,
+        });
+
+        scheduler.option(templateName, defaultValue);
+        await new Promise(process.nextTick);
+
+        expect(isTemplateApplied(POM)).toBe(false);
+
+        if (templateName === 'appointmentTemplate') {
+          const appointment = POM.getAppointments()[0];
+          expect(appointment.getText()).toBe('Appointment 1');
+        } else {
+          const collectorButton = POM.getCollectorButton();
+          expect(collectorButton.textContent).toBe('1');
+        }
+      });
+
       it('should apply current view\'s template', async () => {
         const defaultTemplate = jest.fn();
 
--- a/packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
@@ -28,7 +28,7 @@ describe('New Appointments', () => {
 
   describe('Templates', () => {
     describe('AppointmentTemplate', () => {
-      it('should call template function with correct paramteters', async () => {
+      it('should call template function with correct parameters', async () => {
         const appointmentTemplate = jest.fn();
 
         const appointmentData1 = {
@@ -176,10 +176,6 @@ describe('New Appointments', () => {
 
         const $appointment = $(POM.getAppointments()[0].element);
         expect($appointment.text()).toBe('Appointment 1 Async');
-      });
-
-      it('should call onAppointmentRendered action after appointment is rendered', async () => {
-
       });
     });
 
--- a/packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
@@ -1,0 +1,359 @@
+import {
+  afterEach, beforeEach, describe, expect, it, jest,
+} from '@jest/globals';
+import $ from '@js/core/renderer';
+import type { Properties } from '@js/ui/scheduler';
+
+import { createScheduler as baseCreateScheduler } from './__mock__/create_scheduler';
+import { setupSchedulerTestEnvironment } from './__mock__/m_mock_scheduler';
+
+const createScheduler = (config: Properties) => baseCreateScheduler({
+  ...config,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  _newAppointments: true,
+});
+
+describe('New Appointments', () => {
+  beforeEach(() => {
+    setupSchedulerTestEnvironment();
+  });
+
+  afterEach(() => {
+    const $scheduler = $('.dx-scheduler');
+    // @ts-expect-error
+    $scheduler.dxScheduler('dispose');
+    document.body.innerHTML = '';
+    jest.useRealTimers();
+  });
+
+  describe('Templates', () => {
+    describe('AppointmentTemplate', () => {
+      it('should call template function with correct paramteters', async () => {
+        const appointmentTemplate = jest.fn();
+
+        const appointmentData1 = {
+          text: 'Appointment 1',
+          startDate: new Date(2015, 1, 9, 8),
+          endDate: new Date(2015, 1, 9, 9),
+        };
+        const appointmentData2 = {
+          text: 'Appointment 2',
+          startDate: new Date(2015, 1, 9, 10),
+          endDate: new Date(2015, 1, 9, 11),
+        };
+
+        await createScheduler({
+          dataSource: [appointmentData1, appointmentData2],
+          currentView: 'day',
+          currentDate: new Date(2015, 1, 9, 8),
+          appointmentTemplate,
+        });
+
+        expect(appointmentTemplate).toHaveBeenCalledTimes(2);
+        expect(appointmentTemplate.mock.calls[0]).toHaveLength(3);
+        expect(appointmentTemplate.mock.calls[0]).toEqual([
+          expect.objectContaining({
+            appointmentData: appointmentData1,
+            targetedAppointmentData: expect.objectContaining({
+              ...appointmentData1,
+              displayStartDate: appointmentData1.startDate,
+              displayEndDate: appointmentData1.endDate,
+            }),
+          }),
+          0,
+          expect.any(HTMLElement),
+        ]);
+
+        const container1 = appointmentTemplate.mock.calls[0][2] as HTMLElement;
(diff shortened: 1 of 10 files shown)


packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -169,7 +169,7 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
       this.$commonContainer.empty();
     }
 
-    viewModelDiff.forEach((diffItem) => {
+    viewModelDiff.forEach((diffItem, index) => {
```
Tucchhaa, line 190: Passing index from diff may seem incorrect, but it is exactlye how the old implementation works: https://github.com/DevExpress/DevExtreme/blob/066fbff0a5f7e31324291b616ea4fe1f38fb0c23/packages/devextreme/js/__internal/scheduler/appointments/m_appointment_collection.ts#L320

To avoid making BC I have passed index too, but in general it seems that we need to remove passing index to appointmentTemplate function

packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
```diff
@@ -0,0 +1,355 @@
...
+} from '@jest/globals';
+import $ from '@js/core/renderer';
+import type { Properties } from '@js/ui/scheduler';
+
+import { createScheduler as baseCreateScheduler } from './__mock__/create_scheduler';
+import { setupSchedulerTestEnvironment } from './__mock__/m_mock_scheduler';
+
+const createScheduler = (config: Properties) => baseCreateScheduler({
+  ...config,
+  // eslint-disable-next-line @typescript-eslint/naming-convention
```
aleksei-semikozov: unused eslint-disable — `@typescript-eslint/naming-convention` has nothing to disable here. remove it.

packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
```diff
@@ -0,0 +1,355 @@
...
+  beforeEach(() => {
+    setupSchedulerTestEnvironment();
+  });
+
+  afterEach(() => {
+    const $scheduler = $('.dx-scheduler');
+    // @ts-expect-error
+    $scheduler.dxScheduler('dispose');
+    document.body.innerHTML = '';
+    jest.useRealTimers();
```
aleksei-semikozov: no test in this file uses `jest.useFakeTimers()`, so this restore is dead code. remove.

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.test.ts
```diff
@@ -174,13 +174,24 @@ describe('AppointmentCollector', () => {
       { isCompact: false, expectedText: '1 more' },
     ])('should have correct text for appointmentsCount = 1 and isCompact = %o', ({ isCompact, expectedText }) => {
```
aleksei-semikozov: stale name in description — `appointmentsCount` is no longer a property after this PR. rename to "for items.length === 1" or similar.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
aleksei-semikozov: passing the diff position as the public `itemIndex` to `appointmentTemplate` is wrong. on incremental updates the same appointment can show up at any position in `viewModelDiff` (it's a delta, not the visible list), so the documented stable index becomes random.

you flagged this in your own inline comment further down — calling it BC parity is correct, but it means we ship a documented bug. two ways out:

(a) compute the real visible index (from `appointmentBySortIndex` length / `sortedIndex` order) before passing it through, so the public contract holds.
(b) deprecate and stop passing the index, per the path you suggested in #3612, then drop the second arg in a follow-up.

either way needs a regression test that triggers an incremental update (not a full first render) and asserts the same appointment keeps the same `itemIndex` across update.

sjbur: How about adding a todo to remove index in future? Seems valid in this case

aleksei-semikozov (commented):
### `option('appointmentTemplate', ...)` silent no-op
`m_scheduler.ts` `_optionChanged` does `this._appointments.option('itemTemplate', value)`. new `Appointments` has no `itemTemplate` — only `appointmentTemplate`. so under `_newAppointments: true`, runtime template change does nothing. no test catches it (every test sets the template at construction).

### new `Appointments._optionChanged` ignores templates
`appointments.ts:88-109` only handles `items` and `viewModel`. `appointmentTemplate`, `appointmentCollectorTemplate`, `currentView`, `onAppointmentRendered`, `$allDayContainer` — all no-op.

### stale TODO
`m_scheduler.ts:1055` `// TODO<Appointments>: set custom templates`. this PR closes it. drop.

### naming `appointments_new`
ages badly. you mentioned it becomes `appointments/` after old removed. fine temporarily. while it lives the name says nothing. `_lite` or `_pure` would say "no CollectionWidget bloat". minor.

packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
aleksei-semikozov: dead. no fake timers in file. drop.

packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
```diff
@@ -0,0 +1,355 @@
...
+          }),
+          0,
+          expect.any(HTMLElement),
+        ]);
+
+        const $appointment = $(POM.getAppointments()[0].element);
+        expect($appointment.text()).toBe('Appointment 1 ViewTemplate');
+      });
+
+      it('should render async template', async () => {
```
aleksei-semikozov: only happy path. add:
- `dispose()` during pending promise → no DOM write after dispose
- `option('appointmentTemplate', x)` before previous resolves

`base_appointment.ts:127` `when().done()` has no stale guard.

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.test.ts
aleksei-semikozov: `appointmentsCount` gone, replaced by `appointmentsData`. rename describe to e.g. `for items.length === 1`.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
aleksei-semikozov, line 190: agree with what you wrote in your own comment below — the diff index isn't a stable visible index, this is BC parity with a legacy bug. let's track it as a follow-up so it doesn't get forgotten and fix it the way you suggested below (drop passing the arg).

packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
```diff
@@ -0,0 +1,355 @@
...
+          ],
+          currentView: 'day',
+          currentDate: new Date(2015, 1, 9, 8),
+          maxAppointmentsPerCell: 2,
+          appointmentCollectorTemplate,
+        });
+
+        expect(appointmentCollectorTemplate).toHaveBeenCalledTimes(1);
+        expect(appointmentCollectorTemplate.mock.calls[0]).toHaveLength(2);
+        expect(appointmentCollectorTemplate.mock.calls[0]).toEqual([
```
sjbur, line 272: In this test we want to be sure that appointment collector is rendering properly and is having only one item. How about adding an additional assertion that checks that in items there is only one item?

Something like 
expect(data.items).toHaveLength(1);
expect(data.items[0]).toEqual(expect.objectContaining({
  text: 'Appointment 3',
}));

We have appointmentCount: 1 check here, but if we have a bug and in items being passed more than 1 appt we won't detect it in test

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: added todo comment

packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
Tucchhaa: done

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.test.ts
Tucchhaa: done

packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
Tucchhaa: update test to:
```js
        expect(appointmentCollectorTemplate.mock.calls[0]).toEqual([
          expect.objectContaining({
            appointmentCount: 1,
            isCompact: true,
            items: [
              expect.objectContaining({
                text: 'Appointment 3',
              }),
            ],
          }),
          expect.any(HTMLElement),
        ]);
```

now it should pass only if there is one item in the array

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
```diff
@@ -120,20 +120,20 @@ export class BaseAppointmentView<
       ? this.defaultAppointmentTemplate
       : this.option().appointmentTemplate;
 
-    const $renderPromise = template.render({
```
Tucchhaa, line 123: Turns out template.render doesn't return promise even if async template is passed. In the old impl onAppointmentRendered is called right after the templateFunction is called and it doesn't wait for the promise to complete ([codepen](https://codepen.io/FufelShmertz/pen/QwKZqMB?editors=0010))

Tucchhaa: Because of that, there's no need to test async templates

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -520,6 +525,12 @@ class Scheduler extends SchedulerOptionsBaseWidget {
         this.repaint();
         break;
       case 'appointmentCollectorTemplate':
+        if (this.option('_newAppointments')) {
```
Tucchhaa, line 528: for newAppointments need only to update collector template, instead of the whole scheduler rerender, as collector template doesn't affect viewModel

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -1039,15 +1050,9 @@ class Scheduler extends SchedulerOptionsBaseWidget {
     this._layoutManager = new AppointmentLayoutManager(this);
 
     if (this.option('_newAppointments')) {
-      // TODO<Appointments>: convert 'item' to 'appointment' for compatibility
```
Tucchhaa, line 1042: Moved compatibility normalization to appointments.ts:
https://github.com/DevExpress/DevExtreme/pull/33158/changes#diff-fe8d027bace8707fc88bc8c66b807f3d05219c21aff8edd3b187f70a08c1f641R109

sjbur (approved):

aleksei-semikozov (approved):

PR #33164, opened 2026-04-06
Scheduler — Replace underscore-prefixed: Cherry-pick to 25_1 and validate completion

aleksei-semikozov:


packages/devextreme/js/__internal/scheduler/header/m_header.ts
```diff
@@ -210,7 +210,7 @@ export class SchedulerHeader extends Widget<HeaderOptions> {
 
   private _updateCalendarValueAndCurrentDate(date: Date): void {
     this._updateCurrentDate(date);
-    this._calendar?.option('value', date);
+    this.calendar?.option('value', date);
   }
 
   public _updateCurrentDate(date: Date): void {
```
Tucchhaa: This method still has underscore, as well as some other methods in this file. Was it omitted on purpose? In 26.2 it seems that, it doesn't have:
https://github.com/DevExpress/DevExtreme/blob/066fbff0a5f7e31324291b616ea4fe1f38fb0c23/packages/devextreme/js/__internal/scheduler/header/m_header.ts#L217

I don't want to check if all of the methods still have or not the underscores, because it would be just too much work for the branch that will stop being maintained in several months. 

So I suggest to rename methods in this class, but only if it won't take a lot of your time

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator_month.ts
Tucchhaa: Same  as here: https://github.com/DevExpress/DevExtreme/pull/33164#discussion_r3051491174

The 26.1 doesn't have _ here: 
https://github.com/DevExpress/DevExtreme/blob/066fbff0a5f7e31324291b616ea4fe1f38fb0c23/packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator_month.ts#L54

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
Tucchhaa: Same as here: https://github.com/DevExpress/DevExtreme/pull/33164/changes#r3051491174

26.1 doesn't have _ here:
https://github.com/DevExpress/DevExtreme/blob/26_1/packages/devextreme/js/__internal/scheduler/header/m_calendar.ts#L118

Tucchhaa (dismissed):

packages/devextreme/js/__internal/scheduler/header/m_header.ts
aleksei-semikozov: Renamed to `updateCurrentDate` and updated the call site in `m_date_navigator.ts`.

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator_month.ts
aleksei-semikozov: Kept as `_calculateCellIndex`

packages/devextreme/js/__internal/scheduler/header/m_calendar.ts
aleksei-semikozov: Renamed to `getCalendarOptions`.

Tucchhaa (approved):

PR #33168, opened 2026-04-06
DataGrid - Fix the browser scrolling back to the grid when data is updated (T1310557)

markallenramirez:


packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1809,7 +1813,7 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
     }
   }
 
-  public _updateFocus(isRenderView?, skipFocusEvent = false) {
+  public _updateFocus(isRenderView?: boolean, skipFocusEvent = false) {
```
Raushen: There are three usages of _updateFocus without arguments. Let's update the first argument to
`isRenderView: boolean`
and pass the required false argument in those usages.

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1848,12 +1852,12 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
                 );
                 return;
               }
-              !isFocusedElementDefined && this._focus($cell, false, skipFocusEvent);
+              !isFocusedElementDefined && this._focus($cell, false, skipFocusEvent, !!isRenderView);
```
Raushen: So we could avoid of using !!

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -1795,8 +1795,12 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
       });
       if (!skipFocusEvent) {
         this._applyTabIndexToElement($focusElement);
-        // @ts-expect-error
-        eventsEngine.trigger($focusElement, 'focus');
+        if (preventScroll) {
+          $focusElement.get(0)?.focus({ preventScroll });
+        } else {
+          // @ts-expect-error
+          eventsEngine.trigger($focusElement, 'focus');
```
Raushen, line 1802: As I see eventsEngine.trigger calls the [jQuery.trigger](https://api.jquery.com/trigger/) method which supports additional parameters. 
Can we apply the following code?
eventsEngine.trigger($focusElement, 'focus', { preventScroll });

markallenramirez: We can't. The additional parameters are for the callback's parameters, not for passing options to the native event.
e.g.,
```
$element.on('focus', (event, param1, param2, ...) => {
    // param1 = { preventScroll: true }
})
$element.trigger('focus', { preventScroll: true });

// same as
DevExpress.events.on($element, 'focus', (event, param1, param2, ...) => {
  // param1 = { preventScroll: true }
})
DevExpress.events.trigger($element, 'focus', { preventScroll: true });
```

I don't see any dxEvent or jQuery API that can trigger native events with options.

Raushen (approved):

PR #33188, opened 2026-04-07
Scheduler — Add hiddenDays option to hide arbitrary days of the week

aleksei-semikozov:


packages/devextreme/js/__internal/scheduler/utils/skipped_days.ts
```diff
@@ -0,0 +1,60 @@
...
+  return result;
+};
+
+export const getFirstVisibleDate = (
+  start: Date,
+  skippedDays: number[],
+  nextDate: (date: Date) => Date,
+): Date => {
+  let date = new Date(start);
+  while (isDateSkipped(date, skippedDays)) {
```
aleksei-semikozov, line 38: Infinite loop risk: if `skippedDays` contains all 7 days (bypassing `normalizeHiddenWeekDays`), this loops forever. Add a safety counter:
```ts
while (isDateSkipped(date, skippedDays) && ++guard < 8)
```
or rewrite to `for...`

packages/devextreme/js/__internal/scheduler/workspaces/m_timeline.ts
```diff
@@ -109,6 +109,9 @@ class SchedulerTimeline extends SchedulerWorkSpace {
 
   protected incrementDate(date) {
     date.setDate(date.getDate() + 1);
+    while ((this.option('skippedDays') ?? []).includes(date.getDay())) {
```
aleksei-semikozov: Same infinite loop risk — while loop skipping days has no max iterations guard.

packages/devextreme/js/__internal/scheduler/__tests__/__snapshots__/santiago_timezone.test.ts.snap
```diff
@@ -307,7 +307,7 @@ exports[`scheduler should render correct workspace in Santiago DST for view: Tim
   "Wed 4",
   "Thu 5",
   "Fri 6",
-  "Sat 7",
+  "Mon 9",
```
aleksei-semikozov, line 310: Snapshot changed `"Sat 7"` → `"Mon 9"` in TimelineWorkWeek — is this intentional? Old snapshot had Mon-Sat (7 days), now skips Sat/Sun. Please confirm the new `incrementDate` logic correctly handles TimelineWorkWeek headers.

packages/devextreme/js/__internal/scheduler/utils/skipped_days.ts
sjbur: The risk is theoretical, because skippedDays can't have all days because of our validation, but still makes sense. I added if condition which checks additionally if the skippedDays is 7 and fixes the potential issue

packages/devextreme/js/__internal/scheduler/workspaces/m_timeline.ts
sjbur: Fixed

packages/devextreme/js/__internal/scheduler/__tests__/__snapshots__/santiago_timezone.test.ts.snap
sjbur: This is intended. This is bug not related with timezone. In the snapshot we saved a buggy state that no one noticed is buggy. In timelineWorkWeek we can't have weekend days. You can see the bug here: https://codepen.io/sjbur/pen/Kwgbqdr

Also attaching two screenshots: the first is original state with bug, second one is current implementation without bug.

Image
Image

sjbur (dismissed):

packages/devextreme/js/ui/scheduler.d.ts
```diff
@@ -685,6 +685,12 @@ export interface dxSchedulerOptions extends WidgetOptions<dxScheduler> {
      * @public
      */
     firstDayOfWeek?: FirstDayOfWeek | undefined;
+    /**
+     * @docid
+     * @default undefined
+     * @public
+     */
+    hiddenWeekDays?: Array<number>;
```
Tucchhaa: Suggestion: can we change it to?
```ts
type DayOfWeek = 0 | 1 | 2 | 3 | 4 | 5 | 6;

hiddenWeekDays?: Array<DayOfWeek>
```

[scheduler.firstDayOfWeek](https://github.com/DevExpress/DevExtreme/blob/8749abb6bc51dcd11bd4db2f9c6c844760497ee5/packages/devextreme/js/ui/scheduler.d.ts#L687) for example has a specific type with only correct numbers.

packages/devextreme/js/__internal/scheduler/utils/options/utils.ts
```diff
@@ -1,31 +1,104 @@
...
+    defaultView.skippedDays,
+  );
+  if (skippedDays === defaultView.skippedDays) {
+    return defaultView;
+  }
+  return { ...defaultView, skippedDays };
+};
 
-export const getViews = (views: RawViewType[]): NormalizedView[] => views
+export const getViews = (
```
Tucchhaa: I think current logic to get normalized hiddenDays value is really over complicated.

I understand that existing mechanism (getViewOption) to get view option in scheduler isn't suitable, because if a view has a default value, then global value wouldn't apply. I think existing mechanism is very awkward and should be refactored.

But in the meantime, I think it is possible to greatly simplify the logic you've added. Here's how:

1. add this code to [scheduler_options_base_widget.ts](https://github.com/DevExpress/DevExtreme/blob/8749abb6bc51dcd11bd4db2f9c6c844760497ee5/packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts#L101):

```ts
getViewOption(optionName: string)
    if (optionName === 'hiddenDays') {
      return normalizeViewHiddenDays(
        this.currentView?.type,
        this.currentView?.[optionName] ?? this.option(optionName)
      );
    }

    // ...
}
```

2. Remove skipped days default values from [DEFAULT_VIEW_OPTIONS](https://github.com/DevExpress/DevExtreme/blob/8749abb6bc51dcd11bd4db2f9c6c844760497ee5/packages/devextreme/js/__internal/scheduler/utils/options/constants_view.ts#L25)


3. Implement normalizeViewHiddenDays:

```ts
const normalizeViewHiddenDays = (view: ViewType, hiddenDays?: HiddenDays) => {
   const value = hiddenDays ? hiddenDays : defaultValueByView[view];
   
    // other checks and error logging

    return value;
}
```

What do you think?

Tucchhaa (changes requested):

packages/devextreme/js/__internal/scheduler/utils/options/utils.ts
```diff
@@ -1,31 +1,104 @@
...
+};
+
 const isKnownView = (view: RawViewType): boolean => VIEW_TYPES
   .includes((isObject(view) ? view.type : view) as ViewType);
 const isExistedView = (view: NormalizedView | undefined): view is NormalizedView => Boolean(view);
 
-const normalizeView = (view: RawViewType): NormalizedView | undefined => (isObject(view)
-  ? extend({}, DEFAULT_VIEW_OPTIONS[view.type as string], view) as NormalizedView
-  : DEFAULT_VIEW_OPTIONS[view]);
+const normalizeView = (
```
Tucchhaa: I think current logic to get normalized hiddenDays value is really over complicated.

I understand that existing mechanism (getViewOption) to get view option in scheduler isn't suitable, because if a view has a default value, then global value wouldn't apply. I think existing mechanism is very awkward and should be refactored.

But in the meantime, I think it is possible to greatly simplify the logic you've added. Here's how:

1. add this code to [scheduler_options_base_widget.ts](https://github.com/DevExpress/DevExtreme/blob/8749abb6bc51dcd11bd4db2f9c6c844760497ee5/packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts#L101):

```ts
getViewOption(optionName: string)
    if (optionName === 'hiddenDays') {
      return normalizeHiddenDaysViewOption(
        this.currentView?.type,
        this.currentView?.[optionName] ?? this.option(optionName)
      );
    }

    // ...
}
```

2. Remove skipped days default values from [DEFAULT_VIEW_OPTIONS](https://github.com/DevExpress/DevExtreme/blob/8749abb6bc51dcd11bd4db2f9c6c844760497ee5/packages/devextreme/js/__internal/scheduler/utils/options/constants_view.ts#L25)


3. Implement normalizeHiddenDaysViewOption:

```ts
const normalizeHiddenDaysViewOption = (view: ViewType, hiddenDays?: HiddenDays) => {
   const value = hiddenDays ? hiddenDays : defaultValueByView[view];
   
    // other checks and error logging

    return value;
}
```

Tucchhaa: if need to pass hiddenDays to somewhere, e.g. Header, then just pass it in the config, [like 'currentDate' is passed](https://github.com/DevExpress/DevExtreme/blob/8749abb6bc51dcd11bd4db2f9c6c844760497ee5/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L1293):

```js
private headerConfig(): {
   return {
       hiddenDays: this.getViewOption('hiddenDays')
       // ...
   }
}

```

packages/devextreme/js/__internal/scheduler/utils/skipped_days.ts
```diff
@@ -0,0 +1,64 @@
+export const isValidWeekday = (value: unknown): value is number => (
```
Tucchhaa, line 1: minor: now we have two names for the same entity: skippedDays and hiddenDays.  I would suggest to only use one name in the codebase

packages/devextreme/js/ui/widget/ui.errors.js
```diff
@@ -397,4 +397,8 @@ export default errorUtils(errors.ERROR_MESSAGES, {
      * @name ErrorsUIWidgets.W1028
      */
     W1028: 'Nested/banded columns do not support the following properties: {0}.',
+    /**
+     * @name ErrorsUIWidgets.W1029
+     */
+    W1029: '"hiddenWeekDays" must leave at least one weekday visible.',
```
arman-boyakhchyan:
```suggestion
    W1029: '\'hiddenWeekDays'\ must leave at least one weekday visible.',
```

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
```diff
@@ -98,32 +115,40 @@ const getPeriodEndDate = (
...
   let date = addMS(currentPeriodEndDate);
 
   if (step === 'workWeek') {
-    while (isWeekend(date)) {
-      date = nextDay(date);
-    }
+    date = getFirstVisibleDate(date, skippedDays, nextDay);
+  } else if (step === 'week' && skippedDays.length > 0) {
+    date = getFirstVisibleDate(date, skippedDays, nextDay);
   }
```
Tucchhaa, line 139: This code is equivalent to this:
```js
if (step === 'week' || step === 'workWeek') {
  date = getFirstVisibleDate(date, skippedDays, nextDay);
}
```

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
```diff
@@ -78,15 +87,23 @@ const nextAgendaStart = (
...
+    case 'week': {
+      const weekStart = getPeriodStart(date, step, false, firstDayOfWeek) as Date;
+      if (skippedDays.length > 0) {
+        return getFirstVisibleDate(weekStart, skippedDays, nextDay);
+      }
+      return weekStart;
+    }
     case 'workWeek':
-      return getWorkWeekStart(getWeekStart(date, firstDayOfWeek));
+      return getWorkWeekStart(date, firstDayOfWeek, skippedDays);
```
Tucchhaa: Am I correct that week and workWeek essentially the same, except that 'workWeek' has skippedDays by default?

If yes, then why not handle both cases in the same way?:

```js
    case 'workWeek':
    case 'week': {
      const weekStart = getPeriodStart(date, step, false, firstDayOfWeek) as Date;
      return getFirstVisibleDate(weekStart, skippedDays, nextDay);
    }
```

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
```diff
@@ -46,30 +45,40 @@ const nextMonth = (date: Date): Date => {
...
+  firstDayOfWeek: number | undefined,
+  skippedDays: number[],
+): Date => {
+  if (skippedDays.length === 0) {
+    return nextWeek(weekStartDate);
   }
 
-  return date;
+  return getDateAfterWorkWeek(weekStartDate, firstDayOfWeek, skippedDays);
 };
```
Tucchhaa: Can we unify this functions?:
```js
const getDateAfterWeek = (
  workWeekStart: Date,
  firstDayOfWeek: number | undefined,
  skippedDays: number[],
): Date => {
  const weekStart = getWeekStart(workWeekStart, firstDayOfWeek);
  let lastVisibleDate = addDateInterval(weekStart, { days: 6 }, 1);
  while (isDateSkipped(lastVisibleDate, skippedDays)) {
    lastVisibleDate = addDateInterval(lastVisibleDate, DAY_DURATION, -1);
  }

  return nextDay(lastVisibleDate);
};
```

And use only this function. I think we can do it because results of nextWeek and getDateAfterWorkWeek should be the same in case of `skippedDays.length===0`

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
```diff
@@ -1302,36 +1305,44 @@ class SchedulerWorkSpace extends Widget<WorkspaceOptionsInternal> {
...
+      result = fullInterval - days * hiddenInterval - skippedDaysCount * visibleDayDuration;
     }
 
     return result;
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  protected getWeekendsCount(argument?: any) {
-    return 0;
+  protected getSkippedDaysCount(days: number, startDate: Date = this.getStartViewDate()) {
```
Tucchhaa: minor: The parameters order of `getSkippedDaysCount` and `countSkippedDays` is different which maybe confusing.

I suggest to make it the same:
```
protected getSkippedDaysCount(startDate: Date, days: number)
```

Since the default value of the startDate param is used only once, then it's possible just to pass `this.getStartViewDate()`

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space_work_week.ts
```diff
@@ -10,11 +8,10 @@ const WORK_WEEK_CLASS = 'dx-scheduler-work-space-work-week';
 class SchedulerWorkSpaceWorkWeek extends SchedulerWorkSpaceWeek {
   get type() { return VIEWS.WORK_WEEK; }
 
-  constructor(...args: any[]) {
-    // @ts-expect-error
-    super(...args);
-
-    this.getWeekendsCount = getWeekendsCount;
+  _getDefaultOptions() {
+    return extend(super._getDefaultOptions(), {
```
Tucchhaa: Why do we need this default options if the workWeek view has skippedDays by default?

packages/devextreme/js/__internal/scheduler/workspaces/m_timeline_work_week.ts
```diff
@@ -1,35 +1,23 @@
...
 class SchedulerTimelineWorkWeek extends SchedulerTimelineWeek {
   get type() { return VIEWS.TIMELINE_WORK_WEEK; }
 
-  constructor(...args: any[]) {
-    // @ts-expect-error
-    super(...args);
-
-    this.getWeekendsCount = getWeekendsCount;
+  _getDefaultOptions() {
+    return extend(super._getDefaultOptions(), {
```
Tucchhaa: This same questions as here: https://github.com/DevExpress/DevExtreme/pull/33188/changes#r3085512631

packages/devextreme/js/__internal/scheduler/workspaces/m_timeline.ts
```diff
@@ -108,7 +108,16 @@ class SchedulerTimeline extends SchedulerWorkSpace {
...
+    const skippedDays = this.option('skippedDays') ?? [];
+
+    if (skippedDays.length >= 7) {
+      return;
+    }
+
     date.setDate(date.getDate() + 1);
+    while (skippedDays.includes(date.getDay())) {
+      date.setDate(date.getDate() + 1);
+    }
```
Tucchhaa: It's possible to use `getFirstVisibleDate` helper here

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
```diff
@@ -28,26 +32,82 @@ import type {
...
+  get daysInInterval(): number {
+    if (this.skippedDays.length === 0) {
+      return this.baseDaysInInterval;
+    }
+    const visibleDayCount = 7 - this.skippedDays.length;
+    if (this.baseDaysInInterval >= 7) {
+      return visibleDayCount;
+    }
+    return this.baseDaysInInterval;
+  }
```
Tucchhaa, line 54: Basically what this getter does is:
if currentView is week or workWeek, then `return 7 - skippedDays.length`
else `return 1`

Can you simplify this getter to make it more clear?

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
```diff
@@ -28,26 +32,82 @@ import type {
...
+
+  protected getSkippedDaysAnchorDay(
+    firstDayOfWeekOption: number | undefined,
+    startViewDate: Date, // eslint-disable-line @typescript-eslint/no-unused-vars
+  ): number {
+    return this.getFirstDayOfWeek(firstDayOfWeekOption) ?? 0;
+  }
+
+  protected getVisibleDayOffset(
```
Tucchhaa: this func can be private

packages/devextreme/js/ui/scheduler.d.ts
sjbur: Done

packages/devextreme/js/ui/widget/ui.errors.js
sjbur: Done

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
```diff
@@ -28,26 +32,82 @@ import type {
...
 
   protected tableAllDay = false;
 
   public hiddenInterval = 0;
 
+  public skippedDays: number[] = [];
+
   constructor(public readonly viewType: ViewType) {}
 
+  get daysInInterval(): number {
```
Tucchhaa, line 43: add protected

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
```diff
@@ -28,26 +32,82 @@ import type {
...
+    const idxInWeek = dayIndex % visibleCount;
+    const targetDayOfWeek = rotated[idxInWeek];
+    const naiveDayOffset = dayIndex;
+    const actualDayOffset = week * 7 + ((targetDayOfWeek - anchorDay + 7) % 7);
+    return actualDayOffset - naiveDayOffset;
+  }
+
+  public isSkippedDate(date: Date): boolean {
+    return isDateSkipped(date, this.skippedDays);
+  }
```
Tucchhaa: Use the same name for the method:
```js
public isDateSkipped(date: Date): boolean {
    return isDateSkipped(date, this.skippedDays);
  }
```

Tucchhaa (changes requested):

Tucchhaa:
### Day view Bug

```js
  $('#scheduler').dxScheduler({
    timeZone: 'America/Los_Angeles',
    dataSource: data,
    views: ['day', 'week', 'workWeek', 'month', {
      type: 'day',
      name: 'myday',
      intervalCount: 7,
    }],
    hiddenWeekDays: [0, 1, 2, 3, 4, 5],
    currentView: 'myday',
    currentDate: new Date(2021, 3, 29),
    startDayHour: 9,
    height: 730,
  });
```

**Actual result**: all week days are visible

**Expected result:** hidden week days shouldn't be visible

Looks like we forgot about this case during the spec writing 😔

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
sjbur: The function is being used here /DevExtreme/packages/devextreme/js/__internal/scheduler/workspaces/m_timeline.ts (421,68):
 Property 'daysInInterval' is protected and only accessible within class 'ViewDataGenerator' and its subclasses. 

and in tests. 

Cant make it protected or private

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/workspaces/m_timeline.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/utils/skipped_days.ts
sjbur: Using skippedDays internally makes sense because skippedDays is a normalized value, while hiddenWeekDays is a public API option and may contain any user-provided input.

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
sjbur: Done

sjbur:
> ### Day view Bug
> ```js
>   $('#scheduler').dxScheduler({
>     timeZone: 'America/Los_Angeles',
>     dataSource: data,
>     views: ['day', 'week', 'workWeek', 'month', {
>       type: 'day',
>       name: 'myday',
>       intervalCount: 7,
>     }],
>     hiddenWeekDays: [0, 1, 2, 3, 4, 5],
>     currentView: 'myday',
>     currentDate: new Date(2021, 3, 29),
>     startDayHour: 9,
>     height: 730,
>   });
> ```
> 
> **Actual result**: all week days are visible
> 
> **Expected result:** hidden week days shouldn't be visible
> 
> Looks like we forgot about this case during the spec writing 😔

I think during chat with the PM the feature support for day view was more like 'nice to have' rather than 'must have'. But I looked at the code and it was rather easy to implement it. I tested, works fine. Added necessary tests.

packages/devextreme/js/__internal/scheduler/header/m_utils.ts
sjbur: rewrote to

    case 'week':
    case 'workWeek': {
      const weekStart = getPeriodStart(date, 'week', false, firstDayOfWeek) as Date;
      return getFirstVisibleDate(weekStart, skippedDays, nextDay);
    }

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space_work_week.ts
sjbur: You are right. In current implementation _getDefaultOptions() { here is needed only for pure workspace test without creation of entire Scheduler, as in here:

workspace.base.test.ts

```
  it('should skip weekend by default in timeline work week', () => {
    const { workspace } = createWorkspace(SchedulerTimelineWorkWeek, 'timelineWorkWeek');
    const date = new Date(2026, 3, 10); // Friday

    (workspace as any).incrementDate(date);

    expect(date).toEqual(new Date(2026, 3, 13)); // Monday
  });
```

If remove getDefaultOptions then we need to specify for this and all future tests with (timeline)work week the skippedDays option, like this:

```
  it('should skip weekend in timeline work week when skippedDays are Sat/Sun', () => {
    const { workspace } = createWorkspace(SchedulerTimelineWorkWeek, 'timelineWorkWeek', {
      skippedDays: [0, 6],
    });
    const date = new Date(2026, 3, 10); // Friday

    (workspace as any).incrementDate(date);

    expect(date).toEqual(new Date(2026, 3, 13)); // Monday
  });
```

packages/devextreme/js/__internal/scheduler/utils/options/utils.ts
sjbur: I am not sure this is a good idea.

It moves hiddenWeekDays normalization to getViewOption(), but that method works only for the current view. Because of that, NormalizedView would no longer be fully normalized, and DEFAULT_VIEW_OPTIONS would no longer be true internal defaults.

Also, the hidden days logic would be spread across different places instead of being resolved once in normalizeView(). So even if the current code is not perfect, this change would make the design less clear, not simpler.

currentDate is just one simple value, so it is easy to normalize in getViewOption().

hiddenWeekDays is more complex. Its final value depends on the view type, per-view value, global value, and default view settings. So it is better to resolve it in normalizeView(), where the view config is built.

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space_work_week.ts
Tucchhaa: I suggest to create a util function in the test that would create workSpace with the needed default parameters. 

I think it is a bad practice to make changes to the code only to make tests work. If it's not needed in the code itself, then I suggest to remove it

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
Tucchhaa: Is there a way to check if the currentView is week or workWeek without overriding baseDaysInInterval property in m_view_data_generator_week or m_view_data_generator_work_week?

I mean, something like:
```js
const isWeekLikeView = this.viewType === 'week' || this.viewType === 'workWeek';

return isWeekLikeView
  ? 7 - this.skippedDays.length
   : 1;
```

Thus we wouldn't need baseDaysInInterval prop at all, because in all other viewDataGenerators apart from week don't override the value of `baseDaysInInterval`

packages/devextreme/js/__internal/scheduler/utils/options/utils.ts
sjbur: After discussion, updated to your solution

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space_work_week.ts
sjbur: After discussion decided to delete the default option config and related tests

packages/devextreme/js/__internal/scheduler/workspaces/view_model/m_view_data_generator.ts
sjbur: Updated

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -514,6 +514,8 @@ class Scheduler extends SchedulerOptionsBaseWidget {
         this.updateOption('workSpace', name, value);
         this.repaint();
         break;
+      case 'skippedDays':
+        break;
```
Tucchhaa, line 518: how this option can be changed? Public option name is `hiddenWeekDays`

packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts
```diff
@@ -71,6 +73,7 @@ export class SchedulerOptionsBaseWidget extends Widget<SafeSchedulerOptions> {
     switch (args.name) {
       case 'currentView':
       case 'views':
+      case 'hiddenWeekDays':
```
Tucchhaa: let's remove this, not relevant anymore

packages/devextreme/js/__internal/scheduler/utils/options/utils.test.ts
```diff
@@ -126,7 +136,116 @@ describe('views utils', () => {
         },
       },
     ])('should return normalized $input.type view', ({ input, output }) => {
-      expect(getViews([input] as any)).toEqual([{ ...output, skippedDays: [0, 6] }]);
+      expect(getViews([input as RawViewType])).toEqual([{ ...output, skippedDays: [0, 6] }]);
+    });
+
+    describe('hiddenWeekDays', () => {
```
Tucchhaa: Is it possible to move tests in this describe group to normalize_hidden_days.test.ts?

packages/devextreme/js/__internal/scheduler/scheduler_options_base_widget.ts
sjbur: Done

packages/devextreme/js/__internal/scheduler/utils/options/utils.test.ts
sjbur: Done

Tucchhaa (approved):

PR #33201, opened 2026-04-08
DataGrid: Fix focus getting stuck on Save button when columnRenderingMode is virtual (T1326106)

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts
```diff
@@ -964,6 +921,54 @@ export class KeyboardNavigationController extends KeyboardNavigationControllerCo
     return Deferred().resolve().promise();
   }
 
+  private getNextColumnIndex(
```
Raushen: We can extract it in utils and cover by unit tests as the method does not have any relations within controller.

packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/utils.ts
```diff
@@ -0,0 +1,8 @@
+import type { NavigationDirection } from './types';
```
Raushen, line 20: The main file for utils for KNB is m_keyboard_navigation_utils.ts
Can we add the method there or remane those file to utils.ts?

Raushen (dismissed):

Raushen (approved):

PR #33225, opened 2026-04-10
AIChat: popup and chat default configuration

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/ai_chat/ai_chat.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/ai_chat/ai_chat.ts
@@ -1,19 +1,29 @@
+import messageLocalization from '@js/common/core/localization/message';
 import $ from '@js/core/renderer';
 import type { Properties as ChatProperties } from '@js/ui/chat';
 import Chat from '@js/ui/chat';
-import type { Properties as PopupProperties } from '@js/ui/popup';
+import type { Properties as PopupProperties, ToolbarItem } from '@js/ui/popup';
+import {
+  CHAT_MESSAGELIST_EMPTY_IMAGE_CLASS,
+  CHAT_MESSAGELIST_EMPTY_MESSAGE_CLASS,
+  CHAT_MESSAGELIST_EMPTY_PROMPT_CLASS,
+} from '@ts/ui/chat/messagelist';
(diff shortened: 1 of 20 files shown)


packages/devextreme/js/__internal/grids/grid_core/ai_chat/ai_chat.ts
```diff
@@ -22,19 +32,35 @@ export class AIChat {
   }
 
   private getChatConfig(): ChatProperties {
-    return {};
+    return {
+      ...DEFAULT_CHAT_OPTIONS,
+      emptyViewTemplate: (_data, container): void => {
+        const $image = $('<div>')
```
Alyar666, line 38: Can we extract everything related to the DOM into a separate file – dom.ts? (For example, as it is implemented in sticky_columns)

packages/devextreme/js/__internal/grids/grid_core/ai_chat/ai_chat.ts
```diff
@@ -46,9 +72,41 @@ export class AIChat {
...
+      location: 'after',
+      options: {
+        icon: CLEAR_CHAT_ICON,
+        hint: messageLocalization.format('dxDataGrid-aiAssistantClearButtonText'),
+        onClick: onChatCleared,
+      },
     };
   }
 
+  public updateOptions(options: AIChatOptions, updatePopup: boolean, updateChat: boolean): void {
```
Alyar666, line 98: Minor: rename updatePopup -> needToUpdatePopup, updateChat -> needToUpdateChat

Alyar666 (approved):

packages/devextreme/js/__internal/grids/grid_core/ai_chat/ai_chat.ts
anna-shakhova: Currently there is not so much DOM related code in class, lets do it later if we see it is needed (after messageTemplate for example)

packages/devextreme/js/__internal/grids/grid_core/ai_chat/ai_chat.ts
anna-shakhova: lets fix it in next PR

marker-dao (approved):

r-farkhutdinov (approved):

PR #33238, opened 2026-04-13
Scheduler - Appointments Collection Refactoring - Move getBaseAppointmentViewProperties Mock

Tucchhaa:


packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -65,6 +65,11 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
       appointment: new EmptyTemplate(),
       appointmentCollector: new EmptyTemplate(),
     });
+
+    // TODO: legacy compatibility
+    if (this.option().appointmentTemplate === 'item') {
+      this.option('appointmentTemplate', 'appointment');
```
Tucchhaa, line 71: Need this, so that default template would apply. this will be removed with the old appointments impl

sjbur (approved):

PR #33249, opened 2026-04-14
Scheduler - Appointments Refactoring - Rendering fixes

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/agenda_appointment.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/agenda_appointment.test.ts
@@ -110,25 +110,6 @@ describe('AgendaAppointment', () => {
       expect($element.css('width')).toBe('100%');
       expect($element.css('height')).toBe('50px');
     });
-
-    it('should apply new geometry when resize() is called', async () => {
-      const instance = await createAgendaAppointment({
-        ...getProperties(defaultAppointmentData),
-        geometry: {
-          width: '100%', height: 50,
-        },
-      });
-
-      instance.resize({
-        width: '50%',
-        height: 70,
-      });
-
-      const $element = instance.$element();
-
-      expect($element.css('width')).toBe('50%');
-      expect($element.css('height')).toBe('70px');
-    });
   });
 
   describe('Title', () => {
--- a/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/agenda_appointment.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/agenda_appointment.test.ts
@@ -7,7 +7,7 @@ import type { AppointmentResource } from '@ts/scheduler/utils/resource_manager/a
 
 import fx from '../../../common/core/animation/fx';
 import { getBaseAppointmentViewProperties } from '../__mock__/base_appointment_view';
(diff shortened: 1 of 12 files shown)


packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -45,10 +45,20 @@ export interface AppointmentsProperties extends DOMComponentProperties<Appointme
...
 }
 
-type AppointmentComponent = GridAppointmentView | AgendaAppointmentView | AppointmentCollector;
+type ViewItem = GridAppointmentView | AgendaAppointmentView | AppointmentCollector;
 
 export class Appointments extends DOMComponent<Appointments, AppointmentsProperties> {
-  private appointmentBySortIndex: Record<number, AppointmentComponent> = {};
+  private viewItemBySortedIndex: Record<number, ViewItem> = {};
+
+  private viewItems: ViewItem[] = [];
```
Tucchhaa, line 53: This variable contains the same value as viewItemBySortedIndex, but as a flat array.

It's not used in the current code, but will be used in the next PR with the focus state.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -194,54 +198,53 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
     // TODO: remove passing index to appointmentTemplate, need only to avoid BC
     viewModelDiff.forEach((diffItem, index) => {
       const { allDay, sortedIndex } = diffItem.item;
+      const lookupIndex = diffItem.oldSortedIndex ?? sortedIndex;
```
Tucchhaa, line 203: Without lookup index, this case wouldn't work:
```js
const InitialViewModel = [
   { data: 'A', sortedIndex: 0 },
   { data: 'B', sortedIndex: 1 }
]
renderViewModelDiff(InitialViewModel) ->
viewItemBySortedIndex == {
   0: 'A'
   1: 'B'
}

const nextViewModel = [
   { data: 'A', sortedIndex: 0 },
   { data: 'New', sortedIndex: 1 },
   { data: 'B', sortedIndex: 2 }
]
renderViewModelDiff(InitialViewModel) ->
viewItemBySortedIndex == {
   0: 'A'
   1: 'New',
   2: undefined
}
```

That's because get_view_model_diff doesn't compare `sortedIndex` between items. So since the 'B' item didn't have any changes, it wouldn't be rerendered, but the index to access the element from `this.viewItemBySortedIndex` will be 2, which is not defined.

packages/devextreme/js/__internal/scheduler/appointments_new/appointment/agenda_appointment.test.ts
```diff
@@ -85,6 +85,31 @@ describe('AgendaAppointment', () => {
...
+        ...getProperties(defaultAppointmentData),
+        getResourceColor: () => Promise.resolve('red'),
+      });
+
+      expect(instance.$element().hasClass(APPOINTMENT_TYPE_CLASSES.HAS_RESOURCE)).toBe(true);
+    });
+  });
+
+  describe('Geometry', () => {
+    it('should apply geometry on init', async () => {
```
Tucchhaa, line 100: In agenda view on viewModel change all view items are rerendered, so there's no test for .resize() function, as agenda appointment is resized only at the creation

aleksei-semikozov (approved):

PR #33256, opened 2026-04-14
Lookup: Normalize Unicode form before string comparison to support NFD input (T1326069)

r-farkhutdinov:


marker-dao (approved):

packages/devextreme/js/__internal/core/utils/m_data.ts
```diff
@@ -215,6 +215,7 @@ export const toComparable = function (value, caseSensitive?, options: any = {})
   const isCaseSensitive = caseSensitive || collatorSensitivity === 'case' || collatorSensitivity === 'variant';
 
   if (typeof value === 'string' && !isCaseSensitive) {
+    value = value.normalize('NFC');
```
marker-dao, line 218: You're mutating the input value. It's best to avoid doing that.

r-farkhutdinov: Core part is not refactored intentionally, it should be done separately. Here I just follow the existing pattern, which can be updated, I agree.

PR #33268, opened 2026-04-14
Form: fields should not exceed group column width (T1326645)

pharret31:


packages/devextreme/testing/tests/DevExpress.ui.widgets.form/form.markup.tests.js
```diff
@@ -791,6 +796,44 @@ QUnit.module('Form', () => {
...
+        });
+
+        const $group = $form.find(`.${FORM_GROUP_CLASS}`).first();
+        const $label = $group.find(`.${FIELD_ITEM_LABEL_CLASS}`).first();
+        const $content = $group.find(`.${FIELD_ITEM_CONTENT_CLASS}`).first();
+
+        const groupWidth = $group.outerWidth();
+        const totalItemWidth = $label.outerWidth() + $content.outerWidth();
+
+        assert.strictEqual(totalItemWidth <= groupWidth, true, `field item content width (${totalItemWidth}) should not exceed group width (${groupWidth})`);
```
EugeniyKiyashko: Let’s compare strict values instead of relying on a conditional check. Also, use jQuery methods to retrieve the element’s dimensions

pharret31: The jQuery method is used to get the width: https://api.jquery.com/outerWidth/ but since elements may have padding it's outerWidth instead of a usual width

EugeniyKiyashko (approved):

PR #33269, opened 2026-04-15
Scheduler - Appointments Refactoring - Focus state

Tucchhaa:

### DIFF
--- a/packages/devextreme-scss/scss/widgets/base/scheduler/appointment/regular/_index.scss
+++ b/packages/devextreme-scss/scss/widgets/base/scheduler/appointment/regular/_index.scss
@@ -94,6 +94,7 @@ $reduced-icon-offset: null !default;
   background-clip: padding-box;
   position: absolute;
   cursor: default;
+  outline: none;
 
   @include user-select(none);
 
--- a/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
@@ -1,0 +1,103 @@
+import $ from '@js/core/renderer';
+import type { DxEvent } from '@js/events';
+import type { KeyboardKeyDownEvent } from '@ts/events/core/m_keyboard_processor';
+
+import { getRawAppointmentGroupValues } from '../utils/resource_manager/appointment_groups_utils';
+import type { SortedEntity } from '../view_model/types';
+import type { Appointments } from './appointments';
+
+export class AppointmentsFocusController {
+  private needRestoreFocusIndex = -1;
+
+  private get sortedAppointments(): SortedEntity[] {
+    return this.appointments.option().getSortedAppointments();
+  }
+
+  private get isVirtualScrolling(): boolean {
+    return this.appointments.option().isVirtualScrolling();
+  }
+
+  constructor(private readonly appointments: Appointments) { }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public onAppointmentFocusIn(sortedIndex: number): void { }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public onAppointmentFocusOut(e: DxEvent, sortedIndex: number): void {
+    const focusEvent = e.originalEvent as FocusEvent;
+
+    const $relatedTarget = $(focusEvent.relatedTarget as Element);
+    const { $commonContainer, $allDayContainer } = this.appointments;
+
+    const isFocusOutside = $relatedTarget.length === 0 || (
+      $relatedTarget.closest($commonContainer).length === 0
+      && $relatedTarget?.closest($allDayContainer ?? $()).length === 0
+    );
+
+    if (isFocusOutside) {
+      this.resetTabIndex();
+    }
+  }
+
+  public onAppointmentKeyDown(e: KeyboardKeyDownEvent, sortedIndex: number): void {
+    if (e.key === 'Tab') {
+      this.handleTabKeyDown(e, sortedIndex);
+    }
+  }
+
+  public resetTabIndex(): void {
+    if (this.needRestoreFocusIndex >= 0) {
+      const appointmentView = this.appointments.getViewItemBySortedIndex(
+        this.needRestoreFocusIndex,
+      );
+      appointmentView?.focus();
+      this.needRestoreFocusIndex = -1;
+      return;
+    }
+
+    // TODO: in virtual scrolling no appointment may be rendered in the initial viewport
+    this.appointments.getViewItemByIndex(0)?.makeFocusable();
+  }
+
+  private handleTabKeyDown(e: KeyboardKeyDownEvent, sortedIndex: number): void {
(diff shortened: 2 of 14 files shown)


packages/devextreme-scss/scss/widgets/base/scheduler/appointment/regular/_index.scss
```diff
@@ -94,6 +94,7 @@ $reduced-icon-offset: null !default;
   background-clip: padding-box;
   position: absolute;
   cursor: default;
+  outline: none;
```
Tucchhaa, line 97: In old impl `outline: none` style was set in .dx-item class, which is set by CollectionWidget

packages/devextreme/js/__internal/scheduler/appointments_new/view_item.ts
```diff
@@ -0,0 +1,95 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  extends DOMComponentProperties<ViewItem<any>> {
+  tabIndex: number;
+  sortedIndex: number;
+  onFocusIn: (sortedIndex: number) => void;
+  onFocusOut: (e: DxEvent, sortedIndex: number) => void;
+  onKeyDown: (e: KeyboardKeyDownEvent, sortedIndex: number) => void;
+}
+
+export class ViewItem<
```
Tucchhaa, line 20: This class contains common functionality for both AppointmentView and AppointmentCollector: focus events and keydown event subscription and unsubscription

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -1072,6 +1073,11 @@ class Scheduler extends SchedulerOptionsBaseWidget {
         getResourceManager: () => this.resourceManager,
         getAppointmentDataSource: () => this.appointmentDataSource,
         getDataAccessor: () => this._dataAccessors,
+        getStartViewDate: () => this.getStartViewDate(),
+        getSortedAppointments: () => this._layoutManager.sortedItems,
+
+        isVirtualScrolling: () => this.isVirtualScrolling(),
```
Tucchhaa, line 1079: This props are used only by focus controller

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (this.isVirtualScrolling) {
+      this.scrollToItem(itemData);
+    }
+
+    const appointmentView = this.appointments.getViewItemBySortedIndex(itemData.sortedIndex);
+
+    if (appointmentView) {
+      appointmentView.focus();
+    } else if (this.isVirtualScrolling) {
+      this.needRestoreFocusIndex = itemData.sortedIndex;
```
Tucchhaa, line 105: On virtual scrolling need to restore focus index, because the item to which the scheduler will scroll may not be shown yet. The focus will be restored, when the rendering is finished:

https://github.com/DevExpress/DevExtreme/pull/33269/changes#diff-fe8d027bace8707fc88bc8c66b807f3d05219c21aff8edd3b187f70a08c1f641R265

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (this.needRestoreFocusIndex >= 0) {
+      const appointmentView = this.appointments.getViewItemBySortedIndex(
+        this.needRestoreFocusIndex,
+      );
+      appointmentView?.focus();
+      this.needRestoreFocusIndex = -1;
+      return;
+    }
+
+    // TODO: in virtual scrolling no appointment may be rendered in the initial viewport
```
Tucchhaa, line 77: To fix this we will need to modify the viewModel, so that the first appointment is always rendered

packages/devextreme/js/__internal/scheduler/appointments_new/view_item.ts
```diff
@@ -0,0 +1,94 @@
...
+  TProperties extends ViewItemProperties = ViewItemProperties,
+> extends DOMComponent<ViewItem<TProperties>, TProperties> {
+  private keyboardListenerId?: string;
+
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  override _getSynchronizableOptionsForCreateComponent():
```
Tucchhaa, line 26: This function is overriden only to fix ts error at GridAppointmentView, AgendaAppointmentView and AppointmentCollector creation.

This function should return names of options that should be synchronized with the parent component. E.g. original DOM Component [function](https://github.com/DevExpress/DevExtreme/blob/8749abb6bc51dcd11bd4db2f9c6c844760497ee5/packages/devextreme/js/__internal/core/widget/dom_component.ts#L108) returns only ['rtlEnabled', 'disabled', 'templatesRenderAsynchronously']. 

So nothing in this func is changed except from the return type.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -129,6 +142,13 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
         this.renderViewModel(this.option().viewModel);
         break;
       }
+      case 'tabIndex': {
+        this.viewItems.forEach((item) => {
+          item.setTabIndex(args.value);
+        });
+        this.focusController.resetTabIndex();
```
sjbur, line 146: I think there is a potential focus-management regression here. After tabIndex changes, the currently focusable item keeps the new tabindex through setTabIndex(), but resetTabIndex() always makes the first appointment focusable again. If focus was not on the first item before the update, this may leave two tabbable elements in the list at the same time. I would preserve the currently focused sortedIndex and restore that item

I wrote small test which fails on this case:

      it('should keep only focused appointment tabbable on tabIndex change', () => {
        const instance = createAppointments(getProperties());
        instance.option('viewModel', [
          mockGridViewModel({ ...defaultAppointmentData }, { sortedIndex: 0 }),
          mockGridViewModel({ ...defaultAppointmentData }, { sortedIndex: 1 }),
          mockAppointmentCollectorViewModel({ ...defaultAppointmentData }, { sortedIndex: 2 }),
        ]);

        instance.getViewItemByIndex(1)?.focus();

        instance.option('tabIndex', 2);

        expect(instance.getViewItemByIndex(0)?.$element().attr('tabindex')).toBe('-1');
        expect(instance.getViewItemByIndex(1)?.$element().attr('tabindex')).toBe('2');
        expect(instance.getViewItemByIndex(2)?.$element().attr('tabindex')).toBe('-1');
      });

sjbur (changes requested):

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: fixed and added this test

sjbur (approved):

aleksei-semikozov (approved):

PR #33285, opened 2026-04-15
d.ts - Replace [Smth]Options members of 'Function' type (`smthExpr`) to more specific to keep exact signature in generated wrappers

dmlvr:


packages/devextreme/js/ui/button_group.d.ts
```diff
@@ -19,6 +19,7 @@ import {
 } from '../events';
 
 import {
+    CollectionWidgetOptions,
```
mpreyskurantov: unused?

mpreyskurantov (approved):

EugeniyKiyashko (dismissed):

packages/devextreme/js/ui/tree_list.d.ts
```diff
@@ -967,19 +967,19 @@ export type dxTreeListOptions<TRowData = any, TKey = any> = Omit<GridBaseOptions
...
      */
-    hasItemsExpr?: string | Function;
+    hasItemsExpr?: string | ((item: TRowData) => TKey);
     /**
      * @docid
      * @default "items"
      * @public
      */
-    itemsExpr?: string | Function;
+    itemsExpr?: string | ((item: TRowData, value?: any) => TKey);
```
Tucchhaa: If itemsExpr is a function, then it accepts only one parameter, according to [this code](https://github.com/DevExpress/DevExtreme/blob/67c8004e6fe3f680d7f549f64493f883dffed4cb/packages/devextreme/js/__internal/grids/tree_list/data_source_adapter/m_data_source_adapter.ts#L236) (there are no other usages of itemsGetter).

Moreover, it should return an array of items, not the key.

So the type should be:
```ts
itemsExpr?: string | ((item: TRowData) => TRowData[]);
```

packages/devextreme/js/ui/tree_list.d.ts
```diff
@@ -1119,7 +1119,7 @@ export type dxTreeListOptions<TRowData = any, TKey = any> = Omit<GridBaseOptions
      * @default "parentId"
      * @public
      */
-    parentIdExpr?: string | Function;
+    parentIdExpr?: string | ((item: TRowData, value?: any) => TKey);
```
Tucchhaa, line 1122: if `parentIdExpr` is function, then it accepts only one parameter (check usages of `_parentIdGetter` in [this file](https://github.com/DevExpress/DevExtreme/blob/67c8004e6fe3f680d7f549f64493f883dffed4cb/packages/devextreme/js/__internal/grids/tree_list/data_source_adapter/m_data_source_adapter.ts#L51)).

So its type should be this:
```ts
parentIdExpr?: string | ((item: TRowData) => TKey);
```

packages/devextreme/js/ui/tree_list.d.ts
```diff
@@ -967,19 +967,19 @@ export type dxTreeListOptions<TRowData = any, TKey = any> = Omit<GridBaseOptions
      * @docid
      * @public
      */
-    hasItemsExpr?: string | Function;
+    hasItemsExpr?: string | ((item: TRowData) => TKey);
```
Tucchhaa: I am not sure if it should return TKey. TKey is a type of key, hasItemsExpr doesn't return a key, so its return type should be different. 

Judging by [the usage](https://github.com/DevExpress/DevExtreme/blob/67c8004e6fe3f680d7f549f64493f883dffed4cb/packages/devextreme/js/__internal/grids/tree_list/data_source_adapter/m_data_source_adapter.ts#L141), the type should be this:

```ts
hasItemsExpr?: string | ((item: TRowData) => boolean);
```

packages/devextreme/js/ui/tree_view.d.ts
```diff
@@ -242,13 +242,13 @@ export interface dxTreeViewOptions<TItem extends ItemLike = any, TKey = any>
...
      */
-    expandedExpr?: string | Function;
+    expandedExpr?: string | ((item: TItem) => TKey);
     /**
      * @docid
      * @default 'hasItems'
      * @public
      */
-    hasItemsExpr?: string | Function;
+    hasItemsExpr?: string | ((item: TItem) => TKey);
```
Tucchhaa: I am not sure about treeView impl, but I suspect it should be similar to treeLists hasItemsExpr:
https://github.com/DevExpress/DevExtreme/pull/33285/changes#r3111399000

Please take a look

packages/devextreme/js/ui/tree_view.d.ts
```diff
@@ -242,13 +242,13 @@ export interface dxTreeViewOptions<TItem extends ItemLike = any, TKey = any>
      * @hidden false
      * @public
      */
-    expandedExpr?: string | Function;
+    expandedExpr?: string | ((item: TItem) => TKey);
```
Tucchhaa: I suspect it should return boolean:
```ts
expandedExpr?: string | ((item: TItem) => boolean);
```

Can you check plz?

packages/devextreme/js/ui/tree_view.d.ts
```diff
@@ -335,7 +335,7 @@ export interface dxTreeViewOptions<TItem extends ItemLike = any, TKey = any>
      * @hidden false
      * @public
      */
-    parentIdExpr?: string | Function;
+    parentIdExpr?: string | ((item: TItem, value?: any) => TKey);
```
Tucchhaa: Are you sure it accepts two parameters?

packages/devextreme/js/ui/tree_list.d.ts
Tucchhaa: Please take a look at all other widget's d.ts files. I think they might have similar mistakes, e.g. here: https://github.com/DevExpress/DevExtreme/pull/33285/changes#diff-6698d8890476361148d70e40dd836b58e198b3e1f2932aaea589bb91d264102dR22

Tucchhaa (changes requested):

packages/devextreme/js/ui/tree_list.d.ts
dmlvr: [Here](https://github.com/DevExpress/DevExtreme/blob/67c8004e6fe3f680d7f549f64493f883dffed4cb/packages/devextreme/js/__internal/grids/tree_list/data_source_adapter/m_data_source_adapter.ts#L232)  `_parentIdSetter` called with two arguments (item, value).
In [createParentIdSetter](https://github.com/DevExpress/DevExtreme/blob/67c8004e6fe3f680d7f549f64493f883dffed4cb/packages/devextreme/js/__internal/grids/tree_list/data_source_adapter/m_data_source_adapter.ts#L95), when parentIdExpr is a function, it is returned as-is, so the same user callback is used as a setter too.

Therefore, I think `parentIdExpr` is effectively a dual-mode getter/setter callback in this adapter, and the two-argument signature is expected.

packages/devextreme/js/ui/tree_list.d.ts
dmlvr: Agree on the return type: hasItemsExpr should be boolean-like rather than TKey.
Also, I noticed this callback is used not only as a getter but also as a setter in the adapter [here](https://github.com/DevExpress/DevExtreme/blob/67c8004e6fe3f680d7f549f64493f883dffed4cb/packages/devextreme/js/__internal/grids/tree_list/data_source_adapter/m_data_source_adapter.ts#L119) and [here](https://github.com/DevExpress/DevExtreme/blob/67c8004e6fe3f680d7f549f64493f883dffed4cb/packages/devextreme/js/__internal/grids/tree_list/data_source_adapter/m_data_source_adapter.ts#L539), so I added the second argument to the function signature as well.

``` typescript
hasItemsExpr?: string | ((item: TRowData, value?: boolean) => boolean | undefined);
```

dmlvr: and about `disabledExpr`, agree, fixed returned type

packages/devextreme/js/ui/tree_view.d.ts
dmlvr: Checked it again and removed additional argument

packages/devextreme/js/ui/tree_list.d.ts
```diff
@@ -1119,7 +1119,7 @@ export type dxTreeListOptions<TRowData = any, TKey = any> = Omit<GridBaseOptions
      * @default "parentId"
      * @public
      */
-    parentIdExpr?: string | Function;
+    parentIdExpr?: string | ((item: TRowData, value?: any) => TKey | undefined);
```
Tucchhaa: it should be:
```ts
parentIdExpr?: string | ((item: TRowData, value?: TKey) => TKey | undefined);
```

dmlvr: updated it

Tucchhaa (approved):

EugeniyKiyashko (approved):

PR #33351, opened 2026-04-18
Scheduler - Create appointment by selected range

aleksei-semikozov:


packages/devextreme/js/__internal/scheduler/__tests__/selection_end_event.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const thirdCell = POM.getDateTableCell(2, 0);
+
+    fireEvent.mouseDown(firstCell, { which: 1 });
+    fireEvent.mouseMove(secondCell);
+    fireEvent.mouseMove(thirdCell);
+    fireEvent.mouseUp(thirdCell);
+
+    expect(onSelectionEnd).toHaveBeenCalledTimes(1);
+
+    const { selectedCellData, component } = onSelectionEnd.mock.calls[0][0];
```
sjbur, line 50: Typescript error here, please fix it:
Property 'component' does not exist on type 'unknown'. 
 Property 'selectedCellData' does not exist on type 'unknown'.

sjbur:
Is this a behavior that we expect:
1. Select cells — receive onSelectionEnd event
2. Click on selected cells — we receive onSelectionEnd event and appointment popup also shows up

Maybe we need to add some guard to avoid duplication of onSelectionEnd?

sjbur:
Question: do we want to support this behavior for keyboard navigation? We have keyboard navigation with arrow buttons(shift + arrow buttons), but I don't see the implemention supporting it

sjbur (changes requested):

packages/devextreme/js/__internal/scheduler/__tests__/selection_end_event.test.ts
aleksei-semikozov: fixed

aleksei-semikozov:
> Is this a behavior that we expect:
> 
> 1. Select cells — receive onSelectionEnd event
> 2. Click on selected cells — we receive onSelectionEnd event and appointment popup also shows up
> 
> Maybe we need to add some guard to avoid duplication of onSelectionEnd?

Great catch, thanks for the careful review — fixed and covered with a test

aleksei-semikozov:
> Question: do we want to support this behavior for keyboard navigation? We have keyboard navigation with arrow buttons(shift + arrow buttons), but I don't see the implemention supporting it

  There is no clean "end of keyboard selection" moment:

  - `Shift+Arrow` is incremental — each press is both the start and the end of an extension.
  - Firing on `Shift` `keyup` is fragile: it triggers even with no arrow press at all, and `keyup` can be missed on focus loss / IME / window switch.
  - We did not find an established pattern for this in comparable scheduler widgets.
  - 
  If someone really wants to react to "Shift released after a range" we have WA:
  ```js
  const scheduler = $('#scheduler').dxScheduler('instance');

  $(scheduler.element()).on('keyup', (e) => {
    if (e.key !== 'Shift') return;

    const cells = scheduler.option('selectedCellData');
    if (!cells || cells.length < 2) return;

    scheduler.showAppointmentPopup({
      startDate: cells[0].startDate,
      endDate: cells[cells.length - 1].endDate,
    }, true);
  });

  I'll file a follow-up issue to research proper keyboard support separately.

sjbur:
Found a bug in demo:
- Select cells in all day panel 
- Create appointment
- The appointment is not being created in all day panel

The reason for this is that allDay flag is not being passed to showAppointmentPopup

aleksei-semikozov:
> Found a bug in demo:
> 
> * Select cells in all day panel
> * Create appointment
> * The appointment is not being created in all day panel
> 
> The reason for this is that allDay flag is not being passed to showAppointmentPopup

fixed

sjbur (approved):

PR #33356, opened 2026-04-20
DateBox: Fix validation message not cleared when masked date returns to valid range (T1326750)

r-farkhutdinov:

### DIFF
--- a/packages/devextreme/js/__internal/ui/date_box/date_box.mask.ts
+++ b/packages/devextreme/js/__internal/ui/date_box/date_box.mask.ts
@@ -716,13 +716,16 @@ class DateBoxMask extends DateBoxBase {
     return this._maskValue?.getTime() !== value?.getTime();
   }
 
-  _fireChangeEvent(): void {
+  _hasEditorSpecificValidationError(): boolean {
     const { isValid, validationError } = this.option();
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/ui/date_box/date_box.mask.ts
```diff
@@ -717,9 +717,12 @@ class DateBoxMask extends DateBoxBase {
   }
 
   _fireChangeEvent(): void {
+    const { isValid, validationError } = this.option();
+    const isInvalidEditorSpecific = !isValid && validationError?.editorSpecific;
```
marker-dao: [minor]
I would move that into a separate method. Up to you.

marker-dao (dismissed):

packages/devextreme/js/__internal/ui/date_box/date_box.mask.ts
r-farkhutdinov: Agree, thank you. Changed here: https://github.com/DevExpress/DevExtreme/pull/33356/changes/c690b97f36dc6e5d501478f1cb924c1b1805d060

marker-dao (approved):

PR #33379, opened 2026-04-23
Update to NodeJS v24

dmlvr:


.node-version
```diff
@@ -0,0 +1 @@
+24.15.0
```
pharret31: don't forget to add eols

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "26.1.0",
   "license": "MIT",
   "author": "Developer Express Inc.",
+  "engines": { "node": ">=24" },
```
pharret31: ???
I would suggest use the syntax that renovate bot suggests in the documentation: node: ^24.15.0
https://docs.renovatebot.com/node/#configuring-which-version-of-npm-renovate-uses

UPD: node: 24.15.0, since the updation would be automatic by renovate and it won't update version with caret

dmlvr: We need to fix major version here, renovate bot should update `.node-version` file, based on our last discussion.

PR #33412, opened 2026-04-27
DataGrid - Pager - Show correct pageIndex when virtualScrolling is enabled on pageSize change

Tucchhaa:


packages/devextreme/js/__internal/grids/grid_core/pager/m_pager.ts
```diff
@@ -9,6 +9,10 @@ const PAGER_CLASS = 'pager';
 export const MAX_PAGES_COUNT = 10;
 
 const getPageIndex = function (dataController) {
```
Tucchhaa, line 11: This function is used to get visible index of the current page, that is passed to pager so it could display it.

If in pager the pageSize 'all' is selected, then dataController.pageSize() == 0, and so there's only one page.

packages/devextreme/js/__internal/grids/grid_core/pager/m_pager.ts
```diff
@@ -9,6 +9,10 @@ const PAGER_CLASS = 'pager';
...
+  if (dataController.pageSize() === 0) {
```
anna-shakhova: This lines may confuse later, I think it should be fixed on pageIndex level instead
like changePaging in data controller resets pageIndex to 0 in case of pageSize = 0, same should be here, also you may notice that after changing to pageSize = All grid shows 1st page (rows from 1), not keeps the prev one

anna-shakhova (changes requested):

packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/keyboardNavigation.realControllers.tests.js
```diff
@@ -671,7 +671,7 @@ QUnit.module('Real DataController and ColumnsController', {
         this.cellValue(0, 2, '5');
         this.saveEditData();
         // assert
-        assert.equal(focusedRowChangedFiresCount, 2, 'onFocusedRowChanged fires count');
```
Tucchhaa, line 674: Name of the test suggest that `focusedRowChangedFiresCount` should increase after refresh, but current test doesn't check it.

Initial test did it, though: [link](https://github.com/DevExpress/DevExtreme/blame/cf12bd8edc82b2de8b930fd052a8d254f1e26fb6/testing/tests/DevExpress.ui.widgets.dataGrid/keyboardNavigationParts/realControllers.tests.js)

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -43,6 +43,9 @@ const changePaging = function (that, optionName, value) {
 
         that._skipProcessingPagingChange = true;
         that.option(`paging.${optionName}`, value);
+        if (optionName === 'pageSize' && value === 0) {
+          that.option('paging.pageIndex', 0);
+        }
```
Tucchhaa, line 43: the main fix, this synchronizes pageIndex between dataSource and dataController when pageSize is changed to 'all' .

For other pageSize changes, then pageIndex is synchronized in this way:
1. [loadViewport](https://github.com/DevExpress/DevExtreme/blob/65b83c760ab6f0da7db759d24d11d4674d1b1fc5/packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts#L1134) ->
2. [loadItems](https://github.com/DevExpress/DevExtreme/blob/65b83c760ab6f0da7db759d24d11d4674d1b1fc5/packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts#L1107) ->
3. [_updateVisiblePageIndex](https://github.com/DevExpress/DevExtreme/blob/65b83c760ab6f0da7db759d24d11d4674d1b1fc5/packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts#L1089)

This flow didn't work with pageSize==0, because [_getChangedLoadParams()](https://github.com/DevExpress/DevExtreme/blob/65b83c760ab6f0da7db759d24d11d4674d1b1fc5/packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts#L1057) returns null when pageSize==0, because in [getLoadPageParams](https://github.com/DevExpress/DevExtreme/blob/65b83c760ab6f0da7db759d24d11d4674d1b1fc5/packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts#L981) pageSize==0 is handled separately from other pageSize values.

packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts
```diff
@@ -1004,18 +1004,19 @@ export const data = (Base: ModuleType<DataController>) => class VirtualScrolling
     };
   }
 
-  private _updateVisiblePageIndex(currentPageIndex?) {
+  private _updateVisiblePageIndex(value?: number): void {
```
Tucchhaa, line 1007: just code-style fixes

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -213,11 +216,7 @@ export class DataController extends DataHelperMixin(modules.Controller) {
 
     this._isPaging = false;
     this._currentOperationTypes = null;
-    this._dataChangedHandler = (e) => {
-      this._currentOperationTypes = this._dataSource.operationTypes();
-      this._handleDataChanged(e);
-      this._currentOperationTypes = null;
-    };
```
Tucchhaa, line 220: `_currentOperationTypes` was working incorrectly. It could be set to null before its value is applied to the next `change`, because of:
1. There could be several changes that were merged into one. (In this case [_fireChanged](https://github.com/DevExpress/DevExtreme/pull/33412/changes#diff-67b01513d9b30659c5aa3a93dbd3856e9524fbdb5e7f4c01df14cb96ece9033eL1276) wouldn't be called because of this [if-statement](https://github.com/DevExpress/DevExtreme/pull/33412/changes#diff-67b01513d9b30659c5aa3a93dbd3856e9524fbdb5e7f4c01df14cb96ece9033eL1254))
2. Race condition [here](https://github.com/DevExpress/DevExtreme/pull/33412/changes#diff-67b01513d9b30659c5aa3a93dbd3856e9524fbdb5e7f4c01df14cb96ece9033eL568)

P.S. this is related to the bug only indirectly. I noticed it while looking for the fix.  Changes related to _currentOperationTypes fixed the bug only in several cases, but not completely

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -1231,36 +1233,32 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     }
   }
 
-  public updateItems(change?, isDataChanged?) {
```
Tucchhaa, line 1234: code-style changes only

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -1231,36 +1238,32 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
-    change.isFirstRender = !that.changed.fired();
+  public updateItems(change: any = {}, isDataChanged?: boolean) {
+    change.isFirstRender = !this.changed.fired();
 
-    if (that._repaintChangesOnly !== undefined) {
-      change.repaintChangesOnly = change.repaintChangesOnly ?? that._repaintChangesOnly;
-      change.needUpdateDimensions = change.needUpdateDimensions || that._needUpdateDimensions;
+    if (this._repaintChangesOnly !== undefined) {
+      change.repaintChangesOnly ??= this._repaintChangesOnly;
+      change.needUpdateDimensions ??= this._needUpdateDimensions;
```
anna-shakhova: This is not the same as previous version, now we override `undefined` only, before we also update value for `false`

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -1231,36 +1238,32 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
-      const operationTypes = that.dataSource().operationTypes();
+      const operationTypes = this.dataSource().operationTypes();
 
-      change.repaintChangesOnly = operationTypes && !operationTypes.grouping && !operationTypes.filtering && that.option('repaintChangesOnly');
       change.isDataChanged = true;
-      if (operationTypes && (operationTypes.reload || operationTypes.paging || operationTypes.groupExpanding)) {
-        change.needUpdateDimensions = true;
-      }
+      change.repaintChangesOnly = operationTypes && !operationTypes.grouping && !operationTypes.filtering && this.option('repaintChangesOnly');
+      change.needUpdateDimensions = operationTypes && (operationTypes.reload || operationTypes.paging || operationTypes.groupExpanding);
```
anna-shakhova: Let's keep prev version, it was easier for reading, also we changes logic a bit: now it unconditionally assigns true/false/undefined whereas before it left a pre-existing value alone when the condition was false.

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -42,18 +42,26 @@ const changePaging = function (that, optionName, value) {
         dataSource[optionName](value);
 
         that._skipProcessingPagingChange = true;
+
+        if (optionName === 'pageSize') {
+          that.option('paging.pageIndex', 0);
```
anna-shakhova, line 42: Is there any reason to update paging.pageIndex not in the same moment as dataSource.pageIndex? If not, let's combine them in single condition for readability

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -42,18 +42,26 @@ const changePaging = function (that, optionName, value) {
         dataSource[optionName](value);
 
         that._skipProcessingPagingChange = true;
+
+        if (optionName === 'pageSize') {
```
anna-shakhova: Please, add here condition: `&& value === 0` to reset page index only for 'All' case

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
Tucchhaa: fixed

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
Tucchhaa: applied

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
Tucchhaa: applied

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -1170,6 +1167,8 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     const changeType = change.changeType || 'refresh';
 
     change.changeType = changeType;
+    change.operationTypes = this._currentOperationTypes;
+    this._currentOperationTypes = null;
```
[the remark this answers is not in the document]
Tucchhaa: applied

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
Tucchhaa: combined, but had to update one qunit test: https://github.com/DevExpress/DevExtreme/blob/cf17123a56ce8d1c50662789f965c0dc0c3a6ab8/packages/devextreme/testing/tests/DevExpress.ui.widgets.dataGrid/dataController.tests.js#L3227

Because it checked that pageIndex should be 0, but since we decided to keep the pageIndex between finite pageSize changes, the test is not actual anymore

packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts
```diff
@@ -1231,36 +1230,35 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
         change.needUpdateDimensions = true;
       }
     }
 
-    if (that._updateLockCount && !change.cancel) {
-      that._changes.push(change);
+    if (this._updateLockCount && !change.cancel) {
+      this._changes.push(change);
       return;
     }
```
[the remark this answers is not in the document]
Tucchhaa: no test is failing, I would skip this suggestion

anna-shakhova (dismissed):

anna-shakhova (approved):

PR #33416, opened 2026-04-27
Scheduler: T1327394 — fix appointments display for DST

sjbur:


packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/options/get_minutes_cell_intervals.ts
```diff
@@ -30,8 +31,12 @@ export const getMinutesCellIntervals = ({
 
   let columnIndex = 0;
   filterBySkippedDays(dayIntervals, skippedDays).forEach((dayInterval) => {
-    const date = new Date(dayInterval.min);
-    while (date.getTime() < dayInterval.max) {
+    const firstAvailableDayTime = timeZoneUtils
+      .adjustDayIntervalMinForMidnightDST(dayInterval.min, startDayHour);
+    const intervalMin = Math.max(firstAvailableDayTime, interval.min);
+    const intervalMax = Math.min(dayInterval.max, interval.max);
```
aleksei-semikozov: Why this change?

```ts
const intervalMin = Math.max(firstAvailableDayTime, interval.min);
const intervalMax = Math.min(dayInterval.max, interval.max);
```

splitIntervalByDay already produces dayIntervals inside interval bounds.
dayInterval.min >= interval.min, dayInterval.max <= interval.max — always.
So Math.max picks firstAvailableDayTime, Math.min picks dayInterval.max.
Both clamps are no-ops.

Looks not related to DST fix. Remove?

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -285,6 +297,7 @@ const utils = {
   cacheTimeZones,
 
   isLocalTimeMidnightDST,
+  adjustDayIntervalMinForMidnightDST,
```
aleksei-semikozov: Used only here:
`get_minutes_cell_intervals.ts:35`

Maybe keep helper local there? It is not really a generic time-zone utility.

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
```diff
@@ -258,6 +258,18 @@ const isLocalTimeMidnightDST = (date: Date): boolean => {
...
 
+const adjustDayIntervalMinForMidnightDST = (
+  dayIntervalMin: number,
+  startDayHour: number,
+): number => {
+  const date = createDateFromUTCWithLocalOffset(new Date(dayIntervalMin));
+  const isMidnightDST = startDayHour === 0 && isLocalTimeMidnightDST(date);
+
+  return isMidnightDST
+    ? dayIntervalMin + date.getHours() * 60 * 60 * 1000
```
aleksei-semikozov: `toMs('hour')` already exists in this file (line 23, used at 55, 168, 174).

```ts
return isMidnightDST
  ? dayIntervalMin + date.getHours() * toMs('hour')
  : dayIntervalMin;
```

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
sjbur: Makes sense. Applied

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/options/get_minutes_cell_intervals.ts
sjbur: Yes, we can remove this. Applied

packages/devextreme/js/__internal/scheduler/m_utils_time_zone.ts
sjbur: OK, will use toMs('hour'). Applied

aleksei-semikozov (approved):

PR #33457, opened 2026-04-29
GridCommands: implement paging, filter, group commands

anna-shakhova:


dmirgaev:
I can't write a comment or start a review for some reason, therefore, write it right here:
> Let's move the `searching` command and corresponding tests into separate files

dmirgaev (approved):

PR #33462, opened 2026-04-30
DataGrid - ✨AI: AI assistant - AIChat: Implement the disabled state

Alyar666:


packages/devextreme-scss/scss/widgets/base/gridBase/layout/aiChat/_index.scss
```diff
@@ -66,3 +66,14 @@
   bottom: 0;
   line-height: 0;
 }
+
+.dx-ai-chat--disabled {
+  .dx-chat-messagebox, .dx-ai-chat__message-regenerate-button {
+    opacity: 0.5;
```
anna-shakhova: we have $base-disabled-opacity variable in all themes, that differs from 0.5 in fluent and material, perhaps better to use it

also we have disabled-widget mixin, perhaps, we can use it here, as messagebox is private component

packages/devextreme/js/__internal/grids/grid_core/ai_chat/const.ts
```diff
@@ -38,6 +38,10 @@ export const CLASSES = {
   messageHeaderRow: 'dx-ai-chat__message-header-row',
   messageRegenerateButton: 'dx-ai-chat__message-regenerate-button',
   messageProgressBar: 'dx-ai-chat__message-progressbar',
+  clearChatButton: 'dx-ai-chat__clear-button',
+  disabled: 'dx-ai-chat--disabled',
+  textArea: 'dx-textarea',
```
anna-shakhova: Let's import those constants from m_text_area and speech_to_text files. It is unlikely we will change their values, but this way it'll be clearer

anna-shakhova (approved):

PR #33524, opened 2026-05-07
Security: Update Node version

dmlvr:


pharret31 (dismissed):

pharret31 (dismissed):

.mise.toml
```diff
@@ -0,0 +1,6 @@
+[tools]
+node = "24.15.0"
```
pharret31, line 2: I think a pnpm version should be here as well since this project doesn't work with pnpm != 9.15.9

dmlvr: agree, added it

alexslavr (approved):

PR #33531, opened 2026-05-08
dxDataGrid - AI Assistant: Implement partial re-rendering of messages when their status is updated

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/ai_chat/ai_chat.ts
```diff
@@ -139,25 +147,17 @@ export class AIChat {
...
     message: Message,
   ): void {
     const $row = $('<div>')
       .addClass(CLASSES.messageHeaderRow)
       .appendTo($parent);
-    const headerText = this.getHeaderText(message);
+    // NOTE: dxChat does not support partial re-render for custom message fields.
+    // We use the built-in `text` field as the message header because dxChat
+    // triggers partial re-render when `text` changes via DataSource push updates.
+    const headerText = message.text ?? '';
```
anna-shakhova: we provide `customizeResponseTitle` callback in our api, there might be the case when customer sets same title for different statuses
suggestion: pass status to `text`, we can duplicate it in status field as well for clarity and add additional `title` or `headerText` field for message header, then chat will rerender on status change even if header was not changed

anna-shakhova (approved):

dmirgaev (approved):

PR #33537, opened 2026-05-11
DataGrid - AI Assistant: Implement message cleanup

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/ai_assistant/__tests__/ai_assistant_view.test.ts
```diff
@@ -281,6 +281,20 @@ describe('AIAssistantView', () => {
   });
 
   describe('chat event handlers', () => {
+    describe('onChatCleared', () => {
```
dmirgaev, line 284: Do we have a test which checks whether the button is disabled while commands are executing?

dmirgaev (approved):

PR #33565, opened 2026-05-13
T1328518 - Scrollable - Simulated scrolling breaks after an unhandled JavaScript exception

dmlvr:

### DIFF
--- a/packages/devextreme/js/__internal/ui/scroll_view/scrollable.simulated.ts
+++ b/packages/devextreme/js/__internal/ui/scroll_view/scrollable.simulated.ts
@@ -19,7 +19,6 @@ import {
   // @ts-expect-error ts-error
   deferUpdater,
 } from '@js/core/utils/common';
-import { logger } from '@js/core/utils/console';
 import type { DeferredObj } from '@js/core/utils/deferred';
 import { Deferred, when } from '@js/core/utils/deferred';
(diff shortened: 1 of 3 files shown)


packages/devextreme/js/__internal/ui/scroll_view/scrollable.simulated.ts
```diff
@@ -1094,7 +1095,12 @@ export class SimulatedStrategy<
     const actionHandler = this._createActionByOption(optionName);
 
     return (...args: unknown[]) => {
-      actionHandler(extend(this._createActionArgs(), args));
+      try {
+        actionHandler(extend(this._createActionArgs(), args));
+      } catch (e) {
+        const error = e instanceof Error ? e : new Error(String(e));
+        errors.log('E1069', optionName, error.message);
+      }
```
pharret31, line 1102: Did you check other components? It seems like a common action handler issue and looks like it could produce same issue in other components

dmlvr: Yes, I checked. The problem can be reproduced in the core components as well, but it doesn't lead to behavioral breakage, so we decided to fix it locally inside scrollable

EugeniyKiyashko (approved):

PR #33575, opened 2026-05-13
Scheduler: refactor appointment_popup module (TS)

bit-byte0:
### What does the PR change?
 
- Removed `m_` prefix from files in `appointment_popup/` folder.
- Also added types and replaced old `Deferred` with native `Promise`.
 
 ### How did you achieve this?
 
 Removing from files _m prefix turned strict mode on, and strict mode complained about missing types and the `Deferred` usage (it needed `@ts-expect-error` to even compile). Fixed all that: added interfaces, switched to `async/await` + native `Promise`. 

One more thing - some tests broke after switching to `Promise` because they expected the save to finish synchronously (which Deferred kinda allowed). Native `Promise` always resolves async, so added `await Promise.resolve()` after save calls in those tests to let the chain settle before asserting.
 
 ### How can we verify these changes?
 
 - Run Jest: `pnpm run test-jest`
 - Run QUnit scheduler tests
 - Open demo app, try creating/editing/deleting regular and recurring appointments - should work same as before


packages/devextreme/js/__internal/scheduler/appointment_popup/popup.ts
```diff
@@ -75,28 +124,36 @@ export class AppointmentPopup {
     const popupConfig = this.createPopupConfig();
     this.createPopup(popupConfig);
 
-    this._popup!.show();
+    if (this.popupInstance) {
+      this.popupInstance.show().catch(noop);
+    } else {
+      throw errors.Error('E1033');
```
sjbur, line 130: E1033: 'Unknown step in the date navigator: \'{0}\'',

I am not sure that we need to display this error here. And not sure if we need to return any errors here or in hide step

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -1141,7 +1142,7 @@ class Scheduler extends SchedulerOptionsBaseWidget {
       getEditingConfig: () => this.editing,
       getResourceManager: () => this.resourceManager,
 
-      getFirstDayOfWeek: () => this.option('firstDayOfWeek'),
+      getFirstDayOfWeek: () => this.getFirstDayOfWeek(),
```
sjbur: @aleksei-semikozov I think you had a bug related to this string when you were also passing a function instead of option.

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/allDayAppointments.common-0.tests.js
```diff
@@ -568,6 +568,7 @@ module('All day appointments common', config, () => {
 
                 scheduler.instance.showAppointmentPopup(newItem, true);
                 $('.dx-scheduler-appointment-popup .dx-popup-done').trigger('dxclick');
+                await Promise.resolve();
```
sjbur, line 571: Why do we need this in tests?

sjbur (dismissed):

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/allDayAppointments.common-0.tests.js
bit-byte0: because `Deferred` that was used before refactoring were acting synchronously even that it is `Promise` analogue and all tests also were set up to check synchronous behavior. After refactoring we used `Promise` that acts asynchronous and tests were not ready for it - they didn't have any option to wait a bit (save button clicks before Promise resolves) and that's why I needed to add async behavior to tests

packages/devextreme/js/__internal/scheduler/appointment_popup/popup.ts
aleksei-semikozov: https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/Errors_and_Warnings/#E1033
```
A [Scheduler](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/) internal error.

To solve the issue, please submit a ticket to our [Support Center](https://www.devexpress.com/ask). Include your UI component configuration, fake data, and the steps needed to reproduce the issue in the ticket.
```

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
aleksei-semikozov: In this function, we use the `firstDayOfWeek` option in this way, but if this otpion is not specified, the localization settings are applied. This is the same way we use it in the header

[code link](https://github.com/bit-byte0/DevExtreme/blob/531f0e197b7157c77584b9bef585b21065d12bc8/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L2284)

packages/devextreme/testing/tests/DevExpress.ui.widgets.scheduler/allDayAppointments.common-0.tests.js
sjbur: OK, thank you

packages/devextreme/js/__internal/scheduler/appointment_popup/popup.ts
sjbur: OK, looks like a mistake of mine

aleksei-semikozov (dismissed):

sjbur (dismissed):

aleksei-semikozov (approved):

PR #33577, opened 2026-05-14
Grid dts: add customizeResponseText, customizeResponseTitle

anna-shakhova:
add `customizeResponseText`, `customizeResponseTitle` to dts


packages/devextreme/js/common/grids.d.ts
```diff
@@ -80,14 +80,53 @@ import {
...
+    dataField: string;
+    sortOrder: SortOrder | 'none';
+  };
+  clearSorting: {};
+}
+
+export type CommandInfo<
+  TCommands extends PredefinedCommands = PredefinedCommands,
+> = { [K in keyof TCommands]: { name: K; args: TCommands[K] } }[keyof TCommands]
+    | { name: string; args: Record<string, unknown> };
```
anna-shakhova: hack allows to test type in demos without internal code adjustment, remove before merge (we support only predefined commands for now)

packages/devextreme/js/__internal/grids/grid_core/ai_assistant/__tests__/utils.test.ts
```diff
@@ -1,9 +1,7 @@
-import {
-  describe, expect, it,
-} from '@jest/globals';
+import { describe, expect, it } from '@jest/globals';
 import type { Message } from '@js/ui/chat';
 
-import { AI_ASSISTANT_AUTHOR_ID, MessageStatus } from '../const';
```
Raushen: Question: Is MessageStatus enum user anywhere? Or it can be removed from ../const?

anna-shakhova: It is used in AiAssistantController and AIChat and it differs from ResponseStatus, so it should be here

Raushen (approved):

PR #33587, opened 2026-05-14
DataGrid - AI Assistant: Disable chat suggestions in disabled state

Alyar666:


anna-shakhova (dismissed):

packages/devextreme/js/__internal/grids/grid_core/ai_chat/const.ts
```diff
@@ -48,6 +48,7 @@ export const CLASSES = {
   disabled: 'dx-ai-chat--disabled',
   textArea: TEXTAREA_CLASS,
   speechToTextButton: SPEECH_TO_TEXT_CLASS,
+  chatSuggestions: 'dx-chat-suggestions',
```
Raushen: It seems we do not need this class anymore

Raushen (approved):

PR #33622, opened 2026-05-18
AI Assistant: improve utils test coverage

dmirgaev:


packages/devextreme/js/__internal/grids/grid_core/ai_assistant/utils.ts
Raushen, line 207: How about make tests instead of comment for other util methods also?

dmirgaev: There are all the necessary tests, i'd prefer keep this comment untouched, as it significantly simplify the code clearness

Raushen (approved):

PR #33633, opened 2026-05-19
Scheduler - Appointments Refactoring - Support click & double click

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
@@ -1,20 +1,24 @@
 import messageLocalization from '@js/common/core/localization/message';
 import registerComponent from '@js/core/component_registrator';
-import type { DxElement } from '@js/core/element';
 import type { dxElementWrapper } from '@js/core/renderer';
 import $ from '@js/core/renderer';
 import type { DxEvent } from '@js/events';
+import eventsEngine from '@js/events/core/events_engine';
+import { addNamespace } from '@js/events/utils';
+import type { AppointmentRenderedEvent } from '@js/ui/scheduler';
 import { getPublicElement } from '@ts/core/m_element';
 import { EmptyTemplate } from '@ts/core/templates/m_empty_template';
 import { FunctionTemplate } from '@ts/core/templates/m_function_template';
 import type { TemplateBase } from '@ts/core/templates/m_template_base';
-import { click } from '@ts/events/m_short';
+import { dxClick } from '@ts/events/m_short';
 import type { SafeAppointment, TargetedAppointment } from '@ts/scheduler/types';
 import type { AppointmentDataAccessor } from '@ts/scheduler/utils/data_accessor/appointment_data_accessor';
 
 import { APPOINTMENT_CLASSES, APPOINTMENT_TYPE_CLASSES, FOCUSED_STATE_CLASS } from '../const';
 import { DateFormatType, getDateTextFromTargetAppointment } from '../utils/get_date_text';
 import { EVENTS_NAMESPACE, ViewItem, type ViewItemProperties } from '../view_item';
+
+const DOUBLE_CLICK_EVENT_NAME = addNamespace('dxdblclick', EVENTS_NAMESPACE.namespace);
 
 export interface BaseAppointmentViewProperties
   extends ViewItemProperties {
@@ -23,11 +27,9 @@ export interface BaseAppointmentViewProperties
   targetedAppointmentData: TargetedAppointment;
   appointmentTemplate: TemplateBase;
 
-  onRendered: (e: {
-    element: DxElement;
-    appointmentData: SafeAppointment;
-    targetedAppointmentData: TargetedAppointment;
-  }) => void;
+  onRendered: (e: AppointmentRenderedEvent) => void;
+  onClick: (appointmentView: BaseAppointmentView, event: DxEvent) => void;
+  onDblClick: (appointmentView: BaseAppointmentView, event: DxEvent) => void;
 
   getDataAccessor: () => AppointmentDataAccessor;
   getResourceColor: () => Promise<string | undefined>;
@@ -36,15 +38,26 @@ export interface BaseAppointmentViewProperties
 export class BaseAppointmentView<
   TProperties extends BaseAppointmentViewProperties = BaseAppointmentViewProperties,
 > extends ViewItem<TProperties> {
-  protected get targetedAppointmentData(): TargetedAppointment {
+  get targetedAppointmentData(): TargetedAppointment {
     return this.option().targetedAppointmentData;
   }
 
-  protected get appointmentData(): SafeAppointment {
+  get appointmentData(): SafeAppointment {
     return this.option().appointmentData;
   }
 
   private defaultAppointmentTemplate!: FunctionTemplate;
+
+  override _setOptionsByReference(): void {
+    super._setOptionsByReference();
+
+    // Note: appointmentData object is used as a key in dataSource
+    this._optionsByReference = {
+      ...this._optionsByReference,
+      appointmentData: true,
+      targetedAppointmentData: true,
+    };
+  }
 
   override _init(): void {
     super._init();
@@ -62,6 +75,7 @@ export class BaseAppointmentView<
     this.applyAria();
     this.attachFocusEvents();
     this.attachClickEvent();
+    this.attachDblClickEvent();
     this.attachKeydownEvents();
     this.renderContentTemplate();
   }
@@ -69,7 +83,8 @@ export class BaseAppointmentView<
   override _dispose(): void {
     super._dispose();
 
-    click.off(this.$element(), EVENTS_NAMESPACE);
+    dxClick.off(this.$element(), EVENTS_NAMESPACE);
+    eventsEngine.off(this.$element(), DOUBLE_CLICK_EVENT_NAME);
   }
 
   protected applyElementClasses(): void {
@@ -86,11 +101,20 @@ export class BaseAppointmentView<
   }
 
   private attachClickEvent(): void {
-    click.off(this.$element(), EVENTS_NAMESPACE);
-    click.on(
+    dxClick.off(this.$element(), EVENTS_NAMESPACE);
+    dxClick.on(
       this.$element(),
-      this.onClick.bind(this),
+      (event: DxEvent<MouseEvent>) => this.option().onClick(this, event),
       EVENTS_NAMESPACE,
+    );
+  }
+
+  private attachDblClickEvent(): void {
+    eventsEngine.off(this.$element(), DOUBLE_CLICK_EVENT_NAME);
+    eventsEngine.on(
+      this.$element(),
+      DOUBLE_CLICK_EVENT_NAME,
+      (event: DxEvent<MouseEvent>) => this.option().onDblClick(this, event),
     );
   }
 
@@ -163,8 +187,9 @@ export class BaseAppointmentView<
       },
       index: this.option().index,
       onRendered: () => {
+        // @ts-expect-error 'component' and 'element' are set by action
         this.option().onRendered({
-          element: getPublicElement(this.$element()),
+          appointmentElement: getPublicElement(this.$element()),
           appointmentData: this.appointmentData,
           targetedAppointmentData: this.targetedAppointmentData,
         });
--- a/packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
@@ -9,7 +9,7 @@ import type { ClickEvent as ButtonClickEvent } from '@js/ui/button';
 import Button from '@js/ui/button';
 import { FunctionTemplate } from '@ts/core/templates/m_function_template';
 import type { TemplateBase } from '@ts/core/templates/m_template_base';
-import type { SafeAppointment, TargetedAppointment } from '@ts/scheduler/types';
+import type { TargetedAppointment } from '@ts/scheduler/types';
 
 import type { AppointmentItemViewModel } from '../view_model/types';
 import { APPOINTMENT_COLLECTOR_CLASSES } from './const';
@@ -33,8 +33,6 @@ export interface AppointmentCollectorProperties
 
 export class AppointmentCollector
   extends ViewItem<AppointmentCollectorProperties> {
-  private appointmentsData!: SafeAppointment[];
-
   private defaultAppointmentCollectorTemplate!: FunctionTemplate;
 
   private buttonInstance?: Button;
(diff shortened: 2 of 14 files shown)


packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
```diff
@@ -62,14 +75,16 @@ export class BaseAppointmentView<
...
+    this.attachDblClickEvent();
     this.attachKeydownEvents();
     this.renderContentTemplate();
   }
 
   override _dispose(): void {
     super._dispose();
 
-    click.off(this.$element(), EVENTS_NAMESPACE);
+    dxClick.off(this.$element(), EVENTS_NAMESPACE);
```
Tucchhaa, line 86: old impl also uses dxClick event

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
```diff
@@ -43,6 +62,8 @@ export class AppointmentCollector
     this.defaultAppointmentCollectorTemplate = new FunctionTemplate((options) => {
       this.defaultAppointmentCollectorContent($(options.container));
     });
+
+    this.appointmentsData = this.option().items.map((item) => item.itemData);
```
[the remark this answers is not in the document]
Tucchhaa: no need to support it. On viewModel update the component is recreated instead of being updated

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -378,6 +425,93 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
...
+    }));
+
+    this.option().showAppointmentTooltipCore(
+      collector.$element(),
+      collectorTooltipItems,
+      {
+        dragBehavior: undefined, // TODO
+        isButtonClick: true,
+        tabFocusLoopEnabled: true,
+      },
```
Tucchhaa, line 512: old implementation also passed `clickEvent`property ([here](https://github.com/DevExpress/DevExtreme/blob/21f38fa54bdde5f7e2a96b2035beff31c6936a04/packages/devextreme/js/__internal/scheduler/m_compact_appointments_helper.ts#L60)).

`clickEvent` is used in tooltip to call `onAppointmentClick` handler when tooltip item is clicked.

In new implementation, `onAppointmentClick` is directly used in the tooltip.

Tucchhaa: P.S. when onAppointmentClick is called on the tooltip item click, setting `e.cancel=true` has no effect. 

I think we might need to discuss it with PMs. I would suggest to make `e.cancel` prevent appointment popup from showing if appointment is inside the tooltip. However in this case, we would need to let user know if the clicked appointment is inside tooltip or not.

I would do that after the refactoring is done, because it's not urgent

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -378,6 +417,93 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
...
+      if (isElementInDom($target) && !this.preventSingleAppointmentClick) {
+        this.option().showAppointmentTooltip(
+          appointmentView.appointmentData,
+          $target,
+          appointmentView.targetedAppointmentData,
+        );
+      }
+
+      this.preventSingleAppointmentClick = false;
+    }, SHOW_TOOLTIP_TIMEOUT);
```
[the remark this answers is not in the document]
Tucchhaa: done

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -378,6 +425,93 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
...
```
aleksei-semikozov, line 463: Race condition: two clicks within 300ms → both timers fire, first tooltip flashes before being redrawn for the second.

Browser repro: appointments at 8:00 and 10:00, click A then B within 300ms — `onAppointmentTooltipShowing` fires twice.

Failing test for `appointments_new.test.ts`:

```ts
it('REPRO race: no tooltip for first appointment after clicking another within 300ms', async () => {
  const { POM } = await createScheduler({
    dataSource: [
      { text: 'A', startDate: new Date(2015, 1, 9, 8),  endDate: new Date(2015, 1, 9, 9) },
      { text: 'B', startDate: new Date(2015, 1, 9, 10), endDate: new Date(2015, 1, 9, 11) },
    ],
    currentView: 'day',
    currentDate: new Date(2015, 1, 9, 8),
  });

  const a = POM.getAppointments().find((x) => x.getText() === 'A')!.element!;
  const b = POM.getAppointments().find((x) => x.getText() === 'B')!.element!;

  jest.useFakeTimers();
  a.click();
  jest.advanceTimersByTime(100);
  b.click();
  jest.advanceTimersByTime(200);
  const after = POM.tooltip.isVisible() ? POM.tooltip.getAppointmentItem(0).textContent : null;
  jest.advanceTimersByTime(200);

  expect(POM.tooltip.getAppointmentItem(0).textContent).toContain('B');
  expect(after).toBeNull(); // FAILS: received "A8:00 AM - 9:00 AM"
});
```

Need to clear the previous timer before scheduling a new one.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -378,6 +425,93 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
...
+      event,
+    };
+
+    // @ts-expect-error 'component' and 'element' are set by action
+    this.option().onAppointmentClick(e);
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    if ((e as any).cancel) {
+      return;
+    }
```
aleksei-semikozov: If `preventSingleAppointmentClick` was set by a prior dblclick and the next click is cancelled, this early `return` skips the `setTimeout` that resets the flag. The next normal click then hits `preventSingleAppointmentClick === true` and silently skips `showAppointmentTooltip`.

Can't repro in browser — race above masks it (a leftover timer eventually resets the flag). Reproduces in jest:

```
Expected: true (tooltip should be visible after the normal click)
Received: false
```

Once race is fixed, this surfaces. Can you double-check? Looks like cancel branch should clear pending timer + reset the flag.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -44,20 +55,42 @@ export interface AppointmentsProperties extends DOMComponentProperties<Appointme
...
+    targetedAppointmentData: TargetedAppointment,
+  ) => void;
 }
 
 export class Appointments extends DOMComponent<Appointments, AppointmentsProperties> {
   private focusController!: AppointmentsFocusController;
 
+  private appointmentClickTimeout: number | null = null;
+
+  private preventSingleAppointmentClick = false;
```
aleksei-semikozov: Timer + flag on `Appointments` feel out of place. Debounce-before-tooltip is tooltip behavior, not collection state. Could discuss separately?

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
```diff
@@ -25,16 +28,32 @@ export interface AppointmentCollectorProperties
...
+  override _setOptionsByReference(): void {
+    super._setOptionsByReference();
+
+    // Note: items have appointmentData, which is used as a key in dataSource
+    this._optionsByReference = {
+      ...this._optionsByReference,
+      // @ts-expect-error
+      items: true,
+      // @ts-expect-error
+      targetedAppointmentData: true,
```
aleksei-semikozov: every `@ts-expect-error` — should have comments

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -378,6 +425,93 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
...
+      appointmentData: appointmentView.appointmentData,
+      targetedAppointmentData: appointmentView.targetedAppointmentData,
+      event,
+    };
+
+    // @ts-expect-error 'component' and 'element' are set by action
+    this.option().onAppointmentClick(e);
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    if ((e as any).cancel) {
```
aleksei-semikozov: **`(e as any).cancel` can be replaced with `(e as Cancelable).cancel`.**

`Cancelable` is exported from `@js/common/core/events`:

```ts
import type { Cancelable } from '@js/common/core/events';

// ...
if ((e as Cancelable).cancel) {
  return;
}
```

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -44,20 +55,42 @@ export interface AppointmentsProperties extends DOMComponentProperties<Appointme
...
+    appointment: SafeAppointment,
+    $element: dxElementWrapper,
+    targetedAppointment?: SafeAppointment,
+  ) => void;
+  showAppointmentTooltipCore: (
+    target: dxElementWrapper,
+    data: AppointmentTooltipItem[],
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    options?: any,
+  ) => void;
```
aleksei-semikozov: `AppointmentTooltipExtraOptions` is the right type here.

It's already defined in `tooltip_strategies/tooltip_strategy_base.ts:70`

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
```diff
@@ -25,16 +28,32 @@ export interface AppointmentCollectorProperties
...
   private get appointmentsCount(): number {
-    return this.option().appointmentsData.length;
+    return this.option().items.length;
+  }
+
+  override _setOptionsByReference(): void {
+    super._setOptionsByReference();
+
+    // Note: items have appointmentData, which is used as a key in dataSource
+    this._optionsByReference = {
```
aleksei-semikozov, line 48: Why do we need the _setOptionsByReference() method?
Can you explain?

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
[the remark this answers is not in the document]
aleksei-semikozov: May be drop this cash? 

```ts
// line 139
items: this.option().items.map((item) => item.itemData),
```

packages/devextreme/js/__internal/scheduler/tooltip_strategies/tooltip_strategy_base.ts
```diff
@@ -324,7 +326,21 @@ export abstract class TooltipStrategyBase {
...
+          appointmentElement: e.itemElement,
+          appointmentData: e.itemData.appointment,
+          targetedAppointmentData: e.itemData.targetedAppointment,
+          event: e.event,
+        });
+      }
+    } else {
+      this.extraOptions?.clickEvent?.(e);
+    }
+
```
bit-byte0, line 343: `onAppointmentClick` fires here only when `isButtonClick: true` (i.e. tooltip opened via collector). When tooltip is opened via direct appointment click does clicking an item inside it also fire `onAppointmentClick`? If not, is that intentional?

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: applied

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: Applied

packages/devextreme/js/__internal/scheduler/tooltip_strategies/tooltip_strategy_base.ts
Tucchhaa: Legacy implementation doesn't fire onAppointmentClick when tooltip is opened on appointment click, so to avoid any breaking changes I have made the behavior similar in the refactoring.

But in general, I think we need to rethink how onAppointmentClick works due to the reasons I have described in [this comment](https://github.com/DevExpress/DevExtreme/pull/33633#discussion_r3266339632).

Thanks for noticing, I will take it into account after the appointments refactoring is done

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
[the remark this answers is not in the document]
Tucchhaa: good suggestion, I have applied it

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: I noticed that the old impl has the same problem. But I have applied fix for the new impl and added a unit test

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: fixed and added a unit test

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
Tucchhaa: I have removed the flag, now it's only a timer. I am not sure if we need to move this logic to the tooltip, cuz it's logic may become complicated. 

Right now tooltip is "dumb": it doesn't know about appointments and clicks, which I think is good

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
Tucchhaa: added comments

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
Tucchhaa: Because to access appointment in data source (e.g. to delete or update it), the scheduler uses appointmentData object reference as a key in data source.

This PR adds functionality to show appointment popup, which gets appointmentData as its parameter. When appointment popup saves data it uses the same appointmentData object to update data in the data source.

OptionManager (aka .option()) deep-copies all the options passed to the component. So in order to not lose the object reference, we need to pass appointmentData as a reference for both appointmentView and collector.

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.test.ts
```diff
@@ -966,4 +985,217 @@ describe('Appointments', () => {
...
+      instance.option('viewModel', [
+        mockAppointmentCollectorViewModel(defaultAppointmentData, { sortedIndex: 0 }),
+      ]);
+
+      const viewItem = instance.getViewItemBySortedIndex(0);
+      const element = viewItem?.$element().get(0) as HTMLElement;
+
+      element.click();
+
+      expect(onAppointmentClick).not.toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
Tucchhaa: applied

packages/devextreme/js/__internal/scheduler/appointments_new/appointment_collector.ts
Tucchhaa: This method will be removed when usage of DOMComponent will be removed from AppointmentView and Collector

aleksei-semikozov (approved):

PR #33634, opened 2026-05-19
Scheduler - Appointments Refactoring - Support click & double click

Tucchhaa:


packages/devextreme/js/__internal/scheduler/tooltip_strategies/tooltip_strategy_base.ts
```diff
@@ -73,7 +75,8 @@ interface AppointmentTooltipExtraOptions {
   isButtonClick?: boolean;
   offset?: unknown;
   rtlEnabled?: boolean;
-  tabFocusLoopEnabled?: boolean;
```
Tucchhaa, line 76: `tabFocusLoopEnabled` was published only in 26.1.
For 25.2 it wasn't published, but during the tooltip refactoring, it was accidentally used here/

So I replaced usage of `tabFocusLoopEnabled` back to `_loopFocus`

aleksei-semikozov (approved):

PR #33638, opened 2026-05-19
TagBox: Keeping click order when maxDisplayTag enabled and showMultiTagOnly disabled (T1328498)

vorobey:


packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1113,6 +1116,39 @@ class TagBox<
     return selectedItems;
   }
 
+  _shouldUseClickOrderForTags(values): boolean {
+    return !this.option('showMultiTagOnly')
```
r-farkhutdinov: Replace all `this.option('value')` calls to be `const { value } = this.option();`

This is required for correct types of values as `this.option('value')`'s type is Properties.

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1113,6 +1116,39 @@ class TagBox<
     return selectedItems;
   }
 
+  _shouldUseClickOrderForTags(values): boolean {
```
r-farkhutdinov: Add type signatures to all functions (add type to `values`). Please, also do it for all new methods and callbacks below.

packages/devextreme/testing/tests/DevExpress.ui.widgets.editors/tagBox.tests.js
```diff
@@ -1078,6 +1078,52 @@ QUnit.module('multi tag support', {
         assert.deepEqual(this.getTexts($tagBox.find('.' + TAGBOX_TAG_CLASS)), ['1', '2'], 'tags have correct text');
     });
 
+    QUnit.test('TagBox should preserve direct click order in leading tag when showMultiTagOnly is false', function(assert) {
```
r-farkhutdinov: Looks like this test does not provide any value as it does not check the regression (in this scenario it will always pass even though the behavior is incorrect). Consider removing this test case

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1113,6 +1116,39 @@ class TagBox<
...
+    return !this.option('showMultiTagOnly')
+      && this.option('maxDisplayedTags')
+      && values.length > this.option('maxDisplayedTags');
+  }
+
+  _sortSelectedItemsByValues(selectedItems, values) {
+    if (!this._shouldUseClickOrderForTags(values)) {
+      return selectedItems;
+    }
+    // @ts-expect-error ts-error
```
r-farkhutdinov: Avoid adding new `// @ts-expect-error ts-error`. In this case, `_valueGetterExpr` and `_valueGetter` can be added to a new local interface that extends `TagBoxProperties` so that we don't ignore the errors.

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1046,7 +1046,10 @@ class TagBox<
       this._selectedItems = this._getItemsFromPlain(this._valuesToUpdate);
 
       if (this._selectedItems.length === this._valuesToUpdate.length) {
-        this._tagsToRender = this._selectedItems;
+        this._tagsToRender = this._sortSelectedItemsByValues(
```
r-farkhutdinov, line 1052: The fix only affects rendering the items, but not storing. So, `_tagsToRender` and `_selectedItems` will be inconsistent. Consider updating the storing order as well (if possible).

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1113,6 +1116,39 @@ class TagBox<
...
+  _sortSelectedItemsByValues(selectedItems, values) {
+    if (!this._shouldUseClickOrderForTags(values)) {
+      return selectedItems;
+    }
+    // @ts-expect-error ts-error
+    const isValueExprDefault = this._valueGetterExpr() === 'this';
+
+    const mappedSelectedItems = selectedItems.reduce((result, item) => {
+      // @ts-expect-error ts-error
+      const itemValue = isValueExprDefault ? JSON.stringify(item) : this._valueGetter(item);
```
r-farkhutdinov, line 1142: The path with custom `valueExpr` should be tested as well. Please, add a test case for the scenario.

r-farkhutdinov (changes requested):

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1113,6 +1116,39 @@ class TagBox<
...
+    const selectedByOrderItems = values.reduce((result, currentValue) => {
+      const item = mappedSelectedItems[JSON.stringify(currentValue)];
+      if (isDefined(item)) {
+        result.push(item);
+      }
+
+      return result;
+    }, []);
+
+    return selectedByOrderItems;
```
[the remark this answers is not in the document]
r-farkhutdinov:
```
const products = [
  { ID: 'a', Name: 'HD Video Player' },
  { ID: 'b', Name: 'SuperHD Video Player' },
  { ID: 'c', Name: 'SuperPlasma 50' },
  { ID: 'd', Name: 'SuperLED 50' },
];
```

With this example, the implementation will not work correctly. 

For this scenario (string keys), please add a unit test.

packages/devextreme/js/__internal/ui/m_tag_box.ts
vorobey: _selectedItems is using in several places, i decided not affect them in my first request

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -95,6 +95,10 @@ class TagBox<
 
   _tagsToRender?: any[];
 
+  declare _valueGetterExpr: () => string;
+
+  declare _valueGetter: (item: any) => any;
```
r-farkhutdinov: Please, remove the declarations of `_valueGetterExpr` and `_valueGetter`. As discussed, it can be annotated with:
`// @ts-expect-error _valueGetterExpr is injected by DataExpressionMixin` and
`// @ts-expect-error _valueGetter is injected by DataExpressionMixin`

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -1113,6 +1116,43 @@ class TagBox<
...
+  _shouldUseClickOrderForTags(values: TagBox['_valuesToUpdate']): boolean {
+    const { maxDisplayedTags, showMultiTagOnly } = this.option();
+
+    return !showMultiTagOnly
+      && isDefined(maxDisplayedTags)
+      && values.length > maxDisplayedTags;
+  }
+
+  _sortSelectedItemsByValues(
+    selectedItems: TagBox['_selectedItems'],
```
r-farkhutdinov: Let's remove all references of `TagBox['property']` for clarity and just use `Array<string | number | any>` from `.d.ts` [types](https://github.com/DevExpress/DevExtreme/blob/26_1/packages/devextreme/js/ui/tag_box.d.ts#L274).

r-farkhutdinov (approved):

PR #33643, opened 2026-05-20
Scheduler - Refactor Appointments Collection - KBN

bit-byte0:
## What
 Keyboard navigation for new `appointments_new/` architecture (Del/Home/End/Enter/Space)
 
 ## How
- New props: `editing.allowDeleting`, `onDeleteKeyPress`, `onItemActivate` on `AppointmentsProperties`
- `focus_controller`: switch(true) per key in `onViewItemKeyDown`
- Del passes correct occurrence `startDate`, recurring delete excludes right instance

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
@@ -19,7 +19,7 @@ export class AppointmentsFocusController {
 
   private needRestoreFocusIndex = -1;
 
-  private get sortedAppointments(): SortedEntity[] {
+  private get sortedItems(): SortedEntity[] {
     return this.appointments.option().getSortedItems();
   }
 
@@ -110,7 +110,7 @@ export class AppointmentsFocusController {
 
   private handleTabKeyDown(e: KeyboardKeyDownEvent, sortedIndex: number): void {
     const nextIndex = sortedIndex + (e.shift ? -1 : 1);
-    const nextItemData = this.sortedAppointments[nextIndex];
+    const nextItemData = this.sortedItems[nextIndex];
 
     if (!nextItemData) {
       return;
@@ -134,7 +134,7 @@ export class AppointmentsFocusController {
   }
 
   private handleHomeKeyDown(e: KeyboardKeyDownEvent): void {
-    const firstSortedItem = this.sortedAppointments[0];
+    const firstSortedItem = this.sortedItems[0];
     if (firstSortedItem) {
       e.originalEvent.preventDefault();
       this.focusBySortedItem(firstSortedItem);
@@ -142,7 +142,7 @@ export class AppointmentsFocusController {
   }
 
   private handleEndKeyDown(e: KeyboardKeyDownEvent): void {
-    const lastSortedItem = this.sortedAppointments[this.sortedAppointments.length - 1];
+    const lastSortedItem = this.sortedItems[this.sortedItems.length - 1];
     if (lastSortedItem) {
       e.originalEvent.preventDefault();
       this.focusBySortedItem(lastSortedItem);
--- a/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
@@ -52,6 +52,8 @@ export class AppointmentsFocusController {
   public onViewItemKeyDown(viewItem: ViewItem, e: KeyboardKeyDownEvent): void {
     if (e.key === 'Tab') {
       this.handleTabKeyDown(e, viewItem.option().sortedIndex);
+    } else if (e.key === 'Delete') {
+      this.handleDeleteKeyDown(viewItem.option().sortedIndex);
     }
   }
 
@@ -90,6 +92,15 @@ export class AppointmentsFocusController {
 
     e.originalEvent.preventDefault();
     this.focusByItemData(nextItemData);
+  }
+
+  private handleDeleteKeyDown(sortedIndex: number): void {
+    const { allowDelete, onDeleteKeyPress } = this.appointments.option();
+    if (!allowDelete) { return; }
+
+    const itemData = this.sortedAppointments.find((s) => s.sortedIndex === sortedIndex)?.itemData;
+    if (!itemData) { return; }
+    onDeleteKeyPress({ data: itemData, target: null });
   }
 
   private focusByItemData(itemData: SortedEntity): void {
--- a/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
@@ -10,6 +10,10 @@ import { AppointmentCollector } from './appointment_collector';
 import type { Appointments } from './appointments';
 import type { ViewItem } from './view_item';
 
+interface AppointmentsFocusControllerHandlers {
+  onAppointmentEnterKeyDown: (appointmentView: BaseAppointmentView, event: DxEvent) => void;
+}
+
 export class AppointmentsFocusController {
   private focusableSortedIndex = 0;
 
@@ -27,7 +31,10 @@ export class AppointmentsFocusController {
     return this.appointments.option().tabIndex;
   }
 
-  constructor(private readonly appointments: Appointments) { }
+  constructor(
+    private readonly appointments: Appointments,
+    private readonly handlers: AppointmentsFocusControllerHandlers,
+  ) { }
 
   public onViewItemClick(viewItem: ViewItem): void {
     this.focusViewItem(viewItem);
@@ -119,12 +126,9 @@ export class AppointmentsFocusController {
     const { allowDelete, onDeleteKeyPress } = this.appointments.option();
     if (!allowDelete) { return; }
 
-    const sortedItem = this.sortedAppointments[viewItem.option().sortedIndex];
-    if (!sortedItem) { return; }
-
     const appointmentViewItem = viewItem as BaseAppointmentView;
     onDeleteKeyPress({
-      appointmentData: sortedItem.itemData,
+      appointmentData: appointmentViewItem.appointmentData,
       targetedAppointmentData: appointmentViewItem.targetedAppointmentData,
     });
   }
@@ -146,15 +150,16 @@ export class AppointmentsFocusController {
   }
 
   private handleEnterKeyDown(viewItem: ViewItem, e: KeyboardKeyDownEvent): void {
-    const { onItemActivate } = this.appointments.option();
-    const sortedItem = this.sortedAppointments[viewItem.option().sortedIndex];
-    if (!sortedItem) { return; }
     e.originalEvent.preventDefault();
-    const appointmentViewItem = viewItem as BaseAppointmentView;
-    onItemActivate({
-      data: sortedItem.itemData,
-      targetedAppointmentData: appointmentViewItem.targetedAppointmentData,
-    });
+
+    if (viewItem instanceof AppointmentCollector) {
+      return;
+    }
+
+    this.handlers.onAppointmentEnterKeyDown(
+      viewItem as BaseAppointmentView,
+      e.originalEvent as DxEvent,
+    );
   }
 
   private focusBySortedItem(sortedItem: SortedEntity): void {
--- a/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
@@ -5,6 +5,8 @@ import { focus } from '@ts/events/m_short';
 
 import { getRawAppointmentGroupValues } from '../utils/resource_manager/appointment_groups_utils';
 import type { SortedEntity } from '../view_model/types';
+import type { BaseAppointmentView } from './appointment/base_appointment';
+import { AppointmentCollector } from './appointment_collector';
 import type { Appointments } from './appointments';
 import type { ViewItem } from './view_item';
 
@@ -14,7 +16,7 @@ export class AppointmentsFocusController {
   private needRestoreFocusIndex = -1;
 
   private get sortedAppointments(): SortedEntity[] {
-    return this.appointments.option().getSortedAppointments();
+    return this.appointments.option().getSortedItems();
   }
 
   private get isVirtualScrolling(): boolean {
@@ -55,19 +57,19 @@ export class AppointmentsFocusController {
         this.handleTabKeyDown(e, viewItem.option().sortedIndex);
         break;
       case e.key === 'Delete':
-        this.handleDeleteKeyDown(viewItem.option().sortedIndex);
+        this.handleDeleteKeyDown(viewItem);
         break;
       case e.key === 'Home':
-        this.handleHomeKeyDown();
+        this.handleHomeKeyDown(e);
         break;
       case e.key === 'End':
-        this.handleEndKeyDown();
+        this.handleEndKeyDown(e);
         break;
       case e.key === 'Enter':
-        this.handleEnterKeyDown(viewItem.option().sortedIndex);
+        this.handleEnterKeyDown(viewItem, e);
         break;
       case e.key === ' ':
-        this.handleEnterKeyDown(viewItem.option().sortedIndex);
+        this.handleEnterKeyDown(viewItem, e);
         break;
       default:
         break;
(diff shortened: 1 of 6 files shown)


packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -1071,6 +1071,7 @@ class Scheduler extends SchedulerOptionsBaseWidget {
       const appointmentsConfig: Partial<AppointmentsProperties> = {
         tabIndex: this.option('tabIndex'),
         currentView: this.option('currentView') as ViewType,
+        allowDelete: this.editing.allowUpdating && this.editing.allowDeleting,
```
Tucchhaa: `allowDelete` value is only passed to Appointments on init, but there wasn't added a code that would update `allowDelete` on runtime change.

But actually deleting works as expected, because the legacy implementation does it for new appointments [here](https://github.com/DevExpress/DevExtreme/blob/d1f3a12e667df4fdbe8ee40ff5f1b622ab325550/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L506).

Even though it works, I think we can improve something there:
1) [this repaint call](https://github.com/DevExpress/DevExtreme/blob/d1f3a12e667df4fdbe8ee40ff5f1b622ab325550/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L611) is redundant for the new appointments. 

2) [Weird calculation](https://github.com/DevExpress/DevExtreme/blob/d1f3a12e667df4fdbe8ee40ff5f1b622ab325550/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L601) for allowDelete. In tooltip to delete appointmen just `editing.allowDelete` is [used](https://github.com/DevExpress/DevExtreme/blob/d1f3a12e667df4fdbe8ee40ff5f1b622ab325550/packages/devextreme/js/__internal/scheduler/tooltip_strategies/tooltip_strategy_base.ts#L166).

What I suggest to do:
1) pass scheduler.editing.allowDelete value to Appointments. Scheduler normalizes the 'editing' option [here](https://github.com/DevExpress/DevExtreme/blob/d1f3a12e667df4fdbe8ee40ff5f1b622ab325550/packages/devextreme/js/__internal/scheduler/tooltip_strategies/tooltip_strategy_base.ts#L166). We need to pass this to new Appointments

2) Don't call `bringEditingModeToAppointments` [here](https://github.com/DevExpress/DevExtreme/blob/d1f3a12e667df4fdbe8ee40ff5f1b622ab325550/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L506) for new Appointments. Just update Appointments.option().allowDeleting directly via: `Appointments.option('allowDeleting' this.editing.allowDeleting)`

3) Add integration tests for these cases ([here](https://github.com/DevExpress/DevExtreme/blob/d1f3a12e667df4fdbe8ee40ff5f1b622ab325550/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.test.ts#L1)):

a. Appt can be deleted when editing=true
b. Appt can be deleted when editing.allowDeleting=true
c. Appt can be deleted when editing.allowDeleting=true and editing.allowUpdating=false
d. Appt cant be deleted when editing.allowDeleting=false
e. Appt cant be deleted when editing=false

I think it can be a matrix test.

Please let me know if you have any questions

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -92,6 +111,35 @@ export class AppointmentsFocusController {
     this.focusByItemData(nextItemData);
   }
 
+  private handleDeleteKeyDown(sortedIndex: number): void {
+    const { allowDelete, onDeleteKeyPress, getDataAccessor } = this.appointments.option();
+    if (!allowDelete) { return; }
+
+    const entity = this.sortedAppointments[sortedIndex];
```
Tucchhaa: Let's avoid variable name `entity` because it's very generic and tells nothing. 

I see that the name of the type is `SortedEntity`, but it's also something that we can improve.

I suggest to rename `entity` to `itemData`, the same applies to all other places where name `entity` is used

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -92,6 +111,35 @@ export class AppointmentsFocusController {
...
+    if (!entity) { return; }
```
Tucchhaa: If 'delete' was pressed on a collector item, e.g.:

Image

Then an exception is thrown in the console.

Old appointments implementation also throws some kind of error. I think we might need to discuss it in a call. But I would suggest to prevent appointment deleting in this case

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -92,6 +111,35 @@ export class AppointmentsFocusController {
...
+
+    const occurrence = { ...entity.itemData };
+    getDataAccessor().set('startDate', occurrence, new Date(entity.source.startDate));
+
+    onDeleteKeyPress({ appointment: entity.itemData, occurrence });
```
Tucchhaa: This is a strange way to get occurrence data, because there's a dedicated function for that: [getTargetedAppointmentData](https://github.com/DevExpress/DevExtreme/blob/d1f3a12e667df4fdbe8ee40ff5f1b622ab325550/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts#L357).

I suggest to make `BaseAppointment.targetedAppointmentData` and `BaseAppointment.appointmentData` properties to be public, so we could use it here:

```js
onDeleteKeyPress({
  appointmentData: viewItem.appointmentData,
  targetedAppointmentData: viewItem.targetedAppointmentData
})
```

Actually I have made these props in my PR to [be public](https://github.com/DevExpress/DevExtreme/pull/33633/changes#diff-4ff46915504f4085580834fc78d06198fd3d420e0a5c0eb30cdc1feeb5f2607fR41)

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -92,6 +111,35 @@ export class AppointmentsFocusController {
...
+  }
+
+  private handleHomeKeyDown(): void {
+    const firstAppointment = this.sortedAppointments[0];
+    if (firstAppointment) { this.focusByItemData(firstAppointment); }
```
Tucchhaa: We need `e.preventDefault` here, like in [the legacy impl](https://github.com/DevExpress/DevExtreme/blob/d1f3a12e667df4fdbe8ee40ff5f1b622ab325550/packages/devextreme/js/__internal/scheduler/appointments/m_appointments_kbn.ts#L140).

Otherwise body element may scroll to the bottom if it has large height.

Please add a jest test to check that event is prevented on 'home' press

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -92,6 +111,35 @@ export class AppointmentsFocusController {
...
+  }
+
+  private handleEndKeyDown(): void {
+    const lastAppointment = this.sortedAppointments[this.sortedAppointments.length - 1];
+    if (lastAppointment) { this.focusByItemData(lastAppointment); }
```
Tucchhaa: Same as for Home key down: https://github.com/DevExpress/DevExtreme/pull/33643/changes#r3279707087

Tucchhaa (changes requested):

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
bit-byte0: In case c (`allowDeleting=true, allowUpdating=false`) - should Delete key work? The tooltip only checks `allowDeleting`, but `allowUpdating=false` implies the data is read-only. Should keyboard delete follow the same logic as tooltip, or should it respect allowUpdating too?

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
bit-byte0: `SortedEntity` already has a field called `.itemData` (the appointment data inside). So if we name the variable itemData, you'd write `itemData.itemData` to access the appointment.

How about `sortedItem` instead - consistent with sortedIndex, sortedAppointments, and sortedItems used elsewhere?

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
bit-byte0: Confirmed - the guard doesn't protect here because collector sortedIndex may overlap with `sortedAppointments` indices, potentially deleting the wrong appointment. We'll add an explicit `instanceof AppointmentCollector` check in `onViewItemKeyDown` to skip **Delete** on collectors entirely. This matches your suggestion to prevent deleting in this case. What do you think?

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
Tucchhaa: I think we should follow tooltip's behavior (only check allowDeleting), because [doc](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/Configuration/editing/#allowDeleting) doesn't state that allowUpdating must be enabled too to allow appt deletion

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
Tucchhaa: You're right, I agree with `sortedItem`. Can you please update names of other occurrences of `itemData`?
focusByItemData -> focusBySortedItem

Tucchhaa: What do you think about renaming `getSortedAppointments` to `getSortedItems`, so that it would be consistent everywhere?

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
Tucchhaa: Yep, that's ok for me. Please also add a test. 

And small suggestion: let's move [this](https://github.com/DevExpress/DevExtreme/pull/33643/changes#diff-85b678a5a7516e5d4077bcb3d76162a0a55fe4691983c9322530ce0c42b36190R59) if-statement inside the handleDeleteKeydown and pass viewItem to all the keydown handlers instead of sortedIndex

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
bit-byte0: Let's try it out

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -89,20 +110,64 @@ export class AppointmentsFocusController {
...
+
+    const { allowDelete, onDeleteKeyPress } = this.appointments.option();
+    if (!allowDelete) { return; }
+
+    const sortedItem = this.sortedAppointments[viewItem.option().sortedIndex];
+    if (!sortedItem) { return; }
+
+    const appointmentViewItem = viewItem as BaseAppointmentView;
+    onDeleteKeyPress({
+      appointmentData: sortedItem.itemData,
```
Tucchhaa: it's possible to use `viewItem.appointmentData` here. Then there would be no need for sortedItem at all

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -89,20 +110,64 @@ export class AppointmentsFocusController {
...
+  }
+
+  private handleEnterKeyDown(viewItem: ViewItem, e: KeyboardKeyDownEvent): void {
+    const { onItemActivate } = this.appointments.option();
+    const sortedItem = this.sortedAppointments[viewItem.option().sortedIndex];
+    if (!sortedItem) { return; }
+    e.originalEvent.preventDefault();
+    const appointmentViewItem = viewItem as BaseAppointmentView;
+    onItemActivate({
+      data: sortedItem.itemData,
```
Tucchhaa: let's use `viewItem.appointmentData` here as well

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
```diff
@@ -89,20 +110,64 @@ export class AppointmentsFocusController {
...
+    }
+  }
+
+  private handleEnterKeyDown(viewItem: ViewItem, e: KeyboardKeyDownEvent): void {
+    const { onItemActivate } = this.appointments.option();
+    const sortedItem = this.sortedAppointments[viewItem.option().sortedIndex];
+    if (!sortedItem) { return; }
+    e.originalEvent.preventDefault();
+    const appointmentViewItem = viewItem as BaseAppointmentView;
+    onItemActivate({
```
Tucchhaa: I just found out that in the old impl `onAppointmentClick` event is fired when pressing 'enter' or 'space' on appointment.

However I think the old impl is confusing in this case, because according to the [doc of onAppointmentClick](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/Configuration/#onAppointmentFormOpening) e.cancel should prevent showing tooltip, but on 'enter' keydown the edit popup is shown.

Since the same doc page doesn't specify anything about keyboard events, I suggest to call [Appointments.onAppointmentDblClick](https://github.com/DevExpress/DevExtreme/blob/26_1/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts#L466), on appointment enter/space keydown to open the appointment form (instead of `onItemActivate`).

Please add a test that would check that onAppointmentDblClick is fired on enter/space keydown events

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.test.ts
```diff
@@ -888,6 +888,52 @@ describe('Appointments', () => {
         expect(document.activeElement).toBe(viewItem0?.$element().get(0));
       });
 
+      it('should prevent default browser behavior on Home key', () => {
```
Tucchhaa, line 901: please add the same test for enter and space keys, because they also have e.preventDefault()

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
Tucchhaa: Also there's a bug: when enter/space keydown occurs on appointmentCollector, appointment popup is shown instead of tooltip.

So on collector enter/space keydown, you need to call [Appointments.onCollectorClick](https://github.com/DevExpress/DevExtreme/blob/26_1/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts#L495). 

Please add a test for this case

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.test.ts
```diff
@@ -878,6 +862,220 @@ describe('Appointments', () => {
...
+        const viewItem = instance.getViewItemBySortedIndex(0);
+        (viewItem?.$element().get(0) as HTMLElement).click();
+        fireEvent.keyDown(viewItem?.$element().get(0) as HTMLElement, { key: ' ' });
+
+        expect(onItemActivate).toHaveBeenCalledTimes(1);
+        expect(onItemActivate).toHaveBeenCalledWith(
+          expect.objectContaining({ data: defaultAppointmentData }),
+        );
+      });
+    });
```
Tucchhaa, line 1217: please add unit tests for this cases too:
- on appointment enter keydown appointment popup should be shown
- on appointment space keydown appointment popup should be shown 
- on collector enter keydown tooltip should be shown
- on collector space keydown tooltip should be shown

you can use [this test](https://github.com/DevExpress/DevExtreme/blob/26_1/packages/devextreme/js/__internal/scheduler/appointments_new/appointments.test.ts#L1118) as reference

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.focus_controller.ts
bit-byte0: So I skipped `onCollectorEnterKeyDown` because `AppointmentCollector` contains a **Button** which already handles Enter/Space as a click natively, so adding an explicit handler would trigger `onCollectorClick` twice. We only call `preventDefault()` to prevent page scroll and let the **Button** handle the rest. I hope this is OK approach for you if not - just let me know

Tucchhaa (approved):

PR #33657, opened 2026-05-21
List: expose NoDataText to screen readers when empty (T1329047)

dmlvr:


packages/devextreme/js/__internal/ui/collection/collection_widget.base.ts
```diff
@@ -58,7 +58,7 @@ const ITEM_INDEX_KEY = 'dxItemIndex';
 const ITEM_TEMPLATE_ID_PREFIX = 'tmpl-';
 const ITEMS_OPTIONS_NAME = 'dxItem';
 const ITEM_RESPONSE_WAIT_CLASS = 'dx-item-response-wait';
-const EMPTY_COLLECTION = 'dx-empty-collection';
+export const EMPTY_COLLECTION = 'dx-empty-collection';
```
pharret31: I would rename this to *_CLASS for consistency purposes

packages/devextreme/js/__internal/ui/list/list.base.ts
```diff
@@ -1029,6 +1030,19 @@ export class ListBase extends CollectionWidget<ListBaseProperties, Item> {
...
   }
 
+  _renderEmptyMessage(rootNodes?: Item[]): void {
+    super._renderEmptyMessage(rootNodes);
+    this._updateFocusTargetAriaOnEmptyState();
+  }
+
+  _updateFocusTargetAriaOnEmptyState(): void {
+    const { noDataText } = this.option();
+    const isListEmpty = this.$element().hasClass(EMPTY_COLLECTION_CLASS);
```
pharret31: Is that a good idea to rely on class existence instead of actual items state?

packages/devextreme/js/__internal/ui/collection/collection_widget.base.ts
dmlvr: agree, fixed it

packages/devextreme/js/__internal/ui/list/list.base.ts
dmlvr: I think yes, we can check list items count, but on the widget level we have `_renderEmptyMessage` which do it and toggle class.

I’d skip the extra items check. The empty-state source of truth is already the DOM via EMPTY_COLLECTION_CLASS, while items can drift from the rendered state and hide mismatches.

Keeping only this.$element().hasClass(EMPTY_COLLECTION_CLASS) plus this._$noData?.text() feels cleaner and more accurate.

pharret31 (dismissed):

r-farkhutdinov (dismissed):

EugeniyKiyashko (dismissed):

EugeniyKiyashko (approved):

r-farkhutdinov (approved):

PR #33660, opened 2026-05-21
AI Assistant: DataGrid - write JS frameworks demos (React)

Raushen:


apps/demos/Demos/DataGrid/AIAssistant/React/service.ts
```diff
@@ -0,0 +1,86 @@
...
+  const params: Record<string, unknown> = {
+    messages,
+    model: AzureOpenAIConfig.deployment,
+    max_tokens: 1000,
+    temperature: 0.7,
+    response_format: {
+      type: 'json_schema',
+      json_schema: {
+        name: 'grid_assistant_response',
+        strict: false,
```
dmirgaev, line 27: We can already use `strict: true` if nothing has changed

apps/demos/Demos/DataGrid/AIAssistant/React/service.ts
```diff
@@ -0,0 +1,86 @@
...
+        schema: responseSchema,
+      },
+    },
+  };
+
+  const response = await aiService.chat.completions.create(params as any, { signal });
```
dmirgaev: Can we avoid using `any`? As I can see in other demos, there is no need to use `as any` for params. Otherwise, we can redefine the type in `./types.ts`

apps/demos/Demos/DataGrid/AIAssistant/React/App.tsx
```diff
@@ -0,0 +1,141 @@
...
+      ...suggestions,
+      onItemClick: onSuggestionItemClick,
+    },
+  };
+
+  return (
+    <DataGrid
+      id="gridContainer"
+      dataSource={sales}
+      showBorders={true}
```
anna-shakhova: lets use short syntax for boolean value for `true` here and in other places, it is more react-style
```suggestion
      showBorders
```

apps/demos/Demos/DataGrid/AIAssistant/React/service.ts
anna-shakhova: better use `false` - with strict mode model can fill optional values with `null`, as result validation fails

Raushen: We've discussed and agreed we will take a look on the issue in the future.

apps/demos/Demos/DataGrid/AIAssistant/React/service.ts
```diff
@@ -0,0 +1,87 @@
+import { AIIntegration } from 'devextreme-react/common/ai-integration';
+import type { RequestParams, Response } from 'devextreme-react/common/ai-integration';
+import { AzureOpenAI } from 'openai';
+import type { ChatCompletionCreateParamsNonStreaming } from 'openai/resources/chat/completions';
```
dmirgaev: Use `OpenAI.ChatCompletionCreateParamsNonStreaming` instead + need to import `OpenAI` from `openai`

anna-shakhova (approved):

apps/demos/Demos/DataGrid/AIAssistant/React/service.ts
```diff
@@ -0,0 +1,86 @@
+import { AIIntegration } from 'devextreme-react/common/ai-integration';
+import type { RequestParams, Response } from 'devextreme-react/common/ai-integration';
+import OpenAI, { AzureOpenAI } from 'openai';
```
dmirgaev, line 3: @Raushen Why do we use `import { OpenAI } from 'openai';` everywhere: Vue and Angular, types below - but default export is used here?
`import OpenAI, { AzureOpenAI } from 'openai';`

PR #33665, opened 2026-05-21
AI Assistant: DataGrid - write JS frameworks demos (Angular)

Raushen:


apps/demos/Demos/DataGrid/AIAssistant/Angular/app/app.component.ts
```diff
@@ -0,0 +1,107 @@
...
+          },
+        ] as SuggestionItem[],
+        onItemClick: (e: DxButtonGroupTypes.ItemClickEvent) => {
+          this.onSuggestionItemClick(e);
+        },
+      },
+    };
+  }
+
+  onSuggestionItemClick(e: DxButtonGroupTypes.ItemClickEvent) {
```
anna-shakhova, line 85: ItemClickEvent is generic, DxButtonGroupTypes.ItemClickEvent<SuggestionItem> should work without necessity of assserion

apps/demos/Demos/DataGrid/AIAssistant/Angular/app/ai/ai.service.ts
```diff
@@ -0,0 +1,118 @@
...
+      type: 'json_schema',
+      json_schema: {
+        name: 'grid_assistant_response',
+        strict: false,
+        schema: responseSchema,
+      },
+    },
+  };
+
+  const response = await service.chat.completions.create(params as any, { signal });
```
dmirgaev: Just a reminder about `any` which we want to get rid of (=

apps/demos/Demos/DataGrid/AIAssistant/Angular/app/app.component.ts
Raushen: Let's leave it as is as it writes the ItemClickEvent is not generic

apps/demos/Demos/DataGrid/AIAssistant/Angular/app/ai/ai.service.ts
```diff
@@ -0,0 +1,119 @@
+import { AzureOpenAI, OpenAI } from 'openai';
+import type { ChatCompletionCreateParamsNonStreaming } from 'openai/resources/chat/completions';
```
dmirgaev: use `OpenAI.ChatCompletionCreateParamsNonStreaming` instead, as that is unsafe import (=

anna-shakhova (approved):

PR #33666, opened 2026-05-21
AI Assistant: error handling integration tests

dmirgaev:


Alyar666 (dismissed):

packages/devextreme/js/__internal/grids/grid_core/ai_assistant/__tests__/error_handling.integration.test.ts
```diff
@@ -0,0 +1,900 @@
...
+        expect.objectContaining({
+          status: MessageStatus.Failure,
+          errorText: 'Invalid response from the AI service. Please try again.',
+        }),
+      ]);
+      expect(errors.log).toHaveBeenCalledWith('E1068');
+    });
+  });
+
+  describe('network / API error', () => {
```
anna-shakhova, line 164: in fact network/api error is handled another way - it will go through onComplete branch and will fail here, rendering actual error message. Should be rewritten under success test suite

dmirgaev: discussed personally: it checks pre-defined error message instead of using error message

anna-shakhova (approved):

PR #33668, opened 2026-05-21
AI Assistant: DataGrid - write JS frameworks demos (Vue)

Raushen:


apps/demos/Demos/DataGrid/AIAssistant/Vue/App.vue
```diff
@@ -0,0 +1,144 @@
...
+    prompt: 'Group by Product',
+    width: 170,
+  },
+];
+
+interface SuggestionItem extends DxButtonGroupTypes.Item {
+  prompt: string;
+}
+
+const onSuggestionItemClick = (e: DxButtonGroupTypes.ItemClickEvent) => {
```
anna-shakhova, line 111:
```suggestion
const onSuggestionItemClick = (e: DxButtonGroupTypes.ItemClickEvent<SuggestionItem>) => {
```

apps/demos/Demos/DataGrid/AIAssistant/Vue/service.ts
```diff
@@ -0,0 +1,110 @@
...
+      type: 'json_schema',
+      json_schema: {
+        name: 'grid_assistant_response',
+        strict: false,
+        schema: responseSchema,
+      },
+    },
+  };
+
+  const response = await service.chat.completions.create(params as never, { signal });
```
dmirgaev: We can extend `OpenAI.ChatCompletionCreateParamsStreaming` and use it for `params` if needed

apps/demos/Demos/DataGrid/AIAssistant/Vue/App.vue
Raushen: Let's leave it as is as it writes the ItemClickEvent is not generic

apps/demos/Demos/DataGrid/AIAssistant/Vue/service.ts
```diff
@@ -0,0 +1,111 @@
+import { AzureOpenAI, OpenAI } from 'openai';
+import type { ChatCompletionCreateParamsNonStreaming } from 'openai/resources/chat/completions';
```
dmirgaev: Use `OpenAI.ChatCompletionCreateParamsNonStreaming` instead

anna-shakhova (approved):

PR #33690, opened 2026-05-25
Scheduler - Appointments refactoring - Drag-n-Drop

Tucchhaa:
This PR adds DnD functionality for refactored appointments.
The old DnD implementation was scattered across several files:
1. [/m_appointment_drag_behavior](https://github.com/DevExpress/DevExtreme/blob/f345d0f08ddd634df340a1ff6cc8fa4416ada8c5/packages/devextreme/js/__internal/scheduler/m_appointment_drag_behavior.ts#L1)
2. [ workspaces/m_work_space](https://github.com/DevExpress/DevExtreme/blob/f345d0f08ddd634df340a1ff6cc8fa4416ada8c5/packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts#L2276)
3. [m_compact_appointments_helper](https://github.com/DevExpress/DevExtreme/blob/f345d0f08ddd634df340a1ff6cc8fa4416ada8c5/packages/devextreme/js/__internal/scheduler/m_compact_appointments_helper.ts#L62)
4. m_subscribes/m_scheduler/m_appointments_collection received `dragEvent` and managed it depending on the update results

In the new implementation the whole DnD is located in the [appointments_drag_controller.ts](https://github.com/DevExpress/DevExtreme/blob/f345d0f08ddd634df340a1ff6cc8fa4416ada8c5/packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts#L1) and `dragEvent` is not passed to the scheduler at all.


packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
```diff
@@ -2219,10 +2224,55 @@ class SchedulerWorkSpace extends Widget<WorkspaceOptionsInternal> {
     );
   }
 
+  public getCellFromDragTarget($dragTarget: dxElementWrapper): dxElementWrapper | null {
```
Tucchhaa, line 2227: This method has the same logic as [old method](https://github.com/DevExpress/DevExtreme/blob/fe9a52a5eda5e0547ee0d07ceab62e8aca9880cc/packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts#L3407)

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -2119,6 +2213,7 @@ class Scheduler extends SchedulerOptionsBaseWidget {
         appointment,
         targetedAppointment,
         color: this.resourceManager.getAppointmentColor(appointmentConfig),
+        settings,
```
Tucchhaa, line 2216: This change won't affect anything for the old impl. Needed to match [AppointmentTooltipItem type](https://github.com/DevExpress/DevExtreme/pull/33690/changes#diff-cde56855ae4f038c094cd2c5648584805636ba75496fb6c67edccee67861756fR262)

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -1146,6 +1167,9 @@ class Scheduler extends SchedulerOptionsBaseWidget {
 
   private renderMainContainer() {
     this.mainContainer = $('<div>').addClass('dx-scheduler-container');
+    this.$draggableContainer = $('<div>')
+      .addClass('dx-scheduler-draggable-container')
+      .appendTo(this.mainContainer);
```
Tucchhaa, line 1172: In the old implementation the drag clones were placed in the [workspace.$fixedContainer](https://github.com/DevExpress/DevExtreme/blob/f345d0f08ddd634df340a1ff6cc8fa4416ada8c5/packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts#L2979).

In new implementation I create a container inside scheduler's mainContainer, because I pass the container to the draggable at [DragController's creation](https://github.com/DevExpress/DevExtreme/blob/f345d0f08ddd634df340a1ff6cc8fa4416ada8c5/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L839) at this moment workSpace is not defined yet.
And if feels more right, because draggable container is used not only by workspace, but also by tooltip

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -1885,7 +1967,7 @@ class Scheduler extends SchedulerOptionsBaseWidget {
       dragEvent.cancel = new Deferred();
     }
 
-    if (isPromise(updatingOptions.cancel) && dragEvent) {
+    if (isPromise(updatingOptions.cancel) && (dragEvent || this.option('_newAppointments'))) {
```
Tucchhaa, line 1982: in new appts no need to pass dragEvent, because dragController [receives promise](https://github.com/DevExpress/DevExtreme/blob/f345d0f08ddd634df340a1ff6cc8fa4416ada8c5/packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts#L185) on `updateAppointmentOnDrop`, after which it removes the necessary css classes.

packages/devextreme/js/__internal/scheduler/workspaces/m_work_space.ts
```diff
@@ -2219,10 +2224,55 @@ class SchedulerWorkSpace extends Widget<WorkspaceOptionsInternal> {
...
+    if ($dragTarget.length === 0) {
+      return null;
+    }
+
+    const point = this.getPointFromDragTarget($dragTarget);
+    const elements = document.elementsFromPoint(point.x, point.y);
+
```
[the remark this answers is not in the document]
Tucchhaa: applied

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -1609,6 +1665,30 @@ class Scheduler extends SchedulerOptionsBaseWidget {
...
+      this.checkRecurringAppointment(
+        appointmentData,
+        newAppointmentData,
+        startDate,
+        () => {
+          this.updateAppointmentCore(
+            appointmentData,
+            newAppointmentData,
+          ).done(() => resolve());
+        },
```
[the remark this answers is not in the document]
Tucchhaa: applied

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -107,6 +102,22 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
     return this.viewItemBySortedIndex[sortedIndex];
   }
 
+  public getViewModelBySortedIndex(sortedIndex: number): AppointmentViewModelPlain {
```
Tucchhaa, line 105: needed to create tooltip item for single appt (not collector) on tooltip showing

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -107,6 +102,22 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
...
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const result = this.option().viewModel.find(
+      (viewModel) => viewModel.sortedIndex === sortedIndex,
+    )!;
+    return result;
+  }
+
+  public getAppointmentData($element: dxElementWrapper): SafeAppointment {
```
Tucchhaa: needed by DnD, because to update the dragged appointment, the appointmentData is needed.

Currently it's a linear search, which I think is ok (drop happens very rarely). 

Another approach would be to use `$element.data('appointmentData')` to attach appointmentData to the element, but I don't think this is a good idea, because later it could be used anywhere in the codebase which is not supposed to happen

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -316,17 +325,13 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
     this.focusController.resetTabIndex();
   }
 
+  // TODO: remove passing index to appointmentTemplate, need only to avoid BC
   private renderViewItem(
-    fragment: DocumentFragment,
```
Tucchhaa, line 320: moved `fragment` logic to the viewModel rendering functions, because [renderDragClone](https://github.com/DevExpress/DevExtreme/pull/33690/changes#diff-fe8d027bace8707fc88bc8c66b807f3d05219c21aff8edd3b187f70a08c1f641R409) doesn't pass fragment.

packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
```diff
@@ -0,0 +1,211 @@
...
+    if (isSameCell || !isSameScheduler) {
+      this.removeDraggingClasses($(e.itemElement));
+      return;
+    }
+
+    this.options.updateAppointmentOnDrop(e.itemData, this.$highlightedCell)
+      .then(() => { this.removeDraggingClasses($(e.itemElement)); })
+      .catch(() => { });
+
+    this.removeCellHighlight();
```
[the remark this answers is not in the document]
Tucchhaa: applied

packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
```diff
@@ -0,0 +1,211 @@
...
+    const isSameScheduler = this.$highlightedCell.closest(e.fromComponent.$element()).length > 0;
+
+    if (isSameCell || !isSameScheduler) {
+      this.removeDraggingClasses($(e.itemElement));
+      return;
+    }
+
+    this.options.updateAppointmentOnDrop(e.itemData, this.$highlightedCell)
+      .finally(() => { this.removeDraggingClasses($(e.itemElement)); })
+      .catch((err) => { throw err; });
```
sjbur, line 198:
```
.catch((err) => { throw err; });
```

Seems like a unhandled promise rejection. Maybe its better to write something like 

```
import { noop } from '@js/core/utils/common';
...
.catch(noop);
```

What do you think?

packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    this.removeCellHighlight();
+  }
+
+  private onDragCancel(e): void {
```
sjbur: Can you add type here for ``e``?

packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
```diff
@@ -0,0 +1,211 @@
...
+    this.removeDraggingClasses($(e.itemElement));
+  }
+
+  public removeDraggingClasses($dragSource: dxElementWrapper): void {
```
sjbur: Do we need this method to be public? Its being used only in class

sjbur (changes requested):

sjbur:
I think I found a bug:

1. Create recurrent appointment
2. Try to drag an appointment and cancel/close the recurrent appointment edit window
3. The drag styles for the appointment don't dissappear, it is still semitransparent and it does not change on click or other actions

packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
Tucchhaa: In this code, catch will fire only if function inside `finally` throws, because `updateAppointmentOnDrop` always resolve ([L1685](https://github.com/DevExpress/DevExtreme/blob/d5b45b0a148f56c7b812442cab36234df39bb8e1/packages/devextreme/js/__internal/scheduler/m_scheduler.ts#L1685)).

I don't want to put `noop` here, because in this case we wouldn't notice if the error actually happens. This `.catch` handler is here just to make eslint happy, but basically it's no difference if it's present or not

packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
Tucchhaa: onDragCancel is a private callback, so there's no type for it, but I will add DragEndEvent because they have similar parameters

packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
Tucchhaa: made it private

packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
```diff
@@ -0,0 +1,223 @@
...
+      ...this.getCommonDraggableConfig(),
+      // @ts-expect-error private option
+      filter: `.${APPOINTMENT_CLASSES.CONTAINER}`,
+      dragTemplate: (dragInfo: DragTemplateData) => {
+        this.$dragClone = $(dragInfo.itemElement).clone();
+        this.$dragClone.css({ top: '', left: '' });
+        this.$dragClone.removeClass(APPOINTMENT_DRAG_SOURCE_CLASS);
+
+        return this.$dragClone;
+      },
```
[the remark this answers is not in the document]
Tucchhaa: that's misleading. onDragStart is called before dragTemplate, and e.itemData is set in onDragStart

sjbur (approved):

PR #33741, opened 2026-05-27
AI Assistant: adjust filterValue schema for gemini

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/ai_assistant/commands/filtering.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/ai_assistant/commands/filtering.ts
@@ -12,23 +12,33 @@ const FILTER_OPS = [
 interface BasicFilterExpr {
   type: 'basic';
   field: string;
-  operator: typeof FILTER_OPS[number];
+  operator: SearchOperation;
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/grids/grid_core/ai_assistant/commands/filtering.ts
```diff
@@ -12,23 +12,33 @@ const FILTER_OPS = [
...
 interface NegatedFilterExpr {
   type: 'negated';
-  expression: FilterExprObj;
+  expressionId: string;
 }
 
-type FilterExprObj = BasicFilterExpr | CombinedFilterExpr | NegatedFilterExpr;
+type FilterExpr = BasicFilterExpr | CombinedFilterExpr | NegatedFilterExpr;
+
+interface FilterExprNode {
```
anna-shakhova, line 33: The node id lives on this wrapper, not inside the expr union. Azure OpenAI's strict structured-output mode rejects an `anyOf` whose object branches share an identical property (the variants would all carry `id: string`); keeping id out of the union leaves the branches sharing only their `type` discriminator.

dmirgaev (approved):

PR #33748, opened 2026-05-27
ColorBox: remove m prefix and fix some more type and eslint errors

vorobey:


packages/devextreme/js/__internal/ui/color_box/color_box.ts
```diff
@@ -316,7 +317,7 @@ class ColorBox extends DropDownEditor<ColorBoxProperties> {
 
     this._$colorBoxInputContainer.toggleClass(COLOR_BOX_COLOR_IS_NOT_DEFINED, !hasValue);
 
-    if (hasValue) {
+    if (value !== null && value !== undefined) {
```
marker-dao: Could you please elaborate on that?

packages/devextreme/js/__internal/ui/color_box/color_view.ts
```diff
@@ -401,20 +417,16 @@ class ColorView extends Editor<ColorViewProperties> {
     this._renderAlphaChannelElements();
   }
 
-  _makeTransparentBackground($el: dxElementWrapper, color: any): void {
-    if (!(color instanceof Color)) {
-      color = new Color(color);
-    }
+  _makeTransparentBackground($el: dxElementWrapper, color: string | ColorInstance): void {
+    const colorInstance = color instanceof Color ? color : new Color(color);
```
marker-dao: DRY

packages/devextreme/js/__internal/ui/color_box/color_view.ts
```diff
@@ -401,20 +417,16 @@ class ColorView extends Editor<ColorViewProperties> {
...
 
-    $el.css('backgroundColor', this._makeRgba(color));
+    $el.css('backgroundColor', this._makeRgba(colorInstance));
   }
 
-  _makeRgba(color: any): string {
-    if (!(color instanceof Color)) {
-      color = new Color(color);
-    }
+  _makeRgba(color: string | ColorInstance): string {
```
marker-dao: I believe this method should get only ColorInstance, not string | ColorInstance

packages/devextreme/js/__internal/ui/color_box/color_box.ts
r-farkhutdinov: I believe, `isDefined` utility can be used for this case as it's done in other places.

packages/devextreme/js/__internal/ui/color_box/color_view.ts
```diff
@@ -934,15 +946,17 @@ class ColorView extends Editor<ColorViewProperties> {
   }
 
   _updateColor(isHex: boolean, args: ValueChangedEvent): void {
-    let rgba;
+    let rgba:number[] = [];
```
r-farkhutdinov: Please, apply linter fixes

```suggestion
    let rgba: number[] = [];
```

packages/devextreme/js/__internal/ui/color_box/color_view.ts
```diff
@@ -934,15 +946,17 @@ class ColorView extends Editor<ColorViewProperties> {
...
     let newColor = '';
 
     if (isHex) {
       newColor = this._validateHex(`#${this._hexInput.option('value') as string}`);
     } else {
       rgba = this._validateRgb();
       if (this._alphaChannelInput) {
-        rgba.push(this._alphaChannelInput.option('value'));
+        const { value } = this._alphaChannelInput.option();
+        rgba.push(value !== undefined && this._currentColor.isValidAlpha(value)
```
r-farkhutdinov: Let's simplify this and separate to a variable to increase readability

packages/devextreme/js/__internal/ui/color_box/color_view.ts
```diff
@@ -958,15 +972,17 @@ class ColorView extends Editor<ColorViewProperties> {
...
+    return this._currentColor.isValidHex(hex) ? hex : this._currentColor.toHex();
   }
 
   _validateRgb(): number[] {
     let { value: r } = this._rgbInputs[0].option();
     let { value: g } = this._rgbInputs[1].option();
     let { value: b } = this._rgbInputs[2].option();
 
-    if (!this._currentColor.isValidRGB(r, g, b)) {
+    const isInvalidRgb = r === undefined || g === undefined || b === undefined
```
r-farkhutdinov: `undefined` checks should be performed inside of `isValidRGB` method

r-farkhutdinov (changes requested):

packages/devextreme/js/__internal/ui/color_box/color_box.ts
vorobey: TS wasn't satisfy by checking `Boolean(value);` and highlited line
```
colorUtils.makeTransparentBackground(this._$colorResultPreview, value);
``` 
with error that value might be null or undefined

but my fix is bad, actually, it is better to change the way how  `hasValue` is calculated, i'll make fix

packages/devextreme/js/__internal/ui/color_box/color_view.ts
vorobey: but there was a check in this method in case of string, that's why i left "string" in type
probably it is a good idea to check whether we call this with string anywhere

packages/devextreme/js/__internal/ui/color_box/color_view.ts
vorobey: hm, that's odd, my vscode doesn't apply automatically such formatting, not everywhere, i'll fix

r-farkhutdinov (approved):

PR #33749, opened 2026-05-27
TextBox, TextArea: Improve typing && Refactor

marker-dao:

### DIFF
--- a/packages/devextreme/js/__internal/ui/text_box/m_text_box.ts
+++ b/packages/devextreme/js/__internal/ui/text_box/m_text_box.ts
@@ -95,11 +95,11 @@ class TextBox<
         this._$searchIcon.remove();
       }
 
-      this.option({
-        // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
-        showClearButton: this._showClearButton === undefined ? this.option('showClearButton') : this._showClearButton,
-      });
+      if (this._showClearButton === undefined) {
+        return;
+      }
 
+      this.option({ showClearButton: this._showClearButton });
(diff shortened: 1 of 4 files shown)


r-farkhutdinov (dismissed):

packages/devextreme/js/__internal/ui/text_box/m_text_box.ts
```diff
@@ -95,12 +90,17 @@ class TextBox<
...
+
       if (this._$searchIcon) {
         this._$searchIcon.remove();
       }
-      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
-      this.option('showClearButton', this._showClearButton === undefined ? this.option('showClearButton') : this._showClearButton);
-      delete this._showClearButton;
+
+      this.option({
+        // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
vorobey: Why don't use nullish coalesing here?
also if i read this correct, if there is no `this._showClearButton`, we set `option.showClearButton` with itself value?

marker-dao: Because in the current version, we need to check specifically for undefined; the nullish coalescing operator doesn't apply here.

Yes, you're right — this part could be rewritten differently. It doesn't make sense because a subsequent check verifies that the value hasn't changed, so the optionChange function won't be triggered.

I have refactored it here: eb93b5378196a2dc630da6bdeff3eef4dc477fd9

r-farkhutdinov (approved):

vorobey (approved):

PR #33756, opened 2026-05-28
AI Assistant: Filter command does not create Date objects in filter expression

Raushen:


packages/devextreme/js/__internal/core/ai_integration/commands/utils.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Matches "AIDate(year, month, day)" format used by the filtering command.
+ * The year is the full year; month and day are 1-based.
+ */
+const AI_DATE_REGEX = /^AIDate\((\d+),\s*(\d+),\s*(\d+)\)$/;
+
+export function parseDates(_key: string, value: unknown): unknown {
```
anna-shakhova: Suggested scenario covers only date, not date-time cases, it also does not take into account timezones.
Suggestion: try to use ISO format, with description how it is build
Also it may be usefull to ask to return range instead of one value - it is possible we have different rows for same date but different time

anna-shakhova (changes requested):

packages/devextreme/js/__internal/grids/grid_core/ai_assistant/commands/filtering.ts
```diff
@@ -10,13 +10,13 @@ const FILTER_OPS = [
...
 
-type FilterExprArray = | [string, SearchOperation, string | number | boolean | null]
+type FilterExprArray = | [string, SearchOperation, string | number | boolean | Date | null]
   | [FilterExprArray, 'and' | 'or', FilterExprArray]
   | ['!', FilterExprArray];
 
 const filterOpSchema = z.enum(FILTER_OPS);
 
-const filterValueScalarSchema = z.union([z.string(), z.number(), z.boolean(), z.null()]);
+const filterValueScalarSchema = z.union([z.string(), z.number(), z.boolean(), z.null(), z.date()]);
```
dmirgaev: AI can't return an object of type Date, it'll always be a string, therefore that schema is incorrect

dmirgaev: It's better use a separate description for filter value, like following:
```
z.union([
  z.string().describe(
    'A plain string value. If the value represents a date, it must use the AIDate(year, month, day) format instead of ISO string or natural language.',
  ),
  z.number().describe('A numeric filter value.'),
  z.boolean().describe('A boolean filter value.'),
  z.null().describe('A null filter value.'),
]);
```

dmirgaev (changes requested):

packages/devextreme/js/__internal/core/ai_integration/commands/utils.ts
Raushen: Thank you. I have applied ISO format. However there are issues with making instructions for generating ranges. Let's fix them later.

packages/devextreme/js/__internal/grids/grid_core/ai_assistant/commands/utils.ts
```diff
@@ -64,3 +66,18 @@ export const isKeyShapeValid = (
 
   return keyExpr.every((field) => field in key);
 };
+
+type FilterExprValue = BasicFilterExpr['value'];
+
+export function resolveFilterValue(
```
anna-shakhova, line 72: no need to export utility for filter command only to utils, we can keep it within filtering.ts

packages/devextreme/js/__internal/grids/grid_core/ai_assistant/commands/utils.ts
```diff
@@ -64,3 +66,18 @@ export const isKeyShapeValid = (
...
+  dataType: string | undefined,
+  value: FilterExprValue,
+): FilterExprValue {
+  if (typeof value === 'string' && isDateType(dataType)) {
+    if (!dateUtilsTs.isValidDate(value)) {
+      return value;
```
anna-shakhova, line 78: Will this filter work properly? shouldn't we fail here?

anna-shakhova (approved):

dmirgaev (approved):

PR #33761, opened 2026-05-29
Scheduler - Appointments refactoring - Context Menu

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
+++ b/packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
@@ -573,6 +573,96 @@ describe('New Appointments', () => {
     });
   });
 
+  describe('onAppointmentContextMenu', () => {
+    it('should call onAppointmentContextMenu callback', async () => {
+      const onAppointmentContextMenu = jest.fn();
+
+      const appointment = {
+        text: 'Appointment 1',
+        startDate: new Date(2015, 1, 9, 8),
+        endDate: new Date(2015, 1, 9, 9),
+      };
+
+      const { scheduler, POM } = await createScheduler({
+        dataSource: [appointment],
+        currentView: 'day',
+        currentDate: new Date(2015, 1, 9, 8),
+        onAppointmentContextMenu,
+      });
+
+      const appointmentElement = POM.getAppointments()[0].element;
+      fireEvent(appointmentElement, new Event('dxcontextmenu', { bubbles: true }));
+
+      expect(onAppointmentContextMenu).toHaveBeenCalledTimes(1);
+      expect(onAppointmentContextMenu).toHaveBeenCalledWith({
+        appointmentData: appointment,
+        targetedAppointmentData: {
+          ...appointment,
+          displayStartDate: new Date(2015, 1, 9, 8),
+          displayEndDate: new Date(2015, 1, 9, 9),
+        },
+        appointmentElement,
+        component: scheduler,
+        element: scheduler.$element().get(0),
+        event: expect.objectContaining({ type: 'dxcontextmenu' }),
+      });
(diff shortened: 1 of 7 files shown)


packages/devextreme/js/__internal/scheduler/appointments_new/appointment/base_appointment.ts
```diff
@@ -76,6 +78,7 @@ export class BaseAppointmentView<
     this.attachFocusEvents();
     this.attachClickEvent();
     this.attachDblClickEvent();
+    this.attachContextMenuEvent();
     this.attachKeydownEvents();
```
Tucchhaa, line 82: I suggest to assign event handlers on appointments container instead of each appointment element, because these handlers just delegate event to the Appointments' handlers.

Also it would avoid creating handlers for each appointment, which would decrease memory consumption.

I would do it after this PR is merged

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -516,6 +519,21 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
...
+    appointmentView: BaseAppointmentView,
+    event: DxEvent,
+  ): void {
+    const e = {
+      appointmentElement: getPublicElement(appointmentView.$element()),
+      appointmentData: appointmentView.appointmentData,
+      targetedAppointmentData: appointmentView.targetedAppointmentData,
+      event,
+    };
+
```
Tucchhaa, line 533: unlike `onAppointmentClick` or `onAppointmentDblClick`, `onAppointmentContextMenu` doesn't have e.cancel arg according to [the doc](https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxScheduler/Configuration/#onAppointmentContextMenu)

packages/devextreme/js/__internal/scheduler/appointments_new/appointments.ts
```diff
@@ -516,6 +519,21 @@ export class Appointments extends DOMComponent<Appointments, AppointmentsPropert
...
+    // @ts-expect-error 'component' and 'element' are set by action
+    this.option().onAppointmentContextMenu(e);
+  }
```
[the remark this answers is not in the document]
Tucchhaa: fixed

packages/devextreme/js/__internal/scheduler/__tests__/appointments_new.test.ts
```diff
@@ -573,6 +573,96 @@ describe('New Appointments', () => {
...
+      expect(onAppointmentContextMenu).toHaveBeenCalledWith(expect.objectContaining({
+        targetedAppointmentData: {
+          ...appointment,
+          displayStartDate: new Date(2015, 1, 9, 8),
+          displayEndDate: new Date(2015, 1, 9, 9),
+        },
+      }));
+    });
+  });
+
```
bit-byte0, line 665: Action plan says "tooltip item right click" should fire `onAppointmentContextMenu`, but I can't find test that covers this path. Can you point me out on an integration test for right-clicking a tooltip item?

Tucchhaa: I mentioned that in the issue, because I wasn't sure about how onAppointmentContextMenu is implemented in tooltip.

Turns out the onAppointmentContextMenu is supported separately in tooltip (so both new and old impl are compatible with tooltip) and newAppointments impl doesn't directly affect that scenario. As you can see this PR doesn't have changes related to tooltip. That's why I decided not to add a test for this case.

bit-byte0 (approved):

aleksei-semikozov (approved):

PR #33777, opened 2026-06-01
CardView - "Clear selection" button doesn't clear selection if the option method updates the selectedCardKeys property (T1306438)

markallenramirez:


anna-shakhova (dismissed):

packages/devextreme/js/__internal/ui/selection/selection.ts
```diff
@@ -402,4 +402,15 @@ export default class Selection<
...
+  hasMatchingSelectedItems(keys: TKey[]): boolean {
+    const { selectedItems, keyOf } = this.options;
+    if (keys.length === selectedItems.length) {
+      return keys.every((key) => selectedItems.some((item) => {
+        const isKeysEqual = this._selectionStrategy.equalKeys(keyOf(item), key);
+        return isKeysEqual;
+      }));
+    }
+    return false;
+  }
```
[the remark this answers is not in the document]
markallenramirez: duplicates should be handles externally

packages/devextreme/js/__internal/ui/selection/selection.ts
```diff
@@ -402,4 +402,15 @@ export default class Selection<
   loadSelectedItemsWithFilter(): DeferredObj<unknown> {
     return this._selectionStrategy.loadSelectedItemsWithFilter();
   }
+
+  hasMatchingSelectedItems(keys: TKey[]): boolean {
```
anna-shakhova: method name is confusing a bit, it looks like we check any of keys match, but in fact we check all of them match, let's rename it

markallenramirez: thanks. I renamed it to "selectedItemsMatchAllKeys".

anna-shakhova (approved):

EugeniyKiyashko (approved):

PR #33784, opened 2026-06-02
Autocomplete, DropDownButton, DropDownBox: Refactor and improve typing

vorobey:


packages/devextreme/js/__internal/ui/m_autocomplete.ts
```diff
@@ -34,9 +43,8 @@ class Autocomplete extends DropDownList<AutocompleteProperties> {
         }
         return true;
       },
-      downArrow(e): boolean {
-        // @ts-expect-error ts-error
-        if (parent.downArrow.apply(this, arguments) && !isCommandKeyPressed(e)) {
+      downArrow: (e): boolean => {
```
r-farkhutdinov: Is there any reason to convert the method to an arrow function? The original implementation relied on dynamic this binding and forwarded all arguments via apply(this, arguments). The new version changes both behaviors (arguments -> e which is correct, method -> arrow function), so it may not be fully equivalent.

vorobey: main reason was to fix using "arguments", i removed arrows, but have kept call instead of apply, because we only use `e` here

packages/devextreme/js/__internal/ui/m_drop_down_button.ts
```diff
@@ -186,6 +202,7 @@ class DropDownButton extends Widget<DropDownButtonProperties> {
   }
 
   // T977758
+
```
r-farkhutdinov, line 205: Minor: the empty line can be removed so that the comment is attached to method
```suggestion
```

r-farkhutdinov (approved):

PR #33790, opened 2026-06-02
AI Assistant: Add deselect option to selectByIndexes command

dmirgaev:


packages/devextreme/js/__internal/grids/grid_core/ai_assistant/commands/selection.ts
```diff
@@ -45,15 +45,22 @@ export const selectByKeysCommand = defineGridCommand({
 
 const selectByIndexesCommandSchema = z.object({
   indexes: z.array(z.number().int().min(1)).min(1),
+  deselect: z.boolean().optional(),
```
anna-shakhova: use `optionalNullish` instead of optional - it allows null values and sanitize them on parse, which is necessary for proper work in strict mode

anna-shakhova (changes requested):

anna-shakhova (approved):

PR #33826, opened 2026-06-03
Scheduler: refactor core files (TS) - Part 1

bit-byte0:
## What
rename `m_classes.ts` and `m_loading.ts` to enable strict **TS/ESLint** rules, add proper types to loading module

## How
replace `Deferred` with native `Promise`, type all params/returns, fix call sites with `.get(0)` and `.catch(noop)`, update ~14 import paths


packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -697,9 +697,9 @@ class Scheduler extends SchedulerOptionsBaseWidget {
       });
 
       this._dataSource.isLoading() && showLoading({
-        container: this.$element(),
+        container: this.$element().get(0),
         position: {
-          of: this.$element(),
+          of: this.$element() as any,
         },
       });
```
aleksei-semikozov: Floating promise: no `.catch(noop)` here (popup.ts has it). Same for `hideLoading()` at 691/695. Risk: `no-floating-promises` (the PR goal) + unhandled rejection. We can fix it like this:

```suggestion
      if (this._dataSource.isLoading()) {
        showLoading({
          container: this.$element().get(0),
          position: {
            of: this.$element() as any,
          },
        }).catch(noop);
      }
```
And add `.catch(noop)` to both `hideLoading()` calls above (`noop` is already imported).

packages/devextreme/js/__internal/scheduler/loading.ts
```diff
@@ -0,0 +1,40 @@
...
+export function hide(): Promise<boolean | undefined> {
+  if (!loading) {
+    return Promise.resolve(undefined);
+  }
+
+  const instance = loading;
+  return instance.hide().then((result) => {
+    removeLoadPanel();
+    return result;
+  });
```
aleksei-semikozov, line 41: Race condition: `removeLoadPanel()` clears the global `loading`, so a `show()` started before this `hide()` resolves will drop the new panel. Pre-existing (not a regression), but we can fix it like this:

```suggestion
  return instance.hide().then((result) => {
    if (loading === instance) {
      removeLoadPanel();
    }
    return result;
  });
```

aleksei-semikozov (dismissed):

sjbur (dismissed):

aleksei-semikozov (dismissed):

aleksei-semikozov (approved):

PR #33839, opened 2026-06-04
NxIfraPlugin: add ts-node in devDependencies

vorobey:


marker-dao (changes requested):
I'm sure we don't need that dependency here)

PR #33865, opened 2026-06-05
AI Assistant: Add possibility to select/deselect rows on any page depending on scope

dmirgaev:


packages/devextreme/js/common/grids.d.ts
```diff
@@ -223,6 +223,7 @@ export type PredefinedCommands = {
   selectByIndexes: {
     indexes: number[];
     mode: 'select' | 'deselect';
+    scope: 'dataset' | 'page';
```
arman-boyakhchyan:
```suggestion
    scope: 'page' | 'allPages';
```

arman-boyakhchyan (dismissed):

packages/devextreme/js/__internal/grids/grid_core/ai_assistant/commands/utils.ts
```diff
@@ -67,6 +67,27 @@ export const isKeyShapeValid = (
   return keyExpr.every((field) => field in key);
 };
 
+export const splitIntoContiguousRanges = (indexes: number[]): number[][] => {
+  const sorted = [...new Set(indexes)].sort((a, b) => a - b);
+  const ranges: number[][] = [];
+  let current: number[] = [];
+
+  sorted.forEach((value) => {
+    if (current.length === 0 || value === current[current.length - 1] + 1) {
```
anna-shakhova:
```suggestion
    if (current.length === 0 || value === current.at(-1) + 1) {
```

anna-shakhova: if user asks to select every second line, then we generate n/2 ranges with length 2 and make that much load requests. Perhaps, it is worth make ranges not smaller then page

anna-shakhova (changes requested):

arman-boyakhchyan (approved):

Raushen (approved):

anna-shakhova (approved):

Alyar666 (approved):

PR #33872, opened 2026-06-05
AI Assistant: AI uses "or" filter operation instead of "anyof" - so there is no sync with HeaderFilter

Raushen:


packages/devextreme/js/common/grids.d.ts
```diff
@@ -100,6 +101,13 @@ export type ResponseStatusTexts = {
   failure?: string;
 };
 
+/**
+ * @docid
+ * @public
+ * @namespace DevExpress.common.grids
+ */
+export type ScalarFilterValue = string | number | boolean | null | Date;
```
anna-shakhova: nitpick: more common way to place `| null`  last in union - expected types at the beginning, nullish value (nothing) at the end

anna-shakhova (approved):

PR #33877, opened 2026-06-05
AI Chat: disable clear button when chat is empty

dmirgaev:


packages/devextreme/js/__internal/grids/grid_core/ai_chat/ai_chat.ts
```diff
@@ -116,6 +116,11 @@ export class AIChat {
           Chat,
           this.getChatConfig(),
         );
+
+        this.chatInstance?.getDataSource()?.on('changed', () => {
```
anna-shakhova: Event unsubscribtion is missing. In fact, do we really need make it as event subscription - we check button status on each assistant message update. For each user message sent it will be fired 4 times, wouldn't be enough to keep current disable logic + disabling button on first render according to dataSource state and on clearButton click?

anna-shakhova (changes requested):

packages/devextreme/js/__internal/grids/grid_core/ai_chat/ai_chat.ts
Alyar666: > In fact, do we really need make it as event subscription - we check button status on each assistant message update. 

Messages can be added via the Store Push API, so this event is a more suitable place to handle all relevant scenarios.

anna-shakhova (approved):

PR #33899, opened 2026-06-08
CardView - Header filter search is only triggered once when remoteOperations is true (T1305672)

markallenramirez:


packages/devextreme/js/__internal/grids/new/grid_core/data_controller/deferred_cache.ts
```diff
@@ -19,7 +19,7 @@ export const deferredCache = <TArgs, TResult>(
   ): DeferredObj<TResult> => {
     const hasPreviousCall = lastArgs !== null && cachedResult !== null;
     const isArgsSame = hasPreviousCall
-      ? equalByValue(lastArgs, args)
+      ? equalByValue(lastArgs, args, { maxDepth: 5 })
```
Alyar666: We already have the FILTER_OBJ_COMPARE_DEPTH constant defined in data_controller.ts. I suggest moving it to a more appropriate location (for example, creating a const.ts file inside the data_controller directory) and reusing it wherever needed instead of duplicating the value.

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/deferred_cache.test.ts
```diff
@@ -83,6 +83,22 @@ describe('DataController', () => {
       expect(originFn).toHaveBeenCalledTimes(2);
     });
 
+    it('should call origin fn if args differ on deep value (maxDepth=5)', async () => {
```
Alyar666: The test name should call origin fn if args differ on deep value (maxDepth=5) hardcodes the constant value into the description. If the comparison depth is changed in the future, we'll also need to rename the test. It would be better to keep the test name independent of the specific constant value and focus on the behavior being verified.

Alyar666 (commented):
Please add an end-to-end test that covers the actual user scenario.

packages/devextreme/js/__internal/grids/new/grid_core/data_controller/deferred_cache.ts
Raushen: Agree, also I'd suggest to refactor the following code also:
`if (!equalByValue(prevValue, newValue, { maxDepth: 5 })) {`
in the columns_controller.ts

Alyar666 (approved):

Raushen (approved):

PR #33915, opened 2026-06-08
Scheduler: refactor workspaces module (TS): part 4

sjbur:


packages/devextreme/js/__internal/scheduler/workspaces/view_model/grouped_data_map_provider.ts
```diff
@@ -125,29 +159,47 @@ export class GroupedDataMapProvider {
...
         }
-      }
+        return false;
+      }),
+    );
+
+    return foundPosition;
+  }
+
+  private getRowsForCellSearch(
```
sjbur: Added this helper function to simplify double "for" cycle in findCellPositionInMap function because eslint required to do so

packages/devextreme/js/__internal/scheduler/workspaces/view_model/grouped_data_map_provider.ts
```diff
@@ -91,28 +134,23 @@ export class GroupedDataMapProvider {
...
-      : dateTableGroupedMap[groupIndex] || [];
-
-    for (let rowIndex = 0; rowIndex < rows.length; rowIndex += 1) {
-      const row = rows[rowIndex];
-
-      for (let columnIndex = 0; columnIndex < row.length; columnIndex += 1) {
-        const cell = row[columnIndex];
+    // eslint-disable-next-line no-undef-init
+    let foundPosition: CellPositionData | undefined = undefined;
+    this.getRowsForCellSearch(allDay, groupIndex).some(
```
Tucchhaa: I suggest to use `forEach` instead of `some` in both external and internal loops

packages/devextreme/js/__internal/scheduler/workspaces/view_model/grouped_data_map_provider.ts
```diff
@@ -242,26 +298,26 @@ export class GroupedDataMapProvider {
...
-        groupIndex,
-        startDate: this.getGroupStartDate(groupIndex),
-        endDate: this.getGroupEndDate(groupIndex),
+        allDay: allDay ?? false,
+        groupIndex: groupIndex ?? 0,
+        startDate: this.getGroupStartDate(groupIndex ?? 0),
+        endDate: this.getGroupEndDate(groupIndex ?? 0),
       };
-    }).filter(({ startDate }) => Boolean(startDate));
+    }).filter((info): info is GroupInfo => Boolean(info.startDate) && info.endDate !== undefined);
```
Tucchhaa: minor: `Boolean(info.startDate) && Boolean(info.endDate)`

packages/devextreme/js/__internal/scheduler/workspaces/view_model/grouped_data_map_provider.ts
sjbur: "some" stops as soon as cell is found - no more iterations. forEach has no early exit, it always goes through all rows and cells even after result is already found. So forEach here is worse for performance

packages/devextreme/js/__internal/scheduler/workspaces/view_model/grouped_data_map_provider.ts
sjbur: Applied

packages/devextreme/js/__internal/scheduler/workspaces/view_model/grouped_data_map_provider.ts
Tucchhaa: In the current impl, early return happens only for the row that contains the target cell, all the other rows will still iterate over all the cells.

In the old code, the early return happened the moment the target cell was found. So actually performance has gotten worse.

I see 2 options here:
1. Use 2 `for` loops like in the old impl and disable eslint
2. Use `.forEach` because it has almost the same performance as `.some` in the current impl and doesn't confuse readers. Moreover, the complexities of old impl and `.forEach` impl are both O(n*m), I think performance boost in the old impl is still neglible

Tucchhaa: or you can use `.forEach` to iterate rows and `.filter` to iterate cells, the performance will be the same as using `.some`

packages/devextreme/js/__internal/scheduler/workspaces/view_model/grouped_data_map_provider.ts
```diff
@@ -91,23 +134,32 @@ export class GroupedDataMapProvider {
...
     const {
       allDayPanelGroupedMap,
       dateTableGroupedMap,
     } = this.groupedDataMap;
     const { viewOffset } = this.viewOptions;
 
-    const rows = isAllDay && !this.viewOptions.isVerticalGrouping
-      ? allDayPanelGroupedMap[groupIndex] ? [allDayPanelGroupedMap[groupIndex]] : []
-      : dateTableGroupedMap[groupIndex] || [];
+    const gIdx = groupIndex ?? 0;
```
Tucchhaa: minor: I suggest to use another var name: `normalizedGroupIndex`

Tucchhaa (dismissed):

Tucchhaa (approved):

PR #33975, opened 2026-06-12
PivotGrid - A11y- Make screen reader's output meaningful when focusing pivot grid

Tucchhaa:


packages/devextreme/js/__internal/grids/pivot_grid/data_area/m_data_area.ts
```diff
@@ -58,6 +58,7 @@ class DataArea extends AreaItem {
       rtlEnabled: this.component.option('rtlEnabled'),
       bounceEnabled: false,
       updateManually: true,
+      useKeyboard: false,
```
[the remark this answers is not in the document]
Tucchhaa: that's not true, tabindex is not set when useNative is true

Tucchhaa, line 61: removes tabindex attr from scrollable in the data cells table

packages/devextreme/js/__internal/grids/pivot_grid/headers_area/m_headers_area.ts
```diff
@@ -126,6 +126,7 @@ class HorizontalHeadersArea extends AreaItem {
       direction: 'horizontal',
       rtlEnabled: false,
       updateManually: true,
+      useKeyboard: false,
```
Tucchhaa, line 129: removes tabindex attr from scrollable in the column cells table

packages/devextreme/js/__internal/grids/pivot_grid/headers_area/m_headers_area.ts
```diff
@@ -226,6 +227,7 @@ class VerticalHeadersArea extends HorizontalHeadersArea {
       bounceEnabled: false,
       direction: 'vertical',
       updateManually: true,
+      useKeyboard: false,
     });
   }
```
Tucchhaa, line 232: removes tabindex attr from scrollable in the row cells table

packages/devextreme/js/__internal/grids/pivot_grid/fields_area/m_fields_area.ts
```diff
@@ -45,6 +46,19 @@ class FieldsArea extends AreaItem {
...
+    const localizationMessageMap = {
+      row: 'dxPivotGrid-rowFields',
+      column: 'dxPivotGrid-columnFields',
+      data: 'dxPivotGrid-dataFields',
+      filter: 'dxPivotGrid-filterFields',
+    };
+
+    return $('<table>')
+      .attr('role', 'group')
+      .attr('aria-label', localizationMessage.format(localizationMessageMap[this._area]));
```
Tucchhaa, line 59: Tables that contain fields (filter/row/column/data fields) don't visually look like tables and we don't need the screen reader to announce cells position (each time something is focused in the table screen reader outputs: "1 rows 3 columns, row 1 column 1), so we need to set role="group" to prevent that output.

Now the output when focusing a field is: "Column Fields grouping <field name>"

When the arrow kbn for fields will be implemented, we will need to set role=toolbar

packages/devextreme/js/__internal/grids/pivot_grid/m_widget.ts
```diff
@@ -1084,6 +1071,7 @@ class PivotGrid extends Widget {
   _createTableElement() {
     const that = this;
     const $table = ($('<table>') as any)
+      .attr('role', 'presentation')
```
Tucchhaa, line 1117: This is outer table (pivot grid has one outer table and many nested tables), to prevent complex output of screen reader (it would announce cell position according to outer and inner tables), I set role=presentation. This role tells screen reader to ignore this element in a11y tree (but nested elements still will be visible for screen reader)

packages/devextreme/js/__internal/grids/pivot_grid/m_widget.ts
```diff
@@ -1123,7 +1111,34 @@ class PivotGrid extends Widget {
   _initMarkup() {
     const that = this;
     super._initMarkup();
-    that.$element().addClass(PIVOTGRID_CLASS);
+    that.$element()
+      .addClass(PIVOTGRID_CLASS)
+      .attr('role', 'group');
```
Tucchhaa: role=group was discussed with Andrey D., he approved it.

packages/devextreme/testing/tests/DevExpress.ui.widgets.pivotGrid/pivotGrid.markup.tests.js
```diff
@@ -84,6 +85,221 @@ QUnit.module('PivotGrid markup tests', () => {
...
+        const ariaLabel = pivotGrid.$element().attr('aria-label');
+        assert.ok(ariaLabel, 'aria-label is set');
+        assert.equal(ariaLabel, 'Pivot grid with 3 rows and 2 columns. There are 1 filter fields, 1 data fields, 3 column fields, 2 row fields');
+    });
+
+    QUnit.test('aria-label uses leaf row/column counts', function(assert) {
+        if(!windowUtils.hasWindow()) {
+            assert.expect(0);
+            return;
+        }
```
Tucchhaa: pivotGrid.mark.tests.js are also run on server-side qunit test:
https://github.com/DevExpress/DevExtreme/blob/6fef5067ed22509ea1a8213b42004f0e33e9e14c/packages/devextreme/testing/tests/DevExpress.serverSide/pivotGrid.test.js#L2
Job: https://github.com/DevExpress/DevExtreme/actions/runs/27404209943/job/80990008692

So that's why this guard is needed

packages/devextreme/js/localization/messages/fa.json
```diff
@@ -264,6 +264,7 @@
         "dxPagination-ariaPageSize": "اندازه صفحه",
         "dxPagination-ariaPageNumber": "شماره صفحه",
 
+        "dxPivotGrid-ariaLabel": "Pivot grid with {0} rows and {1} columns. There are {2} filter fields, {3} data fields, {4} column fields, {5} row fields",
```
bit-byte0: When there's 1 filter field, the screen reader will say "1 filter fields", so wouldn't it sound more natural to handle singular/plural here? e.g. "1 filter field" vs "2 filter fields"? I don't know if it handles automatically so that's why I decided to ask

packages/devextreme/js/__internal/grids/pivot_grid/m_widget.ts
```diff
@@ -1123,7 +1111,34 @@ class PivotGrid extends Widget {
...
+      return fields.filter((field) => field.visible !== false).length;
+    };
+
+    const filterFieldsCount = countVisibleFields('filter');
+    const dataFieldsCount = countVisibleFields('data');
+    const columnFieldsCount = countVisibleFields('column');
+    const rowFieldsCount = countVisibleFields('row');
+
+    const ariaLabel = localizationMessage
+      // @ts-expect-error Badly typed format method
```
bit-byte0: can you explain what's badly typed so the next person knows whether it's safe to remove?

packages/devextreme/js/__internal/grids/pivot_grid/fields_area/m_fields_area.ts
```diff
@@ -45,6 +46,19 @@ class FieldsArea extends AreaItem {
       .attr('group', this._area);
   }
 
+  _createTableElement() {
+    const localizationMessageMap = {
+      row: 'dxPivotGrid-rowFields',
+      column: 'dxPivotGrid-columnFields',
+      data: 'dxPivotGrid-dataFields',
+      filter: 'dxPivotGrid-filterFields',
+    };
```
bit-byte0, line 55: As far as I understand this map is static so wouldn't it make sense to move it to module scope so it's not recreated on every _createTableElement call? Probably it's an overkill

packages/devextreme/js/localization/messages/fa.json
Tucchhaa: It would complicate the message formatting because:
1)  there are 6 entities that would need to handle singular/plural nouns.
2) there many languages that we need to support, so it would require us to create a separate localization message for every case (which will be a big number, at least 12 = num of entities * 2)

I think current approach is fine, the same approach is used in datagrid:
image

packages/devextreme/js/__internal/grids/pivot_grid/m_widget.ts
Tucchhaa: Here's the definition of messageLocalization.format: https://github.com/DevExpress/DevExtreme/blob/6fef5067ed22509ea1a8213b42004f0e33e9e14c/packages/devextreme/js/__internal/core/localization/message.ts#L108

It has only 1 argument, but uses `arguments` inside. I am not sure if it's beneficial to add explanation for messageLocalization.format to every @ts-expect-error, because it's easy just to check the function signature

packages/devextreme/js/__internal/grids/pivot_grid/fields_area/m_fields_area.ts
Tucchhaa: Yep, it's possible to move it outside. But I decided to put inside to reduce its scope. This map is only needed in this func and I prefer to avoid creating global var. I think the performance overhead for creating this object is neglible, also this func is called only time for each field container during its creation. ([render func](https://github.com/DevExpress/DevExtreme/blob/6fef5067ed22509ea1a8213b42004f0e33e9e14c/packages/devextreme/js/__internal/grids/pivot_grid/area_item/m_area_item.ts#L557))

bit-byte0 (dismissed):

vladaskorohodova (approved):

bit-byte0 (approved):

PR #33983, opened 2026-06-12
TagBox: Refactor & improve types

r-farkhutdinov:

### DIFF
--- a/packages/devextreme/js/__internal/ui/m_tag_box.ts
+++ b/packages/devextreme/js/__internal/ui/m_tag_box.ts
@@ -638,7 +638,7 @@ class TagBox<
   }
 
   _renderSingleLineScroll(): void {
-    const mouseWheelEvent = addNamespace('dxmousewheel', this.NAME ?? '');
+    const mouseWheelEvent = addNamespace('dxmousewheel', this.NAME as string);
     const $element = this.$element();
     const { multiline } = this.option();
 
@@ -675,7 +675,7 @@ class TagBox<
     super._renderEvents();
 
     const input = this._input();
-    const namespace = addNamespace('keydown', this.NAME ?? '');
+    const namespace = addNamespace('keydown', this.NAME as string);
     eventsEngine.on(input, namespace, (e: KeyboardEvent) => {
       const keyName = normalizeKeyName(e) ?? '';
 
@@ -1399,7 +1399,7 @@ class TagBox<
     if (isObject(item) && isDefined(displayValue)) {
       return item;
     }
-    return ensureDefined(displayValue, '') as string;
+    return ensureDefined(displayValue, '');
   }
 
   _getTag(value: TagBoxItem): dxElementWrapper | false {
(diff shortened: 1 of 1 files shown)


packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -320,25 +368,24 @@ class TagBox<
...
-    const isScrollLeft = xor(direction === 'next', rtlEnabled);
-    // @ts-expect-error ts-error
-    let { left: scrollOffset } = this._$focusedTag.position();
-    let scrollLeft = this._$tagsContainer.scrollLeft();
+  _getFocusedTagPosition(direction: 'next' | 'prev'): number {
+    const { rtlEnabled } = this.option();
+    const isScrollLeft = xor(direction === 'next', Boolean(rtlEnabled));
+
+    let { left: scrollOffset = 0 } = this._$focusedTag?.position() || { };
+    let scrollLeft = this._$tagsContainer.scrollLeft() as unknown as number;
```
marker-dao: It's better to use `@ts-expect-error` here instead of a type cast.

marker-dao: Also, I noticed that this class contains checks like `if (this._$tagsContainer) {` and `if (!this._$tagsContainer) {`.

I think we should make the type of the `_$tagsContainer` property optional, and also, to prevent memory leaks, remove `_$tagsContainer` in the `clean` or `dispose` methods.

packages/devextreme/js/__internal/ui/m_tag_box.ts
marker-dao: It's better to set `undefined`.

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -803,72 +862,68 @@ class TagBox<
...
-
-    this._setTagAria($tag, tagText);
+    this._setTagAria($tag, args.text);
     this._updateElementAria(tagId);
+
     return $tag;
   }
 
-  _getFilter(creator) {
+  _getFilter(creator: FilterCreatorInstance): unknown {
```
marker-dao: `filterExpr` is array, not `unknown`. Could you please dive deeper here?

packages/devextreme/js/__internal/ui/m_tag_box.ts
```diff
@@ -782,15 +841,15 @@ class TagBox<
...
   }
 
-  _renderInputSubstitution() {
+  _renderInputSubstitution(): void {
     super._renderInputSubstitution();
     this._updateWidgetHeight();
   }
 
-  _getValue(): any[] {
+  _getValue(): unknown[] {
```
marker-dao: the `value` type is https://github.com/DevExpress/DevExtreme/blob/dd36a375fd1a6f15c9ea330575dfe8b9fec6df8a/packages/devextreme/js/ui/tag_box.d.ts#L312

, `not unknown[]`, I don't have a big deal with it, just make sure that `unknown[]` is safer in this case.

vorobey (approved):

marker-dao (approved):

PR #33994, opened 2026-06-15
 PivotGrid: open header filter popup by hotkey

sjbur:


packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
```diff
@@ -127,6 +127,16 @@ export class FieldChooserBase extends mixinWidget {
...
 
+  protected _applyColumnState(options) {
+    const $element = super._applyColumnState(options);
+
+    if (options.name === 'headerFilter' && $element) {
+      $element.removeAttr('tabindex');
+    }
+
+    return $element;
+  }
```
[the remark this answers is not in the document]
sjbur: Not sure whether we should do this. Suggestions?

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
```diff
@@ -358,14 +368,32 @@ export class FieldChooserBase extends mixinWidget {
...
+      const field: any = $(e.currentTarget).data('field');
+
+      if (!field) {
+        return;
+      }
+
+      const isAltArrowDown = e.type === 'keydown' && e.altKey && e.key === 'ArrowDown';
+
+      if (isAltArrowDown) {
+        const mainGroupField = getMainGroupField(this._dataSource, field);
```
Tucchhaa: why we need `mainGroupField`? Why not just use `field`?

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
```diff
@@ -358,14 +368,32 @@ export class FieldChooserBase extends mixinWidget {
...
     const handler = (e) => {
+      const field: any = $(e.currentTarget).data('field');
+
+      if (!field) {
+        return;
+      }
+
+      const isAltArrowDown = e.type === 'keydown' && e.altKey && e.key === 'ArrowDown';
+
+      if (isAltArrowDown) {
```
Tucchhaa: Now we have 2 places for opening header filter:
1. here
2. the old place: https://github.com/DevExpress/DevExtreme/pull/33994/changes#diff-12a89dcfa5ebbc8ea7357eaf5c97ea7d1eea821081fd156023cd736aa14d4a72R401

I suggest to unify them, remove the old logic and add to this condition a check for header filter icon click

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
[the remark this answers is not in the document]
Tucchhaa: Copilot has indicated a valid problem, but I don't think that we need to fix it as it suggested.

The problem is that [headerFilterMixin._applyColumnState](https://github.com/DevExpress/DevExtreme/blob/26_1/packages/devextreme/js/__internal/grids/grid_core/header_filter/m_header_filter_core.ts#L547) adds several aria-attrs: label, haspopup, role. However we don't need them and we have to remove all of them from header filter icon.

Using headerFilterMixin._applyColumnState to add these attr, so that later they would be removed in the overriden _applyColumnState seems like bad design for me.

I suggest to remove usage of headerFilterMixin, and replace [this _applyColumnState](https://github.com/DevExpress/DevExtreme/blob/26_1/packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts#L215) with a new function that would render headerFilter icon.

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
Tucchhaa: Ah, I see, because there can grouped fields, in which only the first group item has header filter.

I see that the condition is similar to the condition for rendering header filter icon:
https://github.com/DevExpress/DevExtreme/blob/26_1/packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts#L221. 

Maybe create a helper function, which can be reused?
```js
isFieldHasHeaderFilter = (field) => {
   const mainGroupField = // ...
   const isFirstInGroup = (field.groupIndex === 0 || !isDefined(field.groupIndex);
   return mainGroupField .allowFiltering && isFirstInGroup && field.area !== 'data';
}
```

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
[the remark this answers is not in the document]
sjbur: Applied

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
sjbur: Created helper function

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
sjbur: Done

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
```diff
@@ -428,6 +426,32 @@ export class FieldChooserBase extends mixinWidget {
...
+  private _renderHeaderFilterIcon($fieldElement, mainGroupField, field, showColumnLines): void {
+    const allowFiltering = mainGroupField.allowFiltering && !field.groupIndex;
+
+    if (!allowFiltering) {
+      return;
+    }
+
+    const isEmpty = !mainGroupField.filterValues?.length;
+    const $icon = $('<span>').addClass(CLASSES.headerFilter).toggleClass('dx-header-filter-empty', isEmpty);
+
```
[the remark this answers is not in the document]
sjbur: Same is in CardView I believe

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
```diff
@@ -358,20 +344,32 @@ export class FieldChooserBase extends mixinWidget {
...
         || (e.type === 'keydown' && (e.key === 'Enter' || e.key === ' '));
+      const isAltArrowDown = e.type === 'keydown' && e.altKey && e.key === 'ArrowDown';
+      const isHeaderFilterIconInteraction = isKeyboardOrClick
+        && $(e.target).hasClass(CLASSES.headerFilter);
+
+      if (isAltArrowDown || isHeaderFilterIconInteraction) {
+        if (field.allowFiltering && field.area !== 'data' && !field.groupIndex) {
+          e.preventDefault();
+          this.handleHeaderFilterIconClick(e, field);
+        }
```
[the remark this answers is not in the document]
sjbur: Resolved

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
```diff
@@ -358,20 +349,32 @@ export class FieldChooserBase extends mixinWidget {
...
 
     const handler = (e) => {
-      const shouldHandle = e.type === clickEventName
+      const field: any = $(e.currentTarget).data('field');
+
+      if (!field) {
+        return;
+      }
+
+      const isKeyboardOrClick = e.type === clickEventName
```
Tucchhaa: minor: rename to `isClick`

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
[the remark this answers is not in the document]
Tucchhaa: yes, we don't need to apply this. We have to add aria attributes to field element instead of header filter icon, but this will be done as a separate task

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
```diff
@@ -428,6 +431,31 @@ export class FieldChooserBase extends mixinWidget {
...
+    const $icon = $('<span>').addClass(CLASSES.headerFilter).toggleClass('dx-header-filter-empty', isEmpty);
+
+    let $container = this._getIndicatorContainer($fieldElement);
+
+    if (!$container.length) {
+      const rtlEnabled = this.option('rtlEnabled') as boolean;
+      const indicatorAlignment = rtlEnabled ? 'left' : 'right';
+
+      $container = $('<div>').addClass('dx-column-indicators');
+      this.setAria('role', 'presentation', $container);
```
Tucchhaa: why need role presentation here?

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
sjbur: Done

packages/devextreme/js/__internal/grids/pivot_grid/field_chooser/m_field_chooser_base.ts
sjbur: Initially I copied setAria('role', 'presentation', $container) from old logic of headerFilterMixin. But yes, it makes sense to remove it. Removed

Tucchhaa (approved):

PR #33996, opened 2026-06-15
Color: class refactor and add types

vorobey:


packages/devextreme/js/__internal/m_color.ts
```diff
@@ -527,66 +481,117 @@ function isIntegerBetweenMinAndMax(number, min?, max?) {
...
-  isValidAlpha(a) {
+  static isValidAlpha(a: number): boolean {
     if (isNaN(a) || a < 0 || a > 1 || typeof a !== 'number') {
       return false;
     }
     return true;
-  },
-
-  colorIsInvalid: false,
+  }
```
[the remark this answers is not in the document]
vorobey: Checked all codebase, fixed all consumers.

packages/devextreme/js/__internal/m_color.ts
```diff
@@ -154,11 +154,117 @@ const standardColorNames = {
   yellowgreen: '9acd32',
 };
 
+type ColorParseResult = | [number, number, number]
+  | [number, number, number, number]
+  | [number, number, number, number, [number, number, number]]
+  | [number, number, number, number, null, [number, number, number]];
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+const _round = Math.round;
```
marker-dao: I wouldn't create an alias for the standard `Math.round` method. `Math.round(...)` is immediately clear to any reader, while `_round(...)` adds an unnecessary layer of indirection with no obvious benefit. Plus, you have to disable `@typescript-eslint/naming-convention` here, meaning we’re bypassing a linter rule just for a single local alias. If you need a separate wrapper function with business logic, it’s better to name it explicitly, such as `roundToInteger` or `roundValue`. If you don’t need a wrapper, it’s better to use `Math.round` directly.

marker-dao: Please, also apply to Math.floor, Math.ceil, Array.isArray, etc. in `packages/devextreme/js/__internal/viz/palette.ts`.

packages/devextreme/js/__internal/viz/palette.ts
```diff
@@ -17,9 +17,9 @@
 /* eslint-disable no-else-return */
 /* eslint-disable @typescript-eslint/prefer-optional-chain */
 
-import _Color from '@js/color';
 import { extend } from '@js/core/utils/extend';
 import { isString as _isString } from '@js/core/utils/type';
+import _Color from '@ts/m_color';
```
marker-dao: May we call it just `Color`? Same question for `_isString`.

packages/devextreme/js/__internal/m_color.ts
```diff
@@ -154,11 +154,117 @@ const standardColorNames = {
...
+    case 4: r = vInc; g = vMin; b = v; break;
+    case 5: r = v; g = vMin; b = vDec; break;
+    default:
+      break;
+  }
+
+  return [Math.round(r * 2.55), Math.round(g * 2.55), Math.round(b * 2.55)];
+}
+
+function hslToRgb(h: number, s: number, l: number): [number, number, number] {
```
marker-dao: Have you considered moving the helper functions into a separate helper file? For example, one named `color.utils.ts`/`color.helper.ts` or something like this?

r-farkhutdinov (dismissed):

packages/devextreme/js/__internal/ui/color_box/color_view.ts
```diff
@@ -972,15 +972,15 @@ class ColorView extends Editor<ColorViewProperties> {
   }
 
   _validateHex(hex: string): string {
-    return this._currentColor.isValidHex(hex) ? hex : this._currentColor.toHex();
+    return Color.isValidHex(hex) ? hex : this._currentColor.toHex();
```
marker-dao, line 975: What's the reason of this change? Why do you want to use class method instead of instance?

packages/devextreme/js/__internal/ui/color_box/color_view.ts
```diff
@@ -972,15 +972,15 @@ class ColorView extends Editor<ColorViewProperties> {
...
   }
 
   _validateRgb(): number[] {
     let { value: r } = this._rgbInputs[0].option();
     let { value: g } = this._rgbInputs[1].option();
     let { value: b } = this._rgbInputs[2].option();
 
-    const isInvalidRgb = !this._currentColor.isValidRGB(r, g, b);
+    const isInvalidRgb = !Color.isValidRGB(r, g, b);
```
marker-dao, line 983: Same question.

packages/devextreme/js/__internal/utils/color.helpers.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  return [Math.round(r * 255), Math.round(g * 255), Math.round(b * 255)];
+}
+
+export type ColorParseResult = | [number, number, number]
+  | [number, number, number, number]
+  | [number, number, number, number, [number, number, number]]
+  | [number, number, number, number, null, [number, number, number]];
+
+// array of color definition objects
```
marker-dao: Such comments in the code are not required; we usually only include comments for complex sections, specifying the reason or the user's public ticket number.

packages/devextreme/js/__internal/m_color.ts
```diff
@@ -154,439 +162,117 @@ const standardColorNames = {
...
-        H = 60 * ((b - r) / delta) + 120;
-        break;
-      case b:
-        H = 60 * ((r - g) / delta) + 240;
-        break;
-      default:
-        break;
+  constructor(value?: string) {
+    this.baseColor = value;
+    let color: ColorParseResult | null = null;
```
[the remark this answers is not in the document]
marker-dao: Sounds reasonable.

packages/devextreme/js/__internal/m_color.ts
```diff
@@ -154,439 +162,117 @@ const standardColorNames = {
...
-  },
+  }
 
-  isValidRGB(r, g, b) {
-    if (!isIntegerBetweenMinAndMax(r) || !isIntegerBetweenMinAndMax(g) || !isIntegerBetweenMinAndMax(b)) {
+  static isValidRGB(
+    r: unknown,
+    g: unknown,
+    b: unknown,
+  ): boolean {
```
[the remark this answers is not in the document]
marker-dao: Sounds reasonable.

packages/devextreme/js/__internal/m_color.ts
```diff
@@ -154,439 +162,117 @@ const standardColorNames = {
...
     return true;
-  },
+  }
 
-  isValidAlpha(a) {
-    if (isNaN(a) || a < 0 || a > 1 || typeof a !== 'number') {
+  static isValidAlpha(a: unknown): a is number {
+    if (typeof a !== 'number' || isNaN(a) || a < 0 || a > 1) {
       return false;
     }
```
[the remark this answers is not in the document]
marker-dao: Plus

packages/devextreme/js/__internal/ui/color_box/color_box.ts
```diff
@@ -4,7 +4,7 @@ import $ from '@js/core/renderer';
 import type { DeferredObj } from '@js/core/utils/deferred';
 import type { Properties } from '@js/ui/color_box';
 import type { OptionChanged } from '@ts/core/widget/types';
-import Color from '@ts/m_color';
+import { Color } from '@ts/m_color';
```
marker-dao, line 7: There's no reason to change this line; the Color class is exported by default: https://github.com/vorobey/DevExtreme/blob/f7cc2885c63241519597247b589d70c5bfd0ac39/packages/devextreme/js/__internal/m_color.ts#L295 

But it's up to you.

packages/devextreme/js/__internal/ui/color_box/color_view.ts
vorobey: So, during refactor, I converted all methods inside Color class which aren't use `this` to static methods. I think, that's a right and expected approach and according to our eslint rules. That's why I changed this.
Another approach was - remove arguments from functions like `isValidHex` and return response, based on value, that stored in `Color` instance. But it will lead to more changes in consumer code, probably

packages/devextreme/js/__internal/ui/color_box/color_box.ts
vorobey: I changed it during helpers extraction to other file and didn't return this line to prev state. I think, it's a minor here, let's leave as is

packages/devextreme/js/__internal/ui/color_box/color_view.ts
marker-dao: This approach is possible, but it is rarely used in our codebase. Please try to avoid it in the future.

marker-dao (dismissed):

r-farkhutdinov (approved):

marker-dao (approved):

dmlvr (approved):

PR #34106, opened 2026-06-24
Scheduler - Refactor layout pipeline for auto-height support

bit-byte0:


packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/options/option_manager.ts
```diff
@@ -54,14 +73,22 @@ export class OptionManager {
...
+    const layoutData: WorkspaceLayoutData = this.layoutData ?? this.schedulerStore.getWorkSpace();
+    const panelDOMSize = layoutData.getPanelDOMSize(
       this.options.groupOrientation === 'vertical'
         ? 'regularPanel'
         : panelName,
     );
-
-    return this.cache.memo(`${panelDOMSize.width}.${panelDOMSize.height}.${panelName}`, () => {
+    const groupHeights = layoutData.getGroupHeights?.(panelName);
+    const groupHeightsKey = groupHeights?.join('.') ?? '';
```
sjbur: Since we are dealing with array of pixels here, there is a possibility of such case:

[1.5, 3].join('.') -> '1.5.3'
[1, 5.3].join('.') -> '1.5.3'

In theory its a behavior bug.

I am not sure whether we should overthink this case, but we could use a different separator here, like ',' or we could use stringtify.

What do you think?

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/steps/add_geometry/add_grouping_offset.ts
```diff
@@ -4,6 +4,26 @@ import type {
...
+  defaultHeight: number,
+): number => groupHeights?.[groupIndex] ?? defaultHeight;
+
+const getCumulativeGroupOffset = (
+  groupHeights: number[] | undefined,
+  groupIndex: number,
+  defaultHeight: number,
+): number => {
+  let offset = 0;
+
```
sjbur, line 19: Does it make sense to use cache here as well?

packages/devextreme/js/__internal/scheduler/r1/components/base/date_table_body.tsx
```diff
@@ -63,10 +66,18 @@ export class DateTableBody extends BaseInfernoComponent<DateTableBodyProps> {
                   dateTable.map(({
                     cells,
                     key: rowKey,
-                  }) => (
+                  }) => {
+                    const rowHeight = rowHeights?.[rowIndex];
+                    const rowStyles = rowHeight === undefined
+                      ? undefined
+                      : renderUtils.addHeightToStyle(rowHeight);
+                    rowIndex += 1;
```
sjbur: Increasing rowIndex here in render function by React/Inferno standards is a side effect mutation, which is a bad practice. Can we do it in a different manner?

packages/devextreme/js/__internal/scheduler/shaders/current_time_shader_vertical.ts
```diff
@@ -25,34 +25,47 @@ class VerticalCurrentTimeShader extends CurrentTimeShader {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   renderShader(isHorizontalGroupedWorkSpace: boolean, groupCount: number, cellCount: number): void {
     let shaderHeight = this.getShaderHeight();
-    const maxHeight = this.getShaderMaxHeight();
-    const isSolidShader = shaderHeight > maxHeight;
 
-    if (shaderHeight > maxHeight) {
-      shaderHeight = maxHeight;
-    }
+    if (this.workSpace.isGroupedByDate()) {
```
sjbur, line 29: Maybe we could a test for new logic in current time shader vertical?

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/options/option_manager.ts
bit-byte0: Fixed by serializing the structured cache key with `JSON.stringify`, so fractional pixel values cannot collide

packages/devextreme/js/__internal/scheduler/r1/components/base/date_table_body.tsx
bit-byte0: Removed the mutable `rowIndex` from render and now derive the flattened row index from the group/row indexes

packages/devextreme/js/__internal/scheduler/shaders/current_time_shader_vertical.ts
bit-byte0: Added a focused test for vertical current time shader rendering with per-group max heights

packages/devextreme/js/__internal/scheduler/view_model/generate_view_model/steps/add_geometry/add_grouping_offset.ts
bit-byte0: I thought about this, but I’d prefer to keep the current implementation for now. The cumulative calculation is simple and keeps this helper stateless. We can add cached offsets later if profiling shows this path becomes hot with variable group heights. What do you think?

sjbur: OK, we can leave it as it is

sjbur (approved):

PR #34110, opened 2026-06-24
Scheduler: refactor and remove utils.ts

sjbur:


packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -2203,7 +2202,7 @@ class Scheduler extends SchedulerOptionsBaseWidget {
     targetedAppointment?: SafeAppointment,
   ) {
     if (appointment) {
-      const settings: any = utils.dataAccessors.getAppointmentSettings(element);
+      const settings: any = this._appointments.getAppointmentSettings(element);
```
bit-byte0, line 2205: Are we sure `showAppointmentTooltip(...)` cannot be called when `_newAppointments` is enabled? In that mode `_appointments` is the new `Appointments` component, and it does not have `getAppointmentSettings` or am I wrong?

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
```diff
@@ -1925,7 +1924,7 @@ class Scheduler extends SchedulerOptionsBaseWidget {
 
   // TODO<Appointments>: delete this method when old impl is removed
   getTargetedAppointment(appointment: SafeAppointment, element: dxElementWrapper): TargetedAppointment {
-    const settings = utils.dataAccessors.getAppointmentSettings(element)!;
+    const settings = this._appointments.getAppointmentSettings(element);
```
bit-byte0, line 1927: I think here's the same issue as on line 2205: `getTargetedAppointment(...)` use only the old appointments implementation? `_appointments` can also be the new `Appointments` component, which does not have `getAppointmentSettings`

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
sjbur: Added getAppointmentSettings method to newAppointments as well, according to its implementation

packages/devextreme/js/__internal/scheduler/m_scheduler.ts
sjbur: Added getAppointmentSettings method to newAppointments as well, according to its implementation

bit-byte0 (approved):

PR #34115, opened 2026-06-24
T1331067 - Vertical lines don't work with multiple axes and panes

dmlvr:


packages/devextreme/testing/tests/DevExpress.viz.charts/chart.integration.tests.js
```diff
@@ -2015,6 +2015,43 @@ QUnit.test('Validate Axis on update', function(assert) {
...
+        }
+    });
+
+    try {
+        chart.option('argumentAxis.constantLines', [{ value: new Date(2026, 5, 16) }]);
+        assert.ok(true, 'no error during update');
+    } catch(e) {
+        assert.ok(false, 'error: ' + e.message);
+    }
+});
```
pharret31, line 2048: There's no need to use try-catch, it will hide the actual error stack, you can just write: 

```js
chart.option('argumentAxis.constantLines', [{ value: new Date(2026, 5, 16) }]);
assert.ok(true, 'no error during update');
```

dmlvr: removed try-catch

pharret31 (approved):

PR #34116, opened 2026-06-24
DataGrid: Fix horizontal touch scrolling on freespace row with rowDragging enabled (T1329643)

Alyar666:


packages/devextreme/js/__internal/m_draggable.ts
```diff
@@ -285,6 +285,8 @@ class Draggable extends DOMComponent<Draggable, Properties> {
 
   dragInProgress?: boolean;
 
+  _dragMoveArgs?: any;
```
anna-shakhova: please, add actual type instead of `any`

packages/devextreme/js/__internal/m_draggable.ts
```diff
@@ -790,16 +792,15 @@ class Draggable extends DOMComponent<Draggable, Properties> {
   }
 
   _dragMoveHandler(e) {
```
anna-shakhova: lets add typings to the method, as we touched it

anna-shakhova (dismissed):

packages/devextreme/js/__internal/m_draggable.ts
Alyar666: I've thought about that, but I couldn't find a suitable type. We don't currently have types for drag events. Could you suggest what we should use?”

packages/devextreme/js/__internal/m_draggable.ts
Alyar666: This is the same issue I mentioned above — I couldn't find a suitable type for the `e` parameter. At most, I can specify the method's return type (`void`) and access modifier (`private`).

I think we should create a separate task for proper Draggable/Sortable typing and address it there. For now, I'd prefer to leave it as is.

anna-shakhova: lets start with native DragEvent, we can wrap it inside DxEvent if method required some data from it. It is too small for a separate task

Raushen (dismissed):

anna-shakhova (dismissed):

Raushen (approved):

PR #34170, opened 2026-06-29
DataGrid - fix default types in d.ts

Tucchhaa:


packages/devextreme/js/common/grids.d.ts
```diff
@@ -601,7 +601,6 @@ export interface ColumnBase<TRowData = any> {
   falseText?: string;
   /**
    * @docid GridBaseColumn.filterOperations
-   * @default undefined
```
Tucchhaa, line 604: removed because this is a type of GridBaseColumn which is used in grid.columns[]. There are no default values for .option('column[i].filterOperations') in runtime

packages/devextreme/js/__internal/m_draggable.ts
```diff
@@ -330,16 +330,7 @@ class Draggable extends DOMComponent<Draggable, Properties> {
     return {
       ...super._getDefaultOptions(),
       // @ts-expect-error
-      onDragStart: null,
```
marker-dao, line 333: This is just a question to clarify:

Is this change intentional?

Wouldn't it be better to leave these options in the default options list with the value 'undefined' instead of 'null'? Mainly for ease of search and consistency with the existing options list.

Do we intentionally want to completely remove them from '_getDefaultOptions()'?

Tucchhaa: Yes, this was intentional. For me it was no difference between these two approaches, but now I think that it is better to be a little more explicit. I will bring back to default options such props with `undefined` value

packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -1317,7 +1317,7 @@ export interface dxCardViewOptions<TCardData = unknown, TKey = unknown> extends
      * @type string | Array<any> | Store | DataSource | DataSourceOptions
```
Raushen: Should we add '| undefined' also here?

Tucchhaa: you are right, I have applied it

packages/devextreme/js/__internal/m_draggable.ts
Tucchhaa: Discussed in DM:
if a callback is defined in defaultOptions(), then it should have `@default undefined` and `| undefined`.
If a callback **is not** defined in defaultOptions(), then `@default undefined` and `| undefined` can be removed. 

Since in datagrid&treelist most of the callbacks aren't defined in defaultOptions(), I have removed these from d.ts. But in other components, I kept the doctags as callbacks are defined in defaultOptions()

packages/devextreme-vue/src/data-grid.ts
```diff
@@ -423,7 +423,7 @@ const componentConfig = {
     columnWidth: [String, Number] as PropType<Mode | number>,
     customizeColumns: Function as PropType<((columns: Array<dxDataGridColumn>) => void)>,
     dataRowTemplate: {},
-    dataSource: [Array, Object, String] as PropType<Array<any> | DataSource | DataSourceOptions | null | Store | string | Record<string, any>>,
+    dataSource: [Array, Object, String] as PropType<Array<any> | DataSource | DataSourceOptions | Store | string | Record<string, any>>,
```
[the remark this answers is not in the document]
Tucchhaa: intended BC

marker-dao (approved):

arman-boyakhchyan (approved):

packages/devextreme/js/ui/card_view.d.ts
```diff
@@ -1337,13 +1331,12 @@ export interface dxCardViewOptions<TCardData = unknown, TKey = unknown> extends
...
     /**
      * @docid
-     * @default undefined
      * @type_function_param1 e:object
      * @type_function_param1_field component:this
      * @action
      * @public
      */
-    onDataErrorOccurred?: (e: EventInfo<dxCardView> & DataErrorOccurredInfo) => void;
+    onDataErrorOccurred?: ((e: EventInfo<dxCardView> & DataErrorOccurredInfo) => void);
```
Raushen, line 1339: Brackets here are not required now

Raushen (approved):

PR #34178, opened 2026-06-30
PivotGrid: Add aria-colindex / aria-rowindex to cells

sjbur:


packages/devextreme/js/__internal/grids/pivot_grid/area_item/m_area_item.ts
```diff
@@ -206,23 +257,30 @@ abstract class AreaItem {
           }
 
           cellText = this._getCellText(cell, encodeHtml);
-        }
-
-        const span = domAdapter.createElement('span');
 
-        if (isDefined(cell.wordWrapEnabled)) {
-          span.style.whiteSpace = cell.wordWrapEnabled ? 'normal' : 'nowrap';
+          if (isDefined(cell.wordWrapEnabled)) {
```
sjbur, line 261: moved isDefined(cell.wordWrapEnabled) inside if (cell) check because if cell is undefined, we get type error in runtime

aleksei-semikozov (approved):

PR #34180, opened 2026-06-30
DataGrid - cellRender triggers a second request on loading an initial page (T1326786)

markallenramirez:


markallenramirez:
postponed

PR #34187, opened 2026-07-01
Scheduler - Refactor Appointments Collection - Resizing

bit-byte0:
### What
Added appointment resizing (timed + all-day) to the new appointments collection behind the `_newAppointments` flag with Esc-to-cancel, focus handling, group-bounded drag, and rollback on cancel/failure

### How
Extracted the resize math into small pure, unit-tested helpers; added opt-in `onCancelByEsc` to the `Resizable` widget; wired it into the new grid appointment with a thin `scheduler.ts` handler that computes the new dates (timed vs all-day) and updates the appointment, focusing it on resize-start and snapping it back if the update is cancelled/fails


packages/devextreme/js/__internal/scheduler/scheduler.ts
```diff
@@ -2011,6 +2033,237 @@ class Scheduler extends SchedulerOptionsBaseWidget {
...
+      return this.timeZoneCalculator.createDate(adapter.startDate, 'toGrid', startDateTimeZone);
+    }
+
+    return appointmentStartDate;
+  }
+
+  private onAppointmentResizeEnd(e: AppointmentResizeEvent): void {
+    const $element = $(e.element);
+    const settings = this._appointments
+      .getAppointmentSettings($element) as AppointmentItemViewModel;
```
sjbur, line 2179: settings can return undefined here. We should add guard
if (!settings) {
return;
}

to avoid type error in runtime

packages/devextreme/js/__internal/scheduler/appointments_new/resizing/get_resized_dates.ts
```diff
@@ -0,0 +1,163 @@
...
+  const usesHorizontalHandles = !isVerticalDirection || isAllDay;
+
+  if (usesHorizontalHandles) {
+    return rtlEnabled ? handles.right : handles.left;
+  }
+
+  return handles.top;
+};
+
+const correctEndDateByDelta = (
```
sjbur, line 46: correctStartDateByDelta/correctEndDateByDelta — are they almost the same here?

packages/devextreme/js/__internal/scheduler/scheduler.ts
```diff
@@ -2011,6 +2033,237 @@ class Scheduler extends SchedulerOptionsBaseWidget {
...
+
+  private getEndResizeStartDate(
+    e: AppointmentResizeEvent,
+    rawAppointment: SafeAppointment,
+    appointmentStartDate: Date,
+  ): Date {
+    const adapter = new AppointmentAdapter(rawAppointment, this._dataAccessors);
+    const { startDateTimeZone, isRecurrent } = adapter;
+
+    if (!e.handles.top && !isRecurrent) {
```
sjbur, line 2169: in old _getEndResizeAppointmentStartDate there was a guard check for appointment not being allDay. In new method we don't have it. Is it by a choice?

packages/devextreme/js/__internal/scheduler/scheduler.ts
bit-byte0: Applied the guard

packages/devextreme/js/__internal/scheduler/appointments_new/resizing/get_resized_dates.ts
bit-byte0: They do look almost identical, but they actually work in opposite directions: `correctStartDateByDelta` moves the start of the appointment backward (subtracts the delta, walks days back, clamps to `startDayHour`), while `correctEndDateByDelta` moves the end forward (adds the delta, walks days ahead, clamps to `endDayHour`). Legacy kept them as two separate functions too. I did try to picture merging them into one, but you'd end up passing a bunch of "which direction?" flags, so I left them split on purpose

packages/devextreme/js/__internal/scheduler/scheduler.ts
bit-byte0: Yeah, that's on purpose. In the old code, all-day and regular appointments went through the same `getDateRange`, so `_getEndResizeAppointmentStartDate` had to check `!isAllDay` right there to skip all-day ones. Now the split happens earlier: `onAppointmentResizeEnd` sends all-day appointments to `getResizedAllDayDateRange` and regular ones to `getResizedTimedDateRange`, and this method only gets called from the regular (timed) path, so an all-day appointment never reaches it and that's why the `!isAllDay` check isn't needed here anymore

sjbur (dismissed):

packages/devextreme/js/__internal/ui/resizable/resizable.test.ts
```diff
@@ -0,0 +1,111 @@
+import {
```
Raushen: Let's follow the main flow and place Jest tests into __tests__ folder:
`packages/devextreme/js/__internal/ui/resizable/__tests__/resizable.test.ts‎`

packages/devextreme/js/__internal/ui/resizable/resizable.test.ts
```diff
@@ -0,0 +1,111 @@
...
+import $ from '@js/core/renderer';
+import eventsEngine from '@ts/events/core/m_events_engine';
+
+import Resizable, { type ResizableProperties } from './resizable';
+
+const HANDLE_BOTTOM_CLASS = 'dx-resizable-handle-bottom';
+
+const instances: Resizable[] = [];
+
+const createResizable = (options: Partial<ResizableProperties> = {}): {
```
Raushen: Please create corresponding ResizableModel in the following folder:
`packages/devextreme/js/__internal/ui/__tests__/__mock__/model`

packages/devextreme/js/__internal/ui/resizable/resizable.test.ts
```diff
@@ -0,0 +1,111 @@
...
+  // @ts-expect-error DOMComponent constructor is not typed for direct instantiation
+  const instance = new Resizable($element, options);
+  const $handle = $element.find(`.${HANDLE_BOTTOM_CLASS}`);
+
+  instances.push(instance);
+
+  return { instance, $element, $handle };
+};
+
+const startResize = ($handle: ReturnType<typeof $>): void => {
```
Raushen: Let every such method be a public method of ResizableModel

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -263,6 +283,10 @@ class Resizable extends DOMComponent<Resizable, ResizableProperties> {
     this.$element().toggleClass(RESIZABLE_RESIZING_CLASS, value);
   }
 
+  _isResizing(): boolean {
+    return this.$element().hasClass(RESIZABLE_RESIZING_CLASS);
```
Raushen: The component state should not relate to the DOM state. We need to use business logic. For example we can use internal field
`this._isResizing = false;`
We can set it to true when starting resizing and to false when ending or cancelling.

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -204,12 +217,18 @@ class Resizable extends DOMComponent<Resizable, ResizableProperties> {
         immediate: true,
       });
     });
+
+    if (this.option('onCancelByEsc')) {
+      eventsEngine.on(this.$element(), KEYDOWN_EVENT_NAME, this._keydownHandler.bind(this));
```
Raushen: We need to subscribe the handler every time. The guard
`if (this.option('onCancelByEsc'))`
should work inside of the handler.

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -240,6 +259,7 @@ class Resizable extends DOMComponent<Resizable, ResizableProperties> {
       return;
     }
 
+    this._resizeCanceled = false;
```
Raushen: Could you please check if we can reattach drag handlers while the cancelling instead of creating this field?

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -685,6 +718,37 @@ class Resizable extends DOMComponent<Resizable, ResizableProperties> {
...
+  }
+
+  _cancelResize(e: DxEvent<KeyboardEvent>): void {
+    this._resizeCanceled = true;
+    this._restoreSize();
+
+    this._toggleResizingClass(false);
+
+    this._resizeCancelAction?.({
+      event: e,
```
Raushen: You can obtain an instance of the component and using its options instead of passing arguments.

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -85,6 +88,10 @@ export interface ResizableProperties extends Properties {
   step?: string;
 
   roundStepValue?: boolean;
+
+  onCancelByEsc?: boolean;
```
Raushen: We can improve the name. The prefix on- uses for handlers only. We could name it `cancelOnEscape`.

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -85,6 +88,10 @@ export interface ResizableProperties extends Properties {
...
+
+  onResizeCancel?: (e: Record<string, unknown>) => void;
```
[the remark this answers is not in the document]
Raushen: So, Copilot is right here.

Raushen: I do not see usings of onResizeCancel handler in the scheduler. Do you really need it?

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -685,6 +718,37 @@ class Resizable extends DOMComponent<Resizable, ResizableProperties> {
...
+
+  _restoreSize(): void {
+    this.option({
+      width: this._elementSize.width,
+      height: this._elementSize.height,
+    });
+
+    move(this.$element(), this._elementLocation);
+
+    triggerResizeEvent(this.$element());
```
Raushen, line 737: Please check if it would be better to use internal _triggerResizeAction method instead.

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -721,8 +785,13 @@ class Resizable extends DOMComponent<Resizable, ResizableProperties> {
       case 'onResize':
       case 'onResizeStart':
       case 'onResizeEnd':
+      case 'onResizeCancel':
         this._renderActions();
         break;
+      case 'onCancelByEsc':
```
Raushen: This case should disappear after refactoring `_attachEventHandlers()` method.

Raushen (changes requested):

packages/devextreme/js/__internal/ui/resizable/resizable.test.ts
bit-byte0: moved to `ui/resizable/__tests__/resizable.test.ts`

packages/devextreme/js/__internal/ui/resizable/resizable.test.ts
bit-byte0: added `ResizableModel`

packages/devextreme/js/__internal/ui/resizable/resizable.test.ts
bit-byte0: `startResize` / `moveResize` / `endResize` / `pressEscape` / `isResizing` / `hasResizingClass` are now public methods on `ResizableModel`, and the test drives everything through it

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: replaced the class-based check with an internal `_isResizing` boolean, set on resize start and cleared on end/cancel. The CSS class is now only presentational

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: the keydown handler is now always attached, and the `cancelOnEscape` check moved inside `_keydownHandler`

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: dropped the `_resizeCanceled` field; `_cancelResize` now reattaches the drag handlers so the trailing dragend from the canceled gesture is discarded

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: the payload belonged to `onResizeCancel`, which is now removed, so there are no arguments being passed anymore

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: it originally was created to make the widget's event API "symmetric" (`onResizeStart` / `onResizeEnd` / `onResizeCancel`), mirroring how the drag side has events. It felt like the "proper" complete API. Anyways it was removed

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: I suggest keeping `triggerResizeEvent` here: `_triggerResizeAction` needs a **DragEvent** and also fires the `onResize` action, but a cancel comes from a keydown (no drag event) and firing `onResize` on a cancel would be misleading. `triggerResizeEvent` gives the layout notification without the spurious `onResize`. If you'd prefer to reuse` _triggerResizeAction`, I can extract a small shared helper / make its event arg optional.

Let me know

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: since the keydown handler is now always attached, that `_optionChanged` case is gone

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: renamed

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: Small correction to my earlier reply: the detach/attach logic is gone as you asked, but I had to keep a bare no-op case `cancelOnEscape: break`; The `optionChanged` widget test (_DevExpress.ui.widgets/optionChanged_) requires every option to be handled in `_optionChanged` without falling through to super, otherwise it fails with "Option `cancelOnEscape` is not processed". So the case stays, but only as a no-op alongside area/step/etc

packages/devextreme/js/__internal/ui/__tests__/__mock__/model/resizable.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export class ResizableModel {
+  constructor(protected readonly root: HTMLElement) {}
+
+  public getInstance(): Resizable {
+    return Resizable.getInstance(this.root);
+  }
+
+  public isResizing(): boolean {
+    return this.getInstance()._isResizing;
```
Raushen: Page Object Model should rely on the DOM. So it is needed to check CLASSES.resizing here:
```
public isResizing(): boolean {
    return this.root.classList.contains(CLASSES.resizing);
}
```

Also models should not have access to private class fields of instances.

packages/devextreme/js/__internal/ui/__tests__/__mock__/model/resizable.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  public hasResizingClass(): boolean {
```
Raushen: So, this is a duplicate of the isResizing() method.

packages/devextreme/js/__internal/ui/__tests__/__mock__/model/resizable.ts
```diff
@@ -0,0 +1,49 @@
...
+  public endResize(): void {
+    this.triggerOnHandle('dxdragend');
+  }
+
+  public pressEscape(): void {
+    eventsEngine.trigger(this.root, { type: 'keydown', key: 'Escape' });
+  }
+
+  private getHandle(): dxElementWrapper {
+    return this.getInstance().$element().find(`.${CLASSES.handleBottom}`);
```
Raushen: Let the method work through the this.root, not through the instance.

packages/devextreme/js/__internal/ui/resizable/__tests__/resizable.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const model = new ResizableModel(instance.$element().get(0) as HTMLElement);
+  models.push(model);
+
+  return model;
+};
+
+describe('Resizable cancelOnEscape', () => {
+  afterEach(() => {
+    models.forEach((model) => model.getInstance().dispose());
```
Raushen: It is easier to keep instances:
`resizables.forEach((instanse) => instance.dispose());`

packages/devextreme/js/__internal/ui/resizable/resizable.ts
Raushen: Ok, let's keep it as is.

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -204,12 +212,17 @@ class Resizable extends DOMComponent<Resizable, ResizableProperties> {
         immediate: true,
       });
     });
+
+    eventsEngine.off(this.$element(), KEYDOWN_EVENT_NAME);
```
Raushen: Do we need to detach events first? In which case there are already attached keydown events?

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -682,7 +696,38 @@ class Resizable extends DOMComponent<Resizable, ResizableProperties> {
...
+  }
+
+  _cancelResize(): void {
+    this._isResizing = false;
+    this._restoreSize();
     this._toggleResizingClass(false);
+
+    this._detachEventHandlers();
+    this._attachEventHandlers();
+  }
```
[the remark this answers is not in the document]
bit-byte0: Same pattern as **Draggable** esc-cancel (`_keydownEscapeHandler: detach` → revert → reattach, synchronous). Detach dispose gesture emitter; new emitter never saw pointerdown → no dxdrag/dxdragend from canceled gesture

packages/devextreme/js/__internal/scheduler/appointments_new/resizing/get_appointment_date_range.ts
```diff
@@ -0,0 +1,177 @@
...
+      )
+      : leftCell;
+
+    return {
+      startDate: firstCell.startDate,
+      endDate: firstCell.startDate > appointment.endDate
+        ? firstCell.startDate
+        : appointment.endDate,
+    };
+  }
```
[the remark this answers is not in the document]
bit-byte0: Faithful copy of legacy: _m_core.ts_ `getDateRangeHorizontalRTL` pass same global index to `getCellData`, no modulo. Refactor reproduce legacy 1:1. If real multi-row RTL bug exist in legacy too, deserve separate issue, not silent behavior change in refactor PR

packages/devextreme/js/__internal/ui/__tests__/__mock__/model/resizable.ts
bit-byte0: `isResizing()` now checks the `dx-resizable-resizing` class, the model no longer touches instance fields

packages/devextreme/js/__internal/ui/__tests__/__mock__/model/resizable.ts
bit-byte0: `isResizing()` is the single DOM-based check now

packages/devextreme/js/__internal/ui/__tests__/__mock__/model/resizable.ts
bit-byte0: the model works through `this.root.querySelector` now, no instance access left

packages/devextreme/js/__internal/ui/resizable/__tests__/resizable.test.ts
bit-byte0: the test keeps a resizables array and disposes instances directly

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: As far as I understand we need it on repeated renders. `_attachEventHandlers` re-runs on invalidation (e.g. a handles option change). Drag handlers sit on the handle elements, which `_clean()` removes and recreates, so they never stack. The keydown handler sits on the root element, which survives re-renders. Without the off, each re-render adds another subscription. The QUnit Memory Leaks suite (eventsOnRefresh) caught exactly this in CI. Can move the unbind into `_detachEventHandlers` / `_clean()` instead if you prefer

packages/devextreme/js/__internal/ui/resizable/resizable.ts
```diff
@@ -682,7 +696,38 @@ class Resizable extends DOMComponent<Resizable, ResizableProperties> {
       handles: this._movingSides,
     });
 
+    this._isResizing = false;
+    this._toggleResizingClass(false);
+  }
+
```
Raushen, line 701: Could you please add the comment here like this one:
```
  /**
   * Cancels an active resize when Escape is pressed and `cancelOnEscape` is enabled.
   *
   * The handler is subscribed to the root element, so keydown reaches it only
   * while focus is inside the widget. The root is not focusable by itself —
   * the host must make it focusable: Scheduler appointments have `tabindex`,
   * and clicking a resize handle keeps focus within the appointment element.
   */
```

packages/devextreme/js/__internal/ui/resizable/resizable.ts
Raushen: Yes, please, let's keep detaching inside the lifecycle and move it to the _clean() method.

bit-byte0: Resolved by moving the unbind to `_clean()` as you suggested. `_attachEventHandlers` now only subscribes

packages/devextreme/js/__internal/ui/resizable/resizable.ts
bit-byte0: added

Raushen (approved):

sjbur (approved):

PR #34221, opened 2026-07-07
Grids - Stabilize tests (Part 2 & 3)

Tucchhaa:


e2e/testcafe-devextreme/tests/cardView/columnSortable/functional.ts
```diff
@@ -36,6 +37,8 @@ fixture.disablePageReloads`CardView - ColumnSortable.Functional`
     } else {
       await t.expect(draggingElement.exists).notOk();
     }
+
+    await triggerDragEnd(columnElement);
```
Tucchhaa: This test was stable, but page reloads are disabled in this fixture, so to prevent any leakage between tests I have added this.

e2e/testcafe-devextreme/tests/cardView/columnSortable/functional.ts
```diff
@@ -124,7 +129,6 @@ fixture.disablePageReloads`CardView - ColumnSortable.Functional`
         { dataField: 'Column 1', visible: false },
         { dataField: 'Column 2' },
         { dataField: 'Column 3' },
-        { dataField: 'Column 4' },
```
Tucchhaa, line 127: This matrix tests iterates through [0, 1, 2] for gapIndex. With 4 columns, testcases for gapIndex==1 and gapIndex==2 basically test the same thing: drag from column chooser to the middle of header panel.

With 3 columns, gapIndex==2 now checks the case when column is dragged to the end of header panel

e2e/testcafe-devextreme/tests/common/treeList/toast.ts
```diff
@@ -7,22 +7,39 @@ import { testScreenshot } from '../../../helpers/themeUtils';
...
-test('Toast should be visible after calling and should be not visible after default display time', async (t) => {
+test('Toast should be visible after calling', async (t) => {
   const treeList = new TreeList('#container');
   const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
   await t
     .expect(treeList.isReady())
     .ok();
 
-  await treeList.apiShowErrorToast();
+  await treeList.apiShowErrorToast({ displayTime: 100000 });
```
Tucchhaa: toast could disappear before the screenshot is taken, which can lead to instability. So I have added more display time

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
```diff
@@ -41,16 +43,12 @@ const deleteGlobalVariables = ClientFunction((): void => {
...
 
-test.meta({ unstable: true })('DataGrid should send an AI request for rendered rows after scrolling without changing the page index', async (t) => {
+test('DataGrid should send an AI request for rendered rows after scrolling without changing the page index', async (t) => {
   // arrange
   const dataGrid = new DataGrid(DATA_GRID_SELECTOR);
 
-  // assert
-  await t
-    .expect(dataGrid.getLoadPanel().isVisible())
-    .ok();
```
Tucchhaa, line 55: race condition: load panel can hide before this check runs. The test name doesn't suggest that load panel has to be checked, so I have removed this expect

e2e/testcafe-devextreme/tests/dataGrid/common/focus/focusEvents/newRows_T1162227.ts
```diff
@@ -245,11 +245,15 @@ test('It should not fire row events if focusedRowEnabled: false', async (t) => {
...
 
-test.meta({ unstable: true })('It should fire rowChanged event on initialization if focusedRowKey options is set', async (t) => {
+test('It should fire rowChanged event on initialization if focusedRowKey options is set', async (t) => {
+  const dataGrid = new DataGrid(GRID_SELECTOR);
+
   const expectedRowFocusChanged: FocusRowChangedData[] = [
     [1],
   ];
 
+  await t.expect(dataGrid.getDataRow(1).element).ok();
```
Tucchhaa: wait for the row to appear before checking callback results

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -2970,7 +2970,8 @@ test('New mode. A cell should be focused when the PageDow/Up key is pressed (T89
 
   // act
   await t
-    .pressKey('pagedown');
+    .pressKey('pagedown')
+    .expect(dataGrid.isReady()).ok();
```
Tucchhaa: pagedown/pageup cause dataSource load, so it's needed to wait for it to finish

e2e/testcafe-devextreme/tests/dataGrid/common/scrolling.ts
```diff
@@ -1714,7 +1715,7 @@ test.meta({ unstable: true })('Restoring focus on re-rendering should be done wi
 
   await t.click(dataGrid.getHeaders().getHeaderRow(0).getHeaderCell(19).element);
 
-  await dataGrid.scrollBy(t, { left: 0 });
```
Tucchhaa, line 1717: scrollBy accepts displacement and not the exact position. The ticket suggests that dataGrid has to be scrolled to left, but `.scrollBy(t, { left: 0 });` has no effect

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -167,9 +167,7 @@ test('Navigation via the Tab key should work when cellRender/cellComponent is us
     // eslint-disable-next-line no-underscore-dangle
     dataGrid._getTemplate = () => ({
       render(options) {
-        setTimeout(() => {
```
Tucchhaa, line 170: timeouts can affect other tests, since page reloads are disabled

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -3693,70 +3691,6 @@ test('Window should not be scrolled after clicking on freespace row (T1104035)',
   height: 1500,
 }));
 
-test('edit => scroll => command, should not result in grid scrolling back to edit', async (t) => {
```
Tucchhaa, line 3694: this test written strangely + this behavior won't reproduce in modern browsers, cuz when element is focused, the browser automatically scrolls to it

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -6269,8 +6203,9 @@ test('The last cell should be focused after changing the page size (T1063530)',
     await resetFocusedEventsTestData();
 
     // act
-    await t.pressKey('ctrl+end');
-    await t.wait(100);
+    await t
+      .pressKey('ctrl+end')
+      .expect(dataGrid.isReady()).ok();
```
Tucchhaa: ctrl+end causes datasource load, so need to wait for it

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -6403,7 +6344,7 @@ test('The last cell should be focused after changing the page size (T1063530)',
     })();
   });
 
-  test.meta({ unstable: true })(`Focus events should be called when pressing the Ctrl + End key when infinite scrolling is enabled (scrolling.useNative = ${useNativeScrolling})`, async (t) => {
```
Tucchhaa, line 6406: There's a race condition in the datagrid's code on Ctrl+End in infinite scrolling:
(a) the scroll-driven focus application (_handleScroll → focusFirstOrLastCell), and
(b) the infinite-load append render, which re-renders/settles the rows but skips re-focusing.

I suggest to fix this test separately

e2e/testcafe-devextreme/tests/cardView/columnSortable/functional.ts
```diff
@@ -102,18 +107,18 @@ fixture.disablePageReloads`CardView - ColumnSortable.Functional`
 
 [0, 1].forEach((columnIndex) => {
   [0, 1, 2].forEach((gapIndex) => {
-    test.meta({ unstable: true })(`drag from columnChooser to headerPanel: from index ${columnIndex} to index ${gapIndex}`, async (t) => {
+    test(`drag from columnChooser to headerPanel: from index ${columnIndex} to index ${gapIndex}`, async (t) => {
       const cardView = new CardView('#container');
       await cardView.apiShowColumnChooser();
```
Alyar666, line 119: Shouldn't we check isReady here? Also, shouldn't we verify that the Column Chooser exists before accessing it?

e2e/testcafe-devextreme/tests/cardView/columnSortable/utils.ts
```diff
@@ -79,24 +79,26 @@ export const dragToHeaderPanel = async (
...
   if (gapIndex < columnsNum) {
     const insertBeforeColumn = headers.getHeaderItemNth(gapIndex).element;
+    const { width } = await insertBeforeColumn.boundingClientRect;
 
     await t.dragToElement(
       columnElement,
       insertBeforeColumn,
       {
-        destinationOffsetX: 5,
+        destinationOffsetX: -(Math.floor(width) + 3), // 5px left of the left edge
```
Alyar666: > // 5px left of the left edge

5px or 3px ?

e2e/testcafe-devextreme/tests/cardView/columnSortable/utils.ts
```diff
@@ -79,24 +79,26 @@ export const dragToHeaderPanel = async (
...
   } else {
     const insertAfterColumn = headers.getHeaderItemNth(columnsNum - 1).element;
+    const { width } = await insertAfterColumn.boundingClientRect;
 
     await t.dragToElement(
       columnElement,
       insertAfterColumn,
       {
-        destinationOffsetX: -5,
+        destinationOffsetX: (Math.floor(width) + 3), // 5px right of the right edge
```
Alyar666: > // 5px right of the right edge

5px or 3px?

e2e/testcafe-devextreme/tests/common/treeList/toast.ts
Alyar666: Is the toast destroyed after the test completes?

e2e/testcafe-devextreme/tests/cardView/columnSortable/functional.ts
anna-shakhova: Check, please, do we need `desablePageReload` fixture here or it can be omitted here

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
Alyar666: How is that possible if the AI request is still in progress?

Alyar666: I think we need this check because we want to verify that the load panel is initially displayed while the AI request is in progress. This behavior is also verified later in the test.

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
```diff
@@ -41,16 +43,12 @@ const deleteGlobalVariables = ClientFunction((): void => {
...
-
   // act
+  await t.expect(isAIRequestPending()).ok();
```
Alyar666: Should this really be in the act section?

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
```diff
@@ -76,6 +74,7 @@ test.meta({ unstable: true })('DataGrid should send an AI request for rendered r
     .ok();
 
   // act
+  await t.expect(isAIRequestPending()).ok();
```
Alyar666: Should this really be in the act section?

e2e/testcafe-devextreme/tests/common/treeList/toast.ts
```diff
@@ -7,22 +7,39 @@ import { testScreenshot } from '../../../helpers/themeUtils';
...
+  await t
+    .expect(treeList.isReady())
+    .ok();
+
+  await treeList.apiShowErrorToast({ displayTime: 1000 });
+  await t
+    .expect(treeList.getToast().exists).ok();
+
+  await t
+    .wait(1000)
```
anna-shakhova: lets increase wait time a bit, show/hide toast is made with promise and setTimeout, and may take a little more than displayTime

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
```diff
@@ -144,16 +143,12 @@ test.meta({ unstable: true })('DataGrid should send an AI request for rendered r
     await deleteGlobalVariables();
   });
 
-test.meta({ unstable: true })('DataGrid should send an AI request for rendered rows after scrolling with changing the page index', async (t) => {
+test('DataGrid should send an AI request for rendered rows after scrolling with changing the page index', async (t) => {
```
Alyar666, line 153: The same questions apply as in the previous test (see above).

e2e/testcafe-devextreme/tests/dataGrid/common/editing/functional.ts
```diff
@@ -2316,7 +2329,8 @@ test('Cells should be focused correctly on click when cell editing mode is used
   // act
   await t
     .typeText(dataGrid.getDataCell(0, 0).getEditor().element, '1')
-    .click(dataGrid.getDataCell(1, 0).getEditor().element);
+    .click(dataGrid.getDataCell(1, 0).getEditor().element)
+    .expect(dataGrid.isReady()).ok();
```
Alyar666: Should this really be in the act section?

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
Alyar666: Should this really be in the act section?

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
anna-shakhova: again, can we remove `disablePageReload` fixture here?

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -6112,7 +6048,9 @@ test('The last cell should be focused after changing the page size (T1063530)',
     await resetFocusedEventsTestData();
 
     // act
-    await t.pressKey('ctrl+end');
+    await t
+      .pressKey('ctrl+end')
+      .wait(500);
```
Alyar666: Can we avoid using wait here? For example, could we replace it with a check that a specific cell exists?

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -6189,7 +6127,9 @@ test('The last cell should be focused after changing the page size (T1063530)',
...
```
Alyar666: Can we avoid using wait here? For example, could we replace it with a check that a specific cell exists?

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
anna-shakhova: Usage of `t.wait` is prohibited in testcafe tests and we want to remove all existing. Do we have some other assertion that can help here (ready, some element is visible or not, etc.) 

Some for other added `t.wait`

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -6508,7 +6449,9 @@ test('The last cell should be focused after changing the page size (T1063530)',
...
```
Alyar666: Can we avoid using wait here? For example, could we replace it with a check that a specific cell exists?

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
```diff
@@ -6596,7 +6539,9 @@ test('The last cell should be focused after changing the page size (T1063530)',
...
```
Alyar666: Can we avoid using wait here? For example, could we replace it with a check that a specific cell exists?

e2e/testcafe-devextreme/tests/dataGrid/common/headerPanel.ts
```diff
@@ -87,10 +87,16 @@ test('Drop-down window should be positioned correctly after resizing the toolbar
 // visual: fluent.blue.light
 // visual: fluent.blue.dark
 
-test.meta({ unstable: true })('Disabled toolbar buttons are not grayed out in Material themes (T1217416)', async (t) => {
+test('Disabled toolbar buttons should be grayed out in Material themes (T1217416)', async (t) => {
```
anna-shakhova, line 90: let add testing in Material theme and remove comments above

e2e/testcafe-devextreme/tests/dataGrid/common/toast.ts
```diff
@@ -8,20 +8,39 @@ fixture.disablePageReloads`Toasts in DataGrid`.page(
...
-test.meta({ unstable: true })('Toast should be visible after calling and should be not visible after default display time', async (t) => {
+test('Toast should be visible after calling and should be not visible after default display time', async (t) => {
   const dataGrid = new DataGrid('#container');
   const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
   await t
     .expect(dataGrid.isReady())
     .ok();
 
-  await dataGrid.apiShowErrorToast();
+  await dataGrid.apiShowErrorToast({ displayTime: 100000 });
```
Alyar666: Is the toast destroyed after the test completes?

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/error_handling/m_error_handling.ts
```diff
@@ -158,8 +159,8 @@ export class ErrorHandlingController extends modules.ViewController {
     }
   }
 
-  public showToastError(message: string) {
-    this._toastViewController.showToast(message, { type: 'error' });
+  public showToastError(message: string, options?: ToastProperties) {
```
anna-shakhova: I don't like the idea to extend public method arguments for testing purposes only. I would prefer creating anothe api helper in testcafe model to manage toast behavior with smth like
``` javascript
(window as any).DevExpress.ui.dxToast.defaultOptions(options)
```

e2e/testcafe-devextreme/tests/cardView/columnSortable/functional.ts
Tucchhaa: I think we need `disablePageReloads` in this file, cuz there are a lot of matrix tests. I added this clean-up line for prevention, because it's hard to detect the actual reason of flakiness: locally everything runs smoothly, test starts failing only when CI is slow

e2e/testcafe-devextreme/tests/cardView/columnSortable/utils.ts
Tucchhaa: Fixed. It's 3px because distance between columns in fluent is 6px

e2e/testcafe-devextreme/tests/cardView/columnSortable/utils.ts
Tucchhaa: fixed

e2e/testcafe-devextreme/tests/common/treeList/toast.ts
Tucchhaa: ok, thanks, I have applied

e2e/testcafe-devextreme/tests/common/treeList/toast.ts
Tucchhaa: I have removed `disablePageReloads`, since there are only 2 tests in the file. That will guarantee that there's no leakage between the tests

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
Tucchhaa: moved to the assert section above

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
Tucchhaa: @Alyar666, you're right here. I have brought back the loadPanel visibility check. I think the reason of flakiness was that there was no `await t.expect(isAIRequestPending()).ok();`.  I will run CI several times to verify that

@anna-shakhova this test file, doesn't have `disablePageReloads`

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
Tucchhaa: fixed

e2e/testcafe-devextreme/tests/dataGrid/common/aiColumn/virtualScrolling.functional.ts
Tucchhaa: fixed

e2e/testcafe-devextreme/tests/dataGrid/common/editing/functional.ts
Tucchhaa: fixed

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
Tucchhaa: fixed

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
Tucchhaa: After taking a deeper look, all waits in the modified tests are useless, after keypress there is a check on testcafe Selector (i.e. `.expect(dataGrid.getDataCell(19, 14).element.focused)`), which means that Testcafe will poll cell's state until the condition is true. So testcafe already does the wait for us.

I will replace waits with `dataGrid.isReady`, but that won't fix flakiness completely (I already tried to use `dataGrid.isReady`). I suspect the fix should be on the datagrid's side

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
Tucchhaa: removed all waits from this file

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
Tucchhaa: removed all waits from this file

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
Tucchhaa: I removed all `t.wait` from this file, except this one: https://github.com/DevExpress/DevExtreme/blob/c225896b025eabd39a3681bfe9ac334f9c068100/e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts#L3285
Because:
1. that test is not flaky
2. it's hard to wait for the moment when the filter is applied

e2e/testcafe-devextreme/tests/dataGrid/common/headerPanel.ts
Tucchhaa: We have a lot of occurrences of `// visual: `
If you dont mind, I suggest to apply this change in another task, cuz this PR already has a lot of changes

e2e/testcafe-devextreme/tests/dataGrid/common/toast.ts
Tucchhaa: removed `disablePageReloads` from this file

packages/devextreme/js/__internal/grids/grid_core/error_handling/m_error_handling.ts
Tucchhaa: Nice suggestion, I have applied it

e2e/testcafe-devextreme/tests/cardView/columnSortable/functional.ts
Tucchhaa: isReady won't help here, because it indicates whether dataController finished loading data: https://github.com/DevExpress/DevExtreme/blob/4f7410d48fcc01ab8ff6153b709f2828c47ed7ec/packages/devextreme/js/__internal/grids/grid_core/data_controller/m_data_controller.ts#L389

In this test we don't care whether the data was loaded or not. Explicit wait for column chooser is also not needed, because there's an implement check:

`const columnElement = getColumnItem(cardView, columnIndex, 'columnChooser');`. Column element is a testcafe Selector, so testcafe will wait for the element to appear before doing anything with it

e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts
Tucchhaa: I removed all `t.wait` from this file, except this one: https://github.com/DevExpress/DevExtreme/blob/c225896b025eabd39a3681bfe9ac334f9c068100/e2e/testcafe-devextreme/tests/dataGrid/common/keyboardNavigation/keyboardNavigation.functional.ts#L3285
Because:
1. that test is not flaky
2. it's hard to wait for the moment when the filter is applied

e2e/testcafe-devextreme/tests/cardView/columnChooser/functional.ts
Tucchhaa, line 59: Test is unstable in select mode: https://github.com/DevExpress/DevExtreme/actions/runs/29251945672/job/86861954473?pr=34221. I have tried different approaches to fix it, but none of them worked. So I decided to postpone its fix. I have added this test to the card for fixing tests

anna-shakhova (dismissed):

Alyar666 (dismissed):

anna-shakhova (dismissed):

anna-shakhova (approved):

PR #34247, opened 2026-07-09
Scheduler — Refactor scheduler.d.ts file

sjbur:


packages/devextreme-vue/src/scheduler.ts
```diff
@@ -1006,7 +1005,7 @@ const DxItemConfig = {
     location: String as PropType<ToolbarItemLocation>,
     menuItemTemplate: {},
     name: String as PropType<string | FormPredefinedButtonItem | SchedulerPredefinedToolbarItem>,
-    options: Object as PropType<DateNavigatorItemProperties | Record<string, any> | dxButtonGroupOptions>,
+    options: Object as PropType<DateNavigatorItemProperties | Record<string, any>>,
```
sjbur, line 1008: DateNavigatorItemProperties extends dxButtonGroupOptions, so there is no need for dxButtonGroupOptions here

mpreyskurantov (changes requested):
blocked,
should be a feature branch from the upstream remote (not a fork),
in the feature/* name format

PR #34253, opened 2026-07-09
T1328053 - Draggable - The drag action starts after closing Ctrl + click context menu in Safari/Chrome on MacOS

Raushen:


packages/devextreme/js/__internal/events/utils/index.ts
```diff
@@ -57,6 +57,14 @@ const LEGACY_KEY_CODES = {
   18: 'alt',
```
Raushen, line 57: This file contains only refactoring, without fix.

packages/devextreme/js/__internal/events/gesture/m_emitter.gesture.ts
```diff
@@ -87,7 +87,12 @@ const GestureEmitter = Emitter.inherit({
   },
 
   start(e) {
-    if (e._needSkipEvent || needSkipEvent(e)) {
+    // T1328053: macOS Ctrl+click opens the system context menu. Kept out of needSkipEvent()
+    // because importing m_devices into that low-level events util forces `new Devices()` into
+    // early module init and breaks init order (resizeCallbacks stops firing).
+    const isMacContextMenuClick = isMouseEvent(e) && Boolean(e.ctrlKey) && devices.real().mac;
```
anna-shakhova, line 93: wouldn't it better to move this logic into `needSkipEvent`? It is already covers contextmenu case by right button click

Raushen: Actually no, because it breaks module initialization order and some tests fail. I've left this comment exactly for this situation:
```
Kept out of needSkipEvent()
    // because importing m_devices into that low-level events util forces `new Devices()` into
    // early module init and breaks init order (resizeCallbacks stops firing).
```

anna-shakhova (approved):

pharret31 (approved):

PR #34254, opened 2026-07-09
Scheduler - Refactor Appointments Collection - Support appointmentDragging options

bit-byte0:
### What
Added `appointmentDragging` support to the refactored **Scheduler** appointments (`_newAppointments`) - its callbacks and options, plus drag and drop between two schedulers and with an external component

### How
The new drag controller forwards the `appointmentDragging` callbacks and options to the draggable and resolves cross-component drops by the drop position

### DIFF
--- a/packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
+++ b/packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
@@ -1,9 +1,22 @@
+import eventsEngine from '@js/common/core/events/core/events_engine';
+import { enter as dragEventEnter, leave as dragEventLeave } from '@js/common/core/events/drag';
+import { addNamespace } from '@js/common/core/events/utils/index';
 import type { dxElementWrapper } from '@js/core/renderer';
 import $ from '@js/core/renderer';
+import { extend } from '@js/core/utils/extend';
+import { getWindow } from '@js/core/utils/window';
 import type {
   DragEndEvent, DragMoveEvent, DragStartEvent, DragTemplateData, Properties as DraggableProperties,
 } from '@js/ui/draggable';
-import type { Appointment } from '@js/ui/scheduler';
+import type {
+  Appointment,
+  AppointmentDraggingAddEvent,
+  AppointmentDraggingEndEvent,
+  AppointmentDraggingMoveEvent,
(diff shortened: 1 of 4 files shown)


packages/devextreme/js/__internal/scheduler/appointment_drag_controller.ts
```diff
@@ -139,59 +199,114 @@ export class AppointmentDragController {
...
+    this.draggedItemData = null;
+
+    if (!draggedItemData) {
       this.removeDraggingClasses($(e.itemElement));
       return;
     }
 
-    const isSameCell = this.$initialCell?.is(this.$highlightedCell) ?? false;
-    const isSameScheduler = this.$highlightedCell.closest(e.fromComponent.$element()).length > 0;
+    const isSameComponent = e.fromComponent === e.toComponent;
```
sjbur, line 280: Replacing the old DOM-containment check (this.$highlightedCell.closest(e.fromComponent.$element()).length > 0) with e.fromComponent === e.toComponent drops the guard that verified $highlightedCell was still attached to the live DOM. If the workspace re-renders the date table while a drag is in progress (e.g. virtual scrolling repainting rows during autoScroll), this.$highlightedCell can end up referencing a detached node.

Suggestion — re-validate that $highlightedCell is still connected before using it, treating a detached cell the same as "nothing highlighted" (the existing !this.$highlightedCell branch already handles that safely):

    if (this.$highlightedCell && !this.$highlightedCell.get(0)?.isConnected) {
      this.$highlightedCell = null;
    }

    const isSameComponent = e.fromComponent === e.toComponent;

m_appointment_collection.ts:430 already uses the same !el.isConnected check, so this stays consistent with the rest of the codebase.) Might be worth applying the same guard to $initialCell, and/or in onDragMove too, so a re-render between move events doesn't leave a dangling reference around longer than necessary.

bit-byte0: restored the `.closest(e.fromComponent.$element())` containment check (which also re-establishes the detached-node guard) and added an explicit `isConnected` guard for `$initialCell`

sjbur (approved):

aleksei-semikozov (approved):

PR #34290, opened 2026-07-14
Devextreme: move to nx executors bundle-config tasks from gulp pipeline

vorobey:


packages/devextreme/package.json
```diff
@@ -88,6 +88,7 @@
     "@typescript-eslint/parser": "catalog:",
     "axe-core": "catalog:",
     "babel-plugin-inferno": "6.8.5",
+    "chokidar": "5.0.0",
```
vorobey: @GoodDayForSurf  @ajivanyandev  guys, what do you think, maybe it is better to place on root level? Cuz devextreme-scss uses chokidar also

GoodDayForSurf (approved):

PR #34293, opened 2026-07-14
Grids: fix unstable editing tests

anna-shakhova:

### DIFF
--- a/e2e/testcafe-devextreme/tests/dataGrid/common/editing/editing.functional_matrix.ts
+++ b/e2e/testcafe-devextreme/tests/dataGrid/common/editing/editing.functional_matrix.ts
@@ -167,6 +167,8 @@ const checkCellFocused = async (
   }
 
   await t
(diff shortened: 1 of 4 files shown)


e2e/testcafe-devextreme/tests/dataGrid/common/editing/editing.functional_matrix.ts
```diff
@@ -167,6 +167,8 @@ const checkCellFocused = async (
   }
 
   await t
+    .expect(editor?.element.exists)
+    .ok()
     .expect(editor?.element.focused)
     .eql(true);
```
Tucchhaa, line 173: Afaik `.expect(editor?.element.focused)` will already make testcafe to wait for the element to appear and be focused.

Are you sure that `expect(editor?.element.exists)` fixes instability?

Tucchhaa (approved):

PR #34299, opened 2026-07-14
CardView - cardHeader.items.render does not work in React (T1328152)

markallenramirez:


apps/react/public/js/app/bundle.js
Alyar666, line 1: This is an unnecessary file. Please remove it.

apps/react/public/js/app/bundle.js.map
Alyar666, line 1: This is an unnecessary file. Please remove it.

packages/devextreme/js/__internal/grids/new/card_view/main_view.tsx
```diff
@@ -107,6 +107,9 @@ export class MainView extends View<MainViewProps> {
     rtlEnabled: this.options.oneWay('rtlEnabled').value,
     disabled: this.options.oneWay('disabled').value,
     templatesRenderAsynchronously: this.options.oneWay('templatesRenderAsynchronously').value,
+    integrationOptions: {
```
Alyar666: I suggest memoize the `integrationOptions` object inside the `config` computed.

**Why:** `config` re-evaluates whenever any of its dependencies changes (`rtlEnabled`, `disabled`, `templatesRenderAsynchronously`), and each re-evaluation creates a fresh `integrationOptions` object literal. `InfernoWrapper.updateComponentOptions` compares props by reference, so every unrelated config change makes it call `option('integrationOptions', …)` on all nested widgets (toolbar, popup, form, etc.).

**How:**
```ts
private readonly integrationOptions = computed(() => ({
  templates: this.options.oneWay('integrationOptions.templates').value,
}));

private readonly config = computed(() => ({
  rtlEnabled: this.options.oneWay('rtlEnabled').value,
  disabled: this.options.oneWay('disabled').value,
  templatesRenderAsynchronously: this.options.oneWay('templatesRenderAsynchronously').value,
  integrationOptions: this.integrationOptions.value,
}));
```

packages/devextreme/js/__internal/grids/new/grid_core/core/config_context.ts
```diff
@@ -5,10 +5,12 @@ export interface Config {
   rtlEnabled: WidgetProperties['rtlEnabled'];
   disabled: WidgetProperties['disabled'];
   templatesRenderAsynchronously: unknown; // TODO: Properties['templatesRenderAsynchronously'];
+  integrationOptions: WidgetProperties['integrationOptions'];
```
Alyar666: I suggest narrowing the Config.integrationOptions type to match what is actually being forwarded:
`integrationOptions: Pick<WidgetIntegrationOptions, 'templates'> | undefined;`

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/new/grid_core/core/config_context.ts
markallenramirez: NonNullable is added to resolve an error (see [TS playground](https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgOrACYHMJgApQD2ADtGMBAM7IDeAUMo8gPTPIB0nDTokWUccoRAB5YkJCUA-AC5kAJQgJCUDAB5KYKKCwAaZAFcQAaxCEA7iAB8AbjoBfOnV7R4SZAGFhMYFlrdGVg4uJmQXfkFgYTEJSjk8YARjNXRsXAISMgpKAG0AcnCBCRioyTyAXX08yABbYgAbQSo8q2QAH0MQDAgfEAgMO1DCyOjxUsoAJnjE5IA5YVmDesaAI3qIFMwcfCJSKHIqfOHiseFKCqsq2oam89aOo27e-rt7IA)).

Alyar666 (approved):

PR #34308, opened 2026-07-14
Migrate gulp js-bundles-watch to NX

GoodDayForSurf:


packages/nx-infra-plugin/src/executors/bundle/bundle.impl.ts
```diff
@@ -110,16 +128,96 @@ function runWebpack(webpack: typeof import('webpack'), config: Configuration): P
...
+          logger.error(collectWebpackErrorMessages(stats));
+          return;
+        }
+
+        logWebpackWarnings(stats);
+
+        postBuildChain = postBuildChain
+          .then(() => onSuccess(stats))
+          .then(() => {
+            logger.info('bundle watch: rebuild complete');
```
vorobey, line 205: minor: i propose use verbose here, not info

vorobey (approved):

PR #34325, opened 2026-07-15
DataGrid: Refactor ResizingController._synchronizeColumns

nightskylark:
`ResizingController._synchronizeColumns` and`m_utils.ts`were refactored for readability and type-safety ahead of the fix, without changing behavior:

-   **`ResizingController`**:
    -   Replaced the untyped`_maxWidth: any`field with a small`_maxWidth: MaxWidthController`helper object (`isModified`,`set()`,`clear()`) that encapsulates reading/writing the element's`maxWidth`CSS, instead of manually checking truthiness and mutating the DOM inline in`_synchronizeColumns`.
    -   Extracted the "temporarily switch to best-fit mode, measure, restore focus" logic out of`_synchronizeColumns`into a new`_enableTemporaryBestFitMode()`method that returns a cleanup closure, replacing the previous`resetBestFitMode`boolean flag plus an inline restore-focus block.
    -   Extracted the local`normalizeWidthsByExpandColumns`closure into a proper private method`_normalizeWidthsByExpandColumns(resultWidths, visibleColumns)`, simplified with`findIndex`instead of two separate`each`loops.
    -   Simplified`needBestFit`/`hasMinWidth`computation from imperative loops with early-exit flags into single`.some()`expressions, and narrowed`resultWidths`to`(number | string | undefined)[]`, scoped inside the`deferUpdate`callback instead of the outer closure.

This refactor made`_synchronizeColumns`easier to reason about and was a prerequisite for isolating and fixing the`visibleWidth`/`width`staleness bug (T1329677) in the same best-fit measurement code path.


packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -478,6 +465,21 @@ export class ResizingController extends modules.ViewController {
     return freeWidth / columnCountWithoutWidth;
   }
 
+  private readonly _normalizeWidthsByExpandColumns = (resultWidths, visibleColumns): void => {
```
Alyar666, line 468: It's not clear why this method is marked as readonly. In this case, we're explicitly preventing it from being overridden without any clear reason. That goes against the Grid's architecture, which is built around inheritance and @extended overrides.

Also, the method parameters are untyped.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -478,6 +465,21 @@ export class ResizingController extends modules.ViewController {
...
+    const expandColumnIndex = visibleColumns.findIndex((column) => column.type === 'groupExpand');
```
Alyar666, line 469: Now the Grid uses the width of the first expand column it encounters. Previously, it used the width of the last expand column.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -478,6 +465,21 @@ export class ResizingController extends modules.ViewController {
...
+    const expandColumnWidth = resultWidths[expandColumnIndex];
+
+    if (!isDefined(expandColumnWidth)) {
```
Alyar666, line 472: We no longer ignore the value 0. Is this intentional?

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -122,6 +126,26 @@ export class ResizingController extends modules.ViewController {
 
   public resizeCompleted!: Callback;
 
+  private readonly _maxWidth: MaxWidthController = {
```
Alyar666, line 129: This approach is a bit unusual for the Grid. Having two regular private methods, _setMaxWidth and _clearMaxWidth, along with a boolean flag, would be more consistent with the existing codebase.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -122,6 +126,26 @@ export class ResizingController extends modules.ViewController {
...
+    isModified: false,
+    set: (value): void => {
+      const $element = this.component.$element();
```
Alyar666, line 132: Shouldn't we also check whether $element exists here, just like we do in the clear method?

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
nightskylark: You're right that the arrow-property form was a mistake, though readonly wasn't the actual problem — it doesn't block overriding. The real issue is that an instance arrow property lives on the instance rather than the prototype, so super._normalizeWidthsByExpandColumns(...) wouldn't be reachable from an extender. Fixed: it's a plain prototype method now, and the parameters are typed (ColumnWidth[] / Column[]).

I kept it private rather than protected, though. This method didn't exist before — it was a local closure inside _synchronizeColumns, so nothing could override it and no extension point is being removed. grid_core marks extension points explicitly (242 @extended annotations), which only carries meaning if non-annotated members are genuinely internal. If an extender ever needs it, widening private → protected + @extended is a one-line, deliberate change. Happy to flip it now if you'd rather have it open by default.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
nightskylark: You\re right, this was an unintended behavior change and I've reverted it to "last" and covered it by test.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
nightskylark: Not intentional — reverted to the original truthy check.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
nightskylark: Agreed, switched to _setMaxWidth / _clearMaxWidth plus a private _isMaxWidthSet flag.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
nightskylark: I looked into this and the check isn't needed there. $element() returns this._$element, which is always assigned in ctor (dom_component.ts:109/515), and .css(name, value) goes through setCss, which starts with if (!this[0]?.style) return; (renderer_base.ts:439). The original code called $element.css('maxWidth', ...) unguarded for the same reason; the get(0) check existed only in the clear path because that one dereferences $element[0].style directly, which would throw.

That said, your pointed at a real bug one line over: the old clear() returned early on a missing element without resetting isModified, leaving the flag stuck at true with no style applied. In the new _clearMaxWidth the flag is reset unconditionally and only the DOM write is guarded.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -37,6 +39,8 @@ const GROUP_ROW_SELECTOR = 'tr.dx-group-row';
 
 const HIDDEN_COLUMNS_WIDTH = 'adaptiveHidden';
 
+type ColumnWidth = number | string | undefined;
```
Alyar666: We usually keep types and interfaces in a separate `types.ts` file.

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
```diff
@@ -127,6 +129,8 @@ export class ResizingController extends modules.ViewController {
 
   public resizeCompleted!: Callback;
 
+  private _isMaxWidthSet = false;
```
Alyar666: **Minor**: An underscore in the name.

packages/devextreme/js/__internal/grids/grid_core/m_utils.ts
```diff
@@ -72,6 +72,11 @@ export function isDateType(dataType: string | undefined): boolean {
   return dataType === 'date' || dataType === 'datetime';
 }
 
+export interface SelectionRange {
```
Alyar666: We usually keep types and interfaces in a separate types.ts file.

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
nightskylark: Done

packages/devextreme/js/__internal/grids/grid_core/views/m_grid_view.ts
nightskylark: Done

packages/devextreme/js/__internal/grids/grid_core/m_utils.ts
nightskylark: Done

Alyar666 (approved):

PR #34335, opened 2026-07-16
Implement keyboard navigation improvements in HtmlEditor

Raushen:


vladaskorohodova (dismissed):

packages/devextreme/js/__internal/ui/html_editor/__tests__/focus_escape.test.ts
```diff
@@ -0,0 +1,102 @@
+import {
```
pharret31, line 1: Why a new jest tests file? we have all htmleditor related tests here in QUnit:
packages/devextreme/testing/tests/DevExpress.ui.widgets.htmlEditor/htmlEditorParts

Raushen: It is Jest test, not qUnit. Jest - it is modern testing platform. It allow us use Typescript and create fast, stable unit tests.

pharret31: I know, but all our tests are written using QUnit and starting writing new tests in Jest might lead to a mess, I guess this should be resolved separately: either we move all our tests to Jest or we continue writing QUnits.

Raushen: Actually no, we already have a lot of Jest tests and there is no any mess. Jest tests have their own structure, qUnit - its own, TestCafe - its own. For example we do not need move all TestCafe tests to qUnit or vice versa. We just choose more convenient way to write tests.

pharret31: You're right that we don't need to move TestCafe tests to QUnit, but I think that's a bit different because they solve different problems. TestCafe is for E2E tests, while QUnit and Jest are both unit test frameworks.

I actually agree that Jest is a more modern option. My concern is consistency. All existing HTMLEditor unit tests are in the htmlEditorParts folder, and they're all (as well as other components in Components area) written in QUnit. If we start adding new unit tests in Jest alongside QUnit for the same component, it becomes less obvious where to look for tests and which framework should be used for new ones.

That's why I see two reasonable options:

- decide that we're migrating HTMLEditor (and other component's) unit tests to Jest (or even Vitest, since it's even better and more modern test framework and we have plans to move to Vite as far as I know), and gradually move existing tests using boyscout rule as well as for other components;
- or continue writing HTMLEditor unit tests in QUnit until we make that migration decision.

Without such a decision, I think mixing two unit testing frameworks within the same area of the codebase will make the test suite less consistent over time, even if both frameworks coexist in the repository.

Although I think we should consider this approach but it's up to you :)

pharret31 (dismissed):

pharret31 (dismissed):

arman-boyakhchyan (dismissed):

pharret31 (approved):

arman-boyakhchyan (approved):

PR #34339, opened 2026-07-16
Gulp to nx: Migrate transpile watch and transpile tests. add new watch helper

ajivanyandev:


packages/devextreme/build/gulp/transpile-config.js
```diff
@@ -3,7 +3,6 @@
 const common = {
     plugins: [
         ['babel-plugin-inferno', { 'imports': true }],
-        ['@babel/plugin-transform-object-rest-spread', { loose: true }],
```
vorobey, line 6: Why have youd decided removing this option?

packages/devextreme/project.json
```diff
@@ -507,6 +565,33 @@
...
     },
+    "build:transpile:watch": {
+      "executor": "nx:run-commands",
+      "cache": false,
+      "options": {
+        "commands": [
+          "pnpm nx build:ts:internal:watch devextreme",
+          "pnpm nx build:cjs:watch devextreme",
+          "pnpm nx build:cjs:watch devextreme -c production",
+          "pnpm nx build:cjs:internal:watch devextreme",
```
vorobey, line 575: this and `build:cjs:internal:watch -c production` both depends on `build:ts:internal`. Does spawning those jobs in parallel works correctly?

packages/devextreme/build/gulp/transpile-config.js
ajivanyandev: Because it didn't resolve (couldn't find module from packages/devextreme's own deps) and was dead weight anyway — grep confirmed zero object-rest/spread syntax anywhere in js/**, so it never transformed anything. Verified via byte-parity test that removing it changes no build output.

packages/devextreme/project.json
ajivanyandev: It was functioning correctly, but running it twice. fixed in the "review fixes" commit, thx!

vorobey (approved):

GoodDayForSurf (approved):

packages/devextreme/build/gulp/transpile-config.js
ajivanyandev: re-reviewed this. turned out I was having a cache issue, and there was actually a difference in byte comparison caused by this. Took back by a separate commit as a fix

PR #34342, opened 2026-07-16
Scheduler — Refactor scheduler.d.ts file

sjbur:


mpreyskurantov (changes requested):
check docs generation

PR #34361, opened 2026-07-17
Scheduler/PivotGrid: align @default tags with option types in public d.ts

aleksei-semikozov:


aleksei-semikozov:
Superseded by the feature-branch flow: same commit re-submitted from upstream feature/26_2_pivotgrid-scheduler-jsdoc-defaults

PR #34364, opened 2026-07-17
Grids stabilize tests after toolbar KBN changes

Tucchhaa:


e2e/testcafe-devextreme/tests/dataGrid/common/columnReordering/visual.ts
```diff
@@ -4,97 +4,125 @@ import url from '../../../../helpers/getPageUrl';
...
+      },
+    ],
+    allowColumnReordering: true,
+  });
+}).after(async () => {
+  await MouseUpEvents.enable(MouseAction.dragToOffset);
+  await removeStylesheetRulesFromPage();
+});
 
 test('HeaderRow should be highlighted when dragging column with allowColumnReordering=false', async (t) => {
```
Tucchhaa, line 62: this test was not flaky, but its test case is similar to the flaky one, so I have added styles to prevent its flakiness in the future

e2e/testcafe-devextreme/tests/dataGrid/common/columnReordering/visual.ts
```diff
@@ -4,97 +4,125 @@ import url from '../../../../helpers/getPageUrl';
...
+      },
+    ],
+    allowColumnReordering: false,
+  });
+}).after(async () => {
+  await MouseUpEvents.enable(MouseAction.dragToOffset);
+  await removeStylesheetRulesFromPage();
+});
 
 test('The group separator should not appear when dragging a grouped column to the same position', async (t) => {
```
Tucchhaa, line 110: this test was not flaky, but its test case is similar to the flaky one, so I have added styles to prevent its flakiness in the future

e2e/testcafe-devextreme/tests/dataGrid/common/columnReordering/visual.ts
```diff
@@ -4,97 +4,125 @@ import url from '../../../../helpers/getPageUrl';
...
-      width: 200,
-    }, {
-      dataField: 'field4',
-      width: 200,
+  /*
+    Overrides the toolbar items' :focus-visible outline,
+    because in e2e tests flaky outline would appear
+    */
+  await insertStylesheetRulesToPage(`
+    .dx-group-panel-item[tabindex]:focus-visible:not(.dx-toolbar-item) {
```
bit-byte0: is this `:focus-visible` outline reproducible for real users, or does it only appear with TestCafe's synthetic drag events? if it can happen in real usage, it probably belongs in the product SCSS rather than only in the tests

bit-byte0: is `[tabindex]` intended here instead of `[tabindex="0"]` like in the toolbar rule? matching items with `tabindex="-1"` might catch something we don't want

e2e/testcafe-devextreme/tests/dataGrid/common/columnReordering/visual.ts
```diff
@@ -4,97 +4,125 @@ import url from '../../../../helpers/getPageUrl';
...
-      dataField: 'field3',
-      width: 200,
-    }, {
-      dataField: 'field4',
-      width: 200,
+  /*
+    Overrides the toolbar items' :focus-visible outline,
+    because in e2e tests flaky outline would appear
+    */
+  await insertStylesheetRulesToPage(`
```
bit-byte0: this override block is repeated 4 times across this file and *withGrouping.ts* - could it be extracted into a shared constant in *domUtils.ts*?

e2e/testcafe-devextreme/tests/dataGrid/sticky/common/withGrouping.ts
```diff
@@ -181,23 +182,34 @@ test.meta({ browserSize: [900, 800] })('Sticky columns with grouping - overflow
...
   const { takeScreenshot, compareResults } = createScreenshotsComparer(t);
   const dataGrid = new DataGrid('#container');
 
   await t.expect(dataGrid.isReady()).ok();
-  await t.wait(300);
 
   await t.drag(dataGrid.getGroupPanel().getHeader(0).element, 200, 35);
-  await t.wait(200);
+
+  await t.expect(dataGrid.getHeaders().getHeaderRow(0).isHighlighted()).ok();
```
bit-byte0, line 194: is the outline override still needed now that the explicit `isHighlighted()` wait is in place, or does the flaky outline still appear without it?

e2e/testcafe-devextreme/tests/dataGrid/common/columnReordering/visual.ts
Tucchhaa: outline doesn't appear in real usage only in testcafe

e2e/testcafe-devextreme/tests/dataGrid/sticky/common/withGrouping.ts
Tucchhaa: Yes, it's still needed. I have added `isHighlighted()` to remove `t.wait`, but it doesn't resolve flakiness

e2e/testcafe-devextreme/tests/dataGrid/common/columnReordering/visual.ts
Tucchhaa: Make sense, I have created a helper func

e2e/testcafe-devextreme/tests/dataGrid/common/columnReordering/visual.ts
Tucchhaa: I think it doesn't matter, the tests only need `:focus-visible` selector's outline to be suppressed, because datagrid has its own styles for `:focus` selector. I have removed `[tabindex]` from the selector.

anna-shakhova (approved):

bit-byte0 (approved):

PR #34369, opened 2026-07-17
DataGrid - Fix flaky material drag-icons screenshot test (T1303892 follow-up)

bit-byte0:


bit-byte0:
Note: not a byte-identical pick of #34367 - 25_2 renders the material selection palette differently, so the etalon is taken from a 25_2 CI run

anna-shakhova (approved):

PR #34372, opened 2026-07-17
DataGrid: Fix header borders being cut off above the vertical scrollbar (T1306973)

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
```diff
@@ -1486,15 +1486,16 @@ export class ColumnsView extends ColumnStateMixin(modules.View) {
...
-    $element && $element.css({
-      paddingLeft: rtlEnabled ? width : '',
-      paddingRight: !rtlEnabled ? width : '',
-    });
+    $element && $element
+      .toggleClass(this.addWidgetPrefix(CLASSES.scrollerSpacing), !!width)
+      .css({
+        paddingLeft: rtlEnabled ? width : '',
+        paddingRight: !rtlEnabled ? width : '',
+      });
```
anna-shakhova: Lets replace it with inline start/end paddings

```suggestion
      .css({
        paddingInlineStart: '', 
        paddingInlineEnd: width,
      });
```

e2e/testcafe-devextreme/tests/dataGrid/common/scrolling.ts
```diff
@@ -363,6 +363,69 @@ test('Scroll position after grouping when RTL (T388508)', async (t) => {
...
+      takeScreenshot,
+      `grid-headers-scroller-spacing-border${rtlEnabled ? '-rtl' : ''}.png`,
+      { element: '#container' },
+    );
+
+    await t
+      .expect(compareResults.isValid())
+      .ok(compareResults.errorMessages());
+  }).before(async () => {
+    await insertStylesheetRulesToPage(`
```
anna-shakhova, line 398: let's remove scrollbar customization - it adds logic, but not useful here

Alyar666: On modern Chrome running on Linux, scrollbars are overlay-based and effectively have a width of 0px. As a result, there is no reserved scrollbar gutter, setScrollerSpacing receives 0, and no padding is applied. In this environment, the test would become a no-op: the screenshot would capture a grid where the fix is never actually exercised.

packages/devextreme/js/__internal/grids/grid_core/views/m_columns_view.ts
Alyar666: This behavior was already implemented; I didn't change it. However, let's see if we can adjust it, provided the change is relatively inexpensive.

anna-shakhova (approved):

PR #34382, opened 2026-07-20
T1332109 - DataGrid throws the "Cannot read properties of null (reading 'remoteOperations')" if multiple data operations are executed and autoNavigateToFocusedRow is true

bit-byte0:
### What
**DataGrid** no longer throws a `Cannot read properties of null (reading 'remoteOperations')` error when its data source is disposed or replaced while a focused-row navigation is still resolving (grids using `remoteOperations` together with `autoNavigateToFocusedRow`)

### How
The asynchronous row-index lookup now verifies that its data source is still the current one before continuing, and resolves as "row not found" instead of reading a disposed or swapped source

### DIFF
--- a/packages/devextreme/js/__internal/grids/data_grid/focus/m_focus.ts
+++ b/packages/devextreme/js/__internal/grids/data_grid/focus/m_focus.ts
@@ -88,9 +88,8 @@ const data = (Base: DataControllerBase) => class FocusDataControllerExtender ext
   }
 
   private _calculateGlobalRowIndexByGroupedData(key) {
-    const that = this;
-    const dataSource = that._dataSource;
-    const filter = that._generateFilterByKey(key);
+    const dataSource = this._dataSource;
+    const filter = this._generateFilterByKey(key);
     // @ts-expect-error
     const deferred = new Deferred();
     const isGroupKey = Array.isArray(key);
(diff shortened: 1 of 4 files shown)


packages/devextreme/js/__internal/grids/data_grid/focus/m_focus.ts
anna-shakhova, line 1: Lets cover similar case in TreeView too: packages/devextreme/js/__internal/grids/tree_list/m_focus.ts

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
```diff
@@ -693,6 +693,10 @@ const data = (Base: ModuleType<DataController>) => class FocusDataControllerExte
       skip: 0,
       take: 1,
     }).done((data) => {
+      if (that._dataSource !== dataSource) {
```
anna-shakhova: nitpick: we follow boys scout rule - if you touched smth and it is possible to make small refactoring here - do it.
In this PR we can remove `that` alias for this in touched methods. If it is also clear how to type arguments and function return let's do it also. Only for touched methods, not all file

packages/devextreme/js/__internal/grids/data_grid/focus/m_focus.ts
bit-byte0: done

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
bit-byte0: Added the guard to _tree_list/m_focus.ts_ for parity, but it seems like **TreeList** doesn't actually crash here. `keyOf()` and `pageSize()` are null-safe, so a cleared source just resolves to `-1` instead of throwing. A repro test stays green with or without the guard, so it's basically a no-op safety net. Left it in anyway

anna-shakhova (approved):

PR #34404, opened 2026-07-21
T1331971: DataGrid - Filter Panel and Filter Row values are not synchronized when applyFilter="onClick"

bit-byte0:
### What
**DataGrid** now keeps the filter row value in sync with the filter panel when `filterRow.applyFilter` is `onClick`

### How
The filter synchronization resets the buffered filter row value so the editor reflects the latest applied `filterValue` instead of a stale pending value

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/filter/__tests__/m_filter_row.integration.test.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/filter/__tests__/m_filter_row.integration.test.ts
@@ -5,6 +5,7 @@ import {
 } from '@jest/globals';
 import type { EditorPreparingEvent } from '@js/ui/data_grid';
 import { TagBoxModel } from '@ts/ui/__tests__/__mock__/model/tag_box';
+import { TextBoxModel } from '@ts/ui/__tests__/__mock__/model/textbox';
 
 import {
   afterTest,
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_sync.ts
```diff
@@ -134,11 +134,15 @@ const updateFilterRowCondition = function (columnsController, column, condition)
     filterRowOptions = {
       filterValue,
       selectedFilterOperation,
+      bufferedFilterValue: undefined,
```
Alyar666, line 137: Does it make sense to reset these properties here? They will be restored later in the editor handler anyway (at least bufferedFilterValue will be).

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/filter/__tests__/m_filter_row.integration.test.ts
```diff
@@ -68,4 +69,30 @@ describe('FilterRow', () => {
       expect(tagBox.getTags()).toHaveLength(2);
     });
   });
+
+  describe('sync with filter panel when applyFilter is onClick (T1331971)', () => {
+    it('should update the filter row editor when the condition is changed via the filter panel', async () => {
```
Alyar666, line 74: Let's add a test for filter reset.

packages/devextreme/js/__internal/grids/grid_core/filter/m_filter_sync.ts
bit-byte0: it's actually necessary. The reset is what lets the editor read the freshly-synced `filterValue` during that same `columnOption` change; without it `getColumnFilterValue` returns the stale buffered value and the editor never updates - this is our bug. Yes, the editor handler re-writes `bufferedFilterValue` afterwards, but to the **_new_** `filterValue` and `bufferedSelectedFilterOperation` is not restored by that handler, so its reset is clearly needed. Actually the regression test covers this exact path

packages/devextreme/js/__internal/grids/grid_core/filter/__tests__/m_filter_row.integration.test.ts
bit-byte0: added

Alyar666 (dismissed):

Alyar666 (approved):

PR #34409, opened 2026-07-21
DropDownButton – FilterBuilder editor does not close after clicking outside it when it is rendered via dropDownContentTemplate (T1300811)

markallenramirez:


packages/devextreme/js/__internal/filter_builder/m_filter_builder.ts
```diff
@@ -765,10 +767,15 @@ class FilterBuilder extends Widget<any> {
     this._documentClickHandler = documentClickHandler;
   }
 
-  _isFocusOnEditorParts($editor, target?) {
+  _isFocusOnEditorParts($editor: dxElementWrapper, target?: HTMLElement): boolean {
     const activeElement = target || domAdapter.getActiveElement();
-    return $(activeElement).closest($editor.children()).length
-            || $(activeElement).closest('.dx-dropdowneditor-overlay').length;
+    const isFocusOnEditor = $(activeElement).closest($editor.children()).length > 0;
```
Tucchhaa: I suggest to add an early return after this line:
```js
if (isFocusOnEditor) {
  return true;
}
```

Tucchhaa: Also, I suggest to rewrite this line to (remove `.children()`):
```js
const isFocusOnEditor = $(activeElement).closest($editor).length > 0;
```

packages/devextreme/js/__internal/filter_builder/m_filter_builder.ts
```diff
@@ -765,10 +767,15 @@ class FilterBuilder extends Widget<any> {
...
+    const overlay = $(activeElement).closest(`.${DROPDOWN_EDITOR_OVERLAY_CLASS}`);
+    const overlayID = overlay.find(`.${POPUP_CONTENT_CLASS}`).attr('id');
```
Tucchhaa: How about fetch popupContent element directly:
```js
    const $popupContent = $(activeElement).closest(`.${POPUP_CONTENT_CLASS}`);

    if ($popupContent.length === 0) {
      return false;
    }

    const overlayID = $popupContent.attr('id');
```

markallenramirez: The popupContent element does not occupy the entire overlay.
Screenshot 2026-07-28 125736

In this case, the following line will not find popupContent when clicking outside it even if you clicked inside the overlay area.
```
const $popupContent = $(activeElement).closest(`.${POPUP_CONTENT_CLASS}`);
```

Tucchhaa (approved):

PR #34311, opened 2026-07-15
Popover: support content reading on opening

EugeniyKiyashko:

PR #34413, opened 2026-07-21
Popover: focus management for dialog role

dmlvr:


packages/devextreme/js/__internal/ui/popover/popover.ts
```diff
@@ -894,6 +952,9 @@ class Popover<
   }
 
   _dispose(): void {
+    if (this.option('visible') && this._getEffectiveAriaRole() === 'dialog') {
```
pharret31: avoid such using of options, they'll be untyped

const { visible } = this.option();

if (visible && ...

dmlvr: fixed

packages/devextreme/js/__internal/ui/overlay/overlay.ts
```diff
@@ -992,14 +992,22 @@ class Overlay<
       const $currentElement = $elements?.eq(i) ?? null;
       const $reverseElement = $elements?.eq(elementsCount - i) ?? null;
 
-      // @ts-expect-error is should can get function as callback
-      if (!$first && $currentElement.is(selectors.tabbable)) {
-        $first = $currentElement;
+      if (!$first) {
+        // @ts-expect-error is should can get function as callback
+        const isCurrentTabbable = $currentElement?.is(selectors.tabbable);
+        const isCurrentNotOverlay = $currentElement?.get(0) !== this._$content?.get(0);
```
Raushen: Can we include overlay selector in the previous one:
```
$currentElement
    .is(selectors.tabbable)
    .not(overlayId);
```
?

packages/devextreme/js/__internal/ui/overlay/overlay.ts
```diff
@@ -992,14 +992,22 @@ class Overlay<
...
+        }
       }
 
-      // @ts-expect-error is should can get function as callback
-      if (!$last && $reverseElement.is(selectors.tabbable)) {
-        $last = $reverseElement;
+      if (!$last) {
+        // @ts-expect-error is should can get function as callback
+        const isReverseTabbable = $reverseElement?.is(selectors.tabbable);
+        const isReverseNotOverlay = $reverseElement?.get(0) !== this._$content?.get(0);
```
Raushen: The same comment about selector

packages/devextreme/js/__internal/ui/popover/popover.ts
```diff
@@ -266,9 +267,51 @@ class Popover<
...
+
+    return null;
+  }
+
+  _focusTarget(): dxElementWrapper {
+    return this._getFocusTarget() ?? this.$overlayContent();
+  }
+
+  _restoreTargetFocus(): void {
+    const $targets = this._getAriaDescriptionTargets();
```
Raushen: We can have only one target. We can apply changes from [this comment](https://github.com/DevExpress/DevExtreme/pull/34311/changes#r3664366149).

packages/devextreme/js/__internal/ui/popover/popover.ts
```diff
@@ -266,9 +267,51 @@ class Popover<
...
+
+    if ($targets.length) {
+      // @ts-expect-error trigger should be typed on type 'EventsEngineType'
+      eventsEngine.trigger($targets.first(), 'focus');
+    }
+  }
+
+  _forceFocusLost(): void {
```
Raushen, line 316: Please check base method. May we can move there this logic.
Also please check what exactly base method do and why.

packages/devextreme/js/__internal/ui/popover/popover.ts
```diff
@@ -890,6 +938,10 @@ class Popover<
   }
 
   _dispose(): void {
+    const { visible } = this.option();
+    if (visible && this._getEffectiveAriaRole() === 'dialog') {
+      this._restoreTargetFocus();
```
Raushen, line 966: Should we check if we have focus inside?

packages/devextreme/js/__internal/ui/popover/popover.ts
```diff
@@ -111,6 +111,8 @@ export interface PopoverProperties extends Omit<Properties,
   _overlayContentRole?: string;
 
   _describeTarget?: boolean;
+
+  _preventDialogFocus?: boolean;
```
Raushen: How about _preventDialogContainerFocus?

packages/devextreme/js/__internal/ui/popover/popover.ts
dmlvr: > Please check base method. Also please check what exactly base method do and why.

The base `Overlay._forceFocusLost()` checks if the currently focused element is inside the overlay content (`this._$content`) and blurs (resets) it if so. This prevents focus from being trapped on hidden/invisible elements when the overlay closes, which would break keyboard navigation.

> May we can move there this logic?

No. The base `Overlay` class lacks the role-awareness (`_getEffectiveAriaRole()`) and target resolution mechanics (`_getAriaDescriptionTargets()`) that `Popover` uses to identify and refocus its trigger element.

PR #34444, opened 2026-07-22
T1332624 - HtmlEditor - List item content is not rendered when pasting HTML

Raushen:


packages/devextreme/js/__internal/ui/html_editor/matchers/m_wordLists.ts
```diff
@@ -36,19 +46,38 @@ const getMatcher = (quill) => {
...
+    } else {
+      insertOperation.insert = insertOperation.insert.replace(/^\s+/, '');
+      const listDecoratorMatches = insertOperation.insert.match(/^(\S+)\s+/);
+
+      if (!listDecoratorMatches) {
+        return delta;
+      }
+
+      insertOperation.insert = insertOperation.insert.substring(listDecoratorMatches[0].length);
+      listType = getListType(listDecoratorMatches[1]);
```
dmlvr, line 75: minor: I see two сalls of `getListType` function. One according to the condition, the second without.

I know, it's a small function, but looks a little strange. How about smth like bellow
``` ts
let getListTypeArgument = marker

...

getListTypeArgument = listDecoratorMatches[1]

...

ops.push({ insert: '\n', attributes: { list: getListType(getListTypeArgument), indent } });
```

dmlvr (approved):

packages/devextreme/js/__internal/ui/html_editor/matchers/m_wordLists.ts
Raushen: This function was already written... And it has changes in this request. I'd leave it as is because the next time function fix could be in two places.

PR #34446, opened 2026-07-22
T1307499 - TreeList – Editing the parentIdExpr column causes the row to disappear if the column has setCellValue

bit-byte0:
### What
New rows in **TreeList** now stay visible when you edit the parent id column that uses a `setCellValue` handler

### How
The parent key is captured when the row is added and reused when deciding whether the row is shown, instead of re-reading the value from the cell being edited

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
@@ -859,7 +859,10 @@ class EditingControllerImpl extends modules.ViewController {
   }
 
   private _createInsertInfo(parentKey?: RowKey): InsertInfo {
-    return { [INSERT_INDEX]: this._getInsertIndex(), parentKey };
+    return {
+      [INSERT_INDEX]: this._getInsertIndex(),
+      parentKey,
+    };
   }
 
   private _addInsertInfo(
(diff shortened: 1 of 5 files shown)


packages/devextreme/js/__internal/grids/grid_core/editing/m_editing.ts
```diff
@@ -858,8 +858,8 @@ class EditingControllerImpl extends modules.ViewController {
     }
   }
 
-  private _createInsertInfo(): InsertInfo {
-    return { [INSERT_INDEX]: this._getInsertIndex() };
+  private _createInsertInfo(parentKey?: RowKey): InsertInfo {
+    return { [INSERT_INDEX]: this._getInsertIndex(), parentKey };
```
anna-shakhova: nitpick: hard to read, let's split by line
```
return { 
    [INSERT_INDEX]: this._getInsertIndex(), 
    parentKey 
};
```

packages/devextreme/js/__internal/grids/tree_list/editing/__tests__/parent_id_set_cell_value.integration.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    const $container = $(`#${TREELIST_CONTAINER_ID}`);
+
+    (TreeList.getInstance($container.get(0) as HTMLElement) as TreeList)?.dispose();
+    $container.remove();
+    jest.clearAllMocks();
+    jest.useRealTimers();
+  });
+
+  // T1307499
+  it('should keep the new row visible after editing the parentId column with setCellValue', async () => {
```
anna-shakhova: nitpick: lets add ticket number to test title instead of comment - this way it is much easier to interpret test result

anna-shakhova (approved):

PR #34510, opened 2026-07-27
Grids: throw E1046 errow if key field was removed from data inside onAIColumnRequestCreating

anna-shakhova:


packages/devextreme/js/__internal/grids/grid_core/ai_column/utils.ts
```diff
@@ -33,6 +33,19 @@ export const reduceDataCachedKeys = (
   return newData;
 };
 
+export const isKeyMissingInData = (
+  data: UserData[],
+  keyField: string | string[],
+): boolean => {
+  if (typeof keyField !== 'string') {
+    // The key field should be a string for AI Column functionality.
+    // Return false to avoid unnecessary errors for compound keys.
```
Tucchhaa, line 42: Should we add a note about this to the doc: https://js.devexpress.com/jQuery/Documentation/ApiReference/UI_Components/dxDataGrid/Configuration/columns/ai/

Seems it doesn't currently have such

packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/ai_column.api_handlers.test.ts
```diff
@@ -0,0 +1,427 @@
...
+        ],
+        onAIColumnRequestCreating: (e) => {
+          const reduced = e.data.map((item) => ({ name: item.name, value: item.value }));
+          e.data.splice(0, e.data.length, ...reduced);
+        },
+        onDataErrorOccurred,
+      });
+
+      instance.sendAIColumnRequest('myColumn');
+      jest.advanceTimersByTime(10000);
```
Tucchhaa, line 420: minor: I suggest to use: 
await jest.runAllTimersAsync();
or 
await new Promise(process.nextTick)
instead

packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/ai_column.api_handlers.test.ts
```diff
@@ -0,0 +1,427 @@
...
+
+      expect(columnSendRequestStarted).toHaveBeenCalledTimes(0);
+      expect(onDataErrorOccurred).toHaveBeenCalledTimes(1);
+      expect(errors.Error).toHaveBeenCalledWith('E1046', 'id');
+    });
```
Tucchhaa, line 425: I suggest to add a test to check that error is shown when compound key is used

packages/devextreme/js/__internal/grids/grid_core/ai_column/utils.ts
anna-shakhova: no, I'll fix it in a separate PR

packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/ai_column.api_handlers.test.ts
anna-shakhova: in current implementation error will not be shown for compound key. The tests for compound keys will be added separately

Tucchhaa (approved):

PR #34512, opened 2026-07-27
DataGrid: Fix extra request for the first page on load when starting on another page (T1326786)

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts
```diff
@@ -1122,18 +1122,30 @@ export const data = (Base: ModuleType<DataController>) => class VirtualScrolling
     }
   }
 
+  private isViewportScrolledToPage(): boolean {
+    const initialPageScrollPending = isVirtualPaging(this)
+      && this._rowsScrollController?.getViewportItemIndex() === 0
+      && this.pageIndex() > 0;
```
bit-byte0: the guard reads `viewportItemIndex === 0 && pageIndex() > 0` as "the initial scroll hasn't happened yet" - are those two guaranteed to line up only at init? i'm thinking of a state restore or a focused-row scroll-into-view where `pageIndex()` is legitimately `> 0` while the viewport still sits at item `0`, or a refresh that briefly resets `viewportItemIndex` to `0` - would `updateViewport` then skip a load it actually needs, or does another path cover it?

bit-byte0 (dismissed):

anna-shakhova (dismissed):

anna-shakhova (approved):

PR #34520, opened 2026-07-28
Grids: support compound keys in AI Column

anna-shakhova:


packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/ai_column.cache.test.ts
```diff
@@ -0,0 +1,788 @@
...
+      expect(sendRequestSpy).toHaveBeenCalledTimes(2);
+      expect(sendRequestSpy).toHaveBeenCalledWith({ 2: { id: 2, name: 'Name 2', value: 20 } });
+
+      instance.option('paging.pageIndex', 0);
+      jest.runAllTimers();
+      await Promise.resolve();
+      expect(sendRequestSpy).toHaveBeenCalledTimes(2);
+    });
+
+    it('should use cache with pagination in auto mode (compound key)', async () => {
```
Tucchhaa: I suggest to combine this test with the previous one using matrix, since they almost duplicate each other:

```js
it.each([
  { keyExpr: 'id1' }, expectedKey: 1
  { keyExpr: ['id1', 'id2'] }, expectedKey: '{"id1":2,"id2":"b"}'
])('should use cache with pagination in auto mode when keyExpr = ...', () => {})
```

packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/ai_column.cache.test.ts
```diff
@@ -0,0 +1,788 @@
...
+      instance.option('filterValue', undefined);
+      jest.runAllTimers();
+      await Promise.resolve();
+      expect(sendRequestSpy).toHaveBeenCalledTimes(2);
+      expect(sendRequestSpy).toHaveBeenCalledWith({ 2: { id: 2, name: 'Name 2', value: 20 } });
+    });
+  });
+
+  describe('when a handler replaces the data array', () => {
+    it('should look up the cache using the reassigned e.data, not the original page data', async () => {
```
Tucchhaa, line 329: just curious, why this test was added in this PR? It's not related to compound keys

anna-shakhova: it is related to another issue fixed within PR: in [m_ai_column_integration_controller.ts‎](https://github.com/DevExpress/DevExtreme/pull/34520/changes#diff-1d9e0cdd50d1fd0cd96b9bf33dc9a05ee1478198e48dbca21e2b2dc642b2619e) line 140 -> 146
we used data, not args.data, and it could be outdated if customer replaced, not mutated data inside onAIColumnRequestCreating

packages/devextreme/js/__internal/grids/grid_core/ai_column/__tests__/ai_column.cache.test.ts
anna-shakhova: Grids have different configuration (columns, keyExpr), spies called with different arguments. So even though cases mirror each other, matrix will require heavy configuration, that would make them less readable. I prefer to keep them as is

Tucchhaa: Yeah, I see, but I think it can be resolved by using the same columns in both cases, pls take a look:
```ts
    it.each([
      {
        caseName: 'simple key',
        keyExpr: 'id1',
        expectedFirstKey: 1,
        expectedSecondKey: 2,
      },
      {
        caseName: 'compound key',
        keyExpr: ['id1', 'id2'],
        expectedFirstKey: '{"id1":1,"id2":"a"}',
        expectedSecondKey: '{"id1":2,"id2":"b"}',
      },
    ])('should use cache with pagination in auto mode ($caseName)', async ({
      keyExpr,
      expectedFirstKey,
      expectedSecondKey,
    }) => {
      const aiIntegration = new AIIntegration({
        sendRequest(prompt): RequestResult {
          sendRequestSpy(prompt.data?.data);

          return {
            promise: new Promise<string>((resolve) => {
              const result = {};
              Object.entries(prompt.data?.data).forEach(([key, value]) => {
                result[key] = `Response ${(value as any).name}`;
              });
              resolve(JSON.stringify(result));
            }),
            abort: (): void => {},
          };
        },
      });
      const { instance } = await createDataGrid({
        dataSource: [
          { id1: 1, id2: 'a', name: 'Name 1' },
          { id1: 2, id2: 'b', name: 'Name 2' },
        ],
        keyExpr,
        paging: {
          pageSize: 1,
        },
        columns: [
          { dataField: 'id1', caption: 'ID1' },
          { dataField: 'id2', caption: 'ID2' },
          { dataField: 'name', caption: 'Name' },
          {
            type: 'ai',
            caption: 'AI Column',
            name: 'myColumn',
            ai: {
              aiIntegration,
              prompt: 'Test prompt',
            },
          },
        ],
      });

      await Promise.resolve();
      expect(sendRequestSpy).toHaveBeenCalledTimes(1);
      expect(sendRequestSpy).toHaveBeenCalledWith({ [expectedFirstKey]: { id1: 1, id2: 'a', name: 'Name 1' } });

      instance.option('paging.pageIndex', 1);
      jest.runAllTimers();
      await Promise.resolve();
      expect(sendRequestSpy).toHaveBeenCalledTimes(2);
      expect(sendRequestSpy).toHaveBeenCalledWith({ [expectedSecondKey]: { id1: 2, id2: 'b', name: 'Name 2' } });

      instance.option('paging.pageIndex', 0);
      jest.runAllTimers();
      await Promise.resolve();
      expect(sendRequestSpy).toHaveBeenCalledTimes(2);
    });
```

Tucchhaa (approved):

Alyar666 (approved):

PR #34539, opened 2026-07-29
T1332943: optimize dispose logic in keyboard events

vorobey:


packages/devextreme/js/__internal/grids/new/grid_core/selection/__snapshots__/controller.test.ts.snap
```diff
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
```
ajivanyandev: would probably revert this, to keep consistent with other snapshots

ajivanyandev (approved):

PR #34542, opened 2026-07-29
Grids: add 1px tolerance for elastic scrolling

anna-shakhova:
…e bottom

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
@@ -991,9 +991,12 @@ const rowsView = (Base: ModuleType<RowsView>) => class RowsViewFixedColumnsExten
     if (e.scrollOffset.top < 0) {
       elasticScrollTop = -e.scrollOffset.top;
     } else if (e.reachedBottom) {
-      const $scrollableContent = $(e.component.content());
       const $scrollableContainer = $(e.component.container());
-      const maxScrollTop = Math.max($scrollableContent.get(0).clientHeight - $scrollableContainer.get(0).clientHeight, 0);
+      // Use the container's live scrollTop as the max, not (scrollHeight - clientHeight):
+      // e.scrollOffset.top is derived from that same scrollTop, so they cancel to 0 at the
+      // bottom (browsers can report the two ~1px apart). Correct only because the scroll
+      // handler is synchronous — this scrollTop matches the one e.scrollOffset.top came from.
+      const maxScrollTop = Math.max($scrollableContainer.get(0).scrollTop, 0);
 
       elasticScrollTop = Math.min(maxScrollTop - e.scrollOffset.top, 0);
     }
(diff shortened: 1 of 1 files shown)


packages/devextreme/js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
```diff
@@ -1005,7 +1011,7 @@ const rowsView = (Base: ModuleType<RowsView>) => class RowsViewFixedColumnsExten
     if (this._fixedTableElement) {
       const elasticScrollTop = this._getElasticScrollTop(e);
 
-      if (Math.ceil(elasticScrollTop) !== 0) {
+      if (Math.abs(elasticScrollTop) > ELASTIC_SCROLL_TOLERANCE) {
```
bit-byte0, line 1014: `Math.abs` applies the 1px tolerance to both branches, so the top-bounce case (`-e.scrollOffset.top`) gets relaxed too, even though only the `reachedBottom` case had the Chrome 150 rounding issue

was that intended? we could gate the tolerance behind `e.reachedBottom` and keep the strict check for the top

packages/devextreme/js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
```diff
@@ -1005,7 +1011,7 @@ const rowsView = (Base: ModuleType<RowsView>) => class RowsViewFixedColumnsExten
     if (this._fixedTableElement) {
       const elasticScrollTop = this._getElasticScrollTop(e);
```
bit-byte0, line 1012: the `-1` we tolerate here comes from `Math.floor` in `_getElasticScrollTop` turning a rounding value like `-0.4` into a full `-1`

so rounding (`Math.round`/`Math.trunc`) on that branch would keep it at `0` and fix it at the source - the downstream tolerance may still be safer against jitter though, so curious how you see it

Alyar666 (approved):

packages/devextreme/js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
anna-shakhova: due to accumulation calculation error, difference between scrollHeigth - clientHeight and scrollTop may be up to +-1px, according [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled)
In our particular case it was -1px.
That's why absolute tolerance was added

packages/devextreme/js/__internal/grids/grid_core/column_fixing/m_column_fixing.ts
anna-shakhova: Not quite, in our case both clientHeights and scrollTop are integers, but for clientHeight browser returns rounded values
So here trunc instead of floor in _getElasticScrollTop will not help.
Math.ceil here looks like artifact, as it did nothing

bit-byte0 (approved):

PR #34550, opened 2026-07-29
DataGrid/TreeList - E2E tests: Improve AI Assistant POM

Alyar666:


e2e/testcafe-devextreme/tests/dataGrid/common/aiAssistant/testHelpers.ts
```diff
@@ -28,64 +28,72 @@ export const HANG = '__HANG__';
 
 export const FAIL = '__FAIL__';
 
+// Every piece of test state lives under this single window key, so it is always
```
anna-shakhova: excessive comment, let's remove it

e2e/testcafe-devextreme/tests/dataGrid/common/aiAssistant/testHelpers.ts
```diff
@@ -28,11 +29,17 @@ export const HANG = '__HANG__';
...
 export const FAIL = '__FAIL__';
 
-// Every piece of test state lives under this single window key, so it is always
-// replaced as a whole and no stale key can leak into the next test.
+const DEFERRED_KEY = '__deferredResponse';
+
+// Wraps a response that must stay in flight until resolveAIRequest() is called.
+export const deferred = (response: unknown): unknown => ({ [DEFERRED_KEY]: response });
+
+// The whole test state lives under a single window key: resetAIState replaces it
```
anna-shakhova: excessive comment

packages/testcafe-models/dataGrid/aiAssistantConfirmDialog.ts
```diff
@@ -0,0 +1,30 @@
+import Button from '../button';
+
+const CLASS = {
+  button: 'dx-dialog-button',
+};
+
+const BUTTON_INDEX = {
+  decline: 0,
```
anna-shakhova: Now confirm/decline buttons hardcoded by their index, if we add some button, the test may broke. I suggest differ them by aria-label

packages/testcafe-models/dataGrid/aiAssistantChat.ts
anna-shakhova, line 9: it would be more consistent to import classes from components, in that case rename will not affect tests

Alyar666: I think this is not a good approach. If we accidentally rename a class in the codebase, the tests won't catch it.

anna-shakhova (approved):

PR #34573, opened 2026-07-30
T1332912 - DataGrid - Focused row indicator is duplicated when navigating between grouped rows with repaintChangesOnly

bit-byte0:
### What
Fixed **DataGrid** showing two focused-row highlights at the same time when the focused row moves into a collapsed group while `repaintChangesOnly` is enabled

### How
After a partial update the grid now re-applies the focused-row highlight, so the previously highlighted row is cleared and only the current one stays highlighted

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.integration.test.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.integration.test.ts
@@ -261,7 +261,7 @@ describe('GridCore focus', () => {
     };
 
     const countFocusedRows = (): number => document
-      .querySelectorAll('.dx-data-row.dx-row-focused').length;
+      .querySelectorAll(`#${GRID_CONTAINER_ID} .dx-row.dx-row-focused`).length;
 
     it('should keep a single focused row indicator (T1332912)', async () => {
       const { instance } = await createDataGrid({
--- a/packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.integration.test.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.integration.test.ts
@@ -7,7 +7,7 @@ import type { StoreChange } from '@js/data/store';
 import DataGrid from '@js/ui/data_grid';
 
 import {
-  afterTest, beforeTest, createDataGrid, type DataGridInstance, flushAsync, GRID_CONTAINER_ID,
+  afterTest, beforeTest, createDataGrid, flushAsync, GRID_CONTAINER_ID,
 } from '../__tests__/__mock__/helpers/utils';
 
 describe('GridCore focus', () => {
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.integration.test.ts
```diff
@@ -251,4 +251,53 @@ describe('GridCore focus', () => {
...
     });
   });
+
+  describe('when focusedRowKey is set to a row inside a collapsed group with repaintChangesOnly', () => {
+    const settle = async (): Promise<void> => {
+      for (let i = 0; i < 20; i += 1) {
+        // eslint-disable-next-line no-await-in-loop
+        await flushAsync();
+      }
+    };
```
Tucchhaa: why so many flushAsync are needed? Is it possible to use [waitFor](https://github.com/DevExpress/DevExtreme/blob/27c552d0607d308ef715b92e6f3aaa7a56f771d4/packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.integration.test.ts#L161) instead?

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.ts
```diff
@@ -566,21 +566,40 @@ const data = (Base: ModuleType<DataController>) => class FocusDataControllerExte
...
-      } else if (e.changeType === 'refresh' && e.items.length || isPartialUpdateWithDeleting) {
+      } else if (isRefreshWithItems || isPartialUpdateWithDeleting) {
         this._updatePageIndexes();
         this._updateFocusedRowIfNeeded(e, forceUpdateFocusedRow);
-      } else if (e.changeType === 'append' || e.changeType === 'prepend') {
+      } else if (isAppendOrPrepend) {
         this._updatePageIndexes();
       } else if (isPartialUpdate) {
         this._updateFocusedRowIfNeeded(e, forceUpdateFocusedRow);
+        this._reapplyFocusedRowAfterPartialUpdate(e);
```
anna-shakhova: The naming is confusing, as if we set focus twice.

Tucchhaa: I agree with Anna. By the way, I think we should add a comment about why this function call is needed. Maybe smth like:
`On partial renders the old focused row may not be rerendered, which causes 2 focused rows to appear. To remove stale focused row class, it's needed to reset focus`
That's because the fix actually covers the symptom, instead of fixing the root cause (but I think that would be a complex fix, so I'm ok with the current fix)

packages/devextreme/js/__internal/grids/grid_core/focus/m_focus.integration.test.ts
bit-byte0: switched it to poll a condition instead of the fixed loop. I couldn't reuse the `waitFor` above though - it sleeps on a real `setTimeout`, but this test uses `createDataGrid` which runs on fake timers. I did try the real-timer route (building the grid by hand) and the grouped auto-navigation has a trailing async step that reverts `focusedRowKey` before the wait resolves, so it only passed with a fixed sleep. `flushUntil` keeps the deterministic timer drain and still waits on a real condition instead of a hardcoded count

Tucchhaa (approved):

PR #34614, opened 2026-08-03
Feature/aspnet template compiler csp nonce

vorobey:


packages/devextreme/js/aspnet.js
```diff
@@ -31,6 +31,22 @@
...
+
+    function compileViaScript(src, code) {
+        if(!src || src.tagName !== 'SCRIPT') {
+            return null;
+        }
+        var funcName = src.id.replaceAll('-', '');
+        var func =
+            'function ' + funcName + '(obj,encodeHtml){\n' + code + '\n}';
+        $.globalEval(func, src, window.document);
+        return funcName;
```
vorobey, line 48: it will builds the same string, evals and overwrites window[functionName] with a new, identical, one. Might be added chec for existed function here

PR #34629, opened 2026-08-04
T1333073 - DataGrid - Bound pageIndex does not update after internal page reset

bit-byte0:
### What
**DataGrid** now updates the bound `paging.pageIndex` option when the grid resets to the first page internally (after filtering or searching), so `onOptionChanged` fires and any value bound to `pageIndex` stays in sync with the built-in pager

### How
The filter-apply flow now writes `paging.pageIndex` back to `0` alongside resetting the internal data source, matching how manual paging already keeps the option in sync

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
@@ -1294,7 +1294,7 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     if (dataSource) {
       dataSource.pageIndex(0);
       if (this.option('paging.pageIndex')) {
-        this.resetPageIndexOption();
+        this._silentOption('paging.pageIndex', 0);
       }
       this._isFilterApplying = true;
 
@@ -1307,12 +1307,6 @@ export class DataController extends DataHelperMixin(modules.Controller) {
 
     // @ts-expect-error
     return new Deferred().resolve();
-  }
-
-  private resetPageIndexOption(): void {
-    this._skipProcessingPagingChange = true;
-    this.option('paging.pageIndex', 0);
-    this._skipProcessingPagingChange = false;
   }
 
   public resetFilterApplying() {
--- a/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
@@ -97,7 +97,7 @@ export class DataController extends DataHelperMixin(modules.Controller) {
 
   protected _changes!: any[];
 
-  private _skipProcessingPagingChange: boolean | undefined;
+  private _skipProcessingPagingChange?: boolean;
 
   private _useSortingGroupingFromColumns: boolean | undefined;
 
@@ -1622,7 +1622,7 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     return result;
   }
 
-  private _changePaging(optionName: 'pageIndex' | 'pageSize', value?: number): any {
+  private changePaging(optionName: 'pageIndex' | 'pageSize', value?: number): any {
     const dataSource = this._dataSource;
 
     if (!dataSource) {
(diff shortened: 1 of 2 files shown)


packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -132,7 +97,7 @@ export class DataController extends DataHelperMixin(modules.Controller) {
 
   protected _changes!: any[];
 
-  private readonly _skipProcessingPagingChange: boolean | undefined;
+  private _skipProcessingPagingChange: boolean | undefined;
```
anna-shakhova: We never set `_skipProcessingPagingChange` to `undefined` value
```suggestion
  private _skipProcessingPagingChange?: boolean;
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1648,15 +1622,51 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     return result;
   }
 
+  private _changePaging(optionName: 'pageIndex' | 'pageSize', value?: number): any {
```
anna-shakhova: we do not add _ to the name of private/protected methods anymore, in current convention we distinguish methods visibility with key words private/protected/public

```suggestion
  private _changePaging(optionName: 'pageIndex' | 'pageSize', value?: number): number | DeferredObj<unknown> {
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1648,15 +1622,51 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+        this.pageChanged.fire(pageIndex);
+      });
+  }
+
   /**
    * @extended: virtual_scrolling
    */
-  public pageIndex(value?) {
-    return changePaging(this, 'pageIndex', value);
+  public pageIndex(value?: number): any {
```
anna-shakhova, line 1658:
```suggestion
  public pageIndex(value?: number): number | DeferredObj<unknown> {
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1648,15 +1622,51 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+    return this._changePaging('pageIndex', value);
   }
 
-  public pageSize(value?) {
-    return changePaging(this, 'pageSize', value);
+  public pageSize(value?: number): any {
```
anna-shakhova, line 1662:
```suggestion
  public pageSize(value?: number): number | DeferredObj<unknown> {
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1328,6 +1293,9 @@ export class DataController extends DataHelperMixin(modules.Controller) {
 
     if (dataSource) {
       dataSource.pageIndex(0);
+      if (this.option('paging.pageIndex')) {
```
Alyar666, line 1296: Can we use the _silentOption method here instead of your method?

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
bit-byte0: tried `number | DeferredObj<unknown>`, but callers use the result both as a `number` (arithmetic) and as a Deferred (`.done()`), so a real union breaks ~7 files. kept `: any` to match `load()`/`reload()` here, ok with that?

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
bit-byte0: `_silentOption` wouldn't fire `optionChanged`, but that event is the whole fix here (so a bound `pageIndex` updates). the `_skipProcessingPagingChange` guard only skips the extra pager rebuild. did you have another case in mind?

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
anna-shakhova: let's keep for now, perhaps will be resolvable later during refactoring

anna-shakhova (dismissed):

Alyar666 (approved):

PR #34642, opened 2026-08-04
DataGrid - stabilize tests

Tucchhaa:


e2e/testcafe-devextreme/tests/dataGrid/sticky/common/etalons/datagrid_row_hover_with_fixed_columns (fluent.blue.light).png
anna-shakhova, line 1: new screenshot has another scroll position, is it on purpose?

Tucchhaa: yes, that's on purpose. You can notice in the test that it has no scrolling actions, but scrolling happened because 
`await t.hover(dataRow.element` makes testcafe hover the center of the row. But the row is very wide and its center is beyond the viewport, so testcafe had to scroll. However test only needs to check that row has the hover styles, so just hover the first unfixed cell instead

anna-shakhova (approved):

PR #34670, opened 2026-08-05
Grids: fix eslint errors in data_controller module (part 1)

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -18,11 +18,13 @@
...
 /* eslint-disable prefer-rest-params */
 /* eslint-disable prefer-spread */
+import type { DataSource } from '@js/common/data';
 import ArrayStore from '@js/common/data/array_store';
 import { CustomStore } from '@js/common/data/custom_store';
 import $ from '@js/core/renderer';
+import type { Callback } from '@js/core/utils/callbacks';
 import { deferRender, equalByValue } from '@js/core/utils/common';
-import { Deferred, when } from '@js/core/utils/deferred';
+import { Deferred, type DeferredObj, when } from '@js/core/utils/deferred';
```
anna-shakhova: lets split type imports from methods, classes, etc. this way it is more readable
```suggestion
import { Deferred, when } from '@js/core/utils/deferred';
import type { DeferredObj } from '@js/core/utils/deferred';
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -43,59 +45,32 @@ import type { ValidatingController } from '@ts/grids/grid_core/validating/m_vali
...
-  rowIndex?: number;
-  cells?: unknown[];
-  loadIndex?: number;
-  isSelected?: boolean;
-  removed?: boolean;
-}
-
-export type Filter = any;
+import type {
+  CallbackFlags, DataChange, DataSourceAdapterLike, Filter, HandleDataChangedArguments, Item,
```
anna-shakhova: please, split by lines for readability

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -113,27 +88,27 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
   private _loadingText: string | undefined;
 
-  public dataErrorOccurred: any;
+  public dataErrorOccurred!: Callback;
 
-  public pageChanged: any;
+  public pageChanged!: Callback<[number?]>;
 
-  public pushed: any;
+  public pushed!: Callback<[unknown]>;
```
anna-shakhova, line 112: it is more safe to use optional sign instead of non-null assertions
we may keep `!` for fields defined in `init()` - as they will be defined even though ts doesn't know it

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -43,59 +45,32 @@ import type { ValidatingController } from '@ts/grids/grid_core/validating/m_vali
...
-export type Filter = any;
+import type {
+  CallbackFlags, DataChange, DataSourceAdapterLike, Filter, HandleDataChangedArguments, Item,
+  PagingOptionName, PagingResult,
+} from './types';
 
 export class DataController extends DataHelperMixin(modules.Controller) {
   protected _items!: Item[];
 
   private _cachedProcessedItems!: Item[] | null;
```
anna-shakhova, line 72: lets remove `!` here, it can be `null` from declaration

packages/devextreme/js/__internal/grids/grid_core/data_controller/types.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export interface HandleDataChangedArguments {
+  changeType?: 'refresh' | 'update' | 'loadError';
+  isDelayed?: boolean;
+  isLiveUpdate?: boolean;
+  error?: unknown;
+}
+
+export interface DataChange {
+  [field: string]: unknown;
```
anna-shakhova: what is the purpose of this typing? it allows adding anything in data change, better create more strict type and extend it when required

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
Tucchhaa: `pushed` is always defined, because it's callbackNames: https://github.com/DevExpress/DevExtreme/blob/87e6ce1963b7fb25694d001f29795b1f445a26ec/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts#L215 which is then used in ModuleItem.constrcutor: https://github.com/DevExpress/DevExtreme/blob/87e6ce1963b7fb25694d001f29795b1f445a26ec/packages/devextreme/js/__internal/grids/grid_core/m_modules.ts#L42 to create a callback

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
Tucchhaa: `!` is needed because `null` is set to `_cachedProcessedItems` on `init()` not the constructor

Tucchhaa: or I could set the default value right in the property declaration:
`private _cachedProcessedItems: Item[] | null = null;`

packages/devextreme/js/__internal/grids/grid_core/data_controller/types.ts
Tucchhaa: I guess that was done to avoid changing a lot of lines of code, because change has many fields that are defined across different places. I have made `DataChange` type stricter, but it's not complete (calls to dataController.updateItems are typed). The type will be extended as the work progresses

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
anna-shakhova: this may cause unpredictable effects, cause constructor and init are triggered on different stages of life cycle.
lets then replace with ?, as long as `null` is legitimate value here

anna-shakhova (approved):

PR #34673, opened 2026-08-06
Grids: extract dataController remoteOperations normalization into utils

bit-byte0:
### What
The grids' data controller now resolves the `remoteOperations` option through a dedicated helper instead of inline logiс

### How
The `'auto'`, `true` and `groupPaging` normalization and the store-type checks moved out of `_createDataSourceAdapter` into a new *utils/adapter.ts* module that the controller calls

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
@@ -17,8 +17,6 @@
 /* eslint-disable prefer-rest-params */
 /* eslint-disable prefer-spread */
 import type { DataSource } from '@js/common/data';
-import ArrayStore from '@js/common/data/array_store';
-import { CustomStore } from '@js/common/data/custom_store';
 import $ from '@js/core/renderer';
 import type { Callback } from '@js/core/utils/callbacks';
 import { deferRender, equalByValue } from '@js/core/utils/common';
@@ -26,7 +24,7 @@ import type { DeferredObj } from '@js/core/utils/deferred';
 import { Deferred, when } from '@js/core/utils/deferred';
 import { extend } from '@js/core/utils/extend';
 import { each } from '@js/core/utils/iterator';
-import { isDefined, isObject } from '@js/core/utils/type';
+import { isDefined } from '@js/core/utils/type';
 import errors from '@js/ui/widget/ui.errors';
 import { findChanges } from '@ts/core/utils/m_array_compare';
 import type { EditingController } from '@ts/grids/grid_core/editing/m_editing';
@@ -62,6 +60,7 @@ import type {
   PagingResult,
 } from './types';
 import { resolvePaginate, syncPaging } from './utils/paging';
+import { normalizeRemoteOperations } from './utils/remoteOperations';
 import { generateRowValues } from './utils/row_values';
 
 export class DataController extends DataHelperMixin(modules.Controller) {
@@ -1354,33 +1353,11 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     return dataSourceAdapter;
   }
 
-  public isLocalStore(store = this.store()) {
-    return store instanceof ArrayStore;
-  }
-
-  private isCustomStore(store) {
-    store = store || this.store();
-    return store instanceof CustomStore;
-  }
-
   private _createDataSourceAdapter(dataSource) {
-    let remoteOperations: any = this.option('remoteOperations');
-    const store = dataSource.store();
-    const enabledRemoteOperations = {
-      filtering: true, sorting: true, paging: true, grouping: true, summary: true,
-    };
-
-    // @ts-expect-error
-    if (isObject(remoteOperations) && remoteOperations.groupPaging) {
-      remoteOperations = extend({}, enabledRemoteOperations, remoteOperations);
-    }
-
-    if (remoteOperations === 'auto') {
-      remoteOperations = this.isLocalStore(store) || this.isCustomStore(store) ? {} : { filtering: true, sorting: true, paging: true };
-    }
-    if (remoteOperations === true) {
-      remoteOperations = enabledRemoteOperations;
-    }
+    const remoteOperations = normalizeRemoteOperations(
+      this.option('remoteOperations'),
+      dataSource.store(),
+    );
 
     return this._createDataSourceAdapterCore(dataSource, remoteOperations);
   }
--- a/packages/devextreme/js/__internal/grids/grid_core/data_controller/types.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/data_controller/types.ts
@@ -1,5 +1,7 @@
+import type { Mode } from '@js/common';
 import type { DataSource } from '@js/common/data';
 import type { DeferredObj } from '@js/core/utils/deferred';
+import type { Properties } from '@js/ui/data_grid';
 
 import type { OperationTypes } from '../data_source_adapter/types';
 
@@ -108,3 +110,7 @@ export type PagingResult = number | DeferredObj<unknown> | Promise<unknown>;
 export interface CallbackFlags {
   stopOnFalse: boolean;
 }
+
+export type RemoteOperations = Properties['remoteOperations'];
+
+export type RemoteOperationsOptions = Exclude<RemoteOperations, boolean | Mode | undefined>;
--- a/packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/remoteOperations.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/remoteOperations.ts
@@ -1,0 +1,37 @@
+import { isObject } from '@js/core/utils/type';
+import type { Store } from '@js/data';
+
+import type { RemoteOperations, RemoteOperationsOptions } from '../types';
+import { isCustomStore, isLocalStore } from './store';
+
+export function normalizeRemoteOperations(
+  remoteOperations: RemoteOperations,
+  store: Store,
+): RemoteOperationsOptions | false | undefined {
+  const allExceptGroupPagingEnabled: RemoteOperationsOptions = {
+    filtering: true,
+    sorting: true,
+    paging: true,
+    grouping: true,
+    summary: true,
+  };
+
+  // groupPaging only works when every operation runs remotely.
+  if (isObject(remoteOperations) && remoteOperations.groupPaging) {
+    return { ...allExceptGroupPagingEnabled, ...remoteOperations };
+  }
+  if (remoteOperations === 'auto') {
+    return isLocalStore(store) || isCustomStore(store)
+      ? {}
+      : {
+        filtering: true,
+        sorting: true,
+        paging: true,
+      };
+  }
+  if (remoteOperations === true) {
+    return allExceptGroupPagingEnabled;
+  }
+
+  return remoteOperations;
+}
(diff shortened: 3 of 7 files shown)


packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
```diff
@@ -0,0 +1,38 @@
...
+export function isCustomStore(store: Store): boolean {
+  return store instanceof CustomStore;
+}
+
+export function normalizeRemoteOperations(
+  remoteOperations: RemoteOperationsMode,
+  store: Store,
+): RemoteOperationsOptions | false | undefined {
+  const enabledRemoteOperations = {
+    filtering: true, sorting: true, paging: true, grouping: true, summary: true,
```
anna-shakhova: lets split by lines for readability

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  // groupPaging only works when every operation runs remotely.
+  if (isObject(remoteOperations) && remoteOperations.groupPaging) {
+    return { ...enabledRemoteOperations, ...remoteOperations };
```
anna-shakhova: lets split by lines here too

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/remoteOperations.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import type { RemoteOperationsMode, RemoteOperationsOptions } from '../types';
+
+export function isLocalStore(store: Store): boolean {
+  return store instanceof ArrayStore;
+}
+
+export function isCustomStore(store: Store): boolean {
+  return store instanceof CustomStore;
+}
```
Tucchhaa: duplicated functions:
https://github.com/DevExpress/DevExtreme/blob/760c6f8d240800ccc2445387d31f6fec11429667/packages/devextreme/js/__internal/grids/new/grid_core/data_controller/utils.ts#L42-L48

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/remoteOperations.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function normalizeRemoteOperations(
```
Tucchhaa, line 7: What do you think about moving this func to data_controller/utils.ts? It already has several normalization functions

packages/devextreme/js/__internal/grids/grid_core/data_controller/types.ts
```diff
@@ -22,3 +22,14 @@ export interface PagingDataSource {
...
+export interface RemoteOperationsOptions {
+  filtering?: boolean;
+  sorting?: boolean;
+  paging?: boolean;
+  grouping?: boolean;
+  summary?: boolean;
+  groupPaging?: boolean;
+}
+
+export type RemoteOperationsMode = boolean | 'auto' | RemoteOperationsOptions | undefined;
```
Tucchhaa: I suggest to reuse public type: https://github.com/DevExpress/DevExtreme/blob/760c6f8d240800ccc2445387d31f6fec11429667/packages/devextreme/js/ui/data_grid.d.ts#L1357

DataGrid['remoteOperations']

anna-shakhova: lets just use type from public d.ts rather than create a copy here. I've tried with data_grid, but trellis may differ here, better check

```suggestion
import type { Mode } from '@js/common';
import type { Properties } from '@js/ui/data_grid';

export type RemoteOperations = Properties['remoteOperations'];
export type RemoteOperationsOptions = Exclude<RemoteOperations, boolean | Mode | undefined>;
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1381,32 +1383,14 @@ export class DataController extends DataHelperMixin(modules.Controller) {
   }
 
   public isLocalStore(store = this.store()) {
```
anna-shakhova: can we drop it entirely by always passing argument to guard or this is regression risky?

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function isCustomStore(store: Store): boolean {
+  return store instanceof CustomStore;
+}
+
+export function normalizeRemoteOperations(
+  remoteOperations: RemoteOperationsMode,
+  store: Store,
+): RemoteOperationsOptions | false | undefined {
+  const enabledRemoteOperations = {
```
anna-shakhova: confusing name, maybe smth like allExceptGroupPagingEnabled

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
```diff
@@ -0,0 +1,38 @@
...
+  };
+
+  // groupPaging only works when every operation runs remotely.
+  if (isObject(remoteOperations) && remoteOperations.groupPaging) {
+    return { ...enabledRemoteOperations, ...remoteOperations };
+  }
+  if (remoteOperations === 'auto') {
+    return isLocalStore(store) || isCustomStore(store)
+      ? {}
+      : { filtering: true, sorting: true, paging: true };
```
anna-shakhova: and here

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
bit-byte0: renamed

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
bit-byte0: done

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
bit-byte0: done

packages/devextreme/js/__internal/grids/grid_core/data_controller/types.ts
bit-byte0: reused `Properties['remoteOperations']`

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
bit-byte0: i kept it as a delegate for now - both callers live in _m_rows_view.ts_ and pass no argument (they rely on the `= this.store()` default), so dropping it would pull the utils/adapter import into the **view** layer at both sites; that felt like more coupling than it saves, but i'm happy to inline it if you'd rather?

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
bit-byte0: renamed here too

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/remoteOperations.ts
bit-byte0: _new/grid_core_ already imports from the current core, so pulling the other way (old grid_core importing from new/) would be a reverse dependency - feels like an antipattern, which is why i duplicated instead; do you see a shared spot both could live in, or is keeping them separate ok while the two coexist?

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/remoteOperations.ts
bit-byte0: did you mean a single flat _data_controller/utils.ts_ here in the old grid_core? i went with the per-concern utils/ split (_utils/paging.ts_) to stay consistent with the other utils being added in this area - happy to consolidate into one file if you'd prefer

packages/devextreme/js/__internal/grids/grid_core/data_controller/types.ts
bit-byte0: checked **TreeList** and **CardView** too - their object shapes are subsets of **DataGrid**'s, so the shared alias holds

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/remoteOperations.ts
Tucchhaa: Oh, I see, I didn't notice that the file I mentioned is in new/grid_core. Then I think it's ok to have duplicated functions and a separate file

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/remoteOperations.ts
Tucchhaa: Let's keep it as a separate file, but I suggest to rename the file to `utils/remoteOperations.ts`

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function isCustomStore(store: Store): boolean {
+  return store instanceof CustomStore;
+}
+
+export function normalizeRemoteOperations(
+  remoteOperations: RemoteOperations,
+  store: Store,
+): RemoteOperationsOptions | false | undefined {
+  const allExceptGroupPagingEnabled = {
```
Tucchhaa: minor: let's add a type to the var: `const allExceptGroupPagingEnabled: RemoteOperationsOptions`

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
```diff
@@ -0,0 +1,47 @@
...
+    filtering: true,
+    sorting: true,
+    paging: true,
+    grouping: true,
+    summary: true,
+  };
+
+  // groupPaging only works when every operation runs remotely.
+  if (isObject(remoteOperations) && remoteOperations.groupPaging) {
+    return extend({}, allExceptGroupPagingEnabled, remoteOperations) as RemoteOperationsOptions;
```
Tucchhaa: I suggest to rewrite to:
const result: RemoteOperationsOptions = {
   ....allExceptGroupPagingEnabled,
   ...remoteOperations
}
return result;

to avoid usage of `as`

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/remoteOperations.ts
bit-byte0: sure

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
bit-byte0: done

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/adapter.ts
bit-byte0: applied

Tucchhaa (dismissed):

Tucchhaa (dismissed):

Tucchhaa (approved):

anna-shakhova (approved):

PR #34682, opened 2026-08-06
Grids: extract `generateDataValues` into a standalone utility

anna-shakhova:


packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/row_values.ts
```diff
@@ -0,0 +1,31 @@
+import { AI_COLUMN_NAME } from '@ts/grids/grid_core/ai_column/const';
+import type { Column } from '@ts/grids/grid_core/columns_controller/types';
+
+import type { UserData } from '../types';
+
+export function generateRowValues(
+  data: UserData,
```
Tucchhaa, line 7: minor: rename variable from data to userData

Tucchhaa (approved):

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/row_values.ts
anna-shakhova: I wouldn't rename it - we use `data` in our public types for events that call this method

PR #34701, opened 2026-08-08
 Fluent design tokens pipeline

Raushen:


packages/devextreme-scss/build/tokens/build-tokens.mjs
```diff
@@ -359,6 +358,64 @@ async function validateReferences() {
...
+ * Every `ds.$…` a widget reads must still exist in the token package. validateReferences() above
+ * only checks the generated output against itself, so a release that deletes a token surfaces much
+ * later, as a Sass "Undefined variable" on the first bundle that touches it — one name per rebuild,
+ * with nothing pointing at the bump as the cause.
+ *
+ * The check reads the package's flat index instead of the generated bridge: the two carry the same
+ * names, but the index also carries the version for the message and needs no generated output.
+ * Reusing getBridgeFiles() is what keeps the scope from drifting away from the generator.
+ */
+async function validateConsumedTokens() {
```
EugeniyKiyashko, line 369: The check reads scss/widgets/fluent-next/**/*.scss, but build:tokens declares only {projectRoot}/build/tokens/**/* and pnpm-lock.yaml as inputs. So Nx replays the cache whenever only a stylesheet changed, and the check silently does not run.

Two options:

- add {projectRoot}/scss/widgets/fluent-next/**/* to the build:tokens inputs, or
- move the integration half into tests/consumed-tokens.test.ts - that target already has inputs: ["{projectRoot}/**/*"] and dependsOn: ["build:tokens"], and styles.yml runs it. The pure functions are already there, so the integration case fits naturally.

.github/renovate.json
```diff
@@ -123,6 +123,13 @@
             "matchPackageNames": [
                 "*"
             ]
+        },
+        {
+            "matchPackageNames": [
+                "@devexpress/design-tokens-internal"
+            ],
+            "automerge": false,
+            "minimumReleaseAge": null
```
EugeniyKiyashko, line 133: automerge: false is right, but minimumReleaseAge: null drops the 7-day soak from the * rule, so a PR is opened immediately on every publish of this exact-pinned internal package. That PR then waits for the designers' signal while occupying 1 of only 2 prConcurrentLimit slots, starving every other dependency update for as long as it sits.

For "update manually on a signal", dependencyDashboardApproval: true fits better: the update is visible on the dashboard, no PR until someone approves, no slot taken. minimumReleaseAge: null only makes sense paired with that.

The repo also already has an idiom for packages that must be bumped by hand — {"matchPackageNames": ["devexpress-gantt", "devexpress-diagram", "rrule", "sass-embedded", "systemjs"], "enabled": false} - worth following if no PR is wanted at all.

packages/devextreme-scss/scss/widgets/fluent-next/validation/_sizes.scss
```diff
@@ -24,3 +24,6 @@ $validation-message-font-size: ds.$font-size-caption-md !default;
   $validation-message-padding-inline: ds.$spacing-60 !default;
   $validation-message-line-height: ds.$line-height-120 !default; // dx-no-semantic-role: 120 is off the line-height-role scale
 }
+
+$validation-summary-margin-block-start: ds.$spacing-200 !default;
+$validation-message-padding: ds.$spacing-100 !default;
```
EugeniyKiyashko: This name collides semantically with $validation-message-padding-block / $validation-message-padding-inline declared above: it reads as "the message's padding on all sides", but it actually feeds the base config $validation-message-content-padding - a different element. Please rename to $validation-message-content-padding to match the base parameter.

Also: both new declarations sit after the @if $size branches, ~15 lines away from the padding-block/padding-inline pair they clash with. Moving them up into the unconditional group makes the distinction visible.

packages/devextreme-scss/build/tokens/consumed-tokens.ts
```diff
@@ -0,0 +1,52 @@
...
+/*
+ * Commented-out declarations still spell out token names (stepper/_colors.scss parks a few), so
+ * comments are stripped before scanning — a dead reference must not fail the build.
+ *
+ * Line comments go first, so a `/* … *\/` nested in one disappears with it. The cost is that a `//`
+ * inside a string or a url() swallows the rest of its line: a reference sharing that line would go
+ * uncounted. That under-reports rather than failing wrongly, no theme stylesheet does it today, and
+ * fluent-next-naming.test.ts strips comments the same way.
+ */
+export const stripScssComments = (content: string): string => content
```
EugeniyKiyashko: stripComments (line 65 of this file) is identical copy of stripScssComments.
That commit created the pure module precisely so this could be shared - import { stripScssComments } from '../build/tokens/consumed-tokens'; works from here.

packages/devextreme-scss/build/tokens/consumed-tokens.ts
```diff
@@ -0,0 +1,52 @@
...
+  .filter((_, index) => index % 2 === 0)
+  .join('');
+
+/*
+ * The charset is wider than the kebab-case the generator emits, so a malformed name is captured
+ * whole and fails the check. Matching only [a-z0-9-] would truncate `ds.$spacing-40_typo` to the
+ * valid `spacing-40` and report the stylesheet as verified.
+ */
+export const collectTokenReferences = (content: string): string[] => [
+  ...stripScssComments(content).matchAll(/\bds\.\$([\w-]+)/g),
```
EugeniyKiyashko: The regex only scans ds\.\$, so var(--dxds-…) inside an at-rule parameter (e.g. @use "../../base/x" with ($y: var(--dxds-spacing-100))) survives all three layers: stylelint can't see at-rule params, this test doesn't look for that form, and the build check validates that the name exists but not the form. Adding var\(\s*--dxds- alongside ds\.\$ closes it.

packages/devextreme-scss/build/tokens/consumed-tokens.ts
```diff
@@ -0,0 +1,52 @@
...
+ * comments are stripped before scanning — a dead reference must not fail the build.
+ *
+ * Line comments go first, so a `/* … *\/` nested in one disappears with it. The cost is that a `//`
+ * inside a string or a url() swallows the rest of its line: a reference sharing that line would go
+ * uncounted. That under-reports rather than failing wrongly, no theme stylesheet does it today, and
+ * fluent-next-naming.test.ts strips comments the same way.
+ */
+export const stripScssComments = (content: string): string => content
+  .replace(/\/\/[^\n\r]*/g, '')
+  .split(/\/\*|\*\//)
```
EugeniyKiyashko: .split(/\/\*|\*\//) + even-index filter means a single unbalanced /* or */ (inside a string or a url()) flips parity for the entire rest of the file and silently drops every remaining ds.$ reference from the scan.

The //-swallows-the-line case is documented above; this one isn't, and for a guard a silent pass is the wrong failure direction. Either note it in the same comment or assert that the delimiters balance

packages/devextreme-scss/build/tokens/build-tokens.mjs
```diff
@@ -4,6 +4,11 @@ import { createRequire } from 'node:module';
 import { readdir, readFile, rm } from 'node:fs/promises';
 import StyleDictionary from 'style-dictionary';
 import { registerTransforms } from './transforms.mjs';
+import {
+  buildAvailableNames,
```
EugeniyKiyashko, line 8: This is the only .mjs in the repo that imports a .ts. It carries two undeclared constraints:

- the file is now permanently restricted to erasable-syntax-only TS; an enum/namespace would break the build with a confusing error;
- it sits outside: tconfig.json include: ["./*.ts"], so nothing type-checks it except ts-jest incidentally.

Cheapest fix: make it .mjs (there are three type annotations) or add it to a tsconfig so the constraint is enforced. At minimum a one-line comment saying why it's .ts and what it may not use.

Choose the better option as you decide

packages/devextreme-scss/build/tokens/build-tokens.mjs
```diff
@@ -173,10 +181,13 @@ const getModeFiles = (mode) => [
   `semantic/colors/${THEME_NAME}/${mode}`,
 ];
 
-const getComponentThemeFiles = () => [
-  ...getModeFiles('light'),
-  `components/core/theme/${THEME_NAME}`,
-];
+/*
+ * Source files behind the SCSS bridge. The component tier is deliberately absent: its 601 tokens
```
EugeniyKiyashko: The count 601 is hardcoded in prose here and again in _design-system.scss. The enforced copy lives in tools/naming/registries.json (componentTokenCount: 601, checked by the naming test); these two are unenforced and will rot on the next package bump - which is the exact scenario this PR is about. You already removed 1578 from this file in the last commit; same treatment here (drop the number or say "~600").

packages/devextreme-scss/scss/widgets/fluent-next/map/_sizes.scss
```diff
@@ -0,0 +1,5 @@
+@use "../../../_design-system/variables/ds" as ds;
+
+// adduse
+
+$map-marker-tooltip-margin: ds.$spacing-100 !default;
```
EugeniyKiyashko, line 5: $map-marker-tooltip-margin: $map-marker-tooltip-margin reads as a no-op; it only works because @use "sizes" as * happens to be above and the base config key happens to have the identical name. The same commit uses the clearer namespaced form for validation (validationSizes.$…) - @use "sizes" as mapSizes; + mapSizes.$map-marker-tooltip-margin would be consistent.

packages/devextreme-scss/scss/widgets/fluent-next/common/_mixins.scss
```diff
@@ -3,14 +3,13 @@
 @use "sizes" as *;
 @use "../sizes" as *;
 @use "../../base/mixins" as *;
-@use "../../../_design-system/variables/ds" as ds;
+@use "../validation/sizes" as validationSizes;
 @use "../../base/validation" as baseValidation with (
-  $validation-summary-margin-top: ds.$spacing-200,
-  $validation-message-content-padding: ds.$spacing-100,
+  $validation-summary-margin-top: validationSizes.$validation-summary-margin-block-start,
```
EugeniyKiyashko, line 8: Two blank lines left over from the removed declaration.

packages/devextreme-scss/.stylelintrc.json
```diff
@@ -12,6 +12,10 @@
         "color-function-notation": "legacy",
         "declaration-block-no-redundant-longhand-properties": null,
         "declaration-no-important": true,
+        "declaration-property-value-disallowed-list": [
+            { "/.*/": ["/var\\(\\s*--dxds-/"] },
```
EugeniyKiyashko, line 16: Same rule, same intent as line 17, but the message drops the closing "The public --dx-* properties are unaffected." - that's the sentence that stops someone from thinking the rule bans all custom properties, so it's worth keeping in both. The override above (line 78) also carries a "comment" key while this one doesn't.

packages/devextreme/project.json
```diff
@@ -1224,6 +1224,12 @@
     },
     "build:npm:scss": {
       "executor": "devextreme-nx-infra-plugin:scss-assemble",
```
EugeniyKiyashko, line 1226: [high, question should be resolve] The commit message says "Earlier tokens were included in a package accidently", but this change guarantees they're included: scss-assemble globs scss/**/* with no excludes, so dependsOn: build:tokens makes scss/_design-system/** always present in the published npm scss.

If the message means "inclusion used to depend on build ordering, now it's deterministic" - the diff is right, and then the decision being made is internal design tokens ship in the public devextreme package. fluent-next is marked internal (themebuilder filters it out), yet its SCSS and the bridge do get copied to artifacts/npm/devextreme/scss. Could you state that intent explicitly, and confirm it with whoever owns packaging?

(No clean/copy race here, I checked: targetDefaults.build.dependsOn: ["^build"] makes devextreme-scss:build - which starts with clean:tokens - finish before devextreme:build, and build:npm:scss is uncached, so the new dependsOn can't serve a stale output.)

EugeniyKiyashko (changes requested):

packages/devextreme/project.json
Raushen: Will revert to resolve on the next task

EugeniyKiyashko (approved):

PR #34732, opened 2026-08-11
Grids: type public getters block in DataController

anna-shakhova:


packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1488,94 +1489,106 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+    const rowIndex = this.getRowIndexByKey(key);
 
     if (rowIndex >= 0) {
-      // @ts-expect-error
-      result = new Deferred().resolve(this.items()[rowIndex].data);
+      return Deferred<RawItemData>().resolve(this.items()[rowIndex].data);
     }
 
-    return result || store.byKey(key);
+    return fromPromise(store.byKey(key));
```
anna-shakhova, line 1569: `store.byKey` returns DxPromise that resolves to native or jQuery promise, but then may be used with `done`.
Wrapping with `fromPromise` make this path safe

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1488,94 +1489,106 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+          .done((loadedData: RawItemData[]): void => {
+            const items = this._processItems(
+              this._beforeProcessItems(loadedData),
+              { changeType: 'loadingAll' },
+            );
+            // @ts-expect-error DataGrid-only summary leaks into grid_core
+            d.resolve(items, loadOperation.extra?.summary);
+          })
+          .fail(() => {
+            d.reject();
```
Alyar666: The arrow function isn't needed here because the context of the Deferred object isn't lost. However, we do lose the arguments — for example, the error object.

packages/devextreme/js/__internal/grids/grid_core/data_source_adapter/types.ts
```diff
@@ -24,12 +26,14 @@ export interface OperationTypes {
   paging?: boolean;
 }
 
-export interface LoadOperation extends BaseLoadOperation {
-  data?: unknown[];
-  cachedStoreData?: unknown[];
+export interface LoadOperation extends Omit<BaseLoadOperation, 'operationId'> {
+  operationId?: number;
+  data?: RawItemData[];
```
Alyar666: data can also be a Deferred. I think we should add this type as well.

Alyar666 (changes requested):

packages/devextreme/js/__internal/data/types.ts
Alyar666, line 1: Maybe we should move these types into our DataSource adapter?

anna-shakhova: no, it is not about our DataSource adapter, those types about shared data source, so I'll keep them here

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
anna-shakhova: fixed, this way it requires cast, as args are RawItemData[]

packages/devextreme/js/__internal/grids/grid_core/data_source_adapter/types.ts
anna-shakhova: fixed

Alyar666 (approved):

PR #34741, opened 2026-08-11
GridCore dataController: fix ESLint errors in the refresh/load/push API block

bit-byte0:


packages/devextreme/js/__internal/grids/grid_core/data_controller/types.ts
```diff
@@ -13,6 +13,23 @@ export interface DataSourceAdapterLike {
...
+  reload?: boolean;
+  changesOnly?: boolean;
+  lookup?: boolean;
+  selection?: boolean;
+  allowCancelEditing?: boolean;
+  isPageChanged?: boolean;
+}
+
+export interface UserState {
+  searchText: unknown;
```
anna-shakhova:
```suggestion
  searchText: string | undefined;
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1644,17 +1646,17 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     return this.changePaging('pageSize', value);
   }
 
-  public isCustomLoading() {
-    return this._isCustomLoading || this._dataSource?.isCustomLoading();
+  public isCustomLoading(): boolean {
+    return !!(this._isCustomLoading || this._dataSource?.isCustomLoading());
```
```suggestion
    return this._isCustomLoading || !!this._dataSource?.isCustomLoading();
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1663,140 +1665,143 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+  public refresh(options?: boolean | RefreshOptions): DeferredObj<unknown> {
+    let refreshOptions: RefreshOptions = { reload: true, lookup: true };
     if (options === true) {
-      options = { reload: true, changesOnly: true };
-    } else if (!options) {
-      options = { reload: true, lookup: true };
+      refreshOptions = { reload: true, changesOnly: true };
+    } else if (options) {
+      refreshOptions = options;
     }
```
anna-shakhova: lets extract this as pure utility to get refreshOptions, early returns without mutations

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1663,140 +1665,143 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
     this._disposeDataSource();
     super.dispose();
   }
 
   /**
    * @extended editing
    */
-  public repaintRows(rowIndexes, changesOnly) {
-    rowIndexes = Array.isArray(rowIndexes) ? rowIndexes : [rowIndexes];
+  public repaintRows(rowIndexes: number | (number | undefined)[], changesOnly?: boolean): void {
```
anna-shakhova: `(number | undefined)[]` here is confusing - looks like rowIndexes should always be numbers. Is is legitimate case, when it is array with undefined members?
if case is legitimate, shouldn't we then expect `| undefined` part for symmetry?

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1663,140 +1665,143 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+  public getUserState(): UserState {
     return {
       searchText: this.option('searchPanel.text'),
       pageIndex: this.pageIndex(),
       pageSize: this.pageSize(),
     };
   }
 
-  public getCachedStoreData() {
+  public getCachedStoreData(): unknown {
```
anna-shakhova: I have suspicion that getCachedStoreData() should return RawDataItem[], but better check
We can add cast here if typing in dataSource is too heavy for now. We'll remove cast later, after adapter typing

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1663,140 +1665,143 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
    */
-  public isLastPageLoaded() {
+  public isLastPageLoaded(): boolean {
     const pageIndex = this.pageIndex();
     const pageCount = this.pageCount();
     return pageIndex === (pageCount - 1);
   }
 
-  public load(): any {
+  public load(): DeferredObj<unknown> {
```
anna-shakhova, line 1805: here and rest typed methods: remove eslint disable comments to see eslint errors
dataSource is not typed yet, so we need casts here

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
bit-byte0: the `undefined` only came from the single caller passing `[row.rowIndex]` where `ProcessedItem.rowIndex` is optional. Indices are always set there in practice, so i tightened the param to `number | number[]` and guarded the call site with `isDefined` rather than widen it further, hope it's ok approach to you

anna-shakhova (dismissed):

anna-shakhova (approved):

PR #34743, opened 2026-08-12
GridCore - DataController - Add types for setDataSource/_createDataSourceAdapter/dataSource callbacks

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/data/data_source/types.ts
+++ b/packages/devextreme/js/__internal/data/data_source/types.ts
@@ -1,4 +1,3 @@
-import type { DataSourceEventName as PublicDataSourceEventName } from '@js/common/data';
 import type { DeferredObj } from '@js/core/utils/deferred';
 import type { DataSourceOptionsStub } from '@js/data/data_source';
 import type PublicDataSource from '@js/data/data_source';
@@ -29,7 +28,9 @@ export interface ChangingEvent {
   changes: StoreChange[];
 }
 
-type DataSourceEventName = PublicDataSourceEventName
+type DataSourceEventName = | 'changed'
+  | 'loadError'
+  | 'loadingChanged'
   | 'customizeStoreLoadOptions'
   | 'customizeLoadResult'
   | 'changing';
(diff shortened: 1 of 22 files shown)


packages/devextreme/js/__internal/grids/grid_core/data_source_adapter/m_data_source_adapter.ts
```diff
@@ -506,7 +518,7 @@ export default class DataSourceAdapter extends modules.Controller {
   /**
    * @extended: TreeLists's data_source_adapter
    */
-  protected _handleDataLoaded(options) {
+  protected customizeLoadResultHandler(options) {
```
Tucchhaa, line 519: renamed the function according to the event it is subscribed to: 'customizeLoadResult'

packages/devextreme/js/__internal/data/data_source/types.ts
```diff
@@ -20,3 +24,35 @@ export interface LoadOperation {
...
+export interface ChangingEvent {
+  changes: StoreChange[];
+}
+
+type DataSourceEventName = PublicDataSourceEventName
+  | 'customizeStoreLoadOptions'
+  | 'customizeLoadResult'
+  | 'changing';
+
+export interface DataSource extends PublicDataSource {
```
Tucchhaa, line 38: This interface could be removed if m_data_source used native class instead of Class.inherit:
https://github.com/DevExpress/DevExtreme/blob/7191cc6083d49fb518bc8b6e3798b7e4d6241ca6/packages/devextreme/js/__internal/data/data_source/m_data_source.ts#L28

I suggest to do that in the future

packages/devextreme/js/__internal/grids/grid_core/data_controller/types.ts
```diff
@@ -1,16 +1,22 @@
...
+import type DataSourceAdapter from '../data_source_adapter/m_data_source_adapter';
 import type { ChangedEvent, OperationTypes } from '../data_source_adapter/types';
+import type { InternalGrid } from '../m_types';
 /** data */
 
 export interface DataSourceAdapterLike {
   _dataSource: DataSource;
 }
 
+export interface DataSourceAdapterProvider {
```
anna-shakhova: I think this type should belong to data_source_adapter, not to data_controller
Also this providers now defined in data_grid and tree_list, but they differ only by adapter class, perhaps, we can create some common provider?

anna-shakhova (approved):

PR #34752, opened 2026-08-12
Grids: move _stateStoringController definition from base dataController

anna-shakhova:


packages/devextreme/js/__internal/grids/grid_core/state_storing/extenders/state_storing_data_controller.ts
```diff
@@ -0,0 +1,86 @@
...
+export interface StateStoringDataControllerExtension {
+  stateLoaded: Callback<[]>;
+}
+
+export const stateStoringDataControllerExtender = (
+  Base: ModuleType<DataController>,
+): ModuleType<
+  DataController & StateStoringDataControllerExtension
+> => class StateStoringDataExtender extends Base {
+  // Defined in StateStoringRowsViewExtender.init()
```
Alyar666: Minor: This comment is not accurate.

packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts
```diff
@@ -476,9 +477,8 @@ export const data = (Base: ModuleType<DataController>) => class VirtualScrolling
     super.dispose.apply(this, arguments as any);
   }
 
-  protected _refreshDataSource() {
-    // @ts-expect-error
-    const baseResult = super._refreshDataSource.apply(this, arguments as any) || new Deferred().resolve().promise();
+  protected _refreshDataSource(): DeferredObj<unknown> {
+    const baseResult = super._refreshDataSource() ?? Deferred().resolve();
```
Alyar666: There’s an inconsistency in the return value here. In the other extender, we return a Promise, while here we return resolve. Should we return a Promise here as well?

packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts
```diff
@@ -1434,6 +1434,8 @@ export const resizing = (Base: ModuleType<ResizingController>) => class VirtualS
 };
 
 export const rowsView = (Base: ModuleType<RowsView>) => class VirtualScrollingRowsViewExtender extends Base {
+  protected _dataController!: DataController & Partial<StateStoringDataControllerExtension>;
```
Alyar666, line 1438: Why is _dataController typed differently?

state_storing rowsView: `DataController & StateStoringDataControllerExtension`
virtual_scrolling rowsView: `DataController & Partial<StateStoringDataControllerExtension>`

packages/devextreme/js/__internal/grids/grid_core/state_storing/extenders/state_storing_data_controller.ts
```diff
@@ -0,0 +1,86 @@
...
+
+import type { StateStoringController } from '../m_state_storing_controller';
+
+export interface StateStoringDataControllerExtension {
+  stateLoaded: Callback<[]>;
+}
+
+export const stateStoringDataControllerExtender = (
+  Base: ModuleType<DataController>,
+): ModuleType<
```
Alyar666, line 16: The explicit return type erases the Base type. This is important for TreeList, which extends the result:
`class TreeListStateStoringDataExtender extends stateStoringModule.extenders.controllers.data(Base) `

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts
anna-shakhova: StateStoringDataControllerExtension belongs to state_storing module. So in its RowsViewExtension corresponding callback will always present - we define them simultaneously.
VirtualScrolling is another independent module, and here we add type as Partial - in case stateStoring was not registered before virtualScrolling - we never face it, but it is more honest type

packages/devextreme/js/__internal/grids/grid_core/virtual_scrolling/m_virtual_scrolling.ts
anna-shakhova: Before we used `.promise()` as default value, returned it
But had to add expect-error here due to incorrect typings in Deferred

packages/devextreme/js/__internal/grids/grid_core/state_storing/extenders/state_storing_data_controller.ts
anna-shakhova: there is no issue, TreeListStateStoringDataExtender will have same type as stateStoringModule.extenders.controllers.data - we call here extender function, with proper argument.
Removed m_prefix from TreeList module to make it transparent

Alyar666 (dismissed):

Alyar666 (approved):

aleksei-semikozov (approved):

PR #34760, opened 2026-08-13
feature(demos): add vendor bundles for demos

ajivanyandev:


.github/workflows/visual-tests-demos.yml
```diff
@@ -273,17 +273,80 @@ jobs:
             devextreme-vue-installer.tgz
           retention-days: 1
 
+  # UNTESTED DRAFT — diff-reviewed locally only, never run in real CI.
```
vorobey: this comment is still actual? seems like an leftover from claude/copilot

apps/demos/Demos/PivotGrid/StandaloneFieldChooser/Angular/app/app.component.ts
```diff
@@ -1,108 +1,17 @@
...
-
-let modulePrefix = '';
-// @ts-ignore
-if (window && window.config?.packageConfigPaths) {
-  modulePrefix = '/app';
-}
-
+import { Component } from '@angular/core';
+
+// TODO: restore the original demo (see git history for this file, app.component.html,
```
vorobey, line 4: I prupose to get make this TODO less verbose - am I right, this is a fix for demos that don't build properly because of feature-branch?

apps/demos/utils/server/build-vendor-bundles.js
```diff
@@ -0,0 +1,43 @@
+// CLI: node utils/server/build-vendor-bundles.js [React|Vue ...]
```
vorobey, line 1: comment is too verbose

apps/demos/utils/server/csp-bundle-angular.js
vorobey: i propose to declare smth like `const DEFAULT_SAFE_BATCH_SIZE` and use it here

apps/demos/package.json
```diff
@@ -177,6 +177,7 @@
         "test-testcafe:accessibility": "cross-env STRATEGY=accessibility CONSTEL=jquery node utils/visual-tests/testcafe-runner.ts",
         "csp-server": "node utils/server/csp-server.js 8080",
         "csp-check": "node utils/server/csp-check.js",
+        "build-vendor-bundles": "node utils/server/build-vendor-bundles.js",
```
vorobey, line 180: what do you think to add building vendors in prepare-js or demos:prepare pipeline? i mean, without vendors, demos won't start

PR #34767, opened 2026-08-13
Fluent-next: delivery artifacts to distribution and support accents

Raushen:


packages/devextreme/project.json
```diff
@@ -1226,13 +1226,24 @@
       "executor": "devextreme-nx-infra-plugin:scss-assemble",
       "options": {
         "scssPackagePath": "../devextreme-scss",
-        "outputDir": "./artifacts/npm/devextreme/scss"
+        "outputDir": "./artifacts/npm/devextreme/scss",
+        "exclude": [
+          "widgets/fluent-next/**",
+          "_design-system/**",
```
EugeniyKiyashko: [HIGH] This exclude keeps the token-derived sources out of devextreme/devextreme-internal, but the same sources still ship through a second channel: themebuilder's generate-metadata walks the whole devextreme-scss/scss tree into src/data/scss, and its only filter is filePath.includes('fluent-next') - _design-system/** doesn't match it.

all:build: devextreme-themebuilder-26.2.0.tgz contains 19 data/scss/_design-system/** files (all 11 palettes, semantic tiers, and variables

The fluent-next bundles are only saved by the name-substring accident, so the collector filter needs extending - ideally sharing one exclude list with this target. **Related hygiene: saveScssFiles never cleans the destination, so stale files from earlier checkouts ship too when packing from a dev machine (my local tgz also had 8 leftover widgets/dxdsfluent/** files from before the rename)**

packages/devextreme-scss/scss/widgets/fluent-next/accents/custom.css
```diff
@@ -0,0 +1,25 @@
+/* Set --dx-accent-color to a brand color; load after the theme stylesheet. */
+:root {
```
EugeniyKiyashko: [High] [This should be answered first] What task does custom.css close? Is arbitrary brand accent officially in 26.2 scope, or should we ship only the 11 designed palettes for now and keep custom.css for a follow-up?

Now it's unclear why this file is added in DevExtreme-dist.

JFYI: oklch(from var(...)) requires CSS relative color syntax (Chrome/Edge 119+, Safari 16.4+, Firefox 128+). On a browser without RCS this file doesn't fall back to blue - it makes things worse: the :root block overrides the bundle's valid hex values with declarations that are invalid at computed-value time, so every var(--dxds-primary-*) consumer becomes guaranteed-invalid and loses its value entirely (broken fills instead of the default accent). Wrapping the block in @supports (color: oklch(from red l c h)) { ... } keeps non-supporting browsers on the bundle's designed blue. A docs note on minimum browser versions is needed either way.

packages/devextreme-scss/scss/widgets/fluent-next/accents/custom.css
```diff
@@ -0,0 +1,25 @@
...
+  --dx-accent-color-source: var(--dx-accent-color, #0f6cbd);
+  --dx-accent-lightness-max: 0.95;
```
EugeniyKiyashko: JFYI: With the default blue source, the derived light steps diverge visibly from the designed palette: step 10 comes out at L 0.950 / C 0.040 in OKLCh vs designed #F4F8FC ≈ L 0.977 / C 0.007; step 20 at 0.903/0.052 vs 0.949/0.016 - so the subtle tint surfaces (surface-primary-subdued-*: hover/selected fills) get noticeably darker and ~5× more chromatic than designed. The dark half fits well (ΔL ≤ 0.02). Defaults around --dx-accent-lightness-max: 0.975 and --dx-accent-chroma-min: 0.01 would fit the designed light end much closer. Also, the "Fix comments" commit dropped the header that explained the derivation and its approximation quality (the 0.06 dE note) - now neither the scale rules nor the knobs' meaning are documented anywhere; please restore at least a short version.

packages/devextreme-scss/scss/widgets/fluent-next/accents/custom.css
```diff
@@ -0,0 +1,25 @@
+/* Set --dx-accent-color to a brand color; load after the theme stylesheet. */
+:root {
+  --dx-accent-color-source: var(--dx-accent-color, #0f6cbd);
```
EugeniyKiyashko: JFYI: --dx-accent-color (+ the three tuning knobs) becomes the public API of this feature, but the --dx-* namespace is the evolving component-level tier (declared on component roots), while stable design-system properties live under --dxds-*. Is the short name intentional? If yes, let's document it as a deliberate exception; otherwise --dxds-accent-* would keep the tier split clean.

packages/nx-infra-plugin/src/executors/scss-build/scss-build.impl.ts
```diff
@@ -13,6 +13,10 @@ import { ScssBuildExecutorSchema } from './schema';
 
 const DEFAULT_BUNDLES_DIR = './scss/bundles';
 const DEFAULT_CSS_OUTPUT_DIR = '../devextreme/artifacts/css';
+const GENERATED_ACCENT_PALETTES_DIR = './scss/_design-system/fluent/accents';
+const AUTHORED_ACCENT_STYLES_DIR = './scss/widgets/fluent-next/accents';
+const ACCENT_OUTPUT_DIR_NAME = 'accents';
+const GENERATOR_BANNER_REGEX = /^\s*\/\*[\s\S]*?auto-generated[\s\S]*?\*\/\s*/;
```
EugeniyKiyashko: ^\s*\/\*[\s\S]*?auto-generated[\s\S]*?\*\/ - the lazy [\s\S]*? can cross the end of the first comment: if a file starts with a small unrelated comment and the literal "auto-generated" appears anywhere later in the CSS, everything up to that point gets stripped. Safe for today's inputs, but a footgun. Safer: match only the first comment (/^\s*\/\*[^*]*(?:\*(?!\/)[^*]*)*\*\//) and strip it only if it contains "auto-generated".

packages/devextreme-scss/scss/widgets/fluent-next/accents/custom.css
```diff
@@ -0,0 +1,25 @@
...
+  --dx-accent-lightness-max: 0.95;
+  --dx-accent-lightness-min: 0.15;
+  --dx-accent-chroma-min: 0.04;
+  --dxds-primary-10: oklch(from var(--dx-accent-color-source) calc(l + 9 * (var(--dx-accent-lightness-max) - min(l, var(--dx-accent-lightness-max))) / 9) calc(c - 9 * (max(c, var(--dx-accent-chroma-min)) - var(--dx-accent-chroma-min)) / 9) h);
```
EugeniyKiyashko: JFYI: Watch mode only watches scss/**/*.scss, while compileAccentOverrides also compiles widgets/fluent-next/accents/*.css - edits to custom.css under watch never trigger a rebuild. Worth widening the pattern to **/*.{scss,css}.

packages/devextreme/project.json
```diff
@@ -1226,13 +1226,24 @@
...
       },
       "configurations": {
         "internal": {
           "outputDir": "./artifacts/npm/devextreme-internal/scss"
         }
       },
+      "dependsOn": [
+        {
+          "projects": ["devextreme-scss"],
+          "target": "build:tokens"
```
EugeniyKiyashko: The dependency generates _design-system, which this very target then excludes from the copy - I assume it's for input-hash determinism? A short comment would help. Meanwhile scss/bundles/*.scss, which IS shipped (.npmignore un-ignores it), isn't guaranteed by any dependency: a standalone nx run devextreme:build:npm:scss on a fresh clone silently packs without bundles. Pre-existing, but since we're adding dependencies here - worth considering.

packages/nx-infra-plugin/src/executors/scss-build/scss-build.impl.ts
```diff
@@ -207,6 +224,37 @@ async function compileFile(
...
+
+  if (palettes.length === 0) {
+    throw new Error(`No accent palettes to compile in ${palettesDir}`);
+  }
+
+  const authoredDir = path.resolve(projectRoot, AUTHORED_ACCENT_STYLES_DIR);
+  const accentSources = [...palettes, ...(await globAccentSources(authoredDir, '*.css'))];
+  const accentOutputDir = path.join(cssOutputDir, ACCENT_OUTPUT_DIR_NAME);
+
+  for (const source of accentSources) {
```
EugeniyKiyashko, line 242: Accents deliberately skip postcss/clean-css (minifiers are shaky with relative color syntax) - but nothing in the code says so; a one-line comment would stop someone from "unifying" this later. Separately: each generated palette CSS re-declares all 136 tokens while only the 18 primary steps actually differ across accents - filtering the palette output to primary in createPaletteConfig would shrink each file ~5×. Optional.

packages/devextreme/project.json
```diff
@@ -1226,13 +1226,24 @@
       "executor": "devextreme-nx-infra-plugin:scss-assemble",
       "options": {
         "scssPackagePath": "../devextreme-scss",
-        "outputDir": "./artifacts/npm/devextreme/scss"
+        "outputDir": "./artifacts/npm/devextreme/scss",
+        "exclude": [
```
EugeniyKiyashko: configurations.internal inherits the same exclude, so devextreme-internal also ships without the fluent-next sources. Intended? If internal consumers should keep them, the internal configuration needs exclude: []; if not, fine - just confirming it's a decision, not an accident.

EugeniyKiyashko (changes requested):

packages/nx-infra-plugin/src/executors/scss-build/scss-build.impl.ts
Raushen: We should not apply filtering as source palettes contains all types of tokens. With optimization if the source will change danger-tokens we won't know about it. Palettes should be shrinked on the design-tokens-internal level.
Minification does not work for better readability.

packages/devextreme/project.json
Raushen: After refactoring we do not use `exclude` anymore. The source of truth in the file `internal-scss-paths.json` now.

packages/devextreme/project.json
Raushen: Replaced for "target": "build:themes" - now the target `build:npm:scss` creates bundles.

packages/devextreme-scss/scss/widgets/fluent-next/accents/custom.css
Raushen: Rolled back custom accent feature, will discuss later in a separate scope.

packages/devextreme-scss/scss/widgets/fluent-next/accents/custom.css
Raushen: Rolled back custom accent feature, will discuss later in a separate scope.

packages/devextreme-scss/scss/widgets/fluent-next/accents/custom.css
Raushen: Rolled back custom accent feature, will discuss later in a separate scope.

packages/devextreme-scss/scss/widgets/fluent-next/accents/custom.css
Raushen: Rolled back custom accent feature, will discuss later in a separate scope.

packages/nx-infra-plugin/src/executors/scss-build/scss-build.impl.ts
Raushen: Fixed

packages/devextreme/project.json
Raushen: Fixed

EugeniyKiyashko (approved):

PR #34779, opened 2026-08-14
Numberbox: Refactor and improve typing

Raushen:


packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
```diff
@@ -235,19 +247,21 @@ class NumberBoxMask extends NumberBoxBase<NumberBoxMaskProperties> {
     this._scrollInputTo(direction === MOVE_FORWARD ? 'start' : 'end');
   }
 
-  _shouldMoveCaret(text, caret) {
-    const decimalSeparator = number.getDecimalSeparator();
-    const isDecimalSeparatorNext = text.charAt(caret.end) === decimalSeparator;
-    const moveToFloat = (this._lastKey === decimalSeparator || this._lastKey === '.' || this._lastKey === ',') && isDecimalSeparatorNext;
+  _shouldMoveCaret(text: string, caret: CaretRange | undefined): boolean {
+    const decimalSeparator: string = number.getDecimalSeparator();
+    const isDecimalSeparatorNext = text.charAt(caret?.end ?? 0) === decimalSeparator;
```
EugeniyKiyashko: [blocker] This `?? 0` isn't a neutral fallback, it's a semantically wrong one: 0 is a real index, so
"there is no caret" becomes "the caret is at the start", and the method can answer true and
move the caret. The original was `text.charAt(caret.end)`, which would have thrown instead.
Honest handling:
  _shouldMoveCaret(text: string, caret: CaretRange | undefined): boolean {
    if (!caret) {
      return false;
    }
    ...
  }
This is one of five new places where an absent caret is silently substituted with zero
(see also mask.ts, caret.ts). The root cause is that
_caret() (text_editor.mask.ts) fuses getter and setter into one method dispatching on
arguments.length, so the getter is typed CaretRange | undefined - even though undefined only
happens there when !$input.length, at which point _getInputVal() would already have thrown.
Overloads at the source:
  _caret(): CaretRange | undefined;
  _caret(position: CaretRange, force?: boolean): void;
or splitting into _getCaret()/_setCaret() would remove most of those five fallbacks and let
the remaining case be handled once.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
```diff
@@ -235,19 +247,21 @@ class NumberBoxMask extends NumberBoxBase<NumberBoxMaskProperties> {
...
+    const isSeparatorKey = this._lastKey === decimalSeparator || this._lastKey === '.' || this._lastKey === ',';
 
-    return moveToFloat;
+    return isSeparatorKey && isDecimalSeparatorNext;
   }
 
-  _getInputVal() {
-    return number.convertDigits(this._input().val(), true);
+  _getInputVal(): string {
+    const inputValue: string = number.convertDigits(this._input().val(), true);
```
EugeniyKiyashko, line 271: [blocker] `number` is `any` here, not a typed API. Root cause: numberLocalization is built through
dependencyInjector, and `function injector(object)` in core/utils/m_dependency_injector.ts is
untyped and returns `object`, which under noImplicitAny: false collapses to any. I probed it -
convertDigits / getSign / parse are all any via both import paths, @js/common/... and
@ts/core/... alike.
So the added annotations are unchecked, and three of them are demonstrably false:
  const inputValue: string = number.convertDigits(this._input().val(), true);
      the real signature is (value: string | number) => string | number
  const convertedText: string = number.convertDigits(text, true);           // in _updateFormat
      text = this._format(...), which this PR itself declares as string | undefined -
      the annotation contradicts its own return type two lines up
  const formattedValue: string = number.convertDigits(this._formattedValue, true);  // _formatValue
      _formattedValue?: string, so the input can be undefined and convertDigits returns it
      unchanged, yet the result is declared string
This is worse than leaving them unannotated: the falsehood is now asserted explicitly, and
turning on no-unsafe-assignment later won't catch it.
Options: type injector generically -
  function injector<T extends object>(object: T): T & { inject(o: object): void; resetInjection(): void }
which fixes this for the whole codebase but belongs in its own PR; or, locally, a narrow typed
wrapper such as
  const toStandardDigits = (value: string): string => String(number.convertDigits(value, true));
Either way, please drop the hand-written annotations on `any`.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
```diff
@@ -367,48 +380,49 @@ class NumberBoxMask extends NumberBoxBase<NumberBoxMaskProperties> {
...
     const formatter = isFunction(customFormatter) ? customFormatter : number.format;
 
-    const formattedValue = value === null ? '' : formatter(value, format);
+    const formattedValue: string | undefined = value === null ? '' : formatter(value, format);
 
     return formattedValue;
   }
 
-  _getFormatPattern() {
+  _getFormatPattern(): Format {
```
EugeniyKiyashko, line 439: This is technically not a lie - I checked, Format already includes undefined
(`const x: Format = undefined` compiles) - but it doesn't say what the method actually
returns: either an LDML pattern string, or the user's format passed through as-is.
The price is four identical guards at the call sites, all of the shape
`isString(format) ? format : ''`, in _getTextSeparatorIndex, _isPercentFormat,
_isValueIncomplete, and getCaretAfterFormat in number_box.caret.ts.
Behaviour matches the old code (I verified that getRealSeparatorIndex('') and
getRealSeparatorIndex(object) both yield { occurrence: 1, index: -1 }), but it's "I don't know
what this is" written four times.
Modelling the return explicitly - e.g.
  type FormatPattern = { kind: 'ldml'; pattern: string } | { kind: 'raw'; format: Format }
or at minimum returning the LDML string and handling the raw case once - collapses all four.
Side note: because Format already admits undefined, the `| undefined` in
_getEffectiveFormatOption() and in asFormatObject is redundant, and `_currentFormat?: Format`
is optional twice over.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
```diff
@@ -658,11 +690,11 @@ class NumberBoxMask extends NumberBoxBase<NumberBoxMaskProperties> {
     }
   }
 
-  _detachFormatterEvents() {
+  _detachFormatterEvents(): void {
     eventsEngine.off(this._input(), `.${NUMBER_FORMATTER_NAMESPACE}`);
   }
 
-  _isInputFromPaste(e) {
+  _isInputFromPaste(e: DxEvent<InputEvent>): boolean | undefined {
```
EugeniyKiyashko: `boolean | undefined` documents the tri-state instead of removing it: _isValuePasted is never
initialized. The only caller does `if (isFromPaste)`, so returning `boolean` with `?? false`
is both shorter and honest.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
```diff
@@ -13,32 +13,51 @@ import browser from '@js/core/utils/browser';
...
-const math = Math;
+import type { SpinChangeEvent } from './number_box.spin';
+import SpinButtons from './number_box.spins';
 
 export const WIDGET_CLASS = 'dx-numberbox';
 const FIREFOX_CONTROL_KEYS = ['tab', 'del', 'backspace', 'leftArrow', 'rightArrow', 'home', 'end', 'enter'];
 
 const FORCE_VALUECHANGE_EVENT_NAMESPACE = 'NumberBoxForceValueChange';
 
+export type NumberBoxValue = number | null | undefined;
```
EugeniyKiyashko: [important] This widens rather than sharpens. _parseValue and _normalizeInputValue previously returned
number | null and cannot return undefined, so every caller now has an unreachable case to
consider.
Two knock-on effects:
- `_parseValue(value?: string | NumberBoxValue)` - a `?` on top of a union that already
  contains undefined
- `_parsedValue?: number | null` in number_box.mask.ts doesn't use the alias, so the codebase
  now has two spellings for the same concept
Suggest `type NumberBoxValue = number | null`, and express optionality with `?` / `| undefined`
only where it genuinely exists.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
```diff
@@ -13,32 +13,51 @@ import browser from '@js/core/utils/browser';
...
+
+export type SpinValueChangeEvent = DxEvent<InteractionEvent> | DxMouseWheelEvent;
+
+export type KeyPressEvent = DxEvent<KeyboardEvent> & {
+  originalEvent: InputEvent & KeyboardEvent;
```
EugeniyKiyashko: `InputEvent & KeyboardEvent` describes an event that cannot exist (a real event is never both),
and neither _keyPressHandler override - here or in number_box.mask.ts - ever reads
originalEvent. They only use getChar(e), normalizeKeyName(e), isCommandKeyPressed(e) and
e.preventDefault().
`DxEvent<KeyboardEvent>` is the whole truth here; KeyPressEvent can go away entirely.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
```diff
@@ -397,41 +405,43 @@ class NumberBoxBase<
...
 
-  _isValueIncomplete(value) {
+  _isValueIncomplete(value: string): boolean {
     const incompleteRegex = /(^-$)|(^-?\d*\.$)|(\d+e-?$)/i;
     return incompleteRegex.test(value);
   }
 
-  _isValueInRange(value) {
-    return inRange(value, this.option('min'), this.option('max'));
+  _isValueInRange(value: string | NumberBoxValue): boolean {
```
EugeniyKiyashko: This expands to string | number | null | undefined, i.e. the signature guarantees nothing. Yet
the callers fall into exactly two groups: the base passes a string (_applyValueBoundaries), the
mask passes numbers (_removeHandler, _applyRevertedSign).
Either normalize before the call and keep `value: number`, or split into two functions. Same
comment applies to the override in number_box.mask.ts.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
```diff
@@ -13,32 +13,51 @@ import browser from '@js/core/utils/browser';
...
+): SpinValueChangeEvent => ('event' in e ? e.event : e);
+
 export interface NumberBoxBaseProperties extends TextEditorBaseProperties {
   min?: number;
   max?: number;
   step?: number;
   showSpinButtons?: boolean;
   useLargeSpinButtons?: boolean;
   invalidValueMessage?: string;
+  format?: Format;
```
EugeniyKiyashko: Format includes undefined and excludes null, but the runtime default is null - which is
exactly why the @ts-expect-error above `format: null` in NumberBoxMask._getDefaultOptions() is
still there. The .d.ts meanwhile documents @default "".
Three versions of the truth, and this PR made the situation load-bearing: `isDefined(format)`
in _getEffectiveFormatOption() now looks like dead narrowing to the type system, even though
it's the only thing handling the null default - so a future "simplify to format ?? global"
would break behaviour silently.
Pick one: either `format?: Format | null` here, or default to undefined instead of null (which
also removes that @ts-expect-error).

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
```diff
@@ -136,34 +148,41 @@ class NumberBoxBase<
     this._toggleTabIndex();
   }
 
-  _getDefaultAttributes() {
+  _getDefaultAttributes(): ReturnType<TextEditor<TProperties>['_getDefaultAttributes']> {
```
EugeniyKiyashko: This reaches back into its own base through the generic and is hard to read. The cause is that
the base returns an anonymous inline type. Better to export a named type from
text_editor.base.ts and use it in both places.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
```diff
@@ -13,32 +13,51 @@ import browser from '@js/core/utils/browser';
...
 export interface NumberBoxBaseProperties extends TextEditorBaseProperties {
   min?: number;
   max?: number;
   step?: number;
   showSpinButtons?: boolean;
   useLargeSpinButtons?: boolean;
   invalidValueMessage?: string;
+  format?: Format;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  displayValueFormatter?: ((value: any) => string);
```
EugeniyKiyashko, line 56: With its eslint-disable this is the only remaining `any` in the base. NumberBoxValue fits here:
the value comes from _forceValueRender / _valueChangeEventHandler / _applyValueBoundaries, and
in all three it's a number, null or undefined.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
```diff
@@ -65,8 +71,8 @@ export default class SpinButtons extends TextEditorButton {
 
     this._addToContainer($spinContainer);
 
-    editor?._createComponent($spinUp, SpinButton, extend({ direction: 'up' }, options));
-    editor?._createComponent($spinDown, SpinButton, extend({ direction: 'down' }, options));
+    editor?._createComponent($spinUp, SpinButton, { direction: 'up', ...options });
```
EugeniyKiyashko: Replacing extend with object spread isn't behaviour-preserving: DevExtreme's extend skips
undefined values (core/utils/m_extend.ts: `else if (sourceValue !== undefined)`), object spread
copies them.
`SpinButtonsOptions.disabled` is declared as `boolean | undefined` on a required property, so
`disabled: undefined` now reaches _initOptions → this.option({ disabled: undefined }) and
overwrites Widget's `disabled: false` default.
Today that's harmless only because _toggleDisabledState does Boolean(value). I'd rather not
depend on that.
Fix: make it `disabled?: boolean` and omit the key when there's no value, or keep extend.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
```diff
@@ -3,16 +3,25 @@ import pointer from '@js/common/core/events/pointer';
...
 const SPIN_TOUCH_FRIENDLY_CLASS = 'dx-numberbox-spin-touch-friendly';
 
-export default class SpinButtons extends TextEditorButton {
+interface SpinButtonsOptions {
+  visible: boolean;
+  disabled: boolean | undefined;
+}
+
+export default class SpinButtons extends TextEditorButton<NumberBoxBase> {
+  declare instance?: dxElementWrapper | null;
```
EugeniyKiyashko: This is effectively an @ts-expect-error, just an invisible one. The base declares
`instance?: dxElementWrapper | Button | null`, and the base render() assigns the result of
_create() (typed `Button | dxElementWrapper`) into it - so narrowing it in the subclass
provides no guarantee and has no runtime effect.
Better: make TextEditorButton generic over its instance type
(TextEditorButton<TComponent, TInstance>), so SpinButtons states this once and
_create()/render() stay consistent.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
```diff
@@ -76,57 +82,58 @@ export default class SpinButtons extends TextEditorButton {
...
 
-    return editor?.option('showSpinButtons') && editor?.option('useLargeSpinButtons');
+    return !!showSpinButtons && !!useLargeSpinButtons;
   }
 
   // TODO: get rid of it
-  _legacyRender($editor, isTouchFriendly, isVisible) {
-    $editor.toggleClass(SPIN_TOUCH_FRIENDLY_CLASS, isTouchFriendly);
-    $editor.toggleClass(SPIN_CLASS, isVisible);
+  _legacyRender(
```
EugeniyKiyashko, line 113: All three parameters were made optional, even though both call sites always pass all three.
The optionality exists purely because _create() passes `editor?.$element()`.
Consequences: `$editor?.toggleClass(...)` now silently no-ops instead of failing, and
`toggleClass(cls, undefined)` toggles the class rather than setting it.
_attachEvents already does `if (!editor) return;` - doing the same in _create() lets the
parameters stay required:
  _legacyRender($editor: dxElementWrapper, isTouchFriendly: boolean, isVisible: boolean): void

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
```diff
@@ -76,57 +82,58 @@ export default class SpinButtons extends TextEditorButton {
...
-      const $editor = editor?.$element();
-      const isVisible = this._isVisible();
-      const isTouchFriendly = this._isTouchFriendly();
-      // @ts-expect-error
+    if (shouldUpdate && instance) {
       const $spinButtons = instance.children();
-      const spinUp = SpinButton.getInstance($spinButtons.eq(0));
-      const spinDown = SpinButton.getInstance($spinButtons.eq(1));
+      const spinUp = SpinButton.getInstance<SpinButton>($spinButtons.eq(0));
+      const spinDown = SpinButton.getInstance<SpinButton>($spinButtons.eq(1));
```
EugeniyKiyashko, line 129: `static getInstance<T = any>(element): T` in dom_component.ts - the type argument asserts the
result isn't undefined, but getInstanceByElement can return undefined and the following
.option() would throw. Here and in the three sibling calls an @ts-expect-error was replaced by
an explicit type argument, i.e. a visible escape hatch traded for an invisible one.
Strictly speaking this is no worse than the original, but it isn't a typing improvement either.
The real fix - getInstance(): T | undefined - is probably its own PR; for this one it's enough
not to count it as resolved.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
```diff
@@ -76,57 +82,58 @@ export default class SpinButtons extends TextEditorButton {
...
+    const { editor, instance } = this;
 
-    if (shouldUpdate) {
-      const { editor, instance } = this;
-
-      const $editor = editor?.$element();
-      const isVisible = this._isVisible();
-      const isTouchFriendly = this._isTouchFriendly();
-      // @ts-expect-error
+    if (shouldUpdate && instance) {
```
EugeniyKiyashko: `&& instance` is redundant: the base update() in texteditor_button_collection/button.ts ends
with `return !!this.instance`, so shouldUpdate already implies it. It reads as a semantic guard
but is really a type narrowing.

packages/devextreme/js/__internal/ui/number_box/number_box.caret.ts
```diff
@@ -0,0 +1,163 @@
...
+import type { CaretRange } from '@ts/ui/text_box/utils.caret';
+
+import { getNthOccurrence, getRealSeparatorIndex, splitByIndex } from './utils';
+
+export interface CaretBoundaries {
+  start: number;
+  end: number;
+}
+
+export type CaretPosition = number | CaretRange | undefined;
```
EugeniyKiyashko: Admitting undefined here changed getCaretWithOffset's behaviour: `(range?.start ?? 0) + offset`
now silently returns { start: offset, end: offset } for a missing caret, where the original
(`caret.start === undefined` check on a required argument) would have thrown.
That's the same pattern as the `?? 0` fallbacks in number_box.mask.ts - see the
_shouldMoveCaret comment. If _caret() gets proper getter/setter overloads, this type can go
back to `number | CaretRange` and the `?.`/`?? 0` here disappears.

EugeniyKiyashko (changes requested):

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
Raushen: Fixed in the context of https://github.com/DevExpress/DevExtreme/pull/34807

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
Raushen: Fixed

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
Raushen: Fixed with removing _isValuePasted field

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
Raushen: Fixed with extracting `isString(format) ? format : ''`

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
Raushen: The result doesn't look pretty, there are a lot of NumberBoxValue | undefined usages, but ok, fixed...

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
Raushen: Done here after all — text_editor.base.ts now takes DxEvent, DateBox narrows to DxEvent<InputEvent> where it reads inputType, and KeyPressEvent is gone from NumberBox in favour of DxEvent<KeyboardEvent>.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
Raushen: Fixed with required SpinButtonsOptions.disabled

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
Raushen: Fixed

packages/devextreme/js/__internal/ui/number_box/number_box.caret.ts
Raushen: Applied change

packages/devextreme/js/__internal/ui/number_box/utils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const getNthOccurrence = (str: string, char: string, occurrence: number): number => {
+  let index = -1;
+
+  for (let remaining = occurrence; remaining > 0; remaining -= 1) {
+    index = str.indexOf(char, index + 1);
+  }
+
+  return index;
+};
```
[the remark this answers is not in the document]
Raushen: It is true, but the issue is not a result of this PR. However I decided to fix this also and write tests.

packages/devextreme/js/__internal/ui/number_box/utils.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const getNthOccurrence = (str: string, char: string, occurrence: number): number => {
+  let index = -1;
+
+  for (let remaining = occurrence; remaining > 0; remaining -= 1) {
+    index = str.indexOf(char, index + 1);
+  }
+
+  return index;
```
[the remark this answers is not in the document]
Raushen: It is true, but the issue is not a result of this PR. However I decided to fix this also and write tests.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
Raushen: Fixed

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
Raushen: Fixed

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
Raushen: Fixed

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
Raushen: I tried this and backed it out. NumberBoxValue alone doesn't compile: the property is declared one level up, in TextEditorInternalProperties as (value: string | any[]) => string (with its own eslint-disable), and that type is part of TextEditorPropertiesConstraint, which binds the TextEditor class. Function parameters are contravariant, so a subclass may widen them but not narrow:

`error TS2344: Type 'TProperties' does not satisfy the constraint 'TextEditorPropertiesConstraint'.
  Type '(value: NumberBoxValue | undefined) => string' is not assignable
  to type '(value: string | any[]) => string'.`

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
Raushen: The base does return !!this.instance, so as a condition it is indeed implied — but it can't be dropped: instance is dxElementWrapper | null | undefined, and shouldUpdate is a plain boolean, so nothing narrows the field for instance.children().

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
Raushen: Done as suggested

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
Raushen: Agreed, and I'm not counting it as resolved. I measured the real fix: changing getInstance<T>(): T | undefined in dom_component.ts together with getInstanceByElement<T>(): T | undefined produces 34 errors in 18 files — color_view, slider, range_slider, messagelist, menu, context_menu, multi_view, tree_view, validation_group, hide_toasts, list, collection/item and four spots in number_box.spins — and each one needs a decision rather than a mechanical ?.. getComponentInstance in the same file already has the honest signature, so the target shape is settled; it just doesn't belong in a NumberBox PR.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
EugeniyKiyashko: Verified on 5ff44cd: the early return is in place, and the same guard now covers _arrowHandler, _moveCaret, _setInputText, the dxclick handler and _applyRevertedSign. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
EugeniyKiyashko: Verified on 5ff44cd with a type probe: convertDigits is generic over its input, getSign returns 1 | -1, parse returns number | null | undefined - via both import paths. The annotations are now checked by the compiler. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
EugeniyKiyashko: Centralizing the guard into asPattern() with tests is a reasonable compromise - the four copies are gone and the non-string contract is explicit and tested. The return type itself stays wide, but that can wait. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
EugeniyKiyashko: Checked against the merge base: _isValuePasted was declared, read and reset to false, but never set to true, so removing it is behaviour-preserving. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
EugeniyKiyashko: Agreed it reads noisier - but every `| undefined` now marks a place where undefined really occurs, which is the point. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
EugeniyKiyashko: Verified, including the base and DateBox changes. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
EugeniyKiyashko: Verified. Number(inputValue) coerces exactly the way inRange's comparison did with the raw string, so no behaviour change. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
EugeniyKiyashko: Verified - Format | null in both Properties, the @ts-expect-error is gone and isDefined() now matches the type. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
EugeniyKiyashko: Verified. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.base.ts
EugeniyKiyashko: Correct analysis - under strictFunctionTypes this is a base-level change. A follow-up could relax the TextEditorInternalProperties constraint (for example `(value: never) => string`, or a generic value type) so subclasses can narrow. Fine to leave as is here. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
EugeniyKiyashko: Verified - with disabled always a boolean, the key carries the same effective value extend used to produce. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
EugeniyKiyashko: Verified - and nice that the generic also removed the same hack in DropDownButton and the @ts-expect-error in ClearButton. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
EugeniyKiyashko: Verified. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
EugeniyKiyashko: Agreed - the measurement settles it. Let's track getInstance(): T | undefined separately. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.spins.ts
EugeniyKiyashko: You're right, I conflated the runtime condition with the narrowing. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.caret.ts
EugeniyKiyashko: Verified. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
```diff
@@ -374,48 +392,51 @@ class NumberBoxMask extends NumberBoxBase<NumberBoxMaskProperties> {
...
+  _format(value: NumberBoxValue | undefined, format: Format): string | undefined {
     const formatOption = this._getEffectiveFormatOption();
     const customFormatter = asFormatObject(formatOption)?.formatter ?? formatOption;
-    const formatter = isFunction(customFormatter) ? customFormatter : number.format;
+    const formatter = (isFunction(customFormatter) ? customFormatter : number.format) as (
+      value: NumberBoxValue | undefined, valueFormat: Format,
+    ) => string | undefined;
 
-    const formattedValue = value === null ? '' : formatter(value, format);
+    const formattedValue: string | undefined = value === null ? '' : formatter(value, format);
```
EugeniyKiyashko, line 434: [nit] Now that number.format has a real signature, the cast above asserts something it doesn't satisfy: `(value: NumberBoxValue | undefined, ...) => string | undefined` says number.format accepts undefined, while its actual signature is `(value: number | string, format?)`. The `value === null` guard is what lets undefined slip through to the formatter - the same gap as before, just visible now.

```ts
const formattedValue = isDefined(value) ? formatter(value, format) : '';
```

lets the cast drop `| undefined` from its parameter. Same thought for the parser cast in _parse: number.parse takes `FormatConfig | string`, the cast claims `Format`. Both are honest, local and far better than the `any` they replaced - this is just the last place where the union is asserted rather than narrowed.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
```diff
@@ -703,34 +755,35 @@ class NumberBoxMask extends NumberBoxBase<NumberBoxMaskProperties> {
...
     if (!this._useMaskBehavior()) {
       return super._parseValue(text);
     }
-    // @ts-expect-error ts-error
+
     return this._parsedValue;
   }
 
-  _getPrecisionLimits(text): { min: number; max: number } {
+  _getPrecisionLimits(text: string): { min: number; max: number } {
```
EugeniyKiyashko, line 786: [nit, optional follow-up] This parameter has two meanings: _tryParse passes the current input text, _isValueIncomplete passes asPattern(format). _getPrecisionLimits then calls _getFormatForSign(text) to pick the positive/negative subpattern, so in the second call the sign is derived from the pattern itself. It works today because both subpatterns virtually always share precision, but the name `text` cements the confusion. Pre-existing - Copilot flagged it too and suppressed it.

Cleaner: make it `_getPrecisionLimits(pattern: string)` and resolve _getFormatForSign(text) at the _tryParse call site. Fine as a follow-up.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
Raushen: Fixed

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
Raushen: Fixed

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
EugeniyKiyashko: Verified on f97d92c. The cast now says `(value: number, ...)` and isDefined() narrows to it, so nothing asserts that number.format accepts undefined any more. Checked every _format call site: _hasValueBeenChanged and _setTextByParsedValue both wrap the result in `|| ''`, and the LDML callback in _updateFormat passes a number, so returning '' instead of undefined for a nullish value is not observable. The _parse rewrite is a real improvement too - the custom parser is now called with one argument, which matches the public `parser?: (value: string) => number | Date` contract, and the Date branch is handled explicitly instead of flowing a timestamp into _parsedValue. Resolving.

packages/devextreme/js/__internal/ui/number_box/number_box.mask.ts
EugeniyKiyashko: Verified on f97d92c - _getFormatForSign is hoisted to both call sites and the parameter is honestly named, so the composition is identical to what the method did internally. Resolving.

One residual note, no action needed here: _isValueIncomplete still passes `_getFormatForSign(asPattern(format))`, i.e. the pattern in the position where _getFormatForSign expects input text. That is unchanged behaviour and it is now visible at the call site rather than buried, which was the point. It can only matter for a format whose two subpatterns declare different precision (say `#0.00;(#0.0)`); for the usual symmetric ones both branches yield the same limits. Worth a look only if you ever touch that path again.

EugeniyKiyashko (approved):

PR #34794, opened 2026-08-17
Stabilize CI (grids): mock requests data for demo tests

Alyar666:


apps/demos/Demos/DataGrid/CollaborativeEditing/client-script.js
```diff
@@ -0,0 +1,70 @@
+// TEST-ONLY FILE - not part of the demo.
+// The demo never imports it; the visual-test runner reads it and injects it into
```
anna-shakhova: Lets shorten comment

apps/demos/testing/apiMocks/data/collaborativeEmployees.ts
```diff
@@ -0,0 +1,61 @@
+import { pad } from '../utils';
+
+// Deterministic stand-in for the two endpoints behind DataGrid-CollaborativeEditing:
```
anna-shakhova: The idea of mocks in runner was to mock api routes, not separate demo, so when we use same route we don't have to mock it once again.
So data and handlers should be split by route, not by demo. Split this file to 2, same for other similar places
File names should correspond route for further easier maintenance.

apps/demos/testing/apiMocks/handlers/dataValidation.ts
```diff
@@ -0,0 +1,38 @@
+import type { MockHandler } from '../types';
+import { aspNetLoadHandler, aspNetMutationHandler } from '../aspNetStore';
```
anna-shakhova: I'm not sure about proper place for aspNetStore file.
In widgetsGalleryServiceMock we import file from ./handlers, which in turn import file from same directory as widgetsGalleryServiceMock

apps/demos/testing/etalons/DataGrid-AdvancedMasterDetailView (fluent.blue.light).png
anna-shakhova, line 1: what is the purpose of changing company name in data? We can use same data as it was before and keep existing screenshots

anna-shakhova (changes requested):

apps/demos/testing/apiMocks/aspNetStore.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  if (filter[0] === '!') {
+    return !matches(row, filter[1]);
+  }
+
+  if (Array.isArray(filter[0])) {
+    const parts = filter.filter((part) => Array.isArray(part));
+
+    return filter.includes('or')
```
bit-byte0: i think keying the combinator off `filter.includes('or')` is fragile - it scans the whole array, so a nested value or a mixed `and`/`or` group folds into all-`or`. reading the operator token between conditions (the odd index, `filter[1]`) would be safer?

apps/demos/testing/apiMocks/handlers/dataGridAdvancedMasterDetailView/getOrdersByProduct.ts
anna-shakhova, line 1: lets remove 'get' from filename - keep just route

anna-shakhova (approved):

apps/demos/testing/apiMocks/handlers/dataGridAdvancedMasterDetailView/getOrdersByProduct.ts
Alyar666: We agreed that the file names should match the routing.

Tucchhaa (approved):

PR #34803, opened 2026-08-18
Grids: decouple DataController and ErrorHandlingController

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/error_handling/utils.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/error_handling/utils.ts
@@ -1,12 +1,13 @@
+import { isObject } from '@js/core/utils/type';
 import type { DxError } from '@ts/core/utils/m_error';
 
-import type { ExternalError } from './types';
+import type { GridError } from './types';
 
 export const isDxError = (
-  error: DxError | ExternalError | string,
-): error is DxError => typeof error !== 'string' && 'url' in error && !!error.url;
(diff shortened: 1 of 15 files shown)


packages/devextreme/js/__internal/grids/grid_core/error_handling/utils.ts
```diff
@@ -0,0 +1,22 @@
+import type { DxError } from '@ts/core/utils/m_error';
+
+import type { ExternalError } from './types';
+
+export const isDxError = (
+  error: DxError | ExternalError | string,
```
bit-byte0: `DxError | ExternalError | string` repeats across `isDxError`, `getErrorMessage`, and four methods in *m_error_handling.ts* - would pulling it into one named type (say `GridError`) be worth it, so the shape lives in a single place?

packages/devextreme/js/__internal/grids/grid_core/error_handling/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+): error is DxError => typeof error !== 'string' && 'url' in error && !!error.url;
```
bit-byte0: `'url' in error` throws if `error` is ever `null`/`undefined` (same as the old `error.url` check, so no regression) - worth a leading `error != null` guard here to harden it, or is `error` guaranteed non-nullish by every caller?

bit-byte0 (approved):

PR #34818, opened 2026-08-18
T1329677 DataGrid - Column width changes are not applied immediately

nightskylark:
### Issue

With `columnAutoWidth` enabled, a programmatic width change was not applied until a later repaint:

```ts
dataGrid.columnOption('Task_Assigned_Employee_ID', 'width', 700);
```

The column option was updated, but the grid kept rendering the previously calculated layout.

### Root cause

The grid already has a mechanism for this: `_requireResize` is raised while the component is inside an update cycle and consumed by `GridView._endUpdateCore`, which runs the dimension recalculation once at the end of that cycle. It covers the two ways a width was previously changed:

- `option('columns[0].width', w)` — setting an option opens an update cycle (T590907)
- an explicit `beginUpdate()` / `endUpdate()` around `columnOption` (T628065)

A bare `columnOption` call is not wrapped in anything, so no update cycle exists, the flag is never raised, and no recalculation runs at all. The stale calculated `visibleWidth` therefore keeps winning during rendering. T1329677 is the third instance of the same problem, so it is fixed through the same mechanism rather than a separate one.

Resetting the cached `visibleWidth` would only mask this. The recalculation also produces `bestFitWidth`, the total table width, scrollbar visibility and fixed/sticky column layout, all of which would stay stale.

### Resolution

`columnOption` now opens a component update cycle of its own when it actually changes a width outside any existing cycle, and marks the recalculation as required. All three entry points converge on one code path.

Details worth noting:

- **The flag is raised explicitly** instead of relying on the option-change notification. Command columns have no path in the `columns` option, so they never reach `_updateRequireResize` and would otherwise be skipped, both for a bare call and inside a batch.
- **Internal layout batches are left alone.** They hold only the columns controller lock and apply already resolved dimensions. Interactive column resizing is unaffected.
- **Only a real `beginUpdate` counts.** `Component._lockUpdate`, used around `onInitialized`, raises the component lock without starting a cycle, so a flag raised there would never be consumed and would make every later `resize()` bail out early.
- **The virtual scrolling throttle is dropped for explicit changes.** That controller defers the recalculation by `scrolling.updateTimeout` to coalesce scroll-driven resizes. An explicit width change is a discrete action, so the last resize time is cleared, the way keyboard navigation already does it. This makes the update immediate for all three entry points, including the two older tickets.
- **Equal values do no work.** `columnOptionCore` ignores a value equal to the current one, so the recalculation is forced only when the width really changes, including the object form.

No public API changes.

### Tests

Jest, in `columns_controller.integration.test.ts`:

- a width change through `columnOption`, through the object form, and on a command column each trigger exactly one recalculation
- a command column width change inside a component update cycle triggers one recalculation
- an internal layout batch, a non-width option, and a silent change trigger none
- setting a width to its current value triggers none
- a batch and the `columns` option each trigger exactly one recalculation
- no postponed resize is left behind, either after a width change or after `onInitialized` changes one
- the virtual scrolling throttle is dropped on a width change
- changing a width from `onColumnsChanging` does not recurse

TestCafe, in `columnResizing/functional.ts`: the new width is applied without an explicit repaint, and neighbouring auto-width columns are updated with it.


packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts
```diff
@@ -715,9 +719,54 @@ export const fireOptionChanged = function (that: ColumnsController, options) {
...
+    || changedColumn.visibleWidth === 'auto';
+  const shouldInvalidateChangedColumnVisibleWidth = hasCalculatedVisibleWidth
+    && !isVisibleWidthChangePendingForColumn(that, changedColumn.index);
+
+  if (shouldInvalidateChangedColumnVisibleWidth) {
+    changedColumn.visibleWidth = null;
+  }
+
+  that._columns.concat(that._commandColumns).forEach((column) => {
+    const hasCalculatedVisibleWidth = isNumeric(column.visibleWidth)
```
Alyar666: I think we can avoid duplication by extracting the hasCalculatedVisibleWidth and shouldInvalidateVisibleWidth checks, along with resetting visibleWidth, into a separate function.

packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller_utils.ts
```diff
@@ -715,9 +719,54 @@ export const fireOptionChanged = function (that: ColumnsController, options) {
...
+      && hasCalculatedVisibleWidth
+      && !isVisibleWidthChangePendingForColumn(that, column.index);
+
+    if (shouldInvalidateVisibleWidth) {
+      column.visibleWidth = null;
+    }
+  });
+};
+
+interface ColumnOptionCoreOptions {
```
Alyar666: We usually put types and interfaces in the module's types.ts file.

packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
```diff
@@ -90,6 +90,12 @@ interface IndexedColumns {
   negativeIndexedColumns: Record<string, Column[]>[];
 }
 
+export interface ColumnDimensionsUpdate {
```
Alyar666: We usually put types and interfaces in the module's types.ts file.

packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
```diff
@@ -133,6 +139,8 @@ export class ColumnsController extends modules.Controller {
 
   public _columnChanges?: ColumnsChanges;
 
+  public _pendingVisibleWidthColumnIndices?: Set<number>;
```
Alyar666: Why do we need the underscore here if the property is public? We no longer use this convention even for private properties/methods.

packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
```diff
@@ -322,12 +323,31 @@ export class ColumnsController extends modules.Controller {
     }
   }
 
+  private _setRequireResize(): void {
```
Alyar666: **Minor**: I believe we agreed not to use underscores in private method names.

packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
```diff
@@ -1472,7 +1498,30 @@ export class ColumnsController extends modules.Controller {
...
+        applyOptions();
+        that._setRequireResize();
+      } finally {
+        that.component.endUpdate();
+      }
+    } else {
+      applyOptions();
+
+      if (isWidthChanging) {
+        that._setRequireResize();
```
Alyar666: Why do we need this call? It seems redundant.

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
nightskylark: Setting the flag here does duplicate what `_updateRequireResize` already does, but not in every case.

Writing `width` normally notifies the `columns` option, and that notification ends up in `_updateRequireResize`, which raises `component._requireResize`. Command columns are not part of the `columns` option, so `getColumnFullPath` returns an empty string for them and the notification is never sent. Nothing else raises the flag in that case, so a command column width change inside an update cycle would write the width but never trigger a recalculation.

This call makes sure the flag is raised in any case. It is covered by `should recalculate dimensions when a command column width changes inside a component update cycle`, and deleting the call makes that test, and only that test, fail.

packages/devextreme/js/__internal/grids/grid_core/columns_controller/m_columns_controller.ts
nightskylark: Fixed

Alyar666 (approved):

PR #34833, opened 2026-08-19
DataGrid - Fix column header flickering when data is quickly refreshed (T1333941)

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/column_headers/m_column_headers.ts
```diff
@@ -313,28 +313,30 @@ export class ColumnHeadersView extends ColumnContextMenuMixin(ColumnsView) {
...
-    const deferred = that._updateContent(that._renderTable({ change }), change);
+    const deferred = this._updateContent(this._renderTable({ change }), change);
 
-    $container.toggleClass(MULTI_ROW_HEADER_CLASS, that.getRowCount() > 1);
+    $container.toggleClass(MULTI_ROW_HEADER_CLASS, this.getRowCount() > 1);
 
     // @ts-expect-error
-    super._renderCore.apply(that, arguments);
-    return deferred;
+    super._renderCore.apply(this, arguments);
```
anna-shakhova: lets fix TSlint arguments error

packages/devextreme/js/__internal/grids/grid_core/column_headers/m_column_headers.ts
```diff
@@ -313,28 +313,30 @@ export class ColumnHeadersView extends ColumnContextMenuMixin(ColumnsView) {
...
    */
   protected _renderCore() {
-    const that = this;
-    const $container = that.element();
+    const $container = this.element();
     const change = {};
 
-    if (that._tableElement && !that._dataController.isLoaded() && !that._hasRowElements) {
+    if (this._tableElement && !this._dataController.isLoaded() && !this._hasRowElements) {
       // @ts-expect-error
```
anna-shakhova: what is this expect-error about? lets fix it, type method as well

packages/devextreme/js/__internal/grids/grid_core/column_headers/m_column_headers.ts
```diff
@@ -313,28 +313,30 @@ export class ColumnHeadersView extends ColumnContextMenuMixin(ColumnsView) {
    * @extended: filter_row, virtual_column
    */
   protected _renderCore() {
-    const that = this;
-    const $container = that.element();
+    const $container = this.element();
```
anna-shakhova: lets move variables to the places of their first usage, $container - before line 234, change - before line 330

packages/devextreme/js/__internal/grids/grid_core/column_headers/m_column_headers.ts
Alyar666: The error has been fixed.

Regarding the return type typing, it’s more complicated at the moment. The typing flows through a chain of methods, at least _renderCore -> _updateContent -> waitAsyncTemplates. This would require a lot of additional changes as part of this bug fix.

It would be better to handle this as a separate task.

packages/devextreme/js/__internal/grids/grid_core/column_headers/m_column_headers.ts
Alyar666: I’ve fixed it. However, we used to intentionally move all variable declarations to the beginning of the block. We should agree on a consistent approach for this.

anna-shakhova (approved):

PR #34836, opened 2026-08-19
Feature/remove systemjs from demos. Сleanup

GoodDayForSurf:
- Removed unused Rollup-built demo extras (openai tarball / `create-tgz-packages`) and the Rollup plugins that produced them.
- Dropped the `prepare-js` script and the CI steps that only existed to generate those tgz bundles.
- Angular bundler resolves DevExtreme from ESM (`transpiled-esm-npm/esm`), not CJS.
- ESLint: dropped ignores for `shared/loaders` and `config.js`, the `System` global, and the unused `configs/**` block.
- `modulePrefix` and `window.config?.packageConfigPaths` removed from all Angular demos.
- Localization switched from `npm:…json!json` to normal JSON imports; `systemjs-imports.d.ts` files deleted.
- SystemJS quirks (`npm:`, `!json`) removed from bundlers; only `globalize` / `anti-forgery` aliases remain.
- Dead entries removed from `.gitignore`: `shared/empty-file.js` and `Demos/**/config.js`.
- Demos README rewritten for the current flow: `demos:prepare` / `demos:start`, on-demand bundles, no SystemJS.


pharret31 (approved):

apps/demos/scripts/prepare-js-configs.js
```diff
@@ -1,7 +0,0 @@
-const { init } = require('../utils/shared/config-helper');
```
dxvladislavvolkov, line 1: Do we need this file? It looks like the file can be deleted

apps/demos/package.json
```diff
@@ -59,7 +58,7 @@
         "jszip": "^3.10.1",
         "luxon": "3.4.4",
         "mustache": "4.2.0",
-        "openai": "4.73.1",
```
dxvladislavvolkov, line 62: Are you sure we need to update the library in this pr?

apps/demos/utils/server/tsconfig.csp-bundle-angular.json
```diff
@@ -6,8 +6,9 @@
     "lib": ["ES2022", "DOM", "DOM.Iterable"],
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
-    "esModuleInterop": true,
+      "esModuleInterop": true,
```
dxvladislavvolkov: unnecessary indent

mpreyskurantov (approved):

apps/demos/package.json
GoodDayForSurf: Updating openai the fits very carefully into this PR, since the entire rollup and prepare-js are removed at once

apps/demos/scripts/prepare-js-configs.js
GoodDayForSurf: it is deleted

vorobey (approved):

dxvladislavvolkov (approved):

PR #34838, opened 2026-08-19
Grids: decouple DataController and EditorFactoryController

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
@@ -16,7 +16,6 @@ import type {
 } from '@ts/grids/grid_core/data_source_adapter/types';
 import { isLocalStore } from '@ts/grids/grid_core/data_source_adapter/utils/store';
 import type { EditingController } from '@ts/grids/grid_core/editing/m_editing';
-import type { EditorFactory } from '@ts/grids/grid_core/editor_factory/m_editor_factory';
 import type { FilterSyncController } from '@ts/grids/grid_core/filter/m_filter_sync';
 import type { FocusController } from '@ts/grids/grid_core/focus/m_focus';
 import type { KeyboardNavigationController } from '@ts/grids/grid_core/keyboard_navigation/m_keyboard_navigation';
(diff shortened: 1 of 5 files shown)


packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1151,17 +1150,11 @@ export class DataController extends DataHelperMixin(modules.Controller) {
       change.isLiveUpdate = true;
     }
 
-    this.correctRowIndices(
-      (rowIndex) => this.getRowIndexCorrection(rowIndex, oldItems, newIndexByKey),
+    this.rowIndicesCorrected.fire(
+      (rowIndex: number): number => this.getRowIndexCorrection(rowIndex, oldItems, newIndexByKey),
```
Alyar666, line 1154: **Minor**: I suggest not passing the callback as an argument. Instead, we should pass the necessary data (oldItems, newIndexByKey). Then, make getRowIndexCorrection public and call it in the rowIndicesCorrected callback handler.

Alyar666 (approved):

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
anna-shakhova: `getRowIndexCorrection` computes the delta from two different offsets: 
```
const oldRowIndexOffset = this._rowIndexOffset || 0;   // pre-diff 
const rowIndexOffset = this.getRowIndexOffset();        // post-diff
```

`this._rowIndexOffset` isn't updated to the new value until data_controller.ts:1214, which runs after `_applyChange()` returns at :1198. So the method only produces a correct result inside that window. Making it public advertises "call me whenever" for something that's only meaningful during one synchronousmoment.                                                                                                                                              
                                                                                                                                                       
Secondary points:                                                                                                                                
1. Passing the data widens what a subscriber has to know. Right now the contract is a single question: old index → delta. With oldItems + newIndexByKey the subscriber also has to know those two structures exist, what they mean, and how to combine them. Neither is stored on the controller — both are locals of applyChangesOnly — so the callback signature becomes [ProcessedItem[], RowIndexByKey], i.e. the diff's internals become part of the public event.                                                                                                                     
 2. That also couples subscribers to the diff implementation. The planned row-diffing extraction (and replacing the JSON.stringify equality) changes exactly those structures; a correction function survives that, raw diff inputs don't.

PR #34857, opened 2026-08-20
Grids: Move adaptivity extenders to separate files

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/adaptivity/adaptivity_module.ts
```diff
@@ -0,0 +1,39 @@
...
+import { adaptivityEditingViewControllerExtender } from './extenders/adaptivity_editing_view_controller';
+import { adaptivityEditorFactoryViewControllerExtender } from './extenders/adaptivity_editor_factory_view_controller';
+import { adaptivityHeadersKeyboardNavigationViewControllerExtender } from './extenders/adaptivity_headers_keyboard_navigation_view_controller';
+import { adaptivityKeyboardNavigationViewControllerExtender } from './extenders/adaptivity_keyboard_navigation_view_controller';
+import { adaptivityResizingViewControllerExtender } from './extenders/adaptivity_resizing_view_controller';
+import { adaptivityRowsViewExtender } from './extenders/adaptivity_rows_view';
+import { AdaptiveColumnsController } from './m_adaptivity';
+
+export const adaptivityModule: Module = {
+  defaultOptions() {
```
Tucchhaa: I suggest to add return type:
`defaultOptions(): { columnHidingEnabled: Properties['columnHidingEnabled'] }  {`

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1817,6 +1825,25 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+   * @extended: adaptivity
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public setAdaptiveExpandedKey(key: RowKey): void {}
+
+  /**
+   * @extended: adaptivity
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public toggleExpandAdaptiveDetailRow(key?: RowKey, alwaysExpanded = false): void {}
```
Tucchhaa: I don't think it's a good idea to define adaptivity specific methods in base datacontroller.

How about use this approach:
1. in extenders/adaptivity_data_controller, define an interface:
```ts
interface AdaptivityDataControllerExtension {
  getAdaptiveExpandedKey(): RowKey | undefined;
  public setAdaptiveExpandedKey(key: RowKey): void;
  public toggleExpandAdaptiveDetailRow(key?: RowKey, alwaysExpanded = false): void;
}
```
2. Implement the interface:
```ts
export const adaptivityDataControllerExtender = (
  Base: ModuleType<DataController>,
): ModuleType<DataController> => class AdaptivityDataControllerExtender extends Base implements AdaptivityDataControllerExtension {
```

3. classes that use AdaptivityDataControllerExtension's methods should define dataController explicitly:
```ts
export const adaptivityEditingViewControllerExtender = (
  Base: ModuleType<EditingController>,
): ModuleType<EditingController> => class AdaptivityEditingViewControllerExtender extends Base {
   private dataController!: DataController & AdaptivityDataControllerExtension;

   public init() {
      this.dataController = this.getController('data') as DataController & AdaptivityDataControllerExtension;
       super.init();
    }
```

This way adaptivity-specific methods can be removed from base class. Also it's already an established pattern in our source:
* https://github.com/DevExpress/DevExtreme/blob/debe829323f3ae5486b15a67183b5523095ea8dc/packages/devextreme/js/__internal/grids/grid_core/editing/extenders/editing_form_based_data_controller.ts#L14
* https://github.com/DevExpress/DevExtreme/blob/debe829323f3ae5486b15a67183b5523095ea8dc/packages/devextreme/js/__internal/grids/grid_core/keyboard_navigation/m_keyboard_navigation.ts#L132

packages/devextreme/js/__internal/grids/grid_core/adaptivity/extenders/adaptivity_data_controller.ts
```diff
@@ -0,0 +1,116 @@
...
+  EXPAND_ARIA_NAME,
+  LEGACY_SCROLLING_MODE,
+} from '../const';
+import type { AdaptiveColumnsController } from '../m_adaptivity';
+import { getAdaptiveDetailRowIndex, resolveAdaptiveDetailRowTarget } from '../utils';
+
+export const adaptivityDataControllerExtender = (
+  Base: ModuleType<DataController>,
+): ModuleType<DataController> => class AdaptivityDataControllerExtender extends Base {
+  private adaptiveExpandedKey: RowKey | undefined;
```
anna-shakhova:
```suggestion
  private adaptiveExpandedKey?: RowKey;
```

packages/devextreme/js/__internal/grids/grid_core/adaptivity/extenders/adaptivity_data_controller.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.updateAdaptiveAriaLabels(target.key, oldKey);
+  }
+
+  private updateAdaptiveDetailRows(oldRowIndex: number, newRowIndex: number): void {
+    const rowIndexDelta = this.getRowIndexDelta();
+
+    this.updateItems({
+      allowInvisibleRowIndices: true,
+      changeType: 'update',
+      rowIndices: [oldRowIndex, newRowIndex]
```
anna-shakhova: lets extract definition as separate variable for readability

packages/devextreme/js/__internal/grids/grid_core/adaptivity/adaptivity_module.ts
anna-shakhova:
```suggestion
  defaultOptions(): Pick<GridBaseOptions, 'columnHidingEnabled'> {
```

packages/devextreme/js/__internal/grids/grid_core/adaptivity/utils.ts
```diff
@@ -13,3 +16,22 @@ function isHideableColumn(column: Column): boolean {
...
+    && expandedRowIndex >= 0
+    && expandedRowIndex === rowIndex;
+
+  return isCollapsing ? { key: undefined, rowIndex: -1 } : { key, rowIndex };
+}
+
+export function getAdaptiveDetailRowIndex(dataRowIndex: number, rowIndexDelta = 0): number {
+  const detailRowIndex = dataRowIndex >= 0 ? dataRowIndex + 1 : dataRowIndex;
+
+  return detailRowIndex - rowIndexDelta;
```
anna-shakhova, line 36: what is the meaning of negative index, less than -1? shouldn't we always return -1 if 1st argument is -1?

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1817,6 +1825,25 @@ export class DataController extends DataHelperMixin(modules.Controller) {
   public resetCachedProcessedItems(): void {
     this._cachedProcessedItems = null;
   }
+
+  /**
+   * @extended: adaptivity
+   */
+  public getAdaptiveExpandedKey(): RowKey | undefined {
```
anna-shakhova: extend Extender interface instead of adding adaptivity related methods to base data controller, for other methods as well

anna-shakhova (changes requested):

packages/devextreme/js/__internal/grids/grid_core/adaptivity/utils.ts
Alyar666: Indexes < -1 do not correspond to any row, so they are harmless.

Tucchhaa (dismissed):

packages/devextreme/js/__internal/grids/grid_core/adaptivity/utils.ts
anna-shakhova: then suggestion:

```suggestion
  if (dataRowIndex === -1 ) {
    return -1;
  }
  
  const detailRowIndex = dataRowIndex + 1;
  
  return detailRowIndex - rowIndexDelta;
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -171,6 +171,14 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     this.postInit();
   }
 
+  /**
+   * @extended: adaptivity, master_detail
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  protected getRowIndicesForExpand(key: RowKey): number[] {
```
anna-shakhova, line 182: this method is not related to base dataController, looks like it is grouping leakage.
Is it possible to resolve `changeRowExpand` path now? if it is too big, lets add it to our refactoring path known issues and mark there and it core extenders with comment as group leak

Alyar666: There are actually two separate issues here:

- The coupling between adaptivity and master-detail through getRowIndicesForExpand. As part of the master-detail module refactoring, this coupling should be removed by moving the getRowIndicesForExpand implementation from adaptive behavior to master-detail. I’ll leave a comment in the code about this.
- Expanding groups, master-detail rows, and TreeList nodes currently relies on the same shared mechanism. It should be split into separate feature-specific implementations without changing the public API. I’ve added this to our known issues.

packages/devextreme/js/__internal/grids/grid_core/adaptivity/utils.ts
Alyar666: I looked into this area in more depth. With these changes, we could potentially break virtual scrolling in legacy mode. Therefore, I suggest leaving it as is to avoid regressions.

anna-shakhova (approved):

PR #34858, opened 2026-08-21
Map: OSM provider — Define the public API contract

AlisherAmonulloev:


packages/devextreme/js/ui/widget/ui.errors.js
```diff
@@ -248,9 +248,15 @@
...
     */
+/**
+    * @name ErrorsUIWidgets.W1031
+    */
 /**
     * @name ErrorsUIWidgets.W1032
     */
+/**
+    * @name ErrorsUIWidgets.W1033
+    */
```
pharret31, line 259: I remember there was another error started with E, am I wrong?

AlisherAmonulloev: you remember correctly, it was [E1069](https://github.com/DevExpress/DevExtreme/blob/32f36efb51d34702cd7cb788b635f1e5ead7a0b0/packages/devextreme/js/__internal/ui/errors.ts#L116) and it was added in the previous PR. In this PR, we only add W1031 and W1033 for the new callbacks.

pharret31 (approved):

arman-boyakhchyan (approved):

PR #34860, opened 2026-08-21
CardView: track items dependencies without bare signal reads (T1334012)

anna-shakhova:

### DIFF
--- a/packages/devextreme/build/pure-getters-guard.test.ts
+++ b/packages/devextreme/build/pure-getters-guard.test.ts
@@ -1,16 +1,3 @@
-/**
- * Guard against minifiers deleting signal dependency subscriptions (T1334012)
- *
- * `@preact/signals-core` subscribes as a *side effect* of reading `.value`, so a read whose result
- * is never used can still be load-bearing. Terser's `compress.pure_getters` asserts that no getter
- * has side effects and deletes such reads, silently unsubscribing the reaction.
- *
- * This is not an exotic opt-in: Angular CLI's legacy webpack builder enables
- * `pure_getters` by default via `buildOptimizer`
- *
- * Every module is minified twice and the two outputs are compared. Any `.value` read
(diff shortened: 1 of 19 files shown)


packages/devextreme/js/__internal/grids/new/grid_core/search/controller.ts
```diff
@@ -51,7 +51,7 @@ export class SearchController {
     text: string,
   ): HighlightedTextItem[] | null => splitHighlightedText(
     text,
-    this.highlightTextOptions.peek(),
+    this.highlightTextOptions.value,
```
anna-shakhova, line 54: reads `.value`, not `.peek()`, on purpose. 
Callers inside a computed - ItemsController renders every field through here - then subscribe to the highlight options naturally, instead of the caller having to read the signal separately just to register the dependency.

packages/devextreme/build/pure-getters-guard.test.ts
```diff
@@ -0,0 +1,114 @@
+/**
+ * Guard against minifiers deleting signal dependency subscriptions (T1334012)
+ *
+ * `@preact/signals-core` subscribes as a *side effect* of reading `.value`, so a read whose result
+ * is never used can still be load-bearing. Terser's `compress.pure_getters` asserts that no getter
+ * has side effects and deletes such reads, silently unsubscribing the reaction.
+ *
+ * This is not an exotic opt-in: Angular CLI's legacy webpack builder enables
```
Alyar666: Too many comments. Please reduce them.

Alyar666 (changes requested):

packages/devextreme/build/pure-getters-guard.test.ts
anna-shakhova: fixed

Alyar666 (approved):

PR #34880, opened 2026-08-24
DataGrid - Summary - add types to SummaryDataControllerExtender

Tucchhaa:

### DIFF
--- a/packages/devextreme/js/__internal/grids/data_grid/export/m_export.ts
+++ b/packages/devextreme/js/__internal/grids/data_grid/export/m_export.ts
@@ -617,7 +617,7 @@ export class ExportController extends dataGridCore.ViewController {
       aggregates: totalAggregates,
       visibleColumns: this._columnsController.getVisibleColumns(null, true),
       calculateTargetColumnIndex: (summaryItem, column) => (
-        isDataColumn(column) ? column.index : -1
+        isDataColumn(column) ? (column?.index ?? -1) : -1
       ),
     });
   }
--- a/packages/devextreme/js/__internal/grids/data_grid/export/m_export.ts
+++ b/packages/devextreme/js/__internal/grids/data_grid/export/m_export.ts
@@ -22,6 +22,7 @@ import type { ModuleType } from '../../grid_core/m_types';
 import type { SelectionController } from '../../grid_core/selection/m_selection';
 import type { RowsView } from '../../grid_core/views/m_rows_view';
 import dataGridCore from '../m_core';
+import { isDataColumn } from '../m_utils';
 
 const DATAGRID_EXPORT_MENU_CLASS = 'dx-datagrid-export-menu';
 const DATAGRID_EXPORT_BUTTON_CLASS = 'dx-datagrid-export-button';
@@ -616,8 +617,7 @@ export class ExportController extends dataGridCore.ViewController {
       aggregates: totalAggregates,
       visibleColumns: this._columnsController.getVisibleColumns(null, true),
       calculateTargetColumnIndex: (summaryItem, column) => (
-        // @ts-expect-error
-        this._dataController._isDataColumn(column) ? column.index : -1
+        isDataColumn(column) ? column.index : -1
       ),
     });
   }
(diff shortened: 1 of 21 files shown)


packages/devextreme/js/__internal/grids/data_grid/export/m_export.ts
```diff
@@ -616,8 +617,7 @@ export class ExportController extends dataGridCore.ViewController {
       aggregates: totalAggregates,
       visibleColumns: this._columnsController.getVisibleColumns(null, true),
       calculateTargetColumnIndex: (summaryItem, column) => (
-        // @ts-expect-error
-        this._dataController._isDataColumn(column) ? column.index : -1
+        isDataColumn(column) ? (column?.index ?? -1) : -1
```
anna-shakhova, line 620:
```suggestion
        isDataColumn(column) && isDefined(column.index) ? column.index  : -1
```
also lets convert `isDataColumn` to type guard for easier checks

packages/devextreme/js/__internal/grids/data_grid/summary/extenders/summary_data_controller.ts
```diff
@@ -77,26 +92,33 @@ export const summaryDataControllerExtender = (
...
       let afterGroupColumnIndex = -1;
 
-      each(options.visibleColumns, function (visibleIndex: any) {
+      options.visibleColumns.forEach((column, visibleIndex) => {
         const prevColumn = options.visibleColumns[visibleIndex - 1];
 
-        if (groupItem.groupIndex === this.groupIndex) {
-          groupColumnIndex = this.index;
+        if (groupItem.groupIndex === column.groupIndex) {
+          groupColumnIndex = column.index ?? -1;
```
anna-shakhova: we repeat this construction multiple times, can column.index be undefined? or we can remove such checks with type guard?

packages/devextreme/js/__internal/grids/data_grid/summary/extenders/summary_data_controller.ts
```diff
@@ -262,36 +289,21 @@ export const summaryDataControllerExtender = (
...
+  public optionChanged(e: OptionChanged): void {
+    if (['summary', 'sortByGroupSummaryInfo'].includes(e.name)) {
+      e.name = 'dataSource';
     }
-    super.optionChanged(args);
-  }
 
-  public footerItems() {
-    return this._footerItems;
+    super.optionChanged(e);
```
anna-shakhova: usually we use switch in optionChanged

```suggestion
public optionChanged(args: OptionChanged): void {
  switch (args.name):
    case 'summary':
    case 'sortByGroupSummaryInfo':
      super.optionChanged({
        ...args,
        name: 'dataSource',
      });
      break;
    default:
      super.optionChanged(args);
}

```

packages/devextreme/js/__internal/grids/data_grid/summary/m_summary.ts
```diff
@@ -375,26 +348,76 @@ export const dataSourceAdapterExtender = (Base: ModuleType<DataSourceAdapter>) =
...
+  }
+
+  protected customizeLoadResultHandlerCore(options): void {
+    const groups = normalizeSortingInfo(
+      options.storeLoadOptions.group ?? options.loadOptions.group ?? [],
+    );
+    const remoteOperations = options.remoteOperations ?? {};
     const summary = this.getSummary(remoteOperations);
 
     if (!options.isCustomLoading || options.storeLoadOptions.isLoadingAll) {
```
anna-shakhova: may we decrease nesting depth here?

packages/devextreme/js/__internal/grids/data_grid/export/m_export.ts
Tucchhaa: `column?.index ?? -1` is still needed because `index` is typed as `number | undefined`

packages/devextreme/js/__internal/grids/data_grid/summary/extenders/summary_data_controller.ts
Tucchhaa: column.index can't undefined here, but `Column` type's properties are all optional so that's why  `?? -1` is needed here.

I think that we probably should create a type like `VisibleColumns` that would have non-optional props. I have looked into it and it will take quite some time to correctly define such type.

So I would like to postpone it if you don't mind. For now, I have replaced `?? -1` with `@ts-expect-error`

anna-shakhova (approved):

PR #34924, opened 2026-08-26
DataGrid: separate the kept-row refresh from the row changes lookup

anna-shakhova:
### Summary

`findItemChanges` hided a side effect off the changes lookup: its `isItemEquals` predicate called `updateRowCells` whenever two rows compared equal.

Now `applyChangesOnly` calls `findChanges` with a plain predicate, and a new `updateUnchangedRows` helper refreshes the rows the lookup **reported nothing about** afterwards. 

### Changed behaviour

Previously when `findChanges` return `undefined` (on a moved row, duplicate keys, or a throw), the rows it had already compared were refreshed. And then in `applyChangesOnly` we called `_applyChangeFull`, which replaces `_items` and leaves `changeType: 'refresh'`, so the rows view rerenders the whole table.
Now we do not refresh same state rows in such cases as it will be discarded anyway

Everything else is identical.

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
@@ -1,6 +1,7 @@
 import type { Store } from '@js/common/data';
 import type { Callback } from '@js/core/utils/callbacks';
 import { deferRender } from '@js/core/utils/common';
+import { logger } from '@js/core/utils/console';
 import type { DeferredObj } from '@js/core/utils/deferred';
 import { Deferred, when } from '@js/core/utils/deferred';
 import { isDefined } from '@js/core/utils/type';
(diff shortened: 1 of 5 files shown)


packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1132,13 +1115,27 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+      isItemEquals: this.isSameRowState.bind(this),
+    });
 
+    // Changes cannot be found for a moved row, duplicate keys, or any throw.
     if (!itemChanges) {
       this._applyChangeFull(change);
       return;
     }
 
+    try {
```
anna-shakhova, line 1131: `findChanges` used to catch a throwing row template into a silent full repaint (return `undefined`), since the refresh ran inside its `try/catch`. `applyChangesOnly` now catches for itself to keep that recovery — deliberate and visible instead of incidental.

Alyar666 (approved):

PR #34925, opened 2026-08-26
Editors: get rid of the ctor overrides

EugeniyKiyashko:


packages/devextreme/js/__internal/ui/editor/editor.ts
```diff
@@ -92,15 +92,10 @@ class Editor<
...
-    this.showValidationMessageTimeout = undefined;
-    this.validationRequest = Callbacks();
-
-    super.ctor(element, options);
-  }
-
   _createElement(element: Element): void {
     super._createElement(element);
+    this.showValidationMessageTimeout = undefined;
+    this.validationRequest = Callbacks();
```
Raushen: How about move lines to the _init? Now it seems like side effects while creating element.

EugeniyKiyashko: Good point

Checked two things before moving them: `_init()` still runs before the markup
(`DOMComponent.endUpdate()` reaches `_init()` before `_updateDOMComponent()`), and it
runs exactly once per instance - `_initialized` is only ever set to `true`, never reset -
so the `Callbacks()` object can't be recreated under an already subscribed validator.
Nothing reads either field in between: `validationRequest` is only fired from
`_optionChanged`, which is gated behind `_initialized`.

Raushen (approved):

PR #34937, opened 2026-08-26
DataGrid: refactor applyChangesOnly branch

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/grids/data_grid/summary/extenders/summary_data_controller.ts
+++ b/packages/devextreme/js/__internal/grids/data_grid/summary/extenders/summary_data_controller.ts
@@ -19,7 +19,7 @@ import type {
   CalculateSummaryCellsArgs, ColumnMap, FooterItem, SummaryCellItem,
   SummaryGroupItem,
 } from '../types';
-import { getColumnFromMap, getSummaryCellIndex, hasSummaryCells } from '../utils';
+import { getColumnFromMap, getSummaryCellIndex, isSummaryGroupItem } from '../utils';
 import { getGroupAggregates } from '../utils/get_group_aggregates';
 import { getSummaryItemIndex } from '../utils/get_summary_item_index';
 
@@ -272,7 +272,7 @@ export const summaryDataControllerExtender = (
   }
 
   protected isSameRowState(item1: ProcessedItem, item2: ProcessedItem): boolean {
-    if (hasSummaryCells(item1)
+    if (isSummaryGroupItem(item1)
       && JSON.stringify(item1.summaryCells) !== JSON.stringify(item2.summaryCells)) {
       return false;
     }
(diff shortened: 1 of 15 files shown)


packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/row_changes.ts
```diff
@@ -174,13 +175,27 @@ export function resetChangedRows(change: UpdateChange): ChangedRows {
...
-export function markUpdateChange(change: DataChange, changedRows: ChangedRows): void {
+function toChangedRows(updateRowChanges: UpdateRowChange[]): ChangedRows {
+  return {
+    items: updateRowChanges
+      .map(({ item }) => item)
+      .filter((item): item is ProcessedItem => !!item),
+    rowIndices: updateRowChanges.map(({ rowIndex }) => rowIndex),
+    changeTypes: updateRowChanges.map(({ changeType }) => changeType),
+    columnIndices: updateRowChanges.map(({ columnIndices }) => columnIndices),
+  };
```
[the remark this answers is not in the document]
anna-shakhova: current implementation keeps existing behavior

packages/devextreme/js/__internal/grids/data_grid/summary/extenders/summary_data_controller.ts
```diff
@@ -11,14 +11,15 @@ import type { RawItemData } from '@ts/grids/grid_core/data_source_adapter/types'
 import type { ModuleType, OptionChanged } from '@ts/grids/grid_core/m_types';
 
 import type { ProcessGroupItemsOptions } from '../../grouping/types';
+import { isSameContinuationState } from '../../grouping/utils';
```
bit-byte0, line 14: since both `summary` and `grouping` now consume `isSameContinuationState`, and it's really the continuation half of `isSameGroupRowState`, would it fit better in `grid_core/data_controller/utils` so `summary` doesn't reach into `grouping`?

packages/devextreme/js/__internal/grids/data_grid/summary/utils.ts
```diff
@@ -1,5 +1,12 @@
 import { isDefined } from '@js/core/utils/type';
 import type { Column } from '@ts/grids/data_grid/types';
+import type { ProcessedItem } from '@ts/grids/grid_core/data_controller/types';
+
+import { DATAGRID_GROUP_FOOTER_ROW_TYPE } from './const';
+
+export function hasSummaryCells(item: ProcessedItem): boolean {
```
bit-byte0: `hasSummaryCells` reads like it checks `item.summaryCells`, but it's really a `rowType` predicate - would `isSummaryRow` say the intent more directly?

Alyar666: I agree that the current name is not quite accurate.

If we need to determine which rows can have summary cells, I suggest naming it `canHaveSummaryCells`.

Alternatively, we could name the method `isGroupRow` / `isGroupOrGroupFooterRow` and move it to the grouping utils.

packages/devextreme/js/__internal/grids/data_grid/summary/extenders/summary_data_controller.ts
anna-shakhova: It is not core functionality - it could be placed in some common utils in data_grid, but still not sure about it - looks like grouping leaks into summary in some methods and vice versa grouping handles some cases for summary. They should be unbound, but it is quite big task and out of scope of this task

Also summary already imports types from grouping module

bit-byte0 (approved):

PR #34938, opened 2026-08-26
GridCore data: Merge `DataHelperMixin` into `DataController`

bit-byte0:
### What
Removes the grid_core-only `DataHelperMixin` and folds its data-source lifecycle logic directly into **DataController**, so the controller no longer carries unused generic mixin machinery. Also fixes a pre-existing case where the shared-data-source flag stayed set after switching to a non-shared source

### How
Inlined the live parts (`postInit`, `_refreshDataSource`, ready-watcher wiring) into *data_controller.ts* and dropped the dead generic paths that grid_core never exercised. The raw `DataSource` is now built by a local `_createRawDataSource` helper so `_dataSource` only ever holds the adapter, and `_isSharedDataSource` is now assigned on every path


packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -61,11 +63,20 @@ import {
...
-export class DataController extends DataHelperMixin(modules.Controller) {
+export class DataController extends modules.Controller {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public _dataSource?: any;
+
+  protected _isSharedDataSource?: boolean;
+
+  protected readyWatcher?: (isLoading: boolean) => void;
+
+  protected _ready?: (value?: boolean) => void;
```
Tucchhaa: Looks like the 'ready' callback is never set and used. If that's true, let's remove `readyWatcher` and `_ready`

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -683,11 +709,40 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+
+    this._disposeDataSource();
+
+    if (!dataSourceOptions) {
+      return undefined;
+    }
+
+    const isSharedDataSource = dataSourceOptions instanceof DataSourceClass;
+
+    this._isSharedDataSource = isSharedDataSource;
```
Tucchhaa: optional code-style suggestion, use just this._isSharedDataSource prop and delete var:
```ts
this._isSharedDataSource = dataSourceOptions instanceof DataSourceClass;
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -662,14 +676,26 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     return dataSource;
   }
 
+  public postInit(): void {
+    this.on('disposing', () => {
+      this._disposeDataSource();
+    });
+  }
```
Tucchhaa: Do we really need postInit() method? I see that `dataController.dispose()` already calls `_disposeDataSource()`

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -662,14 +676,26 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+
+  /**
+   * @extended: state_storing, virtual_scrolling
+   */
+  protected _refreshDataSource(): DeferredObj<unknown> | undefined {
```
Tucchhaa: What do you think about renaming the method to `resetDataSource()`?
I think that name expresses what this function does more clearly than the current name

packages/devextreme/js/__internal/grids/grid_core/search/m_search.ts
```diff
@@ -63,7 +63,7 @@ const dataController = (
   }
 
   protected _calculateAdditionalFilter(): DataFilter {
-    const dataSource = this._dataController?.getDataSource?.();
+    const dataSource = this.getDataSource?.();
```
Tucchhaa:
```js
 this.getDataSource();
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
bit-byte0: i don't object about it

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -683,11 +709,40 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     }
   }
 
+  private _createRawDataSource(): DataSource | undefined {
```
anna-shakhova: lets remove _ for added fields and methods (except _dataSource may be - if it assumes multiple changes through code base)

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -683,11 +709,40 @@ export class DataController extends DataHelperMixin(modules.Controller) {
...
+    }
+
+    const isSharedDataSource = dataSourceOptions instanceof DataSourceClass;
+
+    this._isSharedDataSource = isSharedDataSource;
+
+    const dataSource = (isSharedDataSource
+      ? dataSourceOptions
+      : new DataSourceClass(
+        extend(true, {}, normalizeDataSourceOptions(dataSourceOptions, { fromUrlLoadMode: false })),
```
anna-shakhova: a bit difficult to read, after readyWatcher removed, then lets use early return
+ { fromUrlLoadMode: false } is excessive, empty options object can be passed
```suggestion
        extend(true, {}, normalizeDataSourceOptions(dataSourceOptions, {})),
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
anna-shakhova: incorrect naming - this method not only creates new raw data source, but also disposes existing one

Tucchhaa (dismissed):

anna-shakhova (dismissed):

anna-shakhova (dismissed):

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -686,11 +698,30 @@ export class DataController extends DataHelperMixin(modules.Controller) {
     }
```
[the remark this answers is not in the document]
bit-byte0: pre-existing issue

Tucchhaa (dismissed):

Tucchhaa (dismissed):

anna-shakhova (dismissed):

anna-shakhova (approved):

PR #34941, opened 2026-08-26
Core(Events): remove m_ prefix from emitter & pointer event files

EugeniyKiyashko:


EugeniyKiyashko:
Superseded: reworked into a single PR #34940 with two commits in history-preserving order (rename first as pure git mv, then the ES6 conversion).

PR #34986, opened 2026-08-31
DataGrid: extract the row update path from dataController

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/row_changes.ts
```diff
@@ -256,3 +248,98 @@ export function partialUpdateRow(
...
+      return { changeType: 'remove', rowIndex: visibleRowIndex, item };
+    case 'replace':
+      items[rowIndex] = item;
+      return { changeType: 'update', rowIndex: visibleRowIndex, item };
+    default:
+      return undefined;
+  }
+}
+
+export function applyRowOperations(
```
anna-shakhova: suggestion: return UpdateRowChange[], initiated with plain [] instead of initChangedRows - remove one step in utility reading

in `attachChangedRows` take as 2nd argument UpdateRowChange[] and convert it inside method (we already have `toChangedRows`, just move it inside `attachChangedRows` instead of argument

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/row_changes.ts
```diff
@@ -256,3 +248,98 @@ export function partialUpdateRow(
 
   oldItem.update?.(newItem);
 }
+
+function updateRow(rowIndex: number, options: RowOperationOptions): UpdateRowChange {
```
anna-shakhova: function updates row and create UpdateRowChange instance. Let's split it, updateRow will only update data, symmetrical to `applyItemChange`
in fact, are those methods differ? may we use single one for both branches?

Alyar666: I rewrote this method, unified it, and used it in both `applyRowOperation` and `applyItemChange`. I also renamed it to `partialUpdateRow`, and renamed the previous `partialUpdateRow` to `partialUpdateRowCore`.

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/__tests__/row_changes.test.ts
```diff
@@ -463,61 +475,50 @@ describe('convertToUpdateChange', () => {
...
-      { changeType: 'update', rowIndex: 1, item },
-    ]);
-
-    const updateChange = change as UpdateChange;
-    expect(updateChange.items).toEqual([item]);
-    expect(updateChange.rowIndices).toEqual([0, 1]);
-    expect(updateChange.changeTypes).toEqual(['remove', 'update']);
-    expect(updateChange.columnIndices).toEqual([undefined, undefined]);
+describe('partialUpdateRow', () => {
+  const partialUpdate = (
```
anna-shakhova: nitpick: this alias makes tests less readable, better use whole configuration in tests
may be fixed in next PR

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/row_changes.ts
```diff
@@ -258,6 +219,130 @@ export function partialUpdateRow(
...
+
+  return {
+    changeType: 'update',
+    rowIndex: visibleRowIndex,
+    item: newItem,
+    columnIndices,
+  };
+}
+
+function applyRowOperation(
```
anna-shakhova: I would keep this as class method, as it mutates this._items and is symmetrical to applyItemChange
also moving it to utils does not make code easier to read and we have to pass this._items, this.getUpdatedColumnIndices as parameters while in method we can use them in place

anna-shakhova (dismissed):

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/row_changes.ts
```diff
@@ -169,68 +177,61 @@ export function getRowOperation(
...
-  attachChangedRows(change, changedRows);
+  if (newItem) {
+    newItem.rowIndex = index;
+  }
 
-  return changedRows;
+  switch (getRowOperation(items, newItems, index)) {
+    case 'update':
+      if (oldItem.visible !== newItem.visible) {
+        return { type: 'visibility', index, data: newItem };
```
anna-shakhova: other types are verbs, I suggest `updateVisibility` - this way it is clear subcategory of update and also corresponding verb-convention
```suggestion
        return { type: 'updateVisibility', index, data: newItem };
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -849,63 +851,79 @@ export class DataController extends modules.Controller {
...
+      }
 
-    const columnIndices = isPartialUpdate
-      ? this.getUpdatedColumnIndices(oldItem, newItem, visibleRowIndex)
-      : undefined;
+      prevRowIndex = rowIndex;
 
-    partialUpdateRow(oldItem, newItem, columnIndices);
+      const itemChange = getItemChange(this._items, options.newItems, rowIndex);
+      const changedItem = itemChange && this.applyItemChange(itemChange, options);
```
anna-shakhova: it is a bit difficult to read ItemChange && UpdateItemChange
I suggest to add early return for undefined itemChange here or inside this.applyItemChange

anna-shakhova (approved):

PR #34987, opened 2026-08-31
DataGrid: split updateItems and _updateItemsCore into steps

Alyar666:


packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/row_changes.ts
```diff
@@ -256,3 +257,37 @@ export function partialUpdateRow(
...
+  repaintChangesOnly: boolean | undefined,
+): boolean | undefined {
+  if (!operationTypes) {
+    return undefined;
+  }
+
+  return !operationTypes.grouping && !operationTypes.filtering && repaintChangesOnly;
+}
+
+export function syncRowsAfterChange(items: ProcessedItem[], options: SyncRowsOptions): void {
```
Tucchhaa: optional suggestion: define `options` type inline:
```ts
export function syncRowsAfterChange(items: ProcessedItem[], options: {
  newItems: ProcessedItem[];
  oldItems: ProcessedItem[] | null;
  rowIndexDelta: number;
})
```
to avoid types.ts pollution, since `syncRowsAfterChange` is a small util that is used only once

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1146,55 +1148,43 @@ export class DataController extends modules.Controller {
...
-        if (useProcessedItemsCache && cachedProcessedItems) {
-          return cachedProcessedItems;
-        }
+    if (!this._dataSource) {
+      this._items = [];
+      return;
+    }
 
-        // change.items at this stage is defined only if virtualScrolling
-        // + legacyScrollingMode enabled
```
Tucchhaa, line 1164: why this comment was removed? Maybe bring it back?

packages/devextreme/js/__internal/grids/grid_core/data_controller/__tests__/data_controller.update_items.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  afterEach(afterTest);
+
+  describe('_updateItemsCore', () => {
+    it('should carry the rendered cells over to the rows a full change brought', async () => {
+      const dataController = await createGrid();
+      const cells: Cell[] = [{ column: { index: 0 } }];
+
+      dataController._items[0].cells = cells;
+
+      dataController._updateItemsCore(refreshChange());
```
Tucchhaa: I am not sure if it's a good idea to test private/protected methods.

I see 2 options:
1. Test `dataController.updateItems`, this method is public
2. Write integration tests

Also you can use `dataController.items()` instead of `._items`

What do you think?

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1216,35 +1206,49 @@ export class DataController extends modules.Controller {
...
+    if (!isDataChanged) {
       return;
     }
 
-    this._updateItemsCore(change);
+    const operationTypes: OperationTypes | undefined = this.dataSource().operationTypes();
 
-    if (change.cancel) return;
+    change.isDataChanged = true;
+    change.repaintChangesOnly = resolveRepaintChangesOnly(
```
Tucchhaa, line 1248: Is it necessary to move `resolveRepaintChangesOnly` to utils?

I think it shouldn't actually be a util, since `resolveRepaintChangesOnly` doesn't isolate calculation of `change.repaintChangesOnly` completely, but only one case, so it hardly improves readability in my opinion

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1146,55 +1148,43 @@ export class DataController extends modules.Controller {
...
+    if (!this._dataSource) {
+      this._items = [];
+      return;
+    }
 
-        // change.items at this stage is defined only if virtualScrolling
-        // + legacyScrollingMode enabled
-        const dataItems = this._beforeProcessItems(change.items ?? dataSource.items());
-        const processedItems = this._processItems(dataItems, change);
+    const items = this._afterProcessItems(this.getProcessedItems(change));
```
```suggestion
    const newItems = this._afterProcessItems(this.getProcessedItems(change));
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1146,55 +1148,43 @@ export class DataController extends modules.Controller {
...
+    const oldItems = this._items.length === items.length ? this._items : null;
```
anna-shakhova: can be moved inside syncRowsAfterChange - anyway we pass this._items and newItems, then we can pass as args newItems, items and delta

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1216,35 +1206,49 @@ export class DataController extends modules.Controller {
...
+    this.prepareChange(change, isDataChanged);
+
+    if (this._updateLockCount && !change.cancel) {
+      this.changes.push(change);
+      return;
+    }
+
+    this._updateItemsCore(change);
+
+    if (change.cancel) return;
```
```suggestion
    if (change.cancel) {
      return;
    }
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
```diff
@@ -1216,35 +1206,49 @@ export class DataController extends modules.Controller {
...
+
+    this._fireChanged(change);
+  }
+
+  private prepareChange(change: DataChange, isDataChanged?: boolean): void {
```
```suggestion
  private fillChangeFlags(change: DataChange, isDataChanged?: boolean): void {
```

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
Alyar666: This one can't move inside the util, unfortunately — the ternary has to be evaluated *before* `_applyChange`, because that method touches `this._items` in two different ways:

- `applyChangeFull` **replaces** the array: `this._items = (change.items ?? []).slice()`. After it, `this._items[i] === newItems[i]`, so inside the util the comparison would degrade to `newItems.length === newItems.length` (always `true`) and the "old" rows would be the new ones, whose `cells` are `undefined` — every row would get `cells = []` and the rendered-cell reuse would be gone entirely.
- `applyChangesOnly` **mutates** the array in place (`applyItemChange` does `this._items[index] = newItem` and `splice`). There `oldItems === this._items`, so after the mutation the original length is no longer available: the check would again be always `true`, and rows inserted by `splice` would get `cells = []` instead of `undefined`. That difference is observable — `updateRowCells` and `getGroupColumnIndices` branch on `if (!oldItem.cells)`, and `partialUpdateRow` on `oldItem.cells?.forEach`.

packages/devextreme/js/__internal/grids/grid_core/data_controller/utils/row_changes.ts
```diff
@@ -256,3 +257,37 @@ export function partialUpdateRow(
...
+      item.cells = oldItems[index].cells ?? [];
+    }
+
+    const newItem = newItems[index];
+
+    if (newItem) {
+      item.loadIndex = newItem.loadIndex;
+    }
+  });
+}
```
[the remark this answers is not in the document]
Alyar666: `oldItems` is non-`null` only when `this._items.length === items.length`, so an empty `oldItems` implies an empty `items` and the loop body never runs — the out-of-range case is unreachable here. The code is also unchanged from `main`, so a guard would silently alter an edge case in a refactor that is meant to preserve behavior.

packages/devextreme/js/__internal/grids/grid_core/data_controller/data_controller.ts
Alyar666: I'd keep it. This makes the third state explicit, which was previously hidden in the inline operationTypes && … chain: undefined means that the mode is not set, so a subsequent refresh({ changesOnly }) can still set it - it's not equivalent to false.

Extracting this logic into a separate function also made it possible to test all three states (undefined / false / true) in row_changes.test.ts.

anna-shakhova (approved):

PR #34994, opened 2026-08-31
Map: OSM provider - Add markers and location calculation

AlisherAmonulloev:

### DIFF
--- a/packages/devextreme-scss/scss/widgets/generic/map/_colors.scss
+++ b/packages/devextreme-scss/scss/widgets/generic/map/_colors.scss
@@ -1,0 +1,2 @@
+$map-marker-color: #337ab7 !default;
+$map-marker-accent-color: #fff !default;
--- a/packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
+++ b/packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
@@ -3,8 +3,13 @@ import type { MapEngineMarkerOptions } from './provider.dynamic.osm.engine';
 export const MARKER_FALLBACK_WIDTH = 25;
 export const MARKER_FALLBACK_HEIGHT = 41;
 export const DEFAULT_MARKER_SIZE = 44;
-export const DEFAULT_MARKER_CLASS = 'dx-map-marker-default';
 
+const MARKER_CLASS = 'dx-map-marker';
+export const DEFAULT_MARKER_CLASS = `${MARKER_CLASS}-default`;
+const DEFAULT_MARKER_ICON_CLASS = `${DEFAULT_MARKER_CLASS}-icon`;
+const DEFAULT_MARKER_BODY_CLASS = `${DEFAULT_MARKER_CLASS}-body`;
+const DEFAULT_MARKER_CENTER_CLASS = `${DEFAULT_MARKER_CLASS}-center`;
+const IMAGE_MARKER_CLASS = `${MARKER_CLASS}-image`;
 const DEFAULT_MARKER_WIDTH = 24.5;
 const DEFAULT_MARKER_HEIGHT = 36.5;
 const SVG_NAMESPACE = 'http://www.w3.org/2000/svg';
@@ -30,13 +35,13 @@ const createDefaultMarkerElement = (
   const body = ownerDocument.createElementNS(SVG_NAMESPACE, 'path');
   const center = ownerDocument.createElementNS(SVG_NAMESPACE, 'path');
 
-  element.className = `dx-map-marker ${DEFAULT_MARKER_CLASS}`;
+  element.className = `${MARKER_CLASS} ${DEFAULT_MARKER_CLASS}`;
   element.style.alignItems = 'flex-end';
   element.style.display = 'flex';
   element.style.height = `${DEFAULT_MARKER_SIZE}px`;
   element.style.justifyContent = 'center';
   element.style.width = `${DEFAULT_MARKER_SIZE}px`;
-  svg.setAttribute('class', 'dx-map-marker-default-icon');
+  svg.setAttribute('class', DEFAULT_MARKER_ICON_CLASS);
   svg.setAttribute('viewBox', DEFAULT_MARKER_VIEW_BOX);
   svg.setAttribute('width', `${DEFAULT_MARKER_WIDTH}`);
   svg.setAttribute('height', `${DEFAULT_MARKER_HEIGHT}`);
@@ -48,10 +53,10 @@ const createDefaultMarkerElement = (
     'drop-shadow(0 1px 1.5px rgba(0, 0, 0, 0.35))',
   ].join(' ');
   svg.style.overflow = 'visible';
-  body.setAttribute('class', 'dx-map-marker-default-body');
+  body.setAttribute('class', DEFAULT_MARKER_BODY_CLASS);
   body.setAttribute('d', DEFAULT_MARKER_BODY_PATH);
   body.setAttribute('stroke-width', '0.5');
-  center.setAttribute('class', 'dx-map-marker-default-center');
+  center.setAttribute('class', DEFAULT_MARKER_CENTER_CLASS);
   center.setAttribute('d', DEFAULT_MARKER_CENTER_PATH);
   svg.appendChild(body);
   svg.appendChild(center);
@@ -77,7 +82,7 @@ export const createMarkerElement = (
 
   if (options.iconSrc) {
     const element = ownerDocument.createElement('img');
-    element.className = 'dx-map-marker dx-map-marker-image';
+    element.className = `${MARKER_CLASS} ${IMAGE_MARKER_CLASS}`;
     element.src = options.iconSrc;
     element.alt = '';
     element.draggable = false;
--- a/packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
+++ b/packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
@@ -1,0 +1,98 @@
+import type { MapEngineMarkerOptions } from './provider.dynamic.osm.engine';
+
+export const MARKER_FALLBACK_WIDTH = 25;
+export const MARKER_FALLBACK_HEIGHT = 41;
+export const DEFAULT_MARKER_SIZE = 44;
+export const DEFAULT_MARKER_CLASS = 'dx-map-marker-default';
+
+const DEFAULT_MARKER_WIDTH = 24.5;
+const DEFAULT_MARKER_HEIGHT = 36.5;
+const DEFAULT_MARKER_COLOR = '#2d7fbd';
+const DEFAULT_MARKER_ACCENT_COLOR = '#fff';
+const SVG_NAMESPACE = 'http://www.w3.org/2000/svg';
+const DEFAULT_MARKER_VIEW_BOX = '5 2 14 20';
+const DEFAULT_MARKER_BODY_PATH = [
+  'M12 2c-3.85 0-7 3.176-7 7.059',
+  ' 0 5.294 7 12.941 7 12.941',
+  's7-7.765 7-12.941',
+  'c0-3.882-3.15-7.059-7-7.059z',
+].join('');
+const DEFAULT_MARKER_CENTER_PATH = [
+  'M12 6.706c-1.283 0-2.333 1.059-2.333 2.353',
+  's1.05 2.353 2.333 2.353',
+  ' 2.333-1.059 2.333-2.353',
+  'c0-1.294-1.05-2.353-2.333-2.353z',
+].join('');
+
+const createDefaultMarkerElement = (
+  ownerDocument: Document,
+): HTMLElement => {
+  const element = ownerDocument.createElement('div');
+  const svg = ownerDocument.createElementNS(SVG_NAMESPACE, 'svg');
+  const body = ownerDocument.createElementNS(SVG_NAMESPACE, 'path');
+  const center = ownerDocument.createElementNS(SVG_NAMESPACE, 'path');
+
+  element.className = `dx-map-marker ${DEFAULT_MARKER_CLASS}`;
+  element.style.alignItems = 'flex-end';
+  element.style.display = 'flex';
+  element.style.height = `${DEFAULT_MARKER_SIZE}px`;
+  element.style.justifyContent = 'center';
+  element.style.width = `${DEFAULT_MARKER_SIZE}px`;
+  svg.setAttribute('class', 'dx-map-marker-default-icon');
+  svg.setAttribute('viewBox', DEFAULT_MARKER_VIEW_BOX);
+  svg.setAttribute('width', `${DEFAULT_MARKER_WIDTH}`);
+  svg.setAttribute('height', `${DEFAULT_MARKER_HEIGHT}`);
+  svg.setAttribute('aria-hidden', 'true');
+  svg.setAttribute('focusable', 'false');
+  svg.style.display = 'block';
+  svg.style.filter = [
+    'drop-shadow(0 0 0.5px rgba(255, 255, 255, 0.85))',
+    'drop-shadow(0 1px 1.5px rgba(0, 0, 0, 0.35))',
+  ].join(' ');
+  svg.style.overflow = 'visible';
+  body.setAttribute('class', 'dx-map-marker-default-body');
+  body.setAttribute('d', DEFAULT_MARKER_BODY_PATH);
+  body.setAttribute('fill', DEFAULT_MARKER_COLOR);
+  body.setAttribute('stroke', DEFAULT_MARKER_ACCENT_COLOR);
+  body.setAttribute('stroke-width', '0.5');
+  center.setAttribute('class', 'dx-map-marker-default-center');
+  center.setAttribute('d', DEFAULT_MARKER_CENTER_PATH);
+  center.setAttribute('fill', DEFAULT_MARKER_ACCENT_COLOR);
+  svg.appendChild(body);
+  svg.appendChild(center);
+  element.appendChild(svg);
+
+  return element;
+};
+
+export const createMarkerElement = (
+  ownerDocument: Document,
+  options: MapEngineMarkerOptions,
+): { element: HTMLElement; offset: number[]; positioning: string } => {
+  if (options.html) {
+    const element = ownerDocument.createElement('div');
+    element.innerHTML = options.html;
+
+    return {
+      element,
+      offset: [options.htmlOffset?.left ?? 0, options.htmlOffset?.top ?? 0],
+      positioning: 'top-left',
+    };
+  }
+
+  if (options.iconSrc) {
+    const element = ownerDocument.createElement('img');
+    element.className = 'dx-map-marker dx-map-marker-image';
+    element.src = options.iconSrc;
+    element.alt = '';
+    element.draggable = false;
+
+    return { element, offset: [0, 0], positioning: 'bottom-center' };
+  }
+
+  return {
+    element: createDefaultMarkerElement(ownerDocument),
+    offset: [0, 0],
+    positioning: 'bottom-center',
+  };
+};
--- a/packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
+++ b/packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
@@ -7,8 +7,6 @@ export const DEFAULT_MARKER_CLASS = 'dx-map-marker-default';
 
 const DEFAULT_MARKER_WIDTH = 24.5;
 const DEFAULT_MARKER_HEIGHT = 36.5;
-const DEFAULT_MARKER_COLOR = '#2d7fbd';
-const DEFAULT_MARKER_ACCENT_COLOR = '#fff';
 const SVG_NAMESPACE = 'http://www.w3.org/2000/svg';
 const DEFAULT_MARKER_VIEW_BOX = '5 2 14 20';
 const DEFAULT_MARKER_BODY_PATH = [
@@ -52,12 +50,9 @@ const createDefaultMarkerElement = (
   svg.style.overflow = 'visible';
   body.setAttribute('class', 'dx-map-marker-default-body');
   body.setAttribute('d', DEFAULT_MARKER_BODY_PATH);
-  body.setAttribute('fill', DEFAULT_MARKER_COLOR);
-  body.setAttribute('stroke', DEFAULT_MARKER_ACCENT_COLOR);
   body.setAttribute('stroke-width', '0.5');
   center.setAttribute('class', 'dx-map-marker-default-center');
   center.setAttribute('d', DEFAULT_MARKER_CENTER_PATH);
-  center.setAttribute('fill', DEFAULT_MARKER_ACCENT_COLOR);
   svg.appendChild(body);
   svg.appendChild(center);
   element.appendChild(svg);
--- a/packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
+++ b/packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
@@ -1,7 +1,10 @@
+import messageLocalization from '@js/common/core/localization/message';
+
 import type { MapEngineMarkerOptions } from './provider.dynamic.osm.engine';
 
 export const MARKER_FALLBACK_WIDTH = 25;
 export const MARKER_FALLBACK_HEIGHT = 41;
+// NOTE: Keep this value in sync with .dx-map-marker-default in widgets/base/_map.scss.
 export const DEFAULT_MARKER_SIZE = 44;
 
 const MARKER_CLASS = 'dx-map-marker';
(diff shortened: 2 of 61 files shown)


arman-boyakhchyan (approved):

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
```diff
@@ -234,18 +425,108 @@ class OsmProvider extends DynamicProvider {
...
+      // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
+      const iconSrc = options.iconSrc || this._option('markerIconSrc');
+      const markerClickAction = options.onClick
+        ? this._mapWidget._createAction(options.onClick)
+        : undefined;
+      const engineMarker = engineMap.addMarker({
+        html: options.html,
+        htmlOffset: options.htmlOffset,
+        iconSrc,
+        location,
```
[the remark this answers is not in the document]
AlisherAmonulloev: this PR deliberately does not include Tooltip support. This stage implements markers and location calculation

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -90,18 +130,232 @@ class OpenLayersMap implements MapEngineMap {
         zoom: view.zoom ?? 1,
       }),
     });
+    this.originalMap.getOverlayContainer().setAttribute('dir', 'ltr');
+    this.originalMap.getOverlayContainerStopEvent().setAttribute('dir', 'ltr');
+    this.originalMap.getView().on('change:center', this._viewCenterChangeHandler);
```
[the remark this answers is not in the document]
AlisherAmonulloev: fixed in https://github.com/DevExpress/DevExtreme/pull/34994/commits/47073363cc2b94c14d156da9bf828bf888c925f4

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -90,18 +133,247 @@ class OpenLayersMap implements MapEngineMap {
...
+        }
+
+        event.preventDefault();
+        event.stopPropagation();
+        element.click();
+      }
+      : undefined;
+    const imageLoadHandler: EventListener | undefined = element.tagName === 'IMG'
+      ? (): void => this._eventHandlers?.markerSizeChange()
+      : undefined;
```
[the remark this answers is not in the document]
AlisherAmonulloev: fixed in https://github.com/DevExpress/DevExtreme/pull/34994/commits/f478f57b9d590aae274287faf4aa9d5846386377

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
```diff
@@ -45,6 +124,29 @@ class OsmProvider extends DynamicProvider {
...
 
+  _calculatedLocations = new Map<string, MapLocation>();
+
+  _pendingLocationCalculations = new Map<string, Promise<MapLocation>>();
+
+  _boundLocations: MapLocation[] = [];
+
+  _generation = 0;
+
+  render(markers: MarkerOptions[], routes: RouteOptions[]): Promise<unknown> {
```
EugeniyKiyashko: The stale-result guard (`_generation`, this `render` override, `STALE_OPERATION`, the `engineMap !== this._engineMap` checks) works around a race in `map.ts`: `_queueAsyncAction` calls `this._provider[name]` and, when the promise settles, reads `this._provider.map()` of whatever provider is current at that moment. After `_invalidate` that is a new instance, so `onReady` fires with the wrong map. Google and Azure have the same race.

Suggestion: fix it once in `_queueAsyncAction` - capture `const provider = this._provider` before the call and ignore the result and the error when `provider !== this._provider || this._disposed`. Then `_generation`, this override and the `STALE_OPERATION` handling in `addMarkers` can be removed. Only the guard in `_renderMarker` (do not add an overlay to a disposed map) needs to stay.

EugeniyKiyashko (commented):
Review of the solution, not only of the diff. Details are in the line comments. Follow-ups that need no change here:

- Marker keyboard and ARIA handling lives in the OpenLayers adapter. If a second engine is planned, the element-level part belongs next to `createMarkerElement`.
- When tooltips are added, `keyboardInteractive` must include tooltip-only markers - Google makes them clickable.
- The story is tagged `!test` and there is no demo or e2e for OSM, so the default marker look in the three themes has no visual guard yet.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
```diff
@@ -119,13 +227,74 @@ class OsmProvider extends DynamicProvider {
...
+      && Number.isFinite(resolvedLocation.lng)) {
+      return Promise.resolve(resolvedLocation);
+    }
+
+    return typeof location === 'string'
+      ? this._calculateLocation(location)
+      : Promise.resolve(getDefaultLocation());
+  }
+
+  _calculateLocation(query: string): Promise<MapLocation> {
```
EugeniyKiyashko, line 224: `DynamicProvider` already has the geocoding hook: `_geocodeLocation` with the `_geocodedLocations` cache, and providers implement `_geocodeLocationImpl`. This method rebuilds it with a second cache and a pending map.

Pending-request deduplication and result validation are useful for every provider. Suggestion: move them into `DynamicProvider._geocodeLocation` (make the location type generic instead of `GoogleLocation | BingLocation | AzureLocation`) and implement only `_geocodeLocationImpl` here via `providerConfig.calculateLocation`.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
```diff
@@ -119,13 +227,74 @@ class OsmProvider extends DynamicProvider {
...
+      return Promise.resolve(cachedLocation);
+    }
+    const pendingCalculation = this._pendingLocationCalculations.get(query);
+    if (pendingCalculation) {
+      return pendingCalculation;
+    }
+
+    const calculateLocation = this._option('providerConfig')?.calculateLocation;
+    if (!calculateLocation) {
+      errors.log('W1031');
```
EugeniyKiyashko: W1031 is logged for every string location on every call, and the miss is not cached: 100 markers with addresses and no `calculateLocation` produce 100 identical warnings. Log it once per provider instance.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
```diff
@@ -119,13 +227,74 @@ class OsmProvider extends DynamicProvider {
...
+          const result = { lat: location.lat, lng: location.lng };
+          if (generation === this._generation) {
+            this._calculatedLocations.set(query, result);
+          }
+
+          return result;
+        }
+
+        return getDefaultLocation();
+      }, () => getDefaultLocation());
```
EugeniyKiyashko: A rejected or invalid `calculateLocation` result silently becomes `{ lat: 0, lng: 0 }`, so a broken callback puts every marker at Null Island with nothing in the console. Google logs W1006 in the same situation. Log a warning here as well.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
```diff
@@ -119,13 +227,74 @@ class OsmProvider extends DynamicProvider {
...
+      }
+
+      return location;
+    });
   }
 
   _attachHandlers(): void {
     this._engineMap?.attachHandlers({
       click: (event) => this._clickActionHandler(event),
+      markerSizeChange: () => {
```
EugeniyKiyashko, line 261: Every `load` event from a marker image, and `updateDimensions` while `needsLayout` is set, calls `_fitBounds`, which moves the viewport and silently rewrites `zoom`. If an image finishes loading after the user has already panned or zoomed, the map jumps back to the markers. Google fits only in `addMarkers`, `addRoutes` and `adjustViewport`.

Suggestion: refit on size changes only until the first user-initiated move, and reset that on the next markers change or `adjustViewport` call. Wrap own `fit`/`setCenter`/`setZoom` calls in a flag and treat any other `moveend` as user-initiated.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -90,18 +133,288 @@ class OpenLayersMap implements MapEngineMap {
         zoom: view.zoom ?? 1,
       }),
     });
+    this.originalMap.getOverlayContainer().setAttribute('dir', 'ltr');
```
EugeniyKiyashko, line 143: `dir="ltr"` on both overlay containers is inherited by user HTML markers, so in an RTL application their text renders left-to-right. The RTL test checks the attribute only, not the marker content.

Keep the containers LTR for positioning, but set `dir` on the marker root from the widget's `rtlEnabled`.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -90,18 +133,288 @@ class OpenLayersMap implements MapEngineMap {
...
+        if (keyboardEvent.key !== ' ') {
+          return;
+        }
+
+        event.preventDefault();
+        event.stopPropagation();
+        element.click();
+      }
+      : undefined;
+    const hasImage = element.tagName === 'IMG' || element.querySelector('img') !== null;
```
EugeniyKiyashko: The capture `load` listener fires for any descendant that loads (img, iframe, video, lazy content) for the whole lifetime of the marker, and it misses size changes that are not loads (fonts, async content). Once the refit is bounded in time (see `markerSizeChange` in the provider), a `ResizeObserver` on the marker root is a simpler and more complete signal. Disconnect it in `detach`.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -90,18 +133,288 @@ class OpenLayersMap implements MapEngineMap {
...
+  }
+
+  addMarker(options: MapEngineMarkerOptions): MapEngineMarker {
+    const { element, offset, positioning } = createMarkerElement(
+      this._container.ownerDocument,
+      options,
+    );
+    const marker = new this._api.Overlay({
+      element,
+      insertFirst: true,
```
EugeniyKiyashko: `insertFirst: true` (the OpenLayers default) inserts each new overlay before the existing ones, so a marker added at runtime with `addMarker` is drawn under the older markers where they overlap. Google and Leaflet put newer markers on top. Use `insertFirst: false`.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -90,18 +133,288 @@ class OpenLayersMap implements MapEngineMap {
...
+    if (Number.isFinite(worldWidth) && worldWidth > 0) {
+      coordinate[0] += Math.round((center[0] - coordinate[0]) / worldWidth) * worldWidth;
+    }
+
+    return coordinate;
+  }
+
+  private _syncMarkerPositions(): void {
+    this._markers.forEach((marker) => {
+      marker.overlay.setPosition(this._getMarkerPosition(marker.location));
```
EugeniyKiyashko: `change:center` fires on every frame of a drag or animation, and `setPosition` receives a new array each time. OpenLayers compares with `!==`, so every marker gets a change notification and a DOM write on every frame even when nothing moved, in addition to the repositioning OpenLayers already does on `postrender`.

Compare with `areCoordinatesEqual` (already in utils) and call `setPosition` only when the wrapped coordinate changed.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -90,18 +133,288 @@ class OpenLayersMap implements MapEngineMap {
...
+      container.focus({ preventScroll: true });
+    }
+  }
+
+  private _getMarkerFitPadding(): { padding: number[]; needsLayout: boolean } {
+    const padding = [0, 0, 0, 0];
+    let needsLayout = false;
+
+    this._markers.forEach(({ element, offset, positioning }) => {
+      const isDefault = element.classList.contains(DEFAULT_MARKER_CLASS);
```
EugeniyKiyashko: The marker type is derived from a CSS class here. A `kind: 'default' | 'image' | 'html'` field on the marker handle, set by `createMarkerElement`, is more direct and does not depend on class names.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -305,10 +647,17 @@ class OpenLayersMap implements MapEngineMap {
     if (options.zoom !== undefined && view.getZoom() !== options.zoom) {
       view.setZoom(options.zoom);
     }
+
+    this._syncMarkerTabIndexes();
   }
 
-  updateDimensions(): void {
+  updateDimensions(): boolean {
```
EugeniyKiyashko: `updateDimensions` returning `true` means "the provider must refit the viewport". A named method (`needsViewportRefit()`) or a `layoutChange` entry in `MapEngineEventHandlers` reads better than a boolean with this meaning.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
```diff
@@ -0,0 +1,98 @@
...
+      offset: [options.htmlOffset?.left ?? 0, options.htmlOffset?.top ?? 0],
+      positioning: 'top-left',
+    };
+  }
+
+  if (options.iconSrc) {
+    const element = ownerDocument.createElement('img');
+    element.className = `${MARKER_CLASS} ${IMAGE_MARKER_CLASS}`;
+    element.src = options.iconSrc;
+    element.alt = '';
```
EugeniyKiyashko: Image markers become `` (role and label are added in `_attachMarkerElementHandlers`). `alt=""` marks the image as decorative, and ARIA in HTML allows only `none`/`presentation` on such an `img`, so axe reports `aria-allowed-role`. Non-clickable image markers get no name at all, while the Google template uses `alt="Marker icon"`.

Suggestion: set `alt` to the localized marker label for every image marker and do not add `aria-label` to `img`. `role="button"` stays for clickable markers - it is allowed on `img` with a non-empty `alt`.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
```diff
@@ -0,0 +1,98 @@
...
+
+const createDefaultMarkerElement = (
+  ownerDocument: Document,
+): HTMLElement => {
+  const element = ownerDocument.createElement('div');
+  const svg = ownerDocument.createElementNS(SVG_NAMESPACE, 'svg');
+  const body = ownerDocument.createElementNS(SVG_NAMESPACE, 'path');
+  const center = ownerDocument.createElementNS(SVG_NAMESPACE, 'path');
+
+  element.className = `${MARKER_CLASS} ${DEFAULT_MARKER_CLASS}`;
```
EugeniyKiyashko, line 50: Layout, size and the drop shadow are set inline here while the colors live in SCSS, so the default marker is styled in two places. Move the presentation into `base/_map.scss` (`.dx-map-marker-default`, `.dx-map-marker-default-icon`) and keep only `DEFAULT_MARKER_SIZE` in JS for the fit padding, with a note that it must match the CSS.

packages/devextreme-scss/scss/widgets/generic/map/_colors.scss
```diff
@@ -0,0 +1,2 @@
+$map-marker-color: #337ab7 !default;
```
EugeniyKiyashko, line 1: These values are the accent of the light blue scheme only. Generic `#337ab7` is generic.light - carmine, contrast, dark, darkmoon, darkviolet, greenmist and softblue have other accents. Fluent `#0f6cbd` is blue.light, dark uses `#479ef5`. Material `#1564c0` is blue.light - lime, orange, purple, teal and dark differ. ThemeBuilder accent changes do not reach the marker either.

Follow the other `_colors.scss` files:

```scss
@use "../colors" as *;

$map-marker-color: $base-accent !default;
$map-marker-accent-color: $base-bg !default;
```

packages/devextreme/testing/helpers/forMap/openLayersMock.js
```diff
@@ -75,8 +88,14 @@
         fit(extent, options) {
```
EugeniyKiyashko, line 88: The mock decides what `fit` means: padding is recorded but not applied, `calculateExtent` returns a preset, and `toLonLat` normalization is reimplemented by hand. All fit, padding and wrap math in this PR is verified only against this mock.

`ol@10.10.0` is already in the repo (react-storybook). Add it as a devDependency of `packages/devextreme` and cover `createBounds`, `fitBounds` padding and `_getMarkerPosition` with jest against the real OpenLayers `View`.

packages/devextreme/testing/tests/DevExpress.ui.widgets/mapParts/osmTests.js
```diff
@@ -776,13 +802,965 @@ QUnit.module('OSM: initial view', moduleConfig, () => {
...
+            lng: -73.98
+        });
+        add.then(result => {
+            assert.deepEqual(result, [false, []], 'the stale marker operation is canceled');
+            assert.ok(addMarker.notCalled, 'no marker is added to the disposed engine map');
+            done();
+        });
+    });
+});
+QUnit.module('OSM: markers', moduleConfig, () => {
```
EugeniyKiyashko, line 1188: Missing against the issue scope: disabled state with marker `tabindex`, Shadow DOM for marker focus (`_moveMarkerFocusToMap`), a map created inside a hidden container with `autoAdjust` (the hidden-map test only stubs `getBoundingClientRect`), text direction of an HTML marker in RTL, overlay order after `addMarker`, no refit after the user has moved the map, and the attributes of the `img` marker.

packages/devextreme-scss/scss/widgets/generic/map/_colors.scss
AlisherAmonulloev: marker colors were reviewed and approved by the designers. We agreed to use the primary color from each **light** theme for both light and dark variants, since the map background does not change with the DevExtreme theme. So, fixed color values here are intentional

packages/devextreme/testing/helpers/forMap/openLayersMock.js
AlisherAmonulloev: we do not want to add `ol` to packages/devextreme dependencies or devDependencies. This will affect SBOM inventory for everybody, including those who does not use OSM provider. This also might create unnecessary compliance, licensing, and vulnerability-scanner work. Real ESM integration is covered by Storybook and integration examples that have the ol dependency. QUnit tests cover the adapter contract with mocks

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
AlisherAmonulloev: agree, fixed, but required to update other providers

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
AlisherAmonulloev: fixed, will ask techwriters for adjustments of the warning text

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
AlisherAmonulloev: agree, fixed

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.marker.ts
AlisherAmonulloev: fixed

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
AlisherAmonulloev: addressed

packages/devextreme-scss/scss/widgets/generic/map/_colors.scss
EugeniyKiyashko: Understood, the light/dark rule is a design decision. One note for the designers: the fixed values match only the blue schemes. Generic carmine, contrast, greenmist and softblue, and material lime, orange, purple and teal are light schemes with another primary, so those applications get a blue marker on top of their own primary. If that is intended, nothing to change.

packages/devextreme/testing/helpers/forMap/openLayersMock.js
EugeniyKiyashko: Agreed, the SBOM argument settles the dependency question. A cheaper alternative that needs no `ol`: export `createBounds` and `normalizeLongitude` from the provider module and cover them with jest (antimeridian, single location, evenly spaced longitudes). The fit padding math stays under the QUnit mock.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -118,12 +455,42 @@ class OpenLayersMap implements MapEngineMap {
...
+    this._eventHandlers = {
+      click,
+      markerSizeChange: handlers.markerSizeChange,
+      moveEnd,
+      userKeyDown,
+      userViewChange,
+    };
     this.originalMap.on('click', click);
     this.originalMap.on('moveend', moveEnd);
+    this.originalMap.on('pointerdrag', userViewChange);
```
EugeniyKiyashko: User interaction is detected from `pointerdrag`, `wheel` and six keys. Two default ways to change the view are missed: double-click zoom (`DoubleClickZoom` stays enabled in `interaction.defaults`) and the `+`/`-` buttons of the zoom control. After either of them a late marker resize still refits the viewport.

`pointerdown` on the container covers drag, double-click, pinch and the control buttons at once and makes the `pointerdrag` subscription unnecessary.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.openlayers.ts
```diff
@@ -176,7 +549,8 @@ class OpenLayersMap implements MapEngineMap {
     this._container.removeAttribute('tabindex');
   }
 
-  fitBounds(bounds: MapEngineBounds): void {
+  fitBounds(bounds: MapEngineBounds, options?: MapEngineFitBoundsOptions): void {
+    this._markerSizeRefitEnabled = true;
```
EugeniyKiyashko: `fitBounds` is also the path of the `bounds` option - `updateBounds` calls it without options. Re-enabling the refit here means: the application sets `bounds`, a marker image finishes loading, `_fitBounds` moves the view back to the markers. Re-enable only on the autoAdjust path, i.e. when `options?.includeMarkerPadding` is set.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
```diff
@@ -234,18 +394,96 @@ class OsmProvider extends DynamicProvider {
...
+
+  _renderMarker(options: MarkerOptions): Promise<EngineMarkerObject> {
+    const engineMap = this._engineMap;
+    if (!engineMap) {
+      return Promise.reject(errors.Error('E1069'));
+    }
+
+    return this._resolveLocation(options.location).then((location) => {
+      if (engineMap !== this._engineMap) {
+        return Promise.reject(STALE_OPERATION);
```
EugeniyKiyashko: `addMarkers` no longer catches this, so the widget's action now rejects with a bare Symbol. It is swallowed only because the provider has always been replaced by then. Reject with an `Error` so the reason is readable if it ever surfaces.

EugeniyKiyashko (commented):
Re-review of `1dea538599`..`506332829e`. The earlier points are addressed: stale results are ignored in `_queueAsyncAction`, geocoding lives in `DynamicProvider`, warnings, `insertFirst`, position sync, `dir`, `alt`, `kind`, the named `updateDimensions` result, styles in SCSS and the missing tests. Two threads stay open (marker colors, `ol` dependency), replies are there. Three new points on the refit policy below.

packages/devextreme/js/__internal/ui/map/provider.dynamic.osm.ts
```diff
@@ -119,13 +208,61 @@ class OsmProvider extends DynamicProvider {
...
+    return Promise.resolve()
+      .then(() => calculateLocation(query))
+      .then((location) => {
+        if (location
+          && Number.isFinite(location.lat)
+          && Number.isFinite(location.lng)) {
+          return { lat: location.lat, lng: location.lng };
+        }
+
+        errors.log('W1006', 'The calculateLocation callback returned an invalid location.');
```
arman-boyakhchyan:
```suggestion
        errors.log('W1006', 'calculateLocation returned an invalid result.');
```

EugeniyKiyashko (approved):

PR #35009, opened 2026-09-01
Grids: encapsulate FilterSyncController's cross-module surface

anna-shakhova:

### DIFF
--- a/packages/devextreme/js/__internal/grids/grid_core/filter_sync/extenders/filter_sync_column_headers_view.ts
+++ b/packages/devextreme/js/__internal/grids/grid_core/filter_sync/extenders/filter_sync_column_headers_view.ts
@@ -1,0 +1,35 @@
+import { filterHasField } from '@ts/filter_builder/m_utils';
+import type { ColumnHeadersView } from '@ts/grids/grid_core/column_headers/m_column_headers';
+import type { DataController } from '@ts/grids/grid_core/data_controller/data_controller';
+import type { ModuleType, OptionChanged } from '@ts/grids/grid_core/m_types';
+
+import { getColumnIdentifier } from '../utils';
+import type { FilterSyncDataControllerExtension } from './filter_sync_data_controller';
+
+export const filterSyncColumnHeadersViewExtender = (
+  Base: ModuleType<ColumnHeadersView>,
+): ModuleType<ColumnHeadersView> => class ColumnHeadersViewFilterSyncExtender extends Base {
+  declare protected _dataController: DataController & FilterSyncDataControllerExtension;
+
+  public optionChanged(args: OptionChanged): void {
+    if (args.name === 'filterValue') {
+      // @ts-expect-error introduced in HeaderFilter extender
+      this._updateHeaderFilterIndicators();
+    } else {
+      super.optionChanged(args);
+    }
+  }
+
+  private _isHeaderFilterEmpty(column): boolean {
+    if (this._dataController.isFilterSyncActive()) {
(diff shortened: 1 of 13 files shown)


packages/devextreme/js/__internal/grids/grid_core/filter_sync/utils.ts
```diff
@@ -1,6 +1,191 @@
...
+};
+
+export const getColumnIdentifier = (
+  column: FilterSyncColumn,
   // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
-  return column.name || column.dataField;
-}
+): string | undefined => column.name || column.dataField;
+
+export const canSyncHeaderFilterWithFilterRow = (column: FilterSyncColumn): boolean => {
```
Alyar666: Why do we need to export this? This method seems to be used only within this file.

packages/devextreme/js/__internal/grids/grid_core/filter_sync/utils.ts
```diff
@@ -1,6 +1,191 @@
...
+  const filterValues = column.filterValues ?? [];
+  const hasOwnDataSource = Boolean(filterUtils.getGroupInterval(column))
```
Alyar666: The variable name doesn’t match what is actually being checked. I’d split it into two variables: hasGroupInterval and hasOwnDataSource.

packages/devextreme/js/__internal/grids/grid_core/filter_sync/extenders/filter_sync_column_headers_view.ts
```diff
@@ -0,0 +1,35 @@
...
+  public optionChanged(args: OptionChanged): void {
+    if (args.name === 'filterValue') {
+      // @ts-expect-error introduced in HeaderFilter extender
+      this._updateHeaderFilterIndicators();
+    } else {
+      super.optionChanged(args);
+    }
+  }
+
+  private _isHeaderFilterEmpty(column): boolean {
```
Alyar666, line 23: The argument is missing a type

Alyar666: And can we remove the underscore from the method name?

packages/devextreme/js/__internal/grids/grid_core/filter_sync/utils.ts
```diff
@@ -1,6 +1,191 @@
...
+
+  return [field, isExcluded ? 'noneof' : 'anyof', filterValues];
+};
+
+const getEmptyFilterValues = (): HeaderFilterState => ({
+  filterType: FILTER_TYPES_INCLUDE,
+  filterValues: undefined,
+});
+
+export const getHeaderFilterFromCondition = (
```
Alyar666, line 108: Are we going to add tests for the utility methods? Or is that a separate task?

Alyar666 (changes requested):

packages/devextreme/js/__internal/grids/grid_core/filter_sync/extenders/filter_sync_column_headers_view.ts
anna-shakhova: this is columnHeadersView extender, I've extracted it but not typed intentionally as base is not typed

Alyar666 (dismissed):

Alyar666 (approved):

PR #35011, opened 2026-09-02
Fluent-next: ship both colour modes in every bundle, selectable by class

EugeniyKiyashko:


packages/devextreme-scss/scss/widgets/fluent-next/_design-system.scss
```diff
@@ -11,10 +14,77 @@ $accent: colors.$color !default;
...
+    }
+  }
+}
+
+@if colors.$mode != "light" and colors.$mode != "dark" {
+  @error "fluent-next: unknown colour mode #{meta.inspect(colors.$mode)}; expected \"light\" or \"dark\".";
+}
+
+// Inverted first: a named class on the SAME element has to win, and at equal weight source order
+// is what decides.
```
Raushen: Wrong // comment block. That's why CI is red - it is needed to replace to 
/* */
Blocking linter rule: scss/double-slash-comment-empty-line-before

packages/devextreme-scss/tests/data-uri-static-markers.test.ts
```diff
@@ -67,18 +69,34 @@ const expand = (hex: string): string => {
     : value;
 };
 
-function resolve(css: string, property: string, depth = 0): string | null {
-  const found = new RegExp(`${property.replace(/-/g, '\\-')}:([^;}]*)`).exec(css);
-  if (!found) return null;
-  const value = found[1].trim();
+/*
+ * A bundle declares each role more than once: the mode it was built for sits on `:root`, and the
+ * opposite mode sits on the `dx-theme-mode-*` classes (see THEME_MODES.html). A literal baked into
```
Raushen: Dead link to the THEME_MODES.html

packages/devextreme/js/__internal/core/utils/swatch_container.ts
```diff
@@ -1,29 +1,94 @@
...
+];
+
+const getContainerClasses = (
+  $element: dxElementWrapper,
+  $viewport: dxElementWrapper,
+): string[] => {
+  const classes = scopeClasses($element);
+  // A scope the viewport already resolves to needs no container of its own: it would be a wrapper
+  // that repaints nothing, and one that measures nothing - callers reading the container as a
+  // geometric area (popup drag and resize) would be clamped to its zero height.
```
Raushen, line 43: This hazard is real beyond the case the guard covers: a popup owned by a `dx-theme-mode-*` scope
cannot be dragged at all.

`PopupPositionController._getDragResizeContainer()` returns `_$markupContainer` whenever a
viewport is defined (`popup_position_controller.ts:165-169`), so the drag boundary collapses onto
the spawned wrapper.

Not introduced here — I measured the same dead drag for a `dx-swatch-custom` island on this same
build. What changes is who reaches it: a swatch is niche, a mode scope is the ordinary way to use
this feature. The guard above skips the wrapper when the scope matches the viewport; the case the
feature exists for still gets one.

The wrapper carries custom properties, not geometry. Reading the box off the viewport rather than
off `_$markupContainer` would keep those apart. Resizing reads the same `_$dragResizeContainer` —
worth checking, my own resize run was inconclusive.

packages/devextreme/js/__internal/ui/overlay/overlay.ts
```diff
@@ -631,6 +635,39 @@ class Overlay<
...
+    if (!this._isVisible()) {
+      return;
+    }
+
+    // Move only if the scope named a different node: appending is not a no-op for a child already
+    // in place, and the detach takes the focus out, restarts animations and reloads any iframe.
+    const { $container } = this._positionController;
+
+    if ($container && $container.get(0) !== this._$wrapper?.parent().get(0)) {
+      this._moveToContainer();
```
Raushen: The line above names the detach as the reason not to move needlessly — but when the move does
happen, the focus is dropped and nothing puts it back.

So a user typing in a popup loses the caret the moment the application announces a mode change —
silently, with the text still on screen.

Either restore the focused element around `_moveToContainer()` here, or say so in the
`refreshMode()` description so an application can decide when it is safe to call.

packages/devextreme/js/__internal/ui/__tests__/themes.test.ts
```diff
@@ -0,0 +1,103 @@
...
+      refreshMode();
+      refreshMode();
+
+      expect(told).toHaveLength(2);
+    } finally {
+      themeModeChangedCallback.remove(subscriber);
+    }
+  });
+
+  it('stops telling a subscriber that unsubscribed', () => {
```
Raushen: This one passes with the production code deleted: empty out `refreshMode`, and `told` is still
`[]`. What it actually asserts is that `Callbacks.remove` unsubscribes — a core utility this PR
does not touch, already covered by `testing/tests/DevExpress.core/utils.callbacks.tests.js`
(add/remove/fire).

Suggest dropping this one.

packages/devextreme/js/__internal/core/utils/swatch_container.ts
EugeniyKiyashko: Confirmed, and I agree it is not new here. I measured it on this build, with a viewport that has a real size:

| popup sits in | area it may be dragged in | drag by 100,100 | resize by +60,+40 |
|---|---|---|---|
| nothing in particular | 600px tall | moves 100,100 | grows 60,40 |
| a theme-mode scope | 0px tall | does not move | width +60, height **-60** |
| `dx-swatch-custom` | 0px tall | does not move | width +60, height **-60** |

The swatch row and the mode row are identical, so this PR does not create the problem. What it changes is how many people meet it: putting a mode class on a block is the ordinary way to use this feature, while a swatch is a niche thing.

To answer the part you left open: resizing is affected too, and it shows up more plainly than dragging. The popup does not merely refuse to grow - its height collapses instead.

Why the suite never notices: the drag and resize block in `popup.tests.js` creates its popups with `dragAndResizeArea: window`, which sidesteps the element the popup was placed into. Those tests can't reach this path by construction.

I would rather not fix it inside this PR. The fix belongs where a popup decides which area it may be dragged in, it affects every popup whether or not any of this feature is used, and nothing else here goes near that code. Worth its own ticket - happy to raise it.

Raushen (approved):

